PR #17155, opened 2020-08-18
Media libs/partio

redchillipadi:


media-libs/partio/partio-1.10.1.ebuild
```diff
@@ -0,0 +1,48 @@
...
+RDEPEND="${PYTHON_DEPS}
+	media-libs/freeglut
+	sys-libs/zlib
+	virtual/opengl
+"
+DEPEND="${RDEPEND}"
+
+src_configure() {
+	local mycmakeargs=(
+		$(cmake_use_find_package doc Doxygen)
```
thesamesam, line 45: Yes, interestingly, that argument isn't here now ;)

media-libs/partio/partio-1.10.1.ebuild
```diff
@@ -0,0 +1,48 @@
...
+		dev-texlive/texlive-bibtexextra
+		dev-texlive/texlive-fontsextra
+		dev-texlive/texlive-fontutils
+		dev-texlive/texlive-latex
+		dev-texlive/texlive-latexextra
+	)
+"
+RDEPEND="${PYTHON_DEPS}
+	media-libs/freeglut
+	sys-libs/zlib
```
thesamesam, line 38: := on anything we link against which has a subslot

media-libs/partio/partio-1.1.0_p20170829.ebuild
```diff
@@ -1,10 +1,10 @@
 # Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
-PYTHON_COMPAT=( python3_6 )
+EAPI=7
+PYTHON_COMPAT=( python3_{7,8} )
```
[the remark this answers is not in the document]
redchillipadi: I can't find anything official on required versions upstream. I will test all ebuilds with python 3.6 to 3.9 and update accordingly.

media-libs/partio/partio-1.10.1.ebuild
redchillipadi: Looks like only zlib has a subslot, so I have added it for all the ebuilds. swig does not have a slot, so I have removed the :* in 1.1.0

media-libs/partio/partio-1.1.0_p20170829.ebuild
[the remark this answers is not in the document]
redchillipadi: All ebuilds for Partio compile with python 3.6 through 3.9

media-libs/partio/partio-1.10.1.ebuild
redchillipadi: See above, docdir can be removed.

media-libs/partio/partio-1.10.1.ebuild
```diff
@@ -0,0 +1,48 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8} )
```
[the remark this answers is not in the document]
redchillipadi: Done

media-libs/partio/partio-1.1.0_p20170829.ebuild
```diff
@@ -39,16 +39,16 @@ DEPEND="${RDEPEND}
...
+	cmake_src_prepare
 
 	sed -e '/ADD_SUBDIRECTORY (src\/tests)/d' -i CMakeLists.txt || die
 }
 
 src_configure() {
 	local mycmakeargs=(
-		$(cmake-utils_use_find_package doc Doxygen)
+		$(cmake_use_find_package doc Doxygen)
 		-DCMAKE_INSTALL_DOCDIR="share/doc/${PF}"
```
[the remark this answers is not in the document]
redchillipadi: The DOCDIR seems to work when the line is removed, but without the find_package doc Doxygen the docs are always installed regardless of whether the doc USE flag is set.

media-libs/partio/partio-1.1.0_p20170829.ebuild
```diff
@@ -39,16 +39,16 @@ DEPEND="${RDEPEND}
 PATCHES=( "${FILESDIR}/${PN}-1.1.0-Rename-partconv.patch" )
 
 src_prepare() {
-	cmake-utils_src_prepare
+	cmake_src_prepare
 
 	sed -e '/ADD_SUBDIRECTORY (src\/tests)/d' -i CMakeLists.txt || die
```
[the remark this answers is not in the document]
redchillipadi: Thanks, I have put cmake_src_prepare at the bottom of the function.

redchillipadi:
I don't think there will be any need to keep the old versions around here for long. Osl it the only package that depends on partio and it works with the lastest version so once everyone has had a chance to upgrade we should be able to drop everything before 1.10.

media-libs/partio/partio-1.1.0_p20170829.ebuild
```diff
@@ -1,10 +1,10 @@
 # Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
-PYTHON_COMPAT=( python3_6 )
+EAPI=7
+PYTHON_COMPAT=( python3_{6,7,8,9} )
 
-inherit cmake-utils vcs-snapshot python-single-r1
+inherit cmake vcs-snapshot python-single-r1
```
thesamesam, line 7: Sort?

media-libs/partio/partio-1.10.1.ebuild
thesamesam: Yep :)

media-libs/partio/partio-1.10.1.ebuild
```diff
@@ -0,0 +1,49 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{6,7,8,9} )
```
thesamesam, line 6: Just checking: the tests (if they exist) passed on all these, but if they do not exist, at least the program seemed to work (ideally with something using said bindings? or whatever)

redchillipadi: I checked that I could compile each version, and created test.bgeo using makeline (with one of the versions), and opened it using partview with each version. I can see a collection of points in a straight line. Some of the tests like makecircle complain when they use default values, but appear to give valid results when checked using partview.

Tests like testmerge and testcluster only pass if the source is still unpacked as they need src/data/*.bgeo to exist and need portage permissions. I think the tests should eventually be packaged to run only with FEATURES="test" and not be installed on the user's system.

media-libs/partio/partio-1.1.0_p20170829.ebuild
redchillipadi: Done

mgorny (approved):
Thank you. It looks like nice progress!

PR #17157, opened 2020-08-18
net-libs/libtelnet: bump to 0.23, fix QA issues

jsmolic:


net-libs/libtelnet/libtelnet-0.23.ebuild
```diff
@@ -0,0 +1,28 @@
...
+DESCRIPTION="Simple RFC-complient TELNET implementation as a C library"
+HOMEPAGE="https://github.com/seanmiddleditch/libtelnet"
+SRC_URI="https://github.com/seanmiddleditch/libtelnet/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="public-domain"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+src_prepare() {
+	_elibtoolize
```
thesamesam, line 17: This looks a bit odd. I reckon you can get away with simply `eautoreconf` here. We avoid calling these guys individually if we can help it.

net-libs/libtelnet/libtelnet-0.23.ebuild
```diff
@@ -0,0 +1,28 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit eutils autotools
```
thesamesam, line 6: Some background: eutils contained some things which ended up being included in various EAPIs.

So, sometimes you end up with some really common things that once needed it, and it ends up lingering around even though it's long-obsolete. Not needed here, and always good to keep an eye out for it!

(Also, we sort inherits).

net-libs/libtelnet/libtelnet-9999.ebuild
```diff
@@ -22,3 +22,8 @@ IUSE=""
 
 # needed unconditionally for man pages
 DEPEND="app-doc/doxygen"
```
thesamesam, line 24: BDEPEND, no?

net-libs/libtelnet/libtelnet-9999.ebuild
```diff
@@ -22,3 +22,8 @@ IUSE=""
```
thesamesam, line 22: Empty IUSE is not necessary

net-libs/libtelnet/libtelnet-9999.ebuild
```diff
@@ -1,9 +1,9 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-inherit cmake-utils
+inherit cmake
 
 if [[ ${PV} == *9999 ]]; then
```
thesamesam, line 8: Do you want to sync the 9999 with the 0.23 ebuild? That's why this if logic exists.

net-libs/libtelnet/libtelnet-0.23.ebuild
jsmolic: Yeah, I was unsure about this, `eautoreconf` works just fine :)

net-libs/libtelnet/libtelnet-0.23.ebuild
jsmolic: Thanks, it's removed

net-libs/libtelnet/libtelnet-9999.ebuild
jsmolic: Yep, I missed that

net-libs/libtelnet/libtelnet-9999.ebuild
jsmolic: Dropped

net-libs/libtelnet/libtelnet-0.23.ebuild
thesamesam: Perfect. Is it actually needed in the first place at all btw?

net-libs/libtelnet/libtelnet-9999.ebuild
jsmolic: I know :) Upstream switched from autotools to cmake after 0.23, but I left the if logic for future releases

net-libs/libtelnet/libtelnet-9999.ebuild
thesamesam: Np! I often find it's way easier to write things right myself rather than improving other peoples' ebuilds. It is a skill and it takes time, but you will get there (you also end up building up a list of things you want to fix usually!)

net-libs/libtelnet/libtelnet-9999.ebuild
thesamesam: I've got ya.

net-libs/libtelnet/libtelnet-0.23.ebuild
jsmolic: I think so, nothing is actually compiled if it's not called

net-libs/libtelnet/libtelnet-9999.ebuild
jsmolic: Yeah, I definitely agree ;)

PR #17158, opened 2020-08-18
games-emulation/dosbox-staging: add missing dep

AndrewAmmerlaan:
also fix dekstop file to adhere to theme specs

Bug: https://bugs.gentoo.org/733314

### DIFF
--- a/games-emulation/dosbox-staging/dosbox-staging-0.75.0.ebuild
+++ b/games-emulation/dosbox-staging/dosbox-staging-0.75.0.ebuild
@@ -16,7 +16,7 @@ IUSE="alsa debug dynrec opengl opus"
 RDEPEND="alsa? ( media-libs/alsa-lib )
 	debug? ( sys-libs/ncurses:0= )
 	opengl? ( virtual/opengl )
-	opus? ( media-libs/opus )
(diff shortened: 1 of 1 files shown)


games-emulation/dosbox-staging/dosbox-staging-0.75.0.ebuild
```diff
@@ -16,7 +16,10 @@ IUSE="alsa debug dynrec opengl opus"
 RDEPEND="alsa? ( media-libs/alsa-lib )
 	debug? ( sys-libs/ncurses:0= )
 	opengl? ( virtual/opengl )
-	opus? ( media-libs/opus )
+	opus? (
+		media-libs/opus
+		media-libs/opusfile
```
voyageur, line 21: Thanks for testing that! Looking at [configure.ac](https://github.com/dosbox-staging/dosbox-staging/blob/v0.75.0/configure.ac#L500) opusfile is actually the needed dep for dosbox-staging, opus itself comes as dependency of opusfile. So we can drop it from deps here

AndrewAmmerlaan: Fixed

PR #17159, opened 2020-08-18
app-arch/gzip: Various improvements

Whissi:
Various improvements:

- Migrate to EAPI=7
- Get rid of unusual USE=pic -- NO_ASM has no effect on most architectures so we should always set it. That's also Debian/Ubuntu and RHEL default.
- Set `UNALIGNED_OK` which should decrease (de)compression time by ~50%. Default in Debian/Ubuntu.

### DIFF
--- a/app-arch/gzip/gzip-1.10-r1.ebuild
+++ b/app-arch/gzip/gzip-1.10-r1.ebuild
@@ -1,0 +1,39 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+inherit flag-o-matic
+
+DESCRIPTION="Standard GNU compressor"
+HOMEPAGE="https://www.gnu.org/software/gzip/"
+SRC_URI="mirror://gnu/gzip/${P}.tar.xz
+	mirror://gnu-alpha/gzip/${P}.tar.xz
+	mirror://gentoo/${P}.tar.xz"
+
+LICENSE="GPL-3"
+SLOT="0"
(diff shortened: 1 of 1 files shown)


app-arch/gzip/gzip-1.10-r1.ebuild
```diff
@@ -0,0 +1,45 @@
...
+
+EAPI=7
+
+inherit flag-o-matic
+
+DESCRIPTION="Standard GNU compressor"
+HOMEPAGE="https://www.gnu.org/software/gzip/"
+SRC_URI="mirror://gnu/gzip/${P}.tar.xz
+	mirror://gnu-alpha/gzip/${P}.tar.xz
+	mirror://gentoo/${P}.tar.xz"
```
thesamesam, line 12: Is the mirror://gentoo a historical artefact here?

app-arch/gzip/gzip-1.10-r1.ebuild
```diff
@@ -0,0 +1,45 @@
...
+
+LICENSE="GPL-3"
```
thesamesam, line 14: GPL-3+

app-arch/gzip/gzip-1.10-r1.ebuild
Whissi: In the end it is, yes. Will get rid of it.

app-arch/gzip/gzip-1.10-r1.ebuild
Whissi: Thanks.

Polynomial-C:
:+1:

Polynomial-C:
:+1:

ajakk:
Seems obsoleted and nacked by soap

PR #17160, opened 2020-08-18
games-util/pyfa: version bump to 2.28.3

ZeroPointEnergy:
Simple version bump and cleanup of old version


ZeroPointEnergy:
Updated PR to newest version

ZeroPointEnergy:
Updated ebuild to newest version.

games-util/pyfa/pyfa-2.28.1.ebuild
```diff
@@ -0,0 +1,100 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="7"
+
+PYTHON_COMPAT=( python3_{6,7,8} )
```
thesamesam: Python 3.9 too, if tests pass?

games-util/pyfa/pyfa-2.28.1.ebuild
```diff
@@ -0,0 +1,100 @@
...
+PYTHON_REQ_USE="sqlite"
+
+inherit desktop eutils python-single-r1 xdg-utils
```
thesamesam: - May be dos2unix and possibly other classes now, not eutils (a lot of stuff got moved out of there recently, see the gentoo-dev mailing list)
- Use xdg if you can, may even be able to drop the postinst and prerm definitions below.

games-util/pyfa/pyfa-2.28.1.ebuild
```diff
@@ -0,0 +1,100 @@
...
+src_install() {
+	python_moduleinto ${PN}
+	python_domodule eos gui service utils graphs
+	python_domodule config*.py __init__.py version.yml configforced.py db_update.py
+	python_doscript pyfa
+
+	insinto /usr/share/${PN}
+
+	einfo "Creating database ..."
+	python ./db_update.py || die
```
thesamesam: ${EPYTHON} or ${PYTHON}

games-util/pyfa/pyfa-2.28.1.ebuild
ZeroPointEnergy: I will look into this once I get emerge to tell me what the actual issue is when it blocks the install. this is what I currently get:
```
emerge: there are no ebuilds to satisfy ">=dev-python/pyyaml-5.1[python_targets_python3_9(-)]".
(dependency required by "games-util/pyfa-2.28.1::devel[python_single_target_python3_9]" [ebuild])
(dependency required by "pyfa" [argument])
```
That ebuild does have python 3.9 support though :man_shrugging:

games-util/pyfa/pyfa-2.28.1.ebuild
ZeroPointEnergy: replaced with xdg class and dropped the now redundant functions

games-util/pyfa/pyfa-2.28.1.ebuild
ZeroPointEnergy: fixed

games-util/pyfa/pyfa-2.28.1.ebuild
ZeroPointEnergy: There are at least two dependencies that have not yet the python 3.9 support. wxpython and logbook. Will make a PR for logbook.

ZeroPointEnergy:
PR updated to newest version

ZeroPointEnergy:
I would like to quickly discuss if the way this ebuild is maintained is still a good fit.

The utility is only really useful in its latest version as the rules and game DB it contains has to be in sync with the live version of EVE Online. As you can see from the release page here (https://github.com/pyfa-org/Pyfa/tags) there have been 6 newer releases since I opened this PR alone.

I try really hard to update the ebuild as close to upstream as possible. But since every PR takes months I have to constantly rebase it with the newest version and additionally maintain an overlay with the latest ebuilds (https://github.com/ZeroPointEnergy/gentoo-pyfa-overlay). Because of this, probably no one will ever use the version we curate in the Gentoo repository. The only thing this may cause is for users to question why the version is so ancient and updating it takes so long.

I really appreciate the feedback and help you people provide to me to improve the ebuild and I would like to continue to improve my ebuild crafting skills. I just think that this particular ebuild because of the nature of the software it packages is probably a bad fit. Would it be better if we simply move this to GURU so I can maintain it there and I pick another ebuild I can maintain via the proxy maintainer workflow?

juippis:
I understand the current situation, it is rather uncomfortable.

I've been really overworked the past few months now, but it's starting to look better. If we can process the queue, then we should be back at dealing all PRs in 2-3 days max. I checked upstream, they seem to make new release every 5-7 days on average. 

Putting it to GURU definitely makes it a lot easier for you and us, but if you want to see it in ::gentoo repo, give us a couple more weeks to deal with the current work queue. Even after refusing to take new packages, we seem to have more work and less manpower which is a bit confusing... So we can't promise a similar situation does not happen again, and the work queue stretching to +7 days again.

It's up to you. GURU is a great alternative.

ZeroPointEnergy:
@juippis thanks for your answer. It is fine by me if we keep it in Gentoo if you think the queue times will be lowered. We can always move it if it turns out the queue times didn't get lower.

If there is anything I can do as a non-dev to help out feel free to point me in the direction :-)

mgorny:
You can also look for a dedicated dev to process your requests. Maybe somebody from games team?

juippis (approved):

PR #17162, opened 2020-08-18
net-misc/ncp: remove diet use flag, fix bug #737254

mm1ke:
Hi,

this is a simple update to r1 which just updates the sed command to not use colons in order to avoid problems with `-falign-functions=32:25:16`. 

```diff
1c1
< # Copyright 1999-2019 Gentoo Authors
---
> # Copyright 1999-2020 Gentoo Authors
23c23
<       sed -e '/^ncp:/,+5s:strip:#strip:' \
---
>       sed -e '/^ncp:/,+5s|strip|#strip|' \
```


mm1ke:
Please do not merge yet, i want to add another fix for bug: https://bugs.gentoo.org/737254

SoapGentoo:
1. Please use a patch instead of a `sed`
2. Just remove the `diet` USE flag. I'm not convinced by adding special flags to appease someone like fefe, and I don't see the massive win here.

mm1ke:
> 1. Please use a patch instead of a `sed`
Sure, done.
> 2. Just remove the `diet` USE flag. I'm not convinced by adding special flags to appease someone like fefe, and I don't see the massive win here.
Sure, this makes for me easier too :)

net-misc/ncp/files/ncp-1.2.4-fix-makefile.patch
```diff
@@ -0,0 +1,11 @@
+--- ncp-1.2.4/GNUmakefile	2005-05-13 19:17:17.000000000 +0200
++++ ncp-1.2.4/GNUmakefile	2020-08-18 18:11:57.113262402 +0200
+@@ -54,7 +54,7 @@
+ ncp: ncp.o libsocket
+ 	$(CC) $< -o $@ -I. $(CFLAGS) $(LDFLAGS) $(LDLIBS) `cat libsocket`
+ ifeq ($(DEBUG),)
+-	strip -R .note -R .comment ncp || strip ncp
++	#strip -R .note -R .comment ncp || strip ncp
```
thesamesam, line 8: No need to comment out with a patch - we can see what we removed.

net-misc/ncp/files/ncp-1.2.4-fix-makefile.patch
```diff
@@ -0,0 +1,11 @@
+--- ncp-1.2.4/GNUmakefile	2005-05-13 19:17:17.000000000 +0200
```
thesamesam, line 1: I usually like to put a [header](https://dev.gentoo.org/~vapier/clean-patches) with a Gentoo bug ref at least.

net-misc/ncp/files/ncp-1.2.4-fix-makefile.patch
mm1ke: Sorry, i don't understand? I shouldn't use a patch here? (and use `sed` instead again?)

mm1ke:
@thesamesam 
Just rebased and updated the patch with a header. I still not sure about your first comment, please have a look :)

net-misc/ncp/files/ncp-1.2.4-fix-makefile.patch
thesamesam: Sorry, what I'm saying is... if you remove a line with a patch, you can see the missing line, right?

There's no need to put a '#' in, just drop the line, because we still get to see what the old line was. :)

net-misc/ncp/files/ncp-1.2.4-fix-makefile.patch
thesamesam: ping

net-misc/ncp/files/ncp-1.2.4-fix-makefile.patch
mm1ke: ahh.. i understand now :)
i've updated the patch, now it removes the lines (and also the `ifeq .. endif` since it would be empty anyway)

net-misc/ncp/files/ncp-1.2.4-fix-makefile.patch
mm1ke: I've updated the header, but didn't add a bug reference since i couldn't find one. Hope thats enough.

mm1ke:
@thesamesam @SoapGentoo 
Would be nice if anyone could push this :)
Thanks.

thesamesam:
Done, thank you!

PR #17167, opened 2020-08-19
sys-libs/zlib: UNALIGNED_OK

thesamesam:
This is just a test PR, although we did discuss the possible change in future, for Whissi (so not yet tested).

If we really want to do this, we should tidy up the ebuild some more and test extensively with benchmarks. There's a proposal for this with gzip: #17159.


mgorny:
Don't forget `-O99999`.

SoapGentoo:
Closing this for QA reasons. While Debian seems to be getting away with it, unaligned access invokes undefined behaviour ([yes, even on amd64 and x86](https://pzemtsov.github.io/2016/11/06/bug-story-alignment-on-x86.html)) and it's only matter of time until GCC/Clang optimize around this. This is a security disaster waiting to happen.

PR #325, opened 2015-11-03
app-admin/lastpass: bump to 3.2.16 #550834 #551312

gktrk:
Bump app-admin/lastpass to 3.2.16. Upstream does not support versioned
tarballs, so remove the old version. Closes a bump request and a
checksum mismatch bug.

Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=550834
Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=551312

Package-Manager: portage-2.2.20.1

PR #17168, opened 2020-08-19
sox-fmt: validate comments_bytes before use

judyhsiao2020:
(CVE-2019-13590) [bug #325]
Cap the comments size to 1 GB to avoid overflows in subsequent
arithmetic.

The missing null check mentioned in the bug report is bogus since
lsx_calloc() returns a valid pointer or aborts.

patch from:
https://sourceforge.net/p/sox/code/ci/7b6a889217d62ed7e28188621403cc7542fd1f7e/

Bug: https://bugs.gentoo.org/711320
Signed-off-by: judyhsiao@chromium.org


thesamesam:
By the way, I did see your comment on the bug -- just hadn't got a chance to check if it was needed or not yet (to give you a useful reply). I'll handle this if it applies.. I'm surprised this wasn't in the snapshot though?

I will need to check the tag and see what happened.

scardracs:
Ping, is this PR still alive?

ajakk:
Seems like this patch is in tree via the snapshot. Sorry this didn't get tended to sooner!

PR #17170, opened 2020-08-19
dev-haskell/optparse-simple: bump, remove old

solpeth:
Tested in an ~amd64 container.


trofi:
Pulled as is.

Thank you!

PR #17171, opened 2020-08-19
dev-qt/qtwebengine: fix build with >=bison-3.7

Chiitoo:
Tested building 5.15.0 only via 'ebuild' while everything else was at 5.15.9999 (with qtwebengine-5.15.9999 installed as well).


a17r (approved):
Works for me. Thanks!

PR #17178, opened 2020-08-20
Haskell: bump yaml and unicode-transforms, remove old ebuilds

solpeth:
Specifically:

- Remove old versions of `pandoc-citeproc`, `yaml`, `pandoc` and `unicode-transforms`
- Bump `unicode-transforms` and `yaml`.


trofi:
Pulled as is.

Thank you!

PR #17181, opened 2020-08-20
dev-python/jedi: correct parso dependency

AndrewAmmerlaan:
this is how upstream specifies it

Closes: https://bugs.gentoo.org/738178


mgorny:
I'm sorry, I've started working on a fix as soon as you've found the culprit.

AndrewAmmerlaan:
> I'm sorry, I've started working on a fix as soon as you've found the culprit.

That's alright, the important thing is that it's fixed now :smile:

PR #17184, opened 2020-08-20
app-admin/doas: add ~ppc64 keyword

djames1:
Tested on both PPC64 Big and Little Endian.


juippis:
https://github.com/gentoo/gentoo/pull/17186#issuecomment-678065767

PR #17188, opened 2020-08-20
x11-plugins/wmsystemtray: add ~ppc64 keyword

djames1:


x11-plugins/wmsystemtray/wmsystemtray-1.4-r2.ebuild
```diff
@@ -9,7 +9,7 @@ SRC_URI="mirror://sourceforge/${PN}/${P}.tar.gz"
 
 LICENSE="GPL-2"
 SLOT="0"
-KEYWORDS="~amd64 ~x86"
+KEYWORDS="~amd64 ~x86 ~ppc64"
```
voyageur, line 12: CI check fails as this is not sorted else ready to go in! (with added sign-off, same for PR 17189)

djames1: @voyageur Fixed issue with KEYWORDS not being sorted

PR #17192, opened 2020-08-20
sys-apps/texinfo: add missing help2man dependency

matijaskala:


mattst88:
Can you file a bug on bugzilla with a log showing the build failure, etc?

matijaskala:
Actually, on the second try, it compiles fine without help2man.

PR #17198, opened 2020-08-21
media-video/ffmpeg-4.2.4: Make hardware accel work better with 390-series NVIDIA cards

jeeger:
ffmpeg-4.2.4 doesn't actually require nv-codec-headers > 9, but the in-tree version of nv-codec-headers is too low to be compatible with ffmpeg-4.2.4. I've thus bumped the nv-codec-headers ebuild, and removed the version requirement on nv-codec-headers from the ffmpeg ebuild.

I'm happily encoding video here on my 390-series card.


candrews:
media-libs/nv-codec-headers was bumped in 495ee31dbe2ff3c1fab778889f39e347474e69fd to a later version, so I believe this PR can be closed.

Thank you!

PR #17206, opened 2020-08-21
net-analyzer/nikto: version bump to 2.1.6_p20200807

CaseOf:
Package is set for removal but I saw that it is still active upstream.
I updated ebuild then. Maybe will I maintain this package later.


juippis:
734a83031a6028c1dc508e0d7e0b140e71bc7134

CaseOf:
What should I say?… That’s it’s worthless to contribute?

thesamesam:
No, I forgot about your PR. If you are still interested in this package, I guess we can return it.

juippis:
Hi,

of course contributions aren't worthless, but please understand that we can't see everything. If a deadline is approaching to fix a masked package, you should definitely try to contact us directly via IRC or by e-mail. We get hundreds of Github notifications daily.

Another tip I can give for the next time is you can / should've used `Bug: https://bugs.gentoo.org/658448` tag in your commit message footer, so sam would've gotten a notification through bug where the removal of this package was handled. Last-rited packages should always have a bug opened, where the person removing it is involved. Of course these are things not always so obvious, but as I said, for the next time :) 

Nothing personal here, I was mass-closing "outdated" PRs.

PR #17215, opened 2020-08-22
dev-python/influxdb: fix FutureWarning

Hawk777:


Hawk777:
3bf197413f3c rebases and adds Python 3.8 as a supported target. No 3.9 for now because dependency `dev-python/msgpack` doesn’t support it yet.

thesamesam:
@Hawk777 It does now! ;)

dev-python/influxdb/files/5.3.0-pandas-future-warning.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/influxdb/_dataframe_client.py b/influxdb/_dataframe_client.py
```
thesamesam, line 1: Pop the link to the PR at the top of the file (I can do this on merge though).

dev-python/influxdb/influxdb-5.3.0-r1.ebuild
```diff
@@ -30,6 +30,8 @@ BDEPEND="test? (
 		dev-python/requests-mock[${PYTHON_USEDEP}]
 	)"
 
+PATCHES=( "${FILESDIR}/${PV}-pandas-future-warning.patch" )
```
thesamesam, line 33: Normally we prefix name files as: ${PN}-VERSION-DESCRIPTION.patch, where VERSION is the version it was based on.

We will use ${P} in the ebuild if it's an upstreamed patch (it's just a little hint), but ${PN}-5.3.0-foo.patch for example if it's not upstreamed but based on 5.3.0.

dev-python/influxdb/influxdb-5.3.0-r1.ebuild
```diff
@@ -30,6 +30,8 @@ BDEPEND="test? (
 		dev-python/requests-mock[${PYTHON_USEDEP}]
```
thesamesam, line 30: With distutils_enable_tests, there's no need to put ${RDEPEND} in test? or to add nose as a testdep. It's very cconvenient!

dev-python/influxdb/influxdb-5.3.0-r1.ebuild
```diff
@@ -2,7 +2,7 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
-PYTHON_COMPAT=( python3_{6,7} )
```
thesamesam, line 5: One blank line after EAPI as per the [skeleton](https://gitweb.gentoo.org/repo/gentoo.git/tree/skel.ebuild) ebuild (or see the proxy-maint style [guide](https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide/Style_Guide)).

thesamesam (changes requested):

Hawk777:
> @Hawk777 It does now! ;)

3.9 added!

dev-python/influxdb/influxdb-5.3.0-r1.ebuild
Hawk777: Done (by way of explanation, it was like this before; I just copied the old ebuild).

dev-python/influxdb/influxdb-5.3.0-r1.ebuild
Hawk777: Done.

dev-python/influxdb/influxdb-5.3.0-r1.ebuild
Hawk777: Done.

dev-python/influxdb/files/5.3.0-pandas-future-warning.patch
Hawk777: Done.

thesamesam:
Thanks!

PR #17228, opened 2020-08-23
dev-python/pydotplus: PYTHON_COMPAT and EAPI update

TheChymera:
Bug: https://bugs.gentoo.org/695996

### DIFF
--- a/dev-python/pydotplus/files/pydotplus-2.0.2-tests.patch
+++ b/dev-python/pydotplus/files/pydotplus-2.0.2-tests.patch
@@ -1,0 +1,17 @@
+--- a/test/pydot_unittest.py
++++ b/test/pydot_unittest.py
+@@ -142,6 +142,7 @@ class TestGraphAPI(unittest.TestCase):
+         self.assertEqual(g2.get_edges()[0].get_source(), node1)
+         self.assertEqual(g2.get_edges()[0].get_destination(), node2)
+
++    @unittest.skip(reason="Known to fail on Gentoo, reported upstream: https://github.com/carlos-jenkins/pydotplus/issues/22")
+     def test_graph_with_shapefiles(self):
+
+         shapefile_dir = os.path.join(TEST_DIR, 'from-past-to-future')
+@@ -225,6 +226,7 @@ class TestGraphAPI(unittest.TestCase):
+     def test_my_regression_tests(self):
+         self._render_and_compare_dot_files(MY_REGRESSION_TESTS_DIR)
+
++    @unittest.skip(reason="Known to fail on Gentoo, reported upstream: https://github.com/carlos-jenkins/pydotplus/issues/22")
+     def test_graphviz_regression_tests(self):
+         self._render_and_compare_dot_files(REGRESSION_TESTS_DIR)
--- a/dev-python/pydotplus/pydotplus-2.0.2.ebuild
+++ b/dev-python/pydotplus/pydotplus-2.0.2.ebuild
@@ -1,38 +1,32 @@
 # Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
 
-PYTHON_COMPAT=( python3_6 )
+PYTHON_COMPAT=( python3_{7,8} )
 
 inherit distutils-r1
 
 DESCRIPTION="Improved version of the old pydot project"
-HOMEPAGE="http://pydotplus.readthedocs.org/"
+HOMEPAGE="https://pydotplus.readthedocs.org/"
 SRC_URI="mirror://pypi/p/${PN}/${P}.tar.gz"
 
 LICENSE="MIT"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE=""
+IUSE="test"
+RESTRICT="!test? ( test )"
 
-DEPEND="dev-python/setuptools[${PYTHON_USEDEP}]"
-#	test? (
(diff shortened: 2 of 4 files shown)


dev-python/pydotplus/pydotplus-2.0.2.ebuild
```diff
@@ -14,25 +14,22 @@ SRC_URI="mirror://pypi/p/${PN}/${P}.tar.gz"
...
-#		dev-python/sphinx[${PYTHON_USEDEP}]
-#		dev-python/sphinx_rtd_theme[${PYTHON_USEDEP}]
-#		dev-python/tox[${PYTHON_USEDEP}]
-#		)
+IUSE="test"
+RESTRICT="!test? ( test )"
+
+DEPEND="dev-python/setuptools[${PYTHON_USEDEP}]
+	test? (
+		dev-python/pytest[${PYTHON_USEDEP}]
```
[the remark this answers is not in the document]
TheChymera: like so? seems strange to me to use `distutils_enable_tests pytest` when right afterwards I need to define the test phase to be something different from the pytest standard. What “duplication” are you referring to?

dev-python/pydotplus/files/pydotplus-2.0.2-tests.patch
```diff
@@ -0,0 +1,26 @@
+--- a/test/pydot_unittest.py
++++ b/test/pydot_unittest.py
+@@ -18,7 +18,7 @@ import sys
+
+ import pydotplus
+ import unittest
+-
++import pytest
```
mgorny, line 8: Sorry I didn't notice this before. So you're actually adding dependency on pytest? Why aren't you just using unittest's functions?

dev-python/pydotplus/pydotplus-2.0.2.ebuild
```diff
@@ -1,38 +1,32 @@
...
-DEPEND="dev-python/setuptools[${PYTHON_USEDEP}]"
-#	test? (
-#		dev-python/flake8[${PYTHON_USEDEP}]
-#		dev-python/pytest[${PYTHON_USEDEP}]
-#		dev-python/pytest-cov[${PYTHON_USEDEP}]
-#		dev-python/sphinx[${PYTHON_USEDEP}]
-#		dev-python/sphinx_rtd_theme[${PYTHON_USEDEP}]
-#		dev-python/tox[${PYTHON_USEDEP}]
-#		)
+IUSE="test"
```
mgorny, line 17: Redundant.

dev-python/pydotplus/pydotplus-2.0.2.ebuild
```diff
@@ -1,38 +1,32 @@
...
-#the required files:
-#https://github.com/carlos-jenkins/pydotplus/issues/12
-#python_test() {
-#	${EPYTHON} -m unittest discover || die
-#	tox
-#}
+PATCHES=( "${FILESDIR}/${P}-tests.patch" )
+
+distutils_enable_tests pytest
+python_test(){
```
mgorny, line 28: Empty line above, space after `()`.

mgorny (changes requested):

TheChymera:
@mgorny ok

dev-python/pydotplus/pydotplus-2.0.2.ebuild
```diff
@@ -1,38 +1,29 @@
...
-#	test? (
-#		dev-python/flake8[${PYTHON_USEDEP}]
-#		dev-python/pytest[${PYTHON_USEDEP}]
-#		dev-python/pytest-cov[${PYTHON_USEDEP}]
-#		dev-python/sphinx[${PYTHON_USEDEP}]
-#		dev-python/sphinx_rtd_theme[${PYTHON_USEDEP}]
-#		dev-python/tox[${PYTHON_USEDEP}]
-#		)
+
+DEPEND=""
```
mgorny, line 18: Remove the empty var.

dev-python/pydotplus/pydotplus-2.0.2.ebuild
```diff
@@ -1,38 +1,29 @@
...
-#python_test() {
-#	${EPYTHON} -m unittest discover || die
-#	tox
-#}
+PATCHES=( "${FILESDIR}/${P}-tests.patch" )
+
+python_test() {
+	pushd test > /dev/null || die
+		python pydot_unittest.py || die
+	popd > /dev/null
```
mgorny, line 28: ||die

mgorny (commented):
Are you *sure* this package doesn't need to depend on graphviz? Also, you are referencing the wrong bug.

dev-python/pydotplus/pydotplus-2.0.2.ebuild
TheChymera: ok

dev-python/pydotplus/pydotplus-2.0.2.ebuild
TheChymera: most ebuilds have it without die (232 without, 217 with):

```
chymera@silenthost ~/src/pydotplus $ qgrep "popd > /dev/null" | wc -l
449
chymera@silenthost ~/src/pydotplus $ qgrep "popd > /dev/null || die" | wc -l
217
```

Why is this required here? doesn't the dir just change to the correct one after the function exits anyway?

TheChymera:
@mgorny ping?

TheChymera:
@mgorny ping?

mgorny (approved):
Thanks.

PR #17239, opened 2020-08-23
net-mail/getmail: New upstream for package, version 6.9

Remmy:
getmail was removed in commit f63396228c129fdbac0b2b6bf90ca0dbe64bfb08 on
July 27 for being a Py2 only package. This reintroduces this package from
a different upstream that is compatible with Python 3.

Signed-off-by: Remco Rijnders <r+gentoo@r78.nl>


AndrewAmmerlaan:
You should add a sign-off line to your commit message, you can get repoman to do this automatically by setting the following in make.conf:

```
PORTAGE_GPG_KEY='KEY-FINGERPRINT'
DCO_SIGNED_OFF_BY='Your Real Name <someone@example.com>'
```

See [here](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin) for more info

Otherwise this looks good :+1:

Remmy:
Thanks Andrew!

This is probably more of a GitHub than a Gentoo question, but how do I amend my last commit? Do I have to add an extra commit on top of this, or overwrite my existing commit and open a new PR? In other words, what is my next step now to correct this situation?

AndrewAmmerlaan:
> Thanks Andrew!
> 
> This is probably more of a GitHub than a Gentoo question, but how do I amend my last commit? Do I have to add an extra commit on top of this, or overwrite my existing commit and open a new PR? In other words, what is my next step now to correct this situation?

`git --amend` allows you to amend the last commit message (you'll have to add the sign-off line manually)
If there had been more then one commit message, you'd have to use `git rebase git rebase -i HEAD~<number of commits to go back>` and change the word before your commit to `reword` and then edit the commit message.
Afterwards, you'll have to force push, by adding the `-f` option to your push command.

I highly recommend reading up on interactive git rebasing, you'll need it a lot for Pull Requests, as it allows you to drop or squash commits, or change the order of commits. I found rebasing quite difficult when I first started opening PR's, but it does get easier with practice.

Remmy:
Thanks Andrew, I was missing the -f, the force wasn't with me.

AndrewAmmerlaan:
> Thanks Andrew, I was missing the -f, the force wasn't with me.

Great, next you can append `[please reassign]` to this PR's title, to get the bot to re-evaluate this PR. That should make it remove the `no-signoff` tag.

Also, as I understand it, python3.6 is going to be removed at the end of the year: https://wiki.gentoo.org/wiki/Project:Python/Implementations
And as python3.7 is already default, and the Python Team is going through great efforts to add python3.7 and 3.8 support to all the packages, I'm pretty sure the Python Team would prefer to not add python3.6 compat to new packages. So I would suggest to have `PYTHON_COMPAT=( python3_{7,8} )` instead in this ebuild.

I also just tried emerging this ebuild, and I got this warning:
```
* DISTUTILS_USE_SETUPTOOLS value is probably incorrect
*   value:    DISTUTILS_USE_SETUPTOOLS=bdepend (default?)
*   expected: DISTUTILS_USE_SETUPTOOLS=no
```
This indicates that the ebuild should not depend on setuptools. In order to do that, you should add `DISTUTILS_USE_SETUPTOOLS=no` before inheriting the distutils-r1 eclass.

Remmy:
Andrew, thanks again. I've made the requested/suggested changes. I really appreciate you taking the time to review this. Hartelijk dank!

AndrewAmmerlaan:
> Andrew, thanks again. I've made the requested/suggested changes. I really appreciate you taking the time to review this. Hartelijk dank!

No problem, Thank you for re-adding this package to ::gentoo.

@juippis this looks ready to merge IMO :smile:

AndrewAmmerlaan (approved):

net-mail/getmail/getmail-6.03.ebuild
```diff
@@ -0,0 +1,27 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+PYTHON_COMPAT=( python3_{7,8} )
```
juippis, line 5: Put one empty line below EAPI.

net-mail/getmail/getmail-6.03.ebuild
```diff
@@ -0,0 +1,27 @@
...
+DISTUTILS_USE_SETUPTOOLS=no
+inherit distutils-r1
+
+DESCRIPTION="A mail retriever with reliable Maildir and mbox delivery"
+HOMEPAGE="http://www.getmail6.org/"
+SRC_URI="https://github.com/getmail6/getmail6/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc ~x86 ~ppc-macos ~x86-macos ~x64-solaris"
```
juippis, line 17: Let's drop the prefix keywords for now, 
```suggestion
KEYWORDS="~amd64 ~ppc ~x86"
```

net-mail/getmail/getmail-6.03.ebuild
```diff
@@ -0,0 +1,27 @@
...
+IUSE="ssl"
```
juippis, line 18: This USE flag seems to have no function in this ebuild.

net-mail/getmail/getmail-6.03.ebuild
```diff
@@ -0,0 +1,27 @@
...
+
+S="${WORKDIR}/getmail6-${PV}"
+
+python_prepare_all() {
+	sed -i -e "s,'getmail-%s' % __version__,'${PF}'," \
```
juippis, line 23: Please add a comment above this `sed` call to describe what it does. Apparently just renames doc directory / doc file?

juippis (commented):
Yep, looks good! Just few style errors I can't fix for you due to our copyright rules.

vaukai:
> You should add a sign-off line to your commit message, you can get repoman to do this automatically by setting the following in make.conf:
> 
> ```
> PORTAGE_GPG_KEY='KEY-FINGERPRINT'
> DCO_SIGNED_OFF_BY='Your Real Name <someone@example.com>'
> ```
> 
> See [here](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin) for more info
> 
> Otherwise this looks good +1



> PORTAGE_GPG_KEY='KEY-FINGERPRINT'
> DCO_SIGNED_OFF_BY='Your Real Name <someone@example.com>'

Both these lines in make.conf are not relevant for PRs in Github. 
("SIGNED_OFF_BY" still seems needed for GURU, but not in Github)
For Github PRs you want [user.signingkey](https://docs.github.com/en/github/authenticating-to-github/telling-git-about-your-signing-key) in "git config --local -l"

AndrewAmmerlaan:
> Both these lines in make.conf are not relevant for PRs in Github.
> ("SIGNED_OFF_BY" still seems needed for GURU, but not in Github)
> For Github PRs you want [user.signingkey](https://docs.github.com/en/github/authenticating-to-github/telling-git-about-your-signing-key) in "git config --local -l"

Those two variables are not relevant to either GitHub or GURU/Gentoo gitweb.gentoo.org, they are only relevant for repoman. The first line tells repoman which key to use (you can indeed accomplish the same with user.signkey), the second line adds `Signed-off-by: Your Real Name <someone@example.com>` to all commit messages.

net-mail/getmail/getmail-6.03.ebuild
AndrewAmmerlaan: I think the USE flag toggles a dependency on `dev-lang/python[ssl]` with `PYTHON_REQ_USE="ssl?"` in line 6, right?

thesamesam:
@Remmy ping?

Remmy:
Apologies. I will look into the open comments/suggestions and revert this week, also updating to the latest upstream version.

net-mail/getmail/getmail-6.03.ebuild
zpuskas: I have to agree with @juippis here. The SSL use flag does not seem to change how getmail is built. In fact by looking at the code we can see getmail implicitly expects SSL to be available and it cannot be selectively disabled.

So we should drop the `ssl` USE flag and on line 6 we just explicitly require SSL enabled Python.

thesamesam:
Thanks!

net-mail/getmail/getmail-6.03.ebuild
zpuskas, line 5: Could we also add Python 3.9 support here? Looking at https://www.python.org/downloads/release/python-390/ it does not immediately seem there are breaking changes between 3.8 and 3.9 (Unicode update might be interesting though), nor does getmail6 home page specify limits on Python3 compatibility.

Since it's a testing package anyway, I'd say let people try it on higher versions of Python.

zpuskas:
> 
> I also just tried emerging this ebuild, and I got this warning:
> 
> ```
> * DISTUTILS_USE_SETUPTOOLS value is probably incorrect
> *   value:    DISTUTILS_USE_SETUPTOOLS=bdepend (default?)
> *   expected: DISTUTILS_USE_SETUPTOOLS=no
> ```
> 
> This indicates that the ebuild should not depend on setuptools. In order to do that, you should add `DISTUTILS_USE_SETUPTOOLS=no` before inheriting the distutils-r1 eclass.

Interesting, when I tried emerging the ebuild it suggested me the following:

```
 * DISTUTILS_USE_SETUPTOOLS value is probably incorrect
 *   have:     DISTUTILS_USE_SETUPTOOLS=no
 *   expected: DISTUTILS_USE_SETUPTOOLS=bdepend
 * Verifying compiled files in /usr/lib/python3.7/site-packages
 * Verifying compiled files in /usr/lib/python3.8/site-packages
```

Explicitly specifying `DISTUTILS_USE_SETUPTOOLS=bdepend` removes all warnings. However I've slightly modified the ebuild, namely removed IUSE and also upgraded version to 6.8.

net-mail/getmail/getmail-6.03.ebuild
thesamesam: The maintainer should test it, but yeah, it's likely to work.

thesamesam:
@zpuskas You may want to make a new pull request in that case.

net-mail/getmail/metadata.xml
```diff
@@ -0,0 +1,12 @@
...
+<pkgmetadata>
+  <maintainer type="person">
+    <email>r+gentoo@r78.nl</email>
+    <name>Remco Rĳnders</name>
+  </maintainer>
+  <maintainer type="project">
+    <email>proxy-maint@gentoo.org</email>
+    <name>Proxy Maintainers</name>
+  </maintainer>
+</pkgmetadata>
```
zpuskas, line 12: I'd suggest adding more metadata about the package, namely:

```
  <upstream>
    <maintainer status="active">
      <email>roland.puntaier@gmail.com</email>
      <name>Roland Puntaier</name>
    </maintainer>
    <bugs-to>https://github.com/getmail6/getmail6/issues</bugs-to>
    <remote-id type="github">getmail6/getmail6</remote-id>
  </upstream>
```

zpuskas:
@thesamesam I'd like to give the opportunity to @Remmy to update his PR and get it merged, unless he lost interest in finishing it. Until this gets updated and merged into the main tree, I made net-mail/getmail-6.8 ebuild available in my overlay for testing purposes.

Remmy:
>@zpuskas commented on this pull request.
>
>```
>  <upstream>
>    <maintainer status="active">
>      <email>roland.puntaier@gmail.com</email>
>      <name>Roland Puntaier</name>
>    </maintainer>
>    <bugs-to>https://github.com/getmail6/getmail6/issues</bugs-to>
>    <remote-id type="github">getmail6/getmail6</remote-id>
>  </upstream>
>```

Thank you! I will incorporate this. Updated pull request will come today, I am
still interested but got sidetracked a bit, my apologies. I really appreciate
all the feedback I've been given so far by everyone.

Remmy:
>@zpuskas commented on this pull request.
>
>
>
>> @@ -0,0 +1,27 @@
>+# Copyright 1999-2020 Gentoo Authors
>+# Distributed under the terms of the GNU General Public License v2
>+
>+EAPI=7
>+PYTHON_COMPAT=( python3_{7,8} )
>
>Could we also add Python 3.9 support here? Looking at https://www.python.org/downloads/release/python-390/ it does not immediately seem there are breaking changes between 3.8 and 3.9 (Unicode update might be interesting though), nor does getmail6 home page specify limits on Python3 compatibility.
>
>Since it's a testing package anyway, I'd say let people try it on higher versions of Python.

I agree. Also, getmail6 was started explicitly to make getmail work with newer
versions of Python. As such, any incompatibilities are to be considered bugs and
upstream will definately be interested in fixing them. The sooner we run into
these issues (if any), the better.

As a n00b to gentoo packaging, I am a little unsure on how to do this testing
myself with multiple python versions going.

net-mail/getmail/getmail-6.03.ebuild
Remmy: I have a question about this. The sed call (on line 24) deletes the COPYING file. Is this per a gentoo policy to avoid a million copies from being installed? If so, where can I find it? The code refers to the COPYING files in many places, so I wonder if leaving this deletion in isn't actually going to cause more confusion over just leaving it in.

I have "recycled" the ebuild file that was used for the getmail 5.14 package when it was still included in the distribution and used that as a starting point.

juippis (approved):

juippis:
2df6b0f31a5a13fe8d1f5ec0385951bc18c1bf22

PR #17242, opened 2020-08-23
media-gfx/gimp: 9999, change SLOT from 2 to 2.99

band-a-prend:
Depends on gtk+-3 and python-3 so SLOT is required to be changed.

Gimp-2.99.2 developer preview release is expected
before Gimp-3.0 release.

Signed-off-by: Sergey Torokhov <torokhov-s-a@yandex.ru>

### DIFF
--- a/media-gfx/gimp/gimp-9999.ebuild
+++ b/media-gfx/gimp/gimp-9999.ebuild
@@ -15,7 +15,7 @@ HOMEPAGE="https://www.gimp.org/"
 EGIT_REPO_URI="https://gitlab.gnome.org/GNOME/gimp.git"
 SRC_URI=""
 LICENSE="GPL-3 LGPL-3"
-SLOT="2"
+SLOT="2.99"
 KEYWORDS=""
 
 IUSE="aalib alsa aqua debug doc gnome heif javascript jpeg2k lua mng openexr postscript python udev unwind vector-icons webp wmf xpm cpu_flags_ppc_altivec cpu_flags_x86_mmx cpu_flags_x86_sse"


band-a-prend:
There is a number of packages that depends on gimp:2 (i.e. currently gimp-2.10.x branch) and that seems are incompatible with media-gfx/gimp-9999 (here :2.99 SLOT now).

Could I explicitly add SLOT for each of this packages without revision bump even if it's already stable?

@jer-gentoo , @johu : app-misc/emelfm2 - there are 3 revisions;  
media-gfx/dcraw - no maintainer;  
media-gfx/gmic - no maintainer;  
@leio : media-gfx/gtkam;  
media-gfx/ufraw -  no maintainer;  
@akhuettel , @tamiko : media-gfx/xsane;  
media-plugins/gimp-lqr - no maintainer.

thesamesam:
I think we should revbump with a short period before stabling (if at all) so that the slot is applied on systems. 

It is possible to do things with a [slotmove](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-360004.4.4) but I'm not sure that's necessary here.

mgorny:
Can it actually be installed alongside old GIMP?

band-a-prend:
> Can it actually be installed alongside old GIMP?

I assume no. Should it then be a subslot `0/2.99` or listed gimp plugins are required to be restricted with `<gimp-2.99` ?

At first variant, then I need to change SLOT of gimp-2.10 to `0/2` with dropping python2 and pygtk dependencies.

mgorny:
Yes, it should be a subslot if it can't be installed alongside. You don't have to change old subslot, 0/0 is fine.

band-a-prend:
Ok. Then I close this PR.

PR #17254, opened 2020-08-24
dev-libs/rocr-runtime: Added dependency "xxdi"

justxi:


candrews:
If you remove `app-editors/vim-core` and install just `dev-util/xxdi` does this package emerge correctly?

I ask because `dev-util/xxdi` provides `xxdi.pl` - not `xxdi`, so I suspect it's not a straight drop in replacement.

justxi:
Sorry, thought that it is a drop in replacement...
I will check it...

justxi:
According to the script in rocr-runtime only "xxd" will be used.

perkinslr:
I've opened a PR upstream to add support for using xxdi.pl, as it just needs to elide the `-i` in the argument list.  If that gets accepted, this should probably get re-opened.  If it gets rejected, we can consider if it's worth patching here.
https://github.com/RadeonOpenCompute/ROCR-Runtime/pull/103

PR #17255, opened 2020-08-25
www-apps/gitea: Version bump 1.12.4.

tastytea:
#### Differences to 1.12.1 ebuild

* Added `RESTRICT="test"`, because I don't know how to fix the failing tests.
* Addded `QA_PRESTRIPPED="usr/bin/gitea"`.
* Added `pkg_postinst()`.


thesamesam:
Re tests:
- have you asked us for help on it?
- tried to reproduce them on a fresh clone of the git repo (outside of portage)?
- been able to reproduce on your system with the ebuild?
- seen if upstream have any reports?
- reported upstream?

tastytea:
I lack the knowledge to deal with go tests. I hope the maintainer (@nemunaire) will eventually deal with it.

See also: <https://bugs.gentoo.org/733150>.

hydrapolic:
Upstream runs the tests before releasing: https://drone.gitea.io/go-gitea/gitea/30640/2/1

For a complete test we would need mysql, elasticsearch, ldap... so it's kind of out of scope I think. Since this is not a library but an end user app, I think we actually could restrict the tests. Just grabbed the source code and even the unit tests fail:

```
--- FAIL: TestBlob_Data (0.00s)
    blob_test.go:39:
                Error Trace:    blob_test.go:39
                Error:          Received unexpected error:
                                no such file or directory
                Test:           TestBlob_Data
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x8a9e33]

goroutine 23 [running]:
```

Just tested copying 1.12.1 -> 1.12.4 and it works (security issues fixed :)).

r7l:
What exactly is missing here for this version to end up in Portage? How would these upstream tests affect it?

thesamesam:
It needs a rebase, ideally.

PR #17259, opened 2020-08-25
net-nds/openldap: Inherit from preserve-libs for preserve_old_lib

Allen-Webb:
Bug: https://bugs.gentoo.org/739010
Signed-off-by: Allen Webb <allenwebb@google.com>


thesamesam:
Thanks for this - you got pipped [here](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=03efb9dea7e53280aa6399f545607122c2baa16a) unfortunately!

PR #17264, opened 2020-08-26
net-misc/roger-router: New ebuilds

jeeger:
These commits add new ebuilds for roger-router (a desktop frontend for Fritz!Box routers), a dependency (librm) and a bumped version of libhandy (which the git ebuild requires).


ajakk:
Thanks for working on this! I think [GURU](https://wiki.gentoo.org/wiki/Project:GURU) will be the best place for it for now

PR #17269, opened 2020-08-26
dev-libs/nss: Add NS_USE_GCC environment variable

rossbridger:
This variable seems to be required to be be set in order to have certain
files to be compiled on arm64 linux systems, otherwise some symbols will be missing at linking thus causing build failure.

AFAIK the purpose of NS_USE_GCC flag is just to enable some accelerated algorithms written in gcc's C dialect, so there shouldn't be any build failures or behavior changes on other CPU arches.

Upstream's commit on this behavior change: https://hg.mozilla.org/projects/nss/rev/d98bbb6168f4ca2abd534e4c2fce56b7a5d1ad7e

Closes: https://bugs.gentoo.org/738924

### DIFF
--- a/dev-libs/nss/nss-3.56.ebuild
+++ b/dev-libs/nss/nss-3.56.ebuild
@@ -161,6 +161,8 @@ multilib_src_compile() {
 	export USE_SYSTEM_ZLIB=1
 	export ZLIB_LIBS=-lz
 	export ASFLAGS=""
+	# Fix build failure on arm64
+	export NS_USE_GCC=1
 
 	local d
 
--- a/dev-libs/nss/nss-3.56.ebuild
+++ b/dev-libs/nss/nss-3.56.ebuild
@@ -163,6 +163,12 @@ multilib_src_compile() {
 	export ASFLAGS=""
 	# Fix build failure on arm64
 	export NS_USE_GCC=1
+	# Detect compiler type and set proper environment value
+	if tc-is-gcc; then
+		export CC_IS_GCC=1
+	elif tc-is-clang; then
+		export CC_IS_CLANG=1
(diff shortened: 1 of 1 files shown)


Whissi:
I am wondering if this will cause problems for those people using CLANG by default.

rossbridger:
I don't have clang installed so I didn't tested it, but since clang supports gcc's C dialect, i guess it should be a problem.

The only flag specific to clang is CC_IS_CLANG and there is also a CC_IS_GCC flag, these flags are used to detect if the compiler has certain features (eg. int128 support, avx2 support), but that's a different story.

rossbridger:
I just tested using clang to build with NS_USE_GCC enabled, it compiles successfully on both amd64 and arm64.

rossbridger:
@Whissi just in case I added something for detecting compiler types, this should make clang users happy this time.

dev-libs/nss/nss-3.56.ebuild
```diff
@@ -163,6 +163,12 @@ multilib_src_compile() {
 	export ASFLAGS=""
 	# Fix build failure on arm64
 	export NS_USE_GCC=1
+	# Detect compiler type and set proper environment value
+	if tc-is-gcc; then
+		export CC_IS_GCC=1
```
Whissi, line 168: Now shouldn't you move `export NS_USE_GCC=1` into "tc-is-gcc" clause?

rossbridger: This NS_USE_GCC variable is also needed by clang to compile successfully in arm64. As I said earlier, this flag is only to indicate that the compiler supports GCC's C extension, which is also supported by clang.

PR #17270, opened 2020-08-26
app-misc/todo: bump version to 2.12.0, remove old

ferki:
This PR bumps app-misc/todo to version 2.12.0. It also updates upstream repo in metadata, and removes the oldest in-tree version.

Closes: https://bugs.gentoo.org/738094


juippis (approved):

ferki:
Thanks @juippis for merging the metadata update! It looks like this PR remained open after the separate version bump by @monsieurp.

Seemingly the only change between this and the current in-tree version is dropping the previously used Gentoo-specific patch for `todo.sh` command aliases (`todotxt` and `todocli`). However the related `bashcomp_alias` line was not removed, so now it throws the following warnings upon installation:

```
 * Problems with installed bash completions were found:
 *
 *      todocli: incorrect name, no completions for 'todocli' command defined.
 *      todotxt: incorrect name, no completions for 'todotxt' command defined.
 *
 * For more details on installing bash-completions, please see:
 * https://wiki.gentoo.org/wiki/Bash/Installing_completion_files
 *
```

This change of default behavior on Gentoo might be a slight inconvenience for those who relied on the command aliases. Since there's no bug report about it (yet?), while it simplifies maintenance, moves the Gentoo experience closer to the vanilla one, and I guess it's possible to configure the package for the same behavior after installation, I think the benefits are greater than the regression caused. So I'm fine with overall idea of dropping the aliases :+1: 

@juippis @monsieurp : What do you folks think? Should we drop the `bashcomp_alias` line in a subsequent PR (and perhaps add a post-install note about the change)? Or you would prefer to keep the previous behavior and cherry-pick the related bits from here?

juippis:
I'd keep them if it isn't too much work to fix it. It needs fixing to one way or another - you're the maintainer, you decide what's best.

ferki:
> I'd keep them if it isn't too much work to fix it. It needs fixing to one way or another - you're the maintainer, you decide what's best.

Alright, @monsieurp is also a maintainer and seemingly had a reason to change this, so let's wait for his feedback too :)

ferki:
Soo, based on bugzilla feedback, I'll rebase my version on top of current master branch, and will open a new PR. Thanks, folks!

PR #17271, opened 2020-08-26
net-mail/mu: Bump to 1.4.13

00-matt:


net-mail/mu/mu-1.4.13.ebuild
```diff
@@ -0,0 +1,91 @@
...
+IUSE="emacs guile mug"
+
+DEPEND="
+	dev-libs/glib:2
+	dev-libs/gmime:3.0
+	>=dev-libs/xapian-1.4
+	emacs? ( >=app-editors/emacs-24.4:* )
+	guile? ( >=dev-scheme/guile-2.2 )
+	mug? (
+		net-libs/webkit-gtk:4
```
thesamesam: If we link against something and it has a subslot, we should [subscribe](https://devmanual.gentoo.org/general-concepts/dependencies/) to it with :=.

net-mail/mu/mu-1.4.13.ebuild
```diff
@@ -0,0 +1,91 @@
...
+}
+
+pkg_postinst() {
+	if use emacs; then
+		einfo "To use mu4e you need to configure it in your .emacs file"
+		einfo "See the manual for more information:"
+		einfo "https://www.djcbsoftware.nl/code/mu/mu4e/"
+	fi
+
+	use emacs && elisp-site-regen
```
thesamesam: I'd just include elisp-site-regen in the existing if.

00-matt:
Thank you for the tips, updated.

PR #17272, opened 2020-08-26
dev-python/pycurl: add patch for multiple SSL backends

steils:
net-misc/curl supports multiple SSL backends since 5e560782
"net-misc/curl: enable multiple ssl implementations", which is supported
in curl since 7.56.

pycurl supporting this feature is not released yet, so with the latest
pycurl users could get an error like:

ImportError: pycurl: libcurl link-time ssl backend (gnutls) is different
from compile-time ssl backend (openssl)

Add patch from upstream to support multiple SSL backend in pycurl.

Signed-off-by: Stefan Strogin <steils@gentoo.org>


mgorny (approved):
LGTM. Thanks!

PR #17273, opened 2020-08-26
dev-python/pypy* : update SRC_URI

winterheart:
dev-python/pypy-exe, dev-python/pypy, dev-python/pypy3-exe, dev-python/pypy3 uses SRC_URI that not available anymore (see #737896):

``` diff
-SRC_URI="https://bitbucket.org/pypy/pypy/downloads/${MY_P}-src.tar.bz2
+SRC_URI="https://downloads.python.org/pypy/${MY_P}-src.tar.bz2
```

Checksums matches, so there no any revbumps required.


mgorny (approved):
Thanks!

PR #17280, opened 2020-08-27
sys-apps/edac-utils: revbump to resolve #725540 and refresh bundled libtool

jiblime:
Closes: https://bugs.gentoo.org/725540

The system requires GCC >=10 and binutils >=2.34 to take advantage of LDPT_ADD_SYMBOLS_V2, so a check is made if the toolchain is able to support -flto. Clang does not suffer from the same issue that GCC does but requires that `llvm-nm` is used when using LTO.

Built with:
`CFLAGS="-O2 -march=native -flto -pipe" CXXFLAGS="${CFLAGS}"`

`CC=gcc CXX=g++ LD=ld.bfd NM=nm`
`CC=gcc CXX=g++ LD=ld.bfd NM=gcc-nm`
`CC=clang CXX=clang++ LD=ld.bfd NM=llvm-nm`
`CC=clang CXX=clang++ LD=ld.lld NM=llvm-nm`

References:
https://wiki.gentoo.org/wiki/Outdated_libtool
https://sourceware.org/bugzilla/show_bug.cgi?id=25355


sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -38,7 +47,7 @@ src_install() {
 	# The rest is done via the udev-rule
 	rm -rf "${D}/etc/init.d"
```
thesamesam, line 48: || die on external commands, please

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -1,9 +1,9 @@
 # Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=5
+EAPI=7
 
-inherit eutils ltprune
+inherit autotools eutils flag-o-matic
```
thesamesam, line 6: Is eutils used here? If not, drop

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -19,9 +19,18 @@ RDEPEND="${DEPEND}
 	sys-apps/dmidecode"
```
thesamesam, line 19: It's a few lines above but we drop [back](https://devmanual.gentoo.org/keywording/index.html) to ~amd64 when bumping usually.

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -19,9 +19,18 @@ RDEPEND="${DEPEND}
...
 
 src_prepare() {
+	default
+
+	# <GCC-10, <binutils-2.34 do not support LDPT_ADD_SYMBOLS_V2 https://sourceware.org/bugzilla/show_bug.cgi?id=25355
```
thesamesam, line 24: Pop the link on a new line. I would also like a gentoo bug about this we can reference in the commit message?

thesamesam: Maybe you could even fix the calls [CC](https://bugs.gentoo.org/725540) directly bug ;)

thesamesam:
Please use repoman / pkgcheck scan to check commits and include the needed GCO sign off (see what the bot says). Thanks for the PR!

thesamesam:
(GLEP 66 says how to use the commit message to close the [bug](https://www.gentoo.org/glep/glep-0066.html) btw).

jiblime:
Hi @thesamesam , thank you for the helpful feedback! Sorry for the mistakes, I'll take care to read it over before making more edits.

thesamesam:
Hi @jiblime. Please do not apologise for not knowing everything perfectly. Nobody is born knowing ebuilds.

That's exactly what the review process is for, so thank you for your contribution! All we ask is that people take things constructively and learn with us as we go, which you're doing.

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -1,27 +1,41 @@
...
 RDEPEND="${DEPEND}
 	sys-apps/dmidecode"
 
 src_prepare() {
+	default
+
+	# <sys-devel/gcc-10, <sys-devel/binutils-2.34 do not support LDPT_ADD_SYMBOLS_V2
+	# https://sourceware.org/bugzilla/show_bug.cgi?id=25355
+	local gcc_vers=$(gcc-version)
+	local ld_vers=$(ld.bfd -v | awk '{ print $4 }')
```
thesamesam, line 27: Do we know ld.bfd always exists?

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -1,27 +1,41 @@
...
 DEPEND="sys-fs/sysfsutils"
 RDEPEND="${DEPEND}
 	sys-apps/dmidecode"
 
 src_prepare() {
+	default
+
+	# <sys-devel/gcc-10, <sys-devel/binutils-2.34 do not support LDPT_ADD_SYMBOLS_V2
+	# https://sourceware.org/bugzilla/show_bug.cgi?id=25355
+	local gcc_vers=$(gcc-version)
```
thesamesam, line 26: I think some of these functions come from [toolchain-funcs](https://devmanual.gentoo.org/eclass-reference/toolchain-funcs.eclass/index.html), so we must inherit that. I guess flag-o-matic inherits it indirectly for us, but we should be explicit.

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -1,27 +1,41 @@
...
+	local ld_vers=$(ld.bfd -v | awk '{ print $4 }')
+
+	if tc-is-gcc && ver_test ${gcc_vers} -lt 10 ||
```
thesamesam, line 29: A comment here explaining this would help, but maybe we should also put some of the tests into their own variables, like:
local gcc_too_old=....

Maybe I should ask some of the toolchain folks if this could be simplified.

thesamesam (commented):
Left a few more comments. Will ask someone from toolchain/knowledgable about such areas to have a look at the test condition.

No need to have revert commits, do you know about git rebase? It is ok to force push to your PR branch. So just squash all the commits into 1 clean one. :)

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
jiblime: >Do we know ld.bfd always exists?

Cases where it won't exist: the system has a nonstandard arch or sys-devel/binutils-config is built with USE=-native-symlinks.

With regards to multiarch compat, I think the ~amd64 keyword is sufficient for the time being, given the homepage's description. Beyond that, I don't have much experience with how different architectures handle things.

`[...]these modules mainly handle detection of ECC memory errors for many
x86 and x86-64 chipsets and PCI bus parity errors.`

---

Regarding ld.bfd; would this be better? `$(binutils-config -B)##*/}`

And thanks for the rebase tip, I'll look into it!

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
jiblime: Noted on adding comments, but I'm not too sure what you mean after that.

It feels somewhat heavy already for only checking versions. Perhaps a case statement would be cleaner and easier to work with?

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -0,0 +1,76 @@
...
+			filter-flags -flto*
+		fi
+	fi
+
+	# If Clang is used as compiler and LTO is enabled, ensure llvm-nm is used
+	# This applies when using ld.bfd, ld.gold, or ld.lld
+	if tc-is-clang && is-flag -flto*; then
+		einfo "Enforcing NM=llvm-nm for Clang -flto"
+		export NM=llvm-nm
+	fi
```
jiblime, line 45: To explain, I'm now getting the versions from the Portage environment rather than user's. I changed the filter-flags conditions because it would be needlessly filtering out -flto if someone chose to use Clang + ld.bfd/gold or GCC + ld.lld.

Is the second part needed? Is the onus on the user to have a proper Clang LTO build environment in the first place?

jiblime: Note:
Changed ld_vers to:
`ld_vers=$($(tc-getLD) -v | sed 's/.*\(2\.[0-9][0-9]\).*/\1/')`

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -0,0 +1,75 @@
...
+	default
+
+	# <sys-devel/gcc-10, <sys-devel/binutils-2.34 do not support LDPT_ADD_SYMBOLS_V2
+	# https://sourceware.org/bugzilla/show_bug.cgi?id=25355
+	# We use Portage's build environment to determine which GCC and binutils version are
+	# active. If either fail to meet the minimum version, -flto* is stripped from FLAGS
+	local gcc_vers=$(gcc-version)
+	local ld_vers=$($(tc-getLD) -v | sed 's/.*\(2\.[0-9][0-9]\).*/\1/')
+
+	if [[ tc-is-gcc || ! tc-ld-is-lld ]]; then
```
Arfrever, line 31: This line checks if string `tc-is-gcc` is non-empty or if string `tc-ld-is-lld` is empty.
Just drop this check.

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -0,0 +1,75 @@
...
+			einfo "Toolchain is outdated, removed -flto flags"
+			filter-flags -flto*
+		fi
+	fi
+
+	# If Clang is used as compiler and LTO is enabled, ensure llvm-nm is used
+	# This applies when using ld.bfd, ld.gold, or ld.lld
+	if tc-is-clang && is-flag -flto*; then
+		einfo "Enforcing NM=llvm-nm for Clang -flto"
+		export NM=llvm-nm
```
Arfrever, line 43: If `NM` is already set to absolute path (e.g. `NM=/usr/lib/llvm/10/bin/llvm-nm`), then there is no need to re-set `NM`.
So you can check value of `NM`:
```
if [[ ${NM##*/} != llvm-nm ]]; then
	export NM=llvm-nm
fi
```

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -0,0 +1,75 @@
...
+
+src_prepare() {
+	default
+
+	# <sys-devel/gcc-10, <sys-devel/binutils-2.34 do not support LDPT_ADD_SYMBOLS_V2
+	# https://sourceware.org/bugzilla/show_bug.cgi?id=25355
+	# We use Portage's build environment to determine which GCC and binutils version are
+	# active. If either fail to meet the minimum version, -flto* is stripped from FLAGS
+	local gcc_vers=$(gcc-version)
+	local ld_vers=$($(tc-getLD) -v | sed 's/.*\(2\.[0-9][0-9]\).*/\1/')
```
Arfrever, line 29: This regex fails with binutils `2.9` (or any version where minor version consists of single digit) and `3.0` (or any version where major version is not `2`).

jiblime: Thanks for the feedback! I understand these are more hacks than fixes, but I couldn't figure a way that would be compatible with both bfd and gold. what I had in mind is that Gentoo wouldn't use a binutils with a single digit minor version and it would be some years before 3.0. What would be the proper way to do a version check?

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
jiblime: Would something like 

`export NM=""$(llvm-config --prefix)/bin/clang"` (Does this print multiple directories?)

be needed? I don't have more than one slot of clang installed in my system, so I'm unclear on how users decide which version is default besides setting their PATH in order.

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
jiblime: I can see how this is redundant now, thank you for the advice!

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
juippis, line 29: Do you need these variables? Can't you test it via `$(xy)`?

juippis, line 29: Isn't this check already done in the `if` construct below? This sed seems rather redundant.

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
```diff
@@ -0,0 +1,75 @@
...
+		$(use_enable debug)
+}
+
+src_install() {
+	default
+
+	# We don't need this init.d file
+	# Modules should be loaded by adding them to /etc/conf.d/modules
+	# The rest is done via the udev-rule
+	rm -rf "${D}/etc/init.d" || die
```
juippis, line 65: We don't use `rm -f` the f flag since it won't spot errors. You're supposed to know whether the file exists or not, or in what cases it exists.

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
juippis, line 43: What's wrong with Arfever's suggestion though? :) It seems simpler to me. We shouldn't overcomplicate things.

sys-apps/edac-utils/edac-utils-0.18-r1.ebuild
Arfrever: In regular expressions, `[0-9][0-9]` are exactly 2 digits, while `[0-9]+` would be any non-zero number of digits, so it would handle `2.5`, `2.100` etc.
Specifically for `sed` tool, `+` needs to be escaped, so `[0-9]\+`.
For correct handling of major version, expectation of space before major version should be added.
Updated full line:
```
local ld_vers=$($(tc-getLD) -v | sed 's/.* \([0-9]\+\.[0-9]\+\).*/\1/'
```

thesamesam:
@jiblime ping, although let's not worry too much about the perfect fix -- it's been long enough

thesamesam:
OK, so we've now got afcc82fb9728ae744ad890e54a4c1b60145d64fa which I ended up doing independently. Anything left to take now that GCC 10 is stable?

PR #17282, opened 2020-08-27
dev-lang/ispc: Version bump and update to use python3

DarkDefender:
I've bumped and updated the ebuilds to use python3.

Related bug ticket: https://bugs.gentoo.org/735284


dev-lang/ispc/Manifest
```diff
@@ -1 +1,7 @@
-DIST ispc-1.13.0.tar.gz 19202295 BLAKE2B 5fae81429acbda0d65d33e56609d05fb8b5a7db6dd0af5ff779d1e8954f6a98d1fcb0f7ecf0c1b9813630026557c21c2a70bfc87ed75959a8bd24e06fcd2e351 SHA512 d23c127beb9bb17e1ad2509f9ad51dd660c473458b939a6b64ca7f6319b3c2f25318ea2ceac00e4100a059f3f67fb97b6973f18e3432c82ff9691f2a08d15d5d
+AUX ispc-1.14.0-cmake-gentoo-release.patch 507 BLAKE2B e67be17f40dce55895ce307bd5383ad3f9dc8ec460c7195e23ce344f3a0324c53f9f69b695b6feb70d18da813d5d855d675a00e8dfd3e651aaaebf510de1138b SHA512 bdfd7d3f023ace4a0209201e05b7528e50cff644794bc48b5160d8051c2d70dad6690e509ca69102b08843ebb5fd7a76241954b22a0bb1f3c4714708df405d65
```
thesamesam, line 1: Got you! You need to [enable](https://wiki.gentoo.org/wiki/Repository_format/metadata/layout.conf#thin-manifests) thin-manifests on your testing overlay because the AUX/EBUILD/MISC lines are not necessary when using e.g. git which ensures the integrity of the files.

dev-lang/ispc/ispc-1.14.0.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python2_7 )
+PYTHON_COMPAT=( python3_{6,7,8} )
```
thesamesam, line 6: Does 3.9 work?

dev-lang/ispc/ispc-9999.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python2_7 )
+PYTHON_COMPAT=( python3_{6,7,8} )
```
thesamesam, line 6: 3.9 work?

dev-lang/ispc/ispc-9999.ebuild
```diff
@@ -35,9 +35,9 @@ DEPEND="
 	"
 
 PATCHES=(
-	"${FILESDIR}/${PN}-1.13.0-cmake-gentoo-release.patch"
-	"${FILESDIR}/${PN}-1.13.0-llvm-10.patch"
-	"${FILESDIR}/${PN}-1.13.0-werror.patch"
+	"${FILESDIR}/${PN}-1.14.0-cmake-gentoo-release.patch"
+	"${FILESDIR}/${PN}-1.14.0-llvm-10.patch"
+	"${FILESDIR}/${PN}-1.14.0-werror.patch"
```
thesamesam, line 40: It looks like the cmake + werror patches didn't change at all, just bumped the version number. We normally leave these as-is unless they got rebased.

dev-lang/ispc/ispc-9999.ebuild
```diff
@@ -35,9 +35,9 @@ DEPEND="
...
 )
 
 src_configure() {
```
thesamesam, line 43: The $(usex arm) thing is odd, but not wrong. I wonder if we need it on arm64 too? Or any other architectures (different options)?

dev-lang/ispc/ispc-9999.ebuild
```diff
@@ -35,9 +35,9 @@ DEPEND="
 	"
```
thesamesam, line 35: I think bison/flex are BDEPEND (you don't need to run them for CHOST during cross compilation).

Also, if we definitely NEED clang/llvm (cannot use gcc?), we should check in pkg_setup and bail out if not using those.

thesamesam (changes requested):
Left a few comments. Could you also poke at enabling [tests](https://github.com/ispc/ispc/wiki/Regression-tests)?

dev-lang/ispc/ispc-1.14.0.ebuild
DarkDefender: There shouldn't be any issue that I know of.

So I guess I'll just add it?

dev-lang/ispc/ispc-9999.ebuild
DarkDefender: It does work with gcc too. It is just not supported.

There was an issue before were clang only compile flags were used and it would prevent gcc from not erroring out.
But I reported it and it is not fixed upstream: https://github.com/ispc/ispc/pull/1829
(Fixed in the 1.14.0 version and the live version).

I'm using gcc to compile this on my side. So clang is not strictly needed. (llvm is though)

dev-lang/ispc/ispc-1.14.0.ebuild
thesamesam: Does it work with it and do tests pass, ideally?

dev-lang/ispc/ispc-9999.ebuild
thesamesam: OK.

dev-lang/ispc/ispc-1.14.0.ebuild
DarkDefender: It seems to be working and the tests pass.

dev-lang/ispc/ispc-9999.ebuild
DarkDefender: Hmm, I don't have any arm computers to check this with...

DarkDefender:
I have enabled test (they pass on my computer) and fixed most of the issues you pointed out.
The only thing list if the `usex arm thing` I don't know how to resolved that as I can't really test it.

dev-lang/ispc/ispc-9999.ebuild
thesamesam: Well, this isn't even keyworded on arm or arm64, so no need to worry too much. We could look at the build system to see what it's doing for fun though.

thesamesam (commented):
Pending testing, this looks good. If I had to nit pick, I'd say drop the old version in a separate commit (note that you will need to make a copy of the patch file then too for the old one to keep working), just in case we have to revert this one (not that we want that to happen) -- it just makes things cleaner.

dev-lang/ispc/ispc-1.14.0.ebuild
```diff
@@ -57,3 +59,7 @@ src_install() {
 		doins -r "${BUILD_DIR}"/examples/*
```
thesamesam, line 59: This might have been discussed with someone else before, but how big are the examples? If they're small, we'll just install them unconditionally.

DarkDefender:
Just to be clear, you want me to have the commit be:

1. Nuke the ispc ebuilds entirely
2. Commit with the new 1.40.0 and the live ebuilds

?

dev-lang/ispc/ispc-1.14.0.ebuild
DarkDefender: The example directory is 78 MB. So it is not that big but not that small either.

thesamesam:
Other way around:
1. Add your new 1.40.0
2. Improve/change live
3. Drop old/cleanup old

Does that seem ok?

dev-lang/ispc/ispc-1.14.0.ebuild
thesamesam: This is probably enough to justify the USE flag. Thanks!

thesamesam:
Nearly there. Missing the signoffs on the commits. :)

thesamesam:
```
>>> Install dev-lang/ispc-1.14.0 into /var/tmp/portage/dev-lang/ispc-1.14.0/image
>>> Completed installing dev-lang/ispc-1.14.0 into /var/tmp/portage/dev-lang/ispc-1.14.0/image

 * Final size of build directory: 65663 KiB (64.1 MiB)
 * Final size of installed tree:      8 KiB

scanelf: rpath_security_checks(): Security problem NULL DT_RUNPATH in /var/tmp/portage/dev-lang/ispc-1.14.0/image/usr/bin/ispc
   /usr/bin/ispc
./
./usr/
./usr/bin/
./usr/bin/ispc
./usr/share/
./usr/share/doc/
./usr/share/doc/ispc-1.14.0/
./usr/share/doc/ispc-1.14.0/README.md.bz2
>>> Done.
```

DarkDefender:
I have no clue how to fix that issue

thesamesam:
No worries. I did some research into this and -DCMAKE_SKIP_RPATH=ON was familiar to me. It looks like it's the right solution -- don't bother defining RPATH if it's unnecessary.

[This](https://security.stackexchange.com/questions/161799/why-does-checksec-sh-highlight-rpath-and-runpath-as-security-issues), along with some other things, could be useful.

DarkDefender:
Perhaps you can add that check to the qa bot? (With a link to the gentoo wiki explaining it) :thinking:

thesamesam:
You should've seen that message on install -- at least if you change to use my settings now. I think that ELOG one may be needed to show it.

I agree some documentation on the wiki is a good idea for all sorts of QA warnings; it's not always clear how to start fixing them. Thanks again for your contribution and I hope it's the first of many!

PR #17283, opened 2020-08-27
games-engines/odamex: Fix line endings in odalaunch patch

vilhelmgray:
Closes: https://bugs.gentoo.org/739186


chewi:
Think I've fixed this. Will write more later.

chewi:
I don't think `git commit`, `repoman`, our Gentoo git server, or GitHub are to blame here. I'm don't know how you prepared the patch but I know handling DOS line endings in patches can be a bit finicky. To fix it, I ran `unix2dos` on the patch file, opened it with `nano -N`, and then removed the `^M` characters from the end of each line that wasn't part of the patch content, i.e. the first and last few lines.

vilhelmgray:
Thanks, I'll keep an eye on this in the future. :)

PR #17287, opened 2020-08-27
sys-auth/sssd-2.2.3: Fix glibc 2.32 name conflict

webknjaz:
This change applies a patch that is present in newer versions of sssd but hasn't been backported to 2.2.x. It effectively fixes the compatibility with glibc 2.32 that introduced some symbols with the same names that used to exist in sssd source (namely `nss_setnetgrent`).

Ref: https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=499a92df8b9fc64a054cf3b7f728f8967fc1da7d;hp=8a9221415041543aa163071fa547e88138ceec9d

Fixes: https://bugs.gentoo.org/737180

### DIFF
--- a/sys-auth/sssd/files/sssd-2.2.3-glibc-2.32-compat.patch
+++ b/sys-auth/sssd/files/sssd-2.2.3-glibc-2.32-compat.patch
@@ -1,0 +1,71 @@
+From fe9eeb51be06059721e873f77092b1e9ba08e6c1 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Michal=20=C5=BDidek?= <mzidek@redhat.com>
+Date: Thu, 27 Feb 2020 06:50:40 +0100
+Subject: [PATCH] nss: Collision with external nss symbol
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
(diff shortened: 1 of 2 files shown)


webknjaz:
cc @mattst88

mattst88:
Thanks.

I don't see that the patch is being applied in the 2.2.3 ebuild.

Since `sssd` is in such bad shape in Gentoo at the moment, and because I've done a bit of work to v2.3.1 I wonder whether we should bother patching v2.2.3?

webknjaz:
@mattst88 I wasn't sure if I needed to change something in ebuild files: none of them seem to explicitly refer to patches in the `files/` dir so I assumed that they may be picked up automatically by some thing in the framework.

Now that I checked the commits they were added in, I understand that I need to either use `PATCHES=` or `eapply`. I'll update it shortly.


> Since `sssd` is in such bad shape in Gentoo at the moment, and because I've done a bit of work to v2.3.1 I wonder whether we should bother patching v2.2.3?

I think it'd be useful for folks who don't want to unmask the new unstable version.

webknjaz:
@mattst88 is it better now?

mattst88:
Yes, thank you very much!

PR #17289, opened 2020-08-27
net-libs/libblkmaker: Bump to 0.6.0

luke-jr:


luke-jr:
Updated for EAPI 7 and multilib

net-libs/libblkmaker/libblkmaker-0.6.0.ebuild
```diff
@@ -0,0 +1,31 @@
...
+
+SRC_URI="https://github.com/bitcoin/${PN}/archive/v${PV}.tar.gz -> ${P}-github.tgz"
+SLOT="0/7"
+KEYWORDS="~amd64 ~arm ~ppc ~ppc64 ~x86"
+IUSE="test"
+RESTRICT="!test? ( test )"
+
+RDEPEND="
+	>=dev-libs/jansson-2.0.0
+	dev-libs/libbase58
```
juippis, line 20:
```
 * abi_x86_32.x86: running multilib-minimal_abi_src_configure
...
...
checking for libbase58... no
configure: error: Package requirements (libbase58) were not met:

Package 'libbase58', required by 'virtual:world', not found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables libbase58_CFLAGS
and libbase58_LIBS to avoid the need to call pkg-config.
```

libbase58 is not multilib'd so it wouldn't work with a revdep that requires lib32 compatibility.

dev-libs/libbase58/libbase58-0.1.4.ebuild
```diff
@@ -1,8 +1,8 @@
-# Copyright 1999-2017 Gentoo Foundation
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
-inherit autotools
+EAPI=7
```
juippis, line 4: First of all please revbump with resetted KEYWORDS, this'd edit a stabilized ebuild in a dramatic way.

dev-libs/libbase58/libbase58-0.1.4.ebuild
```diff
@@ -11,18 +11,37 @@ LICENSE="MIT"
...
+	elif use test; then
+		myeconf+=( --enable-tool --bindir='/TRASH' )
+	else
+		myeconf+=( --disable-tool )
+	fi
+	econf "${myeconf[@]}"
+}
+
+multilib_src_install_all() {
+	rm -rf "${ED}/TRASH" || die
```
juippis, line 46: We rarely use `-f` in rm, you should know when this dir exists and when not.

dev-libs/libbase58/libbase58-0.1.4.ebuild
```diff
@@ -11,18 +11,37 @@ LICENSE="MIT"
 SRC_URI="https://github.com/luke-jr/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
 SLOT="0/0"
 KEYWORDS="amd64 ~arm ~mips ~ppc ~ppc64 x86"
-IUSE="tools"
+IUSE="test tools"
```
juippis, line 14: I can't get the tests to work, 
```
Testsuite summary for libbase58 0.1.4
# TOTAL: 12
# PASS:  6
# SKIP:  0
# XFAIL: 0
# FAIL:  6
# XPASS: 0
# ERROR: 0
See ./test-suite.log
Please report to luke_libbase58@dashjr.org
make[2]: *** [Makefile:902: test-suite.log] Error 1
make[2]: Leaving directory '/var/tmp/portage/dev-libs/libbase58-0.1.4/work/libbase58-0.1.4-abi_x86_64.amd64'
make[1]: *** [Makefile:1010: check-TESTS] Error 2
make[1]: Leaving directory '/var/tmp/portage/dev-libs/libbase58-0.1.4/work/libbase58-0.1.4-abi_x86_64.amd64'
make: *** [Makefile:1214: check-am] Error 2
 * ERROR: dev-libs/libbase58-0.1.4::testworld failed (test phase):
 *   Make check failed. See above for details.
```
```
=======================================
   libbase58 0.1.4: ./test-suite.log
=======================================

# TOTAL: 12
# PASS:  6
# SKIP:  0
# XFAIL: 0
# FAIL:  6
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: tests/decode
==================

/var/tmp/portage/dev-libs/libbase58-0.1.4/work/libbase58-0.1.4/tests/decode.sh: line 2: base58: command not found
FAIL tests/decode.sh (exit status: 1)

FAIL: tests/decode-small
========================

/var/tmp/portage/dev-libs/libbase58-0.1.4/work/libbase58-0.1.4/tests/decode-small.sh: line 2: base58: command not found
FAIL tests/decode-small.sh (exit status: 1)

FAIL: tests/decode-zero
=======================

/var/tmp/portage/dev-libs/libbase58-0.1.4/work/libbase58-0.1.4/tests/decode-zero.sh: line 2: base58: command not found
FAIL tests/decode-zero.sh (exit status: 1)

FAIL: tests/encode
==================

/var/tmp/portage/dev-libs/libbase58-0.1.4/work/libbase58-0.1.4/tests/encode.sh: line 2: base58: command not found
FAIL tests/encode.sh (exit status: 1)

FAIL: tests/encode-b58c
=======================

/var/tmp/portage/dev-libs/libbase58-0.1.4/work/libbase58-0.1.4/tests/encode-b58c.sh: line 2: base58: command not found
FAIL tests/encode-b58c.sh (exit status: 1)

FAIL: tests/encode-small
========================

/var/tmp/portage/dev-libs/libbase58-0.1.4/work/libbase58-0.1.4/tests/encode-small.sh: line 2: base58: command not found
FAIL tests/encode-small.sh (exit status: 1)
```

If the binary is gotten from this product (although why would it be a library then?) you should be able to edit the PATH variable including correct BUILD_DIR/something there.

juippis (commented):
Also note that your commit summary is wrong, libbase58 is under dev-libs, not net-libs.

dev-libs/libbase58/libbase58-0.1.4.ebuild
juippis: Overall this seems a bit weird. What if I want to keep the tools even though enabling `test` IUSE?

luke-jr: Then you should enable the `tools` USE flag, which will put the native tools in `bin` instead of `TRASH`

To make `rm` conditional would be something like (pseudo-code because I don't know how to do it): if iuse test && more than 1 ABI is enabled (or I guess we can just mkdir TRASH earlier, but what's the point?)

juippis:
```
if multilib_is_native_abi; then
```
should work.

dev-libs/libbase58/libbase58-0.1.4.ebuild
```diff
@@ -11,18 +11,37 @@ LICENSE="MIT"
...
+		--disable-static
+		LIBGCRYPT_CONFIG="${EPREFIX}/usr/bin/${CHOST}-libgcrypt-config"
+	)
+	if multilib_is_native_abi && use tools; then
+		myeconf+=( --enable-tool )
+	elif use test; then
+		myeconf+=( --enable-tool --bindir='/TRASH' )
+	else
+		myeconf+=( --disable-tool )
+	fi
```
juippis, line 41: I'd maybe find it easier to just write `REQUIRED_USE="test? ( tools )`
But make sure you fix the test failure with appending temporary PATH so it finds the compiled binary before it's installed in root.

juippis (commented):
Now it works but I'll have to first emerge it with `FEATURES="-test"` then re-emerge, so it's not working as supposed. During stabilization it won't pass.

dev-libs/libbase58/libbase58-0.1.4.ebuild
luke-jr: `tools` only requires libgcrypt natively built, whereas `test` requires it for all target ABIs.

I'm going to have to fix the test issue upstream somehow.

juippis: I get it. Could you clean this ebuild and `RESTRICT="test"` on this version so we get this PR moved forward, while you fix it upstream?

dev-libs/libbase58/libbase58-0.1.4.ebuild
luke-jr: Not inside `multilib_src_install_all`?

dev-libs/libbase58/libbase58-0.1.4.ebuild
luke-jr: Building in the source dir works, so simply using `multilib_copy_sources` for this version seems fine.

luke-jr:
Fixed the issues. Still no good alternative to `rm -rf TRASH`...

dev-libs/libbase58/libbase58-0.1.4-r1.ebuild
```diff
@@ -0,0 +1,48 @@
...
+		myeconf+=( --enable-tool --bindir='/TRASH' )
+	else
+		myeconf+=( --disable-tool )
+	fi
+	econf "${myeconf[@]}"
+}
+
+multilib_src_install_all() {
+	rm -rf "${ED}/TRASH" || die
+}
```
juippis, line 48: I'm going to add 
```
multilib_src_install_all() {                                                                        
    if use test; then                                                                               
        rm -r "${ED}/TRASH" || die                                                                  
    fi                                                                                              
}                                                                                                   
```
which seems to work and do the job. Rest looks good, thanks!

juippis (approved):

dev-libs/libbase58/libbase58-0.1.4-r1.ebuild
luke-jr: That WON'T work... TRASH won't exist with USE `tools` and only one target ABI.

luke-jr:
Looks like I'm missing a dep on app-editors/vim-core (xxd) for USE=test

dev-libs/libbase58/libbase58-0.1.4-r1.ebuild
juippis: For some reason it doesn't error out with `USE="tools"` but I see where the logic is coming from. It does fail with `ABI_X86="(64) -32 (-x32)"`. I'll fix it in a moment, need to do few test runs. Thanks for pinging!

juippis: Ok, 47b6f9481bd5f547af2134a5dea2e15bd45ca889 I honestly have **no** idea why we didn't just go with the `-f`. Please change it like that on the next version bump. I'm too embarrased to make a commit straight after...

juippis: Okay the final version: 
```
multilib_src_install_all() {
	if use test; then
		# It's hard to control this directory with multilib_is_native_abi && use tools, hence -f.
		rm -rf "${ED}/TRASH" || die
	fi
}
```
Should be good. I needed to leave 15 minutes ago so my brain doesn't bend to this anymore. This should however work, or not work, in every case. But not die while emerging.

PR #17290, opened 2020-08-28
dev-libs/hidapi: Bump to 0.9.0 

luke-jr:


luke-jr:
Also added a fix for https://bugs.gentoo.org/551250

luke-jr:
And https://bugs.gentoo.org/661034

dev-libs/hidapi/hidapi-0.9.0.ebuild
```diff
@@ -0,0 +1,75 @@
...
+		virtual/libudev:0[${MULTILIB_USEDEP}]
+		virtual/libusb:1[${MULTILIB_USEDEP}]
+	)
+	kernel_FreeBSD? (
+		virtual/libusb:1[${MULTILIB_USEDEP}]
+	)
+"
+DEPEND="${RDEPEND}
+	doc? ( app-doc/doxygen )
+	virtual/pkgconfig
```
thesamesam, line 28: BDEPEND usually, but look at how it's used?

dev-libs/hidapi/hidapi-0.9.0.ebuild
```diff
@@ -0,0 +1,75 @@
...
+	fox? ( x11-libs/fox )"
```
thesamesam, line 29: fox not linked against? (I don't know.)

dev-libs/hidapi/hidapi-0.9.0.ebuild
```diff
@@ -0,0 +1,75 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools eutils multilib-minimal
```
thesamesam, line 6: Is eutils being used here?

thesamesam:
Ping

luke-jr:
I don't know either.

dev-libs/hidapi/hidapi-0.9.0.ebuild
netfab: From README.md \:
> HIDAPI also comes with a Test GUI. The Test GUI is cross-platform and uses Fox Toolkit http://www.fox-toolkit.org. It
> will build on every platform which HIDAPI supports. Since it relies on a 3rd party library, building it is optional but
> recommended because it is so useful when debugging hardware.

From configure.ac \:
> if test "x$testgui_enabled" != "xno"; then
>		# On Linux and FreeBSD platforms, use pkg-config to find fox.
>		PKG_CHECK_MODULES([fox], [fox17], [], [PKG_CHECK_MODULES([fox], [fox])]

juippis:
Ping.

jospezial:
Please update this pr as some changes of this are already in gentoo tree or obsolete and update to
https://github.com/libusb/hidapi/releases/tag/hidapi-0.10.1

jospezial:
Maybe you want to take it from
https://github.com/netfab/overlay/tree/master/dev-libs/hidapi

jospezial:
Upstream has added cmake build system and deprecated autotools build. Next release 0.11 will be released in a few days.

jospezial:
https://github.com/libusb/hidapi/releases/tag/hidapi-0.11.0
"NOTE: Starting with this release Autotools build scripts for HIDAPI are deprecated and going to be obsolete in near future.
All users of HIDAPI should switch to CMake build scripts instead."

juippis:
This doesn't seem to go anywhere and we have #22567

PR #17295, opened 2020-08-28
Haskell: bump logict, remove old

solpeth:
Also remove old `logict-state` version `0.1.0.2`. Tested in an ~amd64 container.

**Note:** reverse dependency `logict-state` has not been ported to `MonadFail`, so it does not currently build anyway. It seems unmaintained upstream for the last two years as well. I can try and work on a patch over the next few days, or I'll submit a pull request to mask it and its reverse dependencies.


trofi:
Pulled. Thank you!

PR #17318, opened 2020-08-29
sys-devel/gcc-config: Delete old unpackaged c{8,9}9 wrapper scripts

luke-jr:
Needed to avoid collision-protect complaining
Only deleted if:
1) collision-protect in FEATURES
2) Not replacing a version >=2.3.1
3) Existing files being deleted match the unpackaged scripts' hash

Bug: https://bugs.gentoo.org/737662

### DIFF
--- a/sys-devel/gcc-config/gcc-config-2.3.1.ebuild
+++ b/sys-devel/gcc-config/gcc-config-2.3.1.ebuild
@@ -38,6 +38,39 @@ src_install() {
 	_emake DESTDIR="${D}" install
 }
 
+maybe_remove_old_wrappers() {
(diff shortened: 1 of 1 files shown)


trofi:
I prefer not to maintain fragile logic of deleting files which were created by packages other than `gcc-config` (`toolchain.eclass` in this case): https://bugs.gentoo.org/737662#c2

You can delete files manually.

luke-jr:
This logic *isn't* fragile...

Users shouldn't need to delete them manually.

trofi:
> Users shouldn't need to delete them manually.

Most users don't have to delete these files manually. People who care about stricter collision protection will have to sort it on their own.

PR #17323, opened 2020-08-29
games-sports/vdrift: Add live ebuild, remove outdated ebuilds

MatthiasCoppens:
The 2012 ebuilds don't build anymore. I asked upstream for a new release.
This live ebuild fetches the source from GitHub and the game data from SourceForge as explained right here:
https://github.com/VDrift/vdrift/blob/master/docs/Getting_the_development_version.md.

Signed-off-by: Matthias Coppens <coppens.matthias.abc@gmail.com>


rossbridger:
Why name the live version 9999999 instead of 9999 as per the usual naming convention?

MatthiasCoppens:
> Why do you name the live version 9999999 instead of 9999 as per the usual naming convention?

The versioning scheme uses the date of the release as the version, the previous version for example was `vdrift-20120722`.

rossbridger:
But you also have deleted all the previous versions. I think the usual date related versioning scheme is something like vdrift-0_p20120722.

MatthiasCoppens:
I'll change it to `-9999`. Once upstreams releases a new version we still can choose between `-YYYY.MM.DD` (like `youtube-dl`) or `-0_pYYYYMMDD`.

scardracs:
Kind ping: this package has been dropped so please close that PR or decide to re-add it

ionenwks:
Re-adding and modernizing is still something I'm keeping a note at which I don't know if/when I'll look at -- but guess there's no need to keep this old PR open with the package gone from the tree.

PR #17325, opened 2020-08-30
net-libs/openpgm: version bump to a snapshot with py3 support

jok-ts:
This would allow for removing these mask lines:
profiles/package.mask:
net-libs/openpgm

profiles/base/package.use.mask:
net-libs/zeromq pgm

The zeromq change would still work as is if a new pgm version turns up [upstream](https://github.com/steve-o/openpgm/issues/62).


jok-ts:
Rebasing to use new versions. Upstream openpgm is still a snapshot since the release left the pkgconfig file misnamed.

thesamesam:
@jok-ts We could fix the pkgconfig file with sed or a patch of our own.

net-libs/openpgm/openpgm-5.3.128_p20200831.ebuild
```diff
@@ -0,0 +1,43 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8,9} )
```
thesamesam, line 6: Python 3.6 possible to keep?

net-libs/zeromq/zeromq-4.3.3-r1.ebuild
```diff
@@ -0,0 +1,61 @@
...
+IUSE="doc drafts pgm +sodium static-libs test unwind elibc_Darwin"
+RESTRICT="!test? ( test )"
+
+RDEPEND="
+	!elibc_Darwin? ( unwind? ( sys-libs/libunwind ) )
+	sodium? ( dev-libs/libsodium:= )
+	pgm? ( >=net-libs/openpgm-5.2.122:= )"
+DEPEND="${RDEPEND}
+	!elibc_Darwin? ( sys-apps/util-linux )
+	doc? (
```
thesamesam, line 24: The doc and pkgconfig is probably BDEPEND?

net-libs/openpgm/openpgm-5.3.128_p20200831.ebuild
```diff
@@ -0,0 +1,43 @@
...
+MY_PV="${PV//./-}"
+
+DESCRIPTION="Open source implementation of the Pragmatic General Multicast specification"
+HOMEPAGE="https://github.com/steve-o/openpgm"
+SRC_URI="https://github.com/steve-o/${PN}/archive/${MY_COMMIT}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="LGPL-2.1"
+SLOT="0/5.3"
+KEYWORDS="~amd64 ~arm ~arm64 ~hppa ~ppc ~ppc64 ~sparc ~x86"
+IUSE="static-libs"
```
thesamesam, line 20: Do we need static-libs? Is a package using it?

jok-ts:
Rebasing to use new versions. Upstream openpgm is still a snapshot since the release left the pkgconfig file misnamed.

> @jok-ts We could fix the pkgconfig file with sed or a patch of our own.

Sure, but what's the point? The only difference would be that the ebuild would break if the release tarball gets recut without a new version number.

net-libs/openpgm/openpgm-5.3.128_p20200831.ebuild
jok-ts: It would work just fine, but py3.6 support is going out in Gentoo if I understood correctly.

net-libs/zeromq/zeromq-4.3.3-r1.ebuild
jok-ts: Probably. I'd leave that to the package maintainer.

net-libs/openpgm/openpgm-5.3.128_p20200831.ebuild
jok-ts: net-libs/zeromq has IUSE="static-libs" so maybe. There are no other rdeps according to the mask commit.

jok-ts:
Package treecleaned, PR now moot.

mgorny:
Not at all, you can readd the package if  you need it.

PR #17345, opened 2020-08-31
mail-filter/rspamd: clean old and buggy versions, fix issue with gcc10

arkamar:
- Version 2.3 does not work with libressl, however, the bug was resolved in 2.4 and later
- Version 2.4 contains bugs in `plugins/lua/neural.lua` and `plugins/lua/phishing.lua` and users should not use it.
- GCC-10 fix for 2.5 is taken from upstream.
- Sync live with upstream


djc:
LGTM!

croessner:
If you da cleanups and bug fixes, can you please also make sure that the 9999 release keeps going. Recent updates broke it.

arkamar:
Sure, 9999 is synced with upstream.

croessner:
As far as I saw is that the CMakeFile is patched, which is a problem for the 9999.

arkamar:
Does it mean that the update from my last commit does not work for you?

Dessa:
it did work for me tho. well. until the zstd update happened upstream, then the zstd patch failed.

arkamar:
I have synced the live ebuild again. However, I have temptation to remove it completely. Do you use it? It seems to break quite often since library unbundling.

croessner:
Please do not touch the 9999 ebuild. I will release a new version this day. I fixed the patches and stuff for live ebuild.

arkamar:
Well, sorry, too late, I already touched 9999 in order to fix @Dessa's issue.

croessner:
Hm, I got trouble as proxy maint, because other people are actively working on rspamd ebuild. Therefor I spoke with Jonas. I want to keep proxy maint. So please stay in contact with me ;-)

djc:
@croessner please figure out how to cooperate with other proxy maintainers. @arkamar has made some nice contributions.

arkamar:
@djc, tanks!

Well, I didn't want to touch 9999 previously, but you asked me to do it in https://github.com/gentoo/gentoo/pull/17345#issuecomment-684623206, therefore I updated it.

I started to take care of `rspamd` because I haven't seen enough activity around this ebuild year ago and we use it in our company. I don't want to take proxy maintaining from your hands, so, please keep continue in your contributions, but maybe next time you can write a note to the ticket, that you work on this or create PR. I usually check both before I start with any effort.

croessner:
I looked at the 9999 ebuild changes. One thing is missing, if I am correct.

```
    blas? (
        sci-libs/lapack
        sci-libs/openblas
    )
```

Could you please add this to RDEPEND?

arkamar:
Maybe we should use virtuals instead.
```
	blas? (
		virtual/blas
		virtual/lapack
	)
```
The most of `::gentoo` packages use them and it seems to work with `rspamd` as well. What do you think?

thesamesam:
From the proxy maintainers perspective, it's fine if you're both co-maintainers, just as @djc said, find a way to coordinate if that's the way you want to proceed :).

It is nice to have two people wanting to look after a package!

If it's not ready yet, please mark it as a draft by the way.

croessner:
👍

juippis (approved):
Is -9999 really that important for anyobody? We have to carry large weird patches for it, and it seems to break all the time. We can't react fast enough.

Dessa:
it is for me. in fact 2.7 just got released (haven't looked at it yet) so the patches can likely be re-used for that.
i know this *shouldn't* be a reason but upstream releases happen not too frequently and they also don't *really* support releases (i. e. on bugs, you are supposed to use git version for testing)

im unware on how often this kind of breakage happens though, so i can't comment on the general maintenance overhead of live, i only recently started using it

arkamar:
I just created PR #17834 with version bump to 2.6. The 9999 is preserved as it seems to be used by people.

PR #17346, opened 2020-08-31
Add Coffee Script 2.5.1

Bobbbay:
Hello all,

This is my first time contributing to Gentoo, so please don't roar at me if I say something wrong 😄.

So I saw that there was no Coffee Script 2.5.1 in the repo, so I decided to add it. I decided to simply `cp` the old one (`2.3.2`) and, if errors should arise, adjust as needed. The problem is the following: I'm not sure how I can "test" this out. I'm completely willing to run this on my personal Gentoo installation, I'm just not sure how I set the system to point to `bobbbay/gentoo` instead of `gentoo/gentoo`. Thank you!

Also, I'm aware I need to update the manifest, or do I wait until this is approved? Cheers!

EDIT: Also added long description, straight from [coffeescript.org](https://coffeescript.org/).


Bobbbay:
After tons of debugging, I've found out that all four rebuild files have syntax errors. I'd be thrilled to fix all of these, but can't seem to find the reason. Went over to ebuild.sh and where the error was thrown from, seems like it has something to do with `*DEPEND` or `IUSE`. 

If anyone has thoughts, I'd love to hear them :))

thesamesam:
1. For testing, have a look at this [guide](https://wiki.gentoo.org/wiki/Custom_repository) (then click through to [here](https://wiki.gentoo.org/wiki/Handbook:AMD64/Portage/CustomTree#Defining_a_custom_repository)).

2. We should rebase the commits you made then force push to squash the commits together into just 1 for these changes. Let me know if you want some guidance on doing that.

3. Commit messages need to adhere to [GLEP 66](https://www.gentoo.org/glep/glep-0066.html#commit-messages), which includes the sign off which is missing too (see what the bot said ;)).

4. Just skimming this - not tested yet - I can't see anything glaringly wrong here. What error are you getting?

Bobbbay:
Hey Sam,

1. Will definitely take the read!
2. I know what these are, but have no clue what commands I need to run for them, help needed for sure!
3. I tried doing this in my last commit, is it wrong?
4. In the end I found out it was because I was running repoman as root, silly me ;)
5. That's a very handy dandy command, thanks!

How's this now?

Bobbbay:
Oh wow awesome! In that case I'm going to test the install on a custom overlay of some sort and see if we're 100% clean!

Bobbbay:
Oh wow, it worked like a freaking charm! Never thought ebuilds would be this fun 🤪 (especially once you're done)
```
coffee -v
```
Returns CoffeeScript version 2.5.1

Woohoo! @thesamesam 

I'm clearing this for review!

juippis (approved):

PR #17350, opened 2020-08-31
app-crypt/scrypt: bump to 1.3.1, add tests 

jsmolic:


thesamesam:
need rebase?

jsmolic:
> need rebase?

Done :-)

app-crypt/scrypt/scrypt-1.3.1.ebuild
```diff
@@ -0,0 +1,22 @@
...
+HOMEPAGE="http://www.tarsnap.com/scrypt.html"
+SRC_URI="http://www.tarsnap.com/scrypt/${P}.tgz"
+
+LICENSE="BSD-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+DOCS=( FORMAT )
+
+IUSE="test"
```
thesamesam, line 16: This doesn't add anything because we don't have any test dependencies. It's only needed when we're passing something through to the build system or checking deps.

app-crypt/scrypt/scrypt-1.3.1.ebuild
```diff
@@ -0,0 +1,22 @@
...
+RESTRICT="!test? ( test )"
+
+src_test() {
+	default
```
thesamesam, line 20: The 'check' target seems to do nothing so just dropping it.

thesamesam:
Thanks!

app-crypt/scrypt/scrypt-1.3.1.ebuild
jsmolic: Ah, I see, thanks for explaining

jsmolic:
> Thanks!

Thanks to you as always

PR #17354, opened 2020-09-01
app-misc/pwsafe-0.2.0-r5: revbump

Hawk777:


app-misc/pwsafe/pwsafe-0.2.0-r5.ebuild
```diff
@@ -0,0 +1,44 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DESCRIPTION="A Password Safe compatible command-line password manager"
+HOMEPAGE="http://nsd.dyndns.org/pwsafe/"
```
charIes17, line 7: Have you checked that url, both for homepage and src_uri?
curl --head https://nsd.dyndns.org/pwsafe/
curl: (28) Failed to connect to nsd.dyndns.org port 443: Connection timed out

curl --head dyndns.org/
HTTP/1.1 301 Moved Permanently
Date: Tue, 01 Sep 2020 07:19:14 GMT
Server: Apache
Location: http://www.dyndns.com/
Connection: close
Content-Type: text/html; charset=iso-8859-1

Hawk777: I know that site is dead. I don’t have any other homepage though. The author has a Github account, and pwsafe is there, but the Github project doesn’t have the 0.2.0 tarball or anything that resembles it by name. It’s a bit of a mess, but as long as the Gentoo mirror network still has the 0.2.0 tarball, it can still be installed, and I don’t really see a better option. Unless you suggest something else?

app-misc/pwsafe/pwsafe-0.2.0-r5.ebuild
```diff
@@ -0,0 +1,44 @@
...
+	readline? ( sys-libs/readline:0= )
+	X? ( x11-libs/libSM
+		x11-libs/libICE
+		x11-libs/libXmu
+		x11-libs/libX11 )"
+RDEPEND="${DEPEND}
+	!app-admin/passwordsafe"
+
+src_prepare() {
+	eapply -p0 "${FILESDIR}/${P}-cvs-1.57.patch"
```
thesamesam, line 28: Nowadays (with EAPI 7) we should be able to just define PATCHES=() as an array above src_prepare. -p0 shouldn't be needed but you may need to adjust the patch filenames in the patches (easily done) to make them apply automatically.

Let me know if you need any help.

I'd also then drop the src_prepare definition as it shouldn't be needed.

app-misc/pwsafe/pwsafe-0.2.0-r5.ebuild
```diff
@@ -0,0 +1,44 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+IUSE="X libressl readline"
+
+DEPEND="sys-libs/ncurses:0=
+	!libressl? ( dev-libs/openssl:0= )
+	libressl? ( dev-libs/libressl:0= )
+	readline? ( sys-libs/readline:0= )
+	X? ( x11-libs/libSM
```
thesamesam, line 20: I'd put libSM and libX11 on their own lines

app-misc/pwsafe/pwsafe-0.2.0-r5.ebuild
```diff
@@ -0,0 +1,44 @@
...
+	eapply -p0 "${FILESDIR}/${P}-cvs-1.57.patch"
+	eapply -p0 "${FILESDIR}/${P}-printf.patch"
+	eapply -p0 "${FILESDIR}/${P}-fake-readline.patch"
+	eapply -p0 "${FILESDIR}/${P}-man-page-option-syntax.patch"
+	eapply -p0 "${FILESDIR}/${P}-XChangeProperty.patch"
+	eapply_user
+}
+
+src_configure() {
+	econf $(use_with X x) $(use_with readline)
```
thesamesam, line 37:
```suggestion
	econf \
	    $(use_with X x) \
	    $(use_with readline)
```

app-misc/pwsafe/pwsafe-0.2.0-r5.ebuild
thesamesam: I think _something_ is good because otherwise folks will end up filing bugs about it. Maybe just add the Github project? Or you could archive whatever you can find yourself on Github?

Hawk777: The author has finally replied and said he does not intend to bring the site back, so I have changed `HOMEPAGE` to his Github site. He intends to make a release at some point but has not done so yet, so for now I have pointed `SRC_URI` at `web.archive.org`, which snapshotted the original tarball, so at least `SRC_URI` is now fetchable.

app-misc/pwsafe/pwsafe-0.2.0-r5.ebuild
Hawk777: I thought it better not to modify the patch files that were already there, but if you want then of course that will work. Done.

app-misc/pwsafe/pwsafe-0.2.0-r5.ebuild
Hawk777: OK.

app-misc/pwsafe/pwsafe-0.2.0-r5.ebuild
Hawk777: I thought it was a fairly short line and totally readable as is, but as you don’t like it I’ve changed it.

app-misc/pwsafe/pwsafe-0.2.0-r5.ebuild
```diff
@@ -0,0 +1,47 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DESCRIPTION="A Password Safe compatible command-line password manager"
+HOMEPAGE="https://github.com/nsd20463/pwsafe"
+SRC_URI="http://web.archive.org/web/20171006105548if_/http://nsd.dyndns.org/${PN}/releases/${P}.tar.gz"
```
juippis, line 8: `https`

Since it is mirrored, I wouldn't bother updating SRC_URI. But have you checked that the manifests match?

There's also no need to use `${PN}` in SRC_URI since it's not a changeable variable.

juippis (commented):
Sorry but looking at all these patches, it looks more complicated than a single "revbump". Can you update the ebuild name to be something like `pwsafe-0.2.0_p20201002.ebuild`? Do note you'll then have to use something like `${PV/*_p/}` or introduce it as `MY_PV`, then it should work just fine.

app-misc/pwsafe/pwsafe-0.2.0-r5.ebuild
Hawk777: > `https`

Changed.

> Since it is mirrored, I wouldn't bother updating SRC_URI.

Does that mean you want me to change it back? I can do that, but @thesamesam seemed to prefer that SRC_URI be fetchable.

> But have you checked that the manifests match?

They do.

> There's also no need to use `${PN}` in SRC_URI since it's not a changeable variable.

Sorry, I saw hundreds of other packages doing that so I thought it was common practice. Changed.

> Sorry but looking at all these patches, it looks more complicated than a single "revbump".

I didn’t add any new patches. Those new files are patches that already applied to -r4, just with their headers changed from p0 format to p1 format, as @thesamesam requested so that I could use the `PATCHES` array instead of calls to `eapply`. `git show -C -C` proves that only the filename headers changed when I copied those patch files, none of the content.

juippis: > 
> > Since it is mirrored, I wouldn't bother updating SRC_URI.
> 
> Does that mean you want me to change it back? I can do that, but @thesamesam seemed to prefer that SRC_URI be fetchable.
> 
> > But have you checked that the manifests match?
> 
> They do.

Both are fine for me especially if the source hash matches.

> 
> > There's also no need to use `${PN}` in SRC_URI since it's not a changeable variable.
> 
> Sorry, I saw hundreds of other packages doing that so I thought it was common practice. Changed.

Don't worry, it's not a strict rule really. Just a "common sense" one, 
https://projects.gentoo.org/qa/policy-guide/ebuild-format.html#pg0103
https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide/Style_Guide#Plaintexting_.24.7BPN.7D

> 
> > Sorry but looking at all these patches, it looks more complicated than a single "revbump".
> 
> I didn’t add any new patches. Those new files are patches that already applied to -r4, just with their headers changed from p0 format to p1 format, as @thesamesam requested so that I could use the `PATCHES` array instead of calls to `eapply`. `git show -C -C` proves that only the filename headers changed when I copied those patch files, none of the content.

I'm sorry but that makes little sense to me, maybe the idea was that the previous files could be immediately dropped, but it's not possible because the -r4 is stabilized. We don't want to duplicate patches, suddently you'd have a set of 2x5 identical files in the tree.

But worry not, I will merge this today cherry-picking your file from https://github.com/gentoo/gentoo/commit/b961b1b6d73672afa7ff714873de778a0800eefe and adding whatever other fixes you did. So no action required from you anymore, just patience. Thanks!

juippis (approved):

juippis:
4c5494f2a26d6a80be29e9d1faf942fe4295a9e6

PR #17358, opened 2020-09-01
media-libs/giflib: adjust DEPEND -> BDEPEND

jsmolic:
Closes: https://bugs.gentoo.org/734638
Signed-off-by: David Michael <fedora.dm0@gmail.com>
Signed-off-by: Jakov Smolic <jakov.smolic@sartura.hr>


thesamesam:
But what about the:
" 2. The release archive apparently includes the pregenerated documentation files, so the "emake -C doc" doesn't do anything.  Unless users patch the source doc files, xmlto never runs in the first place."

part?

jsmolic:
> But what about the:
> " 2. The release archive apparently includes the pregenerated documentation files, so the "emake -C doc" doesn't do anything. Unless users patch the source doc files, xmlto never runs in the first place."
> 
> part?

Yeah, I'll have to see about patching that up, just forgot to mark this PR as draft

PR #17359, opened 2020-09-01
lxde-base/lxterminal: port to EAPI 7

jsmolic:
Closes: https://bugs.gentoo.org/739838
Signed-off-by: Jakov Smolic <jakov.smolic@sartura.hr>

### DIFF
--- a/lxde-base/lxterminal/lxterminal-0.3.2-r1.ebuild
+++ b/lxde-base/lxterminal/lxterminal-0.3.2-r1.ebuild
@@ -1,0 +1,36 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools xdg
+
+DESCRIPTION="Lightweight vte-based tabbed terminal emulator for LXDE"
+HOMEPAGE="https://wiki.lxde.org/en/LXTerminal"
+SRC_URI="mirror://sourceforge/lxde/${P}.tar.xz"
+
+LICENSE="GPL-2"
+SLOT="0"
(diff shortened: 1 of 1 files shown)


lxde-base/lxterminal/lxterminal-0.3.2.ebuild
```diff
@@ -1,14 +1,14 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
```
juippis, line 4: EAPI bump requires a revbump. And I'm looking at the ebuild and it's awful, so you'd need to fix a few things before it's accepted.

So just inherit `xdg` and nothing else, or fix the ebuild with a revbump.

jsmolic: Ok, I would like to fix it properly then.\
Can you tell me what else needs to be changed?

juippis: There's no implication that it's installing translation files, I'd put building them behind `nls` USE flag and well build everything instead of having the `PLOCALES` array with it. Those files take a second to build and we're already recommending to use `INSTALL_MASK` for unwanted localization files. This'd allow removing *l10n.eclass* altogether.

Then it has a case for a -9999 ebuild but no such ebuild exists. 

It also has double KEYWORDS line, which breaks some tools.

Also do multi-line the `gtk3` USE conditionals for clarity.

jsmolic: Thanks, I cleaned it up. I didn't find the option to disable building locales with configure options, but I think it's okay to install them unconditionally.

juippis (approved):

PR #17360, opened 2020-09-01
gui-apps/gtkgreet: various fixes

kchibisov:
This commit fixes the git url for live ebuild, since the one that
was there before was giving 404, and it also makes 'gtk-layer-shell'
optional via existing 'layershell' use flag.

Signed-off-by: Kirill Chibisov <contact@kchibisov.com>

### DIFF
--- a/gui-apps/gtkgreet/gtkgreet-9999.ebuild
+++ b/gui-apps/gtkgreet/gtkgreet-9999.ebuild
@@ -10,7 +10,7 @@ inherit meson
 
 if [[ ${PV} == 9999 ]]; then
 	inherit git-r3
-	EGIT_REPO_URI="https://git.sr.ht/~kennylevinsen/gtkgreet.git"
+	EGIT_REPO_URI="https://git.sr.ht/~kennylevinsen/gtkgreet"
 else
 	SRC_URI="https://git.sr.ht/~kennylevinsen/gtkgreet/archive/${PV}.tar.gz -> ${P}.tar.gz"
 	KEYWORDS="~amd64 ~ppc64 ~x86"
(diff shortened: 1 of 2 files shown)


gui-apps/gtkgreet/gtkgreet-0.6.ebuild
```diff
@@ -37,7 +37,7 @@ BDEPEND="
 src_configure() {
 	local emesonargs=(
 		$(meson_feature man man-pages)
-		$(meson_use layershell)
+		$(meson_use layershell layershell)
```
epsilon-0, line 40: Don't need the second `layershell`.

gui-apps/gtkgreet/gtkgreet-9999.ebuild
```diff
@@ -37,7 +37,7 @@ BDEPEND="
 src_configure() {
 	local emesonargs=(
 		$(meson_feature man man-pages)
-		$(meson_use layershell)
+		$(meson_use layershell layershell)
```
epsilon-0, line 40: same

gui-apps/gtkgreet/gtkgreet-0.6.ebuild
kchibisov: Sure, I did it for consistent look initially.

kchibisov: Fixed.

epsilon-0:
There needs to be a version bump to `0.6.0-r1`

epsilon-0:
Look here:https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html

kchibisov:
>Look here:https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html

I decided to avoid version bump initially, since it has `+layershell` by default. But if you think that it makes sense I can do it.

kchibisov:
>Yes, this potentially removes a dependency hence it needs a revision bump.

Sure, done.

PR #17363, opened 2020-09-01
java-pkg-simple.eclass: Fix error message when copying resources

hsudbrock:
I stumbled upon this minor issue in the error message when playing around with the java-ebuilder:
* I think that the `not` is missing in the error message, and should definitely be added
* Adding the resource folder for which copying failed is just nice-to-have, but useful IMHO.


fordfrog:
yes, good catch and improvement, thanks!

PR #17368, opened 2020-09-02
app-text/cmark, dev-python/construct: add support for new python

winterheart:
Added python3_8 and python3_9.


juippis (approved):
cmark fails tests, but it does even without your addition so I guess it's fine.

PR #17369, opened 2020-09-02
media-gfx/gnofract4d: update HOMEPAGE and desktop mimeinfo cache 

jsmolic:
xdg_desktop_database_update is called automatically with xdg eclass, so
close off reported bugs

Closes: https://bugs.gentoo.org/740018
Closes: https://bugs.gentoo.org/740002

### DIFF
--- a/media-gfx/gnofract4d/gnofract4d-4.0.1_p1-r1.ebuild
+++ b/media-gfx/gnofract4d/gnofract4d-4.0.1_p1-r1.ebuild
@@ -12,8 +12,8 @@ inherit distutils-r1 eutils virtualx xdg
 COMMIT=bd600c20921afff7b02fc0a76ab79242ebd0896d
 
 DESCRIPTION="A program for drawing beautiful mathematically-based images known as fractals"
-HOMEPAGE="http://edyoung.github.io/gnofract4d/"
-SRC_URI="https://github.com/edyoung/${PN}/archive/${COMMIT}.tar.gz -> ${P}.tar.gz"
+HOMEPAGE="https://github.com/fract4d/gnofract4d"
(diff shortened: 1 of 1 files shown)


cjmayo:
The bugs are for gnofract4d-4.0.1_p1-r1 - something needs to change to fix them.

i.e. xdg_pkg_postinst added to pkg_postinst.

jsmolic:
> The bugs are for gnofract4d-4.0.1_p1-r1 - something needs to change to fix them.
> 
> i.e. xdg_pkg_postinst added to pkg_postinst.

That is already handled by the xdg eclass. `xdg_pkg_preinst()`, `xdg_pkg_postinst()` and `xdg_pkg_postrm()` are called automatically

cjmayo:
> > The bugs are for gnofract4d-4.0.1_p1-r1 - something needs to change to fix them.
> > i.e. xdg_pkg_postinst added to pkg_postinst.
> 
> That is already handled by the xdg eclass. `xdg_pkg_preinst()`, `xdg_pkg_postinst()` and `xdg_pkg_postrm()` are called automatically

When an ebuild has its own pkg_postinst that takes precedence - the bug reports demonstrate that. (Another eclass that provides a pkg_postinst would also if it is listed after xdg in the inherit line).

`qgrep -v xdg_pkg_postinst` shows hundreds of examples.

jsmolic:
> When an ebuild has its own pkg_postinst that takes precedence - the bug reports demonstrate that. (Another eclass that provides a pkg_postinst would also if it is listed after xdg in the inherit line).
> 

Yes, thank you, I did not take that into account

juippis (approved):

PR #17389, opened 2020-09-02
net-libs/libwebsockets: set LWS_DBUS_INCLUDE2

jsmolic:
CMake only looks for dbus-arch-deps.h at /usr/lib64, so we have to
manually set LWS_DBUS_INCLUDE2 for 32-bit systems where it is located at
/usr/lib

Closes: https://bugs.gentoo.org/740090


jsmolic:
@thesamesam This will be properly fixed in version 4.1 (should be out soon). I think this will be good enough workaround for 4.20 to get it stable.\
On the bright side, at least it's not another USE flag conflict ;)

PR #17394, opened 2020-09-03
games-board/xgammon fix endgame database not found

lmiphay:
Change where the program looks by default for its end game db.

Without this fix, the program reports this on startup:

$ xgammon
endgame database not found
you should create one

Signed-off-by: Paul Healy <lmiphay@gmail.com>


chewi:
That is not the right way to fix this as `lib64` is not correct in every case. This should probably use something like `@LIBDIR@` in conjunction with `sed`. It should also be prefixed with `${EPREFIX}` in the same way.

lmiphay:
Thanks chewi - that does make sense.

For EPREFIX, that appears to be already present in: x11-misc/xorg-cf-files

PR #17397, opened 2020-09-03
app-misc/timew: vbump 1.4.2

janko-j1:
Signed-off-by: David Denoncin <ddenoncin@gmail.com>


thesamesam:
By the way, I think you may want to have a look at https://github.com/GothenburgBitFactory/timewarrior/issues/380 and see if you could help. I haven't looked into the issues they were facing but it'd be great to be able to sort them out. Let me know if any assistance from my end is needed.

Also, could we find a way to run tests? Unfortunately the release tarballs don't include them though.

PR #17399, opened 2020-09-03
dev-lang/erlang: bump up to 23.0.3

arkamar:
`repoman` changed `WX_GTK_VER` to `3.0-gtk3` because `x11-libs/wxGTK:3.0[X,opengl]` is deprecated, but everything seems to work correctly.


trofi:
Looks perfect! Pulled as is.

Thank you!

PR #17400, opened 2020-09-03
dev-tex/chktex: Added patch for incorrect exit status.

jeeger:
chktex has a bug where it returns an incorrect exit status in some
circumstances. The patch from
http://svn.savannah.nongnu.org/viewvc/chktex/trunk/chktex/FindErrs.c?r1=308&r2=314&view=patch
fixes this.

Signed-off-by: Jan Seeger <jan.seeger@thenybble.de>


Zlogene:
The package is going to be removed

PR #17401, opened 2020-09-03
dev-util/{cutter,radare2}: version bump

arkamar:
I decided to bump both in one PR because `cutter-1.12.0` depends directly on `radare2-4.5.1`.


trofi:
Pulled as is.

Thank you!

PR #17404, opened 2020-09-03
x11-misc/openbox-menu: update ebuild

winterheart:
Updated SRC_URI (#739808), migrated to EAPI 7.
Closes: https://bugs.gentoo.org/739808


a17r (approved):
Two major problems:
- PR is dropping the only stable version
- PR in its current form is installing to `/bin/openbox-menu`

I've got a fix for that queued up, but I am also using the opportunity to add an upstream patch for port to GTK3. Only picking your first commit (minus 0.8.0 changes) as stable 0.8.0 cannot be cleaned up like that.

PR #17405, opened 2020-09-03
dev-python/pgspecial-1.11.10: version bump

acerisara:
Most tests are skipped by the runner because upstream is expecting a PostgreSQL instance running on localhost, not sure I can do anything about it.


dev-python/pgspecial/pgspecial-1.11.10.ebuild
```diff
@@ -0,0 +1,25 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8} )
```
jsmolic, line 6: Does 3.9 work as well?

dev-python/pgspecial/pgspecial-1.11.10.ebuild
```diff
@@ -0,0 +1,25 @@
...
+inherit distutils-r1
+
+DESCRIPTION="Python implementation of PostgreSQL meta commands"
+HOMEPAGE="https://github.com/dbcli/pgspecial"
+SRC_URI="https://github.com/dbcli/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
```
jsmolic, line 11: Don't use `${PN}` in SRC_URI :)
(https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide/Style_Guide#Plaintexting_.24.7BPN.7D)

dev-python/pgspecial/pgspecial-1.11.10.ebuild
```diff
@@ -0,0 +1,25 @@
...
+
+LICENSE="BSD MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
```
jsmolic, line 16: Empty `IUSE` can be dropped

dev-python/pgspecial/pgspecial-1.11.10.ebuild
acerisara: Good point, I'll do some testing.

acerisara: I did some testing locally and everything looks good to me, I've added the support for `3.9` as well, although upstream doesn't explicitly mention `python3_{8,9}` as supported.

jsmolic: Thanks! Often times it works without many changes needed even if the upstream does not explicitly declare support for e.g. 3.9. It's better to add 3.9 support now if possible rather than revisiting it again after a couple months :)

juippis (approved):

PR #17409, opened 2020-09-04
media-fonts/iosevka: Ebuild bump to 3.4.7, added variants.

jeeger:
Added the "extras" use flag to install all named variants of Iosevka,
namely slab, curly, curly slab, aile, etoile and sparkle.

Signed-off-by: Jan Seeger <jan.seeger@thenybble.de>


juippis (commented):
Not really a fan of this solution. It becomes a maintenance burden really fast.

Isn't there a single tarball that contains all fonts? Maybe it'd be better to just download the source code and build all the variants within the ebuild?

jeeger:
> Not really a fan of this solution. It becomes a maintenance burden really fast.

Yeah, I was going from the original ebuild. I used some AWK magic to build the SRC_URI, but that doesn't really scale.

> Isn't there a single tarball that contains all fonts? Maybe it'd be better to just download the source code and build all the variants within the ebuild?

Building requires `node.js` unfortunately. There is a zip that contains all variants, so maybe I can just install all the variants?

I'll bump a new ebuild shortly.

juippis:
> 
> As far as I can see, there is no file that contains all variants. Building unfortunately requires `node.js`, I don't know if that's an acceptable dependency for users that just want to install a font.
> 

Thanks for opening a bug upstream. We're better off if this gets resolved there.

I'd personally add `build-extras` USE flag that installs all variants and builds it from the source, then just describe the USE flag in *metadata.xml* file. That shouldn't surprise anyone.

thesamesam:
Could you rebase the PR / open a new one for the variants issue?

PR #17413, opened 2020-09-04
dev-java/icedtea-web: Bump to version 1.8.4

amoskvin:


gyakovlev:
interesting.
upstream ships shell launcher alongside with rust launchers, I have a feeling that it should be a useflag.
does new launcher respect gentoo's `eslect java-vm system|user` switching?

amoskvin:
The --enable-shell-launchers option installs the "deprecated" shell
launchers alongside the rust ones. There's no option to disable rust
ones, which makes it pointless.

    /usr/bin/itweb-javaws
    /usr/bin/itweb-javaws.sh
    /usr/bin/itweb-settings
    /usr/bin/itweb-settings.sh
    /usr/bin/policyeditor
    /usr/bin/policyeditor.sh

For eselect, I only have only icedtea:8 right now, but I see:

    selected jre: /home/alec/.gentoo/java-config-2/current-user-vm/jre/

so I would assume it does.

gyakovlev:
ok I'm testing it right now.
there are many issues with ebuild (mostly old stuff we no longer need, and some QA issues) mind if I commit yours as is + my fixes on top?
I don't want to torture you with all the nits, just want working ebuild committed asap, which will be based directly on your work in this pr.

gyakovlev:
thank you for initial ebuild!
it made it a lot easier.

amoskvin:
Thanks!

Unfortunately -r1 doesn't build for me.

1. The configure script gets confused when I have `_JAVA_OPTIONS=-Dawt.useSystemAAFontSettings=on` env variable set. It was previously filtered out by java-pkg-2 (via java-utils-2.eclass).
```
checking JDK version... ./configure: line 5088: test: Picked up
_JAVA_OPTIONS: : integer expression expected
Picked up _JAVA_OPTIONS:
./configure: line 5093: test: Picked up _JAVA_OPTIONS: : integer expression expected
./configure: line 5095: test: Picked up _JAVA_OPTIONS: : integer expression expected
configure: error: JDK8 or newer is required, detected was: Picked up _JAVA_OPTIONS:
```
2. bash-completions are in /etc not /usr/share, so this fails. `export bashcompdir=$(get_bashcompdir)` in src_configure fixes it.
```
removed '/var/tmp/portage/dev-java/icedtea-web-1.8.4-r1/image/usr/bin/itw-modularjdk.args'
rename: /var/tmp/portage/dev-java/icedtea-web-1.8.4-r1/image/usr/share/bash-completion/completions/*.bash: not accessible: No such file or directory
```
Also, I suppose the cs/de/pl manpages for javaws should also be renamed.

gyakovlev:
for `_JAVA_OPTIONS` I recommend having `/etc/portage/bashrc` with `unset _JAVA_OPTIONS`
this variable breaks a lot of ebuilds, some unset it explicitly, but better do it once and for all.
I will add unset in in ebuild as well.

bash-completions are certainly in `/usr/share`, how did you get them in `/etc`?
it uses `pkg-config`
```
PKG_CHECK_VAR(bashcompdir, [bash-completion], [completionsdir], ,
     bashcompdir="${sysconfdir}/bash_completion.d")
AC_SUBST(bashcompdir)
```
and defaults to `/etc` only if `pkg-config` result is empty.

what does 
`pkg-config --variable=completionsdir bash-completion` print on your system? 
for me it's `/usr/share/bash-completion/completions`


ah, I got it, you probably don't have it installed.  that edge-case needs fixing.

gyakovlev:
ok build issues should be fixed now, thanks for catching!

will rename remaining manpages a bit later, since it requires a revbump.
did not see other manpages, because I filter them out on system I tested on.

amoskvin:
Yep, I don't have it so pkg-config returns nothing for me.

The new ebuild works for me. Thanks a lot!

PR #17415, opened 2020-09-04
sci-electronics/iverilog: fix compile error + clean

xdch47:
Closes: https://bugs.gentoo.org/734760


sci-electronics/iverilog/iverilog-10.3-r2.ebuild
```diff
@@ -0,0 +1,76 @@
...
+# If you are building from git, you will also need gperf to generate
+# the configure scripts.
+RDEPEND="
+	sys-libs/readline:=
+	sys-libs/zlib
+"
+
+DEPEND="
+	dev-util/gperf
+	sys-devel/bison
```
thesamesam, line 38: Are flex and bison BDEPEND?

Looks like gperf should be too.

sci-electronics/iverilog/iverilog-10.3-r2.ebuild
```diff
@@ -0,0 +1,76 @@
...
+}
+
+src_install() {
+	local DOCS=( *.txt )
+	# Default build fails with parallel jobs,
+	# https://github.com/steveicarus/iverilog/pull/294
+	emake installdirs DESTDIR="${D}"
+	default
+
+	if use examples; then
```
thesamesam, line 72: How big are the examples?

thesamesam (changes requested):
Thanks. Left a few comments.

sci-electronics/iverilog/iverilog-10.3-r2.ebuild
xdch47: The size of the examples is 104 kB (compressed - bz2) / 144 kB (uncompressed). Starting from which size the use of the example flag is justified / the use flag can be dropped?

thesamesam: It's relative to the package's installed size, compile time, and how useful they're likely to be. Could you tell me those?

xdch47: Total size of the package is about 6.3M.
The `examples` use flags has been there ever since in the repo (since iverilog-0.9.6 / migration to git)

thesamesam: I'll just drop it on merge.

PR #17416, opened 2020-09-04
dev-haskell/{happy,mysql,regex-posix}: remove unused patches

mm1ke:
Hi,

small cleanup for some haskell related packages.


trofi:
Pulled as is.

Thank you!

PR #17418, opened 2020-09-04
dev-haskell/*, x11-misc/xmonad-contrib: use https

mm1ke:
Hi,

some updates to haskell ebuilds to use https instead of http.


trofi:
Pushed as is.

Thank you!

PR #17419, opened 2020-09-04
net-dns/bind: fix LDAP library path on multilib machines

atomsymbol:
Please merge. Thanks.

Closes: https://bugs.gentoo.org/739062


chutz:
@atomsymbol Please update this to apply against `bind-9.16.6-r1.ebuild`, and revbump the ebuild to `bind-9.16.6-r2.ebuild` in the process.

PR #17420, opened 2020-09-04
Some X11 EAPI 7 updates

dm0-:
These packages have multilib dependencies but do not have an EAPI 7 ebuild.  That breaks cross-compiling to a multilib target (if you don't specify `--root-deps=rdeps` which breaks other things).  Fixing this package set allowed me to cross-compile a graphical multilib system.  The few `x11-apps` packages weren't breaking anything, but I randomly updated them while testing, so they're included here.  (There are some related bugs like [bug #740414](https://bugs.gentoo.org/740414) for core packages with this issue.)

### DIFF
--- a/x11-apps/iceauth/iceauth-1.0.8-r1.ebuild
+++ b/x11-apps/iceauth/iceauth-1.0.8-r1.ebuild
@@ -1,0 +1,14 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit xorg-3
+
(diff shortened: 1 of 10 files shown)


rossbridger:
I don't think there should be a revision change if it doesn't involve USE change/bugfix, saving everyone from unnecessary rebuilding.

dm0-:
I made new ebuilds to be consistent with other changes (e.g. #16935).  If @mattst88 doesn't want them this time, I'll update the PR.

mattst88:
Thanks!

Sorry about the rebuilds. It's kind of required for EAPI changes in most cases. Fortunately all of these packages are quite small so the cost is very low.

PR #17425, opened 2020-09-05
java-pkg-simple.eclass: Fix typos and comments

hsudbrock:
I went over `java-pkg-simple.eclass` and fixed some typos and some comments in the eclass on the way (see explanation for non-typo changes below). Please feel free to reject this PR if reviewing such relatively small changes without a huge benefit takes time that should be better spent elsewhere (I won't feel rejected in that case) - I am simply trying to fix things on the way as good as I can while trying to get a grip on Gentoo Java support. 

Non-typo changes:
* After the recent extensions of this eclass, it now supports resources - hence
  the change in the eclass description.
* The usage description of `java-pkg-simple_getclasspath` mentions the optional
  parameter `--runtime-only`, which appears not to be provided by the function.
* The copyright header should be until `2020` now, I presume.

Signed-off-by: Henning Sudbrock <github@hsudbrock.de>

### DIFF
--- a/eclass/java-pkg-simple.eclass
+++ b/eclass/java-pkg-simple.eclass
@@ -1,4 +1,4 @@
-# Copyright 2004-2015 Gentoo Foundation
+# Copyright 2004-2020 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
 # @ECLASS: java-pkg-simple.eclass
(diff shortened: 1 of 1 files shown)


fordfrog:
hi, nice catches! :-) thank you for the pr. it looks good imo, i just want @IsotropicZ to review the non-functional switch that you dropped for now.

6-6-6:
> hi, nice catches! :-) thank you for the pr. it looks good imo, i just want @IsotropicZ to review the non-functional switch that you dropped for now.

Hello, thanks for the patch and for keeping me involved.
I forgot to delete the switch after I move the code to another place.
The related code of the switch is now placed in
https://github.com/gentoo/gentoo/blob/master/eclass/java-pkg-simple.eclass#L330

So imo the switch could be safely removed.
Cheers!

hsudbrock:
Thanks for checking, @IsotropicZ !

PR #17433, opened 2020-09-05
dev-util/watchman: New package

ThiefMaster:
[watchman](https://facebook.github.io/watchman/) is quite useful as a developer when you want to stuff to happen on file changes.

v4.9.0 is quite old, but still the most widely used version (homebrew has this one, and the official "build from source" docs uses that version as well).

They added release notes for a newer version recently (no updated build docs though), but building it without using their custom tool that fetches all dependencies (including big ones like boost) seems like a major hassle (facebook/watchman#851 has details), and likely not worth it considering that there don't appear to be any important changes besides Python 3 versions of the extra tools provided on top of watchman (watchman-make and watchman-wait).


dev-util/watchman/watchman-4.9.0.ebuild
```diff
@@ -0,0 +1,56 @@
+# Copyright 2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+PYTHON_COMPAT=( python2_7 )
```
hydrapolic, line 5: There is an ongoing effort to remove python2 stuff, please try a newer version with python3 support.

ThiefMaster: The problem is that the newer version they have is [extremely building-unfriendly](https://github.com/facebook/watchman/issues/851#issuecomment-680272404), at least without downloading build dependencies during the build which seems incredibly awful (and blocked by the portage sandbox anyway).

One (likely) easy fix I see would be simply patching the python scripts to work on Python 3 since it's just the CLI scripts that are broken...

ThiefMaster: I've added patches for python 3 compatibility (based on the changes from upstream)

hydrapolic: Maybe those deps can be added to SRC_URI.

ThiefMaster: They don't exist as standalone patches, I got them from the individual git commits between the two tags (minus the ones that just run a python code formatter)

hydrapolic: I meant regarding the build dependencies you mentioned in the first comment (not the patches you added).

monsieurp (approved):
Thanks for the PR. It looks like an interesting program. I will co-maintain it with you. Reworking the PR a bit and then merging it.

ThiefMaster:
Awesome!

PR #17435, opened 2020-09-06
dev-util/cgdb: respect AR

jsmolic:


dev-util/cgdb/cgdb-0.7.0-r2.ebuild
```diff
@@ -40,6 +40,8 @@ DOCS=( AUTHORS ChangeLog FAQ INSTALL NEWS README.md )
 PATCHES=(
 	# Bugs: #730138, #678006, #630512, #682386
 	"${FILESDIR}/${PN}-test.patch"
+	# Bug: #724256
+	"${FILESDIR}/${PN}-env.patch"
 )
 
 src_prepare() {
```
thesamesam, line 47: I think I _might_ have asked this before but we shouldn't call autogen.sh usually as they may not recall auto* in the right order for Gentoo, or allow us to apply the various e.g. libtool patches we want.

eautoreconf should do the job, even if we need to apply some magic variables from the [eclass](https://devmanual.gentoo.org/eclass-reference/autotools.eclass/index.html).

Let me know what the background here is because I'm worried I'm not remembering something which we discussed ;)

dev-util/cgdb/cgdb-0.7.0-r2.ebuild
```diff
@@ -40,6 +40,8 @@ DOCS=( AUTHORS ChangeLog FAQ INSTALL NEWS README.md )
 PATCHES=(
```
thesamesam, line 40: Also, is there a real benefit in having multilib here?

dev-util/cgdb/cgdb-0.6.8.ebuild
```diff
@@ -34,6 +34,11 @@ RDEPEND="
 
 DOCS=( AUTHORS ChangeLog INSTALL NEWS README.md TODO )
 
+PATCHES=(
+	# Bug: #724256
+	"${FILESDIR}/${PN}-env.patch"
```
thesamesam, line 39: Pop in the version you based on, like: ${PN}-0.6.8-respect-AR.patch (provided it was based on 0.6.8).

Also, could you add some information to the patch on its upstream status and a reference to the bug?

dev-util/cgdb/cgdb-0.7.0-r2.ebuild
jsmolic: Ah, yes, I remember, I think we discussed it [here](https://github.com/gentoo/gentoo/pull/17157#discussion_r472937395). This ebuild was using `autogen.sh` before I picked it up, so I didn't touch it immediately.


Okay, so I seem to be having some issues when using default `eautoreconf,`
```
/var/tmp/portage/dev-util/cgdb-0.7.1/work/cgdb-0.7.1/configure: line 6055: RL_LIB_READLINE_VERSION: command not found
configure: error: CGDB requires GNU readline 5.1 or greater to link.
```
It seems like it cannot find readline, even though it is present, and that doesn't happen when using `autogen.sh`.\
I think the issue is that `autogen.sh` runs `aclocal -I config`, while with `eautoreconf` we only call `aclocal`, and the `./config` directory contains [m4 logic for checking readline version](https://github.com/cgdb/cgdb/blob/master/config/readline_check_version.m4).

I will paste a diff below that works for me, it is esentially calling all `eautoreconf` functions, only we now run `aclocal -I config`. Can you please tell me if this is this the "normal" way of changing the `eautoreconf` functions, or is there a more clever solution for this :)

```diff
--- a/dev-util/cgdb/cgdb-0.7.1.ebuild
+++ b/dev-util/cgdb/cgdb-0.7.1.ebuild
@@ -12,7 +12,7 @@ else
        SRC_URI="https://github.com/cgdb/cgdb/archive/v${PV}.tar.gz -> ${P}.tar.gz"
        KEYWORDS="~amd64 ~arm ~ppc ~ppc64 ~x86 ~amd64-linux ~x86-linux"
 fi
-inherit multilib-minimal
+inherit autotools multilib_minimal

 DESCRIPTION="A curses front-end for GDB, the GNU debugger"
 HOMEPAGE="https://cgdb.github.io/"
@@ -45,8 +45,13 @@ PATCHES=(
 )

 src_prepare() {
+       cp configure.{init,ac} || die "cp failed"
        default
-       ./autogen.sh || die
+       eaclocal -I config
+       eautoconf --force
+       eautoheader
+       eautomake --add-missing --copy --foreign --force-missing
+       elibtoolize
 }
```

dev-util/cgdb/cgdb-0.6.8.ebuild
jsmolic: Sure, the fix is upstreamted, I took the patch from git, I will add some information about it to the patch file

dev-util/cgdb/cgdb-0.7.0-r2.ebuild
jsmolic: Hm, not sure really, do you think it should be removed?

jsmolic:
@thesamesam I have changed the autotools stuff and adjusted the patch, let me know about the multilib.\
Thank you for patience and reviews, I always lear a lot from them :)

dev-util/cgdb/cgdb-0.7.0-r2.ebuild
thesamesam: I was just about to comment about ```AT_M4DIR``` because I didn't see the edit!

dev-util/cgdb/cgdb-0.7.0-r2.ebuild
thesamesam: I think so, but we don't have to do it this time. Just for next time around. It removes some complexity from the ebuild.

dev-util/cgdb/cgdb-0.6.8.ebuild
thesamesam: Great!

thesamesam:
@jsmolic Always happy to see one of your contributions! Fast replies and an interesting discussion tends to follow.

I made a minor touchup in eb3b0de81e81e163488b4de04a0bc54018b0eb7b. We'll drop multilib next time around because it's only really useful if it's a dependency of something you _want_ multilib (like Wine).

jsmolic:
@thesamesam Thanks, pleasure to work with you as well, I really appreciate that you take the time to review and explain everything in detail

PR #17437, opened 2020-09-06
xfce-extra/xfce4-systemload-plugin: panel-4.15 compatibility

atomsymbol:
Please merge. Thanks.

Closes: https://bugs.gentoo.org/733952


atomsymbol:
@mgorny Same here, similar to the just merged cpugraph pull request.

xfce-extra/xfce4-systemload-plugin/files/xfce4-sysload-plugin-1.2.3-panel-4.15.patch
```diff
@@ -0,0 +1,25 @@
+From 35b50d425209773c9dd21f125b1bd3f8c3daa46b Mon Sep 17 00:00:00 2001
```
mgorny, line 1: That's not the package name.

PR #17441, opened 2020-09-06
net-mail/notmuch: Bump to version 0.31

rseichter:
Technical changes in Notmuch release 0.31 include support for Emacs 27.1 and Xapian 1.5 (the current development version).


net-mail/notmuch/notmuch-0.31.ebuild
```diff
@@ -0,0 +1,216 @@
...
+		app-doc/doxygen
+		dev-python/mock[${PYTHON_USEDEP}]
+		dev-python/sphinx
+		sys-apps/texinfo
+	)
+"
+COMMON_DEPEND="
+	dev-libs/glib
+	dev-libs/gmime:3.0[crypt]
+	dev-libs/xapian
```
thesamesam, line 43: I restored := as we link against xapian and it exposes a subslot.

net-mail/notmuch/notmuch-0.31.ebuild
```diff
@@ -0,0 +1,216 @@
...
+	sys-libs/talloc
+	sys-libs/zlib
```
thesamesam, line 45: Didn't bother adding for zlib because it's not clear if the subslot will be removed there.

thesamesam:
Thanks!

PR #17445, opened 2020-09-06
net-firewall/fwbuilder: add live ebuild

nilathak:
Signed-off-by: Hannes Schweizer <hschweizer@gmx.net>


juippis:
Hi, 
I'm sorry but we won't be adding live ebuilds for unmaintained packages. They can break any day and maintainer-needed packages aren't really high on anyone's prioritities. Why do you need this? It's hosted in Github so you can take a snapshot from a specific commit and package that. Although looks like there's a major version release, so you could just make an ebuild for that?

nilathak:
Hi,
The latest version in gentoo (5.3.7) is segfaulting during certain operations.
The latest major release (6.0.0_beta) works for me, but needs additional ebuild workarounds for qmake (see fwbuilder/fwbuilder#63).
A version with Cmake support has not been released yet (fwbuilder/fwbuilder@5b2d0e297d5e47f4bcb7e551830848fd713042ab).

I've created a new PR (#17771) with a snapshot ebuild (I looked at games-emulation/dolphin for guidance).

PR #17446, opened 2020-09-06
mail-filter/amavisd-new: Depend on app-arch/arj

rseichter:
Depend on app-arch/arj instead of app-arch/unarj because of licensing.


chewi:
Great but you'll need to address the older versions if we want to last-rite unarj. I know this is a stable package but I don't think pushing this change straight-to-stable would be too controversial.

rseichter:
@chewi I have added a commit that removes the outdated ebuilds. As for going straight to stable, I am not so sure. If you think it is really OK, you can make this change before committing the changes, correct?

chewi:
Turns out we can't do that yet because arj isn't stable on ppc64. I've pushed this now so I'll get that sorted and stabilise this afterwards.

rseichter:
Thanks for checking, and for processing the merge request.

PR #17450, opened 2020-09-07
sys-kernel/gentoo-kernel: enable /proc/config.gz

gyakovlev:
Just enabling on latest kernels on 5.4 and 5.8 branch, so it gets into next bump/stable.

see no reason to enable on EOL 5.7, as it should go away soon-ish.

/proc/config.gz not only useful for users, also linux-info.eclass uses it to check kernel config if kernel src not installed (chroot)


mgorny:
Please don't change this in place. Just make sure it's listed on  https://bugs.gentoo.org/740846 and ping me when new release comes.

mgorny:
As a somewhat weird idea, maybe do it as pseudo-bump to the next version — i.e. bump the ebuild, update config (in the same commit), and I'll regen Manifest when the version is actually released. Or just attach a patch to the ebuild on the bug, as that will probably be easier to tackle than PR.

gyakovlev:
made it a bump

PR #17467, opened 2020-09-08
sys-apps/ibm-powerpc-utils: bump to 1.3.8

dm0-:
This adds `--disable-werror` since it won't cooperate with my compiler version without `CPPFLAGS=-Wno-error=format` and `LDFLAGS=-Wno-error=stack-protector` in the environment.


trofi:
Pushed in a slightly modified form without overriding default git branch.

PR #17468, opened 2020-09-08
media-sound/mumble: Update/Fix live ebuild

DarkDefender:
Upstream now has dropped qmake and uses only cmake.
Updated the ebuild to use cmake instead of qmake.

Also added tests.

Some minor fixes needed to be done to the build system.
I've submitted two pull requests upstream:
https://github.com/mumble-voip/mumble/pull/4466
https://github.com/mumble-voip/mumble/pull/4468

I'm submitting this now so I can get feedback on the ebuild side of things.

### DIFF
--- a/media-sound/mumble/mumble-9999.ebuild
+++ b/media-sound/mumble/mumble-9999.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-inherit desktop multilib-build qmake-utils xdg
+inherit cmake xdg-utils
 
 DESCRIPTION="Mumble is an open source, low-latency, high quality voice chat software"
 HOMEPAGE="https://wiki.mumble.info"
@@ -26,7 +26,8 @@ fi
 
 LICENSE="BSD MIT"
 SLOT="0"
-IUSE="+alsa +dbus debug g15 jack libressl +opus oss pch portaudio pulseaudio +rnnoise speech zeroconf"
+IUSE="+alsa +dbus debug g15 jack libressl portaudio pulseaudio +rnnoise speech test zeroconf"
+RESTRICT="!test? ( test )"
 
 RDEPEND="
 	dev-qt/qtcore:5
@@ -49,7 +50,7 @@ RDEPEND="
 	jack? ( virtual/jack )
 	!libressl? ( >=dev-libs/openssl-1.0.0b:0= )
 	libressl? ( dev-libs/libressl )
-	opus? ( >=media-libs/opus-1.0.1 )
+	>=media-libs/opus-1.3.1
 	portaudio? ( media-libs/portaudio )
 	pulseaudio? ( media-sound/pulseaudio )
 	speech? ( >=app-accessibility/speech-dispatcher-0.8.0 )
@@ -64,91 +65,35 @@ BDEPEND="
 	virtual/pkgconfig
 "
 
-# NB: qmake does not support multilib but it's fine to configure
-# for the native ABI here
 src_configure() {
-	myuse() {
-		[[ -n "${1}" ]] || die "myuse: No use option given"
-		use ${1} || echo no-${1}
-	}
 
-	local conf_add=(
-		bundled-celt
-		no-bundled-opus
-		no-bundled-speex
-		no-embed-qt-translations
-		no-server
-		no-update
-		$(myuse alsa)
-		$(myuse dbus)
-		$(usex debug 'symbols debug' release)
-		$(myuse g15)
-		$(usex jack '' no-jackaudio)
-		$(myuse opus)
-		$(myuse oss)
-		$(myuse portaudio)
-		$(myuse pulseaudio)
-		$(myuse rnnoise)
-		$(usex speech '' no-speechd)
-		$(usex zeroconf '' no-bonjour)
+	local mycmakeargs=(
+		"-Dbundled-celt=ON"
+		"-Dbundled-opus=OFF"
+		"-Dbundled-speex=OFF"
+		"-Doverlay=ON"
+		"-Dtranslations=OFF"
+		"-Dserver=OFF"
+		"-Dupdate=OFF"
+		"-Dalsa=$(usex alsa)"
+		"-Ddbus=$(usex dbus)"
+		"-Dg15=$(usex g15)"
+		"-Djackaudio=$(usex jack)"
+		"-Dportaudio=$(usex portaudio)"
+		"-Dpulseaudio=$(usex pulseaudio)"
+		"-Drnnoise=$(usex rnnoise)"
+		"-Dspeechd=$(usex speech)"
+		"-DBUILD_TESTING=$(usex test)"
+		"-Dzeroconf=$(usex zeroconf)"
 	)
 
-	use pch || conf_add+=( no-pch )
-
-	eqmake5 "${S}/main.pro" -recursive \
-		CONFIG+="${conf_add[*]}" \
-		DEFINES+="PLUGIN_PATH=/usr/$(get_libdir)/mumble"
-}
-
-multilib_src_compile() {
-	local emake_args=(
-		# place libmumble* in a subdirectory
-		DESTDIR_ADD="/${MULTILIB_ABI_FLAG}"
-		{C,L}FLAGS_ADD="$(get_abi_CFLAGS)"
-	)
-	# build only overlay library for other ABIs
-	multilib_is_native_abi || emake_args+=( -C overlay_gl )
-	emake "${emake_args[@]}"
-	emake clean
-}
-
-src_compile() {
-	multilib_foreach_abi multilib_src_compile
-}
-
-multilib_src_install() {
-	local dir=$(usex debug debug release)
-	insinto /usr/$(get_libdir)/${PN}
-	doins "${dir}/${MULTILIB_ABI_FLAG}"/libmumble.so*
-	if multilib_is_native_abi; then
-		dobin "${dir}"/mumble
-		doins "${dir}"/libcelt0.so* "${dir}"/plugins/lib*.so*
-	fi
-}
-
-src_install() {
-	multilib_foreach_abi multilib_src_install
-
-	dodoc CHANGES README.md
-	dobin scripts/mumble-overlay
-
-	insinto /usr/share/services
-	doins scripts/mumble.protocol
-
-	domenu scripts/mumble.desktop
-
-	doicon -s scalable icons/mumble.svg
-
-	doman man/mumble-overlay.1
-	doman man/mumble.1
-}
-
-pkg_preinst() {
-	xdg_pkg_preinst
+	CMAKE_BUILD_TYPE=$(usex debug Debug Release) \
+	cmake_src_configure
 }
 
 pkg_postinst() {
-	xdg_pkg_postinst
+	xdg_icon_cache_update
+	xdg_desktop_database_update
 	echo
 	elog "Visit https://wiki.mumble.info/ for futher configuration instructions."
 	elog "Run mumble-overlay to start the OpenGL overlay (after starting mumble)."
@@ -156,5 +101,6 @@ pkg_postinst() {
 }
 
 pkg_postrm() {
-	xdg_pkg_postrm
+	xdg_icon_cache_update
+	xdg_desktop_database_update
 }
--- a/media-sound/mumble/mumble-9999.ebuild
+++ b/media-sound/mumble/mumble-9999.ebuild
@@ -68,26 +68,39 @@ BDEPEND="
 src_configure() {
 
 	local mycmakeargs=(
+		"-Dalsa=$(usex alsa)"
+		"-DBUILD_TESTING=$(usex test)"
 		"-Dbundled-celt=ON"
 		"-Dbundled-opus=OFF"
 		"-Dbundled-speex=OFF"
-		"-Doverlay=ON"
-		"-Dserver=OFF"
-		"-Dupdate=OFF"
-		"-Dalsa=$(usex alsa)"
-		"-DBUILD_TESTING=$(usex test)"
 		"-Ddbus=$(usex dbus)"
 		"-Dg15=$(usex g15)"
 		"-Djackaudio=$(usex jack)"
+		"-Doverlay=ON"
 		"-Dportaudio=$(usex portaudio)"
 		"-Dpulseaudio=$(usex pulseaudio)"
 		"-Drnnoise=$(usex rnnoise)"
+		"-Dserver=OFF"
 		"-Dspeechd=$(usex speech)"
 		"-Dtranslations=$(usex nls)"
+		"-Dupdate=OFF"
 		"-Dzeroconf=$(usex zeroconf)"
 	)
 
 	cmake_src_configure
+}
+
+src_install() {
+	cmake_src_install
+
+	if use amd64 ; then
+		# The 32bit overlay library gets automatically built and installed on x86_64 platforms.
+		# Install it into the correct 32bit lib dir.
+		libdir_64="/usr/$(get_libdir)/mumble"
+		libdir_32="/usr/$(get_abi_var LIBDIR x86)/mumble"
+		mkdir -p $D/$libdir_32 || die
+		mv $D/$libdir_64/libmumbleoverlay.x86.so* $D/$libdir_32/ || die
+	fi
 }
 
 pkg_postinst() {
@@ -95,7 +108,7 @@ pkg_postinst() {
 	xdg_desktop_database_update
 	echo
 	elog "Visit https://wiki.mumble.info/ for futher configuration instructions."
-	elog "Run mumble-overlay to start the OpenGL overlay (after starting mumble)."
+	elog "Run 'mumble-overlay <program>' to start the OpenGL overlay (after starting mumble)."
 	echo
 }
 
--- a/media-sound/mumble/mumble-9999.ebuild
+++ b/media-sound/mumble/mumble-9999.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-inherit cmake xdg-utils
+inherit cmake xdg
 
 DESCRIPTION="Mumble is an open source, low-latency, high quality voice chat software"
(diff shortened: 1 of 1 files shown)


DarkDefender:
I'm a bit unsure about the multilib nature of this ebuild.

Perhaps it was needed in the past, but now on x86_64, the default cmake build does build a x86 (32bit) overlay library so it can be used in 32bit games.

DarkDefender:
My patches have now been merged upstream so this should be good to go unless there is some changes need to the ebuild.

DarkDefender:
@thesamesam how do I poke the maintainers that are not on github?
It seems like I need to get in touch with `polynomial-c`

Do I file a ticket on the gentoo bug tracker?

Ionic:
NAK from me.

This PR adds a subtle difference: actual support for multilib, i.e., multiple arch builds. This was disabled in the previous version for good reasons: `qmake` can't easily handle multilib stuff (boo), but more importantly, the Qt libraries are not multilib-compatible for some stupid reasons, so this ebuild will fail on any system that enables non-native arches by default. It probably would be good to switch back to `src_configure` again if that makes it fake-multilib for the native arch only.

Also, pinging @Polynomial-C.

Ionic:
Additionally, please make sure to add

`Closes: https://bugs.gentoo.org/741716`

to the commit and read the [Gentoo Certificate of Origin](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin) and sign-off all your commits (`--signoff` option to commit).

DarkDefender:
I've removed the multi lib things and cleaned up the ebuild.

If the `overlay` cmake flag is on, mumble will automatically build a 32bit library if the arch is amd64/x86_64.
So in this case we do not need to do any multi lib magic.

The building of this 32bit library can be turned off with `overlay-xcompile`, but I don't know how useful it will be to expose this in the ebuild. So I now let the build scripts handle this flag.

I have added the requested lines to the commit message.

I've also created a pull request upstream to fix the install rules for the overlay scripts and libs.
As this is a live ebuild I didn't include this patch here as it will probably be merged soon:
https://github.com/mumble-voip/mumble/pull/4495

DarkDefender:
My latest upstream fix is now merged, so this should be good to go.

DarkDefender:
@Ionic Anything preventing this from being merged now?

Ionic:
Mh, not sure. The whole multilib business with the overlay library seems a bit weird, but could be okay, I guess. I'm not sure if it's self-contained or needing other non-native-arch libraries for instance, and given that it's turned on by default, I wonder if it behaves correctly in both cases - foreign arches being enabled and disabled.

I'm not a Gentoo dev, though, merely went through this because of personal interest and reviewed it a bit.

I guess you'll have to get @Polynomial-C's attention for this PR to move on.

Polynomial-C:
This PR is on my todo list but I couldn't give it the attention it deserves yet.
AFAIK, the multilib stuff is important if you have "old" 32bit games that need the libmumble.so for the mumble overlay to work. I hope we can still keep that avaiable because otherwise this could be seen as a regression.

DarkDefender:
The 32bit libs are now automatically built on x86_64 build targets.

So this updated ebuild should still provide support for older 32bit games on a multi lib system.

kripton:
@DarkDefender: Thanks for doing the work on media-sound/mumble! I've created #17991 that does the same to media-sound/murmur. I'll also keep watching this PR to check all remarks by @Polynomial-C  and apply them to my PR as well (if they are valid for murmur)

media-sound/mumble/mumble-9999.ebuild
```diff
@@ -64,97 +65,42 @@ BDEPEND="
...
-		$(myuse pulseaudio)
-		$(myuse rnnoise)
-		$(usex speech '' no-speechd)
-		$(usex zeroconf '' no-bonjour)
+	local mycmakeargs=(
+		"-Dbundled-celt=ON"
+		"-Dbundled-opus=OFF"
+		"-Dbundled-speex=OFF"
+		"-Doverlay=ON"
+		"-Dtranslations=OFF"
```
kripton, line 75: Any reasons why you set this to `OFF`? Does creating the translations need any additional dependencies?

DarkDefender: It is simply mirroring the old settings in the ebuild.
The old flag was "no-embed-qt-translations".

I don't remember if I tried turning it on or not.

But I guess we can have a discussion if we should make this configurable or not if needed.

kripton: Ah, okay. I've just realised that with that setting, mumble is english-only. Setting it to `ON` makes it translated (or English with `LC_ALL=C`). However, since no additional translation-files are installed, I assume they are integrated into the binary.
I will re-check with mumble-1.3.3 tomorrow to see how that behaves.
However, it might be that integrating the translation files might require `dev-qt/linguist-tools` as dependency.

kripton: Okay, I've re-checked mumble-1.3.3: While it is true that the ebuild requests the translations not to be embedded, this doesn't seem to be honored. `lrelease` from `dev-qt/linguist-tools` is called during the build for each language and the resulting binary displays the texts translated. Since the emerge doesn't install any translation files, I assume that the texts are inside the binary (tried `strings` command but didn't find anything. Did not want to investigate how Qt's resource system (qrc) works in detail).
With the CMake-based build system, the configuration option is actually taken into account.
I'd prefer mumble to simply integrate the translations. I wouldn't know how to force them to separate files.

kripton:
And in my murmur-9999-PR (#17991) I got this review that is also valid for your ebuild: https://github.com/gentoo/gentoo/pull/17991#pullrequestreview-515015754

DarkDefender:
@kripton are you sure? When I compile mumble I can see that my flags some make.conf is honored

DarkDefender:
Hmm, I guess we don't want to pull in any extra compile flags, so that is why.

kripton:
> @kripton are you sure? When I compile mumble I can see that my flags from make.conf is honored

Honestly, I simply don't know ;)
Other ebuilds in the tree using the cmake eclass have the debug flag in IUSE but don't have any reference to that flag in the ebuild: https://github.com/gentoo/gentoo/blob/64fe952caa63f4916a5e2f704ebd5306ff00198f/app-admin/keepassxc/keepassxc-2.6.2.ebuild#L27

And the cmake eclass docs mention to not change CMAKE_BUILD_TYPE:
```You usually do *NOT* want nor need to set it as it pulls CMake default build-type specific compiler flags overriding make.conf``` (https://devmanual.gentoo.org/eclass-reference/cmake.eclass/index.html)

**Comment by Polynomial-C** [comment]
So I finally found some time to give this PR some testing. Unfortunately I don't get any 32bit overlay libs on my amd64-multilib system.

**Comment by DarkDefender** [comment]
@Polynomial-C Is does install them on my system. I didn't move them from the default install location as I guess the lib only work with the 64bit client. (But works with 32bit games).

`equery f mumble`:
```
/usr/lib64/mumble
/usr/lib64/mumble/libcelt0.so -> libcelt0.so.0.7.0
/usr/lib64/mumble/libcelt0.so.0.7.0
/usr/lib64/mumble/libmumbleoverlay.x86.so -> libmumbleoverlay.x86.so.1.4.0
/usr/lib64/mumble/libmumbleoverlay.x86.so.1.4.0
/usr/lib64/mumble/libmumbleoverlay.x86_64.so -> libmumbleoverlay.x86_64.so.1.4.0
/usr/lib64/mumble/libmumbleoverlay.x86_64.so.1.4.0
```

File: media-sound/mumble/mumble-9999.ebuild [hunk 2]
```diff
@@ -64,97 +65,42 @@ BDEPEND="
...
+		"-Dalsa=$(usex alsa)"
+		"-Ddbus=$(usex dbus)"
+		"-Dg15=$(usex g15)"
+		"-Djackaudio=$(usex jack)"
+		"-Dportaudio=$(usex portaudio)"
+		"-Dpulseaudio=$(usex pulseaudio)"
+		"-Drnnoise=$(usex rnnoise)"
+		"-Dspeechd=$(usex speech)"
+		"-DBUILD_TESTING=$(usex test)"
+		"-Dzeroconf=$(usex zeroconf)"
```
Polynomial-C, line 87: Can you please adjust the quoting to look like this?
```
-Dfeature1="ON"
-Dfeature2="OFF"
-Dfeature3="$(usex feature3)"
```

media-sound/mumble/mumble-9999.ebuild
Polynomial-C: According to the CMakeLists.txt file `-Dtranslation` triggers other tranlations than english. So please introduce a `nls` USE flag and change the option to `-Dtranslation="$(usex nls)"`

media-sound/mumble/mumble-9999.ebuild
```diff
@@ -49,7 +50,7 @@ RDEPEND="
 	jack? ( virtual/jack )
 	!libressl? ( >=dev-libs/openssl-1.0.0b:0= )
 	libressl? ( dev-libs/libressl )
-	opus? ( >=media-libs/opus-1.0.1 )
+	>=media-libs/opus-1.3.1
```
Polynomial-C, line 53: Move this into the non-conditional block above. Keep it alphabetically sorted please.

media-sound/mumble/mumble-9999.ebuild
Polynomial-C: Oh and please sort the list alphabetically

media-sound/mumble/mumble-9999.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-inherit desktop multilib-build qmake-utils xdg
+inherit cmake xdg-utils
```
Polynomial-C, line 6: Why did you replace `xdg.eclass` with `xdg-utils.eclass`?

Polynomial-C (changes requested):

Polynomial-C:
> @Polynomial-C Is does install them on my system. I didn't move them from the default install location as I guess the lib only works with the 64bit client. (But works with 32bit games).
> 
> `equery f mumble`:
> 
> ```
> /usr/lib64/mumble
> /usr/lib64/mumble/libcelt0.so -> libcelt0.so.0.7.0
> /usr/lib64/mumble/libcelt0.so.0.7.0
> /usr/lib64/mumble/libmumbleoverlay.x86.so -> libmumbleoverlay.x86.so.1.4.0
> /usr/lib64/mumble/libmumbleoverlay.x86.so.1.4.0
> /usr/lib64/mumble/libmumbleoverlay.x86_64.so -> libmumbleoverlay.x86_64.so.1.4.0
> /usr/lib64/mumble/libmumbleoverlay.x86_64.so.1.4.0
> ```

These libs need to go into the 32bit libdir (on 17.0 and older multilib profiles this is `/usr/lib32` and on 17.1 multilib profile this is `/usr/lib`)

DarkDefender:
> > @Polynomial-C Is does install them on my system. I didn't move them from the default install location as I guess the lib only works with the 64bit client. (But works with 32bit games).
> > `equery f mumble`:
> > ```
> > /usr/lib64/mumble
> > /usr/lib64/mumble/libcelt0.so -> libcelt0.so.0.7.0
> > /usr/lib64/mumble/libcelt0.so.0.7.0
> > /usr/lib64/mumble/libmumbleoverlay.x86.so -> libmumbleoverlay.x86.so.1.4.0
> > /usr/lib64/mumble/libmumbleoverlay.x86.so.1.4.0
> > /usr/lib64/mumble/libmumbleoverlay.x86_64.so -> libmumbleoverlay.x86_64.so.1.4.0
> > /usr/lib64/mumble/libmumbleoverlay.x86_64.so.1.4.0
> > ```
> 
> These libs need to go into the 32bit libdir (on 17.0 and older multilib profiles this is `/usr/lib32` and on 17.1 multilib profile this is `/usr/lib`)

It's only one lib `libmumbleoverlay.x86.so`. Is there any automagical way to install into the right place (depending on the profile)?

media-sound/mumble/mumble-9999.ebuild
DarkDefender: I don't really remember, but I think it was because I needed some functions that were not in the xdg class.
After the update I didn't need the xdg eclass

Polynomial-C:
> Is there any automagical way to install into the right place (depending on the profile)?

I don't know any portage function that could help us here. You can get the correct libdir name for x86 by calling `$(get_abi_var LIBDIR x86)` Which either returns `lib32` for 17.0 and older amd64-multilib profiles and `lib` for 17.1 amd64-multilib profiles. Perhaps an additional cmake option that overrides the default install location for the 32bit overlay files could be useful here. Something like: `-D32bit_libdir="/usr/$(get_abi_var LIBDIR x86)"`
Of course, for pure 32bit x86 installations, all this is not necessary as the libdir is always `lib` there.

DarkDefender:
@Polynomial-C I will hopefully update the lib32 stuff later today :crossed_fingers: 

Can you check if my update is good and that this is the only thing left that needs to be fixed?

media-sound/mumble/mumble-9999.ebuild
Polynomial-C: `xdg.eclass` inhrits `xdg-utils.eclass` so I suppose you did it for some other reason. Anyway, please prefer `xdg.eclass`.

media-sound/mumble/mumble-9999.ebuild
```diff
@@ -64,97 +65,41 @@ BDEPEND="
...
+		"-DBUILD_TESTING=$(usex test)"
+		"-Ddbus=$(usex dbus)"
+		"-Dg15=$(usex g15)"
+		"-Djackaudio=$(usex jack)"
+		"-Dportaudio=$(usex portaudio)"
+		"-Dpulseaudio=$(usex pulseaudio)"
+		"-Drnnoise=$(usex rnnoise)"
+		"-Dspeechd=$(usex speech)"
+		"-Dtranslations=$(usex nls)"
+		"-Dzeroconf=$(usex zeroconf)"
```
Polynomial-C, line 87: Still not alphabetically sorted

media-sound/mumble/mumble-9999.ebuild
```diff
@@ -64,97 +65,41 @@ BDEPEND="
...
-
-pkg_preinst() {
-	xdg_pkg_preinst
+	cmake_src_configure
 }
 
 pkg_postinst() {
-	xdg_pkg_postinst
+	xdg_icon_cache_update
+	xdg_desktop_database_update
```
Polynomial-C, line 95: Please change that back when you use `xdg.eclass` again.

Polynomial-C (changes requested):
See my other change request comments

media-sound/mumble/mumble-9999.ebuild
DarkDefender: Ah, now I remember why.
The ebuild fails if I simply use `xdg` with:
```
FATAL: cmake_src_prepare has not been run
```

That is why I changed it.
So it seems like the xdg eclass messes up.

It also seems like there are quite a few other ebuilds that simply uses the utils class to not have to deal with this either.

Polynomial-C: Just add
```
src_prepare() {
        # required because of xdg.eclass also providing src_prepare
        cmake_src_prepare
}
```
to the ebuild.

DarkDefender:
There, now everything should have been taken care of :crossed_fingers:

media-sound/mumble/mumble-9999.ebuild
```diff
@@ -64,94 +65,54 @@ BDEPEND="
...
-
-pkg_preinst() {
-	xdg_pkg_preinst
+	cmake_src_install
+
+	if use amd64 ; then
+		# The 32bit overlay library gets automatically built and installed on x86_64 platforms.
+		# Install it into the correct 32bit lib dir.
+		libdir_64="/usr/$(get_libdir)/mumble"
+		libdir_32="/usr/$(get_abi_var LIBDIR x86)/mumble"
```
Polynomial-C, line 105: Make these two variables `local`

media-sound/mumble/mumble-9999.ebuild
```diff
@@ -64,94 +65,54 @@ BDEPEND="
...
+		mkdir -p $D/$libdir_32 || die
+		mv $D/$libdir_64/libmumbleoverlay.x86.so* $D/$libdir_32/ || die
```
Polynomial-C, line 107: Please use `${D}`, `${libdir_64}` and `${libdir_32}` (curly braces)

Polynomial-C (changes requested):
I know I start to nitpick now but this is good ebuild writing style...

media-sound/mumble/mumble-9999.ebuild
```diff
@@ -64,94 +65,54 @@ BDEPEND="
...
-pkg_preinst() {
-	xdg_pkg_preinst
+	cmake_src_install
+
+	if use amd64 ; then
+		# The 32bit overlay library gets automatically built and installed on x86_64 platforms.
+		# Install it into the correct 32bit lib dir.
+		local libdir_64="/usr/$(get_libdir)/mumble"
+		local libdir_32="/usr/$(get_abi_var LIBDIR x86)/mumble"
+		mkdir -p ${D}/{$libdir_32} || die
```
Polynomial-C, line 106: That line can be reduced by using dodir (sorry, didn't see it when writing my previous comment):
`dodir {$libdir_32}`
but no need to add this with another commit. I will change that when merging this PR.

kripton:
@Polynomial-C : Can you please also review related #17991 ? I've watched your change-requests here and adapted my PR when necessary. But due to not having 32bit libraries, it's a bit simpler ;) I can also squash all commits into one if preferred

Polynomial-C:
Thank you very much for your work and your patience with me.
FYI, I did some final adjustments in commit 3629693f926f2a47c4adc08e28987bb265a7aa4a

PR #17469, opened 2020-09-08
sys-apps/shadow: PDEPEND on acct-group/mail

dm0-:
This depends on the mail group at runtime, but it allows shadow to be installed first so that `groupadd` is executable while creating the mail group.

@floppym


dm0-:
I forgot to mention that this should not be merged before this eclass fix:
https://archives.gentoo.org/gentoo-dev/message/991e77ca67b136bd9b6356038ec5e9e5

Zlogene:
It has nothing to do with shadow.

PR #17471, opened 2020-09-08
media-gfx/inkscape: fix src uri

JellyBrick:
`https://inkscape.org/gallery/item/18460/${P}.tar.xz` -> `https://media.inkscape.org/dl/resources/file/${P}.tar.xz`


Zlogene:
This does not make any sense to change them, once the tarball is mirrorred it does not play a role.

trofi:
Tarball was not mirrored yesterday because the SRC_URI was broken. It had to be fixed as https://gitweb.gentoo.org/repo/gentoo.git/commit/media-gfx/inkscape?id=8b7674eaea637dd035749941ef8e1f5c6a8c09d6 to work at all.

If media-gfx/inkscape SRC_URI is not expected to work longer term it's already a problem. But if you insist on keeping it that way you should use mirror://gentoo instead.

PR #17478, opened 2020-09-09
dev-haskell/logict-state: build with ghc-8.8

solpeth:
I've crafted a patch (also pushed to `::haskell`) that allows `logict-state` to build with `ghc-8.8`.

A function needed a `MonadFail` type constraint in its type signature, and we needed to craft a `MonadFail Identity` instance.

Whether or not the instance satisfies Monad laws has not been extensively verified. I have also not verified that the `CPP` code correctly controls the conditional imports and type signatures for older versions of `ghc` - I just applied what the upstream recommendation is for `MonadFail` compatibility. My main concern was getting this package to build on `ghc-8.8` so we can close the linked bug, until we can clean out `::gentoo` with `logict-state`'s reverse dependencies.

Tested in an ~amd64 container.

Closes: https://bugs.gentoo.org/736098
Package-Manager: Portage-3.0.6, Repoman-3.0.1
Signed-off-by: Jack Todaro <solpeth@posteo.org>


trofi:
Pushed as is.

Thank you!

PR #17481, opened 2020-09-09
www-apps/ampache: bump to 4.2.2, un-last-rite

jsmolic:
Attempt of keeping this package in the repository. If I succeed, I would like to maintain it.


jsmolic:
I'll need few days to investigate other bugs and see if they are reproducible with latest version. In the meantime, any feedback is appreciated.

ajakk:
We probably can't cleanup the stable vulnerable version until stabling a fixed version.

jsmolic:
I think https://bugs.gentoo.org/655558 is still present, the package seems to miss some dependencies. I'm not sure how to work around this now as this is my first time venturing into anything PHP-related...\
Perhaps the package can just be bumped to resolve CVE and keep it in the tree without me taking mainteinership for now and I may try to deal with other issues at later time.\
@thesamesam Any thoughts on this?

PR #17485, opened 2020-09-09
media-gfx/yafaray: add support for python-3.{7,8}

waebbl:
- Remove support for python-3.6
- Remove blender USE flag and support. The blender plugin isn't ready
  for blender-2.8 series yet. Blender-2.79 is python-3.6 only, which
  is going to be removed soon. Support will be re-added, as soon as
  upstream has updated the plugin for blender-2.8*.

Closes: https://bugs.gentoo.org/737392


waebbl:
The reason for not removing the old ebuild, so people who want it, can still use the renderer with blender-2.79 and python-3.6. When these versions get cleaned from the tree, the -r0 version can get cleaned as well.

media-gfx/yafaray/yafaray-3.4.1-r1.ebuild
```diff
@@ -0,0 +1,114 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8} )
```
juippis, line 6: You can edit `PYTHON_COMPAT` without revbumping. I'd rather you do that if there's nothing else happening here.

I can already see this confusing portage a lot, since -r0 has py-3.6 and -r1 has py-3.7 & py-3.8.

waebbl: Ok, renamed it back.

juippis (approved):
After seeing the full diff between these files, maybe revbumping wouldn't been so bad...

but let's go with this since yafaray has no revdeps. This shouldn't break, eh, much.

PR #17492, opened 2020-09-10
net-im/zoom: Version bump to 5.2.458699.0906

dblaber:
Signed-off-by: Darren Blaber <dmbtech@gmail.com>


ulm:
Sorry, I see this only now, after the new version is already in the tree.

Anyway, thank you very much for the contribution!

dblaber:
> Sorry, I see this only now, after the new version is already in the tree.
> 
> Anyway, thank you very much for the contribution!

No problems! Will close, just figured if I was going to manually bump ebuild on my local, might as well push it :)

PR #17497, opened 2020-09-11
net-misc/taylor-uucp: Remove external function declarations in unix/i…

drinkcat:
…nit.c

That makes the code compile witho fewer warnings, and gets rid of this error
with recent Chrome OS toolchains:

```
init.c:100:14: error: redeclaration of 'getcwd' must have the 'overloadable' attribute
extern char *getcwd ();

             ^
/build/amd64-generic/usr/include/bits/unistd.h:158:8: note: previous overload of function is here
__NTH (getcwd (char *const __clang_pass_object_size __buf, size_t __size))
       ^
```

I didn't try to send upstream, I'm not sure if this package is still maintained (https://airs.com/ian/uucp.html still seems to be the canonical source.)

This fork has a very similar patch https://github.com/quinot/taylor-uucp (https://github.com/quinot/taylor-uucp/commit/b599665ae1e085df0f96de5f50a738fceee02150), but it's huge (600K), and this part of the patch is the only thing that seems to be _absolutely_ required.


thesamesam:
Yeah, the fork doesn't look buzzing with activity either...

thesamesam:
Thanks!

drinkcat:
Thanks to you!

PR #17501, opened 2020-09-11
net-news/liferea: version bump to 1.12.9

nE0sIghT:
https://github.com/lwindolf/liferea/releases/tag/v1.12.9


net-news/liferea/liferea-1.12.9.ebuild
```diff
@@ -0,0 +1,76 @@
...
+
+src_install() {
+	default
+	find "${D}" -name '*.la' -delete || die
+}
+
+pkg_preinst() {
+	xdg_pkg_preinst
+	gnome2_schemas_savelist
+}
```
juippis: This shouldn't be needed in your case, I'll remove it for you.

net-news/liferea/liferea-1.12.9.ebuild
```diff
@@ -0,0 +1,76 @@
...
+pkg_postinst() {
+	xdg_pkg_postinst
+	gnome2_schemas_update
+
+	elog "Additional features can be enabled via"
+	elog "\tapp-crypt/libsecret[introspection] for Libsecret Support plugin"
+	elog "\tdev-python/pycairo and x11-libs/gdk-pixbuf[introspection] for Tray Icon (GNOME Classic) plugin"
+	elog "\tmedia-libs/gstreamer[introspection] for Media Player plugin"
+	elog "\tnet-misc/networkmanager for monitoring network status"
+	elog "\tx11-libs/libnotify[introspection] for Popup Notifications plugin"
```
juippis: I've probably suggested this before, but using *optfeature.eclass* is better in here, since it only prints missing packages.

juippis (approved):

PR #17510, opened 2020-09-12
sys-fs/bcache-tools: version bump to 1.1

amoskvin:
Update to version [released](https://git.kernel.org/pub/scm/linux/kernel/git/colyli/bcache-tools.git/log/?h=bcache-tools-1.1) by Coly Li, bcache's [kernel maintainer](https://www.kernel.org/doc/html/latest/process/maintainers.html#bcache-block-layer-cache). This verison is used [by Arch](https://aur.archlinux.org/packages/bcache-tools/).

I verified that setting up a bcache works.

Closes: https://bugs.gentoo.org/722002
Closes: https://bugs.gentoo.org/656884


amoskvin:
> DuplicateFiles | duplicate identical files in FILESDIR: 'files/1.0.8_p20140220/bcache-tools-1.0.8-noprobe.patch', 'files/1.1/bcache-tools-1.0.8-noprobe.patch'
> -- | --

Not really sure what the best practice is correctly reuse the patch with that file structure. Point at the old one?

> 1.0.5 | MatchingChksums | distfile 'bcache-tools-1.0.8_p20140220.tgz' matches checksums for 'bcache-tools-1.0.8.tgz'
> -- | -- | --

Wasn't me!

amoskvin:
Pointed to the old one and now it passes

PR #17516, opened 2020-09-12
games-rpg/freedroidrpg: fix build on gcc 10

DarthGandalf:
And fix prefix and add py3.8

Closes: https://bugs.gentoo.org/706930


chewi:
Not tested this one myself so I'm trusting you here!

PR #17520, opened 2020-09-13
net-misc/gnome-remote-desktop: version bump to 0.1.9

nE0sIghT:
Signed-off-by: Yuri Konotopov <ykonotopov@gnome.org>


nE0sIghT:
Also about package mask - don't it pipewire issue? I mean there is user pipewire.socket, but it's not enabled.

nE0sIghT:
I beleave CI result is false-positive because package is masked as well as screencast USE flag.

mattst88:
Thanks!

PR #17521, opened 2020-09-13
sys-auth/seatd: New Package (plus wlroots and sway commits to add support)

lanodan:


sys-auth/seatd/seatd-0.3.0.ebuild
```diff
@@ -0,0 +1,34 @@
...
+	EGIT_REPO_URI="https://git.sr.ht/~kennylevinsen/seatd"
+else
+	KEYWORDS="~amd64"
+	SRC_URI="https://git.sr.ht/~kennylevinsen/seatd/archive/${PV}.tar.gz -> ${P}.tar.gz"
+fi
+LICENSE="MIT"
+SLOT="0/1"
+IUSE="examples logind"
+
+DEPEND="logind? ( || ( sys-auth/elogind sys-apps/systemd ) )"
```
kchibisov, line 22: I think that it'll be nice to use `systemd`/`elogind` use flags instead of custom `logind` flag? You may also want to add `man` use flag to make `scdoc` an optional dependency the way it's done in sway ebuild?

gui-libs/wlroots/wlroots-9999.ebuild
```diff
@@ -18,7 +18,7 @@ fi
 
 LICENSE="MIT"
 SLOT="0/9999"
-IUSE="elogind icccm systemd x11-backend X"
+IUSE="elogind icccm seatd systemd x11-backend X"
 REQUIRED_USE="?? ( elogind systemd )"
```
thesamesam, line 22: Does seatd change this requirement...?

gui-wm/sway/sway-9999.ebuild
```diff
@@ -20,7 +20,7 @@ fi
 
 LICENSE="MIT"
 SLOT="0"
-IUSE="elogind fish-completion +man +swaybar +swaybg +swayidle +swaylock +swaymsg +swaynag systemd tray wallpapers X zsh-completion"
+IUSE="elogind fish-completion +man +swaybar +swaybg +swayidle +swaylock +swaymsg +swaynag seatd systemd tray wallpapers X zsh-completion"
 REQUIRED_USE="?? ( elogind systemd )
```
thesamesam, line 24: Should this be updated?

sys-auth/seatd/metadata.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+<maintainer type="person">
+	<email>contact@hacktivis.me</email>
```
thesamesam, line 5: Name needed too

gui-libs/wlroots/wlroots-9999.ebuild
lanodan: No because seatd doesn't conflicts with elogind/systemd, it's an additional method.

gui-wm/sway/sway-9999.ebuild
lanodan: Same as for the wlroots part.

gui-libs/wlroots/wlroots-9999.ebuild
thesamesam: You've lost me though. Surely seatd should satisfy the REQUIRED_USE? So that you can use wlroots without elogind or systemd?

lanodan: Yeah, that's the point but seatd isn't a logind implementation. It either provides it itself via it's own API or proxies it to a logind implementation.

sys-auth/seatd/seatd-0.3.0.ebuild
lanodan: Will do for the `systemd`/`elogind` part, for some reason I thought they weren't globals and it turns out seatd needs to be recompiled if the `systemd`/`elogind` target changes.

As for the `man` for `scdoc` I don't really agree there as `scdoc` is really small (~70K on my machines) and doesn't depends on anything more than POSIX/C99, I think this should only be there for gigantic festivals like pandoc, docbook, …

gui-libs/wlroots/wlroots-9999.ebuild
thesamesam: Right, I've got you - thanks for explaining

sys-auth/seatd/files/seatd.initd
```diff
@@ -0,0 +1,4 @@
+#!/sbin/openrc-run
```
kchibisov, line 1: It seems like you've added `initd` file, but you're not installing it in ebuild?

lanodan: Oh, nice catch, looks like I screwed up somewhere as I'm sure I wrote it at some point.

lanodan:
Version bumped it to 0.4.0

PR #17530, opened 2020-09-13
media-radio/direwolf: Bump to v1.5

kripton:
This PR updates media-radio/direwolf to the latest upstream release: v1.5. Latest version in tree right now is v1.4. It uses the same approach to patch the Makefile shipped with the project. In the "dev" branch of the upstream project, cmake is now used so this will hopefully be the last version to use the Makefile-patching approach.
Furthermore, the following things have been achieved in the ebuild:
- Dependency on gpsd is no longer automagic but controlled via a USE flag (https://bugs.gentoo.org/705100)
- Upstream fix for compilation against latest gpsd is included. It should still work against the older (= stable) versions of gpsd as well
- Compilation no longer has the `cc` command hardcoded but it can now be set using the `CC` ENV VAR (https://bugs.gentoo.org/724720)
- LDFLAGs set via ENV VAR are now taken into account (https://bugs.gentoo.org/727636)
- Bump the ebuild to EAPI 7
- Optionally install the `CM108` udev rule and feature (new feature since upstream v1.5) (controlled via USE flag `udev`)
- Optionally depend on hamlib (controlled via USE flag `hamlib`)
- Add local USE-flag description and upstream information to metadata


kripton:
ping @rich0

rich0:
I went ahead and committed this.  Only concern was that we're now carrying quite a few patches.  If you're willing I'd try to submit them upstream (the commenting out in the makefile would be better handled with --enable-foo).  Otherwise we're going to be doing a lot of patch generation on future bumps.

But it looks good overall!  Thanks for contributing!

PR #17532, opened 2020-09-13
games-arcade/pacmanarena: fix build on gcc 10

DarthGandalf:
Fix removing of -g from CFLAGS

Closes: https://bugs.gentoo.org/708066


ConiKost (changes requested):
Please split up in two commits, one for adding patch aud bumping to EAPI=7 and the second for dropping old. While you are at it, you could also drop empty ```USE=""``` and sort stuff in ```RDEPEND```. And ```app-arch/unzip``` must go into ```BDEPEND``` with EAPI=7.

DarthGandalf:
Done

PR #17533, opened 2020-09-14
net-mail/notmuch: Unconditionally install man-pages

rseichter:
Installation of man-pages no longer depends on the "doc" use flag.


rseichter:
@thesamesam This merge request supersedes [MR 15728](/gentoo/gentoo/pull/15728).

PR #17541, opened 2020-09-14
media-gfx/asymptote, sci-libs/{amd,btf,camd,ccolamd,cholmod,libexcelformat,qd}: update HOMEPAGE, use https

mm1ke:
Hi,

This PR updates the homepage of some sci packages and uses https instead of http.

### DIFF
--- a/media-gfx/asymptote/asymptote-2.61.ebuild
+++ b/media-gfx/asymptote/asymptote-2.61.ebuild
@@ -8,7 +8,7 @@ PYTHON_COMPAT=( python3_{6,7} )
 inherit autotools elisp-common latex-package python-r1
 
 DESCRIPTION="A vector graphics language that provides a framework for technical drawing"
-HOMEPAGE="http://asymptote.sourceforge.net/"
+HOMEPAGE="https://asymptote.sourceforge.net/"
 SRC_URI="mirror://sourceforge/asymptote/${P}.src.tgz"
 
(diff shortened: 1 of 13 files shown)


media-gfx/asymptote/asymptote-2.61.ebuild
```diff
@@ -8,7 +8,7 @@ PYTHON_COMPAT=( python3_{6,7} )
 inherit autotools elisp-common latex-package python-r1
 
 DESCRIPTION="A vector graphics language that provides a framework for technical drawing"
-HOMEPAGE="http://asymptote.sourceforge.net/"
+HOMEPAGE="https://asymptote.sourceforge.net/"
```
ConiKost, line 11: Does it work for you? I currently can only reach ```http```, but ```https``` give timeout.

ConiKost: Since I can't reach it also from other PCs, I will drop this commit and merge the rest.

ConiKost (approved):

media-gfx/asymptote/asymptote-2.61.ebuild
mm1ke: Sorry for the late reply, you're right, the change was wrong. I forgot to change the ending to `.io` too.. i'll create a new PR.

PR #17542, opened 2020-09-14
dev-embedded/esptool: add python3.[89] + new maintainer

madmartin:
successfully tested with python3.8 and 3.9
remove maintainer-needed, add proxy maintainer

add python3.9 to depend package dev-python/pyelftools

Package-Manager: Portage-3.0.6, Repoman-3.0.1

### DIFF
--- a/dev-embedded/esptool/esptool-2.8-r1.ebuild
+++ b/dev-embedded/esptool/esptool-2.8-r1.ebuild
@@ -1,0 +1,41 @@
+# Copyright 2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{6,7,8,9} ) # apps work with 3_9 but test depend is not fulfilled
+DISTUTILS_SINGLE_IMPL=1
+DISTUTILS_USE_SETUPTOOLS=rdepend
+
+inherit distutils-r1
+
+DESCRIPTION="Utility to communicate with the ROM bootloader in Espressif ESP8266 and ESP32"
+HOMEPAGE="https://github.com/espressif/esptool"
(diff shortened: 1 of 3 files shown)


madmartin:
@mgorny Hi - AFAIK you're the python eclass expert - can you review please?

And a question: I checked the application with python 3.9 - it works, but the dependent package `dev-python/pyserial` for test is not available for python 3.9 (yet). "pkgcheck" figures this out and therefore rejects the ebuild with `PYTHON_COMPAT=( python3_{6,7,8,9} )`
Is there a way to write this into the ebuild except as a comment?

Greetings
Martin

mgorny:
I don't see anything wrong with it. Could you try pyserial and add py39 there if it works?

madmartin:
>  Could you try pyserial and add py39 there if it works?

I already did that, pathon3.9 works for the app itself without problems, but the depency for "test" dev-python/pyelftools has only  `PYTHON_TARGETS="python3_6 python3_7 python3_8"}`

Do you mean I should try adding python3_9 to dev-python/pyelftools  and see if it works?

madmartin:
> Do you mean I should try adding python3_9 to dev-python/pyelftools and see if it works?
@mgorny  FYI:
I added python3_9 to `dev-python/pyelftools`, merged it, and then added python3_9 to `dev-embedded/esptool` and did a 

USE="test -python_single_target_python3_7 -python_single_target_python3_8 python_single_target_python3_9" ebuild esptool-2.8-r1.ebuild test

which runs without error!

Is that enough to proof that `dev-python/pyelftools` and  `dev-embedded/esptool` both work with python3.9 ?

mgorny:
Yes, that SGTM.

mgorny:
I've added it to my merge queue, will go when pypy3 finishes rebuilding.

PR #17543, opened 2020-09-14
dev-libs/girara: Version bump (v0.3.5)

xdch47:


dev-libs/girara/girara-0.3.5.ebuild
```diff
@@ -0,0 +1,48 @@
...
+
+RESTRICT="!test? ( test )"
+
+RDEPEND="dev-libs/glib:2
+	dev-libs/json-c:=
+	>=x11-libs/gtk+-3.20:3
+	libnotify? ( x11-libs/libnotify )"
+
+DEPEND="${RDEPEND}
+	doc? ( app-doc/doxygen )
```
thesamesam, line 32: I imagine doxygen can go in BDEPEND and we don't run tests in cross so we could put check in there too.

thesamesam:
Thank you!

PR #17545, opened 2020-09-15
games-mud/tf: fix build, fix version number

DarthGandalf:


contyk:
What's the reason for removing the features we have and likely many people enjoy?
At least in my case, I'll have no use for the package without GMCP support.

DarthGandalf:
@contyk GMCP etc support was added by tf-allrootpatch.txt and tf-allsrcpatch.txt which can't be downloaded anymore from SRC_URI. When I switched to debian patches (and dropped tf-*txt which I assumed do the same), `./configure` started complaining about unknown flag --enable-gmcp, which is why I removed it.

Currently tf is installable only because these 2 patches can be downloaded from gentoo mirror (which I forgot about).

DarthGandalf:
@contyk thanks for noticing this, I've added it back. These patches should still be served from somewhere else though, not from "Http/1.1 Service Unavailable"

contyk:
Thanks.

I have not tried this and I'm not familiar with those Debian patches. However, I maintain tf in Fedora and I recall I had to add `-fcommon` to CFLAGS to build it with gcc-10; do those include this fix?

/me should just test it...

DarthGandalf:
Yes, https://sources.debian.org/patches/tf5/5.0beta8-8/0008-Fix-duplicate-world_decl-definitions.patch/

I prefer patch which fixes the issue over hiding it with `-fcommon`…

contyk:
Yeah, I agree a patch is better.

Just running this now. Seems to build and work fine, even though it appears as a downgrade to the previous v50; it also pulled in quilt, as the only package on my system.

DarthGandalf:
Since it's a BDEPEND, `emerge -c` would get rid of it.

DarthGandalf:
The only diff in latest commit is that I just realized that there is mirror://debian

contyk:
> Since it's a BDEPEND, `emerge -c` would get rid of it.

I'd still prefer just carrying the patches as normal patches and dropping the bdep; it's just a handful and makes it easier for everyrone to review them. And makes the `prepare` phase look nicer.

contyk:
By the way, I made another patch to apply on top of all these. Will be adding it to Fedora/EPEL soon. Attached, if you're [interested.

https://pastebin.com/AtUUEwF0

DarthGandalf:
> I'd still prefer just carrying the patches as normal patches and dropping the bdep; it's just a handful and makes it easier for everyrone to review them. And makes the prepare phase look nicer.

But if you convince Debian to include your patch, adding it here would be as simple as version bump - by a simple rename from _p8 to _p9 :)

DarthGandalf:
OTOH, I see Fedora includes these missing files like tf-allrootpatch.txt 

I wonder how hard would it be to declare it the upstream for this package instead of Debian…

DarthGandalf:
Even better option is https://github.com/ingwarsw/tinyfugue which is alive and kicking

contyk:
> OTOH, I see Fedora includes these missing files like tf-allrootpatch.txt
> 
> I wonder how hard would it be to declare it the upstream for this package instead of Debian…

Well indeed, I'm the maintainer of that package :)

> Even better option is https://github.com/ingwarsw/tinyfugue which is alive and kicking

Those aren't the adjectives I'd choose. The quality of the patches it includes is questionable. It doesn't even build and the added `configure` toggles only work one way. If at all. The suggested `--disable-python` still tries to link against python. And fails. Building against lua doesn't work either but at least it can be turned off.

At first I thought it was great someone was reviving the project and then I actually tried it. I'd be interested in the wide character patches, if they actually work. Impossible to test in the current state, though.

DarthGandalf:
But you can send the fixes there. As it is active, it'll probably get accepted.

My point is, I think it's better to have *an upstream* than to maintain the number of patches yourself. Which is why I'm not a fan of listing all the patches manually, and instead just applying whatever Debian applied, where quilt is an easy way to achieve that.

warmstarter:
Hey, we're doing very active work on TF5.

I've been doing extensive research and tracking down all the various computer patches for TF5 and integrating the good ones. In addition to that, a lot of work being done to both modernize and fixing various crash bugs.

Also happy to work with you to make this easier to integrate into your distribution.

I've also been working on getting patches for the severe bugs into any other TF5 that at least shows some people taking an interest.

https://github.com/warmstarter/tinyfugue-1

warmstarter:
> By the way, I made another patch to apply on top of all these. Will be adding it to Fedora/EPEL soon. Attached, if you're [interested.
> 
> https://pastebin.com/AtUUEwF0

Integrated this as a configurable option. Had already been researching it, but it made sense to me, and if redhat signed-off on it, I got no reason not to.

warmstarter:
Also, if you'd just prefer to have the allroot and allsrc patches, this is the original distributor of them. The reason for the 'all' is that each patch integrates there separate patches (GMCP, ACTP, and OPTION102). Those patches are also provided on their own so it's possible to just take in the patch for GMCP or such.

To note: The patch for OPTION102 has a bug in it that as far as I've seen has gone unnoticed by everyone using it.

PR #17546, opened 2020-09-15
sys-cluster/spark-bin: Add 3.0.1

trozamon:


sys-cluster/spark-bin/spark-bin-3.0.1.ebuild
```diff
@@ -0,0 +1,61 @@
...
+
+	insinto usr/lib/spark/bin
+	doins bin/load-spark-env.sh
+
+	insinto usr/lib/spark
+	doins -r conf
+	doins -r jars
+	doins -r python
+	doins -r sbin
+	doins -r yarn
```
juippis, line 49: You should be able to combine these into a single `doins -r`.

juippis (approved):

PR #17548, opened 2020-09-15
dev-python/bibtexparser: add python 3.{8.9} pypy3 support

gerion0:
Tested with `ebuild test`.

I'm not sure, if this needs a revbump, but since it only adds optional features it would trigger a useless rebuild, I guess.

### DIFF
--- a/dev-python/bibtexparser/bibtexparser-1.1.0.ebuild
+++ b/dev-python/bibtexparser/bibtexparser-1.1.0.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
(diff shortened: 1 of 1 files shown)


thesamesam:
Sorry I didn't see this before now. It looks like the changes have been done already: https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=79a660c825fcab60249e532b38ea57e91f4fc474

gerion0:
This is not entirely true, since pypy3 support is missing in your linked commit. I have tested it an it works with pypy3. Can you add it as well?

PR #16554, opened 2020-07-03
gui-libs/display-manager-init: refactor init files out of xorg-server

epsilon-0:
Closes: https://bugs.gentoo.org/730644

---

Changes to script: add support for `greetd`

@gyakovlev has provided the patch

Package-Manager: Portage-2.3.101, Repoman-2.3.22
Signed-off-by: Aisha Tammy <gentoo@aisha.cc>

---

This needs extensive testing!

PR #17557, opened 2020-09-15
gui-libs/greetd: add openrc initscript

JonasToth:
This commit adds an OpenRC-initscript for gui-libs/greetd, that
currently only supports systemd systems.

Bug: https://bugs.gentoo.org/742698
Signed-off-by: Jonas Toth <gentoo@jonas-toth.eu>

### DIFF
--- a/gui-libs/greetd/greetd-0.6.1-r1.ebuild
+++ b/gui-libs/greetd/greetd-0.6.1-r1.ebuild
@@ -104,6 +104,7 @@ src_install() {
 	doins config.toml
 
 	systemd_dounit greetd.service
+	newinitd "${FILESDIR}"/${PN}.init ${PN}
 
 	if use man; then
 		doman agreety.1 greetd.1 greetd.5 greetd-ipc.7
--- a/gui-libs/greetd/files/greetd.init
+++ b/gui-libs/greetd/files/greetd.init
@@ -1,0 +1,8 @@
+#!/sbin/openrc-run
+
+command=/usr/bin/greetd
+pidfile=/var/run/greetd.pid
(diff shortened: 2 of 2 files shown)


gui-libs/greetd/greetd-0.6.1-r1.ebuild
```diff
@@ -103,7 +103,11 @@ src_install() {
 	insinto /etc/greetd
 	doins config.toml
 
-	systemd_dounit greetd.service
+	if use systemd; then
+		systemd_dounit greetd.service
+	else
+		newinitd "${FILESDIR}"/${PN}.init ${PN}
+	fi
```
mm1ke, line 110: Hi,

You can simplify this by removing the `systemd` use flag and simply install both files.
Actually this is even prohibited, see https://wiki.gentoo.org/wiki/Project:Systemd/Ebuild_policy#General_guidelines

JonasToth: Updated. Thanks for the pointer.

epsilon-0:
No, this is incorrect, you are not supposed to add `greetd` as a separate service.
`greetd` is to be added as a display manager and is currently being worked on in this PR - https://github.com/gentoo/gentoo/pull/16554

That PR handles `greetd` integrated as a replacement to all other display managers, `sddm`, `gdm`, etc.

epsilon-0:
If you are keen on trying it out, that would also be appreciated.
Testers are need for https://github.com/gentoo/gentoo/pull/16554

JonasToth:
Closing in favor of #16554. Will support there.

PR #17561, opened 2020-09-16
games-emulation/gngeo: fix build with gcc 10

DarthGandalf:


ConiKost (changes requested):
Please sort ```DEPEND```and drop empty ```USE=""```.

DarthGandalf:
Done

PR #17562, opened 2020-09-16
app-admin/terraform: improve 0.13.2 ebuild

ap4y:
This commit improves ebuild for terraform as follows:

- Switch to builds with -mod=vendor and clean unnecessary EGO_SUM
- Remove PN from SRC_URI per dev manual
- Update licenses, list is generated from binary using golicense
- Correct RESTRICT to include test rather than overwrite
- Align build command with the upstream build script


Zlogene:
not using vendor is ententional, the rest is reeally does not improve anything.

PR #17572, opened 2020-09-17
dev-util/packer: 1.6.2 bump

ap4y:
This patch updates dev-util/packer to 1.6.2 and updates ebuild as following:

- Removed PN from SRC_URI per dev guide
- Updated LICENSE, output from golicense on compiled binary
- Updated RESTRICT to include "test" since we don't have src_test
- Added -ldflags as per upstream build script
- Removed -work to not leave files post build


thesamesam:
is there a reason you've closed this? It's not being ignored, the backlog is just huge and we're working through it. It's about 2-3 weeks atm.

xp-vit:
@ap4y have you closed it to reopen for 1.6.3?

ap4y:
@thesamesam I realize that it's going to take some time and it wasn't the reason for closing my PR. I looked back at this patch and I felt like it's a bit excessive for the purpose of a simple version bump and considering `1.6.3` is out I felt like it's better to let someone else to handle the bump. I'm happy to re-open it if you feel otherwise. Thanks!

PR #17574, opened 2020-09-17
net-firewall/arno-iptables-firewall: bump version to 2.1.1

ferki:
This PR bumps net-firewall/arno-iptables-firewall to its latest 2.1.1 version, along with the following changes:

- upstream removed some configuration options in favor of better autodetection, so we have to search and replace a few default locations in different files than before, but the overall settings/logic didn't change
- install built-in job manager used by some plugins
- install logrotate file
- additional built-in plugin helpers

Overall idea is to get the new version available in the main tree, wait for stabilization, then clean up older versions.

Please review and merge, or let me know how to improve it further!


net-firewall/arno-iptables-firewall/arno-iptables-firewall-2.1.1.ebuild
```diff
@@ -0,0 +1,105 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+inherit readme.gentoo-r1 systemd
+
+DESCRIPTION="Arno's iptables firewall script"
+HOMEPAGE="https://rocky.eld.leidenuniv.nl"
```
juippis, line 8: Could you list the Github here as well for 2ndary HOMEPAGE?

juippis (approved):

net-firewall/arno-iptables-firewall/arno-iptables-firewall-2.1.1.ebuild
ferki: Thanks for the suggestion, @juippis, I agree it might make sense to point to the github repo as well.

I went ahead and included that in my [overlay](https://github.com/ferki/gentoo-overlay/commit/b55360d45e6422f8d59d83cee659b7b0e72487e1), so the next PR here will have it too. If you want me to send a followup PR before the next version bump with that as the sole change, let me know!

PR #17575, opened 2020-09-17
media-sound/spotify: bump to 1.1.42, remove patchelf version from dep, remove 1.1.26-r1

realnc:
patchelf < 0.10 has been removed from portage for a while now, so no need to depend against >= 0.10.


prometheanfire:
you mind rebasing this without the removal of 1.1.26?

realnc:
> you mind rebasing this without the removal of 1.1.26?

1.1.26-r2 is still there. Only r1 was removed.

realnc:
Looks like there is now a different version bump in portage. Closing.

prometheanfire:
ah, in that case, fixing the conflict would be my request, then I'll merge.

realnc:
Pushed.

PR #17303, opened 2020-08-28
dev-libs/aws-checksums: bump to 0.1.9

jsmolic:

PR #17576, opened 2020-09-17
dev-libs/aws-sdk-cpp: Revision bump for fixing dependencies and test suite failing

Yamakuzure:
This PR consists of these three changes, that are all connected.

1. dev-libs/aws-c-common: Bump to 0.4.57 and add static-libs USE flag
2. dev-libs/aws-c-event-stream: Version Bump to 0.1.6 plus two fixes
  * added static-libs USE flag.
  * depend on >=dev-libs/aws-c-common-0.4.26
    From this version on, AwsFindPackage.cmake is available, which this package build system uses.
3. dev-libs/aws-sdk-cpp: Fix dependencies and a typo
  * Depend on
    * &gt;=dev-libs/aws-c-common-0.4.26 and
    * &gt;=dev-libs/aws-c-event-stream-0.1.6
    The latter is needed for the test suite to succeed.
  * A dash was missing, and thus the elastic-inference module could not be activated. Instead, setting USE="elasticinference" lead to the configure stage to fail.

dev-libs/aws-c-common:
Bug: https://bugs.gentoo.org/695418
Bug: https://bugs.gentoo.org/717572
Closes: https://bugs.gentoo.org/695418
Closes: https://bugs.gentoo.org/717572

dev-libs/aws-c-event-stream:
Bug: https://bugs.gentoo.org/695418
Bug: https://bugs.gentoo.org/717574
Closes: https://bugs.gentoo.org/717574

dev-libs/aws-sdk-cpp:
Bug: https://bugs.gentoo.org/742650
Closes: https://bugs.gentoo.org/742650


Yamakuzure:
I do not know why the system put the "maintainer-needed" label in place, as I am in all three metadata.xml files.

jsmolic:
> I do not know why the system put the "maintainer-needed" label in place, as I am in all three metadata.xml files.

I think you were removed from metadata in the meantime.\
https://github.com/gentoo/gentoo/pull/17303 should be merged before this, so you will be restored as the maintainer

jsmolic:
> > I do not know why the system put the "maintainer-needed" label in place, as I am in all three metadata.xml files.
> 
> I think you were removed from metadata in the meantime.
> #17303 should be merged before this, so you will be restored as the maintainer

... or I can close https://github.com/gentoo/gentoo/pull/17303 and let you take over these bumps.\
As you wish

thesamesam:
It all seems to have become a bit complicated 🧩 . It looked like some dependencies need to be added/tightened anyhow based on some of the bugs.

Yamakuzure:
> It all seems to have become a bit complicated  . It looked like some dependencies need to be added/tightened anyhow based on some of the bugs.

Exactly. I had higher local versions as I was preparing bumping the dependencies, before I found #17303. Unfortunately that hindered me to see that both aws-c-common and aws-c-event-stream not only should, but actually have to be bumped before or at the same time as the SDK.

Yamakuzure:
@juippis : Test suite succeeds now with these updates.
As bug 742650 originates from the dependencies not being bumped before the SDK, I took the liberty to whip up an all-in-one PR.
Yes, I know, this is normally frowned upon. But it all is tied together, and the bumps are trivial. So I hope I can get away with this. :frowning:

thesamesam:
It is not frowned upon to do things in the same PR. It is frowned upon to do things in the same _commit_ because it makes review really hard and possible reverts complicated.

dev-libs/aws-c-common/aws-c-common-0.4.57.ebuild
```diff
@@ -12,12 +12,13 @@ SRC_URI="https://github.com/awslabs/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
...
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE="test"
+IUSE="static-libs test"
 
 RESTRICT="!test? ( test )"
 
 src_configure() {
 	local mycmakeargs=(
+		-DBUILD_SHARED_LIBS=$(usex !static-libs)
```
thesamesam, line 21: But why? Is there a real [use](https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0302) for this?

Yamakuzure:
> It is not frowned upon to do things in the same PR. It is frowned upon to do things in the same _commit_ because it makes review really hard and possible reverts complicated.

Ah! good to know!

However, your comment on the static-libs USE flag usage brought something to my mind: The dependencies have to reflect that! Meaning, both aws-c-common and aws-c-event-stream must be built with the same state of the static-libs USE flag as aws-sdk-cpp.

...hang on...

dev-libs/aws-c-common/aws-c-common-0.4.57.ebuild
thesamesam: That doesn't mean we should keep static-libs everywhere, though.

thesamesam:
(Yes, this is where static-libs is a pain as a dependency. I suspect static-libs= will be useful.)

Yamakuzure:
> (Yes, this is where static-libs is a pain as a dependency. I suspect static-libs= will be useful. But I'd honestly prefer it if we just dumped it, if we can, because of the [policy](https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0302).)

static-libs= added.

Well, me too. But I can see a point in having a statically built AWS SDK.

thesamesam:
What is the use? I'm happy with it if there is one.

Yamakuzure:
> What is the use? I'm happy with it if there is one.

1. Write a small app that connects to an S3 storage, or an EDB, and give it to a customer, without having to distribute the libraries (maybe you can't) and without having to ask your customer to install the SDK.
2. For Lambdas this is nice, gives you control
3. Maybe you want to create a Docker container for ECR, that is as small as possible
4. Same with Kubernetes in EKS
5. Some people may think that statically linking the machine learning parts may improve performance. (Well, could, but only very very slightly)
6. With some media tools like "Elemental Media", linking statically might also be a good idea.
7. From AWS developer guide: _"To dynamically link to the SDK, you must define the USE_IMPORT_EXPORT symbol for all build targets using the SDK."_ ; Maybe there is a situation, where this is not possible?

Frankly, I don't need it, at least right now. But there are some use cases. Maybe many many more. Maybe they are all irrelevant.

dev-libs/aws-c-common/Manifest
```diff
@@ -1 +1 @@
-DIST aws-c-common-0.4.11.tar.gz 306933 BLAKE2B 5a2370a3696e4d90b67bc03fbf88f152e1a2d6e4d4a9ff2a9423a2cf18c14a1e3b6d978250c3359f4ebb2cf5e524e416b11604dc09dc45f5c8a5a6f982b344a7 SHA512 1e00f6aa025578978551fc8ca1bc09b448f79ac28bdc969af4312569ffca035d656dc68825da15ad8f4f890bbdb03327c624147dded1ce402e9dab1726290890
+DIST aws-c-common-0.4.57.tar.gz 418954 BLAKE2B cfb268d92e90343e01756b571d0d7fa1269a144b933b5dfae6e5fb92a2b384ff73bf14ad830b53427ad37e9dbd9a0afff5b571108869a69a8d466982a2e6c6ef SHA512 77a9110a067e3057ae3cb90ee0b1fa42e18595e476b803f83711ecc164053db73683df2fccf75ad992b7d08e142cb31d6a70f87ce2f3b87ce28ee1e8151b3c62
```
juippis, line 1: FYI This should be 2 commits, one for adding a new version, 2nd for removing old.

juippis (approved):

PR #17580, opened 2020-09-18
net-analyzer/zmap: tidy up ebuild

jsmolic:
- Port to EAPI 7
- Port to cmake eclass
- Drop unused json USE flag
- Sync live ebuild


juippis (approved):

juippis:
Your last commit is missing sign-off but let's not let it bother us. How did that happen with repoman though? (No need to answer but you may want to figure it out)

PR #17593, opened 2020-09-18
dev-util/nvidia-cuda-toolkit: Bump to 11.0.3

jvesely:
Add new 'sanitizer' use-flag to hide compute-sanitizer.
Split profiler use-flag into profiler and vis-profiler.
The former enables a cmdline tool and library (needed by e.g. pycuda),
the latter enables a gui tool written in Java (nvvm).
Drop eclipse use-flag, the eclipse plugin is part of nvvm which itself
needs java.
Add 'nsight' to enable nVidia nsight profiling/development environment.
Add note about restrictions to access perf counters.

Signed-off-by: Jan Vesely <jano.vesely@gmail.com>


jvesely:
This rewrites much of the ebuild. The directory structure in the downloaded archive no longer matches the installed one.
The new approach goes over every subpackage (based on use-flags) and installs it separately.
There's still a bunch of unbundling to do, but this should be a working and testable version.

I only have an older maxwell GPU so I can't really test Nsight. I've tested that:
 * Building and running a simple program works. Note that the new default target is sm_52 (Pascal).
 * nvprof works with he program built ^^
 * pycuda works

SoapGentoo:
@jvesely thanks for looking into this (because I need CUDA 11 too)

amadio:
The nsight profilers (nsight compute, systems, at least) are available as separate downloads, so they could probably be handled more easily as separate packages. See https://developer.nvidia.com/gameworksdownload#?dn=nsight-compute-2020-1-2 for example.

amadio:
Also, not sure how it's now, but nvvm and other tools don't work out of the box in previous ebuilds. Some of them are not marked executable, and some paths are not added in ld.so.conf so nvvm doesn't work because of that as well. That said, nvvm is considered legacy. The nsight profilers are the new tools.

jvesely:
> Also, not sure how it's now, but nvvm and other tools don't work out of the box in previous ebuilds. Some of them are not marked executable, and some paths are not added in ld.so.conf so nvvm doesn't work because of that as well. That said, nvvm is considered legacy. The nsight profilers are the new tools.

most of that should be fixed.
I've checked that `nvvm`, `ncu-ui` and co at least showed an initial window.
I have only collected data using `nvprof` which works ok.

the installation patter appears to be 'go over `builds/cuda_*` and merge whatever you find to cuda_root' with some special handling of executables in `integrations`', so the ebuild can definitely be improved/consolidated.

amadio:
I will test them all and let you know if I find any problems, or just submit changes for review in case it's not too hard to fix. For now I have nsight tools installed by hand, as well as CUDA 11. I'm stuck with older CUDA because of cudnn for Gentoo packages (for testing sci-physics/root with USE=cudnn).

rossbridger:
Just in time that CUDA 11.1 is released and now we need another version bump all over again...(and finally we have gcc 10 support)

amadio:
One thing I noticed is that prefix is broken in the new package, due to double prefix in installing many files...

amadio:
Prefix support fixed, but now I see this:
```
$ ncu --version
ERROR : Unsupported Architecture: AMD Ryzen 9 3950X 16-Core Processor
```
Sigh, I reported this to Nvidia a few months ago and they even report the issue as fixed in the release notes... https://developer.nvidia.com/nsight-compute-2020_1-new-features

SoapGentoo:
Honestly, the CUDA toolchain is **terrible** and requires tons of workarounds to make it work.

amadio:
Yes, the problem above is simple, but rather than fix in nvidia-cuda-toolkit, I think I will provide nsight-compute and nsight-systems as separate ebuilds instead. If no one is against it, of course. At least nsight-compute and nvvm are not working out of the box on our ebuilds, and for nsight-compute there is a tarball that is much easier to install.

PR #17596, opened 2020-09-19
games-action/garden: fix build with gcc 10

DarthGandalf:


games-action/garden/garden-1.0.9-r2.ebuild
```diff
@@ -1,9 +1,9 @@
-# Copyright 1999-2018 Gentoo Foundation
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
 
-inherit desktop gnome2-utils autotools flag-o-matic
+inherit desktop autotools flag-o-matic xdg-utils
```
ConiKost, line 6: Please sort alphabetically.

ConiKost:
You could also drop empty ```USE=""```.

games-action/garden/garden-1.0.9-r2.ebuild
DarthGandalf: Yeah, that's why I moved xdg-utils to the end. But sorted the rest too now.

ConiKost (approved):

PR #17597, opened 2020-09-19
gnome-extra/nemo: requires xapp 1.8

samuelbernardo:
Requires xapp 1.8 version.
This commit fixes bug 743427.

Bug: https://bugs.gentoo.org/743427
Signed-off-by: Samuel Bernardo <samuelbernardo.mail@gmail.com>


SparkyBluefang:
LGTM. I've been bumping xapp along with the other cinnamon packages, which is why I didn't notice this dependency mismatch.

juippis (commented):
Is the xapp dependency only change you're making in this ebuild? Since it fixes a compile-problem therefore revbumping is not needed. You can edit the old ebuild in place.

samuelbernardo:
> Is the xapp dependency only change you're making in this ebuild? Since it fixes a compile-problem therefore revbumping is not needed. You can edit the old ebuild in place.

Didn't test with the old ebuild, only with the new version.

juippis:
I'm talking about this:
```
>: diff nemo-4.6.5.ebuild nemo-4.6.5-r1.ebuild 
32c32
< 	>=x11-libs/xapps-1.4.0
---
> 	>=x11-libs/xapps-1.8.0
```

Because it's a build failure for some, we shouldn't revbump and force everyone to update. Just edit *nemo-4.6.5.ebuild* changing the dependency requirement.

samuelbernardo:
> Because it's a build failure for some, we shouldn't revbump and force everyone to update. Just edit _nemo-4.6.5.ebuild_ changing the dependency requirement

Done. Removed patch version updating current version dependency.

juippis (approved):

wimmuskee:
I think this PR can be closed, seems like the fix has been applied with: https://github.com/gentoo/gentoo/commit/842b34ecef048eeeeebcd19b9f9518cf6ab96db6

samuelbernardo:
> I think this PR can be closed, seems like the fix has been applied with: [842b34e](https://github.com/gentoo/gentoo/commit/842b34ecef048eeeeebcd19b9f9518cf6ab96db6)

Yes it seems is already applied as juippis instructed and the related bug is already marked as fixed in bugzilla.
So I will close this PR.

PR #17599, opened 2020-09-19
media-gfx/gimp: 2.10.20 drop dev-python/{pygtk,python2} dependencies

band-a-prend:
Bug: https://bugs.gentoo.org/711684

After later stabilization of this ebuilds the bug mentioned above could be closed.


a17r:
Here's a suggestion: Revbump only latest ebuild, start stabilisation. No one is ever going to install another ~arch version of 2.10.18 since 2.10.20 is around.

band-a-prend:
You are right, as 2.10.20 has new features.

PR #17606, opened 2020-09-19
dev-libs/tinyxml: fix libdir path in tinyxml.pc file

band-a-prend:
While using clang/llvm/lld env the on configuration phase
the information about library directory is taken from tinyxml.pc file.

On the amd64 arch system the use of lld linker instead of ldd
results in linking error as LDFLAGS is additionally filled with
`-L/usr/lib -ltinyxml` instead of `-L/usr/lib64 -ltinyxml`.

I.e. it causes linking error for dev-util/codeblocks package
on clang/llvm/lld systems.

Closes: https://bugs.gentoo.org/738948


voyageur:
Thanks for the PR! (and sorry for late pong I only saw your IRC message now)
I will check this, from quick bug read it makes sense

PR #17607, opened 2020-09-20
sys-boot/tboot: Bump to 1.10.3

salahcoronya:
This includes the fix in https://github.com/gentoo/gentoo/pull/15053. It also fixes compiling directly against gcc, so in theory it can be compiled against another compiler (like clang). It won't actually compile with clang, and that will take a A LOT of effort on this codebase, if its even possible.


salahcoronya:
Development on tboot has picked up. a 2.0.0 will be arriving, so I've create a live ebuild. In particular some of the the utilities have been renamed, and some of the command line tools have picked up new, required options (https://wiki.gentoo.org/wiki/Trusted_Boot require revisions already to reflect the new tools parameters. What works on prior to 1.9.11 does not work  on 1.9.11 which does on 1.9.12, which does not work on the current HEAD). I'm hoping the options will stabilize  when the next release is made.

salahcoronya:
Development has slowed down, so i've dropped the live ebuild

Necktwi:
when will this be published?

sys-boot/tboot/tboot-1.10.3.ebuild
```diff
@@ -0,0 +1,79 @@
...
+
+RDEPEND="${DEPEND}
+	sys-boot/grub:2
+	selinux? ( sec-policy/selinux-tboot )"
+
+DOCS=( README.md COPYING CHANGELOG )
+PATCHES=( "${FILESDIR}/${PN}-1.9.11-genkernel-path.patch" )
+
+src_prepare() {
+	sed -i 's/ -Werror//g' Config.mk || die
```
ajakk, line 31: Can we use a patch for these `sed`s instead?

sys-boot/tboot/tboot-1.10.3.ebuild
```diff
@@ -0,0 +1,79 @@
...
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="-* ~amd64 ~x86"
+IUSE="custom-cflags selinux"
+
+# requires patching the kernel src
+RESTRICT="test"
+
+DEPEND="|| ( >=dev-libs/openssl-1.1.1l-r1
```
ajakk, line 20: This seems equivalent to what it was before? Why change it?

sys-boot/tboot/tboot-1.10.3.ebuild
```diff
@@ -0,0 +1,79 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
ajakk, line 4: Can we do EAPI=8 now?

ajakk (changes requested):
I still get `cc` called directly:

```
/bin/sh: line 1: cc: command not found
make[2]: *** [Makefile:136: common/boot.o] Error 127
make[2]: *** Waiting for unfinished jobs....
make[2]: *** [Makefile:133: common/acpi.o] Error 127
/bin/sh: line 1: cc: command not found
/bin/sh: line 1: cc: command not found
make[2]: *** [Makefile:133: common/cmdline.o] Error 127
make[2]: *** [Makefile:133: common/com.o] Error 127
/bin/sh: line 1: cc: command not found
make[2]: *** [Makefile:133: common/e820.o] Error 127
/bin/sh: line 1: cc: command not found
make[2]: *** [Makefile:133: common/vtd.o] Error 127
make[2]: Leaving directory '/var/tmp/portage/sys-boot/tboot-1.10.3/work/tboot-1.10.3/tboot'
make[1]: *** [Makefile:53: build-tboot] Error 2
make[1]: Leaving directory '/var/tmp/portage/sys-boot/tboot-1.10.3/work/tboot-1.10.3'
make: *** [Makefile:48: build] Error 2
```

If this proves to be very painful to fix, I wouldn't mind dropping the Closes: tag for that bug for now as it isn't a regression.

sys-boot/tboot/tboot-1.10.3.ebuild
salahcoronya: mount-boot.eclass is not ported to EAPI 8.

sys-boot/tboot/tboot-1.10.3.ebuild
salahcoronya: CI won't pass without it. As of >=dev-libs/openssl-1.1.1l-r1 (still unstable), the bindist flag is gone. However, there are still versions of openssl with the "bindist" so this construction is needed.

ajakk: What's wrong? Seems to work for me:

```
~/gentoo/gentoo/sys-boot/tboot $ grep openssl tboot-1.10.3.ebuild
DEPEND="dev-libs/openssl[-bindist(-)]"
~/gentoo/gentoo/sys-boot/tboot $ pkgcheck scan
sys-boot/tboot
  UnstableOnly: for arches: [ amd64, x86 ], all versions are unstable: [ 1.10.3 ]
```

salahcoronya: Ah, that's exactly what I was looking for. I did not know about that construction. I will use that.

ajakk: Sorry, seems I missed the subslot so should be:

```
DEPEND="dev-libs/openssl:0=[-bindist(-)]"
```

sys-boot/tboot/tboot-1.10.3.ebuild
salahcoronya: I turned 2 of them (the ones to remove stripping and -Werror) into patches. I combined 10 them into 1, leaving 2 left. The cc issue should be fixed now too.

sys-boot/tboot/tboot-1.10.3.ebuild
salahcoronya: Done, with subslot operator

sys-boot/tboot/tboot-1.10.3.ebuild
salahcoronya: Well it apparently IS now so i updated it to EAPI=8

sys-boot/tboot/tboot-1.10.3.ebuild
```diff
@@ -0,0 +1,78 @@
...
+		-e "s/\(AS[[:space:]]\{1,\}\)?= as/\1= $(tc-getAS)/" \
+		-e "s/\(LD[[:space:]]\{1,\}\)?= ld/\1= $(tc-getLD)/" \
+		-e "s/\(CC[[:space:]]\{1,\}\)?= gcc/\1= $(tc-getCC)/" \
+		-e "s/\(CPP[[:space:]]\{1,\}\)?= cpp/\1= $(tc-getCPP)/" \
+		-e "s/\(AR[[:space:]]\{1,\}\)?= ar/\1= $(tc-getAR)/" \
+		-e "s/\(RANLIB[[:space:]]\{1,\}\)?= ranlib/\1= $(tc-getRANLIB)/" \
+		-e "s/\(NM[[:space:]]\{1,\}\)?= nm/\1= $(tc-getNM)/" \
+		-e "s/\(STRIP[[:space:]]\{1,\}\)?= strip/\1= $(tc-getSTRIP)/" \
+		-e "s/\(OBJCOPY[[:space:]]\{1,\}\)?= objcopy/\1= $(tc-getOBJCOPY)/" \
+		-e "s/\(OBJDUMP[[:space:]]\{1,\}\)?= objdump/\1= $(tc-getOBJDUMP)/" \
```
SoapGentoo, line 41: these are all `?=` defined, you can override them from outside without any patching:
`tc-export AS LD CC CPP AR RANLIB NM OBJCOPY OBJDUMP`
`LD` is a bit cheesy, seeing you generally don't need to invoke this during a build

sys-boot/tboot/tboot-1.10.3.ebuild
```diff
@@ -0,0 +1,78 @@
...
+		Config.mk || die
+
+	sed -i "s/\(CC[[:space:]]\{1,\}\)?= gcc/\1= $(tc-getCC)/" safestringlib/makefile || die
```
SoapGentoo, line 44: same here, `tc-export CC` should obviate the need for a sed here

SoapGentoo (changes requested):

sys-boot/tboot/tboot-1.10.3.ebuild
salahcoronya: Ok, i got rid of all the sed's and change over to tc-export

sys-boot/tboot/tboot-1.10.3.ebuild
```diff
@@ -52,15 +49,15 @@ src_install() {
 	emake DISTDIR="${D}" install
 
 	dodoc "${DOCS[@]}"
-	dodoc docs/*.txt lcptools/*.pdf
+	dodoc docs/*.{txt,md}
 
-	cd "${ED}"
+	cd "${D}"
```
SoapGentoo, line 54:
```suggestion
	cd "${ED}" || die
```
why did you change back to `D`?

sys-boot/tboot/tboot-1.10.3.ebuild
```diff
@@ -17,21 +17,18 @@ IUSE="custom-cflags selinux"
...
-DOCS=( README COPYING CHANGELOG )
-PATCHES=( "${FILESDIR}/${PN}-1.9.11-genkernel-path.patch" )
+DOCS=( README.md COPYING CHANGELOG )
+PATCHES=( "${FILESDIR}/${PN}-1.10.3-disable-Werror.patch"
+	  "${FILESDIR}/${PN}-1.10.3-disable-strip.patch" )
 
 src_prepare() {
-	sed -i 's/ -Werror//g' Config.mk || die
-	sed -i 's/^INSTALL_STRIP = -s$//' Config.mk || die # QA Errors
+	tc-export AS LD CC CPP AR RANLIB NM OBJCOPY OBJDUMP
```
SoapGentoo, line 31: could you look into doing this in `src_configure` instead? I assume there's no `configure` script, and it's general practice to do flag mucking in `src_configure` and not `src_prepare`.

SoapGentoo (changes requested):

sys-boot/tboot/tboot-1.10.3.ebuild
salahcoronya: Wasn't intentional, not sure what happened. I changed it back

sys-boot/tboot/tboot-1.10.3.ebuild
salahcoronya: Its now in src_configure

ajakk:
Thanks!

PR #17609, opened 2020-09-20
sys-devel/dwz: new version 0.13

annulen:
sys-devel/dwz: new version 0.13

* Updated homepage URL
* Updated SRC_URI: now dwz has proper tarball downloads

Signed-off-by: Konstantin Tokarev <annulen@yandex.ru>


sys-devel/dwz/dwz-0.13.ebuild
```diff
@@ -5,11 +5,11 @@ EAPI=7
...
-
 DESCRIPTION="DWARF optimization and duplicate removal tool"
-HOMEPAGE="https://sourceware.org/git/?p=dwz.git;a=summary"
-SRC_URI="https://src.fedoraproject.org/repo/pkgs/dwz/${P}.tar.bz2/${FPKG_HASH}/${P}.tar.bz2"
+HOMEPAGE="https://sourceware.org/dwz"
+SRC_URI="https://sourceware.org/ftp/dwz/releases/dwz-0.13.tar.xz"
+
+S="${WORKDIR}/${PN}"
 
 LICENSE="GPL-2+ GPL-3+"
```
00-matt, line 14: From looking at the COPYING file and the headers in the source files, I think this should just be `GPL-2+`.

```suggestion
LICENSE="GPL-2+"
```

00-matt (commented):
Have you thought about adopting this package (by adding yourself and the proxy-maint project to `metadata.xml`)?

***

Some comments that GitHub won't let me leave in the diff:

* Change the year in the copyright header from 2019 to 2020.
* You can remove the empty `IUSE=""` and `BDEPEND=""` lines.

Also you may want to change the PR title to `[please reassign] sys-devel/dwz: new version 0.13` so that @gentoo-bot can see that you've added the `Signed-off-by` line to your commit message.

sys-devel/dwz/dwz-0.13.ebuild
annulen: It also has COPYING3 file, and at least dwarf2.h and sha1.h use GPL3+ licensing

annulen:
Done.

sys-devel/dwz/dwz-0.13.ebuild
00-matt: Sorry, I missed that. Thank you for double checking :)

DarthGandalf (commented):
Please drop the old version in a separate commit

00-matt:
Forgot to add to add to my previous comment but when you bump a package you should copy the old version and commit that, and avoid git mv.

annulen:
Done.

DarthGandalf:
Your removal commit is not signed off. You can configure git to do it automatically, at least for this repo

annulen:
Fixed.

sys-devel/dwz/dwz-0.13.ebuild
```diff
@@ -5,20 +5,18 @@ EAPI=7
 
 inherit toolchain-funcs
 
-FPKG_HASH="1624afa75b94633e03c6e1bb952fb348"
-
 DESCRIPTION="DWARF optimization and duplicate removal tool"
-HOMEPAGE="https://sourceware.org/git/?p=dwz.git;a=summary"
-SRC_URI="https://src.fedoraproject.org/repo/pkgs/dwz/${P}.tar.bz2/${FPKG_HASH}/${P}.tar.bz2"
+HOMEPAGE="https://sourceware.org/dwz"
+SRC_URI="https://sourceware.org/ftp/dwz/releases/dwz-0.13.tar.xz"
```
ConiKost, line 10: Don't hardcode version number here. Use ```${P}.tar.xz``` instead.

sys-devel/dwz/dwz-0.13.ebuild
```diff
@@ -5,20 +5,18 @@ EAPI=7
...
+
+S="${WORKDIR}/${PN}"
```
ConiKost, line 12: Please put ```S```after ```*DEPEND``` blocks.

ConiKost:
While you are at it, you could also put the sed command in one line, since it's anyway short and drop ```"sed failed"``` since it does not say more information as a normal ```die``` anyway does.

sys-devel/dwz/dwz-0.13.ebuild
annulen: In my copy of `skel.ebuild` `S` goes between `SRC_URI` and `LICENSE`

sys-devel/dwz/dwz-0.13.ebuild
annulen: My bad

annulen:
Done.

sys-devel/dwz/dwz-0.13.ebuild
```diff
@@ -5,25 +5,21 @@ EAPI=7
...
 DEPEND="dev-libs/elfutils"
 RDEPEND="${DEPEND}"
-BDEPEND=""
 
 src_prepare() {
 	default
-	sed -i \
-		-e '/^CFLAGS/d' \
-		Makefile || die "sed failed"
+	sed -i -e '/^CFLAGS/d' Makefile
```
ConiKost, line 23: No, the ```die```command must always stay, since ```sed``` is an external command. I only meant to drop the comment after the die commands :-)

annulen: Oh, after your comment I thought ebuilds use -e mode implicitly nowadays :)

annulen:
Fixed.

sys-devel/dwz/dwz-0.13.ebuild
ConiKost: Lgfm. But please squash your sed commit with the version bump, so you will have 2 commits again.

annulen:
As you wish.

sys-devel/dwz/dwz-0.13.ebuild
ConiKost: Oops, my bad. You are right.

ConiKost (approved):

PR #17611, opened 2020-09-20
sys-devel/llvm-roc: Use flag to optionaly build runtime libraries

justxi:


candrews:
What are the runtime libraries? What are they used for / why would someone want them?

I'm trying to determine if the use flag should default to enabled and what other packages may depend on this use flag.

justxi:
These are the libraries which would be installed by "sys-libs/compiler-rt".
They are needed e.g. by "hip".

candrews:
If they're installed by sys-libs/compiler-rt, then why are they installed by this use flag too?
Should this flag conflict with sys-libs/compiler-rt or replace it?

justxi:
No. 
For the system llvm/clang libs are installed by compiler-rt.
For llvm-roc we need the corresponding libs also.

justxi:
Anything I can do here?

PR #17613, opened 2020-09-20
x11-libs/xapps: remove old and broken xapps-1.4.2.ebuild

laurantino:
Closes: https://bugs.gentoo.org/672676
Signed-off-by: Olivier Laurantin <olivier.laurantin@laposte.net>


ConiKost (changes requested):
No, that won't work. Since ```1.4.2``` is the only stable version, you just can't remove it. You need first to make a bug ticket for stable a newer version. Once the new version is stable, you can remove the old one. Or do you want as a maintainer intentionally drop stable and use testing only? Looking at your revdeps, they all seems to be in testing only.

laurantino:
>  Or do you want as a maintainer intentionally drop stable and use testing only? Looking at your revdeps, they all seems to be in testing only.

Yes, it is because there is no stable package requiring xapps that I think it is more useful to drop this old version instead of fixing the build failure (which has been fixed in more recent versions). A new version is likely to be stabilized later at the same time as cinnamon.

ConiKost (approved):

PR #17615, opened 2020-09-20
Fuse emulator updates

atomsymbol:
Please merge. Thanks.

Closes: https://bugs.gentoo.org/743394


app-emulation/fuse/fuse-1.5.7.ebuild
```diff
@@ -71,6 +71,14 @@ src_configure() {
 		myconf+=("--with-gtk")
 	fi
 
+	if use joystick; then
```
SoapGentoo, line 74: `use joystick && myconf+=( $(use_enable backend-sdl ui-joystick) )`

SoapGentoo (changes requested):

app-emulation/fuse/fuse-1.5.7.ebuild
atomsymbol: Thanks.

juippis (approved):
Thanks, please note that we like to include package name and version where the patch was included in, in the patch name. This way it's easier for people outside Gentoo find and use the patch if needed.

PR #17618, opened 2020-09-20
dev-vcs/git-lfs: Bump to 2.12.1

holgersson32644:


dev-vcs/git-lfs/git-lfs-2.12.0.ebuild
```diff
@@ -0,0 +1,62 @@
...
+
+RESTRICT+=" !test? ( test )"
+
+src_compile() {
+	set -- go build \
+		-ldflags="-X ${EGO_PN}/config.GitCommit=${GIT_COMMIT}" \
+		-mod vendor -v -work -x \
+		-o git-lfs git-lfs.go
+	echo "$@"
+	"$@" || die
```
juippis, line 34: What's the purpose of these `set`, `echo` etc? Can't you just call `go <string> || die`?

dev-vcs/git-lfs/git-lfs-2.12.0.ebuild
```diff
@@ -0,0 +1,62 @@
...
+}
+
+src_test() {
+	set -- go test \
+		-ldflags="-X ${EGO_PN}/config.GitCommit=${GIT_COMMIT}" \
+		-mod vendor \
+		./...
+	echo "$@"
+	"$@" || die
+}
```
juippis, line 54: Here too. And ideally `src_test` comes after `src_compile`.

dev-vcs/git-lfs/git-lfs-2.12.0.ebuild
holgersson32644: Fixed. I took this style from the older versions and didn't touch it as it worked.

dev-vcs/git-lfs/git-lfs-2.12.0.ebuild
holgersson32644: sure

juippis (approved):

PR #17619, opened 2020-09-20
dev-games/ogre: Version bump to 1.12.9

Yamakuzure:
This bump includes the following upgrades

* IMGUI version bumped to 1.77 ; That's what upstream bumped to
* Added USE flag "assimp" to pull in media-libs/assimp
  This also enables building of the AssimpConverter if the USE flag "assimp" is enabled.

Bug: https://bugs.gentoo.org/743742
Closes: https://bugs.gentoo.org/743742


rossbridger:
Is it possible to build it with multilib support?

Yamakuzure:
> Is it possible to build it with multilib support?

Maybe, but that that would be another bug and another PR. I haven't looked into that, yet, as I have no idea why Ogre would need it.

Yamakuzure:
The merge must be halted, as I have hit some strange problems.
Opening an issue with upstream is under way, and I'll link it here later.

thesamesam:
You should mark it as a draft.

Yamakuzure:
Found the culprit, added a patch and reported upstream here: https://github.com/OGRECave/ogre/issues/1713

Everything works normally again.

Yamakuzure:
> You should mark it as a draft.

How? Adding [draft] to the title instead of [nomerge]?

chewi:
> > Is it possible to build it with multilib support?
> 
> Maybe, but that that would be another bug and another PR. I haven't looked into that, yet, as I have no idea why Ogre would need it.

I'm generally in favour of adding multlib support wherever possible but OGRE is sufficiently complicated that I'm happy to only see it done when it's actually needed. This would probably only happen if some old proprietary game was found to use it and that seems unlikely.

Yamakuzure:
@thesamesam : The "do not merge" label is no longer needed, and I can't remove it.

thesamesam:
There's a "mark as draft" button hidden somewhere...

paroj:
just noted that opencv reference from the bug report. FYI the ogre module does not use any deprecated functions starting with opencv 4.2.0

Yamakuzure:
> just noted that opencv reference from the bug report. FYI the ogre module does not use any deprecated functions starting with opencv 4.2.0

That's great to know, thanks!

PR #17620, opened 2020-09-20
app-crypt/autofirma-1.6.5 new package

chusopr:
This is the signing software provided by Spanish government and required when submitting requests online.

The source is GPL/EUPL, but this is a binary-only package. I tried to build it from source, but I'm far from being a Java expert and after spending many hours, I found it impossible: https://github.com/ctt-gob-es/clienteafirma


ajakk:
Thanks for working on this! It'd be best to get this into [GURU](https://wiki.gentoo.org/wiki/Project:GURU) for now if still interested.

PR #17627, opened 2020-09-21
sys-boot/edk2: Rename and bump to version 202008

sveyret:
UDK is a stable subset of EDK2. But UDK is released less often and we need to use more recent versions of EDK2 (especially because of Python 2->3 migration). So I renamed the package to EDK2 and updated the version.

EDK2 is using some extenal git links. I don't know if this is the right way to do, but a use flag can be set to download and install each of these.

Note that license has changed in latest versions of EDK2, requiring the adding of a new license file (in a separate commit).


msva:
Hi there!
I'm about to bump it to 202008 under original naming (and with almost no modifications).

Although, your idea about moving and adding more USE-flags looks to be good.

Can you, please, rebase the work over latest tree state, and then we'll wait @tamiko to disscuss the changes?

Btw, just "change the name" is not enough:
1) Also you'll be need to change it's name in ebuilds of the packages that depends on it (like qemu).
2) Also, you'll need to add `move` entry in the `profiles/updates/4Q-2020`.

sveyret:
Hi @msva. As you may know, the original package was deleted a few days ago by @mgorny as it was last-rited (https://bugs.gentoo.org/735482). What I initially wanted to do is to let the old versions die and create the new version under a new name, so no need to update the depending packages (at least, not at once).

Note that afaik, qemu does not depend on it, only refind used to, but was updated some weeks ago to loose this dependency.

Note also that I will not be able to actively maintain the package. So if you think that, because it does not have any dependent, it would be better to let it die, just let me know, it won't hurt me…

So, can you confirm what you expect me to do? Do I understand well if I say that the only thing I have to do is rebase my changes? Or do I also need to add the `move` entry in `profiles/updates/4Q-2020`?

msva:
oops, sorry, I was a bit overloaded yesterday, and confused this package (or rather "rename" part in PR subject) with `sys-firmware/edk2-ovmf`.

sveyret:
So, @msva, @mgorny, do you think it is better to forget this package and simply close without merging this PR?

juippis:
Could you perhaps push this to the GURU repository? Especially if you can't actively maintain it. We don't need new bugs piling on.

We could experiment and get feedback through GURU.

sveyret:
Thank you for your feedback, @juippis, but if I put it in the GURU repository, I will not manage to use it as a dependency for rEFInd, as it was before. So I think this package will be even more useless…

juippis:
Sorry, what's the status of this one? We just bumped refind. Is this still required?

sveyret:
As you may know, rEFInd can be built with either EDK2 or GnuEfi. In the last version of rEFInd, I completely finished cleaning the links with EDK2 which was removed from the Gentoo main tree.
I gave up with this PR and so will close it.

juippis:
Sorry about that, glad you solved it.

PR #17628, opened 2020-09-21
Removal of media-tv/kodi-18* (and all dependencies on it)

candrews:
See https://bugs.gentoo.org/743535


Karlson2k:
Maybe just drop stable from all Leia components? The infrastructure is not ready for Matrix yet.
Kodi-headless versions still stick to Leia, for example.

Karlson2k:
It was masked, but still usable for those who need it. Why did you remove it?

candrews:
I think https://bugs.gentoo.org/743535#c5 says it all... I waited until the first 19 beta came out, but I didn't think it responsible to wait any longer.

PR #17632, opened 2020-09-21
sys-cluster/charliecloud: Version bump to 0.19.

olifre:
This also adds Python 3.8 since https://bugs.gentoo.org/723490 is finally resolved.


sys-cluster/charliecloud/charliecloud-0.19.ebuild
```diff
@@ -0,0 +1,79 @@
...
+	SRC_URI="https://github.com/hpc/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64 ~x86 ~x86-linux"
+fi
+
+DESCRIPTION="Lightweight user-defined software stacks for high-performance computing"
+HOMEPAGE="https://hpc.github.io/charliecloud/"
+
+SLOT="0"
+LICENSE="Apache-2.0"
+IUSE="ch-grow doc examples +pv +squashfs squashfuse"
```
juippis: What does the `pv` USE flag do here other than pull optional RDEPEND? You should put it under `optfeature`.

olifre: Completely correct, thanks for noticing! 
I have added it with `optfeature` now for both the ebuilds touched in the version bump.

juippis (approved):

PR #17633, opened 2020-09-21
dev-python/openpyxl: update to 3.0.5

winterheart:
Removed live version, updated to 3.0.5, updated SRC_URI and metadata.


dev-python/openpyxl/metadata.xml
```diff
@@ -7,8 +7,6 @@
 	</maintainer>
 	<upstream>
 		<remote-id type="pypi">openpyxl</remote-id>
-		<remote-id type="bitbucket">openpyxl/openpyxl</remote-id>
-		<bugs-to>https://bitbucket.org/openpyxl/openpyxl/issues</bugs-to>
```
sbraz, line 11: Could be replaced with https://foss.heptapod.net/openpyxl/openpyxl/issues

sbraz:
Thanks, merged!

PR #17635, opened 2020-09-22
dev-games/ogre: Re-Add "gles2" USE flag to v1.12.9-r1

Yamakuzure:
### dev-games/ogre: Re-Add "gles2" USE flag and revbump to 1.12.9-r1

* Re-Add "gles2" USE flag
* Change REQUIRED_USE to allow either "opengl" or "gles2" to be enabled

Settings in which a user runs on gles2-only can now install Ogre-1.12.x again.

Bug: https://bugs.gentoo.org/743968
Closes: https://bugs.gentoo.org/743968

### DIFF
--- a/dev-games/ogre/metadata.xml
+++ b/dev-games/ogre/metadata.xml
@@ -91,8 +91,9 @@ Exporters
         not be enabled unless an application really needs it.
     </flag>
     <flag name="freeimage">Support images via media-libs/freeimage</flag>
-    <flag name="gl3plus">Build OpenGL 3+ RenderSystem (EXPERIMENTAL)</flag>
+    <flag name="gl3plus" restrict="~dev-games/ogre-1.9.0">Build OpenGL 3+ RenderSystem (EXPERIMENTAL)</flag>
     <flag name="gles2" restrict="~dev-games/ogre-1.9.0">Build OpenGL ES 2.x RenderSystem</flag>
+    <flag name="gles2" restrict="~dev-games/ogre-1.12.9">Build OpenGL ES2/3 RenderSystem</flag>
     <flag name="gles3" restrict="~dev-games/ogre-1.9.0">Enable OpenGL ES 3.x Features</flag>
     <flag name="json">Use dev-libs/rapidjson (needed by Hlms JSON materials)</flag>
     <flag name="legacy-animations">
--- a/dev-games/ogre/ogre-1.12.9-r1.ebuild
+++ b/dev-games/ogre/ogre-1.12.9-r1.ebuild
@@ -20,23 +20,10 @@ SLOT="0/1.12"
 KEYWORDS="~amd64 ~arm ~x86"
 
 IUSE="assimp +cache cg debug deprecated doc double-precision egl examples +freeimage
-	json openexr +opengl pch profile resman-pedantic tools"
+	gles2 json openexr +opengl pch profile resman-pedantic tools"
 
-# Note: gles2 USE flag taken out for now. It seems like the Ogre Devs now rely
-#       on HLSL2GLSL (https://github.com/aras-p/hlsl2glslfork) unconditionally
-#       for GLES2. So unless we have an ebuild for that, gles2/3 are off the
-#       table.
-#       ~~sed 2020-04-26 (yamakuzure@gmx.net)
-#
-# Note: Without gles2 USE flag, the opengl USE flag is next to useless. But
-#       there are packages which enforce it, so it has to stay.
-#
-# USE="gles2"
-# REQUIRED_USE="
-# 	|| ( gles2 opengl )
-# "
 REQUIRED_USE="
-	examples? ( opengl )
+	|| ( gles2 opengl )
 "
 
 RESTRICT="test" #139905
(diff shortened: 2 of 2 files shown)


dev-games/ogre/metadata.xml
```diff
@@ -91,8 +91,9 @@ Exporters
         not be enabled unless an application really needs it.
     </flag>
     <flag name="freeimage">Support images via media-libs/freeimage</flag>
-    <flag name="gl3plus">Build OpenGL 3+ RenderSystem (EXPERIMENTAL)</flag>
+    <flag name="gl3plus" restrict="~dev-games/ogre-1.9.0">Build OpenGL 3+ RenderSystem (EXPERIMENTAL)</flag>
     <flag name="gles2" restrict="~dev-games/ogre-1.9.0">Build OpenGL ES 2.x RenderSystem</flag>
+    <flag name="gles2" restrict="~dev-games/ogre-1.12.9">Build OpenGL ES2/3 RenderSystem</flag>
```
rossbridger, line 96: Can't we just write 'restrict="&lt;dev-games/ogre-2.0.0' instead of duplicating flag entries?

rossbridger: Actually it seems ogre-2.x also support gles2 rendering system, probably you should also add gles2 support to ogre-2.x ebuild then these restricts are not needed at all.

dev-games/ogre/ogre-1.12.9-r1.ebuild
```diff
@@ -54,6 +41,11 @@ RDEPEND="
 	cg? ( media-gfx/nvidia-cg-toolkit )
 	egl? ( media-libs/mesa[egl] )
 	freeimage? ( media-libs/freeimage )
+	gles2? (
+		dev-util/glsl-optimizer
+		dev-util/hlsl2glslfork
```
rossbridger, line 46: Are these two packages really mandatory for gles2 support? I removed these libraries from RDEPEND and ogre still compiles fine with gles2 USE.

rossbridger (changes requested):

dev-games/ogre/ogre-1.12.9-r1.ebuild
Yamakuzure: OH DAMMIT! You are right.

For all versions in 1.12.x both got included and used unconditionally whenever GLES2 was enabled.

And now that I finally have working ebuilds, they fixed it. :angry: 

...*sigh*... I should be glad, of course, and will reverse those dependencies.

However, as configure states:
````
-----------------------------------------------------------------------------
-- The following OPTIONAL packages could NOT be located on your system.
-- Consider installing them to enable more features from this software.
+ Softimage: Softimage SDK needed for building XSIExporter <FALSE>
+ GLSL Optimizer: GLSL Optimizer <http://github.com/aras-p/glsl-optimizer/>
+ HLSL2GLSL: HLSL2GLSL <http://hlsl2glslfork.googlecode.com/>
-----------------------------------------------------------------------------
````

Both tools _are_ usedful. So I really would like to have them included, although with their own individual USE flags, of course. Do you thik that would be in order?

dev-games/ogre/metadata.xml
Yamakuzure: Yes, yes... of course, you are right.

Yamakuzure: I guess I got caught up in the "ES 2.x" versus "ES2/3" difference. Maybe just write "OpenGL ES 2+" instead?

rossbridger: I noticed there was a 'gles3' flag in older ogre version. Do you think we should follow the convention and make a separate gles3 flag?

dev-games/ogre/ogre-1.12.9-r1.ebuild
rossbridger: Of course they would need their own USE flags, and they probably have their usefulness on full opengl rendering system thus not reliant on gles2 flag. But I wouldn't introduce additional packages only being specifically used for enabling optional features. If anything, python support would be a more worthy feature.

dev-games/ogre/metadata.xml
Yamakuzure: Actually I would like to see ogre-1.9.0 get dumped. But yes, that is what came to my mind, too.
GLES2/3 are separated in Ogre-1.9, while it is unified since ogre-1.10.
Well, the gles2 USE flag is disabled in Ogre-2.1, because that one isn't ported, yet.
So I came to the conclusion, as those USE flag descriptions should be left in place as they are. At least for now.

I also doubt that it ever will, because main development is with Ogre-2.2
v2.1 came out in april, while v2.2 had five release since then.

Once we have sorted out Ogre-1.12.9, I'd like to have another look at v2.2. Maybe we should ditch v2.1 altogether, like we did with v2.0.

Clarification: The v2.0 branch was a hybrid, used to push forward the move from CPU centric (v1.x) to GPU powered (v2.x) rendering. The latter is clearly the thing, but right now it looks to me like v2.1 is a transition stage like v2.0 was.

dev-games/ogre/ogre-1.12.9-r1.ebuild
Yamakuzure: Well, they allow you to import DX9 shaders and have them optimized. That's very useful. 

Actually both hlsl2glsl.h and glsl_optimizer.h are only used in RenderSystems/GLES2.

Yamakuzure:
@rossbridger : I have reduce the PR to re-enabling GLES2. I will update the bug now to reflect that, and have moved inclusion of hlsl2glsl and glsl-optimizer in another branch for re-consideration.

Yamakuzure:
Could I please ... ping?

thesamesam:
Would need rebasing.

PR #17639, opened 2020-09-22
sci-astronomy/stellarium: version bump

DarthGandalf:


dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
```diff
@@ -0,0 +1,59 @@
...
+S="${WORKDIR}/qtcompress-${MY_COMMIT}"
+
+RDEPEND="
+	dev-qt/qtcore:5
+	sys-libs/zlib:=
+"
+DEPEND="${RDEPEND}"
+BDEPEND="
+	virtual/pkgconfig
+	test? ( dev-qt/qttest:5 )
```
juippis, line 27: This is a rather interesting package because it installs tons of headers and few libraries, but no binaries.

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
```diff
@@ -0,0 +1,59 @@
...
+"
+
+PATCHES=(
+	"${FILESDIR}/${P}-remove-zlib.patch"
+	"${FILESDIR}/${P}-test-include.patch"
+)
+
+src_prepare() {
+	default
+	rm -r src/3rdparty/ || die
```
juippis, line 37: Could you add a comment above the `rm` explaining what it does? To me it looks like it removes bundled lib code, but for anyone reading and wondering. People shouldn't be guessing, and then one day upstream might drop the dir and this package is m-n and someone looking at it thinking what was this for.

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
```diff
@@ -0,0 +1,59 @@
...
+		"INCLUDEPATH+=$("${pkg_config}" --cflags zlib)" \
+		"LIBS+=$("${pkg_config}" --libs zlib)"
+}
+
+src_test() {
+	cd tests/auto/unit || die
+	eqmake5 \
+		"INCLUDEPATH+=${S}/src/compress"
+	emake
+	virtx qzip/target_wrapper.sh qzip/tst_qzip || die
```
juippis, line 52: `virtx` is a `nonfatal` function so don't add `|| die` here at all.
https://devmanual.gentoo.org/ebuild-writing/error-handling/index.html
https://devmanual.gentoo.org/eclass-reference/virtualx.eclass/index.html

Basically it opens a new Xvfb session and if you kill the command, the X session may still be left open.

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
```diff
@@ -0,0 +1,59 @@
...
+
+inherit qmake-utils toolchain-funcs virtualx
+
+DESCRIPTION="A public API for QZip in an easy to use module"
+
+HOMEPAGE="https://github.com/nezticle/qtcompress"
+MY_COMMIT="23f8831826cd72aedf99fc3699148b6c994fd677"
+SRC_URI="https://github.com/nezticle/qtcompress/archive/${MY_COMMIT}.tar.gz -> ${P}.tar.gz"
+LICENSE="LGPL-2.1"
+SLOT="0/${PV}"
```
juippis, line 14: Are you absolutely 100 % percent sure this is needed? Note that because this is a snapshot, it'd change with every commit you snapshot into ebuild pretty much. That doesn't sound right...

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
```diff
@@ -0,0 +1,59 @@
...
+EAPI=7
+
+inherit qmake-utils toolchain-funcs virtualx
+
+DESCRIPTION="A public API for QZip in an easy to use module"
+
+HOMEPAGE="https://github.com/nezticle/qtcompress"
+MY_COMMIT="23f8831826cd72aedf99fc3699148b6c994fd677"
+SRC_URI="https://github.com/nezticle/qtcompress/archive/${MY_COMMIT}.tar.gz -> ${P}.tar.gz"
+LICENSE="LGPL-2.1"
```
juippis, line 13: The upstream Git repo has more licenses listed. I know it bundles some stuff and some of the licenses could be related to that, but looking at 
https://github.com/nezticle/qtcompress/blob/master/src/compress/qzip.cpp
it lists a choice of different licenses being used here.

Now I am **not** a license expert but looks to me this could be a case for `LICENSES="|| ( 1 2 3 )` and maybe a comment above it.

dev-libs/qxlsx/qxlsx-1.3.56.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+EAPI=7
+
+inherit cmake qmake-utils
+
+DESCRIPTION="Excel file(*.xlsx) reader/writer library using Qt5"
+
+HOMEPAGE="https://github.com/QtExcel/QXlsx"
+SRC_URI="https://github.com/QtExcel/QXlsx/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+LICENSE="MIT"
```
juippis, line 12: Add a newline here like a default .skel would do.

dev-libs/qxlsx/qxlsx-1.3.56.ebuild
```diff
@@ -0,0 +1,39 @@
...
+SLOT="0/${PV}"
```
juippis, line 13: Does this also change ABI every minor release? Maybe 1.3 -> 1.4 makes more sense. But are you sure about that either?

sci-astronomy/stellarium/stellarium-0.20.3.ebuild
```diff
@@ -0,0 +1,135 @@
...
+	"${FILESDIR}/${P}-unbundle-qxlsx.patch"
+	"${FILESDIR}/${P}-unbundle-qtcompress.patch"
+	"${FILESDIR}/${P}-unbundle-zlib.patch"
+)
+
+src_prepare() {
+	cmake_src_prepare
+	use debug || append-cppflags -DQT_NO_DEBUG #415769
+
+	rm -r src/external/{libindi,qxlsx,qtcompress,zlib}/ || die
```
juippis, line 75: (I'd like a comment above this `rm` too but since this ebuild doesn't need other changes before the first questions are solved, it can be added later)

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
DarthGandalf: It's because it's a snapshot… I can't guarantee they won't change ABI, or even API, because they have no releases.

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
DarthGandalf: ||die applies to the whole virtx command, not the part of it wrapped by virtx, so ||die here is harmless. The guide means that the wrapped command should not die. But sure.

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
DarthGandalf: The soname it uses, is the same as of Qt version it linked to

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
DarthGandalf: Okay?

dev-libs/qxlsx/qxlsx-1.3.56.ebuild
DarthGandalf: Without my patch it can't even be installed. I'll ask upstream in https://github.com/QtExcel/QXlsx/pull/114 though.

sci-astronomy/stellarium/stellarium-0.20.3.ebuild
```diff
@@ -17,15 +25,20 @@ SRC_URI="
...
 
+# Python interpreter is used while building RemoteControl plugin
 BDEPEND="
+	${PYTHON_DEPS}
+	doc? ( app-doc/doxygen[dot] )
 	nls? ( dev-qt/linguist-tools:5 )
 "
 RDEPEND="
+	dev-libs/qtcompress:=
+	=dev-libs/qxlsx-1.3.56-r0:=
```
juippis, line 41: Is this very restrictive dep due to your patch? Since
`=dev-libs/qxlsx-1.3*:=`
or
`=dev-libs/qxlsx-1.3.56*:=`
should also work.

DarthGandalf: Yes. Depending on how upstream accepts my patch, I may need to change how stellarium finds it.

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
DarthGandalf: I see what you mean. I've put it to BDEPEND because python eclasses put `test? ( ${RDEPEND} )` into BDEPEND instead of DEPEND. Because cross-compilation isn't compatible with tests without something like qemu-user

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
DarthGandalf: What about making it 0_p20180513_p5_p14_p2:5.14.2? Then it needs to be updated with every qt update

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
juippis: It's got something to do with how python executables import their modules at runtime which is not the case with compiled products. AFAIK Perl has similar issues.

https://qa-reports.gentoo.org/output/genrdeps/dindex/dev-qt/qttest
https://qa-reports.gentoo.org/output/genrdeps/bindex/dev-qt/qttest

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
juippis: Sounds good to me. Does this work with Qt-5.15?

DarthGandalf: Yes, but for the subslotted version like that, it would need to depend on the exact Qt version

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
DarthGandalf: I see. Done.

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
DarthGandalf: I think I got confused somewhere. This will be much simpler.

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
```diff
@@ -0,0 +1,64 @@
...
+
+HOMEPAGE="https://github.com/nezticle/qtcompress"
+MY_COMMIT="23f8831826cd72aedf99fc3699148b6c994fd677"
+SRC_URI="https://github.com/nezticle/qtcompress/archive/${MY_COMMIT}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="|| ( LGPL-2.1 GPL-3 ) FDL-1.3"
+SLOT="0/5.11.0"
+KEYWORDS="~amd64 ~x86"
+IUSE="test"
+RESTRICT="!test? ( test )"
```
juippis, line 18: You could separate these with an empty line next time touching this.

juippis (commented):
Thanks!

dev-libs/qtcompress/qtcompress-0_p20180513.ebuild
DarthGandalf: Separate them from each other, or from something else?

juippis: Yes, I still find it too clumped together, makes it more readable to add one newline between.
```suggestion
LICENSE="|| ( LGPL-2.1 GPL-3 ) FDL-1.3"
SLOT="0/5.11.0"
KEYWORDS="~amd64 ~x86"
IUSE="test"

RESTRICT="!test? ( test )"
S="${WORKDIR}/qtcompress-${MY_COMMIT}"
```
Although I'd personally even separate `S` from this group, but style things.

PR #17647, opened 2020-09-23
Install ddclient cache directory with proper permissions

emlove:
## Description
Previously, the ddclient cache directory was not created until a reboot. Switching to the new tmpfiles eclass lets us process the tmpfile config after installing to ensure the service can be started immediately.

### DIFF
--- a/net-dns/ddclient/ddclient-3.9.1.ebuild
+++ b/net-dns/ddclient/ddclient-3.9.1.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-inherit systemd
+inherit systemd tmpfiles
 
 DESCRIPTION="Perl client used to update dynamic DNS entries"
 HOMEPAGE="https://ddclient.net/"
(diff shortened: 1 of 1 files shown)


ConiKost:
Eh, why do you need that? The init script (OpenRC) and tmpfiles (Systemd) are already creating that directory.
So it should be created for you?!

ddclient.initd-r6: ```checkpath -d -m 0700 -o ${command_user}:${command_group} /var/cache/ddclient```
ddclient.tmpfiles: ```d /var/cache/ddclient 0700 ddclient ddclient - -```

emlove:
I might have something misconfigured on my end then. I'm using systemd, but it was definitely not creating the directory.

emlove:
> Eh, why do you need that? The init script (OpenRC) and tmpfiles (Systemd) are already creating that directory.
> So it should be created for you?!
> 
> ddclient.initd-r6: `checkpath -d -m 0700 -o ${command_user}:${command_group} /var/cache/ddclient`
> ddclient.tmpfiles: `d /var/cache/ddclient 0700 ddclient ddclient - -`

Thanks for pointing that out! Turns out the problem here is that we're not processing the new tmpfiles config after installing, so the service is broken until a reboot regenerates tmpfiles. I've updated the ebuild to use the newer `tmpfiles` eclass which can process the tmpfiles after installation to ensure the service is ready to start immediately.

ConiKost (approved):
Great catch! lgfm.

PR #17648, opened 2020-09-23
app-crypt/tpm2-tss: Bump to 2.4.3 and 3.0.1 to fix CVE-2020-24455

salahcoronya:
From upstream:

Fix CVE-2020-24455 FAPI PolicyPCR not instatiating correctly
Note that all TPM object created with a PolicyPCR with the currentPcrs
and currentPcrsAndBank options have been created with an incorrect policy
that ommits PCR checks. All these objects have to be recreated!


juippis (approved):
```
 * The tmpfiles processor exited with a non-zero exit code
```
Was there a fix for this? Could you search through bugzilla?

salahcoronya:
As far as the "The tmpfiles processor exited with a non-zero exit code" this is a problem with opentmpfiles. See https://bugs.gentoo.org/741216

PR #17651, opened 2020-09-24
media-libs/libheif: bump to 1.9.1

jsmolic:


novomesk:
libheif upgraded to 1.9.1

jsmolic:
@novomesk Done, thanks for the info :)

media-libs/libheif/libheif-1.9.1.ebuild
```diff
@@ -6,18 +6,18 @@ EAPI=7
...
+	SRC_URI="https://github.com/strukturag/libheif/releases/download/v${PV}/${P}.tar.gz"
 	KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
 fi
 
 DESCRIPTION="ISO/IEC 23008-12:2017 HEIF file format decoder and encoder"
 HOMEPAGE="https://github.com/strukturag/libheif"
 
 LICENSE="GPL-3"
-SLOT="0/1.8"
+SLOT="0/1.9.1"
```
juippis, line 20: Maybe `1.9` is enough since rarely minor versions change ABI. Are you sure this isn't compatible with previous 1.8 either?

juippis (commented):
Actually now that we are here, there are multiple options / automagic deps that aren't dealt with in this ebuild.
https://github.com/strukturag/libheif/blob/2e218ef40440e258b5da1e513f72c7f3b1611c7c/configure.ac#L93

the libaom one is important because the latest version of GIMP supports avif through libheif having libaom support. Could you check through these? 

And you can switch to CMake if you find it easier.

band-a-prend:
The `libheif` supports encode/decode via `libaom`
and support encode via `rav1e` and `dav1d` (I  assume there is a typo and last should enable `have_avif_decoder="yes"` instead of `have_avif_encoder="yes"` on libheif configure.ac).

As gentoo proxied-maintainer of `media-gfx/gimp` I ask to add permanent dependence of `media-libs/libheif` on `media-libs/libaom` as it provide simultaneously encode/decode.

band-a-prend:
There is additional info from communication with 
Dirk Farin, the developer of libheif, about my assumption of dav1d option:

> that is a typo. 
> I've corrected it to "have_avif_decoder" now.

Also I asked:
> Also as a Gentoo Linux proxied-maintainer of Gimp package I'm interesting what library libheif choose to decode and encode AVIF if all of libaom, librav1e, dev1d libraries are in the system at the libheif build moment?
> Could you please clarify this?

And receive the very usefull clarification:
> - rav1e is used for encoding if available, otherwise libaom.
> - dav1d is used for decoding if available, otherwise libaom.
The user may override the encoder used.

> For heif, there is only one option: libde265 for decoding, libx265 for encoding.
```

**Comment by jsmolic** [comment]
So the situation with dependencies is rather messy here, I was working with this ebuild not long ago and I noticed that, but forgot to fix that on this version bump.\
libheif should definitely uncoditionally depend on libaom for now. Even though `rav1e` and `dav1d` can be used for encoding/decoding AVIF, libheif build system only has the option to use local builds of those packages, which is not good for us here.\
There are open pull requests upstream to fix this to an extent (https://github.com/strukturag/libheif/pull/330 and https://github.com/strukturag/libheif/pull/349), same thing should be done for `dav1d` and then add appropriate configure switches, so then we could have USE flags to choose between encoders/decoders. But until that is fixed, I think we'll have to depend on libaom unconditionally.

File: media-libs/libheif/libheif-1.9.1.ebuild (hunk 1 above)
Comment (jsmolic) [reply media-libs/libheif/libheif-1.9.1.ebuild]: I don't think it's compatible, but 1.9 will be enough

**Comment by jsmolic** [comment]
I'll add myself to maintainer as I was working on this package a bit before, it'll be easier for me to track an clean up everything that way.

File: media-libs/libheif/libheif-1.9.1.ebuild (hunk 1 above)
Comment (thesamesam) [reply media-libs/libheif/libheif-1.9.1.ebuild]: Note that this is about ABI compatibility, not API.

i.e. Will a linked binary against the old version still work correctly against the new one? Were the exposed existing symbols changed or removed? (New ones are fine).

Comment (jsmolic) [reply media-libs/libheif/libheif-1.9.1.ebuild]: Few added symbols, but none removed, so it should be backwards compatible. That means the sub-slot doesn't have to be updated I guess? Previously it was changed on every version bump, so I was working in line with that

**Comment by novomesk** [comment]
Regarding the media-libs/libaom dependency for AVIF:
The old libaom-1.0.0 from June 2018 is not very recommended for AVIF.
[libaom v1.0.0-errata1-avif](https://aomedia.googlesource.com/aom/+/refs/tags/v1.0.0-errata1-avif) from December 2019 or newer should be used for AVIF.
media-libs/libaom-2.0.0 is available. People who encode AV1 videos or save AVIF, generally don't use the old libaom-1.0.0 
I would use
`>=media-libs/libaom-2.0.0`

**Comment by jsmolic** [comment]
@novomesk Thank you for pointing that out, I will update the dependency.

**Comment by band-a-prend** [comment]
> There are open pull requests upstream to fix this to an extent, same thing should be done for `dav1d` and then add appropriate configure switches, so then we could have USE flags to choose between encoders/decoders. 

I'm afraid this changes would complicate explicit choose of encode/decode switch as in case of presence of rav1e, dav1d libraries in system they become preferred automaticaly. Maybe I panic ahead of time.

File: media-libs/libheif/libheif-1.9.1.ebuild [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+
+BDEPEND="test? ( dev-lang/go )"
+DEPEND="
+	>=media-libs/libaom-2.0.0:=[${MULTILIB_USEDEP}]
+	media-libs/libde265:=[${MULTILIB_USEDEP}]
+	media-libs/libpng:0=[${MULTILIB_USEDEP}]
+	media-libs/x265:=[${MULTILIB_USEDEP}]
+	sys-libs/zlib:=[${MULTILIB_USEDEP}]
+	virtual/jpeg:0=[${MULTILIB_USEDEP}]
+	gdk-pixbuf? ( x11-libs/gdk-pixbuf )
```
juippis, line 34: Won't this need a `MULTILIB_USEDEP` as well? 

Looks good otherwise.

jsmolic: Done

band-a-prend:
I tested emerge `libaom-2` + rebuild `libheif-1.7.0-r1` and `gimp-2.10.22` and this combinations seems doesn't allow to use heif/avif support.
If I use  `libaom-2` + rebuild `libheif-1.8.0` and `gimp-2.10.22` then heif/avif is presence in export dialog format list.

novomesk:
@band-a-prend That's expected. AVIF support in GIMP is built only when libheif is 1.8.0 or newer.

juippis (approved):
I'm thinking let's give it a go and wait for feedback. Please be active in checking bugs regarding this change.

PR #17655, opened 2020-09-24
WIP: dev-java/openjdk-bin: Latest jdk release jdk15

dblaber:
Adding ebuild for the latest non-lts jdk release, jdk15.

Signed-off-by: Darren Blaber dmbtech@gmail.com


dev-java/openjdk-bin/openjdk-bin-15_p36.ebuild
```diff
@@ -0,0 +1,111 @@
...
+		rm -v lib/src.zip || die
+	fi
+
+	rm -v lib/security/cacerts || die
+	dosym ../../../../etc/ssl/certs/java/cacerts \
+		"${dest}"/lib/security/cacerts
+
+	dodir "${dest}"
+	cp -pPR * "${ddest}" || die
+
```
gyakovlev, line 87: check for `::gentoo` version

I have a stable symlink here

```
# provide stable symlink
    dosym "${P}" "/opt/${PN}-${SLOT}"
```

dev-java/openjdk-bin/openjdk-bin-15_p36.ebuild
```diff
@@ -0,0 +1,111 @@
...
+	$(abi_uri aarch64 arm64)
+	$(abi_uri ppc64le ppc64)
+	$(abi_uri x64 amd64)
+"
+
+DESCRIPTION="Prebuilt Java JDK binaries provided by AdoptOpenJDK. Short Term Supported"
+HOMEPAGE="https://adoptopenjdk.net"
+LICENSE="GPL-2-with-classpath-exception"
+KEYWORDS="~amd64 ~arm64"
+IUSE="alsa cups -gentoo-vm headless-awt source"
```
gyakovlev, line 29: replace the `-` with `+` , it will be masked, but enabled.

gyakovlev (changes requested):
minor nits.
otherwise works =) but I'll test it more a bit later.

ConiKost:
Since we have now jdk17, I would close this, as version 12-16 are non lts, so I don't think we need them anymore.

PR #17361, opened 2020-09-01
sci-geosciences/grass: Bump to version 7.8.3 stable release, and 7.9 dev release

tmszi:
Bump to current GRASS GIS stable version [7.8.3](https://github.com/OSGeo/grass/releases/tag/7.8.3) (which support python3/wxPython4), and dev version 7.9.

PR #15399, opened 2020-04-18
sci-geosciences/grass: Version bump to 7.8.2

dermoench42:
python3, proj-6, build

Signed-off-by: Ervin Peters ervin.peters@ervnet.de
Bug: https://bugs.gentoo.org/711428
Package-Manager: Portage-2.3.89, Repoman-2.3.20

PR #17666, opened 2020-09-25
sci-geosciences/grass: readd package

hamarituc:
Readd package for the GRASS Geographic Information System. GRASS was removed from the tree due to dropping Python 2 Support. This commit readds a recent version of GRASS which doesn't depends on Python 2.

This Pull Request merges the efforts of #15399 and #17361, implements the suggested improvements and fixes minor bugs of the referenced PR:
- Update to EPAI 7
- Drop unneeded dependency to `media-libs/libprojectm`
- Add missing dependency to `dev-python/six`
- Put GRASS 7.8 and 7.9 in the same slot, but different subslots

Bug: https://bugs.gentoo.org/744724
Bug: https://bugs.gentoo.org/682188
Bug: https://bugs.gentoo.org/735458


SoapGentoo:
thanks, will have a look at this tomorrow

PR #17667, opened 2020-09-25
net-libs/pjproject: version bump

jkroonza:
Closes: https://bugs.gentoo.org/744613


jkroonza (commented):
because github doesn't diff nicely.

copy + add vpx USE= flag.

PR #17677, opened 2020-09-26
media-video/vidcutter Fix dependencies and QA issues

chusopr:
* ~~Changed dependency on stale virtual (#744787)~~
* Added missing dependency dev-python/pyopengl (#723378)
* Remove irrelevant doc files installed manually triggering QA warnings.
* Trigger update-mime-database after installing new MIME handlers.

Closes: https://bugs.gentoo.org/723378

PR #17678, opened 2020-09-26
profiles: deprecate virtual/ffmpeg in favour of only provider

chusopr:
See https://bugs.gentoo.org/744787

Another PR is pending to fix the only package still pulling virtual/ffmpeg in `::gentoo`: https://github.com/gentoo/gentoo/pull/17677

This PR adds `virtual/ffmpeg` to `package.deprecated` as adviced in #gentoo-dev-help to trigger a warning if overlays are still using it.

### DIFF
--- a/profiles/package.deprecated
+++ b/profiles/package.deprecated
@@ -17,6 +17,11 @@
 
 #--- END OF EXAMPLES ---
 
+# Jesus P Rey <gentoo@chuso.net> (2020-09-26)
+# Only one provider left; media-video/ffmpeg should be used instead
+# Bug #744787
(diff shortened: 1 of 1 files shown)


chusopr:
Why does @gentoo-bot think this is a new package?

chusopr:
This was marked as broken by @gentoo-repo-qa-bot because it deprecates `virtual/ffmpeg` still pulled by `media-video/videocutter` but https://github.com/gentoo/gentoo/pull/17677 was raised to change that dependency.

zmedico:
Thanks!

PR #17679, opened 2020-09-26
media/gfx-cura-4.6.1 : Fix Startup Crash at Runtime

digitall:
The removal of dev-python/uranium-4.6.1 resulted in cura-4.6.1 using the wrong library dependency. This built fine, but crashes at runtime with various incompatiblities in the Python library API.

This is fixed by restoring the v4.6.1 uranium and associated patches and fixing the cura ebuild to correctly match the dependencies to the version of cura i.e. use "~....${PV}" to match to v4.6.1(-r<n>), but not match to higher or lower versions...


digitall:
I will look at adding a futher Pull Request for cura v4.7.0+ shortly...

digitall:
I also see that the gentoo-bot doesn't update the "no signoff" label on a force-push to fix the commits :/

digitall:
Right have unmerged cura and done a depclean to remove all dependencies and tried a full re-emerge of "emerge -av cura" with successful build and startup. Two minor issues:
1. You need to ensure that you have selected the python 3 interpreter with eselect  to match the one used to build cura i.e. set in PYTHON_TARGETS. If not, it starts but you get a "import Arcus" failure... which is a bit obtuse and confused me for a bit.
2. You need to go through the startup reconfiguration... though that might be as I think I reset the cura configuration to see if that fixed the runtime issue, but there are no problems and once done, the $HOME/.config/cura/<version> files are back as expected.

digitall:
It was easier to add the cura v4.7.0 ebuild to this PR and this also provides another way of solving the overall issue i.e. currently the portage tree has no working version of cura... until / unless this is merged.

Also, the @gentoo-repo-qa-bot broken status is likely spurious as it relates to missing dependencies when building on ARM64 which was likely the case with the builds prior to my changes... Anyway, will now await human comment :)

digitall:
Related... https://bugs.gentoo.org/731274

thesamesam:
The label gets updated if you add '[please reassign]'. The issue on arm64 is that libarcus-1.6.1 isn't keyworded on it. So drop keywords for arm64 on cura, then file a bug to re-add it once this is merged.

digitall:
@thesamesam : OK. Will squash the v4.7.0 QA fix into the previous commit with rebase... Will check the other changes.

[please reassign]

thesamesam:
(Sorry, you have to put it in the PR title -- I wasn't very clear!)

digitall:
@thesamesam : Ah OK!

digitall:
@thesamesam : All good now, prior to review?

digitall:
@expeditioneer : I see you added media-gfx/fdm-materials-4.7.0 a few days ago as https://github.com/gentoo/gentoo/commit/d756f13ebc9125f75727de035cc5f91a74b7277c so my change in this PR to add this was not needed. It was identical anyway.

I have now rebased this to remove that ebuild and rebase onto the latest master to keep the tree consistent.

I hope the fixes to the cura ebuilds can get merged shortly so users have a cura build that actually works in tree :|

PR #17694, opened 2020-09-27
app-emulation/virt-manager: Version updated to 3.1.0.

jpds:
Tested on amd64 system.


gyakovlev:
I'm pretty sure upstream switched from pod2man to python's docutils, or did it happen in 3.1 (which is already out)

gyakovlev:
yup they did

https://github.com/virt-manager/virt-manager/commit/65a39a74950722eecaf165a42bd743ec86b4ef63

so probably can remove perl dependency and add docutils to `DEPEND`

jpds:
@gyakovlev Thank you for the suggestion, I have now added this.

gyakovlev:
thanks, I will test today and merge it it works ok for me.

gyakovlev:
works beautifully, and thanks for catching argcomplete dep!

PR #17699, opened 2020-09-28
dev-embedded/u-boot-tools: bump to 2020.10

jsmolic:


dev-embedded/u-boot-tools/u-boot-tools-2020.07.ebuild
```diff
@@ -0,0 +1,71 @@
...
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~riscv ~x86"
+
+BDEPEND="
+	sys-devel/bison
+	sys-devel/flex"
+
+S=${WORKDIR}/${MY_P}
```
SoapGentoo, line 21: quotes

SoapGentoo (changes requested):

dev-embedded/u-boot-tools/u-boot-tools-2020.07.ebuild
jsmolic: Done ;)

dev-embedded/u-boot-tools/u-boot-tools-2020.10.ebuild
```diff
@@ -12,59 +12,71 @@ SRC_URI="ftp://ftp.denx.de/pub/u-boot/${MY_P}.tar.bz2"
 
 LICENSE="GPL-2"
 SLOT="0"
-KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
```
ajakk, line 15: (Commenting on a different line because the review thing doesn't let me comment on a line you didn't change)

Why does the SRC_URI use the ftp protocol? It seems like the distfiles are also served over HTTPS, can you switch to that for security?

dev-embedded/u-boot-tools/u-boot-tools-2020.10.ebuild
```diff
@@ -12,59 +12,71 @@ SRC_URI="ftp://ftp.denx.de/pub/u-boot/${MY_P}.tar.bz2"
...
-IUSE=""
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~riscv ~x86"
+IUSE="minimal"
```
ajakk, line 16: Why did you add this? It seems like the only difference is about 800kb on my system.

dev-embedded/u-boot-tools/u-boot-tools-2020.10.ebuild
```diff
@@ -12,59 +12,71 @@ SRC_URI="ftp://ftp.denx.de/pub/u-boot/${MY_P}.tar.bz2"
...
-	dobin bmp_logo dumpimage fdtgrep gen_eth_addr img2srec mkenvimage mkimage
 	dobin env/fw_printenv
+
+	if ! use minimal; then
+		dobin bmp_logo dumpimage fdtgrep gen_eth_addr img2srec mkenvimage mkimage
+	fi
+
 	dosym fw_printenv /usr/bin/fw_setenv
 	insinto /etc
 	doins env/fw_env.config
```
ajakk, line 82: Why not wrap the whole string passed to doman in quotes?

dev-embedded/u-boot-tools/u-boot-tools-2020.10.ebuild
```diff
@@ -12,59 +12,71 @@ SRC_URI="ftp://ftp.denx.de/pub/u-boot/${MY_P}.tar.bz2"
...
+		tools-only_defconfig
+
+	if use minimal; then
+		_emake envtools
+	else
+		_emake tools-all
+	fi
 }
 
 src_test() { :; }
```
ajakk, line 70: Whoops! Missed something in the other review. Why is src_test a no-op? It seems like there are tests, if they're broken they should be patched to run or they should have RESTRICT="test" (preferably with an explanation in comments).

dev-embedded/u-boot-tools/u-boot-tools-2020.10.ebuild
```diff
@@ -12,59 +12,71 @@ SRC_URI="ftp://ftp.denx.de/pub/u-boot/${MY_P}.tar.bz2"
 
 LICENSE="GPL-2"
 SLOT="0"
-KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
-IUSE=""
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~riscv ~x86"
+IUSE="minimal"
 
 BDEPEND="
```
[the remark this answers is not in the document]
jsmolic: Sure, this was mostly copy-paste bump, hence the style niggles

dev-embedded/u-boot-tools/u-boot-tools-2020.10.ebuild
jsmolic: Didn't go that far into it, it has been this way for several previous version... but I'll take a look

dev-embedded/u-boot-tools/u-boot-tools-2020.10.ebuild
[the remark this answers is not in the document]
jsmolic: Actually I prefer to do it this way if there are more deps, to keep the tabs consistent

dev-embedded/u-boot-tools/u-boot-tools-2020.10.ebuild
jsmolic: I prefer to quote only `${S}`, looks nicer

dev-embedded/u-boot-tools/u-boot-tools-2020.10.ebuild
jsmolic: Because I have a use-case where only envtools are needed, instead of all binaries, and that size can make a difference on embedded systems

jsmolic:
This is done in https://github.com/gentoo/gentoo/commit/f4f8e821f7c750d3db6e6828a8cd70dc272c4dbd

PR #17702, opened 2020-09-28
app-emulation/crun: bump to 0.15

robertgzr:
~~and drop 0.14.1~~


robertgzr:
I'm going to need some advise here... it looks like sys-process/criu is in testing for `ppc64`, why doesn't this work then?

gyakovlev:
criu is not masked on big-endian ppc64

so you should add a flag mask to

`profiles/arch/powerpc/package.use.mask`
and unmask to
`arch/powerpc/ppc64/64le/package.use.mask`


if you having trouble navigating ppc64 profiles I can do on merge, so let me know.

another 2 changes required:

1) don't drop old version yet, as new one may have bugs, keep at least one so users can use what works, so drop the removal commit.

2) add `Signed-off-by` to the commit. Can't merge without sign-off.

robertgzr:
@gyakovlev looks like there already are masks like what you describe?

```
$> rg criu ../../profiles/arch/powerpc/
../../profiles/arch/powerpc/package.mask
5:# criu does not work on ppc64be and ppc32
6:sys-process/criu

../../profiles/arch/powerpc/ppc64/64le/package.mask
5:# criu works on ppc64le
6:-sys-process/criu
```

I fixed up the pr as suggested, I would appreciate if you looked into the ppc64 stuff o/

gyakovlev:
pushed, thanks!

yes, because criu is masked, but [criu] useflag pulls is, we need also mask app-emulation/crun[criu] flag there.

https://github.com/gentoo/gentoo/commit/154279135ae6f148ed829a75499131f4e99a5202 masking
https://github.com/gentoo/gentoo/commit/e133ccef7f6a5300105f8b2ec400b58040c58c26 unmasking on ppc64le

it's a bit confusing, because ppc64 is a single keyword for both big-endian and little-endian ppc64, so we have to dance a bit.

I actually use crun with docker on ppc64le, works great ;-)

robertgzr:
@gyakovlev TIL thanks a lot o/

PR #17703, opened 2020-09-28
net-p2p/go-ipfs-bin: Bump to 0.7.0

kakra:


stkw0:
I don't know if you were aware of it, but there is a non bin version of this package. I think that we should remove this package as the non binary one is preferred.

kakra:
@stkw0 I noticed that but I figured it was easier to just bump the bin version instead of fiddling around with the huge module list of the source build, at least for now - it also spares me to install and build a lot of tooling and libs just to get this package. The version bump is needed now to be still be able to be an efficient node in IPFS as 0.7 changes some protocol stuff.

From that perspective, I'd say that this version bump should be merged, then focus on the build from source version and deprecating this package later (not bumping it to 0.8).

What do you think?

stkw0:
Last week I was working on go-ipfs-0.7.0 but I had a checksum mismatch on go-msgio which I didn't know how to solve (I opened an issue upstream). It will be easier and faster to merge this binary ebuild than to fix the source-based version, so I agree with you.

juippis (approved):

PR #17706, opened 2020-09-29
dev-lang/pipenv: update RDEPEND.

dave77:
pipenv requires the packages six and appdirs, added these to the ebuild.

Signed-off-by: Dave Wiltshire <davew@mykolab.com>

### DIFF
--- a/dev-python/pipenv/pipenv-2020.6.2-r1.ebuild
+++ b/dev-python/pipenv/pipenv-2020.6.2-r1.ebuild
@@ -1,0 +1,52 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DISTUTILS_USE_SETUPTOOLS=rdepend
+PYTHON_COMPAT=( python3_{6,7,8} )
+
+inherit distutils-r1
+
+MY_PV=${PV/_beta/b}
+DESCRIPTION="Python Development Workflow for Humans"
+HOMEPAGE="https://github.com/pypa/pipenv https://pypi.org/project/pipenv/"
+SRC_URI="https://github.com/pypa/pipenv/archive/v${MY_PV}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}"/${PN}-${MY_PV}
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="test"
+RESTRICT="!test? ( test )"
+
+RDEPEND="
+	dev-python/certifi[${PYTHON_USEDEP}]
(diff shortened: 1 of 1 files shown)


dev-python/pipenv/pipenv-2020.6.2-r1.ebuild
```diff
@@ -0,0 +1,52 @@
...
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="test"
+RESTRICT="!test? ( test )"
+
+RDEPEND="
+	dev-python/certifi[${PYTHON_USEDEP}]
+	>=dev-python/pip-18.0[${PYTHON_USEDEP}]
+	dev-python/virtualenv[${PYTHON_USEDEP}]
+	>=dev-python/virtualenv-clone-0.2.5[${PYTHON_USEDEP}]
+	dev-python/six
```
oz123, line 28: Why are you not using `[${PYTHON_USEDEP}]` here?

oz123: Please see my comments in [the issue ](https://bugs.gentoo.org/738132)regarding these dependencies.

dave77: Added PYTHON_USEDEP on the new dependencies as needed.

oz123: Thanks. I'm still not sure why these dependencies are added. See my comments.
Maybe someone else can back me or contradict me.

juippis: @dave77 can you point us to upstream dependency requirements, or show an error log if pipenv fails without these deps?
Ideally you'd first created a bug about it.

dave77: I wasn't sure if it was related to a different bug because there was a report that dev-lang/pipenv-2020.6.2 fails tests (https://bugs.gentoo.org/738132), so I didn't open a new bug when I couldn't run pipenv.

I've now created this bug describing the problem: https://bugs.gentoo.org/748138

When I install pipenv via pip (appdirs and six both get pulled in).  The stack traces show these packages not being found.

juippis:
b5a4470582c02d92f494de00bcd1c256528fbd9e does this fix this PR?

oz123:
@juippis IMHO yes. We also had a discussion in the bugzilla issue  https://bugs.gentoo.org/748138.

juippis:
Okay, what about https://bugs.gentoo.org/738132?

oz123:
That one too (for completion v 2020.8.13 should be removed, there is a new version https://github.com/pypa/pipenv/releases/tag/v2020.11.15 which I intend to create an ebuild for)

juippis:
Okay, thanks for clarifying.

PR #17715, opened 2020-09-29
dev-vcs/tig: restore tab completion

jsbronder:
The patch applied in the last bump is incorrect and effectively disables
bash completion no matter what order the completions were loaded in as
`type -f` will not find `__git_complete`.

```
$ type -f __git_complete
bash: type: __git_complete: not found
$ type __git_complete | grep 'is a'
__git_complete is a function
```

There is a pull-request upstream that properly addresses the issue of
`__git_complete` not being found but it's not accepted yet,
https://github.com/jonas/tig/pull/1016

Signed-off-by: Justin Bronder <jsbronder@cold-front.org>


gyakovlev:
agreed. let's better have small warning, but working completion.
we require git anyway as RDEPEND.

gyakovlev:
thanks!

PR #17718, opened 2020-09-30
dev-util/ostree: Version updated to 2020.6

jpds:
Tested on amd64 system.


dev-util/ostree/ostree-2020.6.ebuild
```diff
@@ -69,7 +69,6 @@ src_configure() {
 		$(use_with curl)
 		$(use_with dracut)
 		$(use_enable doc gtk-doc)
-		$(use_enable introspection)
```
zmedico: The gentoo dev manual states that "autodetection must be overridden":

https://devmanual.gentoo.org/general-concepts/use-flags/index.html

zmedico:
Merged in c21111531ad46fa9fc0c9e46bdc6de497c561b0b, without the introspection commit.  Thanks!

dev-util/ostree/ostree-2020.6.ebuild
zmedico: Updated in 563bd09fe946701aba0b43a11bdfad5059561d67.

PR #17720, opened 2020-09-30
dev-util/mingw64-runtime: Work with multilib gcc

maharmstone:

This adds a `multilib` USE option to mingw64-runtime, which adds `--enable-lib32` to the configure options when compiling for x86_64-w64-mingw32, which is needed for `x86_64-w64-mingw32-gcc -m32` to work (and thus make i686-w64-mingw32 pretty much redundant).

Without this, cross-x86_64-w64-mingw32/gcc will fail to compile when it has the `multilib` USE flag set - there should probably also be something in crossdev or something, to make sure it's not set for one package and not the other.


maharmstone:
I've closed this in favour of https://github.com/gentoo/gentoo/pull/17721/ - it's not clear beforehand that commits need to have a "Signed-off-by", and I'm not sure whether the bots will get confused by a force-push.

PR #17721, opened 2020-09-30
dev-util/mingw64-runtime: Work with multilib gcc

maharmstone:
This adds a `multilib` USE option to mingw64-runtime, which adds `--enable-lib32` to the configure options when compiling for x86_64-w64-mingw32, which is needed for `x86_64-w64-mingw32-gcc -m32` to work (and thus make i686-w64-mingw32 pretty much redundant).

Without this, cross-x86_64-w64-mingw32/gcc will fail to compile when it has the `multilib` USE flag set - there should probably also be something in crossdev or something, to make sure it's not set for one package and not the other.

Signed-off-by: Mark Harmstone <mark@harmstone.com>

### DIFF
--- a/dev-python/pytables/metadata.xml
+++ b/dev-python/pytables/metadata.xml
@@ -18,6 +18,7 @@
   for interactively save and retrieve large amounts of data.
   </longdescription>
   <upstream>
+    <remote-id type="github">PyTables/PyTables</remote-id>
     <remote-id type="pypi">tables</remote-id>
   </upstream>
 </pkgmetadata>


trofi:
> and thus make i686-w64-mingw32 pretty much redundant

That is probably not quite true. For USE=multilib to work one needs a multilib-capable profile to be able to build most ebuilds in multiple ABIs. I don't think we have such profiles for mingw targets in Gentoo. In my understanding crossdev does not generate multilib profiles either.

USE=multilib is also tricky to bootstrap as gcc normally requires libc binaries to be present before multilib gcc is built. crossdev is known to be broken for linux targets: https://bugs.gentoo.org/744016. Maybe mingw is not that badly affected as it does not build shared libraries for mingw by default and thus avoid circular dependency. Then we could pull your change it to ease future work. I suggest applying it on top of mingw64-runtime-8.0.0 ebuild.

maharmstone:
I've pushed a new version against mingw64-runtime 8.0.0.

Just tested it `crossdev`ing from scratch, and it works fine. The only change I had to make was to crossdev itself, so it wouldn't disable the multilib flag in `/etc/portage/profile/package.use.force/cross-x86_64-w64-mingw32`.

PR #17722, opened 2020-09-30
dev-latex/rubber:  add ebuild for version 1.6.0-alpha1

Flowdalic:


dev-tex/rubber/rubber-1.6.0_alpha1.ebuild
```diff
@@ -0,0 +1,87 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{6,7} )
```
thesamesam, line 6: Any chance of 8, 9?

dev-tex/rubber/rubber-1.6.0_alpha1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+	touch fig2dev-path-inplace/disable || die
+	touch fig2dev-path-into/disable || die
+	touch graphicx-dotted-files/disable || die
+	touch hooks-input-file/disable || die
+	touch knitr/disable || die
+
+	# dev-tex/metapost is hard masked.
+	touch metapost/disable || die
+
+	./run.sh *
```
thesamesam, line 86: || die on external commands

dev-tex/rubber/rubber-1.6.0_alpha1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+	touch fig2dev-dvi/disable || die
+	touch fig2dev-path/disable || die
+	touch fig2dev-path-inplace/disable || die
+	touch fig2dev-path-into/disable || die
+	touch graphicx-dotted-files/disable || die
+	touch hooks-input-file/disable || die
+	touch knitr/disable || die
+
+	# dev-tex/metapost is hard masked.
+	touch metapost/disable || die
```
thesamesam, line 84: I thought this was because it's now in texlive?

dev-tex/rubber/rubber-1.6.0_alpha1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+IUSE="test"
+RESTRICT="!test? ( test )"
+
+RDEPEND="virtual/latex-base"
+
+# Test dependencies:
+# - app-text/texlive-core for rubber's 'cweave' test
+# - dev-lang/R for rubber's 'knitr' test (requires knitr R library, currently disabled)
+# - dev-texlive/texlive-latexextra for rubber's 'combine' test (currently disabled)
+DEPEND="
```
thesamesam, line 36: BDEPEND (and we don't run tests in cross so it's fine to do it all in there)?

dev-tex/rubber/rubber-1.6.0_alpha1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+
+inherit distutils-r1
+
+if [[ ${PV} == "9999" ]] || [[ -n "${EGIT_COMMIT_ID}" ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://gitlab.com/latex-rubber/${PN}.git"
+else
+	UPSTREAM_PV=$(ver_rs 3 -)
+	S="${WORKDIR}/${PN}-${UPSTREAM_PV}"
+	SRC_URI="https://gitlab.com/latex-rubber/${PN}/-/archive/${UPSTREAM_PV}/${PN}-${UPSTREAM_PV}.tar.bz2"
```
thesamesam, line 17: We [usually](https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide/Style_Guide#Plaintexting_.24.7BPN.7D) plaintext ${PN} but it's not mandatory.

profiles/arch/powerpc/package.use.mask
```diff
@@ -1,6 +1,11 @@
 # Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Florian Schmaus <flo@geekplace.eu> (2020-09-30)
+# Rubber's testsuite depends on dev-tex/pythontex, which is not
+# keyworded on ppc.
+dev-tex/rubber test
```
thesamesam, line 7: We should probably do this in profiles/arch/powerpc/ppc32 instead (the structure is confusing and will be fixed soon). Looks ok, just remember to file a bug afterwards for keywording.

dev-tex/rubber/rubber-1.6.0_alpha1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+	test? (
+		app-text/texlive-core
+		dev-texlive/texlive-latexextra
+		media-gfx/asymptote
+		dev-tex/pythontex
+	)
+"
+
+python_install() {
+	insinto /usr/share/zsh/site-functions
```
thesamesam, line 48: I wonder if it makes sense to install this per-Python implementation. Just once would work, right?

dev-tex/rubber/metadata.xml
```diff
@@ -6,6 +6,6 @@
 		<name>Gentoo TeX Project</name>
 	</maintainer>
 	<upstream>
-		<remote-id type="launchpad">rubber</remote-id>
```
thesamesam, line 9: Are you [adding](https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide#Taking_over_an_existing_package) yourself?

dev-tex/rubber/rubber-1.6.0_alpha1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{6,7} )
+DISTUTILS_USE_SETUPTOOLS="no"
+
+inherit distutils-r1
+
+if [[ ${PV} == "9999" ]] || [[ -n "${EGIT_COMMIT_ID}" ]]; then
```
thesamesam, line 11: This logic is only worthwhile if we're keeping a live ebuild too (to allow them to both be kept in sync). But there isn't one right now.

dev-tex/rubber/rubber-1.6.0_alpha1.ebuild
Flowdalic: Good catch, it's now in `src_install`.

dev-tex/rubber/rubber-1.6.0_alpha1.ebuild
Flowdalic: For some reason the metapost tests don't on Gentoo (they succeed, like most other disabled tests here, on debian buster as part of rubber's CI). I have to look into the root cause, but had not time for this yet.

dev-tex/rubber/metadata.xml
Flowdalic: Right, done in https://github.com/gentoo/gentoo/pull/17722/commits/eedd9a95b066c1ab3da322feab6b72ed9521b40b

PR #17724, opened 2020-10-01
sci-electronics/fritzing: version bump to 0.9.4

rossbridger:
The newer fritzing versions rely on libgit2 to refresh the parts repository
directly from upstream. Thus I added a patch to allow users to clone the
parts repository to their home directory.

Bug: https://bugs.gentoo.org/633990


sci-electronics/fritzing/fritzing-0.9.4.ebuild
```diff
@@ -0,0 +1,99 @@
...
+}
+
+src_install() {
+	INSTALL_ROOT="${D}" default
+}
+
+pkg_postinst() {
+	einfo
+	einfo "Fritzing requires parts repository to function properly. To download, run"
+	einfo "\"git clone -b master https://github.com/fritzing/fritzing-parts.git\""
```
thesamesam, line 88: I'm a little bit lost here. Is there a reason we can't include this in the ebuild?

rossbridger: Technically we can, but the new fritzing feature introduced in this version uses libgit to update its parts repo directly from upstream, which requires the parts directory to be a git repo and modifiable by users. So it's a bit tricky indeed.

sci-electronics/fritzing/fritzing-0.9.4.ebuild
```diff
@@ -0,0 +1,94 @@
...
+
+inherit qmake-utils xdg
+
+MY_PV="CD-498"
+PARTS_P="${PN}-parts-${PV}"
+PARTS_COMMIT="e79a69765026f3fda8aab1b3e7a4952c28047a62"
+DESCRIPTION="Electronic Design Automation"
+HOMEPAGE="https://fritzing.org/"
+SRC_URI="https://github.com/fritzing/fritzing-app/archive/${MY_PV}.tar.gz -> ${P}.tar.gz
+	https://github.com/fritzing/fritzing-parts/archive/${PARTS_COMMIT}.zip -> ${PARTS_P}.zip"
```
juippis, line 14: Please make some newlines here to make this mess more readable.
```suggestion
MY_PV="CD-498"
PARTS_P="${PN}-parts-${PV}"
PARTS_COMMIT="e79a69765026f3fda8aab1b3e7a4952c28047a62"

DESCRIPTION="Electronic Design Automation"
HOMEPAGE="https://fritzing.org/"
SRC_URI="https://github.com/fritzing/fritzing-app/archive/${MY_PV}.tar.gz -> ${P}.tar.gz
	https://github.com/fritzing/fritzing-parts/archive/${PARTS_COMMIT}.zip -> ${PARTS_P}.zip"
```

sci-electronics/fritzing/fritzing-0.9.4.ebuild
```diff
@@ -0,0 +1,94 @@
...
+
+EAPI=7
+
+inherit qmake-utils xdg
+
+MY_PV="CD-498"
+PARTS_P="${PN}-parts-${PV}"
+PARTS_COMMIT="e79a69765026f3fda8aab1b3e7a4952c28047a62"
+DESCRIPTION="Electronic Design Automation"
+HOMEPAGE="https://fritzing.org/"
```
juippis, line 12: Please also add the Github page here as 2ndary HOMEPAGE.

sci-electronics/fritzing/fritzing-0.9.4.ebuild
```diff
@@ -0,0 +1,94 @@
...
+SRC_URI="https://github.com/fritzing/fritzing-app/archive/${MY_PV}.tar.gz -> ${P}.tar.gz
+	https://github.com/fritzing/fritzing-parts/archive/${PARTS_COMMIT}.zip -> ${PARTS_P}.zip"
+
+LICENSE="CC-BY-SA-3.0 GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+PLOCALES=( ar bg bn cs da de el en es et fa fr gl hi hu id it ja ko mk mr nl pl pt-BR pt-PT ro ru sk sl sr sv th tr uk ur vi zh-CN zh-TW )
+IUSE="${PLOCALES[@]/#/l10n_}"
+
+REQUIRED_USE="|| ( ${IUSE[@]} )"
```
juippis, line 22: Since these aren't fetched separately, you should just install them all. Helps you to clean and simplify the ebuild a lot.

sci-electronics/fritzing/fritzing-0.9.4.ebuild
```diff
@@ -0,0 +1,94 @@
...
+RDEPEND="dev-qt/qtconcurrent:5=
+	dev-qt/qtcore:5=
+	dev-qt/qtgui:5=
+	dev-qt/qtnetwork:5=
+	dev-qt/qtprintsupport:5=
+	dev-qt/qtserialport:5=
+	dev-qt/qtsql:5=[sqlite]
+	dev-qt/qtsvg:5=
+	dev-qt/qtwidgets:5=
+	dev-qt/qtxml:5=
```
juippis, line 33: I'm not a Qt guy, but my understanding is that Qt:5 has a stable ABI. Since `:5` is forced here, subslots shouldn't be needed at all.

sci-electronics/fritzing/fritzing-0.9.4.ebuild
```diff
@@ -0,0 +1,94 @@
...
+	>=dev-libs/quazip-0.7.2[qt5(+)]
```
juippis, line 34: You can lift off the version restriction, and the USE construct, since lowest version in the tree is 0.9.1

sci-electronics/fritzing/fritzing-0.9.4.ebuild
```diff
@@ -0,0 +1,94 @@
...
+	dev-libs/libgit2:=
+"
+DEPEND="
+	${RDEPEND}
+	>=dev-libs/boost-1.40
```
juippis, line 39: You can remove the version restriction.

sci-electronics/fritzing/fritzing-0.9.4.ebuild
```diff
@@ -0,0 +1,94 @@
...
+"
+
+S="${WORKDIR}/${PN}-app-${MY_PV}"
+
+DOCS="README.md"
```
juippis, line 44: This gets converted to an array, so just specify an array here.

sci-electronics/fritzing/fritzing-0.9.4.ebuild
```diff
@@ -0,0 +1,94 @@
...
+	local translations=
+
+	# fix build with newer quazip - bug #597988
+	sed -i -e "s/#include <quazip/&5/" src/utils/folderutils.cpp || die
+	sed -i -e "s|/usr/include/quazip|&5|" -e "s/-lquazip/&5/" phoenix.pro || die
+
+	# Get a rid of the bundled libs
+	# Bug 412555 and
+	# https://code.google.com/p/fritzing/issues/detail?id=1898
+	rm -rf src/lib/quazip/ pri/quazip.pri src/lib/boost* || die
```
juippis, line 61: We avoid using the `-f` option since it hides errors in the ebuild.

juippis (approved):
Looks good, final question: Does the changed parts directory need to be informed to users? Via `elog` or something? 

First time I launched this it informed me about missing parts with a dialog window, then I saw it had made a directory structure to my home folder, which your patch probably allows. But does the program work like it usually does?

rossbridger:
> Looks good, final question: Does the changed parts directory need to be informed to users? Via `elog` or something?
> 
> First time I launched this it informed me about missing parts with a dialog window, then I saw it had made a directory structure to my home folder, which your patch probably allows. But does the program work like it usually does?

I also saw a dialog informing it can't find git parts repository on the first startup, but it is normal because Fritzing expects the parts repository to be a git repo so it can update the repo with libgit. But other functionalities should still work as usual.

Fritzing always creates that directory for users to store their own circuit parts. My patch allows Fritzing to place a "parts.db" file on that directory instead of the parts repository. I can't find a way to create that file at compile time (which requires running the Fritzing executable but it is a GUI application) so I added the patch to allow user to generate it at runtime.

PR #17725, opened 2020-10-01
dev-db/sqldeveloper-20.2.0.175.1842

fedeliallalinea:
Bug: https://bugs.gentoo.org/632120


juippis (commented):
Could you check the CI issue and fix that? It should be simple.

dev-db/sqldeveloper/sqldeveloper-20.2.0.175.1842.ebuild
```diff
@@ -0,0 +1,102 @@
...
+
+DESCRIPTION="Oracle SQL Developer is a graphical tool for database development"
+HOMEPAGE="http://www.oracle.com/technetwork/developer-tools/sql-developer/overview/index.html"
+SRC_URI="${P}-no-jre.zip"
+
+RESTRICT="fetch"
+
+LICENSE="OTN"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
```
juippis, line 16: Sorry my brains didn't catch this the first time, but if a dependency doesn't exist, shouldn't we just remove the KEYWORD instead of mask? Is there any plan to keyword openjfx for x86 or is it hopeless? I think we can just drop the `~x86` here.

Not like anyone actually uses it anymore.

fedeliallalinea: I don't know if there is any plan for openjfx on x86, I can drop x86 support If you want. I am waiting for clarifications and thanks for review.

juippis (approved):

PR #17726, opened 2020-10-01
 games-simulation/openttd:

LubosKolouch:
New upstream version 1.10.3 ebuild


chewi:
I wasn't able to merge this because we need every commit to have the sign-off and pushing merge commits to fix this afterwards doesn't work. If you don't get it right first time then please amend or rebase the commits instead and then do a forced push.

The ebuild was exactly the same as the previous version anyway. In these cases, it's easier to just file a bug saying that so I don't have to check to see what's changed myself.

PR #17728, opened 2020-10-01
media-libs/openimageio: version bump to 2.2.6.1

epsilon-0:
new maintainer + gentoo science

---

patch Fixes an upstream bug not allowing system-pugixml


dev-cpp/robin-map/robin-map-9999.ebuild
```diff
@@ -0,0 +1,20 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake
+
+DESCRIPTION="c++ fast hash map and hash set using robin hood hashing"
```
thesamesam, line 8: Do we need a live ebuild? Also, C++ instead of c++?

media-libs/openimageio/files/openimageio-2.2.6.1-general.patch
```diff
@@ -0,0 +1,19 @@
+diff --git a/src/cmake/externalpackages.cmake b/src/cmake/externalpackages.cmake
```
thesamesam, line 1: Pop a link to the upstream bug above

media-libs/openimageio/openimageio-2.2.6.1.ebuild
```diff
@@ -0,0 +1,137 @@
...
+X86_CPU_FEATURES=(
+	aes:aes sse2:sse2 sse3:sse3 ssse3:ssse3 sse4_1:sse4.1 sse4_2:sse4.2
+	avx:avx avx2:avx2 avx512f:avx512f f16c:f16c
+)
+CPU_FEATURES=( ${X86_CPU_FEATURES[@]/#/cpu_flags_x86_} )
+
+IUSE="doc ffmpeg field3d gif jpeg2k opencv opengl openvdb ptex python qt5 raw +truetype ${CPU_FEATURES[@]%:*}"
+REQUIRED_USE="python? ( ${PYTHON_REQUIRED_USE} )"
+
+RESTRICT="test" # bug 431412
```
thesamesam, line 27: AFAIK this bug isn't the real reason anymore. It's because they're stored upstream in a separate repo, right?

media-libs/openimageio/openimageio-2.2.6.1.ebuild
```diff
@@ -0,0 +1,137 @@
...
+		opengl? ( dev-qt/qtopengl:5 )
+	)
+	raw? ( media-libs/libraw:= )
+	truetype? ( media-libs/freetype:2= )
+"
+DEPEND="${RDEPEND}"
+
+DOCS=( CHANGES.md CREDITS.md README.md )
+
+PATCHES=( "${FILESDIR}"/${P}-general.patch )
```
thesamesam, line 90: pop a version in too

media-libs/openimageio/openimageio-2.2.6.1.ebuild
```diff
@@ -0,0 +1,137 @@
...
+	python? (
+		${PYTHON_DEPS}
+		$(python_gen_cond_dep '
+			dev-libs/boost:=[python,${PYTHON_USEDEP}]
+			dev-python/numpy:=[${PYTHON_USEDEP}]
+			dev-python/pybind11:=[${PYTHON_USEDEP}]
+		')
+	)
+	qt5? (
+		dev-qt/qtcore:5
```
thesamesam, line 78: I thought Qt stuff was all subslotted and needed :=?

epsilon-0: oh, damn, didn't know that.
qtcore seemed to only be in slot 5.
I should have looked inside the ebuild :O

media-libs/openimageio/openimageio-2.2.6.1.ebuild
epsilon-0: true, will change comment.

dev-cpp/robin-map/robin-map-9999.ebuild
epsilon-0: Its a header only library, I don't think there are a lot problems that could happen with that.

media-libs/openimageio/openimageio-2.2.6.1.ebuild
epsilon-0: version?

media-libs/openimageio/openimageio-2.2.6.1.ebuild
DarthGandalf: I thought := should be used only if using private parts of Qt API.

thesamesam: You're right, @DarthGandalf!

dev-cpp/robin-map/robin-map-9999.ebuild
```diff
@@ -0,0 +1,20 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake
+
+DESCRIPTION="C++ fast hash map and hash set using robin hood hashing"
```
thesamesam, line 8: Do we need a live ebuild for this?

media-libs/openimageio/openimageio-2.2.6.1.ebuild
```diff
@@ -0,0 +1,139 @@
...
+	python? (
+		${PYTHON_DEPS}
+		$(python_gen_cond_dep '
+			dev-libs/boost:=[python,${PYTHON_USEDEP}]
+			dev-python/numpy:=[${PYTHON_USEDEP}]
+			dev-python/pybind11:=[${PYTHON_USEDEP}]
+		')
+	)
+	qt5? (
+		dev-qt/qtcore:5=
```
thesamesam, line 80: Sorry, I was wrong, drop :=

dev-cpp/robin-map/robin-map-0.6.3.ebuild
```diff
@@ -0,0 +1,20 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake
+
+DESCRIPTION="C++ fast hash map and hash set using robin hood hashing"
```
thesamesam, line 8: Does this run the tests btw? (Not tried yet).

epsilon-0: No, its tests are very badly written.
Its a different project, in the test directory, with a new project name.
I would need to patch the cmake so that instead of a new project its added as a subdirectory and then fix the compile flags and other things.

That's very high cost to benefit for a package which installs just two header files...

dev-cpp/robin-map/robin-map-9999.ebuild
epsilon-0: I think it is fine to have it but not necessary.

dev-cpp/robin-map/robin-map-0.6.3.ebuild
thesamesam: OK.

epsilon-0: I don't remember but I did try that, and it didnt work.
Lemme reconstruct the error.

epsilon-0: Hmm, it seems to be just ignoring it, not sure what its trying to do.
Here is what I'm trying - https://bsd.ac/5fn5gfh

Here is the link to testing directory - https://github.com/Tessil/robin-map/blob/master/tests/CMakeLists.txt
(I manually ran the tests and they worked)

I also feel like this has no benefit to testing.
Testing is mainly for checking if installation is working, which in this case is easy enough to see.
Any bug is going to be told to go upstream.
Unless someone wants to fix a bloom filter themselves...

media-libs/openimageio/openimageio-2.2.6.1.ebuild
epsilon-0: done

PR #17730, opened 2020-10-01
x11-libs/libva: bump to 2.9.0

Karlson2k:
Uptream has finally published the tarball.

PR #17733, opened 2020-10-01
Libva intel media driver 2.3.0 r1

Karlson2k:
Added USE flag to enable native host optimisation.


Karlson2k:
I can squash down commits and remove -r1 revision.

Karlson2k:
To be merged after #17730

rossbridger:
Why not just bump a single ebuild and close https://github.com/gentoo/gentoo/pull/17730?

Karlson2k:
> Why not just place everything in a single commit and close #17730?

As I mentioned in the [second comment](https://github.com/gentoo/gentoo/pull/17733#issuecomment-702177851), it could be combined into single commit and single version bump. No problem.
 
> Also what kind of trick does this native-optimisation USE have to offer? Shouldn't passing "-march=native" CFLAG suffice for such feature?
It doesn't work as build system add many flags including `-march=core-i7`. Without patching sources, the only way to override it is to define `UFO_MARCH` which supply argument for -march.
If it is preferable, I can patch cmake files to use user CFLAGS, but it will be untested by upstream.

rossbridger:
> > It doesn't work as build system add many flags including `-march=core-i7`. Without patching sources, the only way to override it is to define `UFO_MARCH` which supply argument for -march.
> > If it is preferable, I can patch cmake files to use user CFLAGS, but it will be untested by upstream.

I think it is preferable to make the build system to respect user's own cflags.

Karlson2k:
Superseded by https://github.com/gentoo/gentoo/pull/18955

PR #17752, opened 2020-10-02
dev-qt/qtwebengine: add use default for libvpx dependency

stha09:
Bug: https://bugs.gentoo.org/678730
Signed-off-by: Stephan Hartmann <sultan@gentoo.org>


stha09:
I don't want to revbump for this minor change, but can do on request.

a17r (changes requested):
Thanks. Will go middle ground here:
- add IUSE default in place for 5.15.1
- leave 5.14.2 unchanged
- start stabilisation of 5.15.1
- please don't drop latest ~arch libvpx w/ svc until after 5.14.2 was cleaned up (immediately after 5.15.1 stable)

PR #17754, opened 2020-10-02
dev-libs/npth: tidy up ebuild, remove unused versions

jsmolic:
- inherit autotools eclass
- unconditionally remove static libs
- drop redundant manual deletion of .la files


jsmolic:
Thanks @thesamesam

PR #17756, opened 2020-10-03
sys-fs/encfs: fix fuse slot, eapi 7, cmake-utils -> cmake

hlein:
encfs is not compatible with fuse-3.x, so we need to specify a slot
with which it is compatible. Also bumped to EAPI 7 while I was
touching it, and migrated from deprecated cmake-utils.eclass
to cmake.eclass.

Also, recent cmake-utils / cmake eclasses turn on shared libs, but
encfs' build system does the wrong thing for libencfs.so, so turn
that knob back off.

Closes: https://bugs.gentoo.org/701698


hlein:
ping @blueness thank you!

hlein:
ping @blueness as maintainer, and @robbat2 because you touched this PR a couple of months ago :-P

blueness:
pushed

PR #17766, opened 2020-10-04
Update kicad to 5.1.7 and drop some old versions

zpuskas:


sci-electronics/kicad/kicad-5.1.7.ebuild
```diff
@@ -131,6 +130,13 @@ src_install() {
 	fi
 }
 
+src_test() {
+	# Test cannot find library in Portage's sandbox. Let's create a link so test can run.
+	ln -s "${S}_build/eeschema/_eeschema.kiface" "${S}_build/qa/eeschema/_eeschema.kiface"
+
+	default
```
juippis, line 137: I doubt the test case did anything as it didn't print any output and was done in under a second, but at least it doesn't fail now. Could you investigate?

juippis (approved):

PR #17767, opened 2020-10-04
media-tv/v4l-dvb-saa716x: fix compile with kernel >=5.6

madmartin:

### DIFF
--- a/media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
+++ b/media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
@@ -1,0 +1,52 @@
+# Copyright 2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit linux-info linux-mod
+
+DESCRIPTION="driver for saa716x based dvb cards like TT S2-6400 or Skystar 2 eXpress HD"
+HOMEPAGE="https://bitbucket.org/powARman/v4l-dvb-saa716x"
+
+REVISION="83f3bfd93a95"
+REVISION_DATE="20160322"
+
+SRC_URI="https://bitbucket.org/powARman/v4l-dvb-saa716x/get/${REVISION}.tar.bz2
+-> v4l-dvb-saa716x-0.0.1_p${REVISION_DATE}.tar.bz2"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+DEPEND=""
+RDEPEND="${DEPEND}
+	sys-firmware/tt-s2-6400-firmware"
+
+S="${WORKDIR}/powARman-v4l-dvb-saa716x-${REVISION}"
+
+BUILD_TARGETS="modules"
+MODULE_NAMES="
+	saa716x_ff(misc:${EROOT}/usr/src/linux:${S}/linux/drivers/media/common/saa716x)
+	saa716x_core(misc:${EROOT}/usr/src/linux:${S}/linux/drivers/media/common/saa716x)
+	saa716x_budget(misc:${EROOT}/usr/src/linux:${S}/linux/drivers/media/common/saa716x)
+	saa716x_hybrid(misc:${EROOT}/usr/src/linux:${S}/linux/drivers/media/common/saa716x)"
+
+CONFIG_CHECK="DVB_CORE DVB_STV6110x DVB_STV090x"
+
+src_prepare() {
+	default
+
+	kernel_is ge 4 4 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-up-to-4.4.patch"
+	kernel_is ge 4 9 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-up-to-4.9.patch"
+	kernel_is ge 4 14 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-up-to-4.14.patch"
+	kernel_is ge 4 17 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-up-to-4.17.patch"
+	kernel_is ge 5 6 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-5.6-fix-compile.patch"
+}
+
+src_compile() {
+	kernel_is le 5 0 && BUILD_PARAMS="SUBDIRS" || BUILD_PARAMS="M"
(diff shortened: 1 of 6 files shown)


media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
```diff
@@ -0,0 +1,62 @@
...
+SRC_URI="https://bitbucket.org/powARman/v4l-dvb-saa716x/get/${REVISION}.tar.bz2
+-> v4l-dvb-saa716x-0.0.1_p${REVISION_DATE}.tar.bz2"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="+firmware"
+
+DEPEND=""
+RDEPEND="${DEPEND}
```
juippis, line 23: This DEPEND inherit is useless.

media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
```diff
@@ -0,0 +1,62 @@
...
+	firmware? ( sys-firmware/tt-s2-6400-firmware )"
```
juippis, line 24: Looks like this has no function in this ebuild, then please use `optfeature` instead.

media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
```diff
@@ -0,0 +1,62 @@
...
+	kernel_is ge 4 6 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-4.6.0-fix-compile.patch"
+	kernel_is ge 4 9 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-4.9-fix-warnings.patch"
+	kernel_is ge 4 10 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-4.10-fix-compile.patch"
+	kernel_is ge 4 14 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-4.14.0-fix-compile.patch"
+	kernel_is ge 4 15 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-4.15-fix-autorepeat.patch"
+	kernel_is ge 4 15 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-4.15-fix-timers.patch"
+	kernel_is ge 4 16 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-4.16-fix-compile.patch"
+	kernel_is ge 4 17 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-4.17-define-AUDIO_GET_PTS.patch"
+	kernel_is ge 5 6 0 && eapply "${FILESDIR}/v4l-dvb-saa716x-5.6-fix-compile.patch"
+}
```
juippis, line 54: Many of these versions have been cleaned out from the portage tree and aren't supported by upstream anymore either. Could you clean this list a bit?

media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
madmartin: yes, you're right...

media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
madmartin: Here I do not really agree. `optfeature` is not the right way, because the driver does not work on the hardware without the firmware. so IMHO "very specific circumstances, such as a package being nonfunctional unless at least one of a set of optional runtime dependencies is installed." applies here. Or may be not, because it is only one useflag, not a set of useflags.
Printing only a postinstall message can alse easily slipped, and the the correlation between driver package name and the firmware package is not easy visible. 
So maybe it would be the best to turn this to 'RDEPEND=sys-firmware/tt-s2-6400-firmware' without a useflag, if you insist of removing it.

media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
madmartin: What should I do - sum up the patches to reduce number of files or reduce the number of "kernel_is ge" expressions?
I see an advantage of having the patches separately, because they are an inspiration for adapting other drivers. At least I myself use this often.
While writing this, I have the idea to concatenate the patch files with some comments inside...

media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
juippis: Sure, depending on it unconditionally sounds better then.

media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
juippis: I didn't think this through, but the idea was that you could just apply the patches without kernel check in many cases, since it's expected the version requirement is fulfilled. But oh well, nevermind about this.

Why none of this can be found from upstream? Why must we carry around these ancient patches?

madmartin: > Why none of this can be found from upstream? Why must we carry around these ancient patches?
Good question, I dont know. This hardware is very rare, AFAIK only a few hundred have been produced + sold in the year 2010 or 2011. There is no active development, its functionally outdated (a raspberry pi can do the same job....)  but nevertheless it still works excellent (at my home and some other gentoo users). 
I will sum up patches to the kernel versions I find in gentoo-sources today, and try to send some pull requests upstream.

madmartin:
Hello @juippis the ebuild is cleaned up,  summed up the patches from 14 down to 4.
You may want to look again?!?

juippis (commented):
Oh well, can't test this but looks good. Let's wait for bug reports.

madmartin:
> Oh well, can't test this but looks good. Let's wait for bug reports.

I have tested compile on 3 different kernel versions, and its already on my "production" video server at home.
No severe surprises expected....

PR #17774, opened 2020-10-04
net-dns/openresolv: remove old unstable version

acerisara:
openresolv-3.11.0: current unstable
openresolv-3.10.0: stable
openresolv-3.9.2: old unstable
openresolv-3.9.0: old stable

Not sure why we have this unstable old version, I think it makes sense to remove it.


juippis (approved):
Yes, let's keep the portage tree clean by removing old redundant versions.

PR #17777, opened 2020-10-04
media-gfx/gimp: 2.10.22 version bump

band-a-prend:
There is no `media-libs/babl-0.1.{80,82}` version bump this time as they aren't required by gegl or gimp versions under review.


thesamesam:
When CI is red due to something like this, remember to rebase so we review. :)

media-gfx/gimp/gimp-2.10.22.ebuild
```diff
@@ -0,0 +1,211 @@
...
+inherit autotools gnome2 toolchain-funcs virtualx
+
+DESCRIPTION="GNU Image Manipulation Program"
+HOMEPAGE="https://www.gimp.org/"
+SRC_URI="mirror://gimp/v2.10/${P}.tar.bz2"
+LICENSE="GPL-3 LGPL-3"
+SLOT="2"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~ppc ~ppc64 ~x86"
+
+IUSE="aalib alsa aqua debug doc gnome heif jpeg2k mng openexr postscript udev unwind vector-icons webp wmf xpm cpu_flags_ppc_altivec cpu_flags_x86_mmx cpu_flags_x86_sse"
```
juippis, line 17: Doesn't this version of GIMP support libavif? We have that in Gentoo tree.

band-a-prend: Yes the news ablut Gimp states that AVIF support was added. And I missed this.

I asked one of Gimp developer and he said that avif support is via libheif. As I can see from libhief configuration.ac file the avif support switch on if libaom is found in system. But currently gentoo libheif package don't depend on libaom.

I.e. before fix that I need gentoo libheif package depends on libaom (maybe pernanent). In case of permanent dependence no changes in gimp ebuild is required.

juippis: **sigh** it also sounds like an automatic dep on libaom which needs to be patched out. So work ahead...

juippis: Ooor maybe a permanent dep isn't that bad.

juippis: Let's discuss what we can do in #17651 here.

PR #17786, opened 2020-10-05
dev-python/lmdb: version bump to 1.0.0, update metadata github id

wimmuskee:
also add python_test and py3_8 python compat

Closes: https://bugs.gentoo.org/744334


dev-python/lmdb/lmdb-1.0.0.ebuild
```diff
@@ -0,0 +1,30 @@
...
+
+RDEPEND="dev-db/lmdb:="
+DEPEND="${RDEPEND}"
+BDEPEND="test? ( dev-python/pytest[${PYTHON_USEDEP}] )"
+
+python_compile() {
+	LMDB_FORCE_SYSTEM=1 distutils-r1_python_compile
+}
+
+python_test() {
```
DarthGandalf, line 28: Replace this and BDEPEND with `distutils_enable_tests pytest`

DarthGandalf: And it adds the IUSE and RESTRICT too

sbraz: Tests aren't named `test_*` so pytest with its default options doesn't detect them. I will be adding `distutils_enable_tests pytest` on merge to remove the need for RESTRICT, IUSE and BDEPEND.

sbraz:
Thanks, pushed with a few changes.

PR #17790, opened 2020-10-05
media-gfx/inkscape: python3_9

vaukai:


thesamesam:
The version number of scour got broken: it should be scour-0.38.1.ebuild, right?

thesamesam:
Thanks!

PR #17795, opened 2020-10-05
sys-apps/i2c-tools: python3_9 and version bump to 4.2

vaukai:


Zlogene:
Version bump is incorrect, I have added a fixed version already but thanks in any way, Will test py3.9 a little bit later.

PR #17813, opened 2020-10-05
dev-lua/luaexpat: additional restore and masking

winterheart:
Additional restore functionality and masking until resolving issue with net-im/prosody.


ConiKost:
Do you see any reason to keep ```1.3.3-r1```? Otherwise you could drop it and won't need ```dev-lua/luaexpat: fix lua slot to :0``` any more?

winterheart:
Well, regarding with nonclear status of fork, I decided enhance 1.3.0 and maintain it for a while. 1.3.3-r2 version has not accepted yet functionality by upstream, so let's keep there revisions.

PR #17817, opened 2020-10-06
net-wireless/crda: EAPI=7, python3_9

vaukai:
Closes: https://bugs.gentoo.org/728666


dm0- (changes requested):
The `python_check_deps` function should not have been dropped.  Sorry, I must have been confused about something when I opened the bug, and all of my build environments succeed without it.  There are scenarios where the build will fail unless that function is defined.

vaukai:
> The `python_check_deps` function should not have been dropped. ...

Sorry for that.  (Done)

PR #17818, opened 2020-10-06
dev-qt/qtdeclarative: python3_9

vaukai:


rossbridger:
Can't help but notice that you have sent out many PRs for adding python 3.9 support, most of which only involve adding a '9' into PYTHON_COMPAT. Maybe it's better to combine all these into a single PR so people can test them in batches.

a17r (approved):
Thanks. I had to drop the first commit for not unnecessarily changing a stable package's `RDEPEND` (without revbump) and rebased adding py39 onto 5.15.2.

a17r:
Thanks for your work, pushed in commit 377bdf92e83d35d6cb3f0382c14ee4ee7da79e32.

@rossbridger combining commits makes sense when they touch packages of the same maintainer, otherwise it gets messy.

PR #17822, opened 2020-10-06
dev-lang/erlang: bump versions up to 23.0.4 and 23.1.1

arkamar:
- 23.0.4 is regular bug-fix release
- 23.1.1 seems to be mainly about changes that make erlang buildable in Haiku OS (which seems to be unrelated with Gentoo) and bug-fixes


trofi:
Pushed as is.

Thank you!

PR #17829, opened 2020-10-06
sys-cluster/glusterfs: cleanup old versions

jkroonza:

This should only be merged after at least amd64 and x86 has stabled:  https://bugs.gentoo.org/746956


ultrabug:
Thanks @jkroonza

PR #17830, opened 2020-10-06
x11-terms/terminator: version 2.0.1

DarthGandalf:


x11-terms/terminator/terminator-2.0.1.ebuild
```diff
@@ -0,0 +1,59 @@
...
+	dev-libs/glib:2
+	dev-python/configobj[${PYTHON_USEDEP}]
+	dev-python/psutil[${PYTHON_USEDEP}]
+	dev-python/pycairo[${PYTHON_USEDEP}]
+	dev-python/pygobject:3[${PYTHON_USEDEP}]
+	x11-libs/gtk+:3
+	x11-libs/vte:2.91[introspection]
+	dbus? ( dev-python/dbus-python[${PYTHON_USEDEP}] )
+	libnotify? ( x11-libs/libnotify[introspection] )
+	X? ( dev-libs/keybinder:3[introspection] )
```
juippis, line 28: Why are none of these in `optfeature`?

DarthGandalf: Good question. Moved X to optfeature, and the rest moved to non-conditional because tests are failing without these packages installed.

I think making it `test? ( … )` *and* adding to `optfeature` would be weird

juippis: It's perfectly fine and used commonly. Just make sure the package does work without those deps, ie why they are needed for test unit?
(most often the test unit just tests all cases so every dep is required)

DarthGandalf: Done.

Is this also fine then?

>  * Consider installing the following for additional features:
> *   dev-python/dbus-python[python_targets_python3_7(-)?,python_targets_python3_8(-)?,python_targets_python3_9(-)?,-python_single_target_python3_7(-),-python_single_target_python3_8(-),-python_single_target_python3_9(-)] for DBus

juippis: It's "correct" but I personally leave the `PYTHON_USEDEP` out from `optfeature` to prevent messages like this. Regular users would be baffled seeing that, and we expect regular users to use our provided profiles to control their active python interpreter. In other words, if you don't change your `PYTHON_TARGETS` via make.conf or package.use, just emerging terminator and dbus-python will get you a working combination.

DarthGandalf: Done

juippis (approved):

PR #17833, opened 2020-10-07
Bump janet 1.12.2

oz123:


dev-lang/janet/janet-1.12.2.ebuild
```diff
@@ -12,10 +12,12 @@ SRC_URI="https://github.com/janet-lang/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.g
 LICENSE="MIT"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE="static-libs"
+IUSE="+static-libs"
```
juippis, line 15: Why is this enabled by default? We're trying to get rid of `static-libs`, not encourage using it.

juippis (commented):
You should get accustomed to **first** doing version bump, **then** removing any older versions. If you sync the tree in a state between your commits, there is **no** janet in the tree.

Also whenever you have to use `repoman --force` you should really stop and think are you doing things correctly.

oz123:
> You should get accustomed to **first** doing version bump, **then** removing any older versions. If you sync the tree in a state between your commits, there is **no** janet in the tree.
> 

@juippis, duely noted: first bump, then remove.

> Also whenever you have to use `repoman --force` you should really stop and think are you doing things correctly.
I don't think I used it,  but I will note this too.

dev-lang/janet/janet-1.12.2.ebuild
oz123: Because without this `jpm` breaks. `jpm build` creates a static executable which can be deployed anywhere.
I can remove this and add `elog` message to notify the users that if this feature is desired they should manually enable it.
WDYT?

oz123: OK. static-libs is no longer the default.

oz123:
@juippis thanks again for all the reviews, I changed the order of commits as you suggested.
Also, I fixed the issues I had with jpm.

juippis (approved):
Looks good, I'll remove *optfeature.eclass* since it doesn't seem to be used. Thanks!

PR #17834, opened 2020-10-07
mail-filter/rspamd: bump to 2.6

arkamar:
New version of rspamd has been released recently. All patches from 9999 are reused as they are but I renamed them appropriately.


thesamesam:
Running tests would be great:
"No tests found. Skipping."

It's not a regression though, and LGTM!

PR #17836, opened 2020-10-07
sci-libs/lapack: and friends cleanup

epsilon-0:
Hunka chunka cleanup for a bunch of lapack and other math libraries


rossbridger:
While we are at it, do you think we should revamp virtual/lapacke as well? Because I was almost certain that the likes of openblas also provide a library implementing lapacke APIs.

epsilon-0:
@rossbridger 
unfortunately it does't :crying_cat_face: 
We don't have a lot of lapacke providers and in general there aren't many reimplementations of lapacke
https://wiki.gentoo.org/index.php?title=BLAS_and_LAPACK_Providers

rossbridger:
I think at least openblas has an openblas implementation according to its [cmake script](https://github.com/xianyi/OpenBLAS/blob/0ac610270809cb6dee8f5587784ceab8df356495/CMakeLists.txt#L386), but the nascent 0.3.11 version hasn't landed yet so I'm not sure the current in-tree versions have it too.

epsilon-0:
> I think at least openblas has an openblas implementation according to its [cmake script](https://github.com/xianyi/OpenBLAS/blob/0ac610270809cb6dee8f5587784ceab8df356495/CMakeLists.txt#L386), but the nascent version 0.3.11 hasn't landed yet so I'm not sure the current in-tree versions have it too.

Oh my! This is amazing news.   
Our openblas doesn't install this library in the ebuild so thats another thing to fix!
Thanks for bringing it to my attention. 

I'm not going to do it here as it is separate but I've opened a bug   
https://bugs.gentoo.org/749921
And we also have a tracker to do A LOT of blas lapack cleanup    
https://bugs.gentoo.org/747136

rossbridger:
I just sent out a [pull request](#17964 ) that added openblas as a conditional dependency on virtual/lapacke

SoapGentoo (approved):
We'll merge this once I sent @epsilon-0's last-rite (in a minute)

heroxbd:
> Ninja edit: I have added a few more potential libraries recently to ::science, as libflame is a lapack{,e} provider (https://github.com/gentoo/sci/tree/master/sci-libs/libflame-amd)

Why not just sci-libs/libflame?   It is not directly developed by amd but by researchers at University of Taxes.

epsilon-0:
> > Ninja edit: I have added a few more potential libraries recently to ::science, as libflame is a lapack{,e} provider (https://github.com/gentoo/sci/tree/master/sci-libs/libflame-amd)
> 
> Why not just sci-libs/libflame? It is not directly developed by AMD but by researchers at University of Taxes.

This is a form by AMD - https://github.com/amd/libflame
I named it in anticipation of also adding the original - https://github.com/flame/libflame

I've added all of these in a separate overlay for testing - https://github.com/bsd-ac/gentoo-blas-lapack

heroxbd:
> n

I suspect that the AMD repo could be catched in a patch on top of the original libflame.

PR #17837, opened 2020-10-07
sys-devel/llvm: build doc conditionaly on doc USE flag

montjoie:
The ebuild has a doc USE flag but still build documentation unconditionaly.
This change mades documentation build and install depending on doc USE flag.

Signed-off-by: Corentin Labbe <clabbe.montjoie@gmail.com>


mgorny:
Skipping manpages is not a valid solution.

PR #17841, opened 2020-10-07
net-im/teams: Version bump to 1.3.00.25560

jvesely:
Signed-off-by: Jan Vesely <jano.vesely@gmail.com>


jvesely:
This drops dependence on libgnome-keyring. In my case teams restarts, and everything works OK even without libgnome-keyring.

PR #17847, opened 2020-10-08
dev-cpp/string-theory: add for ja2-stracciatella-0.17.0

misanthropos:
String Theory is a flexible modern C++ library for string manipulation and storage


dev-cpp/string-theory/Manifest
```diff
@@ -0,0 +1,3 @@
+DIST string-theory-3.3.tar.gz 588360 BLAKE2B ad87857e6d5e3cad1346b759457b434de7612b97e1b19016b1dfc987ac67f0cab318745b14fe1f432cc921ffb1d33b875a2a62f0bc56d815dd35b405a9f08560 SHA512 86209333dce341078c3b973084bd9f3b8ff2ccac0e07a5e6acf5973bd1cfa420897b531b2d1bd6aba9f5ccc8927f85d91f06796ac0e62ec8a735564a0387d2f4
+EBUILD string-theory-3.3.ebuild 518 BLAKE2B d973809215ffc268c7217d7a5a6530d496a63ad6691e51669bcb2ac8f4b5dfc5301b92eebde850f88b6d87c15b24f5df1a9383e4deff889eaf2dc45db3dcdb51 SHA512 db306059624b3d6121a301caf639544ccf1ff6e5b71138e28b06d00124c8a50bee3d703918a5157f509230b52517f76638a472c79b5bc25e23bf98f2dee533a2
```
DarthGandalf, line 2: Remove this and metadata.xml from here. How did you create this file? `repoman manifest` should create it in a proper format.

It depends on which repo format is used. In git, manifests are thin.

dev-cpp/string-theory/string-theory-3.3.ebuild
```diff
@@ -0,0 +1,22 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+if [[ ${PV} = 9999 ]]; then
+	  GIT_ECLASS="git-r3"
+	  EXPERIMENTAL="true"
```
DarthGandalf, line 8: Seems unused?

dev-cpp/string-theory/string-theory-3.3.ebuild
```diff
@@ -0,0 +1,22 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+if [[ ${PV} = 9999 ]]; then
```
DarthGandalf, line 6: I don't see a 9999 ebuild here. But if you want it, SRC_URI should be here, because git version would not have it.

https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide/Style_Guide#Handling_multi_live.2Frelease_ebuilds

dev-cpp/string-theory/string-theory-3.3.ebuild
```diff
@@ -0,0 +1,22 @@
...
+DESCRIPTION="A flexible modern C++ library for string manipulation and storage"
+HOMEPAGE="https://github.com/zrax/string_theory/"
+
+SRC_URI="https://github.com/zrax/string_theory/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+S="${WORKDIR}/string_theory-${PV}"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
```
DarthGandalf, line 22: Did you test it on both architectures?

dev-cpp/string-theory/string-theory-3.3.ebuild
misanthropos: It was intended for a comming 9999 - I can remove it.

dev-cpp/string-theory/string-theory-3.3.ebuild
misanthropos: No - I remove x86 then?

dev-cpp/string-theory/Manifest
misanthropos: I made it with ebuild .. digest

dev-cpp/string-theory/string-theory-3.3.ebuild
DarthGandalf: Yes please. Or find a x86 to test on, via qemu etc

dev-cpp/string-theory/Manifest
DarthGandalf: You probably did it in another overlay, and then copied to here? ebuild digest also follows settings of the current overlay.

misanthropos:
merged with ja2

PR #17852, opened 2020-10-08
mail-filter/rspamd: use bundled hiredis, bug #747274

arkamar:
System hiredis makes the rspamd subprocesses segfault periodically. I have tested this behaviour with hiredis 0.13.3, 0.14.0 and 0.14.1. Therefore I am reverting my changes back to bundled hiredis. Sorry for inconvenience.


thesamesam:
Merged, thanks. Hopefully we can revisit unbundling in future once reported. I've revbumped to force a rebuild.

arkamar:
Thanks! I plan to contine on unbundling. I'll try to push some changes to upstream in this direction.

PR #17853, opened 2020-10-08
dev-libs/hiredis: make ssl support available

arkamar:
New version of hiredis adds ssl support. This PR makes this fature available through `ssl` useflag.


dev-libs/hiredis/hiredis-1.0.0.ebuild
```diff
@@ -10,12 +10,14 @@ HOMEPAGE="https://github.com/redis/hiredis"
 SRC_URI="https://github.com/redis/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
 
 LICENSE="BSD"
-SLOT="0/1.0.0"
+SLOT="0/$(ver_cut 1-2)"
```
juippis, line 13: Please don't do this unless you're **absolutely** sure about upstream breaking ABI every major version. It necessarily doesn't happen.

arkamar: Ok, thanks, I'll remove it.

juippis (approved):

PR #17863, opened 2020-10-09
app-misc/todo: clean up aliases

ferki:
This PR is a follow-up for #17270 to clean up Gentoo-specific `todocli` and `todotxt` aliases.

The bash completion for them was already broken, and they should be easy to re-add as part of the user's configuration. There are also no known reports about any breakage.

With those in mind, and based on a quick discussion with @monsieurp via IRC, we decided to simplify the ebuild by offering the vanilla experience from upstream.


monsieurp (approved):
As per our IRC discussion, let's go ahead with these changes. :+1:

PR #17864, opened 2020-10-09
net-wireless/hostapd: Cleaning up 2.9-r1

alarig:
Bug: https://bugs.gentoo.org/727542


net-wireless/hostapd/hostapd-2.9-r3.ebuild
```diff
@@ -99,9 +99,6 @@ src_configure() {
 	echo "CONFIG_EAP=y" >> ${CONFIG}
 	echo "CONFIG_ERP=y" >> ${CONFIG}
 	echo "CONFIG_EAP_MD5=y" >> ${CONFIG}
```
SoapGentoo, line 101: you're doing I/O with external commands, this always needs a `|| die`. Instead of `echo`ing, I suggest you use the `cat <` trick

SoapGentoo (changes requested):

net-wireless/hostapd/hostapd-2.9-r3.ebuild
alarig: The whole ebuild has `echo CONFIG_BLAH` without `|| die`. Should I do the modification everywhere then?

thesamesam: yes please, but I'll do this on merge

thesamesam:
Thank you!

PR #17872, opened 2020-10-10
licenses: add the CC-BY-NC-2.0 licence

misanthropos:
license for libsmacker which is coming with ja2-stracciatella-0.17.0


misanthropos:
included in ja2-stracciatella

PR #17874, opened 2020-10-10
dev-games/godot: New package

rossbridger:
[Godot](https://godotengine.org/) is an open-source game engine that has gained popularity in the recent years. This PR intends to land this package into the tree.

Closes: https://bugs.gentoo.org/532730


dev-games/godot/godot-3.2.3.ebuild
```diff
@@ -0,0 +1,136 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+PYTHON_COMPAT=( python3_{6..9} )
```
vilhelmgray, line 5: You can drop Python 3.6 from this list because Gentoo no longer supports it.

dev-games/godot/godot-3.2.3.ebuild
```diff
@@ -0,0 +1,136 @@
...
+inherit eutils python-any-r1 scons-utils flag-o-matic desktop xdg
+
+DESCRIPTION="Multi-platform 2D and 3D game engine"
+HOMEPAGE="http://godotengine.org"
+
+SRC_URI="https://github.com/godotengine/${PN}/archive/${PV}-stable.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/${P}-stable"
+KEYWORDS="~amd64 ~x86"
+
+LICENSE="MIT"
```
vilhelmgray, line 16: You should also include licenses for the [Godot logo](https://github.com/godotengine/godot/blob/master/LOGO_LICENSE.md) and the [third-party code](https://github.com/godotengine/godot/blob/master/thirdparty/README.md) used (assuming those are part of the installed files).

dev-games/godot/godot-3.2.3.ebuild
```diff
@@ -0,0 +1,136 @@
...
+		godot_binary="${godot_binary}.64"
+	elif [[ "${ARCH}" == "x86" ]]; then
+		godot_binary="${godot_binary}.32"
+	fi
+
+	newicon icon.svg ${PN}.svg
+	dobin bin/${godot_binary}
+	dodoc AUTHORS.md CHANGELOG.md DONORS.md README.md
+
+	make_desktop_entry ${godot_binary} Godot
```
vilhelmgray, line 135: I see a `dist` directory upstream: https://github.com/godotengine/godot/tree/master/misc/dist/linux

You might be able to use the [`xdg` eclass](https://devmanual.gentoo.org/eclass-reference/xdg.eclass/) to automatically take care of the installation for this desktop entry by calling `xdg_pkg_postinst` instead.

vilhelmgray (changes requested):

rossbridger:
That's a quick response, fixing them right away

vilhelmgray:
> That's a quick response, fixing them right away

I just got excited when I noticed a viable Godot ebuild PR opened up. ;-)

thesamesam:
We're not officially accepting new packages via proxy-maint, but developers can proxy commits personally.

Sell me on it? ;)

rossbridger:
> We're not officially accepting new packages via proxy-maint, but developers can proxy commits personally.
> 
> Sell me on it? ;)

I don't think there are any gentoo devs around here who actually know how to build godot, but regardless I'm keeping this PR open so that I can shape this ebuild into a functional form before sending it to guru or somewhere.

dev-games/godot/godot-3.2.3.ebuild
```diff
@@ -0,0 +1,136 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+PYTHON_COMPAT=( python3_{6..9} )
+
+inherit eutils python-any-r1 scons-utils flag-o-matic desktop xdg
```
thesamesam, line 7: Are you really using eutils?

thesamesam:
Well, all I really wanted was a reason for me to merge it because I'd have to personally keep on top of your PRs.

vilhelmgray:
I'd be willing to co-maintain this package with @rossbridger to reduce the risk of the ebuild becoming stagnant in the official Gentoo tree. I believe Godot is important enough of a package, given it's popularity and use in modern open source gaming.

dev-games/godot/godot-3.2.3.ebuild
vilhelmgray: The [Debian godot package](https://packages.debian.org/sid/godot3) offers a good starting point to locating all the licenses in Godot: https://metadata.ftp-master.debian.org/changelogs//main/g/godot/godot_3.2-stable-2_copyright

rossbridger:
I don't really have a real good reason right now, and if I can actually put a package into official gentoo tree I'd definitely choose to place steam, which offers a godot build in its store that renders this PR unnecessary. However godot does come with a easy-to-use visual editor, a modern realistic rendering system and many other state-of-the-art features that can knock over the existing game engines in the tree. So I'd say it does have a merit to be offered officially by gentoo.

dev-games/godot/godot-3.2.3.ebuild
rossbridger: python is not used by godot per se, but is a runtime dependency for scons, so it is just having what scons have.

dev-games/godot/godot-3.2.3.ebuild
rossbridger: this ebuild has disabled building most of the 3rd party libs and link with the ones provided by gentoo's tree instead, so it might not be necessary. But I will add the logo's license nontheless.

dev-games/godot/godot-3.2.3.ebuild
vilhelmgray: That's all right, but Python 3.6 is not longer provided by Gentoo so the ebuild should not have that version listed in the `PYTHON_COMPAT`. Instead, the line should be like this: `PYTHON_COMPAT=( python3_{7..9} )`

rossbridger: I suppose it won't harm if I remove it. But python 3.6 is still supported until 2021-12-23 according to the website.

dev-games/godot/godot-3.2.3.ebuild
vilhelmgray: If the code under the [`thirdparty`](https://github.com/godotengine/godot/tree/31d0f8ad8d5cf50a310ee7e8ada4dcdb4510690b/thirdparty) is not used, you should remove this directory in `src_prepare()` to ensure that only system libraries are used:

```
src_prepare() {
        # Remove bundled libaries
	rm -rf thirdparty || die
}
```

In addition, those third party libraries don't appear to be the only code with different licenses. For example, [`servers/physics/joints/cone_twist_joint_sw.cpp`](https://github.com/godotengine/godot/blob/31d0f8ad8d5cf50a310ee7e8ada4dcdb4510690b/servers/physics/joints/cone_twist_joint_sw.cpp) is licensed under **zlib** license too.

To be fair, I only see **zlib** as another license in the main Godot code in my quick glace, so adding just that and the icon license might suffice for this ebuild if the thirdparty directory is removed during `src_prepare()`.

rossbridger: I might have overlooked this one. It turns out assimp is still required as a built-in library.

dev-games/godot/godot-3.2.3.ebuild
vilhelmgray: You're right, I misspoke because it is still available in the tree -- rather, it looks like it's just [not receiving bug fixes anymore](https://www.gentoo.org/support/news-items/2020-04-22-python3-7.html); regardless, it'll probably be prudent not to have it in new ebuilds. Thanks! :+1:

dev-games/godot/godot-3.2.3.ebuild
vilhelmgray: That's all right, just remove the directories within `thirdparty` that aren't needed, and include the licenses for the remaining that are used.

rossbridger: There are too many of the directories in there and it certainly takes time to list all of the unneeded directories one by one.

vilhelmgray: Unfortunately, proper documentation of code licenses is important, so that's part of the plight of being an ebuild maintainer. Luckily, there might be a way to automate this a bit with some bash magic.

First remove the directories for the libraries you know aren't used because you're pulling in the system libraries. Then use [`find`](https://man7.org/linux/man-pages/man1/find.1.html) to recursively get all the `COPYRIGHT` and `LICENSE` files, and pass them to [`head`](https://man7.org/linux/man-pages/man1/head.1.html) to print out the first few lines. Finally you can just scroll down the lines printed to see all the license names of the code remaining. :-)

Chiitoo:
I've used Godot via Steam and local builds a little bit, and intend to do more of it in the future.  Would definitely prefer an ebuild over Steam, and will keep an eye on this in case I can help with it.

Thanks so far!

dev-games/godot/godot-3.2.3.ebuild
rossbridger: Alright, I have deleted unused third party libs and enumerated all the licenses for the remaining libs. It should be no longer a problem by now.

dev-games/godot/metadata.xml
```diff
@@ -0,0 +1,34 @@
...
+Games can be exported in one click to a number of platforms, including the major
+desktop platforms (Linux, macOS, Windows) as well as mobile (Android, iOS) and
+web-based (HTML5) platforms.
+	</longdescription>
+	<use>
+		<flag name="bullet">Enable support for Bullet (Physics Engine)</flag>
+		<flag name="enet">Enable support for net-libs/enet</flag>
+		<flag name="lto">Enable support for link time optimization</flag>
+		<flag name="freetype">Enable support for media-libs/freetype</flag>
+		<flag name="mbedtls">Enable support for net-libs/mbedtls</flag>
```
vilhelmgray, line 29: Use `<pkg>` tags for Gentoo packages mentioned in the flags. For example: `<flag name="mbedtls">Enable support for <pkg>net-libs/mbedtls</pkg></flag>`.

dev-games/godot/metadata.xml
```diff
@@ -0,0 +1,34 @@
...
+	</use>
+	<upstream>
+		<remote-id type="github">godotengine/godot</remote-id>
```
vilhelmgray, line 32: Add links for upstream bugs and documentation:

```
<bugs-to>https://github.com/godotengine/godot/issues</bugs-to>
<doc lang="en">https://docs.godotengine.org/en/stable/index.html</doc>
```

dev-games/godot/godot-3.2.3.ebuild
```diff
@@ -0,0 +1,148 @@
...
+	dodoc AUTHORS.md CHANGELOG.md DONORS.md README.md
+}
+
+pkg_postinst() {
+	xdg_pkg_postinst
+}
+
+pkg_postrm() {
+	xdg_pkg_postrm
+}
```
vilhelmgray, line 148: In retrospect, I don't believe `xdg_pkg_postinst` nor `xdg_pkg_postrm` are needed because they are called automatically by the default `pkg_postinst()` and `pkg_postrm()`. You can probably removes these lines then, sorry about the noise.

vilhelmgray (changes requested):

dev-games/godot/godot-3.2.3.ebuild
```diff
@@ -0,0 +1,148 @@
...
+		godot_binary="${godot_binary}.32"
+	fi
+
+	newbin bin/${godot_binary} ${PN}
+	newicon icon.svg ${PN}.svg
+	doman misc/dist/linux/${PN}.6
+	domenu misc/dist/linux/org.godotengine.Godot.desktop
+	insinto /usr/share/metainfo
+	doins misc/dist/linux/org.godotengine.Godot.appdata.xml
+	insinto /usr/share/mime/application
```
vilhelmgray, line 137: You should prefix the absolute installation paths with `${EPREFIX}`. So those two `insinto` should be:

```
insinto "${EPREFIX}/usr/share/metainfo"
...
insinto "${EPREFIX}/usr/share/metainfo"
```

vilhelmgray (changes requested):

dev-games/godot/godot-3.2.3.ebuild
rossbridger: Are you sure we need ${EPREFIX}? The examples in https://devmanual.gentoo.org/ebuild-writing/functions/src_install/index.html say otherwise.

dev-games/godot/godot-3.2.3.ebuild
rossbridger: I will test if they will be triggered automatically.

dev-games/godot/godot-3.2.3.ebuild
vilhelmgray: Hmm, that's a fair point. @thesamesam Is `${EPREFIX}` not necessary for `insinto` and friends?

dev-games/godot/godot-3.2.3.ebuild
vilhelmgray: Thanks, let me know when you find out because I'm recalling this behavior from memory.

rossbridger: It is automatically triggered.

rossbridger:
I guess this package is at least in a good shape for guru. One thing is left out through, I'm unable to tackle C#/Mono support, which seems to be one of the major features for Godot. Need someone familiar with Mono to hone it.

dev-games/godot/godot-3.2.3.ebuild
DarthGandalf: insinto handles EPREFIX by itself

dev-games/godot/godot-3.2.3.ebuild
```diff
@@ -0,0 +1,144 @@
...
+	upnp? ( net-libs/miniupnpc )
+	vorbis? ( >=media-libs/libvorbis-1.3.4 )
+	webp? ( media-libs/libwebp )
+"
+DEPEND="${RDEPEND}"
+BDEPEND="virtual/pkgconfig"
+
+src_prepare() {
+	default
+	rm -rf thirdparty/{bullet,enet,freetype,libogg,libpng,libtheora,libvorbis,libvpx,libwebp,mbedtls,miniupnpc,opus,pcre2,zlib,zstd} || die
```
DarthGandalf, line 68: Remove -f, otherwise it'll succeed even if didn't remove anything, so ||die would not help

dev-games/godot/godot-3.2.3.ebuild
```diff
@@ -0,0 +1,144 @@
...
+	>=sys-apps/tcp-wrappers-7.6.22-r1
+	>=sys-apps/util-linux-2.25.2-r2
+	>=sys-libs/gdbm-1.11
+	>=sys-libs/glibc-2.20-r2
+	>=sys-libs/libcap-2.22-r2
+	>=sys-libs/zlib-1.2.8-r1
+	>=x11-libs/libX11-1.6.2
+	>=x11-libs/libXcursor-1.1.14
+	>=x11-libs/libXinerama-1.1.3
+	media-libs/alsa-lib
```
DarthGandalf, line 42: You listed alsa-lib here twice

Also, please sort them

dev-games/godot/godot-3.2.3.ebuild
```diff
@@ -0,0 +1,144 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+PYTHON_COMPAT=( python3_{7..9} )
+
+inherit python-any-r1 scons-utils desktop xdg
```
DarthGandalf, line 7: Sort please

dev-games/godot/godot-3.2.3.ebuild
vilhelmgray: Thanks @DarthGandalf we can ignore my earlier request then since `insinto` already handles EPREFIX.

rossbridger:
Godot has patched its bundled minizip source to work with Gentoo: https://github.com/godotengine/godot/issues/10539. Probably need a counter patch to make godot link with system minizip lib instead and strip the bundled source.

vilhelmgray:
> Godot has patched its bundled minizip source to work with Gentoo: [godotengine/godot#10539](https://github.com/godotengine/godot/issues/10539). Probably need a counter patch to make godot link with system's minizip lib instead and strip the bundled source away.

Yes, using system-provided minizip library would be the preference for us. By the way, is there a reason why upstream Godot is basing their code on an old version of the zlib API?

vilhelmgray:
It's been over 3 years since that upstream Godot patch was merged. A [comment on the upstream issue page](https://github.com/godotengine/godot/issues/10539#issuecomment-323971937_) mentions that this patch was intended as a stopgap until the version 2.0 was released. I don't think this problem is present in more recent versions of minizip, so perhaps upstream Godot should be pinged to migrate to a newer minizip.

rossbridger:
> It's been over 3 years since that upstream Godot patch was merged. A [comment on the upstream issue page](https://github.com/godotengine/godot/issues/10539#issuecomment-323971937_) mentions that this patch was intended as a stopgap until the version 2.0 was released. I don't think this problem is present in more recent versions of minizip, so perhaps upstream Godot should be pinged to migrate to a newer minizip.

It turns out the bundled minizip source has some substantial API addition, so one can not simply removing it.

vilhelmgray:
> It turns out the bundled minizip source has some substantial API addition, so one can not simply remove it.

Very well, there's not much we can do at that point then. We'll have to treat it as simply part of Godot and patch it as you suggest to get it working on Gentoo.

rossbridger:
Since we have fixed all the known issues, @thesamesam what's your stance on whether putting this package into the main tree? Still waiting for a compelling reason?

rossbridger:
I've decided to uphold this PR and push the package into GURU instead. Will close this PR once it's done.

DarthGandalf:
I think it's fine to keep it open, even when it's in GURU... But up to you

rossbridger:
It's just my work on landing [vscodium](https://vscodium.com/) is about to come to fruition, and I intend to close out backlogs before sending in the big push.

PR #17883, opened 2020-10-11
games-fps/alienarena: fix version and build

DarthGandalf:


games-fps/alienarena/alienarena-7.66.ebuild
```diff
@@ -0,0 +1,78 @@
...
+PATCHES=(
+	"${FILESDIR}"/${P}-format.patch
+	"${FILESDIR}"/${P}-ar.patch
+)
+
+src_prepare() {
+	default
+	eautoreconf
+	# too many variables defined in .h to patch, https://bugs.gentoo.org/707814
+	append-cflags -fcommon
```
SoapGentoo, line 51: just checked, Debian does the same, so this is fine. Can you move it to before the `econf` call pls?

games-fps/alienarena/alienarena-7.66.ebuild
```diff
@@ -0,0 +1,78 @@
...
+	fi
+}
+
+pkg_postinst() {
+	xdg_icon_cache_update
+}
+
+pkg_postrm() {
+	xdg_icon_cache_update
+}
```
SoapGentoo, line 78: drop `xdg-utils` and just rely on `xdg` and remove these explicit definitions

SoapGentoo (changes requested):

PR #17886, opened 2020-10-11
games-arcade/jumpnbump-1.61 version bump

chusopr:
The development has been continued by a different upstream team, but Gentoo package has not been updated keeping a version that is 16 years old and is buggy and uses ancient libraries.

I don't know what to do with `x86` keyword. I don't have a system where I can test that keyword. Should it be just added to the new version without testing it? Or will we need to keep the old version as the provider for the `x86` keyword? Or will we remove support for this keyword from this package? I don't know what's Gentoo's overall plan for keeping `x86`.


mgorny:
We generally assume it's fine to keep old keywords when bumping. If it's broken, we'll know when arch teams try to stabilise it.

chusopr:
Hi @mgorny, thanks for your quick reply.

Then it must be safe to remove the older version.

I just force-pushed a new version of the PR with these changes.

games-arcade/jumpnbump/jumpnbump-1.61.ebuild
```diff
@@ -0,0 +1,46 @@
...
+		$(python_gen_cond_dep '
+			dev-python/pygobject[${PYTHON_USEDEP}]
+		')
+		x11-libs/gdk-pixbuf:2[introspection]
+		x11-libs/gtk+:3[introspection]
+	)
+"
+
+src_compile() {
+	emake PREFIX=/usr
```
DarthGandalf, line 36: Could you use ${EPREFIX}/usr?

chusopr: Fixed.

I also changed `${D}` to `${ED}`, let me know if that's wrong: https://github.com/gentoo/gentoo/pull/17886/commits/47f1a0ab74d03ade3582e51e9cfc207390447e75#diff-ba36f3cd0808c650dfa5029505e26ba1097f732a9613a3cc5e4d34626b498fa9R43

thesamesam:
Thank you! I've pushed this in 0b439abdec321d0a1a7c0b822f527b6308be0c55 but dropped the cleanup old commit because it's a fork to let users revert if there's a problem.

PR #17891, opened 2020-10-11
media-gfx/xfig: patch ghostscript issue

jsmolic:


jsmolic:
Upstream responded with a fix for the issue in https://sourceforge.net/p/mcj/tickets/109/#78cf

Will try to apply this patch tomorrow, marking it as draft until then

PR #17892, opened 2020-10-11
dev-python/greenstalk: add support for python 3.9

acerisara:


juippis:
Hi, same thing here. We don't really let people hoard maintainership without prior contributions to the package. greenstalk looks to be in a good state, but the ebuild is quite outdated and could possibly benefit from python-3.9 bump while updating to match our 2020 standards. 
https://dev.gentoo.org/~mgorny/python-guide/

PR #17893, opened 2020-10-11
dev-python/google-auth-oauthlib: add unit test support

acerisara:


juippis:
Hi, we don't really let people "hoard" package maintainership without prior contributions to said package. From what I can see, the state of this package is pretty good - it's not outdated, and it doesn't have bugs open. You could maybe look into getting python-3.9 support, and unit test support in the ebuild.

acerisara:
> Hi, we don't really let people "hoard" package maintainership without prior contributions to said package. From what I can see, the state of this package is pretty good - it's not outdated, and it doesn't have bugs open. You could maybe look into getting python-3.9 support, and unit test support in the ebuild.

Thanks for point this out @juippis. It wasn't clear to me what was the process to follow when I agreed in the ML to take care of these packages (I should have asked, really). I'll update this PR with what you suggested - worst case scenario I'll close this PR and open a new one when there'll be a new version to support.

acerisara:
Added unit test support and docs. I did some testing with python-3.9 and it seems to work, but one dependency (dev-python/requests-oauthlib) doesn't support python-3.9 yet so not sure if we should add it already.

juippis (approved):

PR #17657, opened 2020-09-25
sci-physics/bullet: version bump to 3.05

rossbridger:
Package-Manager: Portage-3.0.8, Repoman-3.0.1
Signed-off-by: Ross Charles Campbell <rossbridger.cc@gmail.com>

PR #17896, opened 2020-10-12
games-engines/openmw: restrict double-precisionness of bullet

DarthGandalf:
For 0.46.0, require no double precision.
For 9999, require double precision instead.

Closes: https://bugs.gentoo.org/747898

### DIFF
--- a/games-engines/openmw/files/openmw-0.47.0-mygui-license.patch
+++ b/games-engines/openmw/files/openmw-0.47.0-mygui-license.patch
@@ -4,13 +4,13 @@ diff --git a/CMakeLists.txt b/CMakeLists.txt
 index d69352c94..5f97c86e9 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -470,9 +470,6 @@ IF(NOT WIN32 AND NOT APPLE)
-     #    INSTALL(PROGRAMS "${OpenMW_BINARY_DIR}/Plugin_MyGUI_OpenMW_Resources.so" DESTINATION "${LIBDIR}" )
-     #ENDIF(BUILD_MYGUI_PLUGIN)
+@@ -856,9 +856,6 @@ elseif(NOT APPLE)
+             INSTALL(PROGRAMS "${INSTALL_SOURCE}/openmw-wizard" DESTINATION "${BINDIR}" )
+         ENDIF(BUILD_WIZARD)
  
--    # Install licenses
--    INSTALL(FILES "files/mygui/DejaVuFontLicense.txt" DESTINATION "${LICDIR}" )
+-        # Install licenses
+-        INSTALL(FILES "files/mygui/DejaVuFontLicense.txt" DESTINATION "${LICDIR}" )
 -
-     # Install icon and desktop file
-     INSTALL(FILES "${OpenMW_BINARY_DIR}/org.openmw.launcher.desktop" DESTINATION "${DATAROOTDIR}/applications" COMPONENT "openmw")
-     INSTALL(FILES "${OpenMW_SOURCE_DIR}/files/launcher/images/openmw.png" DESTINATION "${ICONDIR}" COMPONENT "openmw")
+         # Install icon and desktop file
+         INSTALL(FILES "${OpenMW_BINARY_DIR}/org.openmw.launcher.desktop" DESTINATION "${DATAROOTDIR}/applications" COMPONENT "openmw")
+         INSTALL(FILES "${OpenMW_SOURCE_DIR}/files/launcher/images/openmw.png" DESTINATION "${ICONDIR}" COMPONENT "openmw")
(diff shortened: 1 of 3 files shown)


rossbridger:
For starters I think we should at least put https://github.com/OpenMW/openmw/pull/2049 and https://github.com/OpenMW/openmw/pull/2848 in the commit message as references.

games-engines/openmw/files/openmw-0.47.0-mygui-license.patch
```diff
@@ -4,13 +4,13 @@ diff --git a/CMakeLists.txt b/CMakeLists.txt
 index d69352c94..5f97c86e9 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -470,9 +470,6 @@ IF(NOT WIN32 AND NOT APPLE)
-     #    INSTALL(PROGRAMS "${OpenMW_BINARY_DIR}/Plugin_MyGUI_OpenMW_Resources.so" DESTINATION "${LIBDIR}" )
-     #ENDIF(BUILD_MYGUI_PLUGIN)
+@@ -856,9 +856,6 @@ elseif(NOT APPLE)
+             INSTALL(PROGRAMS "${INSTALL_SOURCE}/openmw-wizard" DESTINATION "${BINDIR}" )
+         ENDIF(BUILD_WIZARD)
```
rossbridger, line 9: I'm not familiar with openmw, but is this patch required to fix the bug we have in hand?

DarthGandalf:
Good point. I thought these references in the linked bug are enough. Done.

games-engines/openmw/openmw-9999.ebuild
```diff
@@ -96,6 +96,7 @@ src_configure() {
 		-DMORROWIND_DATA_FILES="${EPREFIX}/usr/share/morrowind-data"
 		-DUSE_SYSTEM_TINYXML=ON
 		-DDESIRED_QT_VERSION=5
+		-DBULLET_USE_DOUBLES=ON
 	)
```
rossbridger, line 100: Think we can add a USE flag to flip this variable? I'm not sure if it might result in a subtle behavior change.

rossbridger:
Previously @juippis suggested to remove double-precision USE flag at the [version bump](https://github.com/gentoo/gentoo/pull/17657), but that was before we discovered that openmw is still affected by it. And it might not be a good idea to block the new bullet version because it doesn't introduce any API breakage. Do you think we can re-add the double-precision flag back to bullet-3.05?

games-engines/openmw/files/openmw-0.47.0-mygui-license.patch
DarthGandalf: This patch is for 9999 version, and currently it doesn't build with the old version of the patch. It's not related to the bug.

games-engines/openmw/openmw-9999.ebuild
DarthGandalf: It will result in behavior change, yes. https://github.com/OpenMW/openmw/pull/2049 has a long discussion about it. Yes, we can move the choice of "buggy" vs "correct" to the user, though I don't see much benefit in such USE flag on openmw side.

games-engines/openmw/files/openmw-0.47.0-mygui-license.patch
DarthGandalf: I've moved it to a separate commit

juippis:
@DarthGandalf as the maintainer, what do you think? What'd be easier for you? I can enable the USE flag for bullet no problem there.

DarthGandalf:
@juippis I think, with this PR, if bullet gains the USE flag back, it'll automatically support old openmw too. Also, old bullet built with double precision would break openmw 0.46

I think bullet should get the USE flag back if any of the following is true:
* if there are concerns about having the old bullet version in tree, new bullet should get it, so that existing openmw release can depend on new bullet, and not block removal of old one
* if other programs which use bullet also got broken because of this

Other than that, I don't mind either way

games-engines/openmw/openmw-9999.ebuild
DarthGandalf: I just thought about the case where the USE flag can be useful: if some other program depends on bullet, but requires single precision instead, then user can't install both that program and openmw together, if there is no such USE flag in openmw.

I don't know how important it is though, and whether there are such programs.

rossbridger: I don't think I have the time enumerating all packages that depends on bullet and examine them one by one, but since float-point computing is very dependent on the precision bits, I'd suggest you test the 9999 version with different bullet precisions and make sure it doesn't end up dropping FPS rate dramatically or having other performance impacts.

juippis (approved):

PR #17899, opened 2020-10-12
media-libs/opencv: version bump to 4.5.0

rossbridger:
Starting this arduous work again only days after we put 4.4.0 into the tree...

- Testing if the following bugs will be fixed in this version:
Bug: https://bugs.gentoo.org/747949
Bug: https://bugs.gentoo.org/747163
Bug: https://bugs.gentoo.org/704518

- Should close the following obsolete bugs:
Closes: https://bugs.gentoo.org/666716


thesamesam:
I strongly suspect we need a < dep on eigen for [now](https://bugs.gentoo.org/747949#c4).

rossbridger:
Wait till I have checked compiling with different USE flags. It's possible we might end up with another month to get this version mainlined.

rossbridger:
Confirmed v4.5.0 is still affected by [bug 747949](https://bugs.gentoo.org/747949)

media-libs/opencv/opencv-4.5.0.ebuild
```diff
@@ -0,0 +1,553 @@
...
+SRC_URI="https://github.com/${PN}/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz
+	dnnsamples? ( https://dev.gentoo.org/~amynka/snap/${PN}-3.4.0-res10_300x300-caffeemodel.tar.gz )
+	download? ( https://github.com/rossbridger/opencv-extdep/archive/4.4.0.tar.gz -> ${PN}-4.4.0_extdep.tar.gz )
+	contrib? (
+		https://github.com/${PN}/${PN}_contrib/archive/${PV}.tar.gz -> ${P}_contrib.tar.gz
+		contribdnn? ( https://dev.gentoo.org/~amynka/snap/${PN}-3.4.0-face_landmark_model.tar.gz )
+		contribxfeatures2d? ( https://dev.gentoo.org/~amynka/snap/vgg_boostdesc-3.2.0.tar.gz )
+	)"
+
+LICENSE="BSD"
```
thesamesam, line 22: Did you read the [changelog](https://github.com/opencv/opencv/wiki/ChangeLog#version440)? ;)

media-libs/opencv/opencv-4.5.0.ebuild
```diff
@@ -0,0 +1,553 @@
...
+PYTHON_COMPAT=( python3_{7,8,9} )
+CMAKE_ECLASS=cmake
+inherit java-pkg-opt-2 java-ant-2 cmake-multilib python-r1 toolchain-funcs
+
+DESCRIPTION="A collection of algorithms and sample code for various computer vision problems"
+HOMEPAGE="https://opencv.org"
+TINY_DNN_PV="1.0.0a3"
+SRC_URI="https://github.com/${PN}/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz
+	dnnsamples? ( https://dev.gentoo.org/~amynka/snap/${PN}-3.4.0-res10_300x300-caffeemodel.tar.gz )
+	download? ( https://github.com/rossbridger/opencv-extdep/archive/4.4.0.tar.gz -> ${PN}-4.4.0_extdep.tar.gz )
```
thesamesam, line 15: Does this need a refresh?

media-libs/opencv/opencv-4.5.0.ebuild
```diff
@@ -0,0 +1,553 @@
...
+	contrib? (
+		https://github.com/${PN}/${PN}_contrib/archive/${PV}.tar.gz -> ${P}_contrib.tar.gz
+		contribdnn? ( https://dev.gentoo.org/~amynka/snap/${PN}-3.4.0-face_landmark_model.tar.gz )
```
thesamesam, line 18: I wonder if these are up to date now

thesamesam:
There's no reason to wait if trying an earlier Eigen value fixes it though.

media-libs/opencv/opencv-4.5.0.ebuild
rossbridger: The changelog was released after this PR, so no.

media-libs/opencv/opencv-4.5.0.ebuild
rossbridger: Still testing

thesamesam:
Let me know when ready for testing on my end.

media-libs/opencv/opencv-4.5.0.ebuild
```diff
@@ -0,0 +1,553 @@
...
+	mycmakeargs+=(
+		# python_setup alters PATH and sets this as wrapper
+		# to the correct interpreter we are building for
+		-DPYTHON_DEFAULT_EXECUTABLE=${EPYTHON}
+		-DINSTALL_PYTHON_EXAMPLES=$(usex examples)
+	)
+
+	# Regenerate cache file. Can't use rebuild_cache as it won't
+	# have the Gentoo specific options.
+	rm -rf CMakeCache.txt || die "rm failed"
```
DarthGandalf, line 532: Remove -f, otherwise it will never return an error, which makes ||die useless
And you don't need -r when removing a simple file

media-libs/opencv/opencv-4.5.0.ebuild
```diff
@@ -0,0 +1,553 @@
...
+	cmake_src_configure
+	cmake_src_compile
+	cmake_src_install
+
+	# Remove compiled binary so new version compiles
+	# Avoid conflicts with new module builds as build system doesn't
+	# really support it.
+	rm -rf modules/python2 || die "rm failed"
```
DarthGandalf, line 540: same

rossbridger:
> Let me know when ready for testing on my end. Have you figured out a workaround for that AVX multilib bug?

Unless someone familiar with these SIMD intrinsics comes up with a patch, we might still need upstream to fix it. The only thing we can do is to prevent opencv from building the avx2-related code for x86 by either adding an explicit USE constraint or silently removing the avx2 flag from cmake arguments.

thesamesam:
We'll just drop the avx2 flag when doing x86.

thesamesam:
Ready for testing now? I'll remove the Eigen dep on merge as this has been fixed in the Eigen package now: 2a69391373153bcb32284be0a12d38493b24bddd.

rossbridger:
It should be fine by now.

media-libs/opencv/opencv-4.5.0.ebuild
```diff
@@ -0,0 +1,555 @@
...
+	# have the Gentoo specific options.
+	rm CMakeCache.txt || die "rm failed"
+	cmake_src_configure
+	cmake_src_compile
+	cmake_src_install
+
+	# Remove compiled binary so new version compiles
+	# Avoid conflicts with new module builds as build system doesn't
+	# really support it.
+	rm -r modules/python2 || die "rm failed"
```
mackal, line 542: This needs to remove the correct python version. Currently this new ebuild doesn't even support python2 so this should never work ... We could just have this switched to 3 and it will just work, but that's only because we no longer support 2 with this ebuild.

thesamesam: My unhappiness is that this seems to have not been tested. Mistakes happen and that's okay, but it needs to be fixed.

mackal: It does work as expected removing the modules/python3 folder. No idea if we want to make this less fragile though ... (like if we want to worry about someone adding back python2 support and not have to deal with this line)

rossbridger: Alright I've made some test, and you're right about this

PR #17900, opened 2020-10-12
media-libs/libvpx: drop ppc

stha09:
Signed-off-by: Stephan Hartmann <sultan@gentoo.org>


gyakovlev:
just curious, what's the context here? =)

stha09:
> just curious, what's the context here? =)

Wanted to see how many consumers are left before I look into fixing the test failures. Now I need to setup qemu chroot or find some hardware to reproduce ;)

gyakovlev:
pinged you on IRC, we have quite beefy ppc capable hardware with testing chroots pre-setup, can make you a login there.

PR #17908, opened 2020-10-12
sci-mathematics/normaliz: version bump to 3.8.9

epsilon-0:


sci-mathematics/normaliz/normaliz-3.8.9.ebuild
```diff
@@ -0,0 +1,62 @@
...
+
+# would be nice to package scip and cocoalib
+
+RDEPEND="
+	dev-libs/gmp:=[cxx]
+"
+DEPEND="${RDEPEND}
+	dev-libs/boost
+"
+# Only a boost header is needed -> not RDEPEND
```
DarthGandalf, line 27: Should it use := to use the new header when new boost is installed?

epsilon-0: I'm not even sure but you are probably right.
Just let this sit for a few days, my brain is fried.
Dumb package broke backwards compatibility so I'm not even sure if we want to add a new subslot or not...

sci-mathematics/normaliz/normaliz-3.8.9.ebuild
```diff
@@ -0,0 +1,62 @@
...
+MYP="Normaliz-${PV}"
+
+DESCRIPTION="Tool for computations in affine monoids and more"
+HOMEPAGE="http://www.mathematik.uni-osnabrueck.de/normaliz/"
+SRC_URI="https://github.com/Normaliz/Normaliz/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3"
+SLOT="0/3"
+KEYWORDS="~amd64 ~arm ~x86 ~amd64-linux ~x86-linux"
+IUSE="doc extras openmp static-libs"
```
SoapGentoo, line 17: remove `static-libs` pls

sci-mathematics/normaliz/normaliz-3.8.9.ebuild
SoapGentoo: @DarthGandalf from my quick glance, it only uses C++ headers, and doesn't link, so this should be fine

sci-mathematics/normaliz/normaliz-3.8.9.ebuild
```diff
@@ -0,0 +1,62 @@
...
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	econf \
+		$(use_enable openmp) \
+		$(use_enable static-libs static)
```
SoapGentoo, line 43: `--disable-static`

SoapGentoo (changes requested):

sci-mathematics/normaliz/normaliz-3.8.9.ebuild
epsilon-0: done

sci-mathematics/normaliz/normaliz-3.8.9.ebuild
epsilon-0: done

PR #17916, opened 2020-10-13
app-cdr/iat: fix metadata

oxr463:
Signed-off-by: Lucas Ramage <ramage.lucas@protonmail.com>

Also, while I'm here, would it even make sense to bump EAPI to 7 or does it matter?


juippis (approved):
It wouldn't make sense IMHO. You'd have to reset keywords, wait 30 days, stabilize and drop old. EAPI-6 is still perfectly valid to be used.

If you need to do anything else for the ebuild, then do bump EAPI.

PR #17917, opened 2020-10-13
dev-python/pymilter: Address distutils QA warning

rseichter:
Explicitly set `DISTUTILS_USE_SETUPTOOLS=no` to address QA warning.


Whissi:
No need to do that via rev bump. I'll fix on merge.

rseichter:
@Whissi I deliberately did not bump the revision. It is only a QA warning, so why force a rebuild?

rseichter:
Silly me, I actually forgot to push the change after having realised that a revbump is not needed (which is what I started with).

PR #17922, opened 2020-10-14
Lxd verify sig

juippis:
Another test.


juippis:
This is on hold for a bit because they have multiple people working on these and I haven't been able to find all PGP keys from their publishers yet.

PR #17928, opened 2020-10-14
dev-util/codeblocks: fix gcc-11 compatibility;

band-a-prend:
The GCC 11 now enforces that comparison objects be invocable as const.

Current patch to fix gcc-11 compatibility is proposed by
Sergei Trofimovich <slyfox@gentoo.org>
and fix codeblocks "openfileslist" pligin compilation error.

Patch was sent to upstream :
https://sourceforge.net/p/codeblocks/tickets/1022/
so live-ebuild currently doesn't apply it.

Closes: https://bugs.gentoo.org/732818

Signed-off-by: Sergey Torokhov <torokhov-s-a@yandex.ru>

Additionally use xdg.eclass in 17.12-r304 instead of xdg-utils.eclass. Drop some old revisions.


trofi:
Pushed as is.

Thank you!

PR #17929, opened 2020-10-14
Firejail updates, add proxy maintainer

hlein:
Proxy-maint both firejail and firejail-lts; applied several updates and fixes addressing outstanding bugs in the former.


thesamesam:
@ajakk

sys-apps/firejail/files/firejail-0.9.62.4-nvidia-no-nogroups.patch
```diff
@@ -0,0 +1,19 @@
+diff --git a/src/firejail/profile.c b/src/firejail/profile.c
```
ajakk, line 1: Where did this patch come from? Please add references.

sys-apps/firejail/firejail-0.9.62.4.ebuild
```diff
@@ -0,0 +1,85 @@
...
+	EGIT_REPO_URI="https://github.com/netblue30/firejail.git"
+	EGIT_BRANCH="master"
+fi
+
+DESCRIPTION="Security sandbox for any type of processes"
+HOMEPAGE="https://firejail.wordpress.com/"
+
+LICENSE="GPL-2"
+SLOT="0"
+IUSE="apparmor +chroot contrib debug +file-transfer +globalcfg +network +overlayfs +private-home +seccomp +suid test +userns vim-syntax +whitelist x11"
```
ajakk, line 22: Can `debug` be dropped? I can't find it used anywhere.

sys-apps/firejail/firejail-0.9.62.4.ebuild
```diff
@@ -0,0 +1,85 @@
...
+
+RDEPEND="apparmor? ( sys-libs/libapparmor )"
+
+DEPEND="${RDEPEND}
+	!sys-apps/firejail-lts
+	test? ( dev-tcltk/expect )"
+
+PATCHES=(
+	"${FILESDIR}"/"${P}"-nvidia-no-nogroups.patch
+	"${FILESDIR}"/"${P}"-apparmor-global.patch
```
ajakk, line 33: Quote the whole strings? `"${FILESDIR}/${P}-..."`

ajakk (changes requested):

sys-apps/firejail/firejail-0.9.62.4.ebuild
hlein: Hm, sure looks like it. I'll do some verification, then update with `debug` removed.

sys-apps/firejail/firejail-0.9.62.4.ebuild
hlein: Will do.

sys-apps/firejail/files/firejail-0.9.62.4-nvidia-no-nogroups.patch
hlein: Ah, sure, see:
https://github.com/netblue30/firejail/issues/3644
https://github.com/netblue30/firejail/commit/623e6821687b848d6bb6f78a9deb682746df24ce

I will update the .patch file to include those in comments.

sys-apps/firejail/firejail-0.9.62.4.ebuild
```diff
@@ -0,0 +1,85 @@
...
+)
+
+src_prepare() {
+	default
+
+	find ./contrib -type f -name '*.py' | xargs sed --in-place 's-#!/usr/bin/python3-#!/usr/bin/env python3-g' || die
```
ajakk, line 39: Instead of the ebuild trying to handle Python shebangs and the build system installing them can these be installed with a `python_*script` function from `python-utils-r1.eclass`?

https://devmanual.gentoo.org/eclass-reference/python-utils-r1.eclass/index.html

sys-apps/firejail/firejail-0.9.62.4.ebuild
hlein: Confirmed I think; I can't find anything in the current release that cares about debug (no configure --enable-debug for instance). IUSE=debug was introduced in the firejail-0.9.60 ebuild but seems to serve no purpose; will strip it out.

sys-apps/firejail/firejail-0.9.62.4.ebuild
hlein: Wow, so almost nobody does it the way I did, but there's a lot of both:

`"${FILESDIR}"/${P}-foo.patch`
and:
`"${FILESDIR}/${P}-foo.patch"`

...with the former preferred 2:1, if my quick test is accurate (which it might not be):
```
portage-tree $ egrep -h -o -r '[^    ]+FILESDIR.*patch"?' [a-z]* | sed -E 's/[^"]+/_/g' | \
                        sort | uniq -c | sort -rn
  11055 "_"_
   6056 "_"
     94 _"_"
     82 "_"_"_"_
     74 _"_"_
     46 "_"_"_"
     30 _
     16 _"_"_"_"_
     12 "_
     10 "_"_"_"_"_"_
      3 _"_
      3 "_"_"_
      1 _"_"_"_"
      1 _"
      1 "_"_"_"_"_"
```

I'll go with whole-string as you suggested.

sys-apps/firejail/firejail-0.9.62.4.ebuild
hlein: Huh, actually, I could use help figuring out the right thing to do here. The `find | xargs sed` is not the right thing, but...

The only python scripts in firejail are in `contrib/`, and only installed if `USE=contrib` (so perhaps any mangling should be `if use contrib`'ed). All but one of those scripts already use `#!/usr/bin/env python3` as their shebang. Just because they should probably all be similar, I issued a PR upstream to change that one to be like all the others (https://github.com/netblue30/firejail/pull/3672).

But... that seems to be the opposite of what `python-utils.r1.eclass` functions want: `python_doscript`, `python_newscript`, `python_fix_shebang` all say "Only files having a 'python*' shebang will be modified", so they probably would not act on `#!/usr/bin/env python3`?

So should I in fact invert the current `find | xargs`, have it convert *from* env *to* a python shebang, and then call a `python_*script` function?

hlein: OK, @ajakk I've updated the PR to address everything except this bit. Thanks for the review, please lmk about this part.

ajakk: It looks to me like you can drop that `find` command in src_prepare, disable the `contrib` configure options so the build system doesn't install it, then install those conditionally with `python_doscript`.

hlein: OK, did mostly that. I found and worked around a few issues:

- ended up using `python-single-r1` so that EPYTHON would be set for me
- `python_doscript` would not install without conforming shebangs
- `python_fix_shebang` would not fix upstream's shebangs without -f
- `python_fix_shebang` choked on one contrib/*.py that was all `^M`'ed, so had to clean that up first
- the contrib/ scripts aren't safe to install in `/usr/bin/` - generic names that could collide. So I used `python_scriptinto` to install to the path that the `contrib-install` configure option used (plus, now they will be where upstream expects them for support purposes).

So, provided the way I handled those are acceptable, everything should now be resolved I think.

ajakk:
Firejail 0.9.64 is released. Can you update to it here?

hlein:
> Firejail 0.9.64 is released. Can you update to it here?

Ah, excellent, yup released yesterday. Sure, will update locally and test a bit, and then update this PR.

hlein:
> > Firejail 0.9.64 is released. Can you update to it here?
> 
> Ah, excellent, yup released yesterday. Sure, will update locally and test a bit, and then update this PR.

Done; required some tweaks due to changes in the new version, but also let me drop the interim patches.

sys-apps/firejail/firejail-0.9.64.ebuild
```diff
@@ -0,0 +1,95 @@
...
+	!sys-apps/firejail-lts
+	sys-libs/libseccomp
+	test? ( dev-tcltk/expect )"
+
+REQUIRED_USE="contrib? ( ${PYTHON_REQUIRED_USE} )"
+
+src_prepare() {
+	default
+
+	find -type f -name Makefile.in | xargs sed -i -r \
```
juippis, line 41: Could you use `find -exec` instead of `find | xargs`, since now in theory this would need double `|| die`.

sys-apps/firejail/firejail-0.9.64.ebuild
```diff
@@ -0,0 +1,95 @@
...
+			-e '/^\tinstall .*COPYING /d' \
+			-e '/CFLAGS/s: (-O2|-ggdb) : :g' || die
+
+	sed -i -r -e '/CFLAGS/s: (-O2|-ggdb) : :g' ./src/common.mk.in || die
+
+	# remove compression of man pages
+	sed -i -e '/gzip -9n $$man; \\/d' Makefile.in || die
+	sed -i -e '/rm -f $$man.gz; \\/d' Makefile.in || die
+	sed -i -r -e 's|\*\.([[:digit:]])\) install -m 0644 \$\$man\.gz|\*\.\1\) install -m 0644 \$\$man|g' Makefile.in || die
```
juippis, line 50: You can combine these modifications to a single `sed` call, since they edit the same file. I'd also prefer this as a patch, since it does so much. This solution should also be upstreamed.

sys-apps/firejail/firejail-0.9.64.ebuild
```diff
@@ -0,0 +1,95 @@
...
+src_install() {
+	default
+
+	if use vim-syntax; then
+		insinto /usr/share/vim/vimfiles/ftdetect
+		doins contrib/vim/ftdetect/firejail.vim
+
+		insinto /usr/share/vim/vimfiles/syntax
+		doins contrib/vim/syntax/firejail.vim
+	fi
```
juippis, line 87: I believe you can just install these unconditionally.

hlein:
@juippis I think I've addressed all the things, please take another look, thanks!

sys-apps/firejail/firejail-0.9.64.ebuild
hlein: Done; I copied `find ... -exec ... {} + || die` syntax I found elsewhere in the tree.

sys-apps/firejail/firejail-0.9.64.ebuild
hlein: Combined.

I can create an upstream issue/PR. Is there some standard "precompressed manpages considered harmful" reference you know of that I can cite? This naive grep finds 68 instances of packages un-compressing existing manpages:

```
portage-tree $ find . -name \*.ebuild | xargs egrep '([bg]unzip|[bg]zip2? -d).*(man|\.[1-8])' | wc -l
68
```

sys-apps/firejail/firejail-0.9.64.ebuild
hlein: Hah, as it turns out, upstream's current `make install` target already deploys these unconditionally. I suspect the `if use vim-syntax` is a holdover from a time when it didn't. I've removed it and that USE flag entirely.

sys-apps/firejail/firejail-0.9.64.ebuild
```diff
@@ -0,0 +1,83 @@
...
+SLOT="0"
+IUSE="apparmor +chroot contrib +dbusproxy +file-transfer +globalcfg +network +overlayfs +private-home +suid test +userns +whitelist x11"
+RESTRICT="!test? ( test )"
+
+RDEPEND="apparmor? ( sys-libs/libapparmor )
+	contrib? ( ${PYTHON_DEPS} )
+	dbusproxy? ( sys-apps/xdg-dbus-proxy )"
+
+DEPEND="${RDEPEND}
+	!sys-apps/firejail-lts
```
juippis, line 32: FYI this blocker needs to be in RDEPEND or it won't work properly.

juippis (approved):

sys-apps/firejail/firejail-0.9.64.ebuild
```diff
@@ -0,0 +1,83 @@
...
+	EGIT_REPO_URI="https://github.com/netblue30/firejail.git"
+	EGIT_BRANCH="master"
+fi
+
+DESCRIPTION="Security sandbox for any type of processes"
+HOMEPAGE="https://firejail.wordpress.com/"
+
+LICENSE="GPL-2"
+SLOT="0"
+IUSE="apparmor +chroot contrib +dbusproxy +file-transfer +globalcfg +network +overlayfs +private-home +suid test +userns +whitelist x11"
```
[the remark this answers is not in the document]
hlein: Boo, github never notified me of your re-check. IIRC it tells firejail to turn on some x11-specific support / profiles / whatnot.

Agreed `X` is probably more appropriate; since `x11` is their configure-arg name, I think it would need to be `use_enable X x11`?

sys-apps/firejail/firejail-0.9.64.ebuild
hlein: > FYI this blocker needs to be in RDEPEND or it won't work properly.

Huh, OK, it seemed to behave right (I can't install firejail-lts while firejail is installed, and vice versa) but I will take your word for it.

firejail-lts specifies the reciprocal block the same way as this, but it hasn't had a release so I haven't made any changes to it. If it's necessary, I can add an -r bump just to fix its RDEPEND; if not, I will just get that next time there is a new -lts.

sys-apps/firejail/firejail-0.9.64.ebuild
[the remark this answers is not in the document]
juippis: Yes, that's correct and should work.

sys-apps/firejail/firejail-0.9.64.ebuild
juippis: I'm pretty sure I hit this issue and that's why I noticed it.
https://devmanual.gentoo.org/general-concepts/dependencies/ check "Blockers" and especially the big red warning note in there.

juippis (approved):

sys-apps/firejail/firejail-0.9.64.ebuild
hlein: Ah, gotcha, thanks!

PR #17621, opened 2020-09-20
net-im/discord-bin: Add 'appindicator' USE

Meelo:
Add USE flag 'appindicator' to net-im/discord-bin

Closes: https://bugs.gentoo.org/743808

PR #17931, opened 2020-10-14
net-im/discord-bin: add optfeature

Meelo:
Add optfeature 'appindicator' to net-im/discord-bin

Closes: https://bugs.gentoo.org/743808
Signed-off-by: Niko Hämäläinen <milo@meelo.org>
Package-Manager: Portage-3.0.4, Repoman-3.0.1f

### DIFF
--- a/net-im/discord-bin/discord-bin-0.0.12-r1.ebuild
+++ b/net-im/discord-bin/discord-bin-0.0.12-r1.ebuild
@@ -6,7 +6,7 @@ EAPI=7
 MY_PN=${PN/-bin/}
 MY_BIN="D${MY_PN/d/}"
 
-inherit desktop linux-info pax-utils unpacker xdg
+inherit desktop linux-info optfeature pax-utils unpacker xdg
 
 DESCRIPTION="All-in-one voice and text chat for gamers"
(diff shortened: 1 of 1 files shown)


Meelo:
https://github.com/gentoo/gentoo/pull/17621 - this is the other alternative, but I was educated that runtime deps should not be controlled via USE-flags, but with optfeature instead. That being said, other electron-apps (such as Slack) that have appindicator support are using USE-flag instead of optfeature.

rossbridger:
> #17621 - this is the other alternative, but I was educated that runtime deps should not be controlled via USE-flags, but with optfeature instead. That being said, other electron-apps (such as Slack) that have appindicator support are using USE-flag instead of optfeature.

Who actually educated you? What users enable the USE flag, they'd expect portage pulls in the dependencies automatically, not just putting a message telling them to install the necessary package manually.

Meelo:
This wasn't a hactoberfest merge request though. I was genuinly trying to get this change accepted to the project. This method was adviced to me by someone who had done some of the previous pull request accepts for discord-bin.

But since this is not getting any attraction, I guess I'll just close this and keep the knowledge of discord-bin relying on libappindicator to enable systray icon.

PR #17933, opened 2020-10-15
media-libs/libavif: 0.8.2 version bump

novomesk:
New version with disabled -Werror flag as requested in https://bugs.gentoo.org/744616


juippis (commented):
This version breaks `dev-libs/efl`, currently its only revdep in the tree.

Now I'm thinking if libavif is the faulty version here, or if efl needs to update its avif API calls. Can you think of some other program to test this with?

novomesk:
That's expected. libavif changed API. libavif is known by frequent breaking changes.
I had to adjust my code in qt-avif-image-plugin and gimp-avif-plugin too. darktable uses libavif too, they are not ready for new API yet.

novomesk:
dev-libs/efl can use `~media-libs/libavif-0.8.1` dependency

juippis:
Yes, but regarding future... adding `~media-libs/libavif-0.8.1` to efl now requires a revbump -> rebuild. How do we deal with breakage once more packages start depending on *libavif*? Should we always pinpoint it to a working version?

novomesk:
I started to use libavif in January 2020 and there were several breaking changes this year. Sometimes, there is big change every release.

I think for now we should use specific versions as dependency and maybe we should start adding subslots. Later, the dependencies could be on subslots too (there could be multiple releases with compatible API).

PR #17937, opened 2020-10-15
sci-libs/scikits_learn: version bump to 0.23.2

epsilon-0:

---

Does a lot of things
 - Remove dependency on `sci-libs/scikits`
 - Move to proper names as scikits itself is not a package (the old package is actually scikits.example and is incorrectly named)
 - Remove old deprecated packages and add their new counterparts

Buncha cruncha cleanup and closing bugs


rossbridger:
Any specific reason why you changed the package's category?

epsilon-0:
It was previously in `dev-python`, but was moved to `sci-libs` as it was thought to have a dependency on `sci-libs/scikits`, which is not present as scikit-learn is an independent project from scikits for the last 5 years.
Also the fact that it is a python module, it makes sense to keep it in `dev-python`.

dev-python/scikit-optimize/scikit-optimize-0.8.1.ebuild
```diff
@@ -0,0 +1,26 @@
...
+PYTHON_COMPAT=( python3_{6..9} )
+inherit distutils-r1
+
+DESCRIPTION="Sequential model-based optimization library"
+HOMEPAGE="https://scikit-optimize.github.io/"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64 ~ppc64 ~x86"
```
heroxbd, line 15: Is this intended?  Have you tested the ebuild on ppc64 and arm64?

dev-python/pyaml/pyaml-20.4.0.ebuild
```diff
@@ -0,0 +1,24 @@
...
+PYTHON_COMPAT=( python3_{6..9} )
+inherit distutils-r1
+
+DESCRIPTION="pretty and readable YAML-serialized data generator"
+HOMEPAGE="https://github.com/mk-fg/pretty-yaml"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64 ~ppc64 ~x86"
```
heroxbd, line 15: same here.

dev-python/scikit-optimize/scikit-optimize-0.8.1.ebuild
epsilon-0: Oops, this is a leftover from the previous ebuild.
Sorry about this.
fixed.

dev-python/pyaml/pyaml-20.4.0.ebuild
epsilon-0: fixed

heroxbd:
@chutz So, finally we are going to move the package back to dev-python/scikit-learn!

PR #17938, opened 2020-10-15
net-misc/seafile-client: fix Qt5.15 build

CaseOf:
Hello, this fixes build with Qt5.15.

### DIFF
--- a/net-misc/seafile-client/seafile-client-7.0.9-r1.ebuild
+++ b/net-misc/seafile-client/seafile-client-7.0.9-r1.ebuild
@@ -18,23 +18,24 @@ RESTRICT="!test? ( test )"
 RDEPEND="dev-db/sqlite:3
 	dev-libs/libevent
 	dev-libs/jansson
-	<dev-qt/qtcore-5.15:5
-	<dev-qt/qtdbus-5.15:5
-	<dev-qt/qtgui-5.15:5
-	<dev-qt/qtnetwork-5.15:5
-	<dev-qt/qtwidgets-5.15:5
+	dev-qt/qtcore:5
+	dev-qt/qtdbus:5
+	dev-qt/qtgui:5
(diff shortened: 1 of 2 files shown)


net-misc/seafile-client/seafile-client-7.0.9-r1.ebuild
```diff
@@ -18,23 +18,24 @@ RESTRICT="!test? ( test )"
 RDEPEND="dev-db/sqlite:3
 	dev-libs/libevent
 	dev-libs/jansson
-	<dev-qt/qtcore-5.15:5
-	<dev-qt/qtdbus-5.15:5
-	<dev-qt/qtgui-5.15:5
-	<dev-qt/qtnetwork-5.15:5
-	<dev-qt/qtwidgets-5.15:5
```
a17r, line 25: Please in the future refrain from introducing such version restrictions in the first place. If build with a newer Qt version fails, then that's a bummer, but it ought to be fixed. One lookup in Gentoo's bugzilla would have given you a hint on how to produce a patch anyway.

Doing that completely destroys the ability of users of seafile-client to upgrade their systems.

juippis: That's on me. I noticed it didn't build with newer Qt when testing the contribution, but that it worked with 5.14. Sorry I wasn't aware of an existing patch, not a Qt guy ;)

juippis (approved):
Thanks!

net-misc/seafile-client/seafile-client-7.0.9-r1.ebuild
a17r: @juippis not a ready made patch, but a tracker with instructions for a standard missing header fix. Of course it would help if the tinderbox would produce meaningful summaries containing the error message in the first place.

PR #17940, opened 2020-10-15
games-roguelike/dwarf-fortress: add version 0.47.04, bugfix

sveneusewig:
The bug has already been confirmed but has not yet been fixed.
The changes for the bug fix were posted by the user Cheaterman
https://www.bay12games.com/dwarves/mantisbt/view.php?id=11564

Signed-off-by: Sven Eusewig <sveneusewig@yahoo.de>
Bug: https://bugs.gentoo.org/729002


Cheaterman:
Thank you :-) :+1:

sveneusewig:
@Cheaterman , @chewi 
Who else needs to know?

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.04.ebuild
```diff
@@ -43,6 +43,9 @@ RESTRICT="strip"
 src_prepare() {
 	rm -f libs/*.so* || die
```
DarthGandalf, line 44: Please remove `-f`

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.04.ebuild
```diff
@@ -1,12 +1,12 @@
 # Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
```
DarthGandalf, line 4: In EAPI 7 `BDEPEND` should be split from `DEPEND`

DarthGandalf (commented):
This also closes https://bugs.gentoo.org/717752 so please add it to commit message

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.04.ebuild
sveneusewig: I add this: `BDEPEND="virtual/pkgconfig"` and remove `"virtual/pkgconfig"` from `DEPEND`.

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.04.ebuild
sveneusewig: Ok.

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.04.ebuild
```diff
@@ -31,8 +31,8 @@ RDEPEND="media-libs/glew:0
 DEPEND="${RDEPEND}
 	media-libs/libsndfile
 	media-libs/openal
-	sys-libs/ncurses-compat:5[unicode]
-	virtual/pkgconfig"
+	sys-libs/ncurses-compat:5[unicode]"
```
chewi, line 34: I know you haven't changed this but `ncurses-compat` in `DEPEND` doesn't make any sense. This package does not install any headers. I can see that the sources are hardcoded to load `libncurses.so.5`. This is really bad if you're building with the headers from `libncurses.so.6`. I can see that it falls back to `libncurses.so` but that's no use if it finds `.so.5` first. The fact that it builds means it must be compatible with `.so.6` so I guess you should patch `renderer_curses.cpp` accordingly. This whole thing is really horrible though.

chewi (changes requested):
This isn't new but I really don't like how the launcher for this game just copies the entire thing into `HOME`. Is this really the only way it can work? I don't have the energy to argue with upstream about it.

chewi:
> This isn't new but I really don't like how the launcher for this game just copies the entire thing into `HOME`. Is this really the only way it can work? I don't have the energy to argue with upstream about it.

Ah sorry, I now realise this launcher script is our own. :fearful: Presumably it was done like this for a reason though. Please find out whether it's still necessary.

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.04.ebuild
```diff
@@ -41,8 +41,11 @@ QA_PREBUILT="${gamesdir#/}/libs/Dwarf_Fortress"
 RESTRICT="strip"
 
 src_prepare() {
-	rm -f libs/*.so* || die
+	rm libs/*.so* || die
 	sed -i -e '1i#include <cmath>' g_src/ttf_manager.cpp || die
+
+	eapply "${FILESDIR}/${P}-segfault-fix-729002.patch"
```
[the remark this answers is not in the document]
sveneusewig: Ok. I remove this there.

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.04.ebuild
sveneusewig: Let's see what I can find. In the worst case it looks like this:
https://i.imgur.com/Kx4NvI4.gif

frayment:
You may want to update this to 0.47.05 now that it's out. I've tested your patch in the new version with a custom libgraphic build and it still works.

sveneusewig:
I'm making a new pull request.

PR #17950, opened 2020-10-16
sys-fs/zfs-kmod: 0.8.5: Marked compatiable with 5.9 kernels.

jpds:
This version of ZFS is compatible with 5.9 kernels. I installed this and rebuilt the module against it and it boots a rootfs.


gyakovlev:
thanks!
looks like upstream also updated release notes.

PR #17956, opened 2020-10-17
sci-libs/opencascade: fix check in cmake config

waebbl:
Fix a check for library path in the OpenCASCADEConfig.cmake
file.


juippis (approved):
Can't test this because vtk won't compile, but at least the patch applies.

PR #17961, opened 2020-10-17
games-rpg/adonthell: don't call ar directly

DarthGandalf:
Closes: https://bugs.gentoo.org/746095


chewi:
I was going to say send this upstream but it looks somewhat dead.

PR #17964, opened 2020-10-18
media-libs/opencv: various provisional fixes

rossbridger:
- Explicitly depends on virtual/lapacke to prevent its cmake build scripts from
  implicitly disabling lapack support when it can't find lapacke.h.

 - Links with cblas when sci-libs/lapack is supplimented as the default lapack
  implementation.

 - examples USE requires contribdnn USE to be enabled to prevent build failure.

Closes: https://bugs.gentoo.org/700176
Closes: https://bugs.gentoo.org/749681

### DIFF
--- a/virtual/lapacke/lapacke-3.8.ebuild
+++ b/virtual/lapacke/lapacke-3.8.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
@@ -9,6 +9,9 @@ KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~ppc ~ppc64 ~s390 ~sparc ~x86 ~a
 IUSE="eselect-ldso"
 
 RDEPEND="
-	>=sci-libs/lapack-3.8.0[lapacke,eselect-ldso?]
+	|| (
+		>=sci-libs/lapack-3.8.0[lapacke,eselect-ldso?]
+		>=sci-libs/openblas-0.3.10[eselect-ldso?]
+	)
 "
 DEPEND="${RDEPEND}"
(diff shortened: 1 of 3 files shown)


virtual/lapacke/lapacke-3.8.ebuild
```diff
@@ -9,6 +9,9 @@ KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~ppc ~ppc64 ~s390 ~sparc ~x86 ~a
 IUSE="eselect-ldso"
 
 RDEPEND="
-	>=sci-libs/lapack-3.8.0[lapacke,eselect-ldso?]
+	|| (
+		>=sci-libs/lapack-3.8.0[lapacke,eselect-ldso?]
+		sci-libs/openblas[eselect-ldso?]
```
epsilon-0, line 14: Can't do this before openblas is actually fixed and has `lapacke`, and even in that case we would need to make it dependent only on latest version, by adding a minimum dependency requirement.

Fair warning, there are a large number of changes needed before this PR can even be reviewed, so this PR will be waiting for quite a while.

rossbridger: Good thing I'm not a proxy maintainer for opencv, and I'm merely providing a potential solution for the aforementioned bugs, whether it will be adopted is not really my concern.

epsilon-0: Thanks for the help though, we really need it :+1: .

rossbridger: By the way are you going to bump the latest openblas to 0.3.11? Maybe you can take the opportunity to look into lapacke support for openblas. So far I know that v0.3.10 does install lapacke.h into its include directory.

epsilon-0: I have a lot of backlog of other fixes to do as well.
This is another on the list.
Adding new features (though important) has less priority than fixing packages.

thesamesam: Thoughts on any workarounds until we can fix openblas?

thesamesam:
Note: we can likely merge this (need to review closer) but should merge https://github.com/gentoo/gentoo/pull/17993 first.

rossbridger:
Think it's ready to review

thesamesam:
Thank you!

PR #17971, opened 2020-10-19
Haskell: bump regex-pcre-builtin, drop old

solpeth:
Tested in an `~amd64` container. Synced from my commit in `::haskell`.


trofi:
Pulled.

Thank you!

PR #17976, opened 2020-10-19
Add app-misc/pfetch (new ebuild)

wellWINeo:
I tried to fix all qa issues from previous pull request


juippis:
Hi, 
this can already be found from GURU, 
https://github.com/gentoo/guru/tree/master/app-misc/pfetch
and the ebuild is a little bit better than what you offered here. So we'll be asking the GURU maintainer to bring their work over to ::gentoo once this has matured a bit, and gotten more popular. Thanks!

PR #17977, opened 2020-10-19
net-im/signal-desktop-bin: add new version 1.37.2

robert7k:
Closes: https://bugs.gentoo.org/747775

This adds a new use flag "sound" (enabled by default), which pulls in pulseaudio or apulse. If apulse is installed without the sdk use flag, we need to patch the .desktop entry for signal so that audio calls work.


net-im/signal-desktop-bin/signal-desktop-bin-1.37.2.ebuild
```diff
@@ -90,6 +93,11 @@ src_install() {
 	dodoc changelog
 	doins -r opt
 	insinto /usr/share
+
+	if has_version media-sound/apulse[-sdk]; then
+		sed -i 's/Exec=/Exec=apulse /g' usr/share/applications/signal-desktop.desktop
+	fi
```
SoapGentoo, line 99: what if I happen to have both installed?

SoapGentoo (changes requested):

net-im/signal-desktop-bin/signal-desktop-bin-1.37.2.ebuild
robert7k: You can't, they block each other.

robert7k: I adjusted this: if pulseaudio is installed we don't need to patch the desktop file.

net-im/signal-desktop-bin/signal-desktop-bin-1.37.2.ebuild
```diff
@@ -0,0 +1,115 @@
...
+	x11-libs/libXi
+	x11-libs/libXrandr
+	x11-libs/libXrender
+	x11-libs/libXScrnSaver
+	x11-libs/libXtst
+	x11-libs/pango
+	sound? (
+		|| (
+			media-sound/pulseaudio
+			media-sound/apulse
```
juippis, line 52: The `sed` line below seem to require `-sdk` USE flag for apulse.

robert7k: No, we only need to patch it if apulse is installed without the `sdk` flag. Otherwise, apulse is found automatically and we don't need to patch the `.desktop` file.

juippis (approved):
Okay, let's try it!

PR #17979, opened 2020-10-20
dev-haskell/yesod-core: bump and drop old

solpeth:
Tested in an `~amd64` container. Synced from my commits in `::haskell`.


trofi:
Pulled. Thank you!

PR #17983, opened 2020-10-20
dev-perl/Gtk3-ImageView: new package

DarthGandalf:
This will be future dependency of x11-misc/shutter and media-gfx/gscan2pdf

@cjmayo


kentfredric:
It will probably need love eventually for CFLAGS and LD happening if it has XS parts,  but "Gt" is still ahead of where I am in my audit pass,  so I'll just attack that myself when I get there

DarthGandalf:
For now it's pure perl, without XS. What do you mean by Gt?

kentfredric:
> What do you mean by Gt?

As in, lexicographically I'm going through and san-checking dev-perl/* , I'm at the start of dev-perl/G* right now and so dev-perl/Gt* is logically after it.

dev-perl/Gtk3-ImageView/Gtk3-ImageView-1.ebuild
```diff
@@ -0,0 +1,33 @@
...
+
+EAPI=7
+
+DIST_AUTHOR=RATCLIFFE
+inherit perl-module virtualx
+
+DESCRIPTION="Perl Gtk3 port of the Gtk2::ImageView image viewer widget"
+
+SLOT="0"
+KEYWORDS="~amd64"
```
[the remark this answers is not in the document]
DarthGandalf: I was told that to not do that unless I explicitly test it on x86. But ok

cjmayo:
> @cjmayo

Thanks for the heads-up.

kentfredric:
I note, since you made this ebuild, upstream has already increased by 2 major versions, and now ships a version "3"! 

https://metacpan.org/release/Gtk3-ImageView

This doesn't bode well -_-

DarthGandalf:
I don't think there are any major/minor versions there, just natural numbers.

But yeah, I'll update the ebuild…

kentfredric:
I'll still need to do a local `git am` of this patch and see whatelse fleshes out.

But at this point I'd say it worthwhile adding 
```bash
PERL_RM_FILES=(
    t/90_MANIFEST.t
    t/91_critic.t
)
```
As those tests are basically worthless at best, and harmful at worst.

DarthGandalf:
Done. All the changes between v1 and v4 are to try to make tests work in context of perl smokers, and improving docs.

kentfredric:
Upstream have changed their description to, to `Image viewer widget for Gtk3`, probably should follow it.

Though I'm amused that imagemagick is a test dep. Something seems odd about that, but I'll understand probably when I get around to actually testing it >_>

kentfredric:
Oooh, right, I forget ImageMagick ships .pm files, ew, but I guess. But ew. But I guess. But ew.

DarthGandalf:
What do you suggest replacing `imagemagick[perl]` with?

dev-perl/Gtk3-ImageView/Gtk3-ImageView-4.ebuild
```diff
@@ -0,0 +1,38 @@
...
+inherit perl-module virtualx
+
+DESCRIPTION="Image viewer widget for Gtk3"
+
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="
+	dev-perl/Cairo
+	>=dev-perl/glib-perl-1.210
```
cjmayo, line 16: The only version available is glib-perl-1.329.100.

cjmayo:
I've had a chance to try this out now - gscan2pdf-2.10.0 has been released requiring it, however 2.10.0 also needs PDF-Builder which is also not currently in Gentoo (I have created #18113).

dev-perl/Gtk3-ImageView/Gtk3-ImageView-4.ebuild
DarthGandalf: That's fine. Explicit dependency doesn't hurt, but has benefits of e.g. self-documenting and not breaking overlays

kentfredric:
Sorry I've been tardy at finishing this off, get to it soon :)

kentfredric:
I know this seems "Weird", but for consistency with the scheme we have https://wiki.gentoo.org/wiki/Project:Perl/Version-Scheme ,  this should be version "4.0.0", even though you'll probably never need it, given what upstream are doing.

( But having the version normal up-front makes it clear its been given the relevant love )

kentfredric:
A few corrections to metadata.xml:
```diff
--- a/dev-perl/Gtk3-ImageView/metadata.xml
+++ b/dev-perl/Gtk3-ImageView/metadata.xml
@@ -16,5 +16,9 @@
        <upstream>
                <remote-id type="cpan">Gtk3-ImageView</remote-id>
                <remote-id type="cpan-module">Gtk3::ImageView</remote-id>
+               <remote-id type="cpan-module">Gtk3::ImageView::Tool</remote-id>
+               <remote-id type="cpan-module">Gtk3::ImageView::Tool::Dragger</remote-id>
+               <remote-id type="cpan-module">Gtk3::ImageView::Tool::Selector</remote-id>
+               <remote-id type="cpan-module">Gtk3::ImageView::Tool::SelectorDragger</remote-id>
        </upstream>
 </pkgmetadata>
```

And a rename/DIST_VERSION patch
```diff
--- a/dev-perl/Gtk3-ImageView/Gtk3-ImageView-4.ebuild
+++ b/dev-perl/Gtk3-ImageView/Gtk3-ImageView-4.0.0.ebuild
@@ -4,6 +4,7 @@
 EAPI=7
 
 DIST_AUTHOR=RATCLIFFE
+DIST_VERSION=4
 inherit perl-module virtualx
 
 DESCRIPTION="Image viewer widget for Gtk3"
```

DarthGandalf:
Re metadata: what if different versions have different modules? It should be a union of all?

kentfredric:
> e metadata: what if different versions have different modules? It should be a union of all?

The general approach is to have it map "latest version", partly as this helps when upstream do "splits" or "merges", you can see that right in the metadata if its translated faithfully.

If upstream yank a shipped module, in a future release, and you're using that, you want to stop using that ASAP.

kentfredric:
Few minor qual niggles:
```
>=dev-perl/glib-perl-1.210
```

This would probably be better written "1.210.0" to make it clear its a normalized dep.

And it *might* make sense to add 
```
virtual/perl-Test-Simple
```

Inside `test? ( ... )`

Even though neither of these being missing is a show stopper.

PR #17990, opened 2020-10-21
Update for calcurse to 4.7.0

niyumard:
Signed-off-by: Hosein Naghdbishi <naghdbishi@riseup.net>


juippis (commented):
This works and ebuild looks good, but could you fix the following issue:
```
 * Messages for package app-office/calcurse-4.7.0:

 * 
 *   /usr/share/doc/calcurse
 * 
```
You can either patch or sed the upstream *setup.py*, or simply call `mv` in `src_install`.

juippis (approved):

PR #17991, opened 2020-10-21
media-sound/murmur: Migrate live ebuild to cmake

kripton:
Closes: https://bugs.gentoo.org/745387

Equivalent to  #17468: Convert media-sound/murmur-9999 from qmake to cmake. Drops support for pch but adds support for tests. I was unable to test the systemd-files since all my systems run OpenRC but the server itself is working fine.

### DIFF
--- a/media-sound/murmur/murmur-9999.ebuild
+++ b/media-sound/murmur/murmur-9999.ebuild
@@ -29,7 +29,7 @@ fi
 
 LICENSE="BSD"
 SLOT="0"
-IUSE="+dbus +ice test zeroconf"
+IUSE="+dbus grpc +ice test zeroconf"
 RESTRICT="!test? ( test )"
 
 RDEPEND="
@@ -47,6 +47,7 @@ RDEPEND="
 	sys-apps/lsb-release
 	>=sys-libs/libcap-2.15
 	dbus? ( dev-qt/qtdbus:5 )
+	grpc? ( net-libs/grpc )
 	ice? ( dev-libs/Ice:= )
 	zeroconf? ( net-dns/avahi[mdnsresponder-compat] )
 "
@@ -110,7 +111,7 @@ src_configure() {
 		"-Ddbus=$(usex dbus)"
 		"-Dg15=OFF"
 		"-Dice=$(usex ice)"
-		"-Dgrpc=OFF"
+		"-Dgrpc=$(usex grpc)"
 		"-DBUILD_TESTING=$(usex test)"
 		"-Dzeroconf=$(usex zeroconf)"
 	)
--- a/media-sound/murmur/murmur-9999.ebuild
+++ b/media-sound/murmur/murmur-9999.ebuild
@@ -1,9 +1,9 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-inherit qmake-utils systemd readme.gentoo-r1
+inherit cmake systemd readme.gentoo-r1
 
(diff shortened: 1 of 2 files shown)


kripton:
@Polynomial-C : Any idea why gentoo-bot adds the label "need assignment" and comments `At least one of the listed packages is maintained entirely by non-GitHub developers!`? According to the Gentoo Bugzilla, you are still a Gentoo-dev, right?

kripton:
Fixed issue found by CI (missing line `RESTRICT="!test? ( test )"`)

media-sound/murmur/murmur-9999.ebuild
```diff
@@ -93,36 +94,40 @@ src_prepare() {
...
+		"-Dg15=OFF"
+		"-Dice=$(usex ice)"
+		"-Dgrpc=OFF"
+		"-DBUILD_TESTING=$(usex test)"
+		"-Dzeroconf=$(usex zeroconf)"
 	)
 
-	eqmake5 main.pro -recursive \
-		CONFIG+="${conf_add[*]}"
+	CMAKE_BUILD_TYPE=$(usex debug Debug Release) \
```
DarthGandalf, line 118: You're ignoring CFLAGS from make.conf, use the default build type instead.

kripton: Thanks for the review! I've removed the line after having checked other ebuilds using the `cmake` ECLASS

DarthGandalf: But if `debug` USE flag is not used now, remove it from IUSE

DarthGandalf:
You'll need to ping PolynomialC in IRC, btw. I'm just passing by

media-sound/murmur/murmur-9999.ebuild
kripton: Hm, not so sure about that one... keepassxc also uses the `cmake` eclass, has debug in IUSE but doesn't actually use it inside the ebuild:
https://github.com/gentoo/gentoo/blob/64fe952caa63f4916a5e2f704ebd5306ff00198f/app-admin/keepassxc/keepassxc-2.6.2.ebuild#L27

kripton: Hehe, nice one. I thought maybe the cmake eclass would use that info somehow. I've removed the USE flag from my proposed murmur-update now

DarthGandalf: Usually if a eclass makes use of a USE flag, it adds it to IUSE itself

media-sound/murmur/murmur-9999.ebuild
```diff
@@ -93,36 +95,39 @@ src_prepare() {
...
+	local mycmakeargs=(
+		"-DBUILD_TESTING=$(usex test)"
+		"-Dclient=OFF"
+		"-Ddbus=$(usex dbus)"
+		"-Dg15=OFF"
+		"-Dgrpc=$(usex grpc)"
+		"-Dice=$(usex ice)"
+		"-Doverlay=OFF"
+		"-Dserver=ON"
+		"-Dzeroconf=$(usex zeroconf)"
```
Polynomial-C, line 116: Please also fix the quoting like I did in the mumble ebuild in commit 3629693f926f2a47c4adc08e28987bb265a7aa4a

Polynomial-C (changes requested):
Looking good. There's only the quoting that needs fixing (see my other comment)

media-sound/murmur/murmur-9999.ebuild
kripton: > Looking good. There's only the quoting that needs fixing (see my other comment)

Thanks for the review. Quoting has been adapted and ebuild re-tested

PR #17992, opened 2020-10-21
app-emulation/qemu live ebuild fixes

zippy2:
Upstream has done some cleanups whilst moving towards meson. Update the live ebuild to reflect that.


trofi:
Pulled as is. Thank you!

PR #17994, opened 2020-10-21
sys-auth/pam_u2f: bump to 1.1.0

stintel:
Tested FIDO2 on a YubiKey 5 NFC. Tested U2F on a YubiKey U2F.

Closes: https://bugs.gentoo.org/743385
Signed-off-by: Stijn Tintel <stijn@linux-ipv6.be>


Vicfred:
new yubikeys are not working with older versions, this fixes it :D

gktrk:
Will take a look within the next 24 hours

sys-auth/pam_u2f/pam_u2f-1.1.0.ebuild
```diff
@@ -0,0 +1,34 @@
...
+HOMEPAGE="https://github.com/Yubico/pam-u2f"
+SRC_URI="https://developers.yubico.com/${PN/_/-}/Releases/${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="debug"
+
+RDEPEND="
+	dev-libs/libfido2
```
gktrk, line 18: Per https://github.com/Yubico/pam-u2f/compare/pam_u2f-1.0.8...pam_u2f-1.1.0#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R59, you're missing a dependency on libcrypto (openssl or libressl). It's better to make users of both libraries happy by using something like the following:

```
!libressl? ( dev-libs/openssl:0= )
libressl? ( dev-libs/libressl:0= )
```

sys-auth/pam_u2f/pam_u2f-1.1.0.ebuild
```diff
@@ -0,0 +1,34 @@
...
+
+EAPI=6
+
+inherit autotools flag-o-matic pam
+
+DESCRIPTION="PAM module for FIDO2 and U2F keys"
+HOMEPAGE="https://github.com/Yubico/pam-u2f"
+SRC_URI="https://developers.yubico.com/${PN/_/-}/Releases/${P}.tar.gz"
+
+LICENSE="BSD"
```
gktrk, line 12: Per https://github.com/Yubico/pam-u2f/compare/pam_u2f-1.0.8...pam_u2f-1.1.0#diff-44e0558d931cff7fd8ca7724ad817ffd7c953ec032ba572d1ff1f244656ff4ccR6, they've added a source file with the `ISC` license, so we should update `LICENSE` to reflect that.

sys-auth/pam_u2f/pam_u2f-1.1.0.ebuild
```diff
@@ -0,0 +1,34 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
```
gktrk, line 4: This package doesn't get frequent updates. So this might be a good time to bump its EAPI with this opportunity. Required changes are minimal: move the dependency on `virtual/pkgconfig` from `DEPEND` to `BDEPEND` and have `DEPEND` and `RDEPEND` be identical.

sys-auth/pam_u2f/pam_u2f-1.1.0.ebuild
gktrk, line 18: I also realized that `libfido2` has subslots now. Let's be proactive and add a [slot operator](https://devmanual.gentoo.org/general-concepts/dependencies/#slot-operators) as follows:

```
dev-libs/libfido2:0=
```

gktrk (changes requested):
Almost there. Let me know if you're not willing to work on updating it and I'll just bump it myself.

stintel:
> Almost there. Let me know if you're not willing to work on updating it and I'll just bump it myself.

Like this? Or should the EAPI bump go in a separate commit as it's not really related to the version bump?

sys-auth/pam_u2f/pam_u2f-1.1.0.ebuild
```diff
@@ -0,0 +1,36 @@
...
+
+EAPI=7
+
+inherit autotools flag-o-matic pam
+
+DESCRIPTION="PAM module for FIDO2 and U2F keys"
+HOMEPAGE="https://github.com/Yubico/pam-u2f"
+SRC_URI="https://developers.yubico.com/${PN/_/-}/Releases/${P}.tar.gz"
+
+LICENSE="ISC"
```
gktrk, line 12: They didn't switch to `ISC`, just added additional sources that are licensed with `ISC`. So, we need to list both.

gktrk (approved):
For the most part looks good. I won't waste your time with some of the minor fixes. Good job and thank you for the PR!

sys-auth/pam_u2f/pam_u2f-1.1.0.ebuild
stintel: My bad. Thanks for fixup!

PR #17996, opened 2020-10-21
sys-cluster/charliecloud: Version bump to 0.20.

olifre:
This also drops the oldest version, getting rid of a local `USE` flag which is now unused.


sys-cluster/charliecloud/charliecloud-0.20.ebuild
```diff
@@ -0,0 +1,80 @@
...
+# Extensive test suite exists, but downloads container images
+# directly and via Docker and installs packages inside using apt/yum.
+# Additionally, clashes with portage namespacing and sandbox.
+RESTRICT="test"
+
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+RDEPEND="${PYTHON_DEPS}
+	squashfs? ( sys-fs/squashfs-tools )
+	squashfuse? ( sys-fs/squashfuse )"
```
juippis: RDEPEND-only USE flags are banned by policy, 

You should put these to `optfeature`.

sys-cluster/charliecloud/charliecloud-0.20.ebuild
```diff
@@ -0,0 +1,80 @@
...
+	SRC_URI="https://github.com/hpc/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64 ~x86 ~x86-linux"
+fi
+
+DESCRIPTION="Lightweight user-defined software stacks for high-performance computing"
+HOMEPAGE="https://hpc.github.io/charliecloud/"
+
+SLOT="0"
+LICENSE="Apache-2.0"
+IUSE="ch-grow doc examples +squashfs squashfuse"
```
juippis, line 24: `examples` IUSE doesn't do anything either in here.

sys-cluster/charliecloud/charliecloud-0.20.ebuild
olifre: Thanks for pointing this out, I moved them! 
What I dislike about the policy is that it makes it impossible to express a strong recommendation (which also upstream mentions), in this case to install `squashfs` since noticeable functionality will be missing (that can be expressed via `+squashfs` in `IUSE`). 
But of course, I'll adhere to policy (and understand the reasoning) — and if GLEP 62 is implemented, I'll be very happy about that, since it will bring back this possibility :-).

sys-cluster/charliecloud/charliecloud-0.20.ebuild
olifre: Definetly an oversight, many thanks for catching that, it was used in earlier versions.

sys-cluster/charliecloud/charliecloud-0.20.ebuild
juippis: If you think it'd be worth using it in ~100 % of cases, you can make it an unconditional dependency. Let me know what are your thoughts.

olifre: 100 % is not the case, so I would indeed leave it as it is now: It is useful in >90 % for users who build containers with charliecloud, but not for users running containers only. 
So in short, I'd leave it as is ;-).

PR #17997, opened 2020-10-22
dev-python/aiosmtpd: bump to 1.2.1 and add py3_9

wimmuskee:
py3_9 for dev-python/aiosmtpd and dep dev-python/atpublic

Also added tests for both packages. Skipping one aiosmtpd test because of an expired certificate that requires regular maintenance.


wimmuskee:
@prometheanfire ping

PR #17999, opened 2020-10-22
dev-python/semver: python 3.9 support

ZeroPointEnergy:
Simple addition of python 3.9 support and removal of redundant test use flag

The test are all passing.


gyakovlev (approved):

gyakovlev:
also bumped to new version and added pypy3 in the new one. thanks!

PR #18001, opened 2020-10-22
*/*: QA fixes for my packages

jsmolic:
QA fixes from [gentooqa.levelnine.at](https://gentooqa.levelnine.at/results/gentoo/listings/sort-by-maintainer/jakov.smolic_at_sartura.hr.txt) report:

* net-libs/libwebsockets: `metadata_duplicate_useflag_description` (Remove USE flag descriptions which are already available at
 `profiles/use.desc`)
* media-libs/libheif: `ebuild_unused_patches` (Remove unused patch)
* sys-fs/bashmount: `ebuild_homepage_unsync` (Sync HOMEPAGE)


net-libs/libwebsockets/metadata.xml
```diff
@@ -19,7 +19,6 @@
 	</maintainer>
 	<use>
 		<flag name="access-log">Support generating Apache-compatible access logs</flag>
-		<flag name="cgi">Include CGI (spawn process with network-connected stdin/out/err) APIs</flag>
```
juippis, line 22: The default value is 

>Add CGI script support

Would you say the current metadata flag has a different meaning?

net-libs/libwebsockets/metadata.xml
```diff
@@ -32,9 +31,7 @@
 		<flag name="peer-limits">Track peers and restrict resources a single peer can allocate</flag>
 		<flag name="server-status">Support json + jscript server monitoring</flag>
 		<flag name="smtp">Provide SMTP support</flag>
-		<flag name="socks5">Allow use of SOCKS5 proxy on client connections</flag>
 		<flag name="sqlite3">Require SQLITE3 support</flag>
-		<flag name="zip">Support serving pre-zipped files</flag>
```
juippis, line 37: >Enable support for ZIP archives

Here too. While this can be understood as "support for zip archives", it can also mean the program feeds "bundled" zip archives instead of something else. I'm not familiar with libwebsockets to answer this. But let me know should these be updated or not, these changes do look good after all.

jsmolic: An excerpt from README:
```
Lws now supports serving gzipped files from inside a zip container.

This has the advtantage that if the client can accept GZIP encoding, lws can simply send the gzip-compressed file from inside the zip file with no further processing, saving time and bandwidth.
```

@jkroonza Any thought regarding this?

jkroonza: Enable regardless.  It's quite useful.  My use case don't need it, but I don't see this adding a significant code base.  Exception to this:  If we need additional dependencies.  Most systems already have gzip, so if that's what's required, enable regardless, if we need actual zip ... different story.  But Juippis is right, the flag description could do with improvement, eg:  Support serving content straight from .zip files where browser supports gzip encoding.

juippis: No no, the question is, does the global USE definiton match the feature enabled in this package. Or should it be re-defined locally, using *metadata.xml*. The USE itself is not being removed.

net-libs/libwebsockets/metadata.xml
jkroonza: Global description is correct, so use that.

net-libs/libwebsockets/metadata.xml
jkroonza: Ah ok, my bad. So two issues here:

1.  Should this even be a USE flag.  Refer above, but not the original question.

2.  Does the description need overriding:

global:  Enable support for ZIP archives

So that's exactly what the USE flag does, the question is do we need to provide more info than that, and my opinion is no we don't.

jsmolic: Ok, thanks, I though we could safely drop metadata.xml descriptions but wanted to hear your opinion. Perhaps on the next version bump we can drop some flags to reduce the complexity.

net-libs/libwebsockets/metadata.xml
jsmolic: Yup, thank you

juippis (approved):

jsmolic:
Thanks @juippis

PR #18008, opened 2020-10-23
Haskell: bump darcs and base16-bytestring, drop old

solpeth:
Tested in an `~amd64` container.

The `curl` patch for `darcs` has been incorporated upstream and has therefore been dropped.


trofi:
Pulled. Thank you!

PR #18009, opened 2020-10-23
sci-libs/openlibm: version bump to 0.7.2

epsilon-0:
fix CC AR and other toolchain calls


sci-libs/openlibm/openlibm-0.7.2.ebuild
```diff
@@ -1,23 +1,29 @@
...
 LICENSE="public-domain MIT ISC BSD-2 LGPL-2.1+"
 SLOT="0/${PV}.0"
 KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
 
 IUSE="static-libs"
 
+PATCHES=( "${FILESDIR}"/${PN}-0.7.2-make_inc.patch )
+
+pkg_setup() {
+	tc-export CC CXX FC AR LD
```
SoapGentoo, line 21: do this in `src_configure`

sci-libs/openlibm/openlibm-0.7.2.ebuild
```diff
@@ -1,23 +1,29 @@
...
+}
+
 src_install() {
 	emake DESTDIR="${D}" prefix="${EPREFIX}/usr" \
 		libdir="${EPREFIX}/usr/$(get_libdir)" install
-	use static-libs || rm "${D}/${EPREFIX}/usr/$(get_libdir)/libopenlibm.a" || die "rm failed"
+	use static-libs || rm -f "${ED}/usr/$(get_libdir)/libopenlibm.a" || die "rm failed"
```
SoapGentoo, line 27: remove static-libs, and preferably from the build system

SoapGentoo (changes requested):

sci-libs/openlibm/openlibm-0.7.2.ebuild
epsilon-0: done

sci-libs/openlibm/openlibm-0.7.2.ebuild
epsilon-0: Hoping to upstream this patch, so we don't have to maintain.
They won't accept it we remove static building.
I think this is an okay compromise for now.

sci-libs/openlibm/openlibm-0.7.2.ebuild
```diff
@@ -1,23 +1,28 @@
...
+src_configure() {
+	tc-export CC CXX FC AR LD
+	default
+}
 
 src_install() {
 	emake DESTDIR="${D}" prefix="${EPREFIX}/usr" \
 		libdir="${EPREFIX}/usr/$(get_libdir)" install
-	use static-libs || rm "${D}/${EPREFIX}/usr/$(get_libdir)/libopenlibm.a" || die "rm failed"
+	rm -f "${ED}/usr/$(get_libdir)/libopenlibm.a" || die "rm failed"
```
SoapGentoo, line 26: is preventing the build system from building static libraries hard? While this is better, it still wastes unnecessary cycles on building pointless object code.

epsilon-0: no it wouldn't be hard but I'd rather do it outside a patch
will do in a bit

epsilon-0: done

sci-libs/openlibm/openlibm-0.7.2.ebuild
```diff
@@ -0,0 +1,35 @@
...
+
+src_configure() {
+	tc-export CC CXX FC AR LD
+	default
+}
+
+src_install() {
+	emake DESTDIR="${D}" prefix="${EPREFIX}/usr" \
+		libdir="${EPREFIX}/usr/$(get_libdir)" install
+#	rm -f "${ED}/usr/$(get_libdir)/libopenlibm.a" || die "rm failed"
```
SoapGentoo, line 33: dead code, why leave this in here?

sci-libs/openlibm/openlibm-0.7.2.ebuild
SoapGentoo: Why did you remove the fortran eclass? Please add it back, all fortran ebuilds need it.

SoapGentoo (changes requested):

sci-libs/openlibm/openlibm-0.7.2.ebuild
epsilon-0: whoops, sorry.
fixed.

sci-libs/openlibm/files/openlibm-0.7.2-make_inc.patch
```diff
@@ -0,0 +1,87 @@
...
+ endif
+ 
+@@ -84,9 +84,9 @@ override ARCH := mips
+ endif
+ 
+ # If CFLAGS does not contain a -O optimization flag, default to -O3
+-ifeq ($(findstring -O,$(CFLAGS)),)
+-CFLAGS_add += -O3
+-endif
++#ifeq ($(findstring -O,$(CFLAGS)),)
```
thesamesam, line 69: Don't bother including commented lines in patches because the information is lost anyway (the - lines are in the patch) and it makes the patch shorter.

sci-libs/openlibm/openlibm-0.7.2.ebuild
```diff
@@ -1,23 +1,34 @@
...
-inherit eutils toolchain-funcs fortran-2
+inherit toolchain-funcs
 
 DESCRIPTION="High quality system independent, open source libm"
 HOMEPAGE="https://github.com/JuliaLang/openlibm"
-SRC_URI="https://github.com/JuliaLang/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+SRC_URI="https://github.com/JuliaMath/openlibm/archive/v${PV}.tar.gz -> ${P}.tar.gz"
 
 LICENSE="public-domain MIT ISC BSD-2 LGPL-2.1+"
 SLOT="0/${PV}.0"
```
thesamesam, line 13: Huh. Weird subslot format, but OK.

sci-libs/openlibm/openlibm-0.7.2.ebuild
```diff
@@ -1,23 +1,34 @@
...
 
-inherit eutils toolchain-funcs fortran-2
+inherit toolchain-funcs
 
 DESCRIPTION="High quality system independent, open source libm"
 HOMEPAGE="https://github.com/JuliaLang/openlibm"
-SRC_URI="https://github.com/JuliaLang/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+SRC_URI="https://github.com/JuliaMath/openlibm/archive/v${PV}.tar.gz -> ${P}.tar.gz"
 
 LICENSE="public-domain MIT ISC BSD-2 LGPL-2.1+"
```
thesamesam, line 12: When there's a lot of licences, I like to comment where they come from & why. Not sure if applicable here.

epsilon-0: I dunno why they are there, so I didn't remove them :confused: .

sci-libs/openlibm/openlibm-0.7.2.ebuild
epsilon-0: Not sure why it is present.
But because it was present already, I didn't want to remove in case some esoteric julia thing breaks.

sci-libs/openlibm/openlibm-0.7.2.ebuild
thesamesam: Then go check what the licence should be! Maybe more got added, or fewer, or ...

sci-libs/openlibm/openlibm-0.7.2.ebuild
thesamesam: A subslot is quite common to need to force a rebuild of consumers, I just don't get why they bother with the .0. But it's alright.

sci-libs/openlibm/openlibm-0.7.2.ebuild
epsilon-0: https://github.com/JuliaMath/openlibm/blob/master/LICENSE.md
This seems to be the reason.
 I don't think it needs to be commented if its present in the License file, which is quite straightforward to fine.

sci-libs/openlibm/files/openlibm-0.7.2-make_inc.patch
epsilon-0: done

PR #18014, opened 2020-10-24
dev-libs/hiredis: fix build with USE="-ssl"

arkamar:
I changed `rm` to `rm -f` in order to ignore missing `${ED}/usr/$(get_libdir)/libhiredis.a` when `USE=-ssl` is set.


Whissi:
Thank you but the normal fix for problems like that is `find "${ED}" -name '*.a' -delete || die` which I am going to merge.

arkamar:
Thanks, I'll remeber it for next time.

PR #18015, opened 2020-10-24
GLEP81 pcap

SoapGentoo:


acct-group/pcap/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>netmon@gentoo.org</email>
```
mgorny, line 5: I'm not convinced dumping it on netmon is the best idea.

PR #18026, opened 2020-10-24
 games-fps/yamagi-quake2: Version bump to 7.45

puleglot:


games-fps/yamagi-quake2/yamagi-quake2-7.45.ebuild
```diff
@@ -0,0 +1,149 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit desktop eutils
```
juippis, line 6: *eutils* is being deprecated, you call `make_wrapper` using *wrapper.eclass* now.

juippis (approved):

PR #18029, opened 2020-10-25
profiles: remove virtual/ffmpeg from packages.deprecated (no longer in tree)

chusopr:
After no revdeps were left, the package was removed by @mgorny in 0e5b93b, but it's still listed in `packages.deprecated`.

Shouldn't it be removed there too? Or is there any reason for it to stay there (like, e.g., overlays)?

### DIFF
--- a/profiles/package.deprecated
+++ b/profiles/package.deprecated
@@ -17,11 +17,6 @@
 
 #--- END OF EXAMPLES ---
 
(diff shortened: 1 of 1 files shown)


mgorny:
Yes, it was left deliberately, so that overlays still get a warning about it.

chusopr:
OK, thanks for the clarification. I'm closing https://bugs.gentoo.org/744787 then.

PR #18031, opened 2020-10-25
selinux profile: set sys-process/audit to use python by default.

jpds:
`semanage` requires `auditd` to be compiled with `USE=python` to be usable.


mgorny:
Shouldn't the dependency in libsemanage be corrected too?

perfinion:
I merged this, and also fixed the dep on selinux-python (its not the libsemanage package)

perfinion:
Merged as 7a9ae808043f2ec47ef48c480cd6274767e82a52

PR #18035, opened 2020-10-26
media-fonts/koruri: version bump to 20180915

omasanori:
Note that the project was moved from SourceForge to GitHub.


omasanori:
A new version [20210720](https://github.com/Koruri/Koruri/releases/tag/20210720) was released, so this PR shall be revised. Closing for now.

PR #18037, opened 2020-10-26
net-misc/httpie: Bump to version 2.3.0

rseichter:
Also, delete obsolete ebuild version 2.1.0.


net-misc/httpie/httpie-2.3.0.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DISTUTILS_USE_SETUPTOOLS=rdepend
+PYTHON_COMPAT=( python3_{6,7,8} )
```
Whissi: No Python 3.9 support yet?

net-misc/httpie/httpie-2.3.0.ebuild
```diff
@@ -0,0 +1,39 @@
...
+	>=dev-python/requests-2.22.0[${PYTHON_USEDEP}]"
+DEPEND="test? (
+		${RDEPEND}
+		dev-python/mock[${PYTHON_USEDEP}]
+		dev-python/pyopenssl[${PYTHON_USEDEP}]
+		dev-python/pytest-httpbin[${PYTHON_USEDEP}]
+		dev-python/pytest[${PYTHON_USEDEP}]
+	)"
+
+distutils_enable_tests pytest
```
Whissi: This function will add `dev-python/pytest` to deps so you shouldn't manually add it to `DEPEND`.

net-misc/httpie/httpie-2.3.0.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="test"
+RESTRICT="!test? ( test )"
+
+RDEPEND="dev-python/pygments[${PYTHON_USEDEP}]
+	>=dev-python/requests-2.22.0[${PYTHON_USEDEP}]"
+DEPEND="test? (
```
Whissi: This is EAPI 7, additional test deps belong into BDEPEND.

Whissi (changes requested):

net-misc/httpie/httpie-2.3.0.ebuild
rseichter: According to upstream's `setup.py`, any Python version >= 3.6 will do. I don't have a test environment available, but I guess adding 3.9 anyway won't hurt.

net-misc/httpie/httpie-2.3.0.ebuild
rseichter: Fixed.

net-misc/httpie/httpie-2.3.0.ebuild
rseichter: Fixed.

juippis (commented):
Looks like you're missing some test deps,
```
E   ModuleNotFoundError: No module named 'requests_toolbelt'
______________________________ ERROR collecting tests/test_stream.py _______________________________
ImportError while importing test module '/var/tmp/portage/gpm/portage/net-misc/httpie-2.3.0/work/httpie-2.3.0/tests/test_stream.py'.
```
```
===================================== short test summary info ======================================
ERROR tests/test_auth.py
ERROR tests/test_auth_plugins.py
ERROR tests/test_binary.py
ERROR tests/test_cli.py
ERROR tests/test_compress.py
ERROR tests/test_config.py
ERROR tests/test_defaults.py
ERROR tests/test_docs.py
ERROR tests/test_downloads.py
ERROR tests/test_errors.py
ERROR tests/test_exit_status.py
ERROR tests/test_httpie.py
ERROR tests/test_offline.py
ERROR tests/test_output.py
ERROR tests/test_redirects.py
ERROR tests/test_regressions.py
ERROR tests/test_sessions.py
ERROR tests/test_ssl.py
ERROR tests/test_stream.py
ERROR tests/test_unicode.py
ERROR tests/test_uploads.py
ERROR tests/test_windows.py
!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 22 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!
======================================== 22 errors in 2.50s ========================================
 * ERROR: net-misc/httpie-2.3.0::testworld failed (test phase):
 *   Tests fail with python3.6
```
Although are these run locally or do they require an internet connection? Try to disable them if they require an active connection.

rseichter:
Could this be related to @Whissi asking me to remove dev-python/pytest ? I cannot access to my Gentoo environment to check this from where I am currently located.

juippis:
Nope, it doesn't depend on *requests-toolbelt* at all which I imagine is missing. Can test later today, and add it by myself if that's a simple fix. 
(I just wanted you to check if there are more test deps missing)

rseichter:
Thanks, I'll likely be restricted in terms of Gentoo for at least two weeks due to a new contract. Aside from the changes requested by Whissi, this ebuild is a copy of the previous one.

thesamesam:
(Let's be clear why you could remove the pytest dep, because it's worth understanding why before doing things: distutils_enable_tests pytest [automatically](https://dev.gentoo.org/~mgorny/python-guide/distutils.html#enabling-tests) adds a Pytest dep, the test use flag, and the conditional restrict. See also the [eclass](https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/distutils-r1.eclass#n399)).

juippis:
Aaaaand no, tests didn't work, apparently at least 5 test cases are trying to connect to the Internet. I'll investigate if there's an easy way to disable them tomorrow.

rseichter:
Stepping down as maintainer

PR #18038, opened 2020-10-26
games-server/minecraft-server: Add systemd unit.

rossbridger:
This PR added a systemd unit that allows users to run a minecraft server as a systemd service which is compatible with the existing servers created by the openrc init scripts.

Bug: https://bugs.gentoo.org/750680

### DIFF
--- a/games-server/minecraft-server/files/minecraft-server.service
+++ b/games-server/minecraft-server/files/minecraft-server.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Minecraft Server (World: %I)
+After=network.target
+
+[Service]
+User=minecraft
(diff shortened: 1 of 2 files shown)


ConiKost:
Since you added ~arm64 arch and didn't mentioned that, did you actually tested on that arch?

rossbridger:
> Since you added ~arm64 arch and didn't mentioned that, did you actually tested on that arch?

I tested it on a raspberry pi 4

ConiKost:
Allright. Please mention that in the commit message, that you added ~arm64 arch and tested on your Pi4.

ConiKost (approved):

PR #18040, opened 2020-10-26
www-apps/bugzilla: optional dep fix, update init script

janko-j1:


rossbridger:
Does it also have a coupling systemd service?

janko-j1:
> Does it also have a coupling systemd service?

No, but I'll gladly update the ebuild if one provides such a file :-)

juippis (commented):
You may want to revbump any bugzilla ebuilds using this init scripts, so everyone gets the update. Just use `git mv` on the same commit as you update the init.d file.

juippis (approved):

PR #18042, opened 2020-10-26
sys-block/zram-init: version bump

beatussum:


rossbridger:
I think installing the systemd units into /usr/lib/systemd/ is deprecated in favor of /lib/systemd/

beatussum:
I don't manage to find this depreciation in the freedesktop documentation: could you tell me where you found it?

rossbridger:
It's old news: https://www.gentoo.org/support/news-items/2017-07-16-systemd-rootprefix.html

beatussum:
Thanks, I'll fix that.

telans:
10.9 is available

beatussum:
version bump

PR #18047, opened 2020-10-27
dev-php/composer: bump to 2.0.4

hydrapolic:
Tested on ~amd64.


dev-php/promise/promise-2.8.0.ebuild
```diff
@@ -0,0 +1,21 @@
...
+HOMEPAGE="https://github.com/reactphp/promise/"
+SRC_URI="https://github.com/reactphp/promise/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="dev-lang/php:*
+	dev-php/fedora-autoloader"
+
```
GuillaumeSeren, line 16: Would it be possible to add test for this package ?

dev-php/promise/metadata.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+  <maintainer type="person">
```
GuillaumeSeren, line 4: As you are creating this one please replace my name / mail with yours

dev-php/promise/files/autoload.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/* Autoloader for composer/ca-bundle and its dependencies */
```
GuillaumeSeren, line 2: Bad package name in the comment ?

dev-php/promise/metadata.xml
hydrapolic: It's needed for composer, that's why the maintainers of composer are listed here.

dev-php/promise/promise-2.8.0.ebuild
hydrapolic: Haven't really tried, just tested by running composer itself. They are not included in source code anyway.

dev-php/promise/files/autoload.php
hydrapolic: Yeah, thanks, removed.

dev-php/promise/metadata.xml
GuillaumeSeren: Ah I didn't know that,
to me the metadata maintainer section is only for itself but I may be wrong.

dev-php/promise/promise-2.8.0.ebuild
GuillaumeSeren: Yes tests are not in the release tarball :(

dev-php/promise/metadata.xml
juippis: So @GuillaumeSeren do you agree being added here or not? Do you use composer anymore and care about it still?

dev-php/promise/metadata.xml
```diff
@@ -0,0 +1,18 @@
...
+    <email>guillaumeseren@gmail.com</email>
+    <name>Guillaume Seren</name>
+  </maintainer>
+  <maintainer type="project">
+    <email>proxy-maint@gentoo.org</email>
+    <name>Proxy Maintainers</name>
+  </maintainer>
+  <maintainer type="project">
+    <email>php-bugs@gentoo.org</email>
+  </maintainer>
```
juippis, line 14: And we can't add this project without their consent either :\

hydrapolic: It's a new direct dependency, that's why the current maintainers are added.

Whissi: I am still working on this, currently busy. Hope to get to this in this week.

dev-php/promise/metadata.xml
GuillaumeSeren: Yes I still work on it,
actually I discuss to add more tests on base packages, so you can add me.

dev-php/promise/metadata.xml
juippis: @Whissi ping.

Whissi: Sorry for the delay, had to get phpunit problems out of the way first. Looking forward for the coming week, but PHP 8 will be first.

r7l:
While i appreciate the work on bringing Composer 2 to Gentoo, i wonder if there is any news about this? Composer starts to complain about users having 1.x versions:

`Warning from https://repo.packagist.org: You are using an outdated version of Composer. Composer 2.0 is now available and you should upgrade. See https://getcomposer.org/2`

It's still working for now but Composer 2.x is also said to have some speed improvements.

kuipumu:
> While i appreciate the work on bringing Composer 2.x to Gentoo, i wonder if there is any news about this? Composer starts to complain about users having 1.x versions:
> 
> `Warning from https://repo.packagist.org: You are using an outdated version of Composer. Composer 2.0 is now available and you should upgrade. See https://getcomposer.org/2`
> 
> It's still working for now but Composer 2.x is also said to have some speed improvements.

What is the status on the bump to 2.0.4?, I'm having the same warning from Composer

Whissi:
I am currently working on it.

Whissi:
@GuillaumeSeren You don't have an idea how to teach phpab to set correct dir so we don't have to ship custom autoloader template like I did as workaround via https://github.com/gentoo/gentoo/commit/110f7ca5828a08980e8fa472e30149536e382084#diff-e2bdc0bdfcab3e483d9853480125139029373f69d21bd004c0df83cba5d49aa2R10, do you?

hydrapolic:
Huge thanks @Whissi !

GuillaumeSeren:
@Whissi not really,
as I wanted to get a static autoloader I try to avoid phpab,
but in the case of phpunit I asked upstream I was suggested to use phpab.
see: https://github.com/sebastianbergmann/phpunit/issues/4159

PR #18049, opened 2020-10-27
profiles/package.mask: mask sci-libs/{blas,cblas,lapack}

epsilon-0:


rossbridger:
Doesn't this collide with https://github.com/gentoo/gentoo/pull/17836?

epsilon-0:
@rossbridger 
yea this is being done earlier so that people can have time to prepare.
then we will do the removal later.

epsilon-0:
@SoapGentoo here it is

SoapGentoo (approved):

SoapGentoo:
@epsilon-0 could you rebase this commit pls?

PR #18057, opened 2020-10-28
gui-wm/hikari: version bump to 2.2.2

epsilon-0:
should really fix the build var behaviour.


juippis (approved):
I see both are fixed, thanks!

PR #18058, opened 2020-10-28
app-emulation/firecracker-bin: Bump to 0.23.0

s-hamann:
Trivial version bump for release of Firecracker v0.23.0.

I changed `KEYWORDS` to not disable `arm64` by wildcard. This is for https://bugs.gentoo.org/740886


app-emulation/firecracker-bin/firecracker-bin-0.23.0.ebuild
```diff
@@ -0,0 +1,82 @@
...
+		https://github.com/firecracker-microvm/firecracker/releases/download/v${PV}/jailer-v${PV}-x86_64
+	)
+	arm64? (
+		https://github.com/firecracker-microvm/firecracker/releases/download/v${PV}/firecracker-v${PV}-aarch64
+		https://github.com/firecracker-microvm/firecracker/releases/download/v${PV}/jailer-v${PV}-aarch64
+	)"
+
+LICENSE="|| ( Apache-2.0 MIT Apache-2.0-with-LLVM-exceptions ) MPL-2.0"
+SLOT="0"
+KEYWORDS="-alpha ~amd64 -arm -hppa -ia64 -ppc -ppc64 -sparc -x86"
```
juippis, line 22: This is pretty useless, it's the same as leaving it null. Seeing this would confuse people, let's just take all `-` KEYWORDS out.

s-hamann: Really? The [dev manual](https://devmanual.gentoo.org/keywording/index.html) says that not mentioning an arch means that support is unknown, while `-arch` means that the arch is known to be unsupported. I think the latter is the case here, since upstream releases binaries only for two arches (`amd64` and `arm64`). Binary only packages are explicitly mentioned as an example for when `-arch` is sensible.
I really just "expanded" the `-*` (as used in `app-emulation/firecracker-bin-0.22.1`) to be able to make `arm64` "unknown", so that hopefully that keyword can be added. This was suggested by @thesamesam in https://bugs.gentoo.org/740886#3.
Did I mix something up here?

juippis: In both cases it should be masked by a missing keyword, and be unfetchable anyway. so it doesn't make any difference. This is just ugly.

s-hamann: Alright, changed it.

juippis (approved):

PR #18059, opened 2020-10-28
dev-util/bats: add "threads" USE and pull in rdep

henning-schild:
Users need that rdep when wanting to use the "--jobs" feature.

### DIFF
--- a/dev-util/bats/bats-1.2.1.ebuild
+++ b/dev-util/bats/bats-1.2.1.ebuild
@@ -3,6 +3,8 @@
 
 EAPI=7
 
+inherit optfeature
+
 MY_PN="bats-core"
 DESCRIPTION="Bats-core: Bash Automated Testing System"
 HOMEPAGE="https://github.com/bats-core/bats-core/"
@@ -30,4 +32,8 @@ src_install() {
 
 	dodoc README.md
 	doman man/${PN}.1 man/${PN}.7
+}
+
+pkg_postinst() {
+	optfeature "Parallel Execution" sys-process/parallel
 }


henning-schild:
I am not sure whether "threads" would be the correct USE-flag, it is technically processes ... but did not find a better one.

dev-util/bats/bats-1.2.1-r1.ebuild
```diff
@@ -0,0 +1,38 @@
...
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~ppc64 ~x86"
+
+IUSE="threads"
+
+DEPEND="
+	app-shells/bash:*
+	!ppc64? ( threads? ( sys-process/parallel ) )
```
juippis, line 19:
```suggestion
	ppc64? ( sys-process/parallel )
```
Since `threads` doesn't actually do anything in this ebuild.

gyakovlev:
I've keyworded parallel on ppc64.
the way you omitted ppc64 is no acceptable btw, they way we omit useflag is masking in profiles.

also adding use-flags for `RDEPEND` only is generally prohibited.
you can just print message in postinst or use `optfeature.eclass`

https://projects.gentoo.org/qa/policy-guide/dependencies.html

dev-util/bats/bats-1.2.1-r1.ebuild
henning-schild: switched to postinst printing only

gyakovlev (approved):

gyakovlev:
thanks! merged.

PR #18064, opened 2020-10-29
games-emulation/vbam: don't override CMAKE_AR

DarthGandalf:
Closes: https://bugs.gentoo.org/733938

### DIFF
--- a/games-emulation/vbam/vbam-2.1.4.ebuild
+++ b/games-emulation/vbam/vbam-2.1.4.ebuild
@@ -70,6 +70,7 @@ src_configure() {
 		-DENABLE_ASM_CORE=$(usex x86)
 		-DENABLE_ASM_SCALERS=$(usex x86)
 		-DCMAKE_SKIP_RPATH=ON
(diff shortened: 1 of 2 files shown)


chewi (changes requested):
Sorry but a patch isn't needed here because you can just pass `-DENABLE_LTO=OFF` instead. The extra logic the project has around LTO doesn't seem necessary so there's no harm in doing this.

DarthGandalf:
Doesn't `set()` make it to ignore `-D`? `set(ENABLE_LTO ON CACHE ...)` would work like you describe.

chewi:
You did have me thinking I'd got confused but actually I was right. :stuck_out_tongue: `LTO_DEFAULT` is defined with `set` but `ENABLE_LTO` is already an `option`.

```
option(ENABLE_LTO "Compile with Link Time Optimization (gcc and clang only)" ${LTO_DEFAULT})
```

DarthGandalf:
Indeed. I haven't noticed that there are 2 variables, not one.
Fixed.

PR #18066, opened 2020-10-29
app-admin/rex: bump version to 1.13.x, drop old

ferki:
This PR updates the live ebuild slightly, and then cuts a new ebuild for the latest 1.13.0 version of Rex.

@kentfredric: please review and merge, or let me know how to improve it further.


ferki:
New month, new Rex release; so I've rebased this pending PR on top of current master, included the changes for the latest version, and dropped older ebuilds.

app-admin/rex/rex-1.13.1.ebuild
```diff
@@ -0,0 +1,203 @@
...
+		-e '/^Test::PerlTidy =/d' \
+		-e '/^Test::Pod =/d' \
+		-e '/^\[Test::CPAN::Changes\]/{N;d}' \
+		-e '/^\[OptionalFeature/,/^$/d' \
+		-e '/^\[Test::MinimumVersion\]/{N;d}' \
+		-i dist.ini || die "Can't patch dist.ini"
+}
+dzil_env_setup() {
+	# NextRelease noise :(
+	mkdir -p ~/.dzil/
```
monsieurp: I would add a `|| die` here but it's a nitpick. I will add it before merging the PR.

monsieurp (approved):

app-admin/rex/rex-1.13.1.ebuild
ferki: Right, that wouldn't hurt :+1: I'll make sure to merge back any change to my overlay too.

ferki:
Thank you @monsieurp for the feedback and merge! :heart:

app-admin/rex/rex-1.13.1.ebuild
kentfredric: > I will add it before merging the PR.

I hope you added it in all the files ... because there was "no real diff" involving that line of code. Only stupid github showing copied files as "vanilla diffs".

ferki: If I see correctly, it was only added to the 1.13.1 ebuild for now.

I'll add it to the next iteration of the live ebuild, so when we cut the next releases based on that, the future ones include it too.

Is there any substantial value to add it retrospectively to older ebuilds as well? Want me to do a new PR for that?

kentfredric: I'm still gonna go and have to to properly finish my audit and my fixes, which like, now I have even more pressure on.

No new PR is required, just gets annoying when people can't even review PR's properly, and make changes based on reviewing the PR wrong. 

I feel like taking @gentoo/proxy-maint out of the loop just to discourage people who don't understand what they're doing from doing things.

juippis: No objections from me, you're the best to handle ebuilds like this. Just don't leave @ferki hanging with his contributions ;)

(I specifically ignored this PR for a while hoping @kentfredric would deal with it)
((to my taste the ebuild still has many QA issues, but nitpicking...))

kentfredric: > Is there any substantial value to add it retrospectively to older ebuilds as well? Want me to do a new PR for that?

No, zero. The code in question that was patched *CAN ONLY* run when $PV == 9999. 

So the patching did nothing, and it didn't fix the potential problem in the one place it *could* run.

Fixed now in -9999 as well. https://github.com/gentoo/gentoo/commit/0112abb1935e4582dea0ed1d433528a748826db1

But backporting it is a big waste of time. lol

kentfredric:
@juippis if you have criticisms of the code, I'm all for them. Its just like, not something that the PR submitter has to deal with.

After all, I wrote 99% of it. https://github.com/gentoo/gentoo/blame/0112abb1935e4582dea0ed1d433528a748826db1/app-admin/rex/rex-9999.ebuild 

But ideally, focus on the stuff outside the functions with `dzil_` in their name, as those do not, and cannot run, outside the -9999 version.

Stuff in the -9999 code paths is still worth reviewing, but like, its not exactly the top priority.

PR #18067, opened 2020-10-29
dev-python/markdown2: python 3.9 support

ZeroPointEnergy:
Add python 3.9 support. All test are passing.


sbraz:
Thanks, merged with some small additions.

PR #18076, opened 2020-10-30
app-admin/hddtemp: Taking over package as proxied-mantainer

PPed72:


PPed72:
I wanted to make a commit for taking over and one for adding the drives to hddgentoo.db, but I (actually repoman) made a mess. Closing and doing it again, better.

PR #18077, opened 2020-10-30
gui-apps/gtkgreet: remove Werror build flag

epsilon-0:
Closes: https://bugs.gentoo.org/750707


gui-apps/gtkgreet/gtkgreet-0.6-r1.ebuild
```diff
@@ -23,8 +22,8 @@ IUSE="+layershell +man"
 
 DEPEND="
 	dev-libs/json-c:=
-	x11-libs/gtk+:3
 	layershell? ( gui-libs/gtk-layer-shell )
+	x11-libs/gtk+:3
```
juippis, line 26: Looks good otherwise, but why switch this position?

PR #18078, opened 2020-10-30
app-admin/hddtemp: Take over package as proxied-mantainer

PPed72:


juippis (commented):
This program looks like abandonware, I don't think proxy-maint will want this / we can do much about it. So don't take on the maintainership. 

We can still merge your addition, but it would require a revbump so everyone gets the update. You can just `git mv` the ebuild into a new revision.

Isn't there some place in upstream to get a fully updated database?

PPed72:
Will do, as you prefer.
As far as I know upstream does not update the db since 2010.

juippis (approved):

PR #18083, opened 2020-10-30
net-im/poezio: set DISTUTILS_USE_SETUPTOOLS

jsmolic:
Closes: https://bugs.gentoo.org/751871

### DIFF
--- a/net-im/poezio/poezio-0.13.1.ebuild
+++ b/net-im/poezio/poezio-0.13.1.ebuild
@@ -3,6 +3,7 @@
 
 EAPI=7
 
+DISTUTILS_USE_SETUPTOOLS=rdepend
 PYTHON_COMPAT=( python3_{7,8} )
 
 inherit distutils-r1 readme.gentoo-r1
--- a/net-im/poezio/poezio-9999.ebuild
+++ b/net-im/poezio/poezio-9999.ebuild
@@ -3,6 +3,7 @@
 
 EAPI=7
 
+DISTUTILS_USE_SETUPTOOLS=rdepend
 PYTHON_COMPAT=( python3_{7,8} )
 
 inherit distutils-r1 readme.gentoo-r1


andrey-utkin:
@jsmolic thanks for the PR.

I must admit I still don't understand this topic (the type of QA warnings about DISTUTILS_USE_SETUPTOOLS) very well.
This article (which you have probably have read by now) made me think this is issue uncomfortably complicated: https://blogs.gentoo.org/mgorny/2020/10/21/distutils_use_setuptools-qa-spam-and-more-qa-spam/

To make up for my cluelessness, could you please specify, in commit message, the package-specific facts which you have established to determine the correct setting?
Does anything laid out in the linked article change your conclusion, or make you want to make any additional comments?
Thanks.

jsmolic:
@andrey-utkin Hi :) Sorry for the delay in responding.\
Well, to explain things, the fact the the package imports `pkg_resources` (https://github.com/poezio/poezio/blob/master/poezio/config.py#L17, https://github.com/poezio/poezio/blob/master/poezio/plugin_manager.py#L14) and that it has `console_scripts` type entry points (https://github.com/poezio/poezio/blob/master/setup.py#L156) suggests that it uses `dev-python/setuptools` as a runtime dependency. (Though thinking about it now, 0.13 version should probably be revbumped in case this is added as it's basically changind `RDEPEND`).\
But yeah, as stated in mgorny's article, setuptools upstream changed the way entry points are used in python 3.8 and above, which means that eventually, after python 3.7 support is dropped, lot of `DISTUTILS_USE_SETUPTOOLS` for packages that only have `console_scripts` entry points will probably require changing as entry points will no longer mean requiring setuptools as runtime dependency. This package is a bit different, it imports `pkg_resources` in 2 `.py` files, but I'm not yet sure whether that is only for entry points usage. Perhaps I can contact upstream for clarification.\
What do you think? I'll mark this PR as draft for now

andrey-utkin:
Thanks a lot for the clarification.
I decided to apply this change (with a revbump as you noted).

andrey-utkin:
Pushed as 3a5bcdbfb205341a69492646f43badab4e543b29, b305a7b76d027daed344bc084d4694ff8ea23ceb

Thanks a lot @jsmolic !

jsmolic:
@andrey-utkin Thanks to you as well :)

PR #18085, opened 2020-10-30
sys-cluster/mpich: various fixes

epsilon-0:
mpi-threads forces threads
fix automagic dependency on libunwind
Bug: https://bugs.gentoo.org/686134
build with romio by default


sys-cluster/mpich/mpich-3.3.ebuild
```diff
@@ -16,16 +16,20 @@ SRC_URI="http://www.mpich.org/static/downloads/${PV}/${P}.tar.gz"
 SLOT="0"
 LICENSE="mpich2"
```
SoapGentoo, line 17: revbump due to RDEPEND change

SoapGentoo (changes requested):

PR #18086, opened 2020-10-31
dev-libs/libtpms: Bump to 0.7.4, remove old

salahcoronya:


juippis (approved):
If you see a CI check return false, please wait a bit, re-sync master branch, rebase master branch and force-push. Because if we're busy, we tend to skip broken PRs until they're fixed.

PR #18088, opened 2020-10-31
dev-python/roman: version bump to 3.3

ZeroPointEnergy:
Version bump to 3.3 with some minor ebuild adjustments:
- DISTUTILS_USE_SETUPTOOLS=rdepend was required
- added python 3.9 support

All tests are passing.


ZeroPointEnergy:
Optimized the ebuild a bit by using "distutils_enable_tests" and removing the unused test useflag. Moving the test source is now in prepare.

PR #18089, opened 2020-10-31
 app-misc/editor-wrapper: Avoid infinite recursion

cmplstofB:
If `$EDITOR`/`$PAGER` variables are identical to 'editor'/'pager' respectively, the script recurs infinitely.

e.g.:
```console
$ EDITOR=editor
$ editor # recurs infinitely
```

The users who set these variables to such names probably not intend to cause this behavior, so it would be better to detect infinite recursion and exit with an error.

### DIFF
--- a/app-misc/editor-wrapper/files/editor-wrapper-4.sh
+++ b/app-misc/editor-wrapper/files/editor-wrapper-4.sh
@@ -12,6 +12,10 @@ fi
 if [ -z "${@VAR@}" ]; then
     echo "$0: The @VAR@ variable must be set" >&2
     exit 1
+elif [ "${@VAR@}" = "${0##*/}" ]; then
+    # avoid infinite recursion
+    echo "$0: The @VAR@ variable shall not be identical to '${0##*/}'" >&2
+    exit 1
 fi
 
 exec ${@VAR@} "$@"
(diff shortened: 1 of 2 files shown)


ulm (changes requested):
Some general comments:

- Don't change the script in place, but make a new version 5, plus a new ebuild (straight to stable is fine, the change looks simple enough)
- Keep lines below 80 characters please
- Don't forget to update the copyright years in line 2

cmplstofB:
@ulm Ok, i see.
Sorry for making you tell me about it.

thesamesam:
Fix within this PR please (then force push), thanks!

ulm:
> ## Missing GCO sign-off
> 
> Please read the terms of [Gentoo Certificate of Origin](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin) and acknowledge them by adding a sign-off to _all_ your commits.

This needs to be addressed too, otherwise we cannot merge these commits.

ulm:
Looking at this again, I think it is unlikely for this problem to occur in practice:
- `editor` isn't one of the values offered by editor.eselect;
- `/usr/libexec` is not meant to be in users' path;
- This leaves `EDITOR` set to `/usr/libexec/editor` as a possibility, which the proposed change won't catch however.

So I suggest to keep things as they are, i.e. I am closing this PR.

PR #18092, opened 2020-10-31
games-fps/gzdoom: version bump to 4.5.0

vilhelmgray:
Closes: https://bugs.gentoo.org/752036


vilhelmgray:
I don't know if the WidePix license allows redistribution. Perhaps the Gentoo license team can evaluate it and make sure Gentoo is permitted to mirror WidePix on its servers.

chewi:
Yeah, I wondered what the license implications would be when I saw the release announcement. It doesn't say you cannot redistribute it and it feels like "You MAY make derivative works based on mine" implies that you can but I'll defer to the license gurus.

vilhelmgray:
IANAL, but from the context I think the author just wanted to make clear that the [DOOM assets EULA](https://gitweb.gentoo.org/repo/gentoo.git/plain/licenses/DOOM-COLLECTORS-EDITION) applies to their derivative artwork. Unfortunately, by creating this license file they have effectively created **Yet Another License :tm:**  that we'll have to keep track of. :wink:

chewi:
@gentoo/licenses We request your assistance!

licenses/WidePix
```diff
@@ -0,0 +1,13 @@
...
+
+LICENSE:
+
+You MAY make derivative works based on mine - HOWEVER:
+
+- You may not sell this work
+
+- Your work must comply with id Software's original licenses (i.e. they must only work on the game from which they originated)
+
+- All derivative works must credit the original authors (id Software (Doom), Raven Software (Hexen/Heretic), Rogue Software (Strife), Digital Café (Chex), Nash Muhandes)
```
ulm, line 13: I don't see anything in that license that would permit redistribution, so mirror and bindist restrictions are asked for.

ulm (commented):
Oh no, not gzdoom again. DUMB and cephes licenses weren't enough? :slightly_smiling_face:

licenses/WidePix
chewi: It might be worth asking upstream to clarify this point as that probably wasn't their intention. I'm not sure whether that's compatible with the original licenses though.

vilhelmgray: I'm contacting upstream to see if they can update the license to be more explicit.

vilhelmgray: Upstream has updated the WidePix license to permit distribution: https://github.com/nashmuhandes/WidePix/commit/92738042ca3a37f28153a09809d80a7d61090532

ulm: Still not a free license because of the "no sell" clause, but at least we won't need mirror and bindist restrictions.

vilhelmgray: Yes, unfortunately GZDoom widescreen graphics will have to remain part of the non-free USE flag. I suppose that's to be expected given the similar "no sell" clause of the parent Doom asset license. :disappointed:

chewi:
Fun fact, the new widescreen images have broken the level I've been making. It has laptops and a couple of them had the title screen showing. Now it's not aligned properly! :laughing:

PR #18095, opened 2020-10-31
media-libs/gegl: workaround to fix build on hardened gentoo systems

band-a-prend:
On gentoo hardened fatal error: opencl/* no such file or directory
takes place due to opencl/meson.build doesn't produce appropriate
headers from *.cl files.

Closes: https://bugs.gentoo.org/739816

### DIFF
--- a/media-libs/gegl/gegl-0.4.26-r1.ebuild
+++ b/media-libs/gegl/gegl-0.4.26-r1.ebuild
@@ -1,0 +1,164 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{6,7,8,9} )
+# vala and introspection support is broken, bug #468208
+VALA_USE_DEPEND=vapigen
+
+inherit meson gnome2-utils python-any-r1 vala
+
+if [[ ${PV} == *9999* ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://gitlab.gnome.org/GNOME/gegl.git"
+	SRC_URI=""
+else
+	SRC_URI="http://download.gimp.org/pub/${PN}/${PV:0:3}/${P}.tar.xz"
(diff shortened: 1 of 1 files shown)


media-libs/gegl/gegl-0.4.26-r1.ebuild
```diff
@@ -0,0 +1,167 @@
...
+		-i tests/compositions/meson.build || die
+
+	# fix skipping mipmap tests due to executable not found
+	for item in "invert-crop.sh" "invert.sh" "rotate-crop.sh" "rotate.sh" "unsharp-crop.sh" "unsharp.sh"; do
+		sed -i "s:/bin/gegl:/bin/gegl-0.4:g" "${S}/tests/mipmap/${item}" || die
+		sed -i "s:/tools/gegl-imgcmp:/tools/gegl-imgcmp-0.4:g" "${S}/tests/mipmap/${item}" || die
+	done
+
+	# fix 'build'headers from *.cl on gentoo-hardened, bug 739816
+	pushd "${S}/opencl/"
```
juippis, line 110: `|| die`

media-libs/gegl/gegl-0.4.26-r1.ebuild
```diff
@@ -0,0 +1,167 @@
...
+	for file in *.cl; do
+		if [ -f "$file" ]; then
+			python cltostring.py $file
```
juippis, line 113: `${EPYTHON}` instead of `python` here?
`${file}`.

media-libs/gegl/gegl-0.4.26-r1.ebuild
```diff
@@ -0,0 +1,167 @@
...
+		fi
+	done
+	popd
```
juippis, line 116: `|| die`.

juippis (commented):
Well I did notice more issues in the ebuild, but decided to only focus on what you're attempting to fix. No need to rewrite the whole ebuild due to this.

media-libs/gegl/gegl-0.4.26-r1.ebuild
band-a-prend: add `${PYTHON_DEPS}` to `BDEPEND`;  
remove `pkg_setup()` as now it doesn't depend on `USE="test"`;  
add `|| die` also after `${file}`

juippis (approved):

PR #18096, opened 2020-11-01
sys-fs/cryfs: remove debug warning on sys-fs/cryfs[-debug]

aristocrates:
My comment on the [bug report](https://bugs.gentoo.org/747376) repeated here:

If I add `-DNDEBUG` in the ebuild using append-flags, it gets rid of the warning for me. I'm not 100% sure this is the most correct and future-proof way to disable the development build, but I see something similar done in dev-util/intel-graphics-compiler:

https://github.com/gentoo/gentoo/blob/a50597d0202a4968147c1b5d96ebeeb1872b0ffe/dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.4111.ebuild#L50-L52

From the discussion in an upstream issue (https://github.com/cryfs/cryfs/issues/147) I'm not sure whether there is much of a known performance impact, but I can confirm that the warning goes away and that the following command:

```
file `which cryfs`
```

gives "stripped" which I assume means it does not contain debugging symbols.

### DIFF
--- a/sys-fs/cryfs/cryfs-0.10.2.ebuild
+++ b/sys-fs/cryfs/cryfs-0.10.2.ebuild
@@ -73,6 +73,7 @@ src_configure() {
 		-DBUILD_TESTING=$(usex test)
 	)
 	use custom-optimization || append-flags -O3
+	use debug || append-flags -DNDEBUG
 
 	cmake_src_configure
 }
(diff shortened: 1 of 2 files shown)


juippis (approved):
```
 * Messages for package sys-fs/cryfs-0.10.2:

 * please use append-cppflags for preprocessor flags
```
Wonder what this means?

PR #18097, opened 2020-11-01
media-gfx/gimp: cleanup ebulds depend on py2, droped dev-python/pygtk

band-a-prend:
As media-gimp-2.10.20 is depend on >=media-libs/babl-0.1.78
and >=media-gegl-0.4.24 therefore early of them are dropped too.

Closes: https://bugs.gentoo.org/711684


juippis (commented):
Can you split this into 3 different commits, per package?

If there are problems after, we'd only have to revert the problematic commit, not bring all of these 10 files back.

PR #18098, opened 2020-11-01
media-gfx/gimp: update live; add 2.99.2 developer preview

band-a-prend:
[Gimp 2.99.2 announcement](https://www.gimp.org/news/2020/11/06/gimp-2-99-2-released/)

1a . media-libs/gegl:
Add `PYTHON_COMPAT=( python3_9 )`  
  

1b . media-gfx/gimp: update live-ebuild, switch to SLOT "0/3"

It's time to switch to SLOT `"0/3"` to follow major gimp API

Add USE "vala" as it's optional is fixed by upstream.

Fix `CLEANFILES must be set with '=' before using '+='` error.
Update dependencies with portage tree.

2 . media-gfx/gimp: 2.99.2 version bump, unstable development preview

This package is non-keyworded as depends on
`>=media-libs/gegl-0.4.27` that isn't yet released, so `gegl-9999(**)` is used to build.
_upd: gegl-0.4.26 could be used too as I was noticed by gimp developers._

Add patch to fix compatibility with upcoming autoconf-2.70.

Bug: https://bugs.gentoo.org/750575


band-a-prend:
It seems I have to switch all ebuilds of gimp-2.10.x to `0/2`  SLOT otherwise there will be 2 slot in portage tree assuming parallel installation while the exucutable, man pages and icons paths in new slot didn't change.

band-a-prend:
Unfortunately I don't know how exactly packages conditionally depending on media-gfx/gimp should work 
so just set `media-gfx/gimp:0/2` in *.ebuilds with revision bump without KEYWORDS drop to unstable for stabilized revisions.
If maintainers of this packages could update something else in *.ebuilds, please do it.

app-misc/emelfm2: @johu @jonasstein  could you please switch 0.9.1-r4 to EAPI-7?

media-gfx/gtkam: @leio @pacho2 
media-gfx/xsane, net-print/gutenprint: @akhuettel @tamiko 
media-gfx/iscan: @gringus 

could you please approve changes?

thesamesam:
We need to just revbump in place (git mv), preserving stable keywords.

band-a-prend:
Update pr with changes in `media-gfx/gmic` (dropped old).  
Add `lua? ( dev-lua/lgi )` as dependency for `media-gfx/gimp-{2.99.2,9999}.ebuild`

thesamesam:
I'm inclined to not give it keywords for now, what do you think?

band-a-prend:
@thesamesam 
Thank you. Yes I agree with non-keywording right now especially because it developer preview.

I would like dial with issue about Lua slotting/eclass in separate pr.

thesamesam:
@band-a-prend Sounds great to me. Thank you for your work!

PR #18100, opened 2020-11-01
dev-python/ansible-lint: bump to 4.3.6

aristocrates:
https://bugs.gentoo.org/717088

Also adds dev-python/rich (version 9.1.0), a dependency of ansible-lint 4.3.6 (see https://bugs.gentoo.org/752075) - I added myself as a proxy maintainer. I adapted https://github.com/gentoo-mirror/nest/blob/master/dev-python/rich/rich-6.0.0.ebuild a bit.

Tests for ansible-lint seem to be broken for me here and for 4.1.0 (in tree currently). dev-python/rich[test] passes for me and the following code block from https://pypi.org/project/rich/ gives me the expected output:

```
from rich import print

print("Hello, [bold magenta]World[/bold magenta]!", ":vampire:", locals())
```

~Adding python3_9 to PYTHON_COMPAT is currently blocked by dev-python/commonmark and dev-python/setuptools_scm_git_archive needing 3_9.~ :heavy_check_mark: https://github.com/gentoo/gentoo/pull/18101

PR #18101, opened 2020-11-01
Add python 3_9 to dev-python/{commonmark,setuptools_scm_git_archive} PYTHON_COMPAT

aristocrates:
Would allow for python3_9 support in app-admin/ansible-lint and dev-python/rich (added in https://github.com/gentoo/gentoo/pull/18100)


mgorny:
LGTM.

PR #18103, opened 2020-11-01
app-crypt/pius: pull in minor fixes and add python3.9 support

aristocrates:
Minor fixes now included:
* https://github.com/jaymzh/pius/commit/6a92664fe0cfacffb03e6f3312c1c5fb4d785297
* https://github.com/jaymzh/pius/commit/f54adce7ba47ad8882441aec3b0583ee788ea8e1
* https://github.com/jaymzh/pius/commit/5f24833c50ac6d831c03309e9bbb6b62224ac0ac
* https://github.com/jaymzh/pius/commit/19b61c3b0dc31ee60341b3ca708f8e459e44ce3f

PR #18104, opened 2020-11-01
app-crypt/pius: add patches that include git headers

aristocrates:
@juippis this replaces the diffs added in https://github.com/gentoo/gentoo/pull/18103 with `git format-patch` exports including the git headers.


juippis (approved):
Thanks!

PR #18107, opened 2020-11-01
dev-libs/opensc: disable "pkcs11-register" autostart

henning-schild:
It potentially breaks users configuration continously on every login.
https://github.com/OpenSC/OpenSC/issues/2060

Instead tell users about that tool so they can decide to run it on
demand.


henning-schild:
In the upsteam issue the maintainer of Fedora/RHEL also concluded that the feature is "conflicting", so they disable it. Do we need a bug or how do we proceed here?

henning-schild:
upstream is ignoring attempts to fix this and even claiming there was no issue, sad but there is nothing sustainable to do here

PR #18108, opened 2020-11-01
net-proxy/mitmproxy: Bump to 5.3.0

00-matt:


dev-python/kaitaistruct/kaitaistruct-0.9.ebuild
```diff
@@ -0,0 +1,18 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
```
juippis, line 4: We always aim to use the latest available EAPI, and it's currently 7.

dev-python/kaitaistruct/kaitaistruct-0.9.ebuild
```diff
@@ -0,0 +1,18 @@
...
+
+DESCRIPTION="Kaitai Struct runtime for Python"
+HOMEPAGE="https://kaitai.io/"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+DEPEND="dev-python/setuptools[${PYTHON_USEDEP}]"
```
juippis, line 18: This line is not needed, *distutils-r1.eclass* will set it for you automatically.

00-matt: Thank you, will remove.

juippis (approved):

PR #18112, opened 2020-11-02
net-dialup/freeradius: Support alternate log path in initscript

doctaweeks:

### DIFF
--- a/net-dialup/freeradius/files/radius.conf-r4
+++ b/net-dialup/freeradius/files/radius.conf-r4
@@ -11,6 +11,8 @@ pidfile=/var/run/radiusd/radiusd.pid
 RADIUSD_USER=radius
 RADIUSD_GROUP=radius
 
+RADIUSD_LOGPATH=/var/log/radius
+
 # If you set up logging to syslog in /etc/raddb/radiusd.conf, you want
 # to uncomment the following line.
 #rc_use="logger"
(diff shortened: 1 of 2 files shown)


geaaru:
Hi, for me it's ok. Thanks.
We need to wait for someone with the merge power. :)

mattst88:
Please bump the version of the `.conf` and `.init` files rather than modifying them in place. Modifying them in place means that the changes aren't tied to the freeradius version.

doctaweeks:
@mattst88 Can you please elaborate on why it's necessary to bump everything and tie this to a version?

This change is not version specific; it is applicable to all versions in the tree. It also doesn't change anything - the default is identical to the existing arrangement. This simply provides a documented way for a user to make checkpath do the right thing if the log path in the freeradius config is changed. *If* this parsed the freeradius config to find the log path, I could understand the revbump requirement.

mattst88:
This changes what is installed on disk, does it not? Whenever the files the package installs are different, by policy we should have a revision bump. See https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html

Take for example what would happen without a revision bump in the case of binary packages. Despite the change existing in the ebuild repo, a binary package built before the commit would install something different.

Does that make sense?

geaaru:
From my side it's sufficient bump a new revision (-rX) without creating a new version of the init/conf files. With the new -rX people get the new version automatically.
My cent.

mattst88:
> From my side it's sufficient bump a new revision (-rX) without creating a new version of the init/conf files. With the new -rX people get the new version automatically.

That's fine if you delete the existing freeradius versions; this PR is not doing that.

PR #18116, opened 2020-11-03
net-libs/usrsctp: added new package

Ri0n:
Closes: https://bugs.gentoo.org/752477
Closes: https://bugs.gentoo.org/752447


net-libs/usrsctp/Manifest
```diff
@@ -0,0 +1 @@
+DIST usrsctp-20201103.tar.gz 770742 BLAKE2B 94c7b0417a076d627065c625726eb214ec58be90d0965748610f63c076f1c172cb114dc8162af5604751cfe5e86ebf5444f002d07e43c94e461e21f6a3a4819a SHA512 a4bb4dd0ec81b2314e8590cfe5386179697e64d82915a0ef7e8f7a0baf67f88e3da89329ea73b38e8c417668f971f6580e120ed5be8588761b4b334a65ab4120
```
juippis, line 1: I see they've made a release, but 4 years ago. Could you name this package as `0.9.3.0_p20201102` (note the date, we use a date when commit was made, not when the ebuild was written).

net-libs/usrsctp/usrsctp-20201103.ebuild
```diff
@@ -0,0 +1,28 @@
+# Copyright 2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake flag-o-matic
```
juippis, line 6: Did you forget to use *flag-o-matic* here? I didn't get any flag errors emerging it though.

juippis (commented):
Thanks for fast reaction. You're missing the required sign-off from your commit.

Could you also update `psi-9999` to depend on this package?

net-libs/usrsctp/usrsctp-20201103.ebuild
Ri0n: yep. seems like someting copy/paste-related.

Ri0n:
> Thanks for fast reaction. You're missing the required sign-off from your commit.

strange `repoman commit` doesn't have an option for this. I'll add manually.

> Could you also update psi-9999 to depend on this package?

sure, with next PR(s).

net-libs/usrsctp/Manifest
Ri0n: will do

juippis:
> 
> > Could you also update psi-9999 to depend on this package?
> 
> sure, with next PR(s).

You can add multiple commits to a single PR. An update to psi would be logical, and be processed faster by us if it was here, as a separate commit.

juippis (approved):

PR #18120, opened 2020-11-03
app-editors/vis: Add lua support

lanodan:
Mostly based on prior work comitted into ::lua repository

Closes: https://bugs.gentoo.org/752540

### DIFF
--- a/app-editors/vis/vis-0.6.ebuild
+++ b/app-editors/vis/vis-0.6.ebuild
@@ -4,6 +4,9 @@
 EAPI=7
 
 MY_PTV=0.4
+LUA_COMPAT=( lua5-2 lua5-3 )
+
+inherit lua-single eutils
 
 DESCRIPTION="modern, legacy free, simple yet efficient vim-like editor"
 HOMEPAGE="https://github.com/martanne/vis"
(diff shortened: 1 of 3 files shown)


app-editors/vis/vis-9999.ebuild
```diff
@@ -2,25 +2,27 @@
...
-# - ::lua package done for using >=dev-lang/lua-5.2
-# which is needed for syntax highlighting and settings but masked in ::gentoo
 DEPEND="dev-libs/libtermkey
 	ncurses? ( sys-libs/ncurses:0= )
+	lua? ( ${LUA_DEPS} )
 	tre? ( dev-libs/tre:= )"
 RDEPEND="${DEPEND}
-	app-eselect/eselect-vi"
+	app-eselect/eselect-vi
+	lua? ( lpeg? ( >=dev-lua/lpeg-0.12 ) )"
```
juippis, line 25: Hmm, this probably needs to be wrapped inside `lua_gen_cond_dep` but not sure whether it works, at all, before it's migrated to use new lua eclasses. Thoughts?

juippis: 1a6dc03ce8e11606e794ea6d187771e18432153c it's done now.

app-editors/vis/vis-0.6.ebuild
```diff
@@ -12,17 +15,17 @@ SRC_URI="https://github.com/martanne/vis/releases/download/v${PV}/${P}.tar.gz
...
-# - ::lua package done for using >=dev-lang/lua-5.2
-# which is needed for syntax highlighting and settings but masked in ::gentoo
 DEPEND="dev-libs/libtermkey
 	ncurses? ( sys-libs/ncurses:0= )
+	lua? ( ${LUA_DEPS} )
 	tre? ( dev-libs/tre:= )"
 RDEPEND="${DEPEND}
-	app-eselect/eselect-vi"
+	app-eselect/eselect-vi
+	lua? ( lpeg? ( $(lua_gen_cond_dep '>=dev-lua/lpeg-0.12[${LUA_USEDEP}]') ) )"
```
juippis, line 28: I'm sorry if I missed this the first time, but how is `lpeg` used in this ebuild? Now that I can see, it doesn't have other function that to control optional runtime dependency which is prohibited in Gentoo. You should use `optfeature` instead, which also takes away this `lua_gen_cond_dep` construct...
https://projects.gentoo.org/qa/policy-guide/dependencies.html#optional-runtime-dependencies

juippis (approved):

PR #18127, opened 2020-11-03
app-admin/conky: make xinerama dependency optional

iphands:
I noticed that conky has a configure option for libXinerama but not a use flag for such.
Adding the ability to build conky without xinerama support.

Created ebuilds for 1.10.8 and 1.11.6 versions.

First PR I was not sure if I should make new files or PR against the exiting files... Probably easier to spot diffs if I did the latter, but the convention seemed like new files was correct.

Signed-off-by: Ian Page Hands <iphands@gmail.com>

### DIFF
--- a/app-admin/conky/conky-1.11.6-r2.ebuild
+++ b/app-admin/conky/conky-1.11.6-r2.ebuild
@@ -1,0 +1,193 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+LUA_COMPAT=( lua5-3 )
+
+inherit cmake linux-info lua-single readme.gentoo-r1 xdg
+
+DESCRIPTION="An advanced, highly configurable system monitor for X"
+HOMEPAGE="https://github.com/brndnmtthws/conky"
+SRC_URI="https://github.com/brndnmtthws/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3 BSD LGPL-2.1 MIT"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~ppc ~ppc64 ~sparc ~x86"
+IUSE="apcupsd bundled-toluapp cmus curl doc hddtemp ical iconv imlib iostats
+	ipv6 irc lua-cairo lua-imlib lua-rsvg math moc mpd mysql nano-syntax
+	ncurses nvidia +portmon pulseaudio rss systemd thinkpad truetype
+	vim-syntax weather-metar webserver wifi X xinerama xmms2"
+
+COMMON_DEPEND="
+	X? (
+		x11-libs/libX11
+		x11-libs/libXdamage
+		x11-libs/libXfixes
+		x11-libs/libXext
+	)
+	cmus? ( media-sound/cmus )
+	curl? ( net-misc/curl )
+	ical? ( dev-libs/libical:= )
+	iconv? ( virtual/libiconv )
+	imlib? ( media-libs/imlib2[X] )
+	irc? ( net-libs/libircclient )
+	lua-cairo? ( x11-libs/cairo[X] )
+	lua-imlib? ( media-libs/imlib2[X] )
+	lua-rsvg? ( gnome-base/librsvg )
+	mysql? ( dev-db/mysql-connector-c )
+	ncurses? ( sys-libs/ncurses:= )
+	nvidia? ( x11-drivers/nvidia-drivers[tools,static-libs] )
+	pulseaudio? ( media-sound/pulseaudio )
+	rss? ( dev-libs/libxml2 net-misc/curl dev-libs/glib:2 )
+	systemd? ( sys-apps/systemd )
+	truetype? ( x11-libs/libXft >=media-libs/freetype-2 )
+	wifi? ( net-wireless/wireless-tools )
+	weather-metar? ( net-misc/curl )
+	webserver? ( net-libs/libmicrohttpd )
+	xinerama? ( x11-libs/libXinerama )
+	xmms2? ( media-sound/xmms2 )
+	${LUA_DEPS}
+"
+RDEPEND="
+	${COMMON_DEPEND}
+	apcupsd? ( sys-power/apcupsd )
+	hddtemp? ( app-admin/hddtemp )
+	moc? ( media-sound/moc )
+	nano-syntax? ( app-editors/nano )
+	vim-syntax? ( || ( app-editors/vim app-editors/gvim ) )
+"
+DEPEND="
+	${COMMON_DEPEND}
+	doc? ( app-text/docbook2X dev-libs/libxslt )
+"
+
+REQUIRED_USE="
+	imlib? ( X )
+	nvidia? ( X )
+	truetype? ( X )
+	lua-cairo? ( X  bundled-toluapp )
+	lua-imlib? ( X  bundled-toluapp )
+	lua-rsvg? ( X  bundled-toluapp )
+	xinerama? ( X )
+"
+
+CONFIG_CHECK="~IPV6"
+
+DOCS=( README.md AUTHORS )
+
+PATCHES=(
+	"${FILESDIR}"/${P}-fpermissive.patch
+)
+
+DISABLE_AUTOFORMATTING="yes"
+DOC_CONTENTS="You can find sample configurations at ${ROOT}/usr/share/doc/${PF}.
+To customize, copy to \${XDG_CONFIG_HOME}/conky/conky.conf
+and edit it to your liking.
+
+There are pretty html docs available at the conky homepage
+or in ${ROOT}/usr/share/doc/${PF}/html when built with USE=doc.
+
+Also see https://wiki.gentoo.org/wiki/Conky/HOWTO"
+
+pkg_setup() {
+	use ipv6 && linux-info_pkg_setup
+
+	lua-single_pkg_setup
+}
+
+src_prepare() {
+	cmake_src_prepare
+
+	xdg_environment_reset
+
+	sed -i -e "s|find_program(APP_MAN man)|set(APP_MAN $(which man) CACHE FILEPATH MAN_BINARY)|" \
+		cmake/ConkyPlatformChecks.cmake || die
+}
+
+src_configure() {
+	local mycmakeargs
+
+	if use X; then
+		mycmakeargs=(
+			-DBUILD_ARGB=ON
+			-DBUILD_X11=ON
+			-DBUILD_XDAMAGE=ON
+			-DBUILD_XDBE=ON
+			-DBUILD_XSHAPE=ON
+			-DOWN_WINDOW=ON
+		)
+	else
+		mycmakeargs=(
+			-DBUILD_X11=OFF
+		)
+	fi
+
+	mycmakeargs+=(
+		-DBUILD_APCUPSD=$(usex apcupsd)
+		-DBUILD_AUDACIOUS=OFF
+		-DBUILD_BUILTIN_CONFIG=ON
+		-DBUILD_CMUS=$(usex cmus)
+		-DBUILD_CURL=$(usex curl)
+		-DBUILD_DOCS=$(usex doc)
+		-DBUILD_HDDTEMP=$(usex hddtemp)
+		-DBUILD_HTTP=$(usex webserver)
+		-DBUILD_I18N=ON
+		-DBUILD_IBM=$(usex thinkpad)
+		-DBUILD_ICAL=$(usex ical)
+		-DBUILD_ICONV=$(usex iconv)
+		-DBUILD_IMLIB2=$(usex imlib)
+		-DBUILD_IOSTATS=$(usex iostats)
+		-DBUILD_IPV6=$(usex ipv6)
+		-DBUILD_IRC=$(usex irc)
+		-DBUILD_JOURNAL=$(usex systemd)
+		-DBUILD_LUA_CAIRO=$(usex lua-cairo)
+		-DBUILD_LUA_IMLIB2=$(usex lua-imlib)
+		-DBUILD_LUA_RSVG=$(usex lua-rsvg)
+		-DBUILD_MATH=$(usex math)
+		-DBUILD_MOC=$(usex moc)
+		-DBUILD_MPD=$(usex mpd)
+		-DBUILD_MYSQL=$(usex mysql)
+		-DBUILD_NCURSES=$(usex ncurses)
+		-DBUILD_NVIDIA=$(usex nvidia)
+		-DBUILD_OLD_CONFIG=ON
+		-DBUILD_PORT_MONITORS=$(usex portmon)
+		-DBUILD_PULSEAUDIO=$(usex pulseaudio)
+		-DBUILD_RSS=$(usex rss)
+		-DBUILD_WEATHER_METAR=$(usex weather-metar)
+		-DBUILD_WLAN=$(usex wifi)
+		-DBUILD_XFT=$(usex truetype)
+		-DBUILD_XINERAMA=$(usex xinerama)
+		-DBUILD_XMMS2=$(usex xmms2)
+		-DDOC_PATH=/usr/share/doc/${PF}
+		-DMAINTAINER_MODE=OFF
+		-DRELEASE=ON
+	)
+
+	cmake_src_configure
+}
+
+src_install() {
+	cmake_src_install
+
+	if use vim-syntax; then
+		insinto /usr/share/vim/vimfiles/ftdetect
+		doins "${S}"/extras/vim/ftdetect/conkyrc.vim
+
+		insinto /usr/share/vim/vimfiles/syntax
+		doins "${S}"/extras/vim/syntax/conkyrc.vim
+	fi
+
+	if use nano-syntax; then
+		insinto /usr/share/nano/
+		doins "${S}"/extras/nano/conky.nanorc
+	fi
+
+	readme.gentoo_create_doc
+}
+
+pkg_postinst() {
+	readme.gentoo_print_elog
+
+	xdg_pkg_postinst
+}


billie80:
> I noticed that conky has a configure option for libXinerama but not a use flag for such.
> Adding the ability to build conky without xinerama support.
> 
> Created ebuilds for 1.10.8 and 1.11.6 versions.
> 
> First PR I was not sure if I should make new files or PR against the exiting files... Probably easier to spot diffs if I did the latter, but the convention seemed like new files was correct.

Thank you for your contribution.

I am aware of the xinerama configure flag and deliberately decided against making it a USE flag. If we go this route we also need add a USE flag for x11-libs/libXdamage and x11-libs/libXfixes, x11-libs/libXext.

Do we really need that fine grained dependencies?

iphands:
> > I noticed that conky has a configure option for libXinerama but not a use flag for such.
> > Adding the ability to build conky without xinerama support.
> > Created ebuilds for 1.10.8 and 1.11.6 versions.
> > First PR I was not sure if I should make new files or PR against the exiting files... Probably easier to spot diffs if I did the latter, but the convention seemed like new files was correct.
> 
> Thank you for your contribution.
> 
> I am aware of the xinerama configure flag and deliberately decided against making it a USE flag. If we go this route we also need add a USE flag for x11-libs/libXdamage and x11-libs/libXfixes, x11-libs/libXext.

Do we? like technically?

> Do we really need that fine grained dependencies?

I was specifically trying to not pull in x11-libs/libXinerama  libXinerama... this was how I accomplished that.

billie80:
Well, adding USE flags for the other dependencies would also avoid pulling them in, same as for libXimerama!

iphands:
> Well, adding USE flags for the other dependencies would also avoid pulling them in, same as for libXinerama!

So no not a technical reason.

I mean if someone wants to -xdamage -xfixes why not let them?
I can update the ebuild to support the list if you prefer.

Other packages do indeed do this:
https://github.com/gentoo/gentoo/blob/master/x11-misc/x11vnc/x11vnc-0.9.16-r3.ebuild#L28

iphands:
Also do you prefer
```
		-DBUILD_XINERAMA=$(usex xinerama ON OFF)
		-DBUILD_XEXT=$(usex xext ON OFF)
		-DBUILD_XFIXES=$(usex xfixes ON OFF)
```

As opposed to
```
		-DBUILD_XINERAMA=$(usex xinerama)
		-DBUILD_XEXT=$(usex xext)
		-DBUILD_XFIXES=$(usex xfixes)
```

I think the default yes/no might work just fine, but I looks odd to me in the output.

iphands:
Can do this if you are cool with it
```diff
diff --git a/app-admin/conky/conky-1.10.8-r11.ebuild b/app-admin/conky/conky-1.10.8-r11.ebuild
index 656f330..0c33675 100644
--- a/app-admin/conky/conky-1.10.8-r11.ebuild
+++ b/app-admin/conky/conky-1.10.8-r11.ebuild
@@ -15,7 +15,7 @@ KEYWORDS="~alpha ~amd64 ~arm ~ppc ~ppc64 ~sparc ~x86"
 IUSE="apcupsd cmus curl eve hddtemp ical iconv imlib iostats ipv6 irc
 	lua-cairo lua-imlib lua-rsvg math moc mpd mysql nano-syntax ncurses
 	nvidia +portmon pulseaudio rss systemd thinkpad truetype vim-syntax
-	weather-metar webserver wifi X xmms2 xinerama"
+	weather-metar webserver wifi X xmms2 xdamage xext xfixes xinerama"
 
 COMMON_DEPEND="
 	X? (
@@ -26,9 +26,6 @@ COMMON_DEPEND="
 		nvidia? ( x11-drivers/nvidia-drivers[tools,static-libs] )
 		truetype? ( x11-libs/libXft >=media-libs/freetype-2 )
 		x11-libs/libX11
-		x11-libs/libXdamage
-		x11-libs/libXfixes
-		x11-libs/libXext
 		xmms2? ( media-sound/xmms2 )
 	)
 	cmus? ( media-sound/cmus )
@@ -45,7 +42,10 @@ COMMON_DEPEND="
 	wifi? ( net-wireless/wireless-tools )
 	weather-metar? ( net-misc/curl )
 	webserver? ( net-libs/libmicrohttpd )
-		xinerama? ( x11-libs/libXinerama )
+	xdamage? ( x11-libs/libXdamage )
+	xext? ( x11-libs/libXext )
+	xfixes? ( x11-libs/libXfixes )
+	xinerama? ( x11-libs/libXinerama )
 	|| ( dev-lang/lua:0 dev-lang/lua:5.1 )
 "
 RDEPEND="
@@ -62,6 +62,9 @@ DEPEND="
 "
 
 REQUIRED_USE="
+		xdamage? ( X )
+		xext? ( X )
+		xfixes? ( X )
 		xinerama? ( X )
 "
 
@@ -106,17 +109,16 @@ src_configure() {
 		mycmakeargs=(
 			-DBUILD_X11=ON
 			-DOWN_WINDOW=ON
-			-DBUILD_XDAMAGE=ON
 			-DBUILD_XDBE=ON
-			-DBUILD_XFT=$(usex truetype)
-			-DBUILD_IMLIB2=$(usex imlib)
+			-DBUILD_XFT=$(usex truetype ON OFF)
+			-DBUILD_IMLIB2=$(usex imlib ON OFF)
 			-DBUILD_XSHAPE=ON
 			-DBUILD_ARGB=ON
-			-DBUILD_LUA_CAIRO=$(usex lua-cairo)
-			-DBUILD_LUA_IMLIB2=$(usex lua-imlib)
-			-DBUILD_LUA_RSVG=$(usex lua-rsvg)
-			-DBUILD_NVIDIA=$(usex nvidia)
-			-DBUILD_XMMS2=$(usex xmms2)
+			-DBUILD_LUA_CAIRO=$(usex lua-cairo ON OFF)
+			-DBUILD_LUA_IMLIB2=$(usex lua-imlib ON OFF)
+			-DBUILD_LUA_RSVG=$(usex lua-rsvg ON OFF)
+			-DBUILD_NVIDIA=$(usex nvidia ON OFF)
+			-DBUILD_XMMS2=$(usex xmms2 ON OFF)
 		)
 	else
 		mycmakeargs=(
@@ -129,40 +131,33 @@ src_configure() {
 		)
 	fi
 
-		if use xinerama; then
-			mycmakeargs=(
-			-DBUILD_XINERAMA=ON
-		)
-		else
-				mycmakeargs=(
-						-DBUILD_XINERAMA=OFF
-				)
-		fi
-
 	mycmakeargs+=(
-		-DBUILD_APCUPSD=$(usex apcupsd)
-		-DBUILD_CMUS=$(usex cmus)
-		-DBUILD_CURL=$(usex curl)
-		-DBUILD_EVE=$(usex eve)
-		-DBUILD_HDDTEMP=$(usex hddtemp)
-		-DBUILD_IOSTATS=$(usex iostats)
-		-DBUILD_ICAL=$(usex ical)
-		-DBUILD_ICONV=$(usex iconv)
-		-DBUILD_IPV6=$(usex ipv6)
-		-DBUILD_IRC=$(usex irc)
-		-DBUILD_MATH=$(usex math)
-		-DBUILD_MOC=$(usex moc)
-		-DBUILD_MPD=$(usex mpd)
-		-DBUILD_MYSQL=$(usex mysql)
-		-DBUILD_NCURSES=$(usex ncurses)
-		-DBUILD_PORT_MONITORS=$(usex portmon)
-		-DBUILD_PULSEAUDIO=$(usex pulseaudio)
-		-DBUILD_RSS=$(usex rss)
-		-DBUILD_JOURNAL=$(usex systemd)
-		-DBUILD_IBM=$(usex thinkpad)
-		-DBUILD_HTTP=$(usex webserver)
-		-DBUILD_WEATHER_METAR=$(usex weather-metar)
-		-DBUILD_WLAN=$(usex wifi)
+		-DBUILD_APCUPSD=$(usex apcupsd ON OFF)
+		-DBUILD_CMUS=$(usex cmus ON OFF)
+		-DBUILD_CURL=$(usex curl ON OFF)
+		-DBUILD_EVE=$(usex eve ON OFF)
+		-DBUILD_HDDTEMP=$(usex hddtemp ON OFF)
+		-DBUILD_IOSTATS=$(usex iostats ON OFF)
+		-DBUILD_ICAL=$(usex ical ON OFF)
+		-DBUILD_ICONV=$(usex iconv ON OFF)
+		-DBUILD_IPV6=$(usex ipv6 ON OFF)
+		-DBUILD_IRC=$(usex irc ON OFF)
+		-DBUILD_MATH=$(usex math ON OFF)
+		-DBUILD_MOC=$(usex moc ON OFF)
+		-DBUILD_MPD=$(usex mpd ON OFF)
+		-DBUILD_MYSQL=$(usex mysql ON OFF)
+		-DBUILD_NCURSES=$(usex ncurses ON OFF)
+		-DBUILD_PORT_MONITORS=$(usex portmon ON OFF)
+		-DBUILD_PULSEAUDIO=$(usex pulseaudio ON OFF)
+		-DBUILD_RSS=$(usex rss ON OFF)
+		-DBUILD_JOURNAL=$(usex systemd ON OFF)
+		-DBUILD_IBM=$(usex thinkpad ON OFF)
+		-DBUILD_HTTP=$(usex webserver ON OFF)
+		-DBUILD_WEATHER_METAR=$(usex weather-metar ON OFF)
+		-DBUILD_WLAN=$(usex wifi ON OFF)
+		-DBUILD_XINERAMA=$(usex xinerama ON OFF)
+		-DBUILD_XEXT=$(usex xext ON OFF)
+		-DBUILD_XFIXES=$(usex xfixes ON OFF)
 		-DBUILD_BUILTIN_CONFIG=ON
 		-DBUILD_OLD_CONFIG=ON
 		-DBUILD_I18N=ON
```

billie80:
Very hard for me to compare. Please update the pull request but please leave the default yes/no for usex!

iphands:
> Very hard for me to compare. Please update the pull request but please leave the default yes/no for usex!

Sure thing

iphands:
So apparently repoman does not like:
`unknown USE flags: 'xdamage', 'xext', 'xfixes'`

xinerama alone makes repoman happy... is it in fact wise to include those too?
Or should we reverse it to limit it back to just xinerama?

iphands:
Ahh I see I can add to `metadata.xml` to appease repoman

billie80:
Yes they need to be added to metadata.xml as they are probably local use flags!

billie80:
And please squash the commits into a single one!

iphands:
Yeah I'll squash once it looks good.
You can also squash on merge eh? GitHub lets you do that I think.

If not lemme know when you ack I'll squash.

iphands:
@billie80 hey can you help me out with this `UnnecessaryManifest` error Im not sure how to resolve it, and running `repoman full` locally doesnt even throw this error for me.

billie80:
Just remove the Manifest!?

billie80:
Also you have to check your changes regarding dependencies and build options. There is something amiss! I also would recommend to apply the changes only for 1.11.6 for simplicity. I think this version will be unmasked soon as slotted lua is on it`s way.

iphands:
> Also you have to check your changes regarding dependencies and build options. There is something amiss!

Can you comment on the change using the inline comment / review feature?

app-admin/conky/conky-1.10.8-r11.ebuild
```diff
@@ -0,0 +1,194 @@
...
+HOMEPAGE="https://github.com/brndnmtthws/conky"
+SRC_URI="https://github.com/brndnmtthws/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3 BSD LGPL-2.1 MIT"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~ppc ~ppc64 ~sparc ~x86"
+IUSE="apcupsd cmus curl eve hddtemp ical iconv imlib iostats ipv6 irc
+	lua-cairo lua-imlib lua-rsvg math moc mpd mysql nano-syntax ncurses
+	nvidia +portmon pulseaudio rss systemd thinkpad truetype vim-syntax
+	weather-metar webserver wifi X xmms2 xdamage xext xfixes xinerama"
```
billie80, line 18: Sort alphabetically.
From reading the source code (cmake/ConkyPlatformChecks.cmake) a use flag for xfixes makes no sense as it is required wit xdamage. Also there is no option like BUILD_XFIXES. Did you test your changes?
Instead a USE flag for xshape does make sense.

app-admin/conky/conky-1.10.8-r11.ebuild
```diff
@@ -0,0 +1,194 @@
...
+	iconv? ( virtual/libiconv )
+	irc? ( net-libs/libircclient )
+	mysql? ( dev-db/mysql-connector-c )
+	ncurses? ( sys-libs/ncurses:= )
+	pulseaudio? ( media-sound/pulseaudio )
+	rss? ( dev-libs/libxml2 net-misc/curl dev-libs/glib:2 )
+	systemd? ( sys-apps/systemd )
+	wifi? ( net-wireless/wireless-tools )
+	weather-metar? ( net-misc/curl )
+	webserver? ( net-libs/libmicrohttpd ) || ( dev-lang/lua:0 dev-lang/lua:5.1 )
```
billie80, line 44: Why did you remove the line break between the net-libs/libmicrohttpd and the dev-lang/lua dependency?

app-admin/conky/conky-1.10.8-r11.ebuild
```diff
@@ -0,0 +1,194 @@
...
+}
+
+src_configure() {
+	local mycmakeargs
+
+	if use X; then
+		mycmakeargs=(
+			-DBUILD_X11=ON
+			-DOWN_WINDOW=ON
+			-DBUILD_XDBE=ON
```
billie80, line 111: xdbe is the flag which should trigger the libXext dependency similar to xshape.

app-admin/conky/conky-1.10.8-r11.ebuild
```diff
@@ -0,0 +1,194 @@
...
+		-DBUILD_PULSEAUDIO=$(usex pulseaudio)
+		-DBUILD_RSS=$(usex rss)
+		-DBUILD_JOURNAL=$(usex systemd)
+		-DBUILD_IBM=$(usex thinkpad)
+		-DBUILD_HTTP=$(usex webserver)
+		-DBUILD_WEATHER_METAR=$(usex weather-metar)
+		-DBUILD_WLAN=$(usex wifi)
+		-DBUILD_XINERAMA=$(usex xinerama)
+		-DBUILD_XEXT=$(usex xext)
+		-DBUILD_XFIXES=$(usex xfixes)
```
billie80, line 159: As mentioned above DBUILD_XFIXES does not exist.

Also where is DBUILD_XDAMAGE?

app-admin/conky/conky-1.11.6-r2.ebuild
```diff
@@ -0,0 +1,198 @@
...
+HOMEPAGE="https://github.com/brndnmtthws/conky"
+SRC_URI="https://github.com/brndnmtthws/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3 BSD LGPL-2.1 MIT"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~ppc ~ppc64 ~sparc ~x86"
+IUSE="apcupsd bundled-toluapp cmus curl doc hddtemp ical iconv imlib iostats
+	ipv6 irc lua-cairo lua-imlib lua-rsvg math moc mpd mysql nano-syntax
+	ncurses nvidia +portmon pulseaudio rss systemd thinkpad truetype
+	vim-syntax weather-metar webserver wifi X xmms2 xdamage xext xfixes xinerama"
```
billie80, line 20: Same as for conky-1.10.8-r11

app-admin/conky/conky-1.11.6-r2.ebuild
```diff
@@ -0,0 +1,198 @@
...
+
+COMMON_DEPEND="
+	X? (
+		x11-libs/libX11
+		x11-libs/libXdamage
+		x11-libs/libXfixes
+		x11-libs/libXext
```
billie80, line 27: I thought we wanted to get rid of this dependencies via USE flags?

app-admin/conky/conky-1.11.6-r2.ebuild
```diff
@@ -0,0 +1,198 @@
...
+src_configure() {
+	local mycmakeargs
+
+	if use X; then
+		mycmakeargs=(
+			-DBUILD_ARGB=ON
+			-DBUILD_X11=ON
+			-DBUILD_XDAMAGE=ON
+			-DBUILD_XDBE=ON
+			-DBUILD_XSHAPE=ON
```
billie80, line 123: Why is this still here?

app-admin/conky/conky-1.11.6-r2.ebuild
```diff
@@ -0,0 +1,198 @@
...
+		-DBUILD_PULSEAUDIO=$(usex pulseaudio)
+		-DBUILD_RSS=$(usex rss)
+		-DBUILD_WEATHER_METAR=$(usex weather-metar)
+		-DBUILD_WLAN=$(usex wifi)
+		-DBUILD_XFT=$(usex truetype)
+		-DBUILD_XMMS2=$(usex xmms2)
+		-DDOC_PATH=/usr/share/doc/${PF}
+		-DMAINTAINER_MODE=OFF
+		-DRELEASE=ON
+	)
```
billie80, line 170: And missing here?

app-admin/conky/metadata.xml
```diff
@@ -34,6 +34,9 @@
 		<flag name="thinkpad">Enable support for IBM/Lenovo notebooks</flag>
 		<flag name="weather-metar">Enable support for metar weather service</flag>
 		<flag name="webserver">Enable support to act as a webserver serving conkys output using <pkg>net-libs/libmicrohttpd</pkg></flag>
+		<flag name="xdamage">Enable XDamage support <pkg>x11-libs/libXdamage</pkg></flag>
+		<flag name="xext">Enable Xext support <pkg>x11-libs/libXext</pkg></flag>
```
billie80, line 38: Should be xdbe!

app-admin/conky/metadata.xml
```diff
@@ -34,6 +34,9 @@
...
+		<flag name="xfixes">Enable Xfixes support <pkg>x11-libs/libXfixes</pkg></flag>
```
billie80, line 39: Part of xdamage!

app-admin/conky/metadata.xml
```diff
@@ -34,6 +34,9 @@
...
 		<flag name="xmms2">Enable monitoring of music played by <pkg>media-sound/xmms2</pkg></flag>
 	</use>
```
billie80, line 41: Add xshape!

billie80:
This is what I would do, but due to the review I have basically done it already ;-)

```diff
diff -u conky-1.11.6-r1.ebuild conky-1.11.6-r2.ebuild
--- conky-1.11.6-r1.ebuild	2020-11-02 08:11:40.009944350 +0100
+++ conky-1.11.6-r2.ebuild	2020-11-06 11:38:04.234862197 +0100
@@ -17,16 +17,10 @@
 IUSE="apcupsd bundled-toluapp cmus curl doc hddtemp ical iconv imlib iostats
 	ipv6 irc lua-cairo lua-imlib lua-rsvg math moc mpd mysql nano-syntax
 	ncurses nvidia +portmon pulseaudio rss systemd thinkpad truetype
-	vim-syntax weather-metar webserver wifi X xmms2"
+	vim-syntax weather-metar webserver wifi X xdamage xdbe xinerama xmms2
+	xshape"
 
 COMMON_DEPEND="
-	X? (
-		x11-libs/libX11
-		x11-libs/libXdamage
-		x11-libs/libXinerama
-		x11-libs/libXfixes
-		x11-libs/libXext
-	)
 	cmus? ( media-sound/cmus )
 	curl? ( net-misc/curl )
 	ical? ( dev-libs/libical:= )
@@ -46,7 +40,12 @@
 	wifi? ( net-wireless/wireless-tools )
 	weather-metar? ( net-misc/curl )
 	webserver? ( net-libs/libmicrohttpd )
+	X? ( x11-libs/libX11 )
+	xdamage? ( x11-libs/libXdamage x11-libs/libXfixes )
+	xdbe? ( x11-libs/libXext )
+	xinerama? ( x11-libs/libXinerama )
 	xmms2? ( media-sound/xmms2 )
+	xshape? ( x11-libs/libXext )
 	${LUA_DEPS}
 "
 RDEPEND="
@@ -64,11 +63,15 @@
 
 REQUIRED_USE="
 	imlib? ( X )
-	nvidia? ( X )
-	truetype? ( X )
 	lua-cairo? ( X  bundled-toluapp )
 	lua-imlib? ( X  bundled-toluapp )
 	lua-rsvg? ( X  bundled-toluapp )
+	nvidia? ( X )
+	truetype? ( X )
+	xdamage? ( X )
+	xdbe? ( X )
+	xinerama? ( X )
+	xshape? ( X )
 "
 
 CONFIG_CHECK="~IPV6"
@@ -111,10 +114,6 @@
 		mycmakeargs=(
 			-DBUILD_ARGB=ON
 			-DBUILD_X11=ON
-			-DBUILD_XDAMAGE=ON
-			-DBUILD_XDBE=ON
-			-DBUILD_XINERAMA=ON
-			-DBUILD_XSHAPE=ON
 			-DOWN_WINDOW=ON
 		)
 	else
@@ -156,8 +155,12 @@
 		-DBUILD_RSS=$(usex rss)
 		-DBUILD_WEATHER_METAR=$(usex weather-metar)
 		-DBUILD_WLAN=$(usex wifi)
+		-DBUILD_XDAMAGE=$(usex xdamage)
+		-DBUILD_XDBE=$(usex xdbe)
 		-DBUILD_XFT=$(usex truetype)
+		-DBUILD_XINERAMA=$(usex xinerama)
 		-DBUILD_XMMS2=$(usex xmms2)
+		-DBUILD_XSHAPE=$(usex xshape)
 		-DDOC_PATH=/usr/share/doc/${PF}
 		-DMAINTAINER_MODE=OFF
 		-DRELEASE=ON
```

```diff
--- metadata_old.xml	2019-07-06 11:39:42.000000000 +0200
+++ metadata.xml	2020-11-06 11:40:02.442859816 +0100
@@ -34,6 +34,9 @@
 		<flag name="thinkpad">Enable support for IBM/Lenovo notebooks</flag>
 		<flag name="weather-metar">Enable support for metar weather service</flag>
 		<flag name="webserver">Enable support to act as a webserver serving conkys output using <pkg>net-libs/libmicrohttpd</pkg></flag>
+		<flag name="xdamage">Enable XDamage support <pkg>x11-libs/libXdamage</pkg></flag>
+		<flag name="xdbe">Enable Xdbe (double-buffer) support using <pkg>x11-libs/libXext</pkg></flag>
 		<flag name="xmms2">Enable monitoring of music played by <pkg>media-sound/xmms2</pkg></flag>
+		<flag name="xshape">Enable Xshape support using <pkg>x11-libs/libXext</pkg></flag>
 	</use>
 </pkgmetadata>
```

iphands:
I'll change the PR
```
+		<flag name="xdamage">Enable XDamage support <pkg>x11-libs/libXdamage</pkg></flag>
+		<flag name="xdbe">Enable Xdbe (double-buffer) support using <pkg>x11-libs/libXext</pkg></flag>
 		<flag name="xmms2">Enable monitoring of music played by <pkg>media-sound/xmms2</pkg></flag>
+		<flag name="xshape">Enable Xshape support using <pkg>x11-libs/libXext</pkg></flag>
```

The things in <pkg> still need to be updated right?They all see

iphands:
Oh wait maybe that is intentional actuall.

billie80:
> I'll change the PR
> 
> ```
> +		<flag name="xdamage">Enable XDamage support <pkg>x11-libs/libXdamage</pkg></flag>
> +		<flag name="xdbe">Enable Xdbe (double-buffer) support using <pkg>x11-libs/libXext</pkg></flag>
>  		<flag name="xmms2">Enable monitoring of music played by <pkg>media-sound/xmms2</pkg></flag>
> +		<flag name="xshape">Enable Xshape support using <pkg>x11-libs/libXext</pkg></flag>
> ```
> 
> The things in still need to be updated right?
> They all seem to all be pointing to the same package. That is incorrect eh?

Take a look at the code!

xshape includes:
#include <X11/extensions/shape.h>
which belongs to x11-libs/libXext

xdbe includes:
#include <X11/extensions/Xdbe.h>
which belongs to x11-libs/libXext

but we should mention xdamage and xfixes with the xdamage use flag. I corrected this above!

mattst88:
As the X11 maintainer, I'd advise against adding USE flags for xdamage, xfixes, xext. Xinerama is not likely to be used, so a USE flag makes sense here, but not for the others unless you can completely disable X support (via USE=X for example).

billie80:
Hey @mattst88 as I mentioned above I originally decided against individual USE flags. Currently everything is pulled in or not with USE="X". I don't see much gain in just getting rid of there small libraries as a dependency, libX11 requires less than 3 MiB on disk and the other four are less than 250 KiB.

iphands:
> Hey @mattst88 as I mentioned above I originally decided against individual USE flags. Currently

This is kind of false in this pr though. I originally opened the PR only to be able to exclude xinerama.

You mentioned that we should also add used flags for x damage and others if we are going to add flags for xinerama. I sort of disagree with that but you're the maintainer and I am not.

in any case right now it seems that adding those other packages that were your suggestion was too much.

Should we or shouldn't we limit this xinerama?

iphands:
Fwiw my only real motivation here is getting rid of the libxinerama package. I have a single 4K monitor I don't see what the point of pulling that package in is for me.

Xinerama is already a globally supported use flag. Other packages respect my decision to `-xinerama`. I'm not sure why conky should be any different.

I also don't think that we should necessarily pull in all of the other individual options just because we pull the xinerama. I would let a user who cares open a bug about that if they wanted such a feature.

mattst88:
Right, I'd advise you to add IUSE=xinerama (if you want to), but to not make the others optional outside of USE=X. It's kind of non-sense in a bunch of cases. E.g., libXdamage depends on libXfixes...

billie80:
TL;DR as there there is already a global USE flag we can add it also for conky!
@iphands Sorry, I did not meant that it was your suggestion to add USE flags for all dependencies. It was mine, everybody can read this! With originally I meant the state before your pull request, namely only one flag which is USE=X
Also if you really want to get rid of libXinerama we would also need to add this option for all the other packages in the tree pulling it in unconditionally. Maybe you can get rid of it with a USE flag for conky but I for instance can't. Probably some of the packages even don't support a USE flag. But if there should be a gain for all users all packages have to support it. If there is even one hardcoding it then the dependency will stay on systems using that package.
@mattst88 The way it is now is upstream default. So I thought if we deviate from it we should support all build options. The USE flag xdamage would of course pull in libXdamage and libXfixes. The xshape and xdbe flags would pull in libXext.
I still think however there is not much gain even for libXinerama.

iphands:
> @iphands Sorry, I did not meant that it was your suggestion to add USE flags for all dependencies. It was mine, everybody can read this! With originally I meant the state before your pull request, namely only one flag which is USE=X

No worries.

> Also if you really want to get rid of libXinerama we would also need to add this option for all the other packages in the tree pulling it in unconditionally. Maybe you can get rid of it with a USE flag for conky but I for instance can't. Probably some of the packages even don't support a USE flag. But if there should be a gain for all users all packages have to support it. If there is even one hardcoding it then the dependency will stay on systems using that package.

FWIW I found on my system that this was the only package that was pulling in xinerama against my wishes.
Im sure there might in fact be others... but for me this was it.

Is there a good deterministic way to locate the offending packages? Just looking at the ebuild is not enough right?
Because the upstream might not even support ifdefing / configuring the dependency away. eh?


If you are okay with it.. I can propose a new patch that focuses only on conky and only adding the xinerama flag.

billie80:
> FWIW I found on my system that this was the only package that was pulling in xinerama against my wishes.
> Im sure there might in fact be others... but for me this was it.
> 
> Is there a good deterministic way to locate the offending packages? Just looking at the ebuild is not enough right?
> Because the upstream might not even support ifdefing / configuring the dependency away. eh?

Theoretically you have to locate all packages which unconditionally depend on libXinerama. Maybe there are even some depending on it but don't express this with a dependency but this is another issue. Then you have to check the source code if there is already an option to enable/disable it conditionally. If there is not maybe the upstream maintainer is willing to add such an option. If not the last packages have to be patched ourselves.

However I don't think it is worth the effort for the same reason I was initially against adding it to conky!

> If you are okay with it.. I can propose a new patch that focuses only on conky and only adding the xinerama flag.

Sure go ahead then!

billie80:
@iphands something went wrong here. Now I see 787 commits!

mattst88:
Please fix the commit title of the Xinerama patch to match the MR. Currently it looks like the title is missing, so it's getting the first sentence of the patch description as the title.

iphands:
> @iphands something went wrong here. Now I see 787 commits!

So sorry I rebased on the latest upsteam, I'll undo that and fix the title

iphands:
@mattst88 title should be good now eh?

@billie80 I undid the change that pulled in the newer commits to master, should be clean now

billie80:
Looks better now. Will take a closer look tomorrow!

mattst88:
Looks good to me. Thanks!

billie80:
Applied in fe3fee1f2cb261408b960f50b9f5d61978c61086

PR #18128, opened 2020-11-04
net-libs/nDPI: bump to 3.4, drop 2.x for QA

ajakk:
Bump for security fixes, most patches were upstreamed so they can be dropped with the new version, cleanup older versions to get rid of QA warnings. To be on the safe side I don't want to drop all of the existing versions immediately because this is a potentially fragile package.


net-libs/nDPI/nDPI-3.4.ebuild
```diff
@@ -0,0 +1,63 @@
...
+DEPEND="dev-libs/json-c:=
+	net-libs/libpcap"
+RDEPEND="${DEPEND}"
+
+PATCHES=(
+	"${FILESDIR}/${PN}-3.4-fix-oob-in-kerberos-dissector.patch"
+)
+
+src_prepare() {
+	eval $(grep '^NDPI_MAJOR=' autogen.sh)
```
thesamesam, line 25: I'd like to change what we're doing with evals but now's not the time given the likely sec implications here.

PR #18130, opened 2020-11-04
dev-vcs/tortoisehg: 5.6 version bump

band-a-prend:


band-a-prend:
@mgorny 
Are additional patches from me are required to make CI know about last changes in `xml-schema metadata.dtd` and `dtd/metadata.dtd` where heptapod was added as remote-id?

Also, could you note, please, what project does currently provide packages.gentoo.org, I see several absolete in https://gitweb.gentoo.org/.
I'm interesting where remote-id convert to proper url on package info page.

band-a-prend:
As for packages.gentoo.org i seems I need to send patch for https://gitweb.gentoo.org/sites/soko.git/tree/pkg/app/handler/packages/utils.go is it?

juippis (approved):

juippis:
You can open a bug with patch included, p.g.o team should notice it that way.

PR #18131, opened 2020-11-04
dev-libs/amdgpu-pro-opencl: bump to 20.40.1147286

maxaudron:


Chiitoo:
Cool!

Bug 706406 [1] should probably be linked to.

This seems to still only install the "legacy" files, meaning this probably won't work with Vega or at least not with Navi.  It will of course be difficult for someone without the hardware to test support for those, but I figured I'd mention my findings while I'm here (more details on that in the bug and the forums).

For future reference, when someone does add support for the newer cards, I wonder if it would make sense to add a USE-flag of sorts for "legacy" and the more recent hardware, since the files needed are different, and not exactly tiny (around 100 MiBs each).

Also, I don't think this will be included as is, since the package is currently 'maintainer needed'.

1. https://bugs.gentoo.org/706406

maxaudron:
linked the bug.

I saw some talk about the "legacy" compat before on the forums and been trying around with it locally already, i wanted to introduce this as a use flag then. Currently still haveing some general trouble getting OpenCL to work in general which i gotta figure out.

i might take maintainance of this package, but need to do some reading first

juippis (approved):
Can't test this, but the ebuild looks good enough. Let's wait for bug reports.

PR #18135, opened 2020-11-04
net-im/psi: sync the live version with upstream changes

Ri0n:
Signed-off-by: Sergey Ilinykh <rion4ik@gmail.com>

QCA being compiled without gpg plugin won't let Psi start. But it's purely a QCA bug. So I removed the dependency anyway since it's not used anymore. The bug is fixed in this QCA fork https://github.com/psi-im/qca. But QCA devs are rather dead wrt accepting patches.
So likely I'll end up with something like app-crypt/psi-qca eventually, but not sure when.
I'll make another PR later always forcing gpg plugin to be enabled for qca.

### DIFF
--- a/net-im/psi/psi-9999.ebuild
+++ b/net-im/psi/psi-9999.ebuild
@@ -20,7 +20,7 @@ EGIT_MIN_CLONE_TYPE="single"
 LICENSE="GPL-2"
 SLOT="0"
 KEYWORDS=""
-IUSE="aspell crypt dbus debug doc enchant extras +hunspell iconsets keyring webengine webkit xscreensaver"
+IUSE="aspell dbus debug doc enchant extras +hunspell iconsets keyring webengine xscreensaver"
 
 REQUIRED_USE="
 	?? ( aspell enchant hunspell )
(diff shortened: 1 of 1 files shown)


a17r:
> But QCA devs are rather dead wrt accepting patches.

QCA is active. https://invent.kde.org/libraries/qca

Ri0n:
@a17r , I know where their repo is. After they rejected my last PR I see I won't find common language with them. And I have many more fixes/improvements in my QCA fork.
So I'll add a package for my QCA fork.
Probably I'll name it `app-crypt/qcang`. It won't collide with the original, I'll take care about it.

a17r:
> @a17r , I know where their repo is. After they rejected my last PR I see I won't find common language with them. And I have many more fixes/improvements in my QCA fork.

I don't think that's a sustainable attitude. I doubt you want to fork every project that rejects a contribution once.

a17r:
I'm picking `qtwebkit--` change for now, you'll have to rebase 9999 change afterwards.

"Toxic" is not a term applicable to KDE community, rest assured.

Neustradamus:
@a17r: Any news about this PR?

PR #18139, opened 2020-11-05
games-board/mah-jong: 1.16 version bump

band-a-prend:
Closes: https://bugs.gentoo.org/751040


games-board/mah-jong/mah-jong-1.16.ebuild
```diff
@@ -0,0 +1,51 @@
...
+
+MY_P="mj-${PV}-src"
+DESCRIPTION="A networked Mah Jong program, together with a computer player"
+HOMEPAGE="https://mahjong.julianbradfield.org/"
+SRC_URI="http://mahjong.julianbradfield.org/Source/${MY_P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
```
DarthGandalf, line 16: No need, if it's empty

games-board/mah-jong/mah-jong-1.16.ebuild
```diff
@@ -0,0 +1,51 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit desktop toolchain-funcs xdg
+
+MY_P="mj-${PV}-src"
+DESCRIPTION="A networked Mah Jong program, together with a computer player"
+HOMEPAGE="https://mahjong.julianbradfield.org/"
+SRC_URI="http://mahjong.julianbradfield.org/Source/${MY_P}.tar.gz"
```
DarthGandalf, line 11: https?

games-board/mah-jong/mah-jong-1.16.ebuild
```diff
@@ -0,0 +1,51 @@
...
+
+S=${WORKDIR}/${MY_P}
+
+src_prepare() {
+	default
+
+	sed -i -e '/^.TH/ s/1/6/' xmj.man || die
+	sed -i \
+		-e "/^DESTDIR =/ s:=.*:= ${D}:" \
+		-e "/^BINDIR =/ s:=.*:= /usr/bin:" \
```
DarthGandalf, line 33: Shouldn't this have `${EPREFIX}`?

band-a-prend: ${EPREFIX} is empty here. Should I add it anyway?

DarthGandalf: It's not empty only on Gentoo Prefix: https://wiki.gentoo.org/wiki/Project:Prefix/Technical_Documentation#EPREFIX_variable

band-a-prend: If I replace`-e "/^BINDIR =/ s:=.*:= /usr/bin:" \` with `-e "/^BINDIR =/ s:=.*:= ${EPREFIX}/usr/bin:" \`
it installs the same path `/usr/bin`

If replace with `-e "/^BINDIR =/ s:=.*:= ${EPREFIX}:" \` I get warning 
```
 * 
 *   /mj-player
 *   /mj-server
 *   /xmj

```

band-a-prend: Is EPREFIX for configure phase?
Should I then use prefix.eclass for prepare phase?

DarthGandalf: No. If the full path is e.g. /home/user/gentoo/usr/bin, then EPREFIX part of it is `/home/user/gentoo`. Usually it's empty

band-a-prend: add ${EPREFIX} before /usr/bin

games-board/mah-jong/mah-jong-1.16.ebuild
band-a-prend: fixed

games-board/mah-jong/mah-jong-1.16.ebuild
band-a-prend: fixed

games-board/mah-jong/mah-jong-1.16.ebuild
chewi: I haven't seen the file involved here but the `hprefixify` function from `prefix.eclass` is nicer than manually `sed`ing `EPREFIX` into paths like this.

band-a-prend: The initial line to be replaced is `BINDIR = bin`.
Do you mean besides sed with `-e "/^BINDIR =/ s:=.*:= /usr/bin:"` 
(and instead of `-e "/^BINDIR =/ s:=.*:= ${EPREFIX}/usr/bin:"`)
inherit `prefix` and then after sed expression I need to add `hprefixify -w /^BINDIR/ Makefile`?

chewi: Ah, I see it won't work in this case because it relies on the value being prefixed with `/`. It would have been okay if there was a `PREFIX` variable in the Makefile but they've just abused `DESTDIR` for this purpose. That sucks. You could probably avoid all this hassle by just doing this. I haven't tested it though!

```bash
src_compile() {
	emake \
		CC="$(tc-getCC)" \
		EXTRA_CFLAGS="${CFLAGS}" \
		EXTRA_LDOPTIONS="${LDFLAGS}" \
		CDEBUGFLAGS= \
		TILESETPATH="${EPREFIX}/usr/share/${PN}"
}

src_install() {
	emake \
		DESTDIR="${ED}/usr/"
		MANDIR=man/man6 \
		MANSUFFIX=6 \
		INSTPGMFLAGS= \
		install install.man

	…
}
```

DarthGandalf: That would work if they used `?=` to set these variables inside Makefile, but they used `=`. Perhaps need to add a patch (or sed?) which removes `CC=gcc` etc from Makefile, then such emake call should work.

chewi: Nah, that's not how the rule works. `?=` will prefer an environment variable if it is set but `=` can still be overridden by passing the variables as arguments.

band-a-prend: @chewi thank you for tip.

`TILESETPATH='"${EPREFIX}/usr/share/${PN}"' ` should be (add outer single parentheses) otherwise it passed as value/expression instead of `char*` and fail to compile.

Now I need fix `install.man` part. Currently it results in sandbox fails trying install into /usr/local. Hope fix it soon.

DarthGandalf: But if you pass it like that, `${EPREFIX}` will be compiled into the binary as is, instead of expanding as a bash variable

band-a-prend: :(
I will recheck this as somewhere I saw unexpanded values in the begin of build log  while on compilile flags they was expanded in `-DTILESETPATH`.

chewi: Or try: `TILESETPATH="\"${EPREFIX}/usr/share/${PN}\""`

band-a-prend: Done

DarthGandalf:
Installs and works for me.

Why do you need dev-lang/perl in DEPEND? As it's used during build and called on the build machine (instead of the target arch), it should be in BDEPEND.

band-a-prend:
Always forget about BDEPEND check.
Fixed.

band-a-prend:
@chewi 
ping! 
any additional changes are required before merge?

PR #18143, opened 2020-11-06
net-im/telegram-desktop-bin: bump version to 2.4.7 and cleanup

henning-schild:


henning-schild:
WIP ... already two new releases since that one

henning-schild:
Turns out the two "new" ones are only pre-releases, so they do not count and 2.4.7 stays the most recent.

PR #18145, opened 2020-11-06
dev-embedded/avra: fix CC call, bump to 1.4.2

jsmolic:


juippis (approved):
I hope you're aware you don't have to become a maintainer for every package you fix.

PR #18147, opened 2020-11-06
app-i18n/uchardet: keywording as depency for sys-apps/groff

MatthiasCoppens:
Closes: https://bugs.gentoo.org/750032
Signed-off-by: Matthias Coppens <coppens.matthias.abc@gmail.com>


thesamesam:
We can't do this. Individual arch teams have to test it, sorry.

PR #18156, opened 2020-11-06
media-video/obs-studio: bump to 27.0.1 and add browser source support

ngg:
Closes: https://bugs.gentoo.org/717002
Closes: https://bugs.gentoo.org/795648


Chiitoo:
Thanks!

I'm here just to note that I have seen this, and the bug, but haven't had the time and energy to do anything about them yet.

Will hopefully get to it soon!

media-video/obs-studio/obs-studio-25.0.8-r1.ebuild
```diff
@@ -73,15 +78,37 @@ DEPEND="
...
+
 PATCHES=( "${FILESDIR}/${PN}-25.0.8-gcc-10-build.patch" )
 
 pkg_setup() {
 	use python && python-single-r1_pkg_setup
 }
 
+src_unpack() {
+	default
+	if [ "${PV}" != "9999" ]; then
```
Chiitoo, line 98: I'd prefer '[[' instead of '[', but I can change that later while changing the existing ones, so this is more for information than anything.  :]

Chiitoo:
The changes for non-9999 should probably go into a new ebuild, a 25.0.8-r2.

I'll try to run some tests soon.  In the mean time, @a17r, want to have a peek too?

holgersson32644:
> The changes for non-9999 should probably go into a new ebuild, a 25.0.8-r2.
FYI, upstream already released 26.0.2 ;-)

media-video/obs-studio/obs-studio-25.0.8-r1.ebuild
ngg: I've changed it to [[ to match the condition at the top of the file

ngg:
> The changes for non-9999 should probably go into a new ebuild, a 25.0.8-r2.

I've renamed the ebuild to -r2.

> I'll try to run some tests soon.

I've tried the 25.0.8 and 9999 versions both with and without "browser" flag enabled.

ngg:
> FYI, upstream already released 26.0.2 ;-)

ALSA support is broken in 26.0.2, see https://bugs.gentoo.org/745996 for details.

ngg:
I've also updated the list of supported python versions (I've tested it with 3.8 and 3.9 as well), and updatetd the ebuild to use cmake.eclass instead of the deprecated cmake-utils.eclass

Chiitoo:
Do Python scripts actually work for you with >=3.8?

That should be broken as well... unfortunately: https://bugs.gentoo.org/754006

ngg:
I've reverted the python changes. I've only tested new python versions with python-only scripts which worked perfectly. If I tried to import a binary module (ssl) into those scripts, it was broken as the issue suggests.

WizardCM:
For Python 3.8+ you'll want to keep an eye on https://github.com/obsproject/obs-studio/pull/3689

BtbN:
Missing a dependency on dev-libs/nss when browser is enabled.

ngg:
> Missing a dependency on dev-libs/nss when browser is enabled.
Sorry, I've totally missed the deps of the CEF part, I've updated the dependencies. (Lots of deps were missing, not just nss)

media-video/obs-studio/obs-studio-9999.ebuild
```diff
@@ -55,6 +63,22 @@ DEPEND="
...
+	browser? (
+		app-accessibility/at-spi2-atk
+		dev-libs/atk
+		dev-libs/expat
+		dev-libs/glib
+		dev-libs/nspr
+		dev-libs/nss
+		media-libs/fontconfig
+		x11-libs/libXScrnSaver
+		x11-libs/libXcursor
```
Chiitoo, line 75: Possible nitpick: it looks like some like to sort upper case before lower case, while some disregard case.

I might prefer disregarding case, but I'm not even entirely sure at this time.  In any... case... it will ultimately be up to the person who'll do the actual commit (which I'm unable to, still).

media-video/obs-studio/obs-studio-9999.ebuild
```diff
@@ -73,13 +97,35 @@ DEPEND="
 "
 RDEPEND="${DEPEND}"
 
+QA_PREBUILT="
+	/usr/lib*/obs-plugins/swiftshader/libEGL.so
+	/usr/lib*/obs-plugins/swiftshader/libGLESv2.so
+	/usr/lib*/obs-plugins/libEGL.so
+	/usr/lib*/obs-plugins/chrome-sandbox
+	/usr/lib*/obs-plugins/libcef.so
+	/usr/lib*/obs-plugins/libGLESv2.so
```
Chiitoo, line 106: Another nit: perhaps these could be sorted as well?

Chiitoo:
Thanks again!

Regarding the non-9999, please leave the r1 be around for now, and instead or renaming, create a new file for r2.  I'll clean the r1 off at a later time (usually it would be preferable that the removal happened in a separate commit).

I can't spot anything else at this time, and it does seem to work for me.

I'll add asturm as a reviewer in case they've got time to look into it, and do the commit if all looks good.

ngg:
I've brought back the r1 version and sorted the lists.

Chiitoo:
Since no one that can push this has arrived yet, if you're willing and like to work on this some more, adding another commit which would fix bug 752768 [1] could be nice, as I've been waiting to get that done already before that report was opened.

I think it should be a pretty simple thing to do.

If you'd rather not, that's completely fine as well!

1. https://bugs.gentoo.org/752768

ngg:
I've added r100 with lua-single (just like other packages handled this), and updated 9999 to use that as well. It seems to work correctly for me.

Chiitoo:
Cool, thanks!

Hm, it looks like there's a "competing" implementation of the browser feature in bug 745996 [1].  I haven't got to really comparing yet at all, so I have no comments regarding which would be better...

1. https://bugs.gentoo.org/745996#c6

Chiitoo:
Finally managed to have a look again, though not sure if being awake enough... but in any case, I do at this time prefer this implementation since it seems the alternative would make 'git' and friends a hard dependency.

Now... the lua (and some unrelated changes) got merged, so I'm afraid this could use some minor adjustments again.  :]

Chiitoo:
I was thinking about this for a bit at the start, and asturm brought it back to my mind: what do you think about making the browser plug-in a separate package?  Can you think of any issues with that approach?

ngg:
> Now... the lua (and some unrelated changes) got merged, so I'm afraid this could use some minor adjustments again. :]

I've rebased the PR on current master, and made the new ebuild r101 instead of r2, because r100 is already unmasked since https://github.com/gentoo/gentoo/commit/ff4bb4ec63894a8586160c8b54e88715bb8bc245

> I was thinking about this for a bit at the start, and asturm brought it back to my mind: what do you think about making the browser plug-in a separate package? Can you think of any issues with that approach?

I'm not sure how the build system works, if it's possible to compile the plugin separately and how important it is to keep the version of OBS and the plugin in sync. In theory, it can work, I'll take a look in the following days.

ngg:
> > I was thinking about this for a bit at the start, and asturm brought it back to my mind: what do you think about making the browser plug-in a separate package? Can you think of any issues with that approach?
> 
> I'm not sure how the build system works, if it's possible to compile the plugin separately and how important it is to keep the version of OBS and the plugin in sync. In theory, it can work, I'll take a look in the following days.

I think this current way is the proper one to support the browser plugin. The plugin itself does not have a separate build system, so building it as a separate package would still need the full OBS source and some modifications to the CMake files (or at least to the install part). On top of that, it seems that the plugin version and the OBS version must almost always be in sync in order to compile and work correctly.
I'm sure it's possible to create a separate ebuild for the plugin, but it seems it's much easier to maintain them this way. (The only thing to do with this method is to update the submodule commit hash when bumping the version)

BtbN:
Something I've been wondering: It shouldn't at all be impossible to compile CEF, instead of downloading pre-built binaries, should it?
Granted, building a browser alongside OBS would make the build time explode, but pulling along a binary strikes me as a bit odd for an ebuild.
On top of that, the downloaded binary CEF/Chromium version is _really_ old and Chromium had quite a few security issues fixed since 75.0.3770.100, which is what CEF 3770 corresponds to.

Chiitoo:
Yeah, there's bug 529604 [1] for CEF, but no one seems to have been interested enough in working on it.

By the by, I started going for the version bump at 18883 [2], but I would not mind if this still got merged as is, I can add the changes to the new version later, but if you want, you could also do it here.  That PR needs some adjustments in any case, since I got part of it upstreamed sooner than anticipated (also 26.1.2 is out, and wondering a bit if there will be another hotfix release very soon).

1. https://bugs.gentoo.org/529604
2. https://github.com/gentoo/gentoo/pull/18883

ngg:
> On top of that, the downloaded binary CEF/Chromium version is _really_ old and Chromium had quite a few security issues fixed since 75.0.3770.100, which is what CEF 3770 corresponds to.

Unfortunately CEF has lots of API changes between releases, and the OBS plugin is not compatible with newer versions, they explicitly say that only the 3770 version works. Let's hope they'll update the required CEF version soon.

ngg:
The version bump was merged yesterday, I've updated this PR accordingly (now it applies to 26.1.2 and 9999). I've also updated the CEF versionfrom 3770 to 4280 which is currently the newest that OBS claims to support.

Chiitoo:
Since 27.0.0 is getting close, perhaps this can wait for that since no one got to merging it yet (I'm still fine with it).

I haven't looked at all, but does the CEF version get even newer with that release?

ngg:
> I haven't looked at all, but does the CEF version get even newer with that release?
Their current master branch still uses CEF 4280

nipafx:
Hey there, frequent OBS-on-Gentoo user here. Just wanted to let you know that I appreciate your work and would love this PR to get merged. I'm eagerly awaiting pimping my live stream with the browser source. 😃

Chiitoo:
Some updates will be necessary now due to the Python (and other) changes.

@nipafx, if you'd like to already use this, you could set up a local overlay [1] if you don't have one, and install from there.

1. https://wiki.gentoo.org/wiki/Handbook:Parts/Portage/CustomTree#Defining_a_custom_repository

ngg:
@Chiitoo I've rebased this PR onto the current master branch to include the Python and other changes.
Can I help in any other way to have this merged?

Chiitoo:
Thanks!

Not much you can do, I think, since we'll either need me to finally complete the ebuild-maintainer quiz (well, I did that, but haven't sent it out for review) so that I can push the changes to the Gentoo repository, or we need someone else with that power to do it instead.

I'll try asking around in the coming days.

ngg:
27.0.0 is released, I've updated this PR to both include browser support and the version update in 2 separate commits.

@a17r Can you please help us merging this?

l33tlinuxh4x0r:
I can confirm that the 27.0.0 ebuild works correctly. Thanks!

Chiitoo:
A point release also happened, and a bug [1] was filed for it.

1. https://github.com/gentoo/gentoo/pull/18156

ngg:
> A point release also happened, and a bug [1] was filed for it.
> 
>     1. https://bugs.gentoo.org/795648

I've updated this PR with the new version, seems to work correctly.

media-video/obs-studio/metadata.xml
```diff
@@ -12,6 +12,7 @@
     internally.
   </longdescription>
   <use>
+    <flag name="browser">Enable browser source support via CEF.</flag>
```
vilhelmgray, line 15: I see this USE flag allows users to use a browser source in OBS. The ebuild downloads the browser source plugin as a binary -- which is understandable given how long it takes to compile the plugin from source -- but perhaps this USE flag should reflect that so users know what's happening. Would it make more sense to call this USE flag `browser-bin` or something similar to make it obvious that the plugin is downloaded as a binary?

mjeveritt: Possibly `browser-src-bin` might be a better name, but the description should definitely be updated to reflect this, as you rightly note! :smiley:

ngg: The plugin itself is compiled from the git submodule. Only CEF is downloaded as a binary. I'll update the description but not sure about changing the flag's name.

Chiitoo: Right, and I've seen some talk about ebuilds only including '-bin' if there is a 'from source' option available, which as far as I know, we don't have have.

Of course this isn't about an ebuild name, and I don't remember if there was any sort of consensus reached though.

I don't mind if it's somehow mentioned in the description of the flag.

vilhelmgray: I'm sorry, yes it's the CEF that is downloaded as a binary, so mentioning it in the description would be good. Since the plugin itself is compiled from source, I suppose `browser` is fine as a USE flag name. If in the future we decide to allow users to compile the CEF, then perhaps a `cef` USE flag would be added, so I reckon there's no naming conflict there between the USE flags.

ngg: I've updated the description of the USE flag

nipafx:
@Chiitoo Thanks for the tip with the local overlay. I think I got everything set up correctly (this is my first local ebuild)

Unfortunately, the build fails. Maybe somebody can help me out? ([Here's the emerge info and build log.](https://gist.github.com/nipafx/42099ab4a169a918825fb2bbb97e4338))

stefantalpalaru:
> Unfortunately, the build fails. Maybe somebody can help me out?

Problem described here: https://github.com/obsproject/obs-studio/issues/4827

The workaround is to disable the "imagemagick" USE flag.

nipafx:
Thanks @stefantalpalaru, that did the trick! 🙏🏾

Chiitoo:
Unfortunately asturm (who usually helps me out with the packages I maintain) has been very busy lately, and as such has not had the time for this.

On another note, since building with USE="imagemagick" indeed seems to be currently broken (thought I last tested it not too long ago...), it's probably a good time to remove it, as the option for it will be removed upstream sometimesoon'ish [1].

1. https://github.com/obsproject/obs-studio/pull/4560#issuecomment-826345608

ionenwks (commented):
> Unfortunately asturm (who usually helps me out with the packages I maintain) has been very busy lately, and as such has not had the time for this.

I'll take a closer look later and merge myself if still needed, but get on it with the tree access thing :)

For now just a quick ebuild check.

media-video/obs-studio/metadata.xml
```diff
@@ -12,6 +12,7 @@
     internally.
```
ionenwks, line 12: `<maintainer type="person">`
This is remind me, at g.o address without commit access is one of the reason we got these new `proxied="yes"` tags. I could make the change (on merge, don't worry about this) and add myself as proxy if @Chiitoo acks it so these PRs don't get lost.

I don't use obs (last time I streamed I used standalone ffmpeg :eyes:) but I still want it to see timely updates.

media-video/obs-studio/obs-studio-27.0.1.ebuild
```diff
@@ -0,0 +1,220 @@
...
+	KEYWORDS="~amd64 ~ppc64 ~x86"
+fi
+SRC_URI+=" browser? ( https://cdn-fastly.obsproject.com/downloads/${CEF_DIR}.tar.bz2 )"
+
+DESCRIPTION="Software for Recording and Streaming Live Video Content"
+HOMEPAGE="https://obsproject.com"
+
+LICENSE="GPL-2"
+SLOT="0"
+IUSE="+alsa browser decklink fdk imagemagick jack lua nvenc pipewire pulseaudio python speex +ssl truetype v4l vlc wayland"
```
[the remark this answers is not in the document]
Chiitoo: I think it's a bit easier to read and also more consistent with other things in the file when the quotes sort of mimic what the braces do, although in general I much prefer "Allman style" always.  :]

media-video/obs-studio/obs-studio-27.0.1.ebuild
```diff
@@ -0,0 +1,220 @@
...
+		elog "The 'sys-apps/dbus' package is not installed, but"
+		elog "could be used for disabling hibernating, screensaving,"
+		elog "and sleeping.  Where it is not installed,"
+		elog "'xdg-screensaver reset' is used instead"
+		elog "(if 'x11-misc/xdg-utils' is installed)."
+		elog
+	fi
+}
+
+pkg_postinst() {
```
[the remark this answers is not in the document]
Chiitoo: That's funny.  Sort of.  Must have been pretty sleepy when I did that (okay that's me always).

media-video/obs-studio/obs-studio-27.0.1.ebuild
```diff
@@ -0,0 +1,220 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+CMAKE_REMOVE_MODULES_LIST=( FindFreetype )
+LUA_COMPAT=( luajit )
+PYTHON_COMPAT=( python3_{7..9} )
```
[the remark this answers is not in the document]
Chiitoo: I did try with Python 10 on May 6th, and it worked for what I tested, but asturm opined it would have been a bold move at the time.  Not that that necessarily meant not to do it, I maybe guess.

media-video/obs-studio/metadata.xml
Chiitoo: I would not mind that, although I do wonder a bit if there might be someone better and willing to take care of the package (or all of the packages I maintain) altogether.

ionenwks: I think you've been doing great with this personally, and hope you keep at it.

Given haven't heard back from @ngg after last short review, and this PR has waited and for a long time already, I'll go ahead and do the last few checks / fixes (just warning ahead of time so hopefully no work duplication).

I need to mask USE=browser on ppc64/x86 too given it's using amd64 binaries. Haven't checked if there's binaries for the rest but I have no real way to test this and I'll leave it how it is.

ngg: Hi! Sorry for not responding, I've been busier than usually. I'll have time to make changed on the weekend if that's ok.

ionenwks: Oh don't worry about it, unless you really want to do it yourself I'm happy to just do it (especially if you're busy).

You've done a lot here already.

ionenwks: I could ask another question though, are you interested in co-maintaining this officially?

ionenwks:
Well, I went ahead with it. Thanks for all the work.

See https://github.com/gentoo/gentoo/commit/2c4e606350f81335053fc77a81bc9790379adcb9 for the few extra changes I did, or the sync live https://github.com/gentoo/gentoo/commit/e2fee93fb7bebb6e7bbd43fb724154267d44ab2c  for differences from original.

Don't hesitate to tell me if anything else, or make new PRs as needed. This hopefully won't wait as long next time.

PR #18162, opened 2020-11-07
app-crypt/trousers: Bump to 0.3.15

salahcoronya:
Signed-off-by: Salah Coronya <salah.coronya@gmail.com>


app-crypt/trousers/trousers-0.3.15.ebuild
```diff
@@ -0,0 +1,65 @@
...
+CONFIG_CHECK="~TCG_TPM"
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	# econf --with-gui=$(usex gtk gtk openssl)
+	econf --with-gui=openssl
```
monsieurp, line 50: `--with-gui=openssl` ?! I believe there's a typo here.

monsieurp (changes requested):

app-crypt/trousers/trousers-0.3.15.ebuild
salahcoronya: No typo. That's the correct value. There 3 options for the GUI popup backend: gtk (which is broken), openssl (which works), and none (which violates the spec).

PR #18172, opened 2020-11-07
dev-libs/gobject-introspection: fix LTO compilation

atomsymbol:
This patch fixes compilation of packages which depend on gobject-introspection on machines which have -flto in CFLAGS.

Please merge. Thanks.

### DIFF
--- a/dev-libs/gobject-introspection/files/gobject-introspection-1.64.1-lto.patch
+++ b/dev-libs/gobject-introspection/files/gobject-introspection-1.64.1-lto.patch
@@ -1,0 +1,50 @@
+From cea7ef1e4a5626de1332bc34e1962befb75c2976 Mon Sep 17 00:00:00 2001
+From: Christoph Reiter <reiter.christoph@gmail.com>
+Date: Thu, 14 May 2020 09:41:43 +0200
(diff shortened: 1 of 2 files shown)


mattst88:
I don't think we want a revision bump of `gobject-introspection` for this. Can we just apply the patch directly to the latest stable?

Or better yet, would you be happy with a version bump of gobject-introspection (1.66.1) that contains this patch?

atomsymbol:
I updated the pull request to apply the patch to the latest stable, as I haven't verified yet whether 1.66.1 requires bumped versions of its dependencies.

atomsymbol:
1.66.1 emerges fine on my machine without any changes, and already includes the patch in its source code. Do you recommend an ~amd64/~etc in the 1.66.1 ebuild, or just a verbatim copy of 1.64.1-r1?

mattst88:
Thanks. I've opened a pull request here with some version bumps: https://github.com/gentoo/gentoo/pull/18174

atomsymbol:
I suppose if the intention was to merge this MR then it would already have been merged, so I am closing this MR. Thanks for adding 1.66.1 to the Portage tree.

PR #18173, opened 2020-11-07
dev-libs/ustr: Fix compilation using clang/LLVM (Bug 732026) and clean up

Yamakuzure:
This package depends on old-style GCC inline functions, but the patch to maintain this behaviour on newer compilers only works for GCC specifically.
Replace this with a check for `__GNUC_STDC_INLINE__`, which is documented to work with both GCC and clang.

Bug: https://bugs.gentoo.org/732026
Closes: https://bugs.gentoo.org/732026
Signed-off-by: Fergus Dall <sidereal@google.com>


dev-libs/ustr/ustr-1.0.4-r9.ebuild
```diff
@@ -20,13 +21,19 @@ MULTILIB_WRAPPED_HEADERS=(
...
+
 src_prepare() {
-	epatch "${FILESDIR}/${P}-gcc_5-check.patch"
+	default
 	multilib_copy_sources
 }
 
 _emake() {
 	emake \
+		USE_STATIC=$(usex static-libs) \
```
SoapGentoo, line 36: is there a reason you need static-libs?

Yamakuzure: No idea. I did not change anything here, and haven't searched for consumers.

Yamakuzure: I just saw that 'static-libs' was dropped by Aaron Bauman anyway.

PR #18174, opened 2020-11-07
dev-libs/glib and friends version bump

mattst88:


mattst88:
Merged

atomsymbol:
`glib-2.66.2` with use flag `gtk-doc` requires `gtk-doc-1.33.0` in order to build.

There is a circular dependency between `glib` and `gtk-doc`. Is there a plan for automatic resolution of circular dependencies in /usr/bin/emerge by emerging one of the packages twice with different use flags?

mattst88:
Not as far as I'm aware :(

PR #18180, opened 2020-11-08
net-proxy/mitmproxy: Require at least LibreSSL 3.2.0

00-matt:


net-proxy/mitmproxy/mitmproxy-5.3.0.ebuild
```diff
@@ -43,6 +43,8 @@ RDEPEND="
 	>=dev-python/urwid-2.1.1[${PYTHON_USEDEP}]
 	>=dev-python/wsproto-0.14.0[${PYTHON_USEDEP}]
 	>=dev-python/zstandard-0.11.0[${PYTHON_USEDEP}]
+	!libressl? ( dev-libs/openssl:0 )
+	libressl? ( >=dev-libs/libressl-3.2.0:* )
```
00-matt, line 47: Suggestion on Bugzilla was to use `=` subslot operator, but I'm not sure that it's necessary since we don't actually link to OpenSSL/LibreSSL and use dev-python/cryptography to do all of that.

PR #18182, opened 2020-11-08
net-vpn/i2pd-2.34.0: version bump

khumarahn:
Nothing special, same ebuild as for 2.33.0. Removed 2.31.0 (so we only keep last 3 versions) and a patch which is no more used.

As always, any suggestions are very welcome.


thesamesam:
Looks good but could you add tests in a new PR? Thanks!

```
>>> Test phase: net-vpn/i2pd-2.34.0
 * Working in BUILD_DIR: "/var/tmp/portage/net-vpn/i2pd-2.34.0/work/i2pd-2.34.0_build"
No tests found. Skipping.
```

khumarahn:
Thanks! I'll look into tests on next version bump.

PR #18183, opened 2020-11-08
sci-misc/boinc: V7.16.14 bump, EAPI 7 upgrade, bug fixes, GLEP 81

Yamakuzure:
This version bump also fixes the following issues:
* The user/group management is now done according to GLEP 81
* /etc/init.d/boinc: Ensure correct RUNTIMEDIR permissions (#684074)
* /etc/init.d/boinc: The password no longer needs to be submitted
* Hardcoded library search paths to /lib and /usr/lib have been removed or switched to /lib64 and /usr/lib64, if those are the regular library directories on the emerging system. (#732024)

Closes: https://bugs.gentoo.org/684074
Closes: https://bugs.gentoo.org/717580
Closes: https://bugs.gentoo.org/732024


acct-user/boinc/boinc-0.ebuild
```diff
@@ -0,0 +1,17 @@
...
+DESCRIPTION="User for sci-misc/boinc"
+ACCT_USER_ID=405
+ACCT_USER_HOME=/var/lib/${PN}
+ACCT_USER_HOME_PERMS=0750
+ACCT_USER_GROUPS=( boinc video )
+
+acct-user_add_deps
+
+DEPEND+=" acct-group/boinc "
+RDEPEND+=" acct-group/boinc "
```
heroxbd, line 17: This is not needed, because `acct-user_add_deps` adds `acct-group/*` dependencies deduced from `ACCT_USER_GROUPS`.

heroxbd:
> [...]
> 
> Bug: https://bugs.gentoo.org/684074
> Bug: https://bugs.gentoo.org/717580
> Bug: https://bugs.gentoo.org/732024
> Closes: https://bugs.gentoo.org/684074
> Closes: https://bugs.gentoo.org/717580
> Closes: https://bugs.gentoo.org/732024
> [...]

I usually remove `Bug` fields if the same `Closes` fields exist. Otherwise two nearly identical comments will be created on each of the related bugs.

acct-user/boinc/boinc-0.ebuild
Yamakuzure: Oh, that is good to know.
Actually I just copied that from an ebuild in the tree. :shrug: ( acct-user/kismet )

Yamakuzure:
> > [...]
> > Bug: https://bugs.gentoo.org/684074
> > Bug: https://bugs.gentoo.org/717580
> > Bug: https://bugs.gentoo.org/732024
> > Closes: https://bugs.gentoo.org/684074
> > Closes: https://bugs.gentoo.org/717580
> > Closes: https://bugs.gentoo.org/732024
> > [...]
> 
> I usually remove `Bug` fields if the same `Closes` fields exist. Otherwise two nearly identical comments will be created on each of the related bugs.

So the "Closes" field also does the Bug assiciation? Then the extra Bug field is of course superfluous. Thanks alot!

acct-user/boinc/boinc-0.ebuild
Yamakuzure: Done

rossbridger:
Does the package also provide a systemd service? I thinkwe can steal one from [arch linux](https://wiki.archlinux.org/index.php/BOINC)

Yamakuzure:
> Does the package also provide a systemd service? I think we can steal one from [arch linux](https://wiki.archlinux.org/index.php/BOINC)

It does:
````
sed@sed-m7550 ~ $ equery f sci-misc/boinc | grep service
/usr/lib/systemd/system/boinc-client.service
````

juippis (approved):

PR #18195, opened 2020-11-09
app-admin/graylog: security bump to 3.3.8

hydrapolic:
https://github.com/Graylog2/graylog2-server/pull/9098

Tested on amd64, thanks.


app-admin/graylog/graylog-3.3.8.ebuild
```diff
@@ -0,0 +1,83 @@
...
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc64 ~x86"
+RESTRICT="strip"
+
+RDEPEND="!app-admin/graylog2
+	acct-group/graylog
+	acct-user/graylog
+	>=virtual/jdk-1.8:*"
```
monsieurp: The correct way to declare a dependency on the JVM and have Portage get the JVM ready for an ebuild that might require it is as follows:
```
inherit java-pkg-opt-2

pkg_setup() {
        java-pkg-opt-2_pkg_setup
}

RDEPEND="
    >=virtual/jre-1.8
"

DEPEND="
    >=virtual/jdk-1.8
"
```

monsieurp (changes requested):

app-admin/graylog/graylog-3.3.8.ebuild
hydrapolic: Hello @monsieurp, actually we don't need Java for building itself, it's a binary package, so Java will be needed just for running which is expressed in the current RDEPEND.

hydrapolic:
Thanks @Whissi

PR #18196, opened 2020-11-09
 net-dns/https_dns_proxy: add snapshot ebuild

jsmolic:


net-dns/https_dns_proxy/Manifest
```diff
@@ -0,0 +1 @@
+DIST https_dns_proxy-0_p20201109.tar.gz 16270 BLAKE2B 2677f63662d030fa1b6955f819eb6cbab6916356aa06ede13c4942824ea5c9c88f93d5c53fe725989ec668b3152c8242755279752abb1ed9941b57b83d137191 SHA512 8d0bc083447eaf5973a3ed3efbd7946c62fcabca7286266d2ac445aa522a11bff41201f54e1acb49483957a7f488578cfbb7053cd0cc83ca42c4146eab4e44ca
```
juippis, line 1: Snapshot needs to be versioned after commit date, not after ebuild-write date. Because in theory you could bump it everyday and it's easier to find the commit per date.

Also use `_pre` instead of `_p`.
https://devmanual.gentoo.org/ebuild-writing/file-format/index.html#snapshots-and-live-ebuilds

net-dns/https_dns_proxy/https_dns_proxy-0_p20201109.ebuild
```diff
@@ -0,0 +1,31 @@
...
+	EGIT_REPO_URI="https://github.com/aarond10/https_dns_proxy.git"
+	inherit git-r3
+else
+	MY_COMMIT="2d9285e2b94bce21c588c8160f8fac660806987a"
+	SRC_URI="https://github.com/aarond10/https_dns_proxy/archive/${MY_COMMIT}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64 ~x86"
+	S="${WORKDIR}/${PN}-${MY_COMMIT}"
+fi
+
+DESCRIPTION="A lightweight DNS-over-HTTPS proxy."
```
juippis, line 17: We don't write the `.` here.

net-dns/https_dns_proxy/Manifest
jsmolic: Thanks, done

net-dns/https_dns_proxy/https_dns_proxy-0_p20201109.ebuild
jsmolic: Fixed

juippis (approved):

PR #18199, opened 2020-11-09
app-emulation/xen: add security patches

hydrapolic:
https://bugs.gentoo.org/753692

4.13.2 boot tested on amd64

Thanks!


hydrapolic:
Thank you @Whissi

PR #18204, opened 2020-11-09
dev-python/uncertainties: add future as dependency

hanmertens:
The main thing is that it adds `dev-python/future` as a dependency in `(R)DEPEND`. This dependency [is necessary](https://github.com/lebigot/uncertainties/blob/3.1.4/setup.py#L336) to prevent the following traceback upon `import uncertainties`:

```
  File "/usr/lib/python3.7/site-packages/uncertainties/__init__.py", line 225, in <module>
    from .core import *
  File "/usr/lib/python3.7/site-packages/uncertainties/core.py", line 22, in <module>
    from past.builtins import basestring
ModuleNotFoundError: No module named 'past'
```

I accompanied this with a revision bump to `r1`, but let me know if that's not necessary. While I was touching the ebuild anyway, I also made some minor changes:
- Update to EAPI 7 (and moved relevant `DEPEND` to `BDEPEND` because of that)
- Remove explicit `setuptools` dependency, as that is handled by the `distutils-r1` eclass already
- Remove unused `ppython_compile_all` function

Related to (closes?): https://bugs.gentoo.org/732832

### DIFF
--- a/dev-python/uncertainties/uncertainties-3.1.4-r1.ebuild
+++ b/dev-python/uncertainties/uncertainties-3.1.4-r1.ebuild
@@ -1,7 +1,7 @@
 # Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
 
 PYTHON_COMPAT=( python3_{6..9} )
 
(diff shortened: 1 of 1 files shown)


monsieurp (approved):
Hi! A revision bump is definitely required when changing dependencies. Thanks for that. However please avoid making the required changes to the faulty and then renaming it straight to r1. For your next PR: leave the faulty ebuild in the repository and add the revision bump with your changes to the PR. I will make an exception here and merge your PR but keep that in mind.

hanmertens:
Thanks for letting me know the proper way to do a revision bump, I'll keep that in mind in the future.

PR #18213, opened 2020-11-10
media-libs/libavif: version bump to 0.8.3

novomesk:
@juippis
Just a side note: libavif can use also another encoder called SVT-AV1. This encoder is now in GURU overlay ( media-libs/svt-av1 ). If we want to use it in the future, we will need to move it from ::guru -> ::gentoo


juippis:
> Just a side note: libavif can use also another encoder called SVT-AV1. This encoder is now in GURU overlay ( media-libs/svt-av1 ). If we want to use it in the future, we will need to move it from ::guru -> ::gentoo

Please go ahead if you're willing to maintain it too.

novomesk:
In the near future I am not going to focus on the SVT-AV1. It is not a top priority yet because preferred encoder by libavif is libAOM, then rav1e and svt-av1 is on the last place.

juippis (approved):
Really weird, this works again with efl. So 0.8.1 and 0.8.3 do work while 0.8.2 is broken. I'm tempted to believe 0.8.2 is a broken release by upstream? I'm going to remove it after merging this.

As for my interest towards svt-av1; I just want this to work with efl, so this would get minimum attention from me :P I'm glad you took on the maintainership and seem to be doing a good job here. I'll see this package stabilized on amd64+x86 at some point and then leave the rest to you.

novomesk:
I don't know if efl was adjusted to new libavif API or they ceased to use libavif.
libavif API in 0.8.2 is almost the same as in 0.8.3, there is new possibility to use SVT-AV1 encoder.
My programs worked with 0.8.2 and compile without any change with 0.8.3

I think it is quite unlikely that old code suddenly starts working in 0.8.3

juippis:
Yes, as little as it makes sense, 
https://github.com/gentoo/gentoo/pull/17933#pullrequestreview-510129692
I don't get it either.
Just sounds like upstream made some changes and reverted them back between versions...

PR #18219, opened 2020-11-11
games-fps/avp: update to 20170505

winterheart:
Fix compilation with -fno-common and migrate to EAPI 7.
Closes: https://bugs.gentoo.org/708416


thesamesam:
Could you rebase given Soap did GCC 10?

winterheart:
Patch for 20150214-r1 not applies for new version, other changes are minor.

ionenwks:
The old version now also received a gcc11 patch, I'd assume it's still relevant for this version, rebasing to integrate everything from current would be nice (can also remove the bugref). If it runs fine I'll give it the test run and hopefully won't make this wait much longer.

There's also this [git repo fork](https://github.com/neuromancer/avp) with a handful of commits, maybe check if anything's important in there if haven't yet (but not very active, so I'm not giving it much consideration yet).

I'd prefer if patches didn't use _ in their names too, same naming as in-tree patches is fine (just use new version for new or rebased patches).

games-fps/avp/avp-20170505.ebuild
```diff
@@ -0,0 +1,43 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+inherit cmake
+
+DESCRIPTION="Linux port of Aliens vs Predator"
+HOMEPAGE="http://www.icculus.org/avp/"
+SRC_URI="http://www.icculus.org/avp/files/${P}-a1.tar.gz"
```
ConiKost, line 9: I am not sure, is a1 maybe alpha1?

PR #18221, opened 2020-11-11
net-irc/weechat: Version updated to 3.0

jpds:


gyakovlev:
thanks, but this needs a port to new `lua-single.eclass`
https://bugs.gentoo.org/752813
I'll be looking at it within next ~12hrs.

gyakovlev:
thanks again, bumped using your PR + some changes in next commit from me.
lua change can wait, as it will require masking.

PR #18223, opened 2020-11-11
net-misc/gallery-dl: add required python use

xgqt:

Fixes: https://bugs.gentoo.org/754030

### DIFF
--- a/net-misc/gallery-dl/gallery-dl-1.15.2-r1.ebuild
+++ b/net-misc/gallery-dl/gallery-dl-1.15.2-r1.ebuild
@@ -4,6 +4,7 @@
 EAPI=7
 
 PYTHON_COMPAT=(python3_{7,8})
+PYTHON_REQ_USE="sqlite,ssl,xml"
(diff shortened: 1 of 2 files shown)


net-misc/gallery-dl/gallery-dl-1.15.2-r1.ebuild
```diff
@@ -4,6 +4,7 @@
 EAPI=7
 
 PYTHON_COMPAT=(python3_{7,8})
+PYTHON_REQ_USE=(sqlite ssl xml)
```
juippis, line 7: This doesn't seem to work, you can test it by emerging any python without these USE flags then emerging gallery-dl.

```suggestion
PYTHON_REQ_USE="sqlite,ssl,xml"
```

xgqt: Oh, ok

juippis (approved):

PR #18226, opened 2020-11-11
app-emulation/xen: add security patches

hydrapolic:
Adding more security patches, fixing building with USE=ovmf


hydrapolic:
Thank you @Whissi

PR #18231, opened 2020-11-12
mail-filter/rspamd: fix bug with lua-lpeg and sync live

arkamar:
The bundled lua-lpeg has some modifications necessary to work properly (see https://github.com/rspamd/rspamd/issues/3533). I also synced the live ebuild.

### DIFF
--- a/mail-filter/rspamd/files/rspamd-2.5-unbundle-lua.patch
+++ b/mail-filter/rspamd/files/rspamd-2.5-unbundle-lua.patch
@@ -1,22 +1,8 @@
 diff --git a/CMakeLists.txt b/CMakeLists.txt
-index bbc141170..8e3665c18 100644
+index bbc141170..7fbb1b485 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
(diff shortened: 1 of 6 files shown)


Dessa:
while at it, it should probably be discussed if blas dependency is switched back to openblas because rspamd automagically depends on it

```
NEEDED:/usr/lib/rspamd/librspamd-kann.so libopenblas.so.0,libc.so.6
NEEDED:/usr/lib/rspamd/librspamd-server.so libstemmer.so.0d,libopenblas.so.0,libpthread.so.0,libluajit-5.1.so.2,libglib-2.0.so.0,libpcre2-8.so.0,libzstd.so.1,libsqlite3.so.0,libicui18n.so.67,libicuuc.so.67,libssl.so.1.1,libcrypto.so.1.1,libz.so.1,libsodium.so.23,librt.so.1,librspamd-replxx.so,librspamd-actrie.so,librspamd-ev.so,librspamd-kann.so,libm.so.6,libstdc++.so.6,libgcc_s.so.1,libc.so.6
```

while reference blas is selected by eselect.
i suppose openblas dependency is actually automagic since when openblas was unmerged at least -kann didn't like reference blas,  -server did link iirc but yelled at me for using workarounds in cmake output

juippis (approved):

PR #18242, opened 2020-11-13
dev-libs/rocm-device-libs: Version bump to 3.9.0

justxi:


justxi:
@candrews 
The ebuild in Gentoo portage installs to a location which does not match FHS.

candrews:
> The ebuild in Gentoo portage installs to a location which does not match FHS.

That's not good... can you submit a PR (or update this one) to fix it?

justxi:
Yes.

PR #18244, opened 2020-11-13
dev-libs/rocm-device-libs: Correcting install location according to FHS

justxi:


justxi:
@candrews 
Here it is ...

PR #18246, opened 2020-11-13
dev-python/pytest-xvfb: fix tests

AndrewAmmerlaan:
tests do not work at all if the package is
not already installed to the system

Closes: https://bugs.gentoo.org/754270

### DIFF
--- a/dev-python/pytest-xvfb/pytest-xvfb-2.0.0.ebuild
+++ b/dev-python/pytest-xvfb/pytest-xvfb-2.0.0.ebuild
@@ -24,6 +24,10 @@ DEPEND="
 distutils_enable_tests pytest
 
 python_test() {
-	distutils_install_for_testing
-	pytest -vv || die "Tests failed with ${EPYTHON}"
+	local -x PYTHONPATH=${BUILD_DIR}/install/lib
+	esetup.py install \
+		--root="${BUILD_DIR}"/install \
+		--install-lib=lib
+
+	pytest -vv || die "Tests fail with ${EPYTHON}"
 }


dev-python/pytest-xvfb/pytest-xvfb-2.0.0.ebuild
```diff
@@ -21,9 +21,8 @@ DEPEND="
 	x11-base/xorg-server[xvfb]
 "
 
-distutils_enable_tests pytest
+# Needs to be installed to system for tests to work at all
+# distutils_install_for_testing does not work
+BDEPEND="test? ( dev-python/pytest-xvfb[${PYTHON_USEDEP}] )"
```
Whissi, line 26: This looks wrong to me. Please have a look at https://dev.gentoo.org/~mgorny/python-guide/distutils.html#installing-the-package-before-testing

Whissi (changes requested):

dev-python/pytest-xvfb/pytest-xvfb-2.0.0.ebuild
AndrewAmmerlaan: >  Please have a look at https://dev.gentoo.org/~mgorny/python-guide/distutils.html#installing-the-package-before-testing
I tried that (in fact that is how it was before) and it didn't work, this is not the first pytest-foobar package where I found that the `distutils_install_for_testing` function is not sufficient to get the tests working. The only solution I can think of is to have the tests of the package BDEPEND on itself, though it is a very ugly solution IMO.

sbraz: Are you sure that was the reason tests were failing? They work fine here and the original bug doesn't seem to have much to do with the plugin not being found.
I'd say it looks more like a X-related failure. I'm attempting to reproduce on a cleaner system since it works on my desktop.

AndrewAmmerlaan: > Are you sure that was the reason tests were failing?
Yes I could reproduce the error by removing the package and trying to reinstall it.
emerge -C pytest-xvfb && emerge -1 pytest-xvfb fails
USE="-test" emerge -1 pytest-xvfb && emerge -1 pytest-xvfb works

On the other hand, the fact that it is the way it is in the ebuild, implies that I got it working before with the `distutils_install_for_testing` function, though it is too long ago for me to remember.

sbraz: Can you paste the failure with the current ebuild (pre-PR)? I'm curious.

sbraz: OK I can see the failure on another system when the package isn't installed. Installing it makes it work. That's really weird.

sbraz:
It would be nice if you could add Python 3.9 support to pyvirtualdisplay and pytest-xfvb.

AndrewAmmerlaan:
> It would be nice if you could add Python 3.9 support to pyvirtualdisplay and pytest-xfvb.

@sbraz  I would if I could, but I tested that earlier and if I remember correctly on of the deps fails tests with python3.9

sbraz:
I looked at https://qa-reports.gentoo.org/output/gpyutils/38-to-39.svg too quickly and assumed pyvirtualdisplay was a leaf package. You're right, it has some deps that lack Python 3.9 support. Maybe I'll work on these; in that case I'll let you know.

AndrewAmmerlaan:
> You're right, it has some deps that lack Python 3.9 support.

The additional problem is that pyvirtualdisplay has circular dependencies with 3 other packages if USE="test" and FEATURES="test", which means that py3.9 would have to be added to all these 4 packages simultaneously

PR #18247, opened 2020-11-13
sys-cluster/k9scli: Bump version to 0.23.10

shnups:
- Bump version to 0.23.10
- Use RESTRICT="mirror", to avoid unnecessarily hitting gentoo mirrors
- Pass ${PV} to emake so that the program can properly display its version (in `k9s version` and in the main UI)


thesamesam:
Newer version already in tree. Feel free to ping in future if things seem stuck though!

PR #18252, opened 2020-11-13
app-crypt/tpm2-tss: Fix building with autoconf-2.70

salahcoronya:
Closes: https://bugs.gentoo.org/750845
Signed-off-by: Salah Coronya <salah.coronya@gmail.com>

Using trial and error, I was able to figure out the problem, I think. Someone with more autotools experience should double-check, but based on my limited knowledge of autoconf, the fix seems correct.


juippis (approved):
It does look a bit weird to me, but because I'm not an autoconf expert, if it works...

PR #18257, opened 2020-11-14
dev-python/pid: version bump to 3.0.4, add PYTHON_COMPAT 3_9

wimmuskee:
- dev-python/mock not required anymore
 - let distutils handle setuptools dep


mgorny (approved):
Thank you!

PR #18259, opened 2020-11-14
app-cdr/bchunk: Fix build without native symlinks, pass flags

DarthGandalf:


app-cdr/bchunk/bchunk-1.2.2.ebuild
```diff
@@ -13,6 +15,13 @@ KEYWORDS="amd64 ppc ppc64 sparc x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-maco
 
 DOCS=( "${PN}.spec" README ChangeLog )
 
+src_compile() {
+	emake \
+		"CC=$(tc-getCC)" \
+		"CFLAGS=${CFLAGS}" \
+		"LD=$(tc-getCC)"
```
SoapGentoo, line 22:
```
	emake \
		CC="$(tc-getCC)" \
		CFLAGS="${CFLAGS}" \
		LD="$(tc-getCC)"
```
don't quote the key, only the value, improves readability

SoapGentoo (changes requested):

juippis (approved):

PR #18266, opened 2020-11-15
dev-libs/rocr-runtime: Fixes dependency due to changed bitcode location

justxi:
Closes: https://bugs.gentoo.org/754351


candrews:
Instead, how about cleaning up the old 3.8 version?

justxi:
I would like to keep 3.8 until I have updated my repository to 3.9.
Or I have to copy all 3.8 ebuilds into  my repository.

PR #18268, opened 2020-11-15
dev-python/cherrypy: Drop Python 2 specific patches

webknjaz:
CherryPy 18+ is Python 3 only, plus `PYTHON_COMPAT` is set to py3.6-py3.9 so no Python 2 specific patches are necessary.


webknjaz:
Looks like this was addressed in https://github.com/gentoo/gentoo/commit/d8f8f0deae65eedcced8f55e3d2304f5c1c795fd by @mgorny.

PR #18282, opened 2020-11-16
sys-devel/slibtool: version bump

doughdemon:


sys-devel/slibtool/slibtool-0.5.31.ebuild
```diff
@@ -0,0 +1,16 @@
...
+DESCRIPTION="A skinny libtool implementation, written in C"
+HOMEPAGE="https://git.midipix.org/cgit.cgi/slibtool"
+SRC_URI="https://git.midipix.org/cgit.cgi/${PN}/snapshot/${P}.tar.xz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm"
+
+src_configure() {
+	./configure --host=${CHOST} --prefix="${EPREFIX}"/usr || die
```
juippis, line 15: Have this been asked before, but why wouldn't `econf` work here? Please add a comment above this line explaining why, if it doesn't.

Rest looks good.

doughdemon:
> Have this been asked before, but why wouldn't `econf` work here?

The package uses a home-grown configure script (not generated by
autoconf).

> Please add a comment above this line explaining why, if it doesn't.

I will try to do so later today.

juippis (approved):

orbea:
I just want to point out that the build system slibtool uses is called sofort.

https://dev.midipix.org/zippy/sofort

PR #18283, opened 2020-11-16
dev-db/mycli: version bump to 1.22.2

jok-ts:
The version bump allows use of dev-python/prompt_toolkit-3.0.7.
It also fixes an issue with dev-python/pymysql-0.10.1 and convert_mysql_timestamp.
The patches fix building against dev-python/sqlparse-0.4.1.

Patches are based on upstream [PR900](https://github.com/dbcli/mycli/pull/900) that is already in the master but does not apply cleanly to the versioned tree.

Closes: https://bugs.gentoo.org/754666
Signed-off-by: Jouni Kosonen jouni.kosonen@tukesoft.com


jok-ts:
Rejected by package maintainer, closing.

PR #18285, opened 2020-11-16
dev-libs/opensc: fix CardOS 5.x support

henning-schild:
This is a backport of a patch that was merged upstream to fix a
regression in 0.20.0.


henning-schild:
Not sure if we need a bug for that, please let me know and i will open one.

fedora also did that backport:
https://bugzilla.redhat.com/show_bug.cgi?id=1830528
https://src.fedoraproject.org/rpms/opensc/c/663c939ecd4b825be0513823306aba0199c63449?branch=master

And that is where the patch is coming from.

I was hoping for an upstream release but they do not seem to move, so backport while we wait ...
https://github.com/OpenSC/OpenSC/issues/2061

henning-schild:
First tried to carry the patch in FILESDIR, but it is too big for that. Later tried applying exactly the 6 upstream commits

```
git format-patch -o <gentoo>/dev-libs/opensc/files/ 797414be05220fa371972606d7a6c2e94dd6d8db...db41cd9ab1033e0c63c894e2dcdd3da14edec3cf
```

But they need backporting and do not apply. So i decided to take the patch where the ferdora maintainer already did the backport.

henning-schild:
In fact there is 0.21.0 now, keeping this open until we have 0.21.0 ...

PR #18297, opened 2020-11-17
dev-qt/qtnetwork doesn't support old openssl

Qeenon:
I've found reason here

```json
"openssl_headers": {
    "label": "OpenSSL Headers",
    "export": "openssl",
    "test": {
        "tail": [
            "#if !defined(OPENSSL_VERSION_NUMBER) || OPENSSL_VERSION_NUMBER-0 < 0x10101000L",
            "#  error OpenSSL >= 1.1.1 is required",
            "#endif",
            "#if !defined(OPENSSL_NO_EC) && !defined(SSL_CTRL_SET_CURVES)",
            "#  error OpenSSL was reported as >= 1.1.1 but is missing required features, possibly it's libressl which is unsupported",
            "#endif"
        ]
    },
```

### DIFF
--- a/dev-qt/qtnetwork/qtnetwork-5.15.1.ebuild
+++ b/dev-qt/qtnetwork/qtnetwork-5.15.1.ebuild
@@ -22,7 +22,7 @@ DEPEND="
 	networkmanager? ( ~dev-qt/qtdbus-${PV} )
 	sctp? ( kernel_linux? ( net-misc/lksctp-tools ) )
 	ssl? (
-		!libressl? ( >=dev-libs/openssl-1.1.1:0/1.1=[bindist=] )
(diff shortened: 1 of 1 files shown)


Qeenon:
Hi @a17r this accepted? can be closed?

a17r:
In principle yes, but was only added to qt overlay for now.

If you want this landed with you as author, you need to
- squash your commits into one
- add missing GCO sign-off (more on that see first reply by gentoo-bot)
- as you can see in my commit, please keep the slot definition for openssl: `1.1.1:0=`

Otherwise a similar commit as in qt overlay is going to enter Gentoo ebuild repo soon.

Qeenon:
@a17r thank you, I'm good with your change, closing this.

May I suggest/ask for using `Co-authored-by:` instead of `Thanks-to:` for future.

PR #18298, opened 2020-11-17
games-emulation/ppsspp: Add discord-presence USE flag.

orbea:
Adds a `discord-presence` USE flag similar to `games-emulation/dolphin`.


juippis:
Well, you already know we can't merge this due to our sign-off policy. But looking at the state of this package overall, and the fact that I can't compile it, it looks like a candidate for removal.

PR #18299, opened 2020-11-17
Dev-python/spyder: eutils --> optfeature

AndrewAmmerlaan:


AndrewAmmerlaan:
Ooh oopsie, Just noticed that this has already been fixed

PR #18306, opened 2020-11-17
eapi8-dosym.eclass

ulm:


eclass/eapi8-dosym.eclass
```diff
@@ -0,0 +1,110 @@
...
+		[[ ${target} == /* ]] \
+			|| die "${FUNCNAME}: -r specified but no absolute target path"
+
+		target=$(_dosym8_canonicalize "${target}")
+		dir=$(_dosym8_canonicalize "/${link#/}")
+		dir=${dir%/*}			# poor man's dirname
+		dir=${dir:-/}			# use "/" if we ended up with a null string
+
+		while [[ -n ${dir} ]]; do
+			c=${dir%%/*}
```
mgorny, line 96: This is the part where more verbose variable names would be helpful.

eclass/tests/eapi8-dosym.sh
```diff
@@ -0,0 +1,71 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+source tests-common.sh
+
+inherit eapi8-dosym
+
+dosym() {
+    echo "$1"
```
mgorny, line 12: Space indent.

ulm: Fixed.

eclass/eapi8-dosym.eclass
ulm: I went for `linkdir` and `comp`.

eclass/tests/eapi8-dosym.sh
```diff
@@ -0,0 +1,71 @@
...
+	teq b/f ${f} /a/b///./c/d/../e/..//../f /a/././///g/../h
+	teq ../h ${f} /a/././///g/../h /a/b///./c/d/../e/..//../f
+done
+
+randompath() {
+	dd if=/dev/urandom count=1 2>/dev/null \
+		| LC_ALL=C sed 's/[^a-zA-N]//g;s/[A-E]/\/.\//g;s/[F-J]/\/..\//g;s/[K-N]/\//g;s/^/\//;q'
+}
+
+# set the "crazy" variable to a positive number to enable crazy random tests
```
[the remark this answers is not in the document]
ulm: Yeah, don't take this too serious. :smile:
Though it exposed a bug for some weird corner case.

eclass/tests/eapi8-dosym.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
```
SoapGentoo, line 1: portable shebang pls ;)

SoapGentoo (changes requested):

eclass/tests/eapi8-dosym.sh
ulm: It's the same as in all other tests. And I hate these `env` shebangs (which assumes that `env` is installed in a particular location, so just shifts the problem).

SoapGentoo: what if I run these tests on some old CentOS 6 Prefix installation?

mgorny: Or on NetBSD where bash is in /usr/pkg/bin?

ulm: Yeah, sure. Continue like this and I'm going to drop tests altogether. :smile:

It's `/bin/bash` everywhere, and `tests-common.sh` sources `/lib/gentoo/functions.sh`. Does the latter exist on Prefix?

SoapGentoo: we should make that relative then

ulm: This is all part of a global issue that has little relation to this particular eclass and its tests.

PR #18308, opened 2020-11-17
sci-libs/mkl: add chainloading mechanism

epsilon-0:
makes dummy libraries that will pull in the
correct BLAS LAPACK library during runtime
for eselect-ldso

---

Need someone to triple check multilib support
I've tested it locally and it seems work, more eyes are good though


epsilon-0:
closing for larger PR with all changes for bunch of packages.

PR #18311, opened 2020-11-18
x11-wm/i3-gaps: add 4.19, drop old

ajakk:


x11-wm/i3-gaps/i3-gaps-4.19.ebuild
```diff
@@ -0,0 +1,93 @@
...
+HOMEPAGE="https://github.com/Airblader/i3"
+SRC_URI="https://github.com/Airblader/i3/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc64 ~x86"
+IUSE="doc test"
+RESTRICT="!test? ( test )"
+
+CDEPEND="dev-libs/glib:2
```
juippis: Use `COMMON_DEPEND` instead here.

x11-wm/i3-gaps/i3-gaps-4.19.ebuild
```diff
@@ -0,0 +1,93 @@
...
+	x11-libs/libxkbcommon[X]
+	x11-libs/pango[X]
+	x11-libs/startup-notification
+	x11-libs/xcb-util
+	x11-libs/xcb-util-cursor
+	x11-libs/xcb-util-keysyms
+	x11-libs/xcb-util-wm
+	x11-libs/xcb-util-xrm
+"
+DEPEND="${CDEPEND}
```
juippis: And here.

x11-wm/i3-gaps/i3-gaps-4.19.ebuild
```diff
@@ -0,0 +1,93 @@
...
+		x11-base/xorg-server[xephyr,xvfb]
+		x11-misc/xvfb-run
+	)
+"
+BDEPEND="app-text/asciidoc
+	app-text/xmlto
+	dev-lang/perl
+	virtual/pkgconfig
+"
+RDEPEND="${CDEPEND}
```
juippis: And here.

x11-wm/i3-gaps/i3-gaps-4.19.ebuild
```diff
@@ -0,0 +1,93 @@
...
+}
+
+pkg_postinst() {
+	einfo "There are several packages that you may find useful with ${PN} and"
+	einfo "their usage is suggested by the upstream maintainers, namely:"
+	einfo "  x11-misc/dmenu"
+	einfo "  x11-misc/i3lock"
+	einfo "  x11-misc/i3status"
+	einfo "Please refer to their description for additional info."
+}
```
juippis: `optfeature` is better because it only prints deps that are missing.

x11-wm/i3-gaps/i3-gaps-4.19.ebuild
```diff
@@ -0,0 +1,93 @@
...
+
+DOCS=( RELEASE-NOTES-$(ver_cut 1-3) )
+
+PATCHES=(
+	"${FILESDIR}/${PN}-4.18-musl.patch"
+)
+
+src_configure() {
+	local emesonargs=(
+		-Ddocdir="/usr/share/doc/${P}"
```
juippis, line 68:
```
 * Messages for package x11-wm/i3-gaps-4.19:

 * 
 *   /usr/share/doc/i3
 * 
```

We use `${PF}` for doc dir, however it doesn't seem to be problem here. This installs an empty dir called `/usr/share/doc/i3`.

mid-kid:
All of the suggested changes should be applied to the baseline `x11-wm/i3` ebuild, too, and both ebuilds synced. There shouldn't be a big difference between two ebuilds for essentially the same software.

juippis:
I can make a PR to sync these ebuilds once this has been merged. The real blocker here is the doc path really, rest I can modify later too. Just want the author to check whether that directory can be somehow dealth with using their build system.
(If not, it can be removed in `src_install` with a comment stating where it's coming from, or patched away in `src_prepare`)

x11-wm/i3-gaps/i3-gaps-4.19.ebuild
ajakk: Reported upstream and I've just pushed a patch here for this ebuild.

https://github.com/i3/i3/issues/4268

juippis (approved):
Tests seem to fail everytime `-doc` is used. With `+doc` they pass. Can you test it?

PR #18312, opened 2020-11-18
*/*: sci@ python updates

jsmolic:
@SoapGentoo


dev-python/line_profiler/line_profiler-3.0.2.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{6,7,8} )
+PYTHON_COMPAT=( python3_{6..9} )
```
jsmolic, line 6: Should we drop 3.6 support or leave it for now? (https://bugs.gentoo.org/754294)
btw I'm not able to hit https://bugs.gentoo.org/742635 right now

SoapGentoo: Never drop 3.6 if it still works, it makes handling the depgraph and transitions harder. I know it's mostly dead at this point, but removing 3.6 prematurely causes more problems than it solves.

jsmolic: Yeah, I thought the same, just wanted to check to make sure

PR #18315, opened 2020-11-18
sys-devel/gcc-config: Sync from 9999 ebuild

m-gupta:
Sync from 9999 ebuild to bring in new USE flag cc-wrappers.


trofi:
Ah, good catch! I completely forgot about it.

I also USE-force'd the flag for most users: https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=63b46c711f9b64d7300e36428a55759b595d7b22. You might need to unmask it in local profile override similar to `native-symlinks`: https://wiki.gentoo.org/wiki/Project:Toolchain/use_native_symlinks#Removing_native_symlinks

I hope it's usable for you in this form.

m-gupta:
Thanks for merging 👍 .

PR #18321, opened 2020-11-19
media-radio/KochMorse: Remove myself from maintainers

rseichter:
I no longer use this software.


monsieurp (changes requested):
You have to move the package to maintainer-needed and get someone from the proxy-maint team to send a last-rites emails on your behalf. If you're really not interested at all, you might want to ask for the package to be wiped (removed) from the tree.

rseichter:
Does this really need to be last-rited? There are no open bug reports, and somebody else might find this package useful.

juippis (approved):
This is fine, packages-up-for-grabs e-mail has also been sent. Thanks!

PR #18326, opened 2020-11-19
x11-terms/st: install terminfo files

ferki:
This PR contains two logical changes:

- adds a new x11-terms/st-terminfo ebuild to install terminfo files for st
- makes x11-terms/st depend on the above ebuild to make sure terminfo files are installed when st is present

As previously described on [bug 592332](https://bugs.gentoo.org/592332#c3), this approach is inspired by x11-terms/kitty-terminfo, and makes it possible to install the terminfo files without st itself (which might be useful for administrators who would like to support their st-users logging in to their machines remotely, but don't need st itself on a machine).


x11-terms/st-terminfo/st-terminfo-0.8.4.ebuild
```diff
@@ -0,0 +1,48 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~hppa ~ppc64 ~x86"
+RESTRICT="test"
+
+DEPEND=">=sys-libs/ncurses-6.0:0="
+
+S="${WORKDIR}/${MY_P}"
+
+src_prepare() {
+	mkdir terminfo
```
gyakovlev, line 23: add ` || die`

x11-terms/st-terminfo/st-terminfo-0.8.4.ebuild
```diff
@@ -0,0 +1,48 @@
...
+	default
+}
+
+src_configure() {
+	return
+}
+
+src_compile() {
+	tic -sxo terminfo st.info
```
gyakovlev, line 32: add ` || die`

x11-terms/st-terminfo/st-terminfo-0.8.4.ebuild
```diff
@@ -0,0 +1,48 @@
...
+
+S="${WORKDIR}/${MY_P}"
+
+src_prepare() {
+	mkdir terminfo
+	default
+}
+
+src_configure() {
+	return
```
gyakovlev, line 28: we usually use `:` as shell true in such cases instead of return.

gyakovlev (changes requested):
thanks and sorry for missing it.
I inherited this package from another developer but have forgotten to set up monitoring of bugs/issues/prs for it.

I thought new ncurses should contain `st` terminfo already, or something changed?

x11-terms/st-terminfo/st-terminfo-0.8.4.ebuild
ferki: Done!

x11-terms/st-terminfo/st-terminfo-0.8.4.ebuild
ferki: Done!

x11-terms/st-terminfo/st-terminfo-0.8.4.ebuild
ferki: Replaced!

ferki:
> thanks and sorry for missing it.
> I inherited this package from another developer but have forgotten to set up monitoring of bugs/issues/prs for it.

Thanks for coming back to me about this!

> I thought new ncurses should contain `st` terminfo already, or something changed?

As I was trying to explain in this [bugzilla comment](https://bugs.gentoo.org/592332#c2), there are various inevitable delays before a new terminfo lands on the systems of end users through ncurses after a new st release.

As far as I understand today's stable ncurses-6.2.-r1 (and upstream's git HEAD) only has terminfo files from st-0.8.2, which was released on 2019-02-09. At the same time, st-0.8.4 is the latest version as of now (released on 2020-06-09), containing 7 changesets to the terminfo files compared to 0.8.2.

That means terminfo files for st through ncurses are lagging behind more than two years and 7 changesets already, and that's exactly what this ebuild is supposed to solve by decoupling the st terminfo update process from ncurses development cycles.

I'm using this approach via my overlay for almost a year now, and I feel it's working well. I hope you would also agree it would be nice to enable all Gentoo users to gain the benefits of it.

gyakovlev:
I lost track if this PR again, sorry.
but finally change is in, starting with version 0.8.5 it will pull standalone terminfo, which also can be installed somewhere else without pulling st.

ferki:
@gyakovlev: thank for coming back and merging this idea! :heart: 

Funnily enough, I have just bumped the versions in my overlay about an hour ago, and planned to update this PR; but I don't have to do that anymore :smile:

PR #18329, opened 2020-11-19
dev-vcs/git-cola: verbump to 3.8

stkw0:


dev-vcs/git-cola/git-cola-3.8.ebuild
```diff
@@ -0,0 +1,109 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8,9} )
+DISTUTILS_SINGLE_IMPL=true
```
bearsh, line 7: DISTUTILS_USE_SETUPTOOLS=no  (although I'm not sure if this is still needed)

dev-vcs/git-cola/git-cola-3.8.ebuild
```diff
@@ -0,0 +1,109 @@
...
+python_install_all() {
+	cd share/doc/${PN}/ || die
+	emake \
+		DESTDIR="${D}" \
+		docdir="${EPREFIX}/usr/share/doc/${PF}" \
+		prefix="${EPREFIX}/usr" \
+		install
+
+	python_fix_shebang "${ED}/usr/bin/git-cola" "${ED}"/usr/bin/git-cola
+	python_optimize "${ED}/usr/share/git-cola/lib/cola"
```
bearsh, line 95: I think `python_fix_shebang` and `python_optimize` is not needed anymore

dev-vcs/git-cola/git-cola-3.8.ebuild
```diff
@@ -0,0 +1,109 @@
...
+		cola/resources.py || die
+
+	# fix ssh-askpass directory reference
+	sed -i -e 's/resources\.share/resources\.prefix/' cola/app.py || die
+
+	distutils-r1_python_prepare_all
+}
+
+python_configure_all() {
+	mydistutilsargs=( --no-vendor-libs )
```
bearsh, line 60: maybe also pass `--no-private-libs`

bearsh (commented):
changes I have in my version which could also apply here...

dev-vcs/git-cola/git-cola-3.8.ebuild
stkw0: Seems that distutils was replacing "env python" with a hardcoded path with python3.7 (in my case) and python_fix_shebang was not fixing it, the output was the same. I have no idea why it doesn't works as expected but I will replace it with the "--no-env-python" flag.

bearsh: I think it's not needed at all as git-cola's installation method changed, it's now install in python's `site-packages`, see also https://github.com/git-cola/git-cola/blob/main/CHANGES.rst#v38

bearsh: > Seems that distutils was replacing "env python" with a hardcoded path with python3.7 (in my case)

isn't that what distutils-r1 in single implementation mode is supposed to do?

stkw0: > > Seems that distutils was replacing "env python" with a hardcoded path with python3.7 (in my case)
> 
> isn't that what distutils-r1 in single implementation mode is supposed to do?

Yes, but python_fix_shebang should change it to env python again I think

stkw0:
@bearsh thank you for your recommendations. It works now. btw, as you already worked on git-cola ebuilds, do you want to maintain this package?

bearsh:
> @bearsh thank you for your recommendations. It works now. btw, as you already worked on git-cola ebuilds, do you want to maintain this package?

no as a sole maintainer but as a co-maintainer would be ok...

stkw0:
> > @bearsh thank you for your recommendations. It works now. btw, as you already worked on git-cola ebuilds, do you want to maintain this package?
> 
> no as a sole maintainer but as a co-maintainer would be ok...

I added you to the metadata.xml

bearsh:
@stkw0 just for your information, I've opened https://bugs.gentoo.org/756088 to get rid of the PyQt5 dep by requesting a network useflag for QtPy

stkw0:
> @stkw0 just for your information, I've opened https://bugs.gentoo.org/756088 to get rid of the PyQt5 dep by requesting a network useflag for QtPy

Nice, I added myself to the CC list to be aware of any change.

dev-vcs/git-cola/git-cola-3.8.ebuild
```diff
@@ -0,0 +1,107 @@
...
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="doc test"
+
+RESTRICT="!test? ( test )"
+
+RDEPEND="
+	$(python_gen_cond_dep '
```
juippis, line 22: Did you mean to use python-single implementation on this ebuild? This gen_cond is not necessary with pure distutils.

dev-vcs/git-cola/git-cola-3.8.ebuild
```diff
@@ -0,0 +1,107 @@
...
+		dev-python/numpy[${PYTHON_USEDEP}]
+		dev-python/pygments[${PYTHON_USEDEP}]
+		dev-python/PyQt5[network,${PYTHON_USEDEP}]
+		dev-python/QtPy[gui,${PYTHON_USEDEP}]
+		dev-python/send2trash[${PYTHON_USEDEP}]
+	')
+	dev-vcs/git"
+BDEPEND="sys-devel/gettext
+	$(python_gen_cond_dep "
```
juippis, line 31: Here too.

dev-vcs/git-cola/git-cola-3.8.ebuild
```diff
@@ -0,0 +1,107 @@
...
+		doc? ( dev-python/sphinx[\${PYTHON_USEDEP}] )
```
juippis, line 32: You should use `distutils_enable_sphinx` to control this dep.

dev-vcs/git-cola/git-cola-3.8.ebuild
```diff
@@ -0,0 +1,107 @@
...
+		test? (
+			${VIRTUALX_DEPEND}
+			dev-python/mock[\${PYTHON_USEDEP}]
+			dev-python/nose[\${PYTHON_USEDEP}]
+		)
```
juippis, line 37:
```
ERROR: Failure: ModuleNotFoundError (No module named 'pytest')

    import pytest
ModuleNotFoundError: No module named 'pytest'

Ran 170 tests in 6.493s

FAILED (errors=1)
 * ERROR: dev-vcs/git-cola-3.8::testworld failed (test phase):
 *   Failed to run 'nosetests --verbose --with-id --with-doctest --exclude=sphinxtogithub'
```
```

dev-vcs/git-cola/git-cola-3.8.ebuild
bearsh: thanks @juippis, maybe the doc at https://dev.gentoo.org/~mgorny/python-guide/distutils.html#dependencies should be adjusted then, it states 

> Dependencies on Python packages are declared using the same method as the underlying eclass — that is, python-r1 or python-single-r1.

 and also the example uses `python_gen_cond_dep`

juippis: Yep, `DISTUTILS_SINGLE_IMPL` is not declared here, so it uses the functionality from *python-r1.eclass*. So now you want to be looking at 
https://dev.gentoo.org/~mgorny/python-guide/multi.html#dependencies
instead. The example you linked, is correct when using *python-single-r1.eclass*, or when `DISTUTILS_SINGLE_IMPL` is declared with *distutils-r1.eclass*. That's why I asked what is the intention with this package, is it supposed to work multi-impl, or single-impl only?

bearsh: ok, sorry I had _my version_ (from my overlay) in mind where DISTUTILS_SINGLE_IMPL is set and implied it is also set here.
I think setting this / make it single-impl makes sense for this package...

dev-vcs/git-cola/git-cola-3.8.ebuild
bearsh: it this supposed to work with `DISTUTILS_SINGLE_IMPL` set? 

>  * External commands disallowed while sourcing ebuild: python_gen_any_dep 
>  *                      dev-python/sphinx[${PYTHON_USEDEP}]

I think I'll create a new PR with the whole ebuild converted to `DISTUTILS_SINGLE_IMPL=true`

stkw0:
I am closing this as #18445 already fixed some of the requested changes and replaces this PR

PR #18331, opened 2020-11-19
Haskell: bump yesod-core, drop old

solpeth:
Tested in an `~amd64` container. Pulled from my commits in `::haskell`.


trofi:
Pulled. Thank you!

PR #18345, opened 2020-11-21
sys-apps/firejail: restored removed USE flag description

hlein:
Restored and renamed the old 'x11' USE flag description now that that
flag has been renamed to 'X'.


hlein:
ping @juippis because I'm afraid this will get lost otherwise. You correctly saw `USE=x11` ought to be renamed `USE=X` and did so when merging another PR for me.  ...But since the `metadata.xml` reference didn't get updated, you then removed the flag description a few hours ago :-P

juippis (approved):

hlein:
Thanks!

PR #18346, opened 2020-11-21
dev-python/pyusb: version bump to 1.1.0, support PyPy3+Python 3.9

wimmuskee:
updated github id in metadata


wimmuskee:
added with https://github.com/gentoo/gentoo/commit/75eec14cda77354d357710efacf66a8d4839e00c

PR #18347, opened 2020-11-21
Net irc/inspircd 3.8.1

clinew:
Version bump for CVE issue & bug https://bugs.gentoo.org/show_bug.cgi?id=755854

I didn't add the bug tag because this alone doesn't fix the issue in the older, stabilized build; the newer version will need stabilization.


clinew:
Fixed commit message and added bug tags...

thesamesam:
Use a Bug tag in future so we get a notification when the commit lands, just not Closes, but thank you!

PR #18352, opened 2020-11-21
dev-python/pynput: version bump pynput-1.7.1

bdouxx:
version required for recent version of app-misc/solaar

new dependency : dev-python/python-evdev
new PYTHON_COMPAT : python3_9

Signed-off-by: Etienne d'Hautefeuille bdouxx-git@yahoo.fr


thesamesam:
Newer version already in tree.

PR #18364, opened 2020-11-22
app-i18n/transifex-client: add new version 0.14.2

robert7k:


app-i18n/transifex-client/transifex-client-0.14.2.ebuild
```diff
@@ -0,0 +1,34 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{6,7,8} )
```
juippis: Can you try to add 3.9?

app-i18n/transifex-client/transifex-client-0.14.2.ebuild
```diff
@@ -0,0 +1,34 @@
...
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="test"
+RESTRICT="!test? ( test )"
+
+BDEPEND="test? ( dev-python/mock[${PYTHON_USEDEP}] )"
+RDEPEND="dev-python/GitPython[${PYTHON_USEDEP}]
+	<dev-python/python-slugify-5.0.0[${PYTHON_USEDEP}]
```
juippis: Could you verify this...

app-i18n/transifex-client/transifex-client-0.14.2.ebuild
```diff
@@ -0,0 +1,34 @@
...
+	dev-python/requests[${PYTHON_USEDEP}]
+	<dev-python/six-2.0.0[${PYTHON_USEDEP}]
```
juippis: ... and this, and if upstream file says these are required, could you patch off these restrictions and run python tests?

juippis (approved):

app-i18n/transifex-client/transifex-client-0.14.2.ebuild
robert7k: Done in #18603

PR #18365, opened 2020-11-22
Convert sddm to acct eclass

floppym:


a17r:
Thanks, but this is already in preparation including reserved UID/GID.

floppym:
> Thanks, but this is already in preparation including reserved UID/GID.

Where is this being prepared? I looked in bugzilla and in the kde overlay.

a17r:
Not very discoverable, should make a PR at some point: https://github.com/a17r/gentoo/tree/sddm-glep81

A bit stalled currently due to https://github.com/sddm/sddm/issues/1316

Not decided yet whether we shouldn't try and drop `video` membership and hard-depend on `pam` instead.

floppym:
Ok. Do you want to close this then?

a17r:
Yes please, I'll make sure to add you for my upcoming PR review then.

PR #18367, opened 2020-11-22
media-gfx/alembic: bump to 1.7.16

waebbl:
Second commit drops old 1.7.11-r1l

Closes: https://bugs.gentoo.org/733456


media-gfx/alembic/alembic-1.7.16.ebuild
```diff
@@ -0,0 +1,59 @@
...
+	local mycmakeargs=(
+		-DALEMBIC_SHARED_LIBS=ON
+		# C++-11 and thus {shared,unique,weak}_ptr are common nowadays, so these
+		# are no longer needed and using boost fails. Options will get removed by
+		# upstream soon
+		-DALEMBIC_LIB_USES_BOOST=OFF
+		-DALEMBIC_LIB_USES_TR1=OFF
+		-DUSE_ARNOLD=OFF
+		-DUSE_BINARIES=ON
+		-DUSE_EXAMPLES=$(usex examples)
```
juippis, line 42: Could you check whether this works or not, the only "possible" files I see related to this installed are
```
/usr/lib64/cmake/Alembic/AlembicConfig.cmake
/usr/lib64/cmake/Alembic/AlembicConfigVersion.cmake
/usr/lib64/cmake/Alembic/AlembicTargets-gentoo.cmake
/usr/lib64/cmake/Alembic/AlembicTargets.cmake
```
which seem weird to justify a USE flag.

media-gfx/alembic/alembic-1.7.16.ebuild
```diff
@@ -0,0 +1,59 @@
...
+
+	cmake_src_configure
+}
+
+pkg_postinst() {
+	if use prman; then
+		einfo "If you're looking for an ebuild for renderman, you may want to"
+		einfo "try the waebbl overlay: 'eselect repository enable waebbl'"
+		einfo "followed by 'emerge renderman'"
+	fi
```
juippis, line 58: Want to package that for ::gentoo as well?

juippis (approved):

media-gfx/alembic/alembic-1.7.16.ebuild
waebbl: IIRC it builds only one additional executable file which doesn't get installed. I agree this will not justify a USE flag. I remember this were more examples in the past, and/or they also got installed.
It's a good idea to remove the flag.

media-gfx/alembic/alembic-1.7.16.ebuild
waebbl: I thought about this several times. But the package has several characteristics, so I decided against it so far:
- it's a binary only package
- fetch restricted, user has to register and download manually
- license restricted to one year, use has to re-license every year
- blender integration, which was the reason, I build the package at first, is far behind blender development, currently there is only a renderman for blender-2.79, a version for 2.8+ is questionable, just checked their forum
- is there demand for a build of the package?
Probably a good idea to remove the flag too. What do you think?

juippis: We've probably seen worse, but it does look like a mess I wouldn't want to welcome after all :P

waebbl: Yes it is. I'm removing the flag for it.

PR #18371, opened 2020-11-23
mail-client/neomutt: fix live ebuild src dir

Austin-Ray:
Commit e4ef3e09eaf dropped the assignment of "S", changing its value from
"${WORKDIR}/${PN}-${P}" to the default value of "${WORKDIR}/${P}".
However, the live ebuild still checkouts out source code to previous
location so portage says the source directory doesn't exist.

This commit drops the custom checkout directory and uses the default
value provided, which corresponds to the default value of "S".

Closes: https://bugs.gentoo.org/756166

Signed-off-by: Austin Ray <austin@austinray.io>


nicolasbock:
Thanks @Austin-Ray ! Merged in bff7b7f99b533258eb6c5094de4f415b03fceb53.

PR #18375, opened 2020-11-23
media-sound/pulseaudio fixes

Polynomial-C:
The first commit fixes [bug #756196](https://bugs.gentoo.org/756196) by correctly quoting `SYSTEMDLOGIN_CFLAGS` and `SYSTEMDLOGIN_LIBS`.
The second commit fixes configure with non-bash shell (two bashisms haven been removed).


leio:
The quoting fix was merged, and the bashism fix isn't applicable (yet), as the option and problem isn't there in 13.0 release

PR #18377, opened 2020-11-23
lxqt-base/lxqt-powermanagement: add a missing dependency

Chiitoo:


lxqt-base/lxqt-powermanagement/lxqt-powermanagement-0.16.0-r1.ebuild
```diff
@@ -8,6 +8,8 @@ inherit cmake
 DESCRIPTION="LXQt daemon for power management and auto-suspend"
 HOMEPAGE="https://lxqt.github.io/"
 
+MY_PV="$(ver_cut 1-2)*"
```
a17r, line 11: hm, that's a bit oddly named when not used for the ebuild itself. how about `LXQTPV' instead or similar?

Chiitoo: Works for me.  I just copied this from one of the other ebuilds, for which I could change it probably the next time I make any changes to them.

a17r (approved):

PR #18383, opened 2020-11-24
net-dns/dnscrypt-proxy log rotation needs a hard restart

ftobin:
There is a race condition where copytruncate  causes dnscrypt-proxy to end up with a logfile it is writing to that is not truncated, since it could be in the middle of a write (dnscrypt-proxy's file descriptor is still to the middle of the logfile, I believe).  The logfile ends up having a bunch of null data at the beginning as it continues to write to it.  This has happened multiples times on different machines for me.  It's much safer to do a hard restart after a normal rotation.

### DIFF
--- a/net-dns/dnscrypt-proxy/files/dnscrypt-proxy.logrotate
+++ b/net-dns/dnscrypt-proxy/files/dnscrypt-proxy.logrotate
@@ -1,6 +1,9 @@
 /var/log/dnscrypt-proxy/*.log {
 	su dnscrypt-proxy dnscrypt-proxy
-	copytruncate
 	missingok
 	notifempty
+        sharedscripts
+        postrotate
+                /etc/init.d/dnscrypt-proxy restart > /dev/null
+        endscript
 }


gyakovlev:
restarting invalidates cache, which is not nice. I don't remember, does it handle `-HUP`? In that case initrscript should be updated to provide reload and logrotate should use that.

ftobin:
It does not handle any signals like many daemons do (it tries to be cross-platform, and apparently signal handling on Windows is tough; see https://github.com/DNSCrypt/dnscrypt-proxy/issues/278).  I agree that cache invalidation is not terrific, but caches shouldn't be *that* long relative to the lifetime of a logfile.

ftobin:
I'll also suggest that the cost of invalidating the cache is outweighed by the logfile corruption.

fkobi:
@ftobin seeing as it got no response could you please open a bug instead of this PR?

thesamesam:
Merged finally, I'm sorry I took so long to get to it, I'd completely forgot.

PR #18385, opened 2020-11-24
app-emulation/qemu: Work around libusb bug

zippy2:
In its 6.9.0 relelase Libvirt changed way how USB devices are
specified on QEMU's command line. While previously .hostbus= and
.hostaddr= were used they were not really good fit and often get
in way of another libvirt feature - namespaces (see
v6.9.0-rc1~456 libvirt commit for more info). The new way of
configuring USB devices is .hostdevice=/dev/bus/usb/.../... but
it exposed a bug in libusb which is worked around in qemu (see
v5.1.0-599-g202d69a715 QEMU commit for more info). Backport the
workaround for the time being. It can be dropped when we rebase
to 5.2.0.

Closes: https://bugs.gentoo.org/753305


zippy2:
If somebody could fix the title of https://bugs.gentoo.org/753305 that would be great. I don't have enough permissions to do that myself.

trofi:
Pulled. Thank you!

PR #18388, opened 2020-11-24
dev-python/confuse: verbump to 1.4.0

stkw0:


dev-python/confuse/confuse-1.4.0.ebuild
```diff
@@ -0,0 +1,44 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8,9} pypy3 )
+inherit distutils-r1
+
+DISTUTILS_USE_SETUPTOOLS=pyproject.toml
```
juippis, line 9: AFAIK this needs to be set before `inherit distutils-r1`.

dev-python/confuse/confuse-1.4.0.ebuild
```diff
@@ -0,0 +1,44 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="doc test"
+RESTRICT="!test? ( test )"
+
+BDEPEND="
+	dev-python/pyproject2setuppy[${PYTHON_USEDEP}]
+	dev-python/setuptools[${PYTHON_USEDEP}]
+	doc? ( dev-python/sphinx )
+	test? ( dev-python/nose[${PYTHON_USEDEP}] )
```
juippis, line 24: This isn't wrong, but you can also use `distutils_enable_tests` which adds many other things you've already specified in this ebuild.
https://dev.gentoo.org/~mgorny/python-guide/distutils.html#enabling-tests

dev-python/confuse/confuse-1.4.0.ebuild
```diff
@@ -0,0 +1,44 @@
...
+	dev-python/pyyaml[${PYTHON_USEDEP}]
+"
+RDEPEND="${DEPEND}"
+
+python_compile_all() {
+	if use doc; then
+		emake -C docs html
+		rm -r docs/_build/html/_sources || die
+		HTML_DOCS=( docs/_build/html/. )
+	fi
```
juippis, line 36:
```
make: Entering directory '/var/tmp/portage/gpm/portage/dev-python/confuse-1.4.0/work/confuse-1.4.0/docs'
make: *** No rule to make target 'html'.  Stop.
make: Leaving directory '/var/tmp/portage/gpm/portage/dev-python/confuse-1.4.0/work/confuse-1.4.0/docs'
 * ERROR: dev-python/confuse-1.4.0::testworld failed (compile phase):
 *   emake failed
```

If it's simple like this, you should be able to do it with `distutils_enable_sphinx`, https://dev.gentoo.org/~mgorny/python-guide/distutils.html#building-documentation-via-sphinx

dev-python/confuse/confuse-1.4.0.ebuild
stkw0: Using distutils the result is much more clean :)

juippis (approved):

PR #18395, opened 2020-11-25
net-misc/anydesk: Remove x11-libs/pangox-compat dependency

PPed72:
Closes: https://bugs.gentoo.org/756613

### DIFF
--- a/net-misc/anydesk/anydesk-6.0.1-r1.ebuild
+++ b/net-misc/anydesk/anydesk-6.0.1-r1.ebuild
@@ -3,7 +3,7 @@
 
 EAPI="7"
 
-inherit desktop eutils systemd xdg-utils
+inherit desktop optfeature systemd xdg-utils
 
 DESCRIPTION="Feature rich multi-platform remote desktop application"
 HOMEPAGE="https://anydesk.com"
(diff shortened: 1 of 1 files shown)


net-misc/anydesk/anydesk-6.0.1-r1.ebuild
```diff
@@ -58,10 +58,12 @@ QA_PREBUILT="opt/${PN}/*"
 src_install() {
 	local dst="/opt/${PN}"
 
-	dodir ${dst}
 	exeinto ${dst}
 	doexe ${PN}
 
+	# bug 706344
+	patchelf --remove-needed libpangox-1.0.so.0 "${ED}"${dst}/${PN} || die
```
telans, line 65:
```suggestion
	patchelf --remove-needed libpangox-1.0.so.0 "${ED}${dst}/${PN}" || die
```

This should all be quoted (along with everything else in the ebuild), right?

PPed72: I just copied that from 6.0.0-r1 ebuild. Will fix it as you indicated.

PPed72:
@telans : fixed quotations. Thanks for the tip.

a17r (changes requested):
Thanks for your fix. Don't be surprised that I dropped the revbump: No one was able to build this package yet because of the inexistent dependency.

a17r (approved):

PR #18406, opened 2020-11-25
Haskell: bump conduit, drop old

solpeth:
Tested in an `~amd64` container (awaiting successful rebuild of revdeps).


trofi:
Pulled as is. Thank you!

PR #18408, opened 2020-11-26
sci-libs/med: bump to 4.1.0

waebbl:
Closes: https://bugs.gentoo.org/756208

Fortran support is no longer a cmake option by upstream, but an automagic dependency, so I removed the USE flag.

Second commit drops old 4.0.0-r1


juippis (commented):
> Fortran support is no longer a cmake option by upstream, but an automagic dependency, so I removed the USE flag.

Could you try to patch this, and possible any other automagic deps, to be conditional again? We really hate these automagic deps in Gentoo as they're unpredictable.
See, 
https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies#CMake
https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/check/files/check-0.14.0-r2-disable-automagic-dep.patch
for examples.

Rest looks good.

waebbl:
I see what I can do

waebbl:
@juippis I was able to re-add an option into the CMakeLists.txt file which works as intended.
Other than that, I haven't noticed any other automagic dependencies.

juippis (approved):
Thanks, seems to work as intended!

PR #18410, opened 2020-11-26
Asterisk updates

jkroonza:


Whissi:
Looks like you maintain multiple branches. Could you please introduce sub slots like I demonstrated in bcec778fe190845b5bfa79a3979306c537d0b6fa? This would allow us to target these version in GLSAs.

PR #18412, opened 2020-11-26
net-misc/netkit-telnetd: change sed delimiter

PPed72:
Bug: https://bugs.gentoo.org/715706


juippis (approved):
Weird, never seen `/` break sed.

PR #18415, opened 2020-11-26
app-emulation/crun: bump to 0.16

robertgzr:
Signed-off-by: Robert Günzler <r@gnzler.io>


robertgzr:
@juippis let me know when it's time to drop older version... I was thinking of taking 0.14 out or at least replacing 0.15 with 0.15.1

juippis:
If it were me, I'd just have 0.16 in tree and delete everything else. If there's nothing depending on 0.15 series, people will install 0.16 by default anyway since they all have matching KEYWORDS.

So can we not at least add 0.15.1?

app-emulation/crun/crun-0.16.ebuild
```diff
@@ -0,0 +1,68 @@
...
+DOCS=( README.md )
+
+src_configure() {
+	local myeconfargs=(
+		$(use_enable bpf) \
+		$(use_enable caps) \
+		$(use_enable criu) \
+		$(use_enable seccomp) \
+		$(use_enable systemd) \
+		$(usex static-libs '--enabled-shared  --enabled-static' '--enable-shared --disable-static' '' '')
```
juippis: This doesn't seem to work, when `+static-libs`:
```
 * econf: updating crun-0.16/build-aux/config.guess with /usr/share/gnuconfig/config.guess
./configure --prefix=/usr --build=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu --mandir=/usr/share/man --infodir=/usr/share/info --datadir=/usr/share --sysconfdir=/etc --localstatedir=/var/lib --disable-dependency-tracking --disable-silent-rules --docdir=/usr/share/doc/crun-0.16 --htmldir=/usr/share/doc/crun-0.16/html --with-sysroot=/ --libdir=/usr/lib64 --enable-bpf --disable-caps --enable-criu --disable-seccomp --disable-systemd --enabled-shared --enabled-static
configure: error: unrecognized option: `--enabled-shared'
Try `./configure --help' for more information
 * ERROR: app-emulation/crun-0.16::testworld failed (configure phase):
 *   econf failed
 * 
```

juippis (commented):
Also don't push 0.15.1, as it has same KEYWORDS as 0.16 so probably no one will end up installing it. Rather just leave the 0.15 in tree for people to go back if they can't compile 0.16.

robertgzr:
alright, sorry for the long silence...

- kept 0.15 unchanged
- dropped 0.14.1
- introduced 0.16
  - with the static-build fixes and
  - dropping the `man` use flag (thanks @epsilon-0)

app-emulation/crun/crun-0.16.ebuild
```diff
@@ -0,0 +1,61 @@
...
+inherit autotools python-any-r1
+
+DESCRIPTION="A fast and low-memory footprint OCI Container Runtime fully written in C"
+HOMEPAGE="https://github.com/containers/crun"
+SRC_URI="https://github.com/containers/${PN}/releases/download/${PV}/${P}.tar.gz"
+
+LICENSE="GPL-2+ LGPL-2.1+"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64"
+IUSE="+bpf +caps criu +seccomp systemd static-libs"
```
juippis, line 17: Btw we are trying to get rid of `static-libs` where they're not needed.

app-emulation/crun/crun-0.16.ebuild
```diff
@@ -0,0 +1,61 @@
...
+DOCS=( README.md )
+
+src_configure() {
+	local myeconfargs=(
+		$(use_enable bpf) \
+		$(use_enable caps) \
+		$(use_enable criu) \
+		$(use_enable seccomp) \
+		$(use_enable systemd) \
+		$(usex static-libs '--enable-shared --enable-static' '--enable-shared --disable-static' '' '')
```
juippis: Backlashes not needed with array, one of the arrays strong points. Will clean after you.

juippis (approved):
```
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: ./.libs/libcrun.so: undefined reference to `seccomp_init'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: ./.libs/libcrun.so: undefined reference to `seccomp_rule_add'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: ./.libs/libcrun.so: undefined reference to `seccomp_arch_add'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: ./.libs/libcrun.so: undefined reference to `seccomp_export_bpf'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: ./.libs/libcrun.so: undefined reference to `seccomp_release'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: ./.libs/libcrun.so: undefined reference to `seccomp_rule_add_array'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: ./.libs/libcrun.so: undefined reference to `seccomp_syscall_resolve_name'
make: *** [Makefile:1143: crun] Error 1
 * ERROR: app-emulation/crun-0.16::testworld failed (compile phase):
 *   emake failed
```
Is this just my system? This doesn't seem to emerge with `-seccomp` USE at all.

app-emulation/crun/crun-0.16.ebuild
robertgzr: thanks o/

app-emulation/crun/crun-0.16.ebuild
robertgzr: what's the reason? I always thought it a strong point of gentoo to be able to get static libraries easily when I need them...

juippis: https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0302

PR #18418, opened 2020-11-26
media-libs/osl: Version bump to 1.10.13

redchillipadi:
This version works with openimageio 2 (fixes build failure of 1.10.10)

Add support for LLVM 10, which requires C++14

Enable shared libraries
Add github page as second homepage

Added subslot operator otherwise upgrading to 1.11 will leave behind
preserved libraries (blender is the only current client in the tree)

Thanks to Mehw, Cyrillic, Peter Karlsson, Reggie, Luke A Guest, and
Joonas Niilola for their contributions to the bugs mentioned below
which forms the bulk of this update.

Closes: https://bugs.gentoo.org/735750
Closes: https://bugs.gentoo.org/751580


juippis (approved):
Would osl-1.11 support llvm-11 already?

redchillipadi:
I tried osl-1.11 with llvm 11 but an assert in LLVM_Util::call_function in src/liboslexec/llvm_util.cpp kept triggering.

PR #18420, opened 2020-11-27
www-client/otter: Added support for qtwebengine and qtwebkit

JellyBrick:
Otter can use QtWebkit and QtWebEngine at the same time.

### DIFF
--- a/www-client/otter/metadata.xml
+++ b/www-client/otter/metadata.xml
@@ -2,6 +2,10 @@
 <!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
 <pkgmetadata>
 	<!-- maintainer-needed -->
+	<use>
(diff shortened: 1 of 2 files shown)


Chiitoo:
Thanks!

Howver, QtWebKit is kind of on its way out at this time, which is why it was removed from this ebuild (I believe).

JellyBrick:
> Thanks!
> 
> However, QtWebKit is kind of on its way out at this time, which is why it was removed from this ebuild (I believe).

But as long as Otter supports QtWebkit, I think to should be able to use QtWebkit as a backend.

juippis:
I can't even compile this with `+qtwebkit`. I really don't see a point keeping this package in gentoo, it's supposed to be in high-development mode yet this release is from 2 years ago. That's not very convincing in browser market right now.

a17r:
I didn't file https://bugs.gentoo.org/712172 out of the blue. QtWebKit must not be used for browsers.

PR #18424, opened 2020-11-27
dev-lang/php: version bump to 8.0.0

dcolt:


dcolt:
The json flag was dropped as it's now a required feature of PHP 8.0.0, it's no longer an optional extension

Whissi:
Thank you but there is more to do than just cp'ing ebuild of previous version. PHP team will take care of that but please be patient and give us a little bit time.

PR #18427, opened 2020-11-27
dev-lang/ispc: Iscp only supports up to LLVM 10

redchillipadi:
Iscp does not yet have llvm 11 support, so limit clang and llvm in both
ebuilds to version 10 maximum.

Thanks to Toralf Förster, brothermechanic and Marco Genasci for their
contributions to fixing this bug, I am only repackaging it.

A working ispc is needed for adding embree support to blender.

See https://github.com/ispc/ispc/issues/1896

Closes: https://bugs.gentoo.org/749573

### DIFF
--- a/dev-lang/ispc/ispc-1.14.1.ebuild
+++ b/dev-lang/ispc/ispc-1.14.1.ebuild
@@ -24,7 +24,7 @@ LICENSE="BSD BSD-2 UoI-NCSA"
 SLOT="0"
 IUSE="examples"
 
-RDEPEND="<sys-devel/clang-11:=[llvm_targets_AMDGPU(+)]"
+RDEPEND="<sys-devel/clang-11:="
 
 DEPEND="
 	${RDEPEND}
@@ -42,7 +42,7 @@ PATCHES=(
 )
 
 llvm_check_deps() {
-	has_version -d "sys-devel/clang:${LLVM_SLOT}[llvm_targets_AMDGPU(+)]"
+	has_version -d "sys-devel/clang:${LLVM_SLOT}"
 }
 
 src_prepare() {
--- a/dev-lang/ispc/ispc-1.14.1.ebuild
+++ b/dev-lang/ispc/ispc-1.14.1.ebuild
@@ -5,7 +5,9 @@ EAPI=7
 
 PYTHON_COMPAT=( python3_{6,7,8,9} )
 
-inherit cmake toolchain-funcs python-any-r1
+inherit cmake toolchain-funcs python-any-r1 llvm
+
+LLVM_MAX_SLOT=10
 
 DESCRIPTION="Intel SPMD Program Compiler"
 HOMEPAGE="https://ispc.github.com/"
@@ -22,10 +24,8 @@ LICENSE="BSD BSD-2 UoI-NCSA"
 SLOT="0"
 IUSE="examples"
 
-RDEPEND="
-	>=sys-devel/clang-3.0:*
-	>=sys-devel/llvm-3.0:*
-	"
+RDEPEND="<sys-devel/clang-11:=[llvm_targets_AMDGPU(+)]"
+
 DEPEND="
 	${RDEPEND}
 	${PYTHON_DEPS}
@@ -40,6 +40,10 @@ PATCHES=(
 	"${FILESDIR}/${PN}-1.14.0-llvm-10.patch"
 	"${FILESDIR}/${PN}-1.13.0-werror.patch"
 )
+
+llvm_check_deps() {
+	has_version -d "sys-devel/clang:${LLVM_SLOT}[llvm_targets_AMDGPU(+)]"
+}
 
 src_prepare() {
 	if use amd64; then
--- a/dev-lang/ispc/ispc-9999.ebuild
+++ b/dev-lang/ispc/ispc-9999.ebuild
@@ -5,7 +5,9 @@ EAPI=7
 
 PYTHON_COMPAT=( python3_{6,7,8,9} )
 
-inherit cmake toolchain-funcs python-any-r1
+inherit cmake toolchain-funcs python-any-r1 llvm
+
+LLVM_MAX_SLOT=10
 
 DESCRIPTION="Intel SPMD Program Compiler"
 HOMEPAGE="https://ispc.github.com/"
@@ -22,10 +24,7 @@ LICENSE="BSD BSD-2 UoI-NCSA"
 SLOT="0"
 IUSE="examples"
 
-RDEPEND="
-	>=sys-devel/clang-3.0:*
-	>=sys-devel/llvm-3.0:*
-	"
+RDEPEND="<sys-devel/clang-11:=[llvm_targets_AMDGPU(+)]"
 DEPEND="
 	${RDEPEND}
 	${PYTHON_DEPS}
@@ -40,6 +39,10 @@ PATCHES=(
 	"${FILESDIR}/${PN}-1.14.0-llvm-10.patch"
 	"${FILESDIR}/${PN}-1.13.0-werror.patch"
 )
+
+llvm_check_deps() {
+	has_version -d "sys-devel/clang:${LLVM_SLOT}[llvm_targets_AMDGPU(+)]"
+}
 
 src_prepare() {
 	if use amd64; then
--- a/dev-lang/ispc/ispc-9999.ebuild
+++ b/dev-lang/ispc/ispc-9999.ebuild
@@ -24,7 +24,7 @@ LICENSE="BSD BSD-2 UoI-NCSA"
 SLOT="0"
 IUSE="examples"
 
-RDEPEND="<sys-devel/clang-11:=[llvm_targets_AMDGPU(+)]"
+RDEPEND="<sys-devel/clang-11:="
 DEPEND="
 	${RDEPEND}
 	${PYTHON_DEPS}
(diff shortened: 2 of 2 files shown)


dev-lang/ispc/ispc-1.14.1.ebuild
```diff
@@ -22,10 +24,8 @@ LICENSE="BSD BSD-2 UoI-NCSA"
 SLOT="0"
 IUSE="examples"
 
-RDEPEND="
-	>=sys-devel/clang-3.0:*
-	>=sys-devel/llvm-3.0:*
-	"
+RDEPEND="<sys-devel/clang-11:=[llvm_targets_AMDGPU(+)]"
```
waebbl, line 27: Why do you pass the llvm_target? I successfully compiled it against clang/llvm without this target defined.

dev-lang/ispc/ispc-1.14.1.ebuild
```diff
@@ -41,6 +41,10 @@ PATCHES=(
 	"${FILESDIR}/${PN}-1.13.0-werror.patch"
 )
 
+llvm_check_deps() {
+	has_version -d "sys-devel/clang:${LLVM_SLOT}[llvm_targets_AMDGPU(+)]"
```
waebbl, line 45: same here

dev-lang/ispc/ispc-9999.ebuild
```diff
@@ -22,10 +24,7 @@ LICENSE="BSD BSD-2 UoI-NCSA"
 SLOT="0"
 IUSE="examples"
 
-RDEPEND="
-	>=sys-devel/clang-3.0:*
-	>=sys-devel/llvm-3.0:*
-	"
+RDEPEND="<sys-devel/clang-11:=[llvm_targets_AMDGPU(+)]"
```
waebbl, line 27: and here

dev-lang/ispc/ispc-9999.ebuild
```diff
@@ -41,6 +40,10 @@ PATCHES=(
 	"${FILESDIR}/${PN}-1.13.0-werror.patch"
 )
 
+llvm_check_deps() {
+	has_version -d "sys-devel/clang:${LLVM_SLOT}[llvm_targets_AMDGPU(+)]"
```
waebbl, line 44: and here

waebbl (commented):
Having llvm/clang 10 and 11 installed, I needed to restrict the `find_package` call in `${WORKDIR}/cmake/LLVMConfig.cmake` to 10. Better would be a version restriction to 9..10, but this needs cmake-3.19 or later and would need to update the `cmake_minimum_required` call too. Else it will use clang-11, even with the modifications from the ebuild. See the attached file for the patch I used.

Additionally, I get build errors for the examples, both when using gcc and when using clang as default compiler. So I needed to add `-DISPC_INCLUDE_EXAMPLES=$(usex examples)` to `src_configure` which is enabled by default, even if the USE flag is unset.

[ispc-1.14.1-0001-restrict-llvm-version.patch.txt](https://github.com/gentoo/gentoo/files/5610501/ispc-1.14.1-0001-restrict-llvm-version.patch.txt)

dev-lang/ispc/ispc-1.14.1.ebuild
redchillipadi: With the current settings ispc needs AVX2 on x86_64 or x86. I thought it was coming from that target, but now I see that it probably should be the x86 target instead, which includes both x86_64 and x86.

redchillipadi:
I have llvm 9, 10 and 11 on my system, and will double check which version is being linked with ispc with and without the patch. There seems to be a push to support llvm 10 at present, so I think hard coding llvm 10 support (without 9) should be OK.

dev-lang/ispc/ispc-1.14.1.ebuild
waebbl: AVX2 seems not strictly needed. I'm having a bulldozer family CPU which doesn't support AVX2 (according to cpuid2cpuflags and /proc/cpuinfo) and the package builds fine.

redchillipadi: Thanks, I think that means we can just go with llvm/clang 10 without the target specified.

Maybe if you don't have AVX2 it automatically builds with SSE4 or something else instead?

Perhaps this is for down the track, but I would be interested to know if you can run open image denoiser with blender on your cpu, or whether it is necessary to add USE flags to control which version is built for either ispc or oidn in order to do so. The install_deps script in blender builds oidn with AVX2 support by default, but I don't know if it a hard requirement or not.

waebbl: It builds for AVX support.

Currently I'm not able to build oidn from your overlay. The denoiser works however with the official binary blender package. The issues are the same, like when I was trying to build the ispc examples and tests and point to an internal ispc compiler error. I will probably report this upstream.

```FAILED: local__core/output_copy.dev.o local__core/output_copy.dev_sse4.o local__core/output_copy.dev_avx2.o local__core/output_copy.dev_avx512skx.o output_copy_ispc.h 
cd /var/tmp/portage/media-libs/oidn-1.2.4/work/oidn-1.2.4_build && /usr/bin/cmake -E make_directory /var/tmp/portage/media-libs/oidn-1.2.4/work/oidn-1.2.4_build/local__core && /usr/bin/ispc -I /var/tmp/portage/media-libs/oidn-1.2.4/work/oidn-1.2.4 --arch=x86-64 --addressing=64 -O3 --target=sse4-i32x8,avx2-i32x8,avx512skx-i32x16 --woff --opt=fast-math --pic --opt=disable-assertions -h /var/tmp/portage/media-libs/oidn-1.2.4/work/oidn-1.2.4_build/output_copy_ispc.h -MMM /var/tmp/portage/media-libs/oidn-1.2.4/work/oidn-1.2.4_build/local__core/output_copy.dev.idep -o /var/tmp/portage/media-libs/oidn-1.2.4/work/oidn-1.2.4_build/local__core/output_copy.dev.o /var/tmp/portage/media-libs/oidn-1.2.4/work/oidn-1.2.4/core/output_copy.ispc
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
/var/tmp/portage/dev-lang/ispc-1.14.1/work/ispc-1.14.1/src/main.cpp(372): FATAL ERROR: Unhandled signal sent to process; terminating.
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
```

This exception throws whenever an ispc program is compiled (ispc examples, tests, or oidn programs using ispc).

I noticed this, because cuda isn't working in blender, neither with the build from your overlay, nor with the offical binary. But I hope I found the issue. Current cuda-11.1 needs >=nvidia-driver-450.80.02, but we only have 455.38-r1. Now trying to downgrade nvidia-cuda-toolkit to 11.0.

redchillipadi: I tried the same commands you pasted above, and it completes without error on my machine. The local_core directory is filled with
```
output_copy.dev.idep  output_copy.dev_avx2.o       output_copy.dev_sse4.o
output_copy.dev.o     output_copy.dev_avx512skx.o
```

I notice there is avx2 in your target, do you think this might be a contributing factor?

I saw a warning the other day that blender only officially supports nvidia-cuda-toolkit 10 and 10.1, see https://developer.blender.org/D8063. If downgrading to 10 fixes your issue we might need to patch the blender ebuild. Having said that, it seems to work with 11.1 and nvidia-drivers-455.28-r1 for me.

There is also the issue with the gpu architectures supported which changed with CUDA 11, see https://bugs.gentoo.org/744517

waebbl: > I notice there is avx2 in your target, do you think this might be a contributing factor?

Might be. I noticed this file too in the error output and was thinking whether this might be an issue.

> I saw a warning the other day that blender only officially supports nvidia-cuda-toolkit 10 and 10.1, see https://developer.blender.org/D8063. If downgrading to 10 fixes your issue we might need to patch the blender ebuild. Having said that, it seems to work with 11.1 and nvidia-drivers-455.28-r1 for me.
> 
> There is also the issue with the gpu architectures supported which changed with CUDA 11, see https://bugs.gentoo.org/744517

Yep they changed a lot, but my GPU should still be compatible. It's a compute platform 35, Kepler architecture, GK208 type, GeForce GT 730, which should still be supported according to https://en.wikipedia.org/wiki/CUDA#GPUs_supported

waebbl:
> The examples also build for me, can you please post the error you are getting?

It's identical to the error I posted above for oidn.

dev-lang/ispc/ispc-1.14.1.ebuild
redchillipadi: I have fixed the unnecessary dependency on llvm targets in both ebuilds.

I am not sure that the patch to restrict llvm find_package is necessary. I have tested compiling ispc without it on a few systems now and it never chooses LLVM 11 for me.

waebbl:
Thanks, I will test it.

BTW the problem seems to be my nvidia-uvm module which fails to load, due to unknown symbol. Need to investigate what's wrong with it.

juippis (approved):

juippis:
BTW test phase failed so expect a bug if one does not exist already.

PR #18433, opened 2020-11-27
net-misc/unison: bump to 2.51.3_p20201127

vaukai:


tillschaefer:
Why was a stable ebuild removed without stabilizing the successor first? This leads to downgrading of unison for many users to an older version (which might cause compatibility problems) 
BTW: GUI was working for me with the current stable removed.

juippis:
I'd imagine the intention was to remove the **older** stable, not the newer one. Let's revert and take it from there again. Thanks for letting us know!

tillschaefer:
THX. As long it does not contain some serious bug, I would like to urge to keep the older stable, too. It is still quite useful for compatibility reasons with other operaing systems, e.g., ubuntu 18.04.

PR #18442, opened 2020-11-28
app-admin/cdist: bump to 6.9.2

jsmolic:


app-admin/cdist/cdist-6.9.2.ebuild
```diff
@@ -3,20 +3,28 @@
...
 
 LICENSE="GPL-3+"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
 IUSE="doc"
 
+distutils_enable_tests unittest
+
+python_prepare_all() {
+	echo "VERSION='${PV}'" > "${S}"/cdist/version.py
```
SoapGentoo, line 23: `echo "VERSION='${PV}'" > cdist/version.py || die`
1. you're doing I/O with an external command, which always requires a `|| die`
2. you're already in `${S}`, hence `"${S}"/` is unnecessary

SoapGentoo (changes requested):

app-admin/cdist/cdist-6.9.2.ebuild
jsmolic: Thanks, fixed :smiley:

app-admin/cdist/cdist-6.9.2.ebuild
```diff
@@ -0,0 +1,34 @@
...
+PATCHES=( "${FILESDIR}"/${P}-tests.patch )
+
+python_prepare_all() {
+	echo "VERSION='${PV}'" > cdist/version.py || die
+
+	distutils-r1_python_prepare_all
+}
+
+python_install_all() {
+	use doc && local HTML_DOCS=( docs/src/_build/html/*.html docs/src/_build/html/man1/*.html )
```
SoapGentoo, line 32: you sure this is necessary? `distutils_enable_sphinx` injects its own install helpers

SoapGentoo (changes requested):

app-admin/cdist/cdist-6.9.2.ebuild
jsmolic: Dropped, we don't need it

app-admin/cdist/cdist-6.9.2.ebuild
```diff
@@ -0,0 +1,30 @@
...
+inherit distutils-r1
+
+DESCRIPTION="A usable configuration management system"
+HOMEPAGE="https://www.cdi.st/ https://code.ungleich.ch/ungleich-public/cdist"
+SRC_URI="https://code.ungleich.ch/ungleich-public/cdist/-/archive/${PV}/cdist-${PV}.tar.gz"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="doc"
```
jsmolic, line 18: I think this is handled with `distutils_enable_sphinx`, so I'll drop it

SoapGentoo (approved):

PR #18443, opened 2020-11-29
media-tv/mythtv: cleanup Python 2.7, drop dev-qt/qtwebkit:5

thebitpit:
Closes: https://bugs.gentoo.org/735986
Closes: https://bugs.gentoo.org/750887


thebitpit:
I removed the mythtv-30 to get rid of Python 2.7 dependency.
The dependency on QtWebKit is not needed.
Please review and merge.

media-tv/mythtv/mythtv-31.0-r4.ebuild
```diff
@@ -105,7 +105,6 @@ RDEPEND="
 	xml? ( dev-libs/libxml2:2 )
 	xmltv? (
 		dev-perl/XML-LibXML
-		dev-qt/qtwebkit:5
```
a17r, line 108: Without pushing a build system button here, doesn't that mean `qtwebkit:5` is an automagic dependency after that?

thebitpit: I learned a bit about my own 20 year old gentoo system as a result of this bug. The `USE webkit` flag was set in `make.conf` on my system. I remove it, emerge changed use, and run depclean. Several other packages were also updated as a result of the `USE webkit` flag removal. qtwebkit is no longer used by anything in my system and is unmerged. I use mythtv several hours daily and see no problems.

emerge mythtv with the `USE xmltv` flag set works fine. I emerged `media-tv/xmltv` as a test without qtwebkit on the system and saw no issues relating to webkit. I did need to `unmask dev-perl/DateTime-Format-SQLite`.

 `dev-qt/qtwebkit` is not required by mythtv any more.

`media-tv/mythplugins` is still using `dev-qt/qtwebkit`. It does not compile without it. I am working on fixing it.

mgorny: He is asking if mythtv doesn't link to qtwebkit when it is installed but the flag is off/missing.

thebitpit: Mythtv does not link to `media-tv/qtwebkit` when it is installed. The optional `media-plugins/mythplugins` does link to qtwebkit for some, not all, plugins. The `webkit` flag is still in `media-plugins/mythplugins`.

a17r: Thanks. Can you add your signoff please?

thebitpit: Earthlink was dropped by my internet provider. I had to switch to Spectrum and loose my old email addresses. I had to update a lot of accounts as a result. I changed my email to a domain I own so it should not change again. I think I have fixed up gpg and git to work with the new email in the certificate. We'll see.

PR #18444, opened 2020-11-29
app-text/mandoc: allow as default man provider

epsilon-0:
~~Still need to see how to recompress all man pages to `gz`.~~
Seems like that is not going to be possible.
There needs to be a warning stating that if using this as the man command, you should first set the PORTAGE_COMPRESS and then emerge the whole tree again.
It is either emerge the whole tree or portage should have a function which recompresses files with a different compression.

### DIFF
--- a/app-text/mandoc/mandoc-1.14.5-r1.ebuild
+++ b/app-text/mandoc/mandoc-1.14.5-r1.ebuild
@@ -1,0 +1,118 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit toolchain-funcs
+
+DESCRIPTION="Suite of tools compiling mdoc and man"
+HOMEPAGE="https://mdocml.bsd.lv/"
+SRC_URI="https://mdocml.bsd.lv/snapshots/${P}.tar.gz"
+
+LICENSE="ISC"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc64 ~x86"
+IUSE="cgi system-man"
+
+RDEPEND="sys-libs/zlib
+	system-man? ( !sys-apps/man-db )
+"
+DEPEND="${RDEPEND}
+	cgi? ( sys-libs/zlib[static-libs] )
+"
+BDEPEND="
+	cgi? ( app-text/highlight )
+"
+
+PATCHES=( "${FILESDIR}"/${PN}-1.14.5-r1-www-install.patch )
+
+pkg_pretend() {
+	if use system-man ; then
+		# only support uncompressed and gzip
+		[[ -n ${PORTAGE_COMPRESS+unset} ]] && \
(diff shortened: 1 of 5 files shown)


app-text/mandoc/mandoc-1.14.5-r1.ebuild
```diff
@@ -0,0 +1,96 @@
...
+IUSE="system-man"
+
+RDEPEND="sys-libs/zlib
+	system-man? ( !sys-apps/man-db )
+"
+DEPEND="${RDEPEND}"
+
+pkg_setup() {
+	use system-man && \
+	[[ "${PORTAGE_COMPRESS}" =~ "gzip"* ]] || \
```
Arfrever, line 24: `${PORTAGE_COMPRESS}` is a command used for compression, not list of used compression formats or something, so no need for that regular expression. And `p*` would match 0 or more of `p`.

I suggest simply:
`[[ ${PORTAGE_COMPRESS} != gzip ]] && \`

epsilon-0: done!

app-text/mandoc/mandoc-1.14.5-r1.ebuild
```diff
@@ -0,0 +1,100 @@
...
+	system-man? ( !sys-apps/man-db )
+"
+DEPEND="${RDEPEND}"
+
+pkg_setup() {
+	if use system-man ; then
+		# allow uncompressed and gzip
+		[[ ! -z "${PORTAGE_COMPRESS+unset}" ]] && \
+		( [[ "${PORTAGE_COMPRESS}" == "gzip" ]] || [[ "${PORTAGE_COMPRESS}" == "" ]] ) || \
+		die "only PORTAGE_COMPRESS=gzip or '' is allowed for system wide support"
```
Arfrever, line 27: What about alternative package managers which do not use `PORTAGE_COMPRESS` variable?
Maybe just print a message without calling `die`?

zmedico: Not sure that die is appropriate in any case, maybe an error or ewarn is enough. If it's going to die then you could at least add a way out with `[[ -n ${I_KNOW_WHAT_I_AM_DOING} ]]`.

The check should go in pkg_pretend. Instead of a subshell you can do this:
```
[[ "${PORTAGE_COMPRESS}" == "gzip" || "${PORTAGE_COMPRESS}" == "" ]]
```

epsilon-0: good point!
done.

floppym:
The mandoc ebuild goes out of its way to rename the binaries to avoid installing standard commands like "man" and "apropos". I don't think it makes sense to add it to virtual/man if it doesn't provide the standard commands.

floppym:
Oh, I see you made some changes to the mandoc ebuild as well. Nevermind.

PR #18445, opened 2020-11-29
dev-vcs/git-cola: verbump to 3.8

bearsh:
replaces #18329

@stkw0 are OK with this?


stkw0:
> replaces #18329
> 
> @stkw0 are OK with this?

Yeah, awesome.

Please add "GIT_CONFIG_NOSYSTEM=true" to both ebuilds to avoid system configuration to conflict with tests.
And it should also close this bug: https://bugs.gentoo.org/739026. It's related with the QtNetwork dependency.

juippis (approved):

PR #18448, opened 2020-11-29
*/*: Port to EAPI 7

jsmolic:


net-mail/dbmail/dbmail-3.2.3-r3.ebuild
```diff
@@ -0,0 +1,132 @@
...
+		--enable-systemd \
+		--sysconfdir=/etc/dbmail \
+		--disable-static \
+		$(use_with sieve) \
+		$(use_with ldap auth-ldap)
+}
+
+src_install() {
+	default
+	emake DESTDIR="${D}" SYSTEMD_UNIT_DIR="$(systemd_get_systemunitdir)" install
```
SoapGentoo, line 96: doesn't this double install (because you're already calling `emake install` in `default`)?

net-mail/dbmail/dbmail-3.2.3-r3.ebuild
```diff
@@ -0,0 +1,132 @@
...
+	newinitd "${FILESDIR}/dbmail-pop3d.initd" dbmail-pop3d
+	newinitd "${FILESDIR}/dbmail-timsieved.initd" dbmail-timsieved
+
+	dobin contrib/mailbox2dbmail/mailbox2dbmail
+	doman contrib/mailbox2dbmail/mailbox2dbmail.1
+
+	# ldap schema
+	if use ldap; then
+		insinto /etc/openldap/schema
+		doins "${S}/dbmail.schema"
```
SoapGentoo, line 118: `doins dbmail.schema`

SoapGentoo (changes requested):

net-mail/dbmail/dbmail-3.2.3-r3.ebuild
jsmolic: Oh yes, I should remove `default` call then

net-mail/dbmail/dbmail-3.2.3-r3.ebuild
jsmolic: Fixed

net-mail/dbmail/dbmail-3.2.3-r3.ebuild
jsmolic: Done

PR #18452, opened 2020-11-30
sys-cluster/impi: standalone Intel MPI

epsilon-0:
it is available with an ISSL license
which we can use to separate it out from
intel parallel studio

---

Ugliest crap ever...
Please do tonnes of runtime checking before merging.


scardracs:
Ping: still interested on it?

PR #18453, opened 2020-11-30
app-shells/bash: update shipped bashrc to add support for TERM=tmux

redneb:
Closes: https://bugs.gentoo.org/757711
Signed-off-by: Marios Titas <redneb@gmx.com>


chutz:
This probably merits a revbump, while we are here we should probably add `alacritty`, which has also been gaining a lot of popularity lately.

PR #18454, opened 2020-11-30
sci-libs/libxc: version bump 4.3.4

TheChymera:
Partly credited to Reinis Danne:
https://github.com/gentoo/sci/pull/903/files


sci-libs/libxc/libxc-4.3.4.ebuild
```diff
@@ -0,0 +1,52 @@
...
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	econf --enable-shared \
+		$(use_enable fortran) \
+		$(use_enable static-libs static)
```
SoapGentoo, line 29: disable static libs

SoapGentoo (changes requested):

sci-libs/libxc/libxc-4.3.4.ebuild
TheChymera: It seems to work afaict. Why disable it?

PR #18458, opened 2020-11-30
dev-python/pipenv: bump version to 2020.11.15

oz123:
Signed-off-by: Oz Tiram <oz.tiram@gmail.com>


juippis (commented):
Your commits seem to be mixed a bit and you're removing some older version in your bump.

Is the `<pytest-5` requirement still true? Can't it be patched?

oz123:
@juippis I should split that commit. You are right.

As for pytest ...  this is taken[ from here](https://github.com/pypa/pipenv/blob/48e1f18930aec29d3d8dc1e346a7726f5d324907/setup.py#L45).
```
    "tests": ["pytest<5.0", "pytest-timeout", "pytest-xdist<2.0", "flaky", "mock"],
```
However, I can check if this also works with pytest 5 or later..

juippis (approved):

juippis:
Please make/find a comment upstream about it, it could just be they forgot to update README or something.

oz123:
Already done here: https://github.com/pypa/pipenv/issues/4559

The development of pipenv was taken over by a new team pretty abruptly. I think there is still a lot of old cruft hiding there.
Is the update OK besides this? Can the PR be merged as is?

juippis:
Yep, sorry for the delay, I usually attempt to push as many finished PRs as possible at once.

PR #18464, opened 2020-11-30
dev-lisp/c2ffi: Bump version to c2ffi-11.0.0.20201113

hjudt:
Update from c2ffi-10 (deprecated upstream) to c2ffi-11.


dev-lisp/c2ffi/c2ffi-11.0.0.20201113.ebuild
```diff
@@ -16,8 +16,8 @@ SLOT="0"
 KEYWORDS="~amd64"
 IUSE=""
 
-DEPEND="sys-devel/clang:10=
-	sys-devel/llvm:10=
+DEPEND="sys-devel/clang:11=
+	sys-devel/llvm:11=
```
juippis, line 20: Have we talked about this before, but I can have llvm-10 clang-10 and -11 installed, but how does this ebuild force -11 to be used? You probably need some checks from *llvm.eclass*.

juippis: Or does upstream handle this with their build system properly?

hjudt: AFAICT, the build system will look for llvm-11, not llvm-10. So if llvm-11 is not installed, it will not build.

juippis (approved):

PR #18465, opened 2020-11-30
dev-python/reedsolomon: fix ebuild, add necessary dependency

madmartin:
Unfortunately the buildtime dependency dev-python/cython is missing.
This ebuild fixes this.

Closes: https://bugs.gentoo.org/757486

### DIFF
--- a/dev-python/reedsolomon/reedsolomon-1.5.4.ebuild
+++ b/dev-python/reedsolomon/reedsolomon-1.5.4.ebuild
@@ -16,6 +16,10 @@ KEYWORDS="~amd64 ~x86"
 IUSE="test"
 RESTRICT="!test? ( test )"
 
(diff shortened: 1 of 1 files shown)


juippis (commented):
It's a build-time fix so no revbump needed. Just edit the 1.5.4 ebuild adding required BDEPEND.

madmartin:
> It's a build-time fix so no revbump needed. Just edit the 1.5.4 ebuild adding required BDEPEND.

@juippis  Yes, you're right, I changed the PR. My first thinking was that cython is also needed at runtime, then a -r1 would have been correct I think.

juippis (approved):
That's correct, RDEPEND changes require a revbump most of the time. Adding a python compatibility being a special case.

PR #18473, opened 2020-12-01
sci-libs/lapack: lots of changes to BLAS/LAPACK switch

epsilon-0:
move all BLAS/LAPACK providers to do chainloading of libraries.
introduces two new eclasses for helping deal with the chainload linking and installation + registration.
removes the eselect-ldso USE flag and makes it redundant.


epsilon-0:
dude wth...

PR #18478, opened 2020-12-02
Net wireless/wavemon

alarig:


net-wireless/wavemon/wavemon-0.9.1.ebuild
```diff
@@ -0,0 +1,54 @@
...
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~hppa ~ppc ~sparc ~x86"
+
+IUSE="caps"
+RDEPEND="dev-libs/libnl:3
+	sys-libs/ncurses:0=
+	caps? ( sys-libs/libcap )"
+DEPEND="${RDEPEND}
+	virtual/pkgconfig"
```
DarthGandalf, line 21: this should be in BDEPEND instead

https://devmanual.gentoo.org/general-concepts/dependencies/index.html

net-wireless/wavemon/Manifest
```diff
@@ -1,2 +1,4 @@
...
 DIST wavemon-0.8.0.tar.gz 196353 BLAKE2B 90d358d325b62ba9379ea525c4399e38c38043e67460b20188c4d81261d290e8e3f154a33240882f8b2611e4db1862d7bd33c87e168569be9d751d431602da81 SHA512 2df48eb300e9b30e0cb7a0ab41ba5e10b2a6e8170dbd11a4c978462c8cd6dc1171f79ceadfdb7344c495420a4e87b7dd723efea4f1af59a6d0312d93049ee798
+DIST wavemon-0.9.1.tar.gz 209240 BLAKE2B d7c423cd23efb0ff998461b1e29fd2310c597d7a5b67e967206fcaaa2ea574682646d6fa9dfca89153ecb7e45de886cc07fcffad3d7b65fb52d63894d1f73b0a SHA512 c05e9ea488d3fb9bfdb15419aadc54bed08d338dba7c118d2a2ff6423ffe31c16ff8e9e05fc2e39fffb069f28fa7daef30f079ca63186064c0ade4f9fe98864c
+DIST wavemon-0.9.2.tar.gz 202575 BLAKE2B 0a1afe1081bbbeb2ba0fec4741df7da846b65bbf182e886b2185c5e406258a1ecf912509b1a8f4862f37ba92832240dc6db4305a8d5cfb6f4876f3e93b04ccc7 SHA512 f1ee2e4602e93f0639c10067d23df021101bfc0496e53b221cd1553ebc90876cdd963cd64d5ef503da6d4dc4fbcf640e78873df1bdf64272e70dfa3c53ec1706
```
DarthGandalf, line 4: If you're adding 0.9.2 already, why do you need 0.9.1 too?

alarig: It’s because the bug https://bugs.gentoo.org/649632 is about 0.9.1

DarthGandalf: I see. The bump for 0.9.2 would satisfy such bug as well. 0.9.1 was just the current latest version at the time. It was even requesting 0.8.2 before.

So, no need.

alarig: I’m deleting 0.9.1 then

net-wireless/wavemon/wavemon-0.9.1.ebuild
alarig: I moved it to `BDEPEND` on the 0.9.2 ebuild (as I removed 0.9.1 as stated in the other thread)

juippis (approved):

PR #18482, opened 2020-12-03
dev-python/pybtex: Bump to 0.23.0

WPettersson:
Includes change of testing dependency from nose to pytest

Package-Manager: Portage-3.0.11, Repoman-3.0.1


WPettersson:
Oops, this QA issue was caused by me fat-fingering and removing python3_6 from python targets, somehow. Pushing new commit without that now.

juippis (commented):
Please split adding and removing ebuilds into their separate own commits. So this should be 2 commits if you want to remove 0.22.2 version.

juippis (approved):

PR #18490, opened 2020-12-03
net-libs/webkit-gtk add minibrowser use flag

nkichukov:
fixes: https://bugs.gentoo.org/758299


leio:
Thanks, I've added it to USE=examples instead together with some important fixes in a revbump

PR #18496, opened 2020-12-04
app-editors/vim: do not create symlink vimdiff for minimal

dongsupark:
A symlink `vimdiff` should not be created, if the USE flag `minimal` is enabled.
Otherwise running `vimdiff` results in failure like that:

```
$ vimdiff aaa bbb
This Vim was not compiled with the diff feature.
```


dongsupark:
Added GCO sign-off.

monsieurp (approved):
Thanks for the fix. Merging underway.

PR #18503, opened 2020-12-04
games-emulation/higan: Version bump to 116_pre

vilhelmgray:
Closes: https://bugs.gentoo.org/709674


vilhelmgray:
This ebuild isn't yet ready; when I use `emake` the GNU make invocation has trouble finding some directories. Oddly, when I manually execute `make` in the source directory for higan, it finds directories just fine and completes the compilation without trouble.

Here's an example `make` line that runs correctly when I manually execute, yet seems to fail when invoked via `emerge`:
```
make -j5 platform=linux compiler=x86_64-pc-linux-gnu-g++ hiro=gtk3 cores="cv fc gb gba md ms msx n64 ngp pce ps1 sfc sg ws" profile=accuracy build=performance local=false -C higan-ui
```

Any idea what I have incorrect in my ebuild to cause this problem?

chewi:
My first thought was a race condition but I see you've called `make -j5` there. Are you sure you didn't just get lucky that time? Try emerging with `MAKEOPTS="-j1"`.

vilhelmgray:
I'm still trying to resolve the ebuild issue mentioned above. Unrelated, I wonder if we should define the emulation cores similar to how `app-emulation/qemu` handles it: by using [QEMU_SOFTMMU_TARGETS](https://github.com/gentoo/gentoo/blob/master/profiles/desc/qemu_softmmu_targets.desc) rather than regular USE flags. Would it make sense to define something similar for `games-emulation/higan`, or leave the cores as individual USE flags?

vilhelmgray:
> My first thought was a race condition but I see you've called `make -j5` there. Are you sure you didn't just get lucky that time? Try emerging with `MAKEOPTS="-j1"`.

I figured out what's happening: **emake** is removing the quotes for the ` cores="cv fc gb gba md ms msx n64 ngp pce ps1 sfc sg ws"` option. So essentially the line that it finally executes is this:
```
make -j5 platform=linux compiler=x86_64-pc-linux-gnu-g++ hiro=gtk3 cores=cv fc gb gba md ms msx n64 ngp pce ps1 sfc sg ws profile=accuracy build=performance local=false -C higan-ui
```

Without the quotation, **make** interprets `fc` as a target rather than a directory path, and thus I see the error issued.

Do you know a way to keep **emake** from stripping the quotes from `core="..."`?

vilhelmgray:
Investigating a bit further, I tried executing with some debug info:
```
MAKEOPTS="-p" emerge higan
```

In the printout of variables I found the following:
```
# command line
cores = "cv
```

So it looks like it's actually delimiting based on whitespace for some reason and ignoring the quotations. Does this have something to do with how **emake** constructs the `cmd` list? https://github.com/gentoo/portage/blob/ff270ae58eca457a1e2bbf9507d8438123082ca0/bin/ebuild-helpers/emake#L14-L16

vilhelmgray:
I was able to resolve the issue. Seems it was having trouble combining the bash arrays together as a string, so I was able to avoid the problem by creating an intermediate bash array called `uiopts`. This turned out to make the ebuild more concise too, so that's a plus in the end. :+1: 

I forced-pushed the fixed ebuild, so feel free to test it and review.

vilhelmgray:
I decided to have the `byuu` user interface be the default USE flag as this is the configuration recommended by upstream for casual users; the `higan-ui` user interface and `icarus` library manager are apparently intended more so for advanced users.

games-emulation/higan/higan-116_pre.ebuild
```diff
@@ -0,0 +1,164 @@
...
+HOMEPAGE="https://github.com/higan-emu/higan"
+SRC_URI="https://github.com/higan-emu/${PN}/archive/nightly.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="
+	alsa ao +byuu +core-cv +core-fc +core-gb +core-gba +core-md +core-ms
+	+core-msx +core-n64 +core-ngp +core-pce +core-ps1 +core-sfc +core-sg
+	+core-ws +gtk higan-ui icarus openal +opengl oss +pulseaudio qt5 +sdl udev
```
a17r, line 18: Please reconsider adding all those `core-*` flags.
- They do not pull in any additional dependency
- What is the net benefit (in size) of disabling individual cores? If it is not substantial by modern standards, why bother?
- Is there any practical gain (besides install size) from being able to disable one or several cores?

games-emulation/higan/higan-116_pre.ebuild
```diff
@@ -0,0 +1,164 @@
...
+	xv"
+REQUIRED_USE="
+	^^ ( gtk qt5 )
```
a17r, line 21: This will introduce a default conflict for every single `desktop` profile user. You don't want that.

Is there a preferred gui according to upstream?

a17r (changes requested):

games-emulation/higan/higan-116_pre.ebuild
vilhelmgray: I suppose most users of higan are using it specifically because it supports multiple cores. So perhaps there's not much benefit afterall in providing these USE flags because I imagine most users would just leave them all enabled anyway. I think it's best to just remove the USE flags then and just build all cores unconditionally.

games-emulation/higan/higan-116_pre.ebuild
vilhelmgray: The user interface for higan is planned to be re-implemented some time in the future: https://github.com/higan-emu/higan/issues/28

Upstream believes the GTK2 backend is the most solid one right now for Linux systems, but it's best to not use that one because GTK2 support will definitely go away soon. So I think GTK3 should be the one we use here instead because it will be the most similar to the GTK2 backend used by higan in our past ebuild versions. I'll remove the qt5 USE flag then and build the GTK3 backend unconditionally.

a17r: That doesn't mean you have to remove the Qt5 option altogether. You can drop the `qt5` flag and build it if the user has unset `gtk` instead.

vilhelmgray: So we would have something like the following instead?
```
!gtk? ( qt5? (
		dev-qt/qtcore:5
		dev-qt/qtgui:5
		dev-qt/qtwidgets:5
	)
)
```

a17r: yup.

vilhelmgray: All right, that makes sense. I've gone ahead and removed the qt5 use and left the gtk use flag with a metadata entry to describe the new behavior.

games-emulation/higan/higan-116_pre.ebuild
```diff
@@ -0,0 +1,138 @@
...
+inherit desktop toolchain-funcs xdg
+
+DESCRIPTION="A multi-system emulator focused on accuracy, preservation, and configurability"
+HOMEPAGE="https://github.com/higan-emu/higan"
+SRC_URI="https://github.com/higan-emu/${PN}/archive/nightly.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="
```
a17r, line 15: We do not usually make a newline here.

games-emulation/higan/higan-116_pre.ebuild
```diff
@@ -0,0 +1,138 @@
...
+	alsa ao +byuu +gtk higan-ui icarus openal +opengl oss +pulseaudio +sdl udev
+	xv"
+REQUIRED_USE="|| ( byuu higan-ui icarus )"
```
a17r, line 18: If it is correct that `alsa`, `ao`, `openal`, `opengl`, `pulseaudio`, `sdl`, `udev` and `xv` flags are only effective with either `byuu` or `higan-ui` then you'll need to add a bunch of additional constraints not to make them ambiguous. Unless it is a mistake -  at least previous ebuilds did not make that distinction.

games-emulation/higan/higan-116_pre.ebuild
```diff
@@ -0,0 +1,138 @@
...
+		$(usex sdl input.sdl "")
+		$(usex udev input.udev "")
+		$(usex xv video.xvideo "")
+	)
+
+	local coreopts=(
+			cores="cv fc gb gba md ms msx n64 ngp pce ps1 sfc sg ws"
+			ruby="${drivers[@]}"
+			build=performance
+			local=false
```
a17r, line 84: too much indendation

a17r (changes requested):
I'm sure it has come up before, but a "version bump" commit should never include the removal of an old version. That shall be in a separate commit.

games-emulation/higan/higan-116_pre.ebuild
vilhelmgray: For some historical context: previous versions of higan did not include `byuu` which is a new alternative user interface for casual users that don't need the advanced controls of the full `higan-ui` user interface; `icarus` is [a separate tool bundled with higan that allows you to bulk-import ROM files into higan’s game library](https://higan.readthedocs.io/en/latest/interface/icarus/).

Because `icarus` is just a game library management tool, the driver USE flags (`alsa`, `ao`, `openal`, etc..) don't apply to it. I believe `icarus` is intended to be run in conjunction with the `higan-ui` user interface, but I created a USE flag option for it because the tool could in theory be run independently since it's just used to construct and maintain the layout of your games directory on your filesystem.

If you think it'll be simpler for users, I can remove the `icarus` flag and build it only when `higan-ui` is selected -- this would match how upstream bundles `icarus` with `higan-ui` while `byuu` is alone.

tobiasjakobi:
Any news on this one? A version bump would be very much appreciated, in particular given how 106 is from 2017.

Also thanks to @vilhelmgray for working on this! :)

vilhelmgray:
Upstream has restructured the higan build to remove `byuu` as a separate UI and have it all under a unified `higan-ui` interface in order to reduce confusion to end-users. This ebuild will need to be updated again to account for this minor change. I'll try to update this ebuild some time this month when I find the time.

a17r:
> Also thanks to @vilhelmgray for working on this! :)

++

`byuu` removal will automatically get rid of the last thing that needed to be fixed in this PR.

games-emulation/higan/higan-116_pre.ebuild
a17r: Given that `icarus` flag also does not drive any additional dependencies, it makes sense to build it unconditionally and (together with `byuu`) drop the REQUIRED_USE complexity.

vilhelmgray:
> > Also thanks to @vilhelmgray for working on this! :)
> 
> ++
> 
> `byuu` removal will automatically get rid of the last thing that needed to be fixed in this PR.

You're right, and it looks like this was the commit that removed `byuu` along with some useful history and explanation: https://github.com/higan-emu/higan/commit/a8d19935fac5a9ae2a0a2b00284af937da269d59

This should actually be a simple update so I'll push the change in a few minutes.

vilhelmgray:
After removing the `profile=accuracy` line, I'm encountering the bash array combination issue again: https://github.com/gentoo/gentoo/pull/18503#issuecomment-739424474

Using an intermediate bash array does not seem to avoid this issue, so I need some help figuring out how to overcome this.

games-emulation/higan/higan-116_pre.ebuild
```diff
@@ -0,0 +1,100 @@
...
+		$(usex oss audio.oss "")
+		$(usex pulseaudio "audio.pulseaudio audio.pulseaudiosimple" "")
+		$(usex sdl input.sdl "")
+		$(usex udev input.udev "")
+		$(usex xv video.xvideo "")
+	)
+
+	local coreopts=(
+		cores="cv fc gb gba md ms msx ngp pce sfc sg ws"
+		ruby="${drivers[@]}"
```
a17r, line 66: fwics you want this to be a string list, so `${drivers[*]}` should work instead.

games-emulation/higan/higan-116_pre.ebuild
```diff
@@ -0,0 +1,100 @@
...
+	# Make higan
+	emake "${makeopts[@]}" "${coreopts[@]}" -C higan-ui
+
+	# Make icarus
+	emake "${makeopts[@]}" -C icarus
+}
+
+src_install() {
+	# Install higan
+	dobin higan-ui/out/higan
```
a17r, line 80:
```
>>> Install games-emulation/higan-116_pre into /var/tmp/portage/games-emulation/higan-116_pre/image
!!! dobin: higan-ui/out/higan does not exist
```

Looks like it should be `dobin higan-ui/out/higan-ui` instead.

a17r (changes requested):

games-emulation/higan/higan-116_pre.ebuild
vilhelmgray: That worked. Thanks! :+1:

games-emulation/higan/higan-116_pre.ebuild
vilhelmgray: No, this was renamed in https://github.com/higan-emu/higan/commit/a8d19935fac5a9ae2a0a2b00284af937da269d59

Is it trying to build an old version of the tarball?

games-emulation/higan/higan-116_pre.ebuild
```diff
@@ -0,0 +1,100 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit desktop toolchain-funcs xdg
+
+DESCRIPTION="A multi-system emulator focused on accuracy, preservation, and configurability"
+HOMEPAGE="https://github.com/higan-emu/higan"
+SRC_URI="https://github.com/higan-emu/${PN}/archive/nightly.tar.gz -> ${P}.tar.gz"
```
a17r, line 10: Right. This doesn't work. They don't have a recent versioned tarball available?'

a17r (changes requested):

games-emulation/higan/higan-116_pre.ebuild
a17r: If you want to use what they call `higan nightly 2021-03-23`, since they do not make it available as a reliable tarball, you'll have to make a snapshot pinned to the specific commit 44cbb88d49d5a77f19e353fad4bc7fbab6909ca9 and the ebuild will then be versioned `116_pre20210323`.

vilhelmgray: Ah, I see what's going on: this nightly version isn't static but is periodically updated. I talked with the team last December and they hadn't chosen a date at the time for the 116 proper release; if I recall correctly, the intention was to release at some time this year.

Pinning this ebuild to a snapshop might be the best path forward for now. Does Github provide a way to get the tarball of a specific commit rather than a tag?

vilhelmgray: Looks like it's as simple as replacing `nightly` with the commit ID, so I'll do that force-push the update.

games-emulation/higan/higan-116_pre20210404.ebuild
```diff
@@ -0,0 +1,100 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit desktop toolchain-funcs xdg
+
+DESCRIPTION="A multi-system emulator focused on accuracy, preservation, and configurability"
+HOMEPAGE="https://github.com/higan-emu/higan"
+SRC_URI="https://github.com/higan-emu/${PN}/archive/44cbb88d49d5a77f19e353fad4bc7fbab6909ca9.tar.gz -> ${P}.tar.gz"
```
a17r, line 10: Please make that use `${COMMIT}` and set `COMMIT=44cbb88d49d5a77f19e353fad4bc7fbab6909ca9` before inherit.
You can use `COMMIT` for setting `S` as well then, and ideally move that line below `SRC_URI` since ultimately here S is linked to SRC_URI.

vilhelmgray: Do you mean move `S` up closer to `SRC_URI` so that `S` is right below it?

tobiasjakobi:
@vilhelmgray If I could make a suggestion. Can we make the two drivers video.xshm and input.xlib dependant on the X useflag?

games-emulation/higan/metadata.xml
```diff
@@ -7,6 +7,10 @@
 	</maintainer>
 	<use>
 		<flag name="icarus">Build icarus library manager</flag>
-		<flag name="udev">Enable udev based input</flag>
```
tobiasjakobi, line 10: Is it intentional to remove this? I'm asking since this still holds true.

vilhelmgray:
> @vilhelmgray If I could make a suggestion. Can we make the two drivers video.xshm and input.xlib dependant on the X useflag?

X11 libraries are unconditionally liked by the Makefile so I don't think there would be much benefit to adding the `X` use flag here because we'll be pulling in and linking to the dependencies anyway.

games-emulation/higan/metadata.xml
vilhelmgray: This local description is essentially the same as the global description for the udev USE flag, so I'm removing it as redundant.

games-emulation/higan/higan-116_pre20210404.ebuild
a17r: yep

games-emulation/higan/metadata.xml
tobiasjakobi: Ah, OK, that makes sense!

a17r:
Please update commit message :)

tobiasjakobi:
> > @vilhelmgray If I could make a suggestion. Can we make the two drivers video.xshm and input.xlib dependant on the X useflag?
> 
> X11 libraries are unconditionally linked by the Makefile so I don't think there would be much benefit to adding the `X` use flag here because we'll be pulling in and linking to the dependencies anyway.

Right, so something that would need fixing upstream first. Was just a thought of a Wayland user :)

vilhelmgray:
> > > @vilhelmgray If I could make a suggestion. Can we make the two drivers video.xshm and input.xlib dependant on the X useflag?
> > 
> > 
> > X11 libraries are unconditionally linked by the Makefile so I don't think there would be much benefit to adding the `X` use flag here because we'll be pulling in and linking to the dependencies anyway.
> 
> Right, so something that would need fixing upstream first. Was just a thought of a Wayland user :)

That's a fair concern and we should expect that Wayland would be gaining more users over the coming years. Perhaps you should open a Github issue for upstream about adding Wayland support. Once upstream implements it we can then integrate it here in the ebuild so users have a choice between X11 and Wayland when building.

a17r:
Btw, couldn't help but notice in old ebuild:
```
        #elog "optional dependencies:"
        #elog "  dev-games/higan-ananke (extra rom load options)"
        #elog "  games-util/higan-purify (Rom purifier)"
```

Is any of that relevant?

a17r (approved):

vilhelmgray:
> ananke

I believe these have been superseded by icarus: https://github.com/higan-emu/higan/commit/b113ecb5a3b69b9a2152c993f40dc2360493a48b

PR #18510, opened 2020-12-05
net-im/discord-bin: version bump to 0.0.13

weedoe:
Closes: https://bugs.gentoo.org/758479
Signed-off-by: William Paterson weedoe@gmail.com


weedoe:
Made the dumbest mistake of not adding the bug link to the commit messages. Force pushed amendments to the messages and everything should be okay now!

00-matt:
Thanks for trying to update it :)

One commit should be to bump the version (so it adds a new ebuild and the hash to the manifest), and you can drop the old broken version in a commit afterwards: https://wiki.gentoo.org/wiki/Gentoo_git_workflow#Atomicity

Also your commit messages should follow [the format](https://wiki.gentoo.org/wiki/Gentoo_git_workflow#Commit_message_format), e.g. `net-im/discord-bin: Bump to 0.0.13' and `net-im/discord-bin: Drop old`.

weedoe:
> Thanks for trying to update it :)
> 
> One commit should be to bump the version (so it adds a new ebuild and the hash to the manifest), and you can drop the old broken version in a commit afterwards: https://wiki.gentoo.org/wiki/Gentoo_git_workflow#Atomicity
> 
> Also your commit messages should follow [the format](https://wiki.gentoo.org/wiki/Gentoo_git_workflow#Commit_message_format), e.g. `net-im/discord-bin: Bump to 0.0.13` and `net-im/discord-bin: Drop old`.

Thanks, I will do this for future contributions. Does this need to be changed now or will it be acceptable as it is?

coalwater:
Hi @weedoe,
I'm also new to this and I only found this because I wanted to update my discord

By reading the comment from @00-matt and checking the history of the file here's what I believe you need to do according to what I understood

Here's the history of the manifest file
https://github.com/gentoo/gentoo/commits/e38c331bed70419879e60cd9ba096a4a243db66a/net-im/discord-bin/Manifest

Your first commit should add to the manifest but not remove the old hash also copy the old ebuild but not rename it, this is to have the two versions available at the same time

Reference: https://github.com/gentoo/gentoo/commit/0156a2ed617b036847a2995dc3f7658fa787c114#diff-7424a3c669de7bcf774541e9f32cafca9dd19bc04a5ec8124bbb5fe8eaf95506

After that a clean up PR would come which would remove the old hash and the old ebuild.

steils:
Closed in https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=300bf4d5a04c02a932de111ba013ac439972f8ce

PR #18516, opened 2020-12-05
media-libs/opensubdiv: Allow compilation against CUDA 11

redchillipadi:
This ebuild contains two changes which support compilation against
nvidia-cuda-toolkit-11

Compared to nvidia-cuda-toolkit-10 which was gcc < 9 only, version 11
can be compiled with later versions of gcc.

Currently the supported versions are
<=nvidia-cuda-toolkit-9.0  ( <=sys-devel/gcc-5.4 )
nvidia-cuda-toolkit-9.1 ( <=sys-devel/gcc-6.4 )
nvidia-cuda-toolkit-9.2 to 10.1 ( <=sys-devel/gcc-7.3 )
nvidia-cuda-toolkit 10.1 to 10.2 ( <=sys-devel/gcc-8.4 )
nvidia-cuda-toolkit-11.0 ( <=sys-devel/gcc-9.3 )
nvidia-cuda-toolkit-11.1 ( <=sys-devel/gcc-10.4 )

Putting this testing in every ebuild that uses nvcc at build time is
fragile and will be easily broken by updates to nvidia-cuda-toolkit.

This version removes the gcc version checks in pkg_pretend, as
the user has already been warned about selecting appropriate versions
of gcc in a message that appears when installing nvidia-cuda-toolkit.

The second change is that nvidia-cuda-toolkit 11 removed support for
cards supporting compute_30. A patch is included that builds opensubdiv
against compute_35 when nvidia-cuda-toolkit 11 is used.

This is only a workaround as a proper solution would be a USE_EXPAND
variable to allow the user to select their exact compute version
supported, and compile optimised cuda binary files for their specific
card.

In particular this version chooses the lowest version of compute to
support the largest range of graphics cards, but this prevents using
features provided by later versions.

Closes: https://bugs.gentoo.org/751382
Bug: https://bugs.gentoo.org/744517


redchillipadi:
I tried rebasing this as was suggested to fix the too many broken packages warning. but now there are 524 extra commits that I don't need. My two commits for opensubdiv work fine here. Sometimes I think it is easier just to start again.

khumarahn:
I don't think you did the rebasing right... your pull request should only have a few commits, not 526.

juippis:
Yes please do start over. I guess you merged into your branch, not rebased?

(Obviously you probably can still do a hard reset or rebase and drop those 526 commits, then wget your work from github and re-commit)

redchillipadi:
I will close and recreate this PR

PR #18519, opened 2020-12-05
dev-libs/roct-thunk-interface: Version bump to 3.10.0

justxi:


justxi:
Please keep the old ebuilds if possible until I have a backup of the old ones.

PR #18522, opened 2020-12-06
games-roguelike/tomenet: version bump, 4.7.3

DarthGandalf:


games-roguelike/tomenet/tomenet-4.7.3.ebuild
```diff
@@ -35,42 +36,44 @@ PATCHES=(
 
 src_prepare() {
 	default
-	use server || use dedicated || { rm -r ../lib/{config,data,save} || die ;}
+	use server || { rm -r ../lib/{config,data,save} || die ;}
```
SoapGentoo, line 39: this is a terrible construct, turn it into a normal `if ! use server; then ... fi` construct

games-roguelike/tomenet/tomenet-4.7.3.ebuild
```diff
@@ -80,20 +83,10 @@ src_install() {
 	doins ../.tomenetrc
 }
 
-pkg_preinst() {
-	gnome2_icon_savelist
-}
-
 pkg_postinst() {
```
SoapGentoo, line 86: `xdg_postinst` won't get called here

games-roguelike/tomenet/files/tomenet-4.7.3-makefile.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/makefile b/makefile
+index ee56a33..7f3c81f 100644
+--- a/makefile
++++ b/makefile
```
SoapGentoo, line 4: can you just check whether the makefile also respects `CPPFLAGS`? This is a user flag that should be respected

SoapGentoo (changes requested):

games-roguelike/tomenet/files/tomenet-4.7.3-makefile.patch
DarthGandalf: Done

SoapGentoo (approved):

PR #18526, opened 2020-12-06
app-editors/texworks: bump PYTHON_COMPAT, migrate to lua-single.eclass

telans:
Closes: https://bugs.gentoo.org/758662
Closes: https://bugs.gentoo.org/752534


telans:
It looks like this has already been done now

PR #18528, opened 2020-12-06
dev-libs/jansson: bump revision to r1

Zer0-One:
This revision removes the dependency on sphinx < 3.0 .

Closes: https://bugs.gentoo.org/731668
Signed-off-by: David Zero <zero-one@zer0-one.net


juippis (approved):
This is good, but you could try to use *docs.eclass* to control the doc-building. I'd be interested to see how it works with *multilib-minimal.eclass*.

@AndrewAmmerlaan anything worth warning you see here?

AndrewAmmerlaan:
> This is good, but you could try to use _docs.eclass_ to control the doc-building. I'd be interested to see how it works with _multilib-minimal.eclass_.
> 
> @AndrewAmmerlaan anything worth warning you see here?

```
PYTHON_COMPAT=( python3_{7,8,9} )
DOCS_BUILDER="sphinx"
DOCS_DIR="doc"

inherit python-any-r1 docs multilib-minimal

....

multilib_src_compile() {
	default
	docs_compile
}

....
````
Should do the trick (haven't tested though). `doc` can be removed from IUSE and `sphinx` can be removed from BDEPEND, this is added by the eclass automatically.

juippis:
Is inheriting *python-any-r1* necessary for `docs_compile`, as that's a bit counter-intuitive? I can kind of see the connection with sphinx and it, but can it be dealth within *docs.eclass*?

AndrewAmmerlaan:
> Is inheriting _python-any-r1_ necessary for `docs_compile`, as that's a bit counter-intuitive? I can kind of see the connection with sphinx and it, but can it be dealth within _docs.eclass_?

Inheriting either python-any-r1, python-r1 or distutils-r1 is necessary for any python based documentation builder (sphinx and mkdocs), because we need `PYTHON_COMPAT`, `PYTHON_USEDEP` and `python_gen_any_dep` to get the dependencies right (it is not technically required for the `docs_compile` function). Previously the eclass would inherit python-any-r1 if none of the necessary eclasses had been inherited, but this led to confusion when using the eclass as python-any-r1 and python-r1 can not be inherited simultaneously and in that case the order eclasses are inherited in becomes important. People got confused that `inherit docs distutilts-r1` would lead to the error that python-any-r1 and python-r1 could not be inherited simultaneously, while `inherit distutils-r1 docs` would work just fine. Therefore the eclass now dies with a helpful error message if none of the required eclases are inherited, instead of assuming which eclass should be inherited. The disadvantage is that you now have to manually inherit a python-* eclass for sphinx and mkdocs, but hopefully this behaviour is less confusing.

PR #18529, opened 2020-12-06
media-libs/libquvi-scripts: migrate to lua-single.eclass

telans:
I think I'm doing this right

Closes: https://bugs.gentoo.org/752960


ConiKost:
This has been already migrated.

PR #16366, opened 2020-06-21
Dev python/mkdocs: add py3_8, version bump

AndrewAmmerlaan:
Version bumps and py_3_8 update for mkdocs and deps as requested in bug 728832

PR #18530, opened 2020-12-06
dev-libs/efl: use docs.eclass for building docs

AndrewAmmerlaan:
As requested by @juippis , an example on how to use
the new docs.eclass with doxygen

See-Also: https://github.com/gentoo/gentoo/pull/16366

### DIFF
--- a/dev-libs/efl/efl-1.25.1-r10.ebuild
+++ b/dev-libs/efl/efl-1.25.1-r10.ebuild
@@ -5,7 +5,10 @@ EAPI=7
 
 LUA_COMPAT=( lua5-{1..3} luajit )
 
-inherit lua-single meson xdg-utils
+DOCS_BUILDER="doxygen"
+DOCS_DIR="${S}/doc"
+
+inherit docs lua-single meson xdg-utils
 
 DESCRIPTION="Enlightenment Foundation Libraries all-in-one package"
 HOMEPAGE="https://www.enlightenment.org"
@@ -14,7 +17,7 @@ SRC_URI="https://download.enlightenment.org/rel/libs/${PN}/${P}.tar.xz"
 LICENSE="BSD-2 GPL-2 LGPL-2.1 ZLIB"
 SLOT="0"
 KEYWORDS="~amd64 ~arm ~arm64 ~hppa ~ppc ~ppc64 ~x86"
-IUSE="+X avif bmp connman cpu_flags_arm_neon dds debug doc drm +eet efl-one elogind examples fbcon
+IUSE="+X avif bmp connman cpu_flags_arm_neon dds debug drm +eet efl-one elogind examples fbcon
 	+fontconfig fribidi gif gles2-only gnutls glib +gstreamer harfbuzz hyphen ibus ico libressl
 	jpeg2k json nls mono opengl +pdf physics pmaps postscript psd pulseaudio raw scim
 	sdl +sound +ssl +svg +system-lz4 systemd tga tgv tiff tslib unwind v4l vnc wayland webp xcf
@@ -125,7 +128,6 @@ RDEPEND="${LUA_DEPS}
 	zeroconf? ( net-dns/avahi )"
 DEPEND="${RDEPEND}"
 BDEPEND="virtual/pkgconfig
-	doc? ( app-doc/doxygen )
 	nls? ( sys-devel/gettext )"
 
 pkg_setup() {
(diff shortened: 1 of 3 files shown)


juippis (commented):
I'm getting,
```
 * ERROR: dev-libs/efl-1.25.1-r10::testworld failed (compile phase):
 *   doxygen_compile: doc/Doxyfile not found, DOCS_DIR=doc or DOCS_CONFIG_NAME=Doxyfile wrong
 * 
 * Call stack:
 *     ebuild.sh, line  125:  Called src_compile
 *   environment, line 2908:  Called docs_compile
 *   environment, line 1329:  Called doxygen_compile
 *   environment, line 1339:  Called die
 *       [[ -f ${doxyfile} ]] || die "${FUNCNAME}: ${doxyfile} not found, DOCS_DIR=${DOCS_DIR} or DOCS_CONFIG_NAME=${DOCS_CONFIG_NAME} wrong";
```
with this applied, while the *::gentoo* one installs fine.
```
 >:  equery f efl | grep "/html" | wc -l
654
```

I kind of have a vague memory of `doxygen` failing as-is, but `doxygen .` working for some reason. It's probably related?

juippis:
Other feedback: Using the eclass looks really simple. Kind of surprised to see nothing is needed in `src_install`, but I do get and see the logic with: `HTML_DOCS+=( "${DOCS_OUTDIR}/." )` in `docs_compile`.

AndrewAmmerlaan:
> I'm getting,
> 
> ```
>  * ERROR: dev-libs/efl-1.25.1-r10::testworld failed (compile phase):
>  *   doxygen_compile: doc/Doxyfile not found, DOCS_DIR=doc or DOCS_CONFIG_NAME=Doxyfile wrong
>  * 
>  * Call stack:
>  *     ebuild.sh, line  125:  Called src_compile
>  *   environment, line 2908:  Called docs_compile
>  *   environment, line 1329:  Called doxygen_compile
>  *   environment, line 1339:  Called die
>  * The specific snippet of code:
>  *       [[ -f ${doxyfile} ]] || die "${FUNCNAME}: ${doxyfile} not found, DOCS_DIR=${DOCS_DIR} or DOCS_CONFIG_NAME=${DOCS_CONFIG_NAME} wrong";
> ```
> 
> with this applied, while the _::gentoo_ one installs fine.
> 
> ```
>  >:  equery f efl | grep "/html" | wc -l
> 654
> ```
> 
> I kind of have a vague memory of `doxygen` failing as-is, but `doxygen .` working for some reason. It's probably related?

Oopsie, my bad, I was editing the -r10 version and emerging the -r2 version, I'll look into it

AndrewAmmerlaan:
> I kind of have a vague memory of `doxygen` failing as-is, but `doxygen .` working for some reason. It's probably related?

@juippis So turns out there is a meson.build file in the doc directory that is supposed to create Doxyfile from Doxyfile.in. I could not get that to work so I just do the necessary edits in src_prepare, and now docs compilation works just fine.

So turns out the ebuild was not actually compiling the docs correctly before, because by calling `doxygen .` the config file is bypassed and the generated docs are not complete. It should be complete now, in any case the generation of the docs takes significantly longer now indicating that it is now doing things that it was not doing before.

Ideally we would get the automatic generation of the Doxyfile from the Doxyfile.in working, but I lack the knowledge and experience with meson.build files to get it working.

dev-libs/efl/efl-1.25.1-r10.ebuild
```diff
@@ -5,7 +5,10 @@ EAPI=7
 
 LUA_COMPAT=( lua5-{1..3} luajit )
 
-inherit lua-single meson xdg-utils
+DOCS_BUILDER="doxygen"
+DOCS_DIR="${S}/doc"
```
juippis, line 9: Does the `${S}` need to get specified now with doxygen? Or is this specific to efl?

juippis (approved):
Heh, my only experience with doxygen and it was wrong. No wonder I had a weird feeling about it. Thanks for looking in! Left a question about `DOCS_DIR` regarding future reviews.

Also you're welcome to port any maintainer-needed / proxy-maint packages to use this eclass if you want, and they'll get reviewed. Can't promise that for packages with other maintainers.

dev-libs/efl/efl-1.25.1-r10.ebuild
AndrewAmmerlaan: For sphinx you can get away with using a relative path. For doxygen and mkdocs DOCS_OUTDIR has to be a absolute path because those docbuilders have to be run in the directory where the config file is, which means that if DOCS_DIR is not equal to S (the default) the docbuilder and `einstalldocs` are not run from the same directory which will cause `einstalldocs` to fail because it can't find the directory specified in `HTML_DOCS`. Because DOCS_OUTDIR is set relative to DOCS_DIR this means that DOCS_DIR has to be an absolute path.

This morning I thought of a solution that will allow the eclass to be used with relative paths in DOCS_DIR (changing the default DOCS_OUTDIR from DOCS_DIR/_build/html to S/_build/html). That way DOCS_OUTDIR is an absolute path by default, and it is not longer strictly necessary for DOCS_DIR to be an absolute path as well.

PR #18536, opened 2020-12-06
app-emulation/libvirt-6.8.0: Properly handle libpcap if it's explicitly disabled

zippy2:
This is a direct backport of an upstream commit of:

  c8be9ccc36a32cc756d05b2c0838c43a57be15f7

It was merged in 6.9.0 release, so 6.8.0 is the only release
which is still buggy.

Resolves: https://bugs.gentoo.org/754888


zippy2:
CC @trofi

trofi:
Pulled as is. Thank you!

PR #18537, opened 2020-12-06
dev-python/libvirt-python-9999: Update PYTHON_COMPAT

zippy2:
All other dev-python/libvirt-python ebuilds support through
python3.7 to python3.9 except for the live ebuild. This doesn't
make much sense - the live ebuild does support those python
versions too.


zippy2:
CC @trofi

trofi:
Pulled as is. Thank you!

PR #18538, opened 2020-12-06
net-analyzer/zabbix: introduce subslots

Whissi:
Needed for proper targeting in GLSAs.


Whissi:
Status broken is false positive by CI which is flagging usage of user eclass. But we need this because we are only rev bumping to promote subslot.

fordfrog:
@Whissi i took your idea and did the requested changes with respect to the upstream releases: https://www.zabbix.com/download_sources
i also reused your commit comment :-)

PR #18542, opened 2020-12-06
net-libs/mbedtls: introduce slots

Whissi:
While we still don't support installing multiple net-libs/mbedtls
GLSAs.


blueness:
wait i have to back out of this

blueness:
sorry I accidentally pushed this and reverted immediately.  i'm okay with introducing slots, but you need an upgrade path.

blueness:
maybe we can introduce them upon bumping to new versions.

Arfrever:
Why introducing more fake slots, which collide with each other?

winterheart:
@thesamesam can this PR be closed as #39536 superseded it?

thesamesam:
Yes.

PR #18543, opened 2020-12-06
app-emulation/virtualbox: introduce subslots

Whissi:
While we still don't support installing multiple app-emulation/virtualbox
GLSAs.


Whissi:
@Polynomial-C please review

PR #18545, opened 2020-12-06
net-vpn/i2p: bump to 0.9.48

tharvik:
* bump to new upstream version
* also reduce the number of unpacked deps, some where unused, so made the new version fails startup
* remove the user directory migration, it was added in 0.9.42 and it's been a few version since.
* drop old version


net-vpn/i2p/i2p-0.9.48.ebuild
```diff
@@ -0,0 +1,159 @@
...
+
+src_test() {
+	# store built version of jars, overwritten by testing
+	mv "${S}/pkg-temp/lib/"{i2p,router}.jar "${T}" ||
+		die "unable to save jars before tests"
+
+	# generate test classpath
+	local cp
+	cp="$(java-pkg_getjars --build-only junit-4,hamcrest-core-1.3,hamcrest-library-1.3)"
+	EANT_TEST_EXTRA_ARGS="-Djavac.classpath=${cp}" java-pkg-2_src_test
```
juippis: Could you rename this variable as it immediately made me think of the Unix `cp` command which in return churned my brain to a know for a moment.

juippis (approved):

net-vpn/i2p/i2p-0.9.48.ebuild
tharvik: Will do in next bump then.

PR #18546, opened 2020-12-07
dev-lang/go: introduce slots

Whissi:
While we still don't support installing multiple dev-lang/go
GLSAs.

### DIFF
--- a/dev-lang/go/go-1.14.12-r1.ebuild
+++ b/dev-lang/go/go-1.14.12-r1.ebuild
@@ -21,7 +21,7 @@ case ${PV}  in
 	case ${PV} in
 	*_beta*|*_rc*) ;;
 	*)
-		KEYWORDS="-* ~amd64 ~arm ~arm64 ~ppc64 ~s390 ~x86 ~amd64-linux ~x86-linux ~x64-macos ~x64-solaris"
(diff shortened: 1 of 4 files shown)


Whissi:
Note: I ignored live ebuild, you probably have to find another solution for it.

Whissi:
@williamh please review

thesamesam (approved):

dev-lang/go/go-1.14.12-r1.ebuild
```diff
@@ -30,12 +30,13 @@ DESCRIPTION="A concurrent garbage collected and typesafe programming language"
...
 LICENSE="BSD"
-SLOT="0/${PV}"
+SLOT="$(ver_cut 1-2)/${PV}"
 
 BDEPEND="|| (
 		dev-lang/go
 		dev-lang/go-bootstrap )"
-RDEPEND="!<dev-go/go-tools-0_pre20150902"
+RDEPEND="!<dev-go/go-tools-0_pre20150902
+	!dev-lang/go:1.15"
```
floppym, line 39: You would also need to block slot 0 to force portage to unmerge currently installed versions.

floppym (changes requested):
Overall, this is an abuse of slots to compensate for limitations in the GLSA tooling.

I suggest enhancing the version matching syntax used by the GLSA tools to allow for version ranges with lower and upper bounds that will be applied together.

thesamesam:
If we're going to do this, we should really just slot Go properly (which is not a bad idea).

PR #18548, opened 2020-12-07
media-gfx/librecad: bump to 2.2.0_rc1

Moult:


media-gfx/librecad/files/librecad-2.2.0_rc1-qt-5.15.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/librecad/src/lib/engine/lc_splinepoints.cpp b/librecad/src/lib/engine/lc_splinepoints.cpp
```
DarthGandalf, line 1: 1. These files are already in files/, no need to copy them if they don't need changes for the new version. Simply replace `${P}` in patches with the correct name.

2. Please don't remove the header with description, otherwise after a while it will be impossible to say what is the purpose of this patch.

media-gfx/librecad/librecad-2.2.0_rc1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+
+inherit desktop qmake-utils
+
+DESCRIPTION="Generic 2D CAD program"
+HOMEPAGE="https://www.librecad.org/"
+SRC_URI="https://github.com/LibreCAD/LibreCAD/archive/${PV//_/-}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="amd64 ~ppc64 x86 ~amd64-linux ~x86-linux"
```
DarthGandalf, line 14: New version shouldn't go straight to stable, mark it as ~

media-gfx/librecad/librecad-2.2.0_rc1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+src_configure() {
+	eqmake5 -r
+}
+
+src_install() {
+	dobin unix/librecad
+	use tools && dobin unix/ttf2lff
+	insinto /usr/share/${PN}
+	doins -r unix/resources/*
+	use doc && docinto html && dodoc -r librecad/support/doc/*
```
DarthGandalf, line 51: Is the USE flag necessary here? I think this falls into the "When not to use USE flags?" section of https://devmanual.gentoo.org/general-concepts/use-flags/index.html

media-gfx/librecad/files/librecad-2.2.0_rc1-qt-5.15.patch
Moult: The old patch doesn't apply because the line numbers have changed, I think. That is why I created a new patch file, even though the contents are the same. Is there a better way?

I will add the header with description.

DarthGandalf: If just the line numbers changed, the patch would still apply. But if context changed, like the includes nearby... I see how it wouldn't apply anymore.

Moult: Ah, OK then the context must've changed. Because originally I did try to reuse the original patch, and only when it failed to apply I created a new one.

media-gfx/librecad/librecad-2.2.0_rc1.ebuild
Moult: Fixed

media-gfx/librecad/librecad-2.2.0_rc1.ebuild
Moult: Fixed

media-gfx/librecad/files/librecad-2.2.0_rc1-qt-5.15.patch
Moult: I think you are right for one of the patches, will check and fix.

Moult: You are absolutely right! I've removed one patch. The other remains relevant, though.

Moult:
Anything else I can do to help this process?

media-gfx/librecad/librecad-2.2.0_rc1.ebuild
```diff
@@ -0,0 +1,56 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit desktop qmake-utils
+
+DESCRIPTION="Generic 2D CAD program"
+HOMEPAGE="https://www.librecad.org/"
```
juippis, line 9: Please add the Github page here as 2ndary homepage.

media-gfx/librecad/librecad-2.2.0_rc1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+src_install() {
+	dobin unix/librecad
+	dobin unix/ttf2lff
+	insinto /usr/share/${PN}
+	doins -r unix/resources/*
+	docinto html && dodoc -r librecad/support/doc/*
+	insinto /usr/share/appdata
+	doins unix/appdata/librecad.appdata.xml
+	doicon librecad/res/main/${PN}.png
+	make_desktop_entry ${PN} LibreCAD ${PN} Graphics
```
juippis, line 55:
```suggestion
    dobin unix/librecad
    dobin unix/ttf2lff

    insinto /usr/share/${PN}
    doins -r unix/resources/*

    docinto html
    dodoc -r librecad/support/doc/*

    insinto /usr/share/appdata
    doins unix/appdata/librecad.appdata.xml

    doicon librecad/res/main/${PN}.png
    make_desktop_entry ${PN} LibreCAD ${PN} Graphics
```

juippis (commented):
```
 * Messages for package media-gfx/librecad-2.2.0_rc1:

 * 
 * This package seems to install metainfo files into the following
 * location(s):
 * 
 *   /usr/share/appdata
 * 
 * This location is deprecated, it should not be used anymore by new software.
 * Appdata/Metainfo files should be installed into /usr/share/metainfo directory.
 * For more details, please see the freedesktop Upstream Metadata guidelines at
 * https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html
 * 
 * Gentoo Tracker bug: https://bugs.gentoo.org/709450
 * 
```
This can be solved by a patch, or simply using `mv`, `mkdir`, `rmdir` in `src_install`. Rest looks good here.

media-gfx/librecad/librecad-2.2.0_rc1.ebuild
juippis: Note use tabs, not spaces there

media-gfx/librecad/librecad-2.2.0_rc1.ebuild
juippis: Still not added.

media-gfx/librecad/librecad-2.2.0_rc1.ebuild
```diff
@@ -0,0 +1,61 @@
...
+    doins -r unix/resources/*
+
+    docinto html
+    dodoc -r librecad/support/doc/*
+
+    insinto /usr/share/appdata
+    doins unix/appdata/librecad.appdata.xml
+
+    doicon librecad/res/main/${PN}.png
+    make_desktop_entry ${PN} LibreCAD ${PN} Graphics
```
juippis, line 60: My mistake here, but I did ask to use tabs instead of spaces. Well that's an easy fix for me too.

juippis (commented):
And this needs to be fixed, 
```
 * Messages for package media-gfx/librecad-2.2.0_rc1:

 * 
 * This package seems to install metainfo files into the following
 * location(s):
 * 
 *   /usr/share/appdata
 * 
 * This location is deprecated, it should not be used anymore by new software.
 * Appdata/Metainfo files should be installed into /usr/share/metainfo directory.
 * For more details, please see the freedesktop Upstream Metadata guidelines at
 * https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html
 * 
 * Gentoo Tracker bug: https://bugs.gentoo.org/709450
 * 
```

Looking at
https://github.com/LibreCAD/LibreCAD/blob/master/scripts/postprocess-unix.sh
it's a very simple patch or a single sed you need to get it fixed.

arkamar:
Obsoleted by 2.2.0 which is in the tree since commit 12765a744522.

PR #18550, opened 2020-12-07
net-firewall/xtables-addons: Version updated to 3.13

jpds:
Closes: https://bugs.gentoo.org/755926


blueness:
pushed.  thanks for keeping your eye on this!

PR #18558, opened 2020-12-07
Cascadia code

janhenke:


telans:
You should use `Closes: https://bugs.gentoo.org/758950` instead of `Bug:`

PR #18559, opened 2020-12-08
app-i18n/librime: Add plugins for lua support and bump to 1.6.1

goduck777:
Signed-off-by: Chang Liu goduck777@gmail.com


Arfrever:
This approach with bundling of other packages (in this case plugins) in `app-i18n/librime` is not sustainable.
It does not allow to independently update/install plugins.

I am working on proper solution:
1. RIME plugins will be in separate packages (`app-i18n/librime-charcode`, `app-i18n/librime-lua`, `app-i18n/librime-octagram` etc.), not bundled inside `librime.so`.
They will be installed in `/usr/$(get_libdir)/rime-plugins` directory (e.g. `/usr/lib64/rime-plugins`) as separate `.so` files (e.g. `/usr/lib64/rime-plugins/librime-lua.so`).
2. `librime.so` from `app-i18n/librime` will dynamically load all installed plugins using `dlopen()`.

Do you need any other plugins beside those 3 aforementioned?

goduck777:
@Arfrever Glad to know you are working on it.

Actually I only need librime-lua. I just followed https://github.com/archlinux/svntogit-community/blob/packages/librime/trunk/PKGBUILD to put all the plugins in that file.

I think the official suggested way to install plugin is to do make merged-plugins for librime, which means that you need to reinstall librime when installing plugins. The cmake files in librime-lua does not seems to be an independent one.

Anyway, if you can find another way working, it would be preferred.

Arfrever:
It seems to be ready. (I will test it more...)
`app-i18n/librime-charcode` and `app-i18n/librime-octagram` will be added later.
I will send patch for librime to upstream.

You can test patch for `gentoo.git` repository now: [librime.patch.txt](https://github.com/gentoo/gentoo/files/5683370/librime.patch.txt)

`app-i18n/librime-lua` depends on currently masked Lua, so you need to add:
* To `/etc/portage/package.unmask`:
```
app-eselect/eselect-lua
dev-lang/lua
```
If you have other packages depending on Lua, then you should unmask currently masked versions of them.
See this list: https://gitweb.gentoo.org/repo/gentoo.git/tree/profiles/package.mask?id=6b525b817ffb07fb04c85fa898916b9a0623d5df#n475
* To `/etc/portage/package.use` (example):
```
app-i18n/librime-lua LUA_SINGLE_TARGET: lua5-4
```
(Alternatively `LUA_SINGLE_TARGET="lua5-4"` can be set in `/etc/portage/make.conf`.)

goduck777:
@Arfrever It works. This is a lot of work. Thank you.

Do you want to add it to Gentoo repo or you want to wait for the upstream?

goduck777:
@Arfrever A minor suggestion: Add lua as a use flag to install librime-lua.

Arfrever:
I attached patch for upstream in `https://github.com/rime/librime/issues/431`.
I will wait a while for potential comments from upstream, but not more than several days.

I will add these packages in Gentoo.

About USE flag, maybe I will add `app-i18n/rime-meta` package with several USE flags.

goduck777:
@Arfrever This is a good idea. AFAIK, most users of fcitx has moved to fcitx5, which is available in gentoo-zh overlay. I hope you can help packing fcitx5 to the official repo. You can check the existing ebuild in gentoo-zh as a starting point.

I will submit a bug to gentoo bugzilla to track this.

PR #18560, opened 2020-12-08
dev-util/cmake: Backport FPHSA fix

redchillipadi:
Changes to Find Package Handle Standard Args in 3.19.0 result in
an error when compiling openvdb-7.1.0.

This is a known problem in cmake and the fix has been merged into
3.19.2 upstream.

See https://gitlab.kitware.com/cmake/cmake/-/issues/21505

To produce the bug:
emerge dev-util/cmake-3.19.0 or 3.19.1 (the bug not exist in <= 3.18)
emerge media-gfx/openvdb-7.1.0-r1

Compilation fails with
find_package_check_version(): Cannot be used outside a 'Find Module'

To fix the bug:
Apply the patch provided in this bug fix to cmake, then repeat the
above steps and compilation of openvdb succeeds.

The patch ensures that the CMAKE_FPHSA_PACKAGE_NAME package is made
available outside find_package.

Thanks to Dennis Schridde for finding the upstream fix, produced by
Marc Chevrier.

Closes: https://bugs.gentoo.org/755743

### DIFF
--- a/dev-util/cmake/cmake-3.19.0.ebuild
+++ b/dev-util/cmake/cmake-3.19.0.ebuild
@@ -70,6 +70,7 @@ PATCHES=(
 	"${FILESDIR}"/${PN}-3.18.0-filter_distcc_warning.patch # bug 691544
 
 	# upstream fixes (can usually be removed with a version bump)
+	"${FILESDIR}"/${P}-use-FPHSA-outside-find_package.patch # 21505
 )
 
 cmake_src_bootstrap() {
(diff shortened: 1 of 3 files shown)


a17r (changes requested):
Thanks for your contribution.
- Please use git format-style for the upstream patch to include full header (should be possible to get through their git web interface as well).
- This requires a revision bump for `3.19.1` so it reaches all ~arch users, otherwise it remains broken for those who have it already installed.
- Forget about fixing `3.19.0`, we will just drop that version.

PR #18563, opened 2020-12-08
www-apps/gitea: security bump

nemunaire:
Hi!

Please find a simple version bump for gitea.


www-apps/gitea/gitea-1.13.0.ebuild
```diff
@@ -33,14 +33,11 @@ RDEPEND="${COMMON_DEPEND}
...
+	custom/conf/app.example.ini CONTRIBUTING.md README.md
 )
 FILECAPS=(
 	cap_net_bind_service+ep usr/bin/gitea
 )
-PATCHES=(
-	"${FILESDIR}/1.12-fix-vendoring.patch"
-)
 
 RESTRICT="test"
```
thesamesam, line 42: I'd still really love even a small number of tests to be running in future if possible

PR #18564, opened 2020-12-08
gui-wm/hikari: fix suid flag usage

epsilon-0:
was changed from WITH_SUID to WITHOUT_SUID

Closes: https://bugs.gentoo.org/758968


epsilon-0:
no merge yet...
Seems like its a bit weird.

juippis:
In the bug it was reported this needs a fix in `src_install` too, you could try that?

epsilon-0:
I got thrown off by the fact that `hikari-unlocker` is always set to suid, only `hikari` suid is controllable. Upstream has said, that is how it should behave, I didn't ask too many questions.

PR #18565, opened 2020-12-08
dev-util/lldb-9.0.1-r2 compile with newer PYTHON_TARGETS

mercuriete:
Closes: https://bugs.gentoo.org/758404
Signed-off-by: Abel Garcia Dorta <mercuriete@yahoo.es>


telans:
`~dev-python/lit-9.0.1` does not support PYTHON_COMPAT >python3_7

Perhaps you should make a PR for that package too

mercuriete:
@telans 
I did a PR but I dont know if is correct:
https://github.com/gentoo/gentoo/pull/18645

mercuriete:
this was fixed by
https://bugs.gentoo.org/735756

closing

PR #18568, opened 2020-12-08
net-misc/networkmanager: When use=json, add cflag -fcommon

candrews:
Closes: https://bugs.gentoo.org/738614
Signed-off-by: Craig Andrews <candrews@gentoo.org>

### DIFF
--- a/net-misc/networkmanager/networkmanager-1.26.6.ebuild
+++ b/net-misc/networkmanager/networkmanager-1.26.6.ebuild
@@ -7,7 +7,7 @@ GNOME2_LA_PUNT="yes"
 VALA_USE_DEPEND="vapigen"
 PYTHON_COMPAT=( python3_{6,7,8} )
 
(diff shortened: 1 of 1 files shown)


winterheart:
This is not right solution for GCC 10 porting.

candrews:
> This is not right solution for GCC 10 porting.

Agreed - but it is a workaround.

Perhaps instead of closing the issue, I update the PR to mention the issue, and we just keep this as a workaround until the real fix is in place?

BTW, it looks like this issue has been fixed in 1.27.1: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/329#note_723086 Could we bump to that?

winterheart:
Well, package needs version bump to 1.28.0 anyway. Commit 341111ff looks as real solution for -fno-common, but can't be used as individual patch due huge amount of changes.

candrews:
Done in https://github.com/gentoo/gentoo/commit/ee7c0b87152636485019566cf0ffabffb12a9da3

PR #18574, opened 2020-12-09
Clean up some GNOME packages after the eclass updates

dm0-:
@mattst88 Let me know which of these should get a new revision.  They all seem to produce the same binpkgs.

### DIFF
--- a/dev-libs/glib/glib-2.62.6.ebuild
+++ b/dev-libs/glib/glib-2.62.6.ebuild
@@ -250,14 +250,7 @@ pkg_postinst() {
 		gnome2_giomodule_cache_update \
 			|| die "Update GIO modules cache failed (for ${ABI})"
 	}
-	if ! tc-is-cross-compiler ; then
-		multilib_foreach_abi multilib_pkg_postinst
-	else
-		ewarn "Updating of GIO modules cache skipped due to cross-compilation."
-		ewarn "You might want to run gio-querymodules manually on the target for"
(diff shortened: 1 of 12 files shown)


leio:
I don't think librsvg-2.40 should get an EAPI bump, as it is security vulnerable and just promotes cross-compiling people to expose themselves to security issues more.

dm0-:
@leio But librsvg-2.40 is the only version available for three CPUs I'm targeting: armv5, i686 with no SSE, and sh4.  Rust versions of librsvg are unkeyworded and/or masked in the arch profiles where there is no `dev-lang/rust-bin`.  Anyone with such CPUs is prevented from using the updated versions.

mattst88:
> I don't think librsvg-2.40 should get an EAPI bump, as it is security vulnerable and just promotes cross-compiling people to expose themselves to security issues more.

I don't think this hurts anything. If its so vulnerable, we should just remove it from the tree, right? Otherwise, allowing easier cross-compilation to architectures without rust support seems fine...

leio:
> @leio But librsvg-2.40 is the only version available for three CPUs I'm targeting: armv5, i686 with no SSE, and sh4. Rust versions of librsvg are unkeyworded and/or masked in the arch profiles where there is no `dev-lang/rust-bin`. Anyone with such CPUs is prevented from using the updated versions.

I'm sold with these arguments; unsure about revbump or not.

rust without SSE is possible to my knowledge, but not in Gentoo without lots of hoops.

PR #18576, opened 2020-12-09
dev-vcs/git-cola: add py3.8, fix qa warnings

bearsh:
Bug: https://bugs.gentoo.org/750068
Bug: https://bugs.gentoo.org/713930
Signed-off-by: Martin Gysel <me@bearsh.org>


bearsh:
I don't know about the stabilization process in detail but may this revision go straight to stable?

juippis (commented):
If the only change you do in here is adding python-3.8 compatibility, you can do it without revbumping. It's counted as a "USE flag change", 

bearsh:
@juippis thanks for your feedback, I've updated the pull request. the introduced patch is only relevant for the newly added python version...

juippis (approved):

PR #18577, opened 2020-12-09
media-gfx/img2pdf: version bump to 0.4.0

fedeliallalinea:
- removed ~x86 keyword because no pikepdf ebuild available

Closes: https://bugs.gentoo.org/759184


media-gfx/img2pdf/files/img2pdf-0.4.0_disable-error-tests.patch
```diff
@@ -0,0 +1,33 @@
...
+ 
+ 
+-@pytest.mark.skipif(
+-    sys.platform in ["darwin", "win32"],
+-    reason="test utilities not available on Windows and MacOS",
+-)
+-def test_png_icc(tmp_path_factory, png_icc_img, png_icc_pdf):
+-    tmpdir = tmp_path_factory.mktemp("png_icc")
+-    compare_ghostscript(tmpdir, png_icc_img, png_icc_pdf, icc=True)
++#@pytest.mark.skipif(
```
thesamesam, line 24: There's no need to comment these out, just remove them instead. The information is still retained in the patch as - lines.

media-gfx/img2pdf/img2pdf-0.4.0.ebuild
```diff
@@ -0,0 +1,38 @@
...
+LICENSE="LGPL-3+"
+SLOT="0"
+KEYWORDS="~amd64"
+
+BDEPEND="test? (
+	dev-python/numpy[${PYTHON_USEDEP}]
+	dev-python/pdfrw[${PYTHON_USEDEP}]
+	dev-python/pytest[${PYTHON_USEDEP}]
+	dev-python/scipy[${PYTHON_USEDEP}]
+	<media-gfx/imagemagick-7.0.0[jpeg,jpeg2k,png,q8,q32,tiff]
```
thesamesam, line 24: This is really unfortunate. Avoidable at all?

media-gfx/img2pdf/img2pdf-0.4.0.ebuild
```diff
@@ -0,0 +1,38 @@
...
+)"
+RDEPEND="dev-python/pikepdf[${PYTHON_USEDEP}]
+dev-python/pillow[${PYTHON_USEDEP}]"
```
thesamesam, line 27: Indent

media-gfx/img2pdf/files/img2pdf-0.4.0_disable-error-tests.patch
```diff
@@ -0,0 +1,33 @@
+diff -Naru a/src/img2pdf_test.py b/src/img2pdf_test.py
```
thesamesam, line 1: Source for the patch / some bug references?

thesamesam (commented):
Left a few comments. Please remember to file a bug for rekeywording img2pdf on ~x86 when this is merged.

media-gfx/img2pdf/files/img2pdf-0.4.0_disable-error-tests.patch
fedeliallalinea: In the ebuild I added reference
https://gitlab.mister-muffin.de/josch/img2pdf/issues/85

media-gfx/img2pdf/img2pdf-0.4.0.ebuild
fedeliallalinea: I tried to use imagemagick-7 but first problem is that formatDescription change, for example, from "Portable Network Graphics" to "PNG" so it would be necessary to patch.
Even doing this other tests fail.

fedeliallalinea:
> Left a few comments. Please remember to file a bug for rekeywording img2pdf on ~x86 when this is merged.

I should re-add ~x86 keyword?

thesamesam:
No, leave the keyword out, but when we merge this, you need to file a bug asking for it to be restored.

media-gfx/img2pdf/files/img2pdf-0.4.0_endianness.patch
```diff
@@ -0,0 +1,245 @@
...
++    assert identify[0]["image"].get("endianness") in ["Undefined", "LSB",], str(
+         identify
+     )  # FIXME: should be LSB
+     assert identify[0]["image"].get("depth") == 1, str(identify)
+@@ -3334,7 +3334,7 @@ def tiff_ccitt_msb_m2l_white_img(tmp_pat
+     }, str(identify)
+     assert identify[0]["image"].get("colorspace") == "Gray", str(identify)
+     assert identify[0]["image"].get("type") == "Bilevel", str(identify)
+-    assert identify[0]["image"].get("endianess") in [
++    assert identify[0]["image"].get("endianness") in [
```
juippis, line 206: Is this `endianess` just a typo across this file? I'd say `sed` works better for this than a patch file.

media-gfx/img2pdf/files/img2pdf-0.4.0_disable-error-tests.patch
juippis: We like to keep the headers **in** the patch file itself, because other distros/users may just look at the patch and not ebuild. It should tell everything relevant.

juippis (commented):
Please clarify the situation with that `endianess` patch, rest LGTM.

fedeliallalinea:
I think this version bump can be closed because tests depend on dev-python/pdfrw that has been masked, or a solution is disable all tests?

juippis:
Yeah, free to restrict for now. Add a comment above `RESTRICT` stating one dep is missing, then in future please try to work up whether tests can somehow be ran without the missing dep. Ie, disable part of them.

media-gfx/img2pdf/img2pdf-0.4.0.ebuild
```diff
@@ -0,0 +1,40 @@
...
+#BDEPEND="test? (
+#	dev-python/numpy[${PYTHON_USEDEP}]
+#	dev-python/pdfrw[${PYTHON_USEDEP}]
+#	dev-python/pikepdf[${PYTHON_USEDEP}]
+#	dev-python/pytest[${PYTHON_USEDEP}]
+#	dev-python/scipy[${PYTHON_USEDEP}]
+#	<media-gfx/imagemagick-7.0.0[jpeg,jpeg2k,png,q8,q32,tiff]
+#)"
+RDEPEND="dev-python/pillow[${PYTHON_USEDEP}]"
+
```
juippis, line 35:
```
INSTALL_REQUIRES = (     "Pillow",     "pikepdf", ) 
```
Would this also need to depend on *dev-python/pikepdf* outside tests?

juippis, line 35:
```
>: img2pdf-gui 
  File "/usr/lib/python-exec/python3.8/img2pdf-gui", line 33, in <module>
    sys.exit(load_entry_point('img2pdf==0.4.0', 'gui_scripts', 'img2pdf-gui')())
  File "/usr/lib/python3.8/site-packages/img2pdf.py", line 2371, in gui
    import tkinter
ModuleNotFoundError: No module named 'tkinter'
```
Is this a problem? Do we have tkinter in Gentoo? Not sure should you try to package it or remove the *-gui* executable :\

ionenwks: > Do we have tkinter in Gentoo?

Haven't tested but USE=tk on python should provide it
https://wiki.gentoo.org/wiki/Tkinter

juippis: Yep, that did it :)

I guess you could include a `gui` USE flag and I guess this should work then:
`gui? ( $(python_gen_impl_dep tk) )`
(But if it is too hard just ignore for now)

fedeliallalinea: Thanks @ionenwks and @juippis.
Sorry for mess and the little care I took in testing for this package, I am not the proxy maintainer and I bumped only for add it as ocrmypdf dependency in my overlay.

media-gfx/img2pdf/img2pdf-0.4.0.ebuild
```diff
@@ -0,0 +1,46 @@
...
+RDEPEND="dev-python/pikepdf[${PYTHON_USEDEP}]
+	dev-python/pillow[${PYTHON_USEDEP}]
+	gui? ( $(python_gen_impl_dep tk) )"
+
+distutils_enable_tests pytest
+
+src_prepare() {
+	default
+	use !gui && sed -i '/gui_scripts/d' setup.py
+}
```
juippis, line 42:
```suggestion
    distutils-r1_python_prepare_all                                                                 
                                                                                                    
    if ! use gui; then                                                                              
        sed -i '/gui_scripts/d' setup.py || die                                                     
    fi
```

juippis (approved):

media-gfx/img2pdf/img2pdf-0.4.0.ebuild
juippis: Eeh, it's fine. I hope you understand we rather get it as close to perfect in here than deal with bug reports later on.

I'm fine with dropping `~x86`, I seriously have doubts how popular that arch is anymore.

PR #18583, opened 2020-12-09
dev-python/numpy: remove linking to flame library

epsilon-0:


mgorny:
@SoapGentoo, could you confirm?

PR #18584, opened 2020-12-09
dev-vcs/tortoisehg: 5.6.1 version bump

band-a-prend:
Add python3_9 compat. as `dev-python/iniparse`, `dev-python/qscintilla-python` now support it


dev-vcs/tortoisehg/tortoisehg-5.6.1.ebuild
```diff
@@ -0,0 +1,62 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+PYTHON_COMPAT=( python3_{6,7,8,9} )
```
telans, line 5:
```suggestion
PYTHON_COMPAT=( python3_{7..9} )
```
dev-python/iniparse doesn't support PYTHON_COMPAT python3_6

band-a-prend: Thank you for advise. I thinked I need mask python3_6 as iniparse 0.4-r2 still has it. But as I know python3_6 is planning to be dropped soon, isn't it? Then it's really better to remove it here.

telans: > I thinked I need mask python3_6 as iniparse 0.4-r2 still has it

There's no need to mask anything aside from limiting PYTHON_COMPAT as I believe portage will only attempt to install the latest version which fits within the constraints. If this package used python-single-r1 either version of `iniparse` would be okay (?) but because all python targets could be enabled, the version of iniparse selected needs to support every PYTHON_COMPAT here (in this case `dev-python/iniparse-0.5` is the only one that can). `dev-python/iniparse-0.42-r2` is esentially masked should `dev-vcs/tortoisehg` be installed. I believe I have that all correct

band-a-prend: Done

juippis (commented):
```
 * python3_9: running distutils-r1_run_phase distutils-r1_python_compile
python3.9 setup.py build -j 16
  File "/var/tmp/portage/portage/dev-vcs/tortoisehg-5.6.1/work/tortoisehg-5.6.1/setup.py", line 30, in <module>
    from i18n.msgfmt import Msgfmt
  File "/var/tmp/portage/portage/dev-vcs/tortoisehg-5.6.1/work/tortoisehg-5.6.1/i18n/msgfmt.py", line 39, in <module>
    from mercurial import (
ModuleNotFoundError: No module named 'mercurial'
 * ERROR: dev-vcs/tortoisehg-5.6.1::testworld failed (compile phase):
 *   (no error message)
 * 
```
so it looks like you want to add `dev-vcs/mercurial[${PYTHON_USEDEP}]` to `RDEPEND` list. And as mercurial doesn't have py-3.9 support yet, you can't add it here either.

band-a-prend:
> so it looks like you want to add `dev-vcs/mercurial[${PYTHON_USEDEP}]` to `RDEPEND` list. And as mercurial doesn't have py-3.9 support yet, you can't add it here either.

I think changes in assignment of `HG_DEPEND` variable should fix. But I could check it later.

band-a-prend:
rebased, checked by repoman, successfully merged

juippis (approved):

PR #18585, opened 2020-12-10
media-libs/osl: require <media-libs/openimageio-2.2

EoD:
Only OSL 1.10.13 is compatible with OIIO 2.2, see
https://github.com/imageworks/OpenShadingLanguage/releases/tag/Release-1.10.13

### DIFF
--- a/media-libs/osl/osl-1.10.10.ebuild
+++ b/media-libs/osl/osl-1.10.10.ebuild
@@ -27,7 +27,7 @@ RDEPEND="
 	dev-libs/boost:=
 	dev-libs/pugixml
 	media-libs/openexr:=
-	media-libs/openimageio:=
+	<media-libs/openimageio-2.2:=
 	<sys-devel/clang-10:=
 	sys-libs/zlib:=
 	partio? ( media-libs/partio )


EoD:
@redchillipadi it might be better to stabilise v1.10.13 and drop v1.10.10 to avoid the conflicts.

redchillipadi:
Thanks for researching this, I can no longer build 1.10.10 now that openimageio-1.8.7 has been removed. But it looks like you are correct. I plan to remove 1.10.10 as soon as I can stabilise 1.10.13 which should be at the end of this month

juippis:
@redchillipadi what do you think to do here: bring back openimagei-1.8.7 or stabilize osl-1.10.13 immediately? If the current stable is broken, some action needs to be done.

redchillipadi:
I prefer to stabilise 1.10.13 immediately. Openimageio 1.8.7 is really old to resurrect, and osl works for me here. I will keep an eye out for any bugs

EoD:
@redchillipadi stabilising OSL 1.10.13 is even better. Thanks!

PR #18586, opened 2020-12-10
media-gfx/blender: Version bump to 2.83.10

EoD:
It builds fine for me and I was able to load the 2.83 demo scene successfully

### DIFF
--- a/media-gfx/blender/blender-2.83.10.ebuild
+++ b/media-gfx/blender/blender-2.83.10.ebuild
@@ -1,0 +1,320 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_7 )
+
+inherit check-reqs cmake flag-o-matic pax-utils python-single-r1 \
+	toolchain-funcs xdg-utils
+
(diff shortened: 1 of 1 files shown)


redchillipadi:
Thanks for letting me know that you are interested in 2.83 LTS, I will keep it up to date as well.

And thank you for your submission. You have already done most of the work with the pull request, I think there are just a few small steps to get your commit merged yourself if you like.

Each time you commit Gentoo needs a declaration that your contribution can be modified and redistributed under a free license, which you can give by amending your commit and adding Signed-off-by: Legal Name <email@address> to the end of your commit message. This is explained more at https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin

When you add a new ebuild, the manifest needs to be updated, which you can do by running repoman manifest and then committing the new manifest file, again with a sign off. This should fix most of the failing checks.

The only other issue is the python compatibility. There is nothing to do here for now as we should stick to python 3.7 to prevent breaking addons until blender supports a later version of python (it follows the VFX reference platform which is currently python 3.7 only)

Unfortunately I am working this weekend, but will go through your ebuild next week. It looks like the only change was the keywording, so I am confident it should be pretty easy to test and merge.

EoD:
@redchillipadi No worries, I don't mind if you open a PR directly.

PR #18587, opened 2020-12-10
sys-fs/bees: bump to 0.6.4; sync 9999; drop old

telans:
Closes: https://bugs.gentoo.org/753542
Closes: https://bugs.gentoo.org/725996


sys-fs/bees/bees-9999.ebuild
```diff
@@ -48,7 +45,14 @@ pkg_pretend() {
 			ewarn "# WARNING: CPU: 3 PID: 18172 at fs/btrfs/backref.c:1391 find_parent_nodes+0xc41/0x14e0"
 			ewarn
 		fi
-
+		if kernel_is -lt 5 0 4; then
+			ewarn "IMPORTANT: With kernel versions below 5.0.4, you may experience data corruption"
+			ewarn "due to bees using compression in btrfs. You are adviced to use a chronologically"
+			ewarn "later kernel, that includes older LTS versions released after 5.0.4:"
+			ewarn "Fixed in: 5.1+, 5.0.4+, 4.19.31+, 4.14.108+, 4.9.165+, 4.4.177+, 3.18.137+"
```
kakra, line 52: The `if` should probably check for 5.1.0 then. This was probably my fault in a previous commit.

sys-fs/bees/bees-0.6.3.ebuild
```diff
@@ -29,21 +29,11 @@ RDEPEND="${DEPEND}"
 CONFIG_CHECK="~BTRFS_FS"
 ERROR_BTRFS_FS="CONFIG_BTRFS_FS: bees does currently only work with btrfs"
 
-PATCHES=(
-	"${FILESDIR}/6001-lib-fix-non-local-lambda-expression-cannot-have-a-ca.patch"
```
kakra, line 33: I did not yet check if 0.6.3 included these commits. But I didn't bump it because it caused errors during launch of bees and never managed to make progress. Is this still the case? Does 0.6.3 work for you?

kakra (changes requested):
Thanks for the heads up, I'll look into it. I think I prepared the systemd changes already somewhere in my local repos. But I won't get to it before end of next week probably.

kakra:
@telans Could you split updating 9999 into a dedicated commit and order it first? This makes syncing up with my own branch easier.

sys-fs/bees/bees-0.6.3.ebuild
telans: I haven't used bees much, but from what I can tell it's working fine for me. I checked and all the 600* patches have been merged into 0.6.3

telans:
I've separated the commit and changed the kernel check to `-lt 5 1 0`. Does this work?

kakra:
> I've separated the commit and changed the kernel check to `-lt 5 1 0`. Does this work?

Let me figure this out in the weekend.

kakra (commented):
Version 0.6.4 has been tagged... Do you want to update?

sys-fs/bees/bees-9999.ebuild
```diff
@@ -29,14 +29,11 @@ RDEPEND="${DEPEND}"
 CONFIG_CHECK="~BTRFS_FS"
 ERROR_BTRFS_FS="CONFIG_BTRFS_FS: bees does currently only work with btrfs"
 
+PATCHES=( "${FILESDIR}/0001-HACK-musl-does-not-define-pthread_getname_np.patch" )
+
 pkg_pretend() {
 	if [[ ${MERGE_TYPE} != buildonly ]]; then
-		if kernel_is -lt 4 4 3; then
```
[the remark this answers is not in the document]
kakra: I think we can: 4.4.240 is the lowest version in portage.

kakra (approved):
LGTM but you may want to bump to 0.6.4

telans:
bumped

sys-fs/bees/bees-0.6.4.ebuild
```diff
@@ -0,0 +1,92 @@
...
+		DEFAULT_MAKE_TARGET=all
+	EOF
+	if [[ ${PV} != "9999" ]] ; then
+		cat >>localconf <<-EOF || die
+			BEES_VERSION=v${PV}
+		EOF
+	fi
+	if use tools; then
+		echo OPTIONAL_INSTALL_TARGETS=install_tools >>localconf || die
+	fi
```
juippis: What exactly this installs? Would it be sane to always install these tools, without having an USE flag for it?

juippis (commented):
```
 * /usr/libexec/bees
```
could you fix this? It's not a fatal issue but you'll have to fix it at sometime anyway.

Just let me know what you think about the `tools` IUSE.

sys-fs/bees/bees-0.6.4.ebuild
kakra: It installs some development and debugging tools to walk extent maps and test some syscalls. The ordinary user should not use them nor see them, I think. If anything, we could install them unconditionally to a directory outside of `PATH`.

kakra:
> could you fix this?

I could look into this but apparently this is not my PR so I'd defer it to a future version bump. It's probably best to upstream the fixes needed to respect  LDFLAGS.

juippis (approved):

PR #18598, opened 2020-12-10
dev-python/aiohttp: fix chardet dependency

mm1ke:
Hi,

Since `dev-python/multidict-4.7.6` was removed with commit f7e2af533f3836caa996963d67f9da27f39c9b32 `net-misc/gns3-server` won't start anymore because aiohttp, which depends on multidict, doesn't work with >multidict-5.0. 
I've reverted the commit and also fixed the dependency for aiohttp in order to reflect this issue. However, since I'm not sure how to depend on a version range (`>=dev-python/multidict-4.5.0` and `<=dev-python/multidict-5.0.0`) I've simply changed it to `=dev-python/multidict-4*`. It's not completely correct either, but since we don't have any older versions it shouldn't have any impact.

### DIFF
--- a/dev-python/aiohttp/aiohttp-3.6.2-r1.ebuild
+++ b/dev-python/aiohttp/aiohttp-3.6.2-r1.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
@@ -139,6 +139,10 @@ python_prepare_all() {
 		-i tests/test_frozenlist.py || die
 	sed -e 's:test_proxy_https_bad_response:_&:' \
 		-i tests/test_proxy_functional.py || die
+
+	# newer chardet works too
+	sed -e 's|chardet>=2.0,<4.0|chardet>=2.0|' \
+		-i setup.py aiohttp.egg-info/requires.txt || die
 
 	distutils-r1_python_prepare_all
 }
--- a/dev-python/aiohttp/aiohttp-3.7.1-r1.ebuild
+++ b/dev-python/aiohttp/aiohttp-3.7.1-r1.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
@@ -69,6 +69,10 @@ python_prepare_all() {
 	# Internet
 	sed -e 's:test_mark_formdata_as_processed:_&:' \
 		-i tests/test_formdata.py || die
+
+	# newer chardet works too
+	sed -e 's|chardet>=2.0,<4.0|chardet>=2.0|' \
+		-i setup.py aiohttp.egg-info/requires.txt || die
 
 	# takes a very long time, then fails
 	rm tests/test_pytest_plugin.py || die
--- a/dev-python/aiohttp/aiohttp-3.7.2-r1.ebuild
+++ b/dev-python/aiohttp/aiohttp-3.7.2-r1.ebuild
@@ -71,6 +71,10 @@ python_prepare_all() {
 	sed -e 's:test_mark_formdata_as_processed:_&:' \
 		-i tests/test_formdata.py || die
 
+	# newer chardet works too
+	sed -e 's|chardet>=2.0,<4.0|chardet>=2.0|' \
+		-i setup.py aiohttp.egg-info/requires.txt || die
+
 	# takes a very long time, then fails
 	rm tests/test_pytest_plugin.py || die
 
--- a/dev-python/aiohttp/aiohttp-3.7.3.ebuild
+++ b/dev-python/aiohttp/aiohttp-3.7.3.ebuild
@@ -71,6 +71,10 @@ python_prepare_all() {
 	sed -e 's:test_mark_formdata_as_processed:_&:' \
 		-i tests/test_formdata.py || die
 
+	# newer chardet works too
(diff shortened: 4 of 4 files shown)


mm1ke:
ping, can anyone have a look at it? would be great if we could fix this.

floppym:
I would suggest you un-pin the dependencies in gns3-server so that you can upgrade aiohttp to a version that doesn't require the old version of multidict.

zmedico:
> I would suggest you un-pin the dependencies in gns3-server so that you can upgrade aiohttp to a version that doesn't require the old version of multidict.

Indeed. It's been over a year since the last update to its aiohttp version pin here: https://github.com/GNS3/gns3-server/commit/8dd2831890b001c7e1af6e879ec7142918df0291

Generally, I would use sed to remove all of the pinned versions from requirements.txt, like this:
```
$ cat requirements.txt 
jsonschema==3.2.0
aiohttp==3.6.2
aiohttp-cors==0.7.0
aiofiles==0.5.0
Jinja2>=2.7.3
sentry-sdk>=0.14.4
psutil==5.6.7
async-timeout==3.0.1
distro>=1.3.0
py-cpuinfo==7.0.0
$ sed 's:[>=].*::' requirements.txt
jsonschema
aiohttp
aiohttp-cors
aiofiles
Jinja2
sentry-sdk
psutil
async-timeout
distro
py-cpuinfo
```

mm1ke:
Hi,

sorry for the rather late reply. I've now tried using newer versions of `aiohttp` and `multidict` and also removing the pinned versions in requirements.txt and ... it failed with a error in aiohttp.

However, this time a completely different error but it again has todo with a wrong dependency in `aiohttp`. Looking at https://github.com/aio-libs/aiohttp/blob/3.7/setup.py#L69 `aiohttp` doesn't work with latest `dev-python/chardet-4.0.0` which is why it creates following error in gns3: 
```
=> Local server process has stopped
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 567, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 884, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 775, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (chardet 4.0.0 (/usr/lib/python3.8/site-packages), Requirement.parse('chardet<4.0,>=2.0'), {'aiohttp'})

During handling of the above exception, another exception occurred:

  File "/usr/lib/python-exec/python3.8/gns3server", line 33, in <module>
    sys.exit(load_entry_point('gns3-server==2.2.17', 'console_scripts', 'gns3server')())
  File "/usr/lib/python-exec/python3.8/gns3server", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.8/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/lib/python3.8/site-packages/gns3server/main.py", line 29, in <module>
    import gns3server.utils.get_resource
  File "/usr/lib/python3.8/site-packages/gns3server/utils/get_resource.py", line 19, in <module>
    import pkg_resources
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3239, in <module>
    def _initialize_master_working_set():
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3222, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3251, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 569, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 582, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 770, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'chardet<4.0,>=2.0' distribution was not found and is required by aiohttp
```

I've now changed the PR to only fix the dependency in `aiohttp`. With downgrading chardet on my system `gns3` also seems to work flawless (even with newer aiohttp/multidict). If needed i can also create a bug about this issue.

thesamesam:
Sorry, just to double check: have you tried stripping the < in requirements.txt from aiohttp and seeing if it actually works with the newer chardet?

mm1ke:
> Sorry, just to double check: have you tried stripping the < in requirements.txt from aiohttp and seeing if it actually works with the newer chardet?

Hi, I've checked that now too and it works too.
I've updated the PR and instead of changing the dependency i've simply stripped the pinned version. Just to be clear - ignoring upstream dependency requirements is recommended for python packages? :)

thesamesam:
@mm1ke Yep, if it works, stripping out useless < deps from requirements.txt is something quite common we do. You might want to let upstream know in such cases that things work if tests passs.

zmedico:
Thanks!

PR #18599, opened 2020-12-10
dev-util/idea-community: fix JBR installation

HGuillemet:
The JBR (custom Jetbrains JDK) was downloaded but never installed.
Also add idea-ultimate 
The LICENSE variable should probably contains the IDEA licenses, like it's done in clion ebuild.


fee1-dead:
idea-ultimate does not use the same license. it has a EULA (because they provide educational/open source licenses that prohibits commercial use)

mpagano:
So I don't see any issue with the unpacking of the source tarballs with the current ebuild.  I also don't plan on maintaining ultimate

PR #18603, opened 2020-12-10
app-i18n/transifex-client: fix dependencies, add support for Python 3.9

robert7k:
Closes: https://bugs.gentoo.org/758809

### DIFF
--- a/app-i18n/transifex-client/transifex-client-0.14.2.ebuild
+++ b/app-i18n/transifex-client/transifex-client-0.14.2.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{6,7,8} )
+PYTHON_COMPAT=( python3_{6,7,8,9} )
 DISTUTILS_USE_SETUPTOOLS=rdepend
 
 inherit distutils-r1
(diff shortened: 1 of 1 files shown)


app-i18n/transifex-client/transifex-client-0.14.2.ebuild
```diff
@@ -18,7 +18,7 @@ KEYWORDS="~amd64 ~x86"
 IUSE="test"
 RESTRICT="!test? ( test )"
 
-BDEPEND="test? ( dev-python/mock[${PYTHON_USEDEP}] )"
+BDEPEND="test? ( <dev-python/mock-4.0[${PYTHON_USEDEP}] )"
```
juippis, line 21: I hope this is patchable... has upstream been noted/asked about it?

juippis (approved):

app-i18n/transifex-client/transifex-client-0.14.2.ebuild
robert7k: Seems we can easily patch it: https://github.com/gentoo/gentoo/pull/19057

robert7k: I also proposed to change that upstream: https://github.com/transifex/transifex-client/pull/314

PR #18604, opened 2020-12-11
gui-libs/libhandy: bump to 1.0.2

xgqt:
Changed the URI, bumped version, removed MY_P, reordered some items in DEPEND.

Gentoo's libhandy is very outdated, I namely need it for new version of komikku (https://github.com/gentoo/guru/tree/master/app-text/komikku) other than GNOME apps I'd like to push to GURU.


xgqt:
https://bugs.gentoo.org/759439

leio:
This needs to be in a separate slot and fully parallel-installable with the previous slot

leio:
Use gnome.org eclass for SRC_URI to use gnome distfiles mirror, not a tag from gitlab

xgqt:
I haven't had any conflict with USE="glade gtk-doc introspection vala"

gui-libs/libhandy/libhandy-1.0.2.ebuild
```diff
@@ -0,0 +1,57 @@
...
+
+EAPI=7
+
+inherit gnome.org meson xdg vala virtualx
+
+DESCRIPTION="Library with GTK widgets for mobile phones"
+HOMEPAGE="https://gitlab.gnome.org/GNOME/libhandy/"
+
+LICENSE="LGPL-2.1+"
+SLOT="1.0/0"
```
leio, line 12: This should be SLOT="1", because they (as I understand it) guarantee ABI stability now within the parallel-installable package, and the package is called libhandy-1, not libhandy-1.0

leio (changes requested):

PR #18611, opened 2020-12-11
Games board/cockatrice version bump 20200823

GuillaumeSeren:


games-board/cockatrice/cockatrice-20200823.ebuild
```diff
@@ -38,32 +39,34 @@ BDEPEND="
...
 	cmake_src_configure
 }
+
+pkg_postinst() {
+	xdg_icon_cache_update
+}
+
+pkg_postrm() {
+	xdg_icon_cache_update
+}
```
chewi, line 72: You're using these xdg functions without inheriting the `xdg-utils` eclass. You should just inherit the `xdg` eclass though so you don't have to define these functions at all. If you order it `xdg cmake` rather than `cmake xdg` then you can avoid a bad interaction between the two.

chewi (changes requested):

games-board/cockatrice/cockatrice-20200823.ebuild
GuillaumeSeren: Hey @chewi thank you for the feedback, I have upgraded the branch.

GuillaumeSeren:
Thank you for the merge @chewi

PR #18619, opened 2020-12-12
sci-libs/geos: Upstream dropped ruby and python support

doctaweeks:
Bug: https://bugs.gentoo.org/759295

### DIFF
--- a/sci-libs/geos/geos-3.9.0-r1.ebuild
+++ b/sci-libs/geos/geos-3.9.0-r1.ebuild
@@ -10,16 +10,11 @@ SRC_URI="http://download.osgeo.org/geos/${PN}-${PV}.tar.bz2"
 LICENSE="LGPL-2.1"
 SLOT="0"
 KEYWORDS="~amd64 ~arm ~arm64 ~ia64 ~ppc ~ppc64 ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~x86-macos ~sparc-solaris ~sparc64-solaris ~x64-solaris"
-IUSE="doc ruby static-libs"
+IUSE="doc static-libs"
 
 BDEPEND="
 	doc? ( app-doc/doxygen )
-	ruby? ( dev-lang/swig:0 )
 "
-RDEPEND="
-	ruby? ( dev-lang/ruby:* )
-"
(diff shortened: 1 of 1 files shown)


titanofold:
Thanks for submitting a fix.

However, there's another wrinkle: Upstream doesn't have a 3.9.0 released. They only have 3.9.0rc1. So, we need to drop 3.9.0 and replace it with 3.9.0rc1. I think they accidentally made the rc1 available as a release, briefly.

Would you alter your PR to do that as well?

sci-libs/geos/geos-3.9.0-r1.ebuild
```diff
@@ -10,16 +10,11 @@ SRC_URI="http://download.osgeo.org/geos/${PN}-${PV}.tar.bz2"
 LICENSE="LGPL-2.1"
 SLOT="0"
 KEYWORDS="~amd64 ~arm ~arm64 ~ia64 ~ppc ~ppc64 ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~x86-macos ~sparc-solaris ~sparc64-solaris ~x64-solaris"
-IUSE="doc ruby static-libs"
+IUSE="doc static-libs"
 
 BDEPEND="
 	doc? ( app-doc/doxygen )
-	ruby? ( dev-lang/swig:0 )
 "
```
titanofold, line 17: Might as well make this one line since it only has one depend.

sci-libs/geos/geos-3.9.0-r1.ebuild
```diff
@@ -30,8 +25,6 @@ src_prepare() {
 
 src_configure() {
 	local myeconfargs=(
-		--disable-python
-		$(use_enable ruby)
 		$(use_enable static-libs static)
 	)
```
titanofold, line 29: Same deal here. Make it a single line.

titanofold (changes requested):
Thanks for submitting a fix.

However, there's another wrinkle: Upstream doesn't have a 3.9.0 released. They only have 3.9.0rc1. So, we need to drop 3.9.0 and replace it with 3.9.0rc1. I think they accidentally made the rc1 available as a release, briefly.

Would you alter your PR to do that as well?

PR #18626, opened 2020-12-12
www-plugins/lightspark: bump to 0.8.3

000exploit:


mgorny:
> Signed-off-by: 000exploit <illialoo99@gmail.com>

Please forgive me if I'm wrong but that doesn't look like a real name.

PR #18628, opened 2020-12-12
Haskell: bump packages and drop old

solpeth:
Tested in an `~amd64` container. Commits pulled from `::haskell`.


trofi:
Pulled. Thank you!

solpeth:
@trofi another PR on its way shortly to remove `tasty-hedgehog-1.0.0.2-r0` which requires `hedgehog-1.0.2`. Sorry to have missed this!

PR #18630, opened 2020-12-12
dev-haskell/tasty-hedgehog: drop old

solpeth:
Drop `-r0` due its dependency on old and no longer available versions of `hedgehog`.


trofi:
Pulled as well. Thank you!

PR #18634, opened 2020-12-13
media-libs/openimageio, sci-geosciences/qmapshack: remove unused patches

mm1ke:


epsilon-0:
openimageio looks good with me.

PR #18637, opened 2020-12-13
Further merges from ::prefix

thesamesam:
Affected ebuilds:
- ~sys-libs/zlib: minor changes for Cygwin only~
- ~app-arch/unzip: across the board Prefix changes but limited in scope (just CHOST parsing)~
- app-arch/bzip2: still needs work, here as a draft

All changes are prefix-only and erring on the side of caution (i.e. do nothing in non-prefix case, even if harmless).

This is mostly here to help me organise my work. They're not necessarily ready to review or going to be merged as-is. Comments from passers-by in prefix or base-system are welcome but not being solicited yet.

TODO:
- Thoroughly test and examine the bzip2 changes (especially with lib naming)

Future areas:
- ncurses
- readline
- [db.eclass](https://bugs.gentoo.org/show_bug.cgi?id=673322)?

### DIFF
--- a/app-arch/bzip2/bzip2-1.0.8-r1.ebuild
+++ b/app-arch/bzip2/bzip2-1.0.8-r1.ebuild
@@ -6,7 +6,7 @@
 
 EAPI=7
 
-inherit toolchain-funcs multilib-minimal usr-ldscript
+inherit toolchain-funcs multilib-minimal usr-ldscript prefix
 
 DESCRIPTION="A high-quality data compressor used extensively by Gentoo Linux"
 HOMEPAGE="https://sourceware.org/bzip2/"
@@ -14,7 +14,7 @@ SRC_URI="https://sourceware.org/pub/${PN}/${P}.tar.gz"
 
 LICENSE="BZIP2"
 SLOT="0/1" # subslot = SONAME
-KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86 ~ppc-aix ~x64-cygwin ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~x86-macos ~m68k-mint ~sparc-solaris ~sparc64-solaris ~x64-solaris ~x86-solaris ~x86-winnt"
 IUSE="static static-libs"
 
 PATCHES=(
@@ -25,6 +25,8 @@ PATCHES=(
 	"${FILESDIR}"/${PN}-1.0.3-no-test.patch
 	"${FILESDIR}"/${PN}-1.0.8-mingw.patch #393573
 	"${FILESDIR}"/${PN}-1.0.8-out-of-tree-build.patch
+
+	"${FILESDIR}"/${PN}-1.0.6-r7-checkenv.patch # for AIX, Darwin?
 )
 
 DOCS=( CHANGES README{,.COMPILATION.PROBLEMS,.XML.STUFF} manual.pdf )
@@ -41,6 +43,28 @@ src_prepare() {
 		-e 's:ln -s -f $(PREFIX)/bin/:ln -s -f :' \
 		-e 's:$(PREFIX)/lib:$(PREFIX)/$(LIBDIR):g' \
 		Makefile || die
+
+	if use prefix ; then
+		hprefixify -w "/^PATH=/" bz{diff,grep,more}
+
+		# This is a makefile for Darwin, which already "includes" sane so
+		cp "${FILESDIR}"/${PN}-1.0.6-Makefile-libbz2_dylib Makefile-libbz2_dylib || die
+
+		if [[ ${CHOST} == *-hpux* ]] ; then
+			sed -i -e 's,-soname,+h,' Makefile-libbz2_so || die "cannot replace -soname with +h"
+			if [[ ${CHOST} == hppa*-hpux* && ${CHOST} != hppa64*-hpux* ]] ; then
+				sed -i -e '/^SOEXT/s,so,sl,' Makefile-libbz2_so || die "cannot replace so with sl"
+				sed -i -e '/^SONAME/s,=,=${EPREFIX}/lib/,' Makefile-libbz2_so || die "cannt set soname"
+			fi
+		fi
+
+		if [[ ${CHOST} == *-cygwin* ]] ; then
+			sed -i -e "s/-o libbz2\.so\.${PV}/-Wl,--out-implib=libbz2$(get_libname ${PV})/" \
+				   -e "s/-Wl,-soname -Wl,libbz2\.so\.1/-o cygbz2-${PV%%.*}.dll/" \
+				   -e "s/libbz2\.so/libbz2$(get_libname)/g" \
+				Makefile-libbz2_so
+		fi
+	fi
 }
 
 bemake() {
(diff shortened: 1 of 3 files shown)


app-arch/bzip2/bzip2-1.0.8-r1.ebuild
```diff
@@ -6,15 +6,15 @@
...
+inherit toolchain-funcs multilib-minimal usr-ldscript prefix
 
 DESCRIPTION="A high-quality data compressor used extensively by Gentoo Linux"
 HOMEPAGE="https://sourceware.org/bzip2/"
 SRC_URI="https://sourceware.org/pub/${PN}/${P}.tar.gz"
 
 LICENSE="BZIP2"
 SLOT="0/1" # subslot = SONAME
-KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86 ~ppc-aix ~x64-cygwin ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~x86-macos ~m68k-mint ~sparc-solaris ~sparc64-solaris ~x64-solaris ~x86-solaris ~x86-winnt"
```
grobian, line 17: drop aix and mint

app-arch/bzip2/bzip2-1.0.8-r1.ebuild
```diff
@@ -25,6 +25,8 @@ PATCHES=(
 	"${FILESDIR}"/${PN}-1.0.3-no-test.patch
 	"${FILESDIR}"/${PN}-1.0.8-mingw.patch #393573
 	"${FILESDIR}"/${PN}-1.0.8-out-of-tree-build.patch
+
+	"${FILESDIR}"/${PN}-1.0.6-r7-checkenv.patch # for AIX, Darwin?
```
grobian, line 29: sync with prefix, this is dropped now

app-arch/bzip2/bzip2-1.0.8-r1.ebuild
```diff
@@ -41,6 +43,28 @@ src_prepare() {
...
 		-e 's:$(PREFIX)/lib:$(PREFIX)/$(LIBDIR):g' \
 		Makefile || die
+
+	if use prefix ; then
+		hprefixify -w "/^PATH=/" bz{diff,grep,more}
+
+		# This is a makefile for Darwin, which already "includes" sane so
+		cp "${FILESDIR}"/${PN}-1.0.6-Makefile-libbz2_dylib Makefile-libbz2_dylib || die
+
+		if [[ ${CHOST} == *-hpux* ]] ; then
```
grobian, line 53: the hpux bit is now dropped

app-arch/bzip2/bzip2-1.0.8-r1.ebuild
```diff
@@ -41,6 +43,28 @@ src_prepare() {
 		-e 's:ln -s -f $(PREFIX)/bin/:ln -s -f :' \
 		-e 's:$(PREFIX)/lib:$(PREFIX)/$(LIBDIR):g' \
 		Makefile || die
+
+	if use prefix ; then
+		hprefixify -w "/^PATH=/" bz{diff,grep,more}
```
grobian, line 48: hprefixify is already non-prefix-safe, so no need for a conditional

app-arch/bzip2/bzip2-1.0.8-r1.ebuild
```diff
@@ -41,6 +43,28 @@ src_prepare() {
...
+
+		# This is a makefile for Darwin, which already "includes" sane so
+		cp "${FILESDIR}"/${PN}-1.0.6-Makefile-libbz2_dylib Makefile-libbz2_dylib || die
```
grobian, line 51: this file affects noone but Darwin users, so no need for a conditional either

app-arch/bzip2/bzip2-1.0.8-r1.ebuild
```diff
@@ -41,6 +43,28 @@ src_prepare() {
...
+
+		if [[ ${CHOST} == *-hpux* ]] ; then
+			sed -i -e 's,-soname,+h,' Makefile-libbz2_so || die "cannot replace -soname with +h"
+			if [[ ${CHOST} == hppa*-hpux* && ${CHOST} != hppa64*-hpux* ]] ; then
+				sed -i -e '/^SOEXT/s,so,sl,' Makefile-libbz2_so || die "cannot replace so with sl"
+				sed -i -e '/^SONAME/s,=,=${EPREFIX}/lib/,' Makefile-libbz2_so || die "cannt set soname"
+			fi
+		fi
+
+		if [[ ${CHOST} == *-cygwin* ]] ; then
```
grobian, line 61: this is already conditionalised (for a Prefix-only target) so no need for conditional

-> drop the if use prefix

app-arch/bzip2/bzip2-1.0.8-r1.ebuild
```diff
@@ -53,9 +77,26 @@ bemake() {
...
-	# Make sure we link against the shared lib #504648
-	ln -s libbz2.so.${PV} libbz2.so || die
+	local checkopts=
+	case "${CHOST}" in
+		*-darwin*)
+			bemake PREFIX="${EPREFIX}"/usr -f "${S}"/Makefile-libbz2_dylib all
+			# FWIW, #504648 like for .so below
+			ln -sf libbz2.${PV}.dylib libbz2.dylib || die
+		;;
+		*-mint*)
```
grobian, line 87: this bit was dropped

app-arch/bzip2/bzip2-1.0.8-r1.ebuild
```diff
@@ -53,9 +77,26 @@ bemake() {
...
+			# do nothing, no shared libraries
+			:
+		;;
+		*)
+			bemake -f "${S}"/Makefile-libbz2_so all
+			# Make sure we link against the shared lib #504648
+			if [[ $(get_libname) != $(get_libname ${PV}) ]] ; then
```
grobian, line 94: this is a mint remnant, I've dropped it in prefix alreay, please sync it up

app-arch/bzip2/bzip2-1.0.8-r1.ebuild
```diff
@@ -66,17 +107,24 @@ multilib_src_install() {
...
-	for v in libbz2.so{,.{${PV%%.*},${PV%.*}}} ; do
-		dosym libbz2.so.${PV} /usr/$(get_libdir)/${v}
-	done
+	dolib.so libbz2$(get_libname ${PV})
+
+	if [[ ${CHOST} == *-cygwin* ]] ; then
+		dobin cygbz2-${PV%%.*}.dll
+	fi
+
+	if [[ $(get_libname) != $(get_libname ${PV}) ]] ; then
```
grobian, line 116: revisit this

app-arch/bzip2/files/bzip2-1.0.6-checkenv.patch
```diff
@@ -0,0 +1,21 @@
+--- a/Makefile	2005-02-17 05:28:24.000000000 -0600
```
grobian, line 1: drop

grobian (commented):
this is a nice illustration of why I'm dropping old cruft from the overlay first

In some cases (which is the majority of remaining packages in the prefix tree) we can reduce the differences between gx86 and prefix quite a lot when we drop the bits for the difficult targets.  Cygwin is the remaining "difficult" target, unfortunately, but many dubious constructs can be dropped, which really is good for everybody :)

grobian:
I've been thinking about this one, to see if we can get the changes down.  The Cygwin changes are basically what the Darwin changes tried to avoid, by making a new Makefile.  Yet, instead of sedding the heck out of everything, perhaps just (over)writing the Makefile-so thing with the appropriate few rules would be enough.

thesamesam:
Good point! Not forgot about this, just got exams right now. Happy to finish the work.

grobian:
Odd time for exams!  Anyway, focus on those, priorities seem clear to me in that regard :)  Good luck!

PR #18639, opened 2020-12-13
x11-terms/kitty: disable builtin automatic update check

TheJJ:
disable the automatic update check which periodically contacts
the developer's webserver.

Signed-off-by: Jonas Jelten <jj@sft.lol>

### DIFF
--- a/x11-terms/kitty/kitty-0.19.1.ebuild
+++ b/x11-terms/kitty/kitty-0.19.1.ebuild
@@ -80,6 +80,7 @@ src_compile() {
 	"${EPYTHON}" setup.py \
 		--verbose $(usex debug --debug "") \
 		--libdir-name $(get_libdir) \
+		--update-check-interval=0 \
 		linux-package || die "Failed to compile kitty."
 }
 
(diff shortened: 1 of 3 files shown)


juippis:
@end222 can you comment on this, is this the right way to disable it, and is disabling it a good idea?

juippis:
@TheJJ please ping us again after weekend if there's no response from the maintainer and we'll take a deeper look.

TheJJ:
The update check basically fetches a url from the developer's server and checks if there's a newer version. I'd like to disable that since i don't want my kitties phoning home :) I've disabled it with a patch currently.

Arch has already disabled it: https://github.com/archlinux/svntogit-community/blob/557466e354c15b4e303b52fbb2b4affe10f9ec1b/trunk/PKGBUILD#L22

end222:
It is it said in the documentation link that @TheJJ shared, it is the correct way of doing it.
I personally think that it would be a good idea to disable it as it will improve the privacy of the package by not pinging to the dev's website and it will also remove the update notifications, which get quite annoying as they are shown too often.

juippis (approved):
Thanks! I didn't doubt the motivation as it's a good one, just that this was the cohesive way of doing it.

PR #18643, opened 2020-12-13
net-misc/openntpd: version bump to 6.8_p1

epsilon-0:

---

Gets rid of the patch.


juippis (approved):

net-misc/openntpd/openntpd-6.8_p1.ebuild
```diff
@@ -0,0 +1,65 @@
...
+
+inherit systemd
+
+MY_P="${P/_p/p}"
+
+DESCRIPTION="Lightweight NTP server ported from OpenBSD"
+HOMEPAGE="http://www.openntpd.org/"
+SRC_URI="mirror://openbsd/OpenNTPD/${MY_P}.tar.gz"
+
+LICENSE="BSD GPL-2"
```
juippis, line 14: By the way there's a catch to this combination, but as you're not the maintainer of this package, I won't have you do the required detective work for it.

But since it's also a quiz question, you should ;)
(ask your mentor about it)

PR #18644, opened 2020-12-13
dev-python/subunit: remove -Werror from configure.ac

crabbedhaloablution:
Closes: https://bugs.gentoo.org/744313


floppym (changes requested):
I would prefer this be done via a patch instead of sed.

prometheanfire (changes requested):
I think I agree, a patch would be better (and could include the bug link in the patch header).

crabbedhaloablution:
> I would prefer this be done via a patch instead of sed.
Updated.

prometheanfire (commented):
Is it possible to use the `PATCHES=()` syntax here?  I feel like it should be.

crabbedhaloablution:
> Is it possible to use the `PATCHES=()` syntax here? I feel like it should be.

Sure. Updated.

PR #18649, opened 2020-12-14
app-shell/bash: add Alacritty as a colorfull terminal

vitaly-zdanevich:


vitaly-zdanevich:
Found that I have a file `/etc/DIR_COLORS` but for some reasons Alacritty after emerging was without colors, please review - I am not expert in this.

vitaly-zdanevich:
Wrong change - this does not change anything...

PR #18651, opened 2020-12-14
dev-libs/s2n: Fix USE="test" and remove -Werror

Yamakuzure:
USE="test" was broken, because building of a shared library does not work at the moment.
* Fixed by removing hidden default visibility.
* See: https://github.com/awslabs/s2n/issues/2401

`-Werror` is not recommended for releases and should always be disabled when encountered in build-logs, because there are numerous cases where this breaks without purpose.
* Removed from CMakeLists.txt and s2n.mk

Closes: https://bugs.gentoo.org/759796
Closes: https://bugs.gentoo.org/759799


juippis (approved):
This doesn't really require a revbump, but let's go with it.

PR #18652, opened 2020-12-14
gui-apps/swappy: New package

emlove:
## Description
Swappy is a Wayland native snapshot and editor tool, inspired by Snappy on macOS. Works great with grim, slurp and sway. But can easily work with other screen copy tools that can output a final PNG image to stdout.


juippis:
@epsilon-0, @Freed0mBear your territory. Would you be interested in this?

epsilon-0:
I have this in my overlay. After revisiting that, there seem to be quite a few things missing...
I've attached my version of the ebuild at the end, please make your ebuild use the additional things that I have enabled.
The postinst/rm are the important ones. man-pages should be optional. The libnotify is should be tied to the meson option.
It needs grim at runtime not at build/link time.

```

EAPI=7

inherit meson xdg-utils

DESCRIPTION="wayland native snapshot editing tool"
HOMEPAGE="https://github.com/jtheoof/swappy"

MY_PV=$(ver_rs 2 '-')
if [[ ${PV} == 9999 ]]; then
	inherit git-r3
	EGIT_REPO_URI="https://github.com/jtheoof/swappy"
else
	SRC_URI="https://github.com/jtheoof/swappy/archive/v${PV}.tar.gz -> ${P}.tar.gz"
	KEYWORDS="~amd64 ~arm64"
fi

LICENSE="MIT"
SLOT="0"
IUSE="+libnotify man"

DEPEND="
	x11-libs/cairo
	x11-libs/gtk+:3
	x11-libs/pango
	libnotify? ( x11-libs/libnotify )
"
RDEPEND="${DEPEND}
	gui-apps/grim
"
BDEPEND="
	virtual/pkgconfig
	man? ( app-text/scdoc )
"

src_configure() {
	local emesonargs=(
		$(meson_feature man man-pages)
		$(meson_feature libnotify)
	)
	meson_src_configure
}

pkg_postinst() {
	xdg_icon_cache_update
	xdg_desktop_database_update
}

pkg_postrm() {
	xdg_icon_cache_update
	xdg_desktop_database_update
}
```

emlove:
Thanks! I didn't realize there was an ebuild out there already.

I've incorporated what you have in the ebuild, the only thing I'm not sure about is the dependency on grim. It looks like grim is a tool suggested for use together with swappy, but swappy doesn't actually depend on it.

media-gfx/swappy/swappy-1.2.1.ebuild
```diff
@@ -0,0 +1,54 @@
...
+SLOT="0"
+IUSE="+libnotify +clipboard man"
+
+DEPEND="
+	dev-libs/glib:2
+	x11-libs/gtk+:3
+	x11-libs/cairo
+	x11-libs/pango
+	media-fonts/fontawesome[otf]
+	clipboard? ( gui-apps/wl-clipboard )
```
epsilon-0, line 29: This is an optional runtime dependency.
This should not be a USE flag, instead look at the `optfeature.eclass` and in `pkg_postinst` to give an message saying that wl-clipboard can be used for optional clipboard support.

media-gfx/swappy/swappy-1.2.1.ebuild
```diff
@@ -0,0 +1,54 @@
...
+LICENSE="MIT"
+SLOT="0"
+IUSE="+libnotify +clipboard man"
+
+DEPEND="
+	dev-libs/glib:2
+	x11-libs/gtk+:3
+	x11-libs/cairo
+	x11-libs/pango
+	media-fonts/fontawesome[otf]
```
epsilon-0, line 28: fontawesome should be in `RDEPEND`.

epsilon-0 (changes requested):
Rest looks good :+1:

epsilon-0:
> It looks like grim is a tool suggested for use together with swappy, but swappy doesn't actually depend on it.

Yes, you are correct. I made this ebuild quite a while ago and didn't remember the details.

media-gfx/swappy/swappy-1.2.1.ebuild
emlove: TIL!

epsilon-0:
Please also change the package category to `gui-apps`.
All wayland gui applications are using that, including grim and slurp.

Along with that change, please also change the commit message to satisfy https://www.gentoo.org/glep/glep-0066.html.
Something like:
```
gui-apps/swappy: wayland native snapshot tool

allows editing and modifying snapshots along
with support for other native wayland tools
like grim and wl-clipboard

Signed-off-by: Emily Mills <emily@emlove.me>
```

emlove:
Done!

epsilon-0:
Whoops, don't know how I overlooked this but please you add yourself as maintainer to metadata.xml.
Not sure who the second maintainer should be just yet. 
I'm not a full dev yet so I can't take the second spot.

epsilon-0:
Perfect! Did install and runtime test on my system, barring a warning about an incorrect `.desktop` file, package is working nicely :+1: 
The .desktop file error is something upstream should fix, this is not our problem :neutral_face: 
`swappy` is still found in menus so its OK.

Someone from proxy-maintainers will merge when they get time.

emlove:
OK! I've got the metadata updated with the maintainers.

I read through https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers , and it looks like I'll be a Proxied Maintainer for this project. I don't see any actionable steps though, other than having a willingness to maintain the package. Please let me know if I missed something!

epsilon-0:
proxy-maint will tell you if they need anything, but for now you are good :)

emlove:
Thanks for your help!

gui-apps/swappy/Manifest
```diff
@@ -0,0 +1 @@
+DIST swappy-1.2.1.tar.gz 648773 BLAKE2B f60b079d7bcb28708afb8d016f1b997a7b7860b84e614812832e4d6e9bcd2b7f546f544ee2dba6e0e913434f5d13c2377e84419811b3661ffbfb9be1ca61d0be SHA512 e724283845d5555e560129e976dafa1c03c134a3b685b7264138b18b3c7c6a6ea7c4de5862e0b8b5ec3101ec3ba0aeff939e8acee44d672d5eae67517609bb07
```
juippis, line 1: Could you please update to the latest release? Sorry it's taken a while, we prioritize reviewing existing packages before adding new packages to main tree.

gui-apps/swappy/swappy-1.2.1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit meson xdg-utils optfeature
+
+DESCRIPTION="A Wayland native snapshot and editor tool, inspired by Snappy on macOS"
+HOMEPAGE="https://github.com/jtheoof/swappy"
+
+if [[ ${PV} == 9999 ]]; then
```
juippis, line 11: (You can use `*9999` for a bit more accurate live matching, but it is not necessary - a taste preference)

gui-apps/swappy/swappy-1.2.1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+	x11-libs/cairo
+	x11-libs/pango
+	libnotify? ( x11-libs/libnotify )
+"
+RDEPEND="${DEPEND}
+	media-fonts/fontawesome[otf]
+"
+BDEPEND="
+	virtual/pkgconfig
+	man? ( app-text/scdoc )
```
juippis, line 35: Since `scdoc` is a **very** cheap BDEP, could you consider always building man page unconditionally? Yes, it just installs one file.

gui-apps/swappy/swappy-1.2.1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+"
+
+src_configure() {
+	local emesonargs=(
+		$(meson_feature man man-pages)
+		$(meson_feature libnotify)
```
juippis, line 41: FYI it seems like the latest release has gotten rid of libnotify option. I'm not sure whether they disabled it, or made it unconditional. Please do confirm.

gui-apps/swappy/swappy-1.2.1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+	xdg_icon_cache_update
+	xdg_desktop_database_update
+
+	optfeature "Persist clipboard after closing" gui-apps/wl-clipboard
+}
+
+pkg_postrm() {
+	xdg_icon_cache_update
+	xdg_desktop_database_update
+}
```
juippis, line 56: If you want, you can replace a lot of this text simply by inheriting *xdg.eclass*. Then you can do 
```suggestion
pkg_postinst() {
	xdg_pkg_postinst
	optfeature "persist clipboard after closing" gui-apps/wl-clipboard
}
```
Oh yeah, the `optfeature` already adds a word for you, so currently it looks like this:
` *   gui-apps/wl-clipboard for Persist clipboard after closing`. I don't mind the capital `P`, but someone might ;) 

All the above comments apply to -9999 ebuild as well.

juippis (commented):
I am also getting:
```
 * Messages for package gui-apps/swappy-9999:

 * QA Notice: This package installs one or more .desktop files that do not
 * pass validation.
 * 
 * 	/usr/share/applications/swappy.desktop: error: value "Utility;Graphics;Annotation;" for key "Categories" in group "Desktop Entry" contains an unregistered value "Annotation"; values extending the format should start with "X-"
```
You can test the .desktop file with `desktop-file-validate /usr/share/applications/swappy.desktop`. Please `sed` or patch it so it passes xdg standards.

emlove:
Thanks for the review! Pushed up those changes.

juippis (approved):

PR #18654, opened 2020-12-14
dev-libs/aws-c-event-stream: Add missing dev-libs/aws-c-io dependency

Yamakuzure:
Sorry, I forgot that one. :facepalm: 

Closes: https://bugs.gentoo.org/759805


juippis (approved):
No need to revbump for build failures.

PR #18655, opened 2020-12-14
Haskell: bump packages, drop old

solpeth:
Tested in an `~amd64` container. Commits pulled from my commits in `::haskell`.


trofi:
Pulled as is. Thank you!

PR #18656, opened 2020-12-14
dev-libs/aws-c-io: Disable tests requiring an internet connection

Yamakuzure:
Some of the tests require internet connectivity. On an offline machine these tests will therefore fail.

As network sandboxing is enabled by default for years now, it is very unlikely that anybody would perform these tests anyway.

Closes: https://bugs.gentoo.org/759802

### DIFF
--- a/dev-libs/aws-c-io/aws-c-io-0.7.0-r1.ebuild
+++ b/dev-libs/aws-c-io/aws-c-io-0.7.0-r1.ebuild
@@ -38,5 +38,15 @@ src_configure() {
 		-DBUILD_SHARED_LIBS=$(usex !static-libs)
 		-DBUILD_TESTING=$(usex test)
 	)
+
+	if use test; then
+		# (#759802) Due to network sandboxing of portage, internet connectivity
+		# tests will always fail. If you need a USE flag, because you want/need
+		# to perform these tests manually, please open a bug report for it.
+		mycmakeargs+=(
+			-DENABLE_NET_TESTS=OFF
+		)
+	fi
+
 	cmake_src_configure
 }


dev-libs/aws-c-io/aws-c-io-0.7.0-r1.ebuild
```diff
@@ -38,5 +42,12 @@ src_configure() {
 		-DBUILD_SHARED_LIBS=$(usex !static-libs)
 		-DBUILD_TESTING=$(usex test)
 	)
+
+	if use test; then
+		mycmakeargs+=(
+			-DENABLE_NET_TESTS=$(usex net-tests)
```
thesamesam, line 48: This isn't the right approach here unfortunately, we need to disable them unconditionally.

FEATURES="network-sandbox" is on by default (for a few years now), do you not have this enabled?

Yamakuzure: This is what emerge --info gives me:

````
FEATURES="assume-digests binpkg-docompress binpkg-dostrip binpkg-logs ccache compressdebug
config-protect-if-modified distlocks ebuild-locks fixlafiles ipc-sandbox merge-sync multilib-strict
network-sandbox news parallel-fetch pid-sandbox preserve-libs protect-owned qa-unresolved-soname-deps
sandbox sfperms split-log splitdebug strict strict-keepdir unknown-features-warn unmerge-logs
unmerge-orphans userfetch userpriv usersandbox usersync xattr"
````

So, yes, network-sandbox is enabled.

However, I can do
````
USE="test net-tests" ebuild aws-c-io-0.7.0-r1.ebuild test
````
as a regular user, and it just works. As root it does not, and I think that's completely fine. ('portage' is a non-root user, too, right?)

But if you think that it would be better to completely remove the possibility to perform network related tests, instead of disabling them by default, then I could do that, yes.
From my point of view, it does not make much sense to remove something that is a working default, just because it _might_ not work for some settings. But then, I could just happen to have a setting where it works but being a minuscule minority.

:man_shrugging: That's the reason why I highly appreciate any input/feedback from others. What's the bes solution here?

Comment (juippis) [reply dev-libs/aws-c-io/aws-c-io-0.7.0-r1.ebuild:48]: No arch tester will ever have network-sandbox open. This is most likely why the test phase hangs for me, 
```
146/149 Test  #99: tls_server_multiple_connections .........................................***Failed   10.06 sec
^C

Exiting on signal 2
```
If you can identify the exact tests that require network connection and disable them, it'd be fantastic. And in general you should enable `net-tests` with `tests` IUSE, not make them a separate one.

Yamakuzure: @juippis : The net-tests are those that require an internet connection, and they are enabled by default.
The USE flag has the purpose of disabling them by default, but keeping an option to enable them if someone wants these tests to run.

But I am still wondering whether it is really useful to enable these tests. How many regular users (apart from me) would actually have them run? Zero?

I guess I'll go with what @thesamesam suggested, and disable them unconditionally. I could leave a message via comment there, that if there was a good reason to make them available again, a bug against that could always be opened.

juippis (approved):

PR #18658, opened 2020-12-14
net-misc/chrony: lots of little nits.

crabbedhaloablution:
Setup alternatives for sechash support, Nettle and NSS.
NSS is no longer a null-option. Do note that with --as-needed,
chronyd will only link against libfreebl3.so, not other parts of
NSS.
NSS, though, *is* kind of useless since it's only good for sechash
support.

Removed pkg-config sed since upstream fixed $PKG_CONFIG support in
commit 9a9c0d7b99d7478e1431bd3fa8c225bdc46b3df4

Removed RELOADDNS sed since upstream fixed the bug in
commit ff466439fc22d6a47594f5dda33a5bdb726e92b8
Bug: https://bugs.gentoo.org/739684

Added debug use-flag. Use -dd option to chronyd to get DEBUG_LOG
output. Added in 2013(!)
commit 4bbc5520b8e340a08795e245784beb42c1254c0b

Added "export SOURCE_DATE_EPOCH=1607976314".  Update when updating
ebuild. If not set, "date '+%s'" is used instead.
This is a problem when the build host is not synchronized to NTP,
since if time obtained from NTP is prior to build date, NTP time
may underflow.

Enabled NTS support in ebuild. Needs net-libs/gnutls and dev-libs/nettle.
NTS is the only thing net-libs/gnutls is used for.
Reported-By: Matt Whitlock <gentoo@mattwhitlock.name>
Closes: https://bugs.gentoo.org/759526

### DIFF
--- a/net-misc/chrony/chrony-4.0-r1.ebuild
+++ b/net-misc/chrony/chrony-4.0-r1.ebuild
@@ -1,0 +1,190 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit systemd tmpfiles toolchain-funcs
+
+DESCRIPTION="NTP client and server programs"
+HOMEPAGE="https://chrony.tuxfamily.org/"
+
+if [[ ${PV} == "9999" ]]; then
+	inherit git-r3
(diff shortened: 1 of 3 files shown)


thesamesam:
Thank you!

PR #18659, opened 2020-12-14
x11-misc/i3status: make pulseaudio optional

creich:
added patches
bumped version to 2.13-r1


trofi:
Pulled (in a squashed form). Thank you!

PR #18665, opened 2020-12-15
media-gfx/blender: Version bump to 2.83.10 LTS

redchillipadi:
Version bump to blender for 2.83.10 LTS

The only change from the current 2.83.4 ebuild is the keywording ~amd64


juippis (approved):
LTS or not, I believe python-3.7 packages are being looked to being removed. So this may not be as long-lived.

redchillipadi:
Blender plans to support 2.83 LTS until April 2022, but will bring out 2.93 LTS in April 2021.

However the VFX platform is python 3.7 only for 2021, so if we do last-rite python 3.7 in June, two years ahead of the upstream python EOL, we might be in a position where we need to force all versions of blender to run on unsupported python 3.8.

Hopefully we will start to get some user bug reports once the default python target changes to python 3.8 (this month?) to determine how many people will have addon breakage.

PR #18673, opened 2020-12-15
games-emulation/openmsx: resurrect after python2 purge (with 16.0)

ajakk:


games-emulation/openmsx/openmsx-16.0.ebuild
```diff
@@ -0,0 +1,106 @@
...
+inherit desktop flag-o-matic python-any-r1 readme.gentoo-r1
+
+DESCRIPTION="MSX emulator that aims for perfection"
+HOMEPAGE="http://openmsx.org/"
+SRC_URI="https://github.com/openMSX/openMSX/releases/download/RELEASE_${PV//./_}/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="test"
```
DarthGandalf, line 17: This is not necessary. You're not using this USE flag. And if you remove it, RESTRICT won't be needed either

games-emulation/openmsx/openmsx-16.0.ebuild
```diff
@@ -0,0 +1,106 @@
...
+	cp "${S}/Contrib/catch2/catch.hpp" "${S}/src/unittest" || die
+}
+
+src_compile() {
+	# Doesn't compile without this, and build system doesn't consistently add
+	# it properly for all compilers
+	append-cxxflags '-std=c++17'
+	emake \
+		CXXFLAGS="${CXXFLAGS}" \
+		INSTALL_SHARE_DIR="/usr/share/${PN}" \
```
DarthGandalf, line 65: `${EPREFIX}`?

chewi:
This is broken in several ways but I don't think I'll have time to go round and round in a review so I may just handle it myself.

games-emulation/openmsx/openmsx-16.0.ebuild
```diff
@@ -0,0 +1,111 @@
...
+src_configure() {
+	tc-export CXX
+
+	default
+}
+
+src_compile() {
+	# Doesn't compile without this, and build system doesn't consistently add
+	# it properly for all compilers
+	append-cxxflags '-std=c++17'
```
chewi, line 66: Now you have `src_configure`, I feel like this could go there instead.

games-emulation/openmsx/openmsx-16.0.ebuild
```diff
@@ -0,0 +1,111 @@
...
+
+	for i in 16 32 48 64 128 256 ; do
+		newicon -s "${i}" "share/icons/openMSX-logo-${i}.png" "${PN}.png"
+	done
+	make_desktop_entry "${PN}" "openMSX"
+}
+
+pkg_postinst() {
+	readme.gentoo_print_elog
+	xdg_icon_cache_update
```
chewi, line 110: I'd rather call `xdg_pkg_postinst` so that we don't have to worry about the details of what that eclass is doing, especially as it could change.

chewi (changes requested):
Much better! Just a couple of minor things now.

games-emulation/openmsx/openmsx-16.0.ebuild
```diff
@@ -0,0 +1,112 @@
...
+	media-libs/alsa-lib
+	media-libs/libogg
+	media-libs/libpng:0=
+	media-libs/libsdl2[sound,video]
+	media-libs/libtheora
+	media-libs/libvorbis
+	media-libs/sdl2-image[png]
+	media-libs/sdl2-ttf
+	>=media-libs/glew-1.3.2:0=
+	virtual/opengl"
```
chewi, line 29: Sorry, I was doing one more final check and noticed through `readelf -d` that the dependencies are outdated.

* libxml2 was dropped in 2014.
* SDL_image was dropped in 2010!!
* It should now depend on zlib.

chewi (changes requested):

chewi:
Merged now, thanks! I took the liberty of removing the subslot operator from zlib. The package does have a subslot but this was added in error and it makes Portage work a fair bit harder so you should just leave out the operator in this case.

PR #18675, opened 2020-12-16
dev-libs/ustr: Do not call ranlib directly (#719160)

Yamakuzure:
ranlib may or may not be available. It is therefore always better to use `ar s`, which is fully equivalent.

Closes: https://bugs.gentoo.org/719160


thesamesam:
For what it's worth, all the bug needed you to do was respect RANLIB via tc-getRANLIB, but this is fine too AFAIK. Thank you!

Yamakuzure:
> For what it's worth, all the bug needed you to do was respect RANLIB via tc-getRANLIB, but this is fine too AFAIK.

Well, the bug did not mention anything like that, so how was I supposed to know? Thanks for making me look stupid.

> No revbump is needed for build [failures](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html).

Right. My bad. But there is _NO_ build failure unless you have the corner case of the corner case where no `ranlib` is available. That's why I decided to revbump.

> Not much point in merging now IMO given we've last-rited it, but thank you! Let me know if you disagree though?

Exactly. There is no need for this PR any more. Thanks for reminding me, I appreciate that! :+1:

PR #18678, opened 2020-12-16
Add rust-lld use flag for dev-lang/rust

gamozolabs:
Rust is frequently using `rust-lld` in targets now, and that number is growing. It's currently in the dozens. https://github.com/rust-lang/rust/search?q=rust-lld+path%3Acompiler%2Frustc_target%2Fsrc%2Fspec&type=code

Due to this, many embedded targets will not build with the Rust from Gentoo. This PR adds a new USE flag for Rust called `rust-lld`, which will enable building of `rust-lld` for non-`system-llvm` use cases. There's already some overrides in place for handling `wasm`, which uses `rust-lld` for linking, and we do some regex to patch the Rust code to change the usage of `rust-ldd`. Unfortunately, as seen in the link above, `wasm` is not the only use of `rust-lld` by any means.

I highly recommend that we make this the default behavior (and probably just remove this option), but this is my first Gentoo repo push and I'm trying to make this as light of a change as possible. I suspect that over time, Rust will switch more and more targets over to using `rust-lld`. I'm unsure how this will be fixed to work with `system-llvm`.

Please note, this is my first Gentoo ebuild PR, and thus I don't really know what I'm doing. I removed the old `1.48.0`, made a new `1.48.0-r2` ebuild, added an IUSE entry, and changed the emission of the Rust configuration file to reflect the new use flag.

-B

### DIFF
--- a/dev-lang/rust/metadata.xml
+++ b/dev-lang/rust/metadata.xml
@@ -15,6 +15,7 @@
 		<flag name="nightly">Enable nightly (UNSTABLE) features</flag>
 		<flag name="parallel-compiler">Build a multi-threaded rustc</flag>
 		<flag name="rls">Install rls, Rust Language Server (used with IDEs supporting RLS protocol)</flag>
+		<flag name="rust-lld">Enable building of the rust-lld linker (not compatible with system-llvm)</flag>
 		<flag name="rustfmt">Install rustfmt, Rust code formatter</flag>
 		<flag name="system-bootstrap">Bootstrap using installed rust compiler</flag>
 		<flag name="system-llvm">Use the system LLVM install</flag>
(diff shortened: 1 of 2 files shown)


gamozolabs:
Hmm, just ran the build and it appears `rust-lld` is installed to `/usr/lib/rust/1.48.0/lib/rustlib/x86_64-unknown-linux-gnu/bin/rust-lld`.

I'm not sure how to set up a symlink as I'm not sure where the `x86_64-unknown-linux-gnu` could be derived during the ebuild. Seems like an annoying path.

telans:
You should add `rust-lld? ( !system-llvm )` into `REQUIRED_USE`

gyakovlev:
yeah it needs to manage symlink via eselect. current host abi can be referenced via `$(rust_abi)` call.
I think it would be better just to build it without extra useflag if not using system-llvm, so it can be untied from `wasm` useflag to use with all other targets.

hhoffstaette:
This will soon be obsolete as upstream Rust are working on making lld the default:
https://blog.rust-lang.org/2024/05/17/enabling-rust-lld-on-linux.html

PR #18689, opened 2020-12-17
media-libs/lilv: rewrite to use the python-single-r1 eclass

redneb:
Closes: https://bugs.gentoo.org/760402
Signed-off-by: Marios Titas <redneb@gmx.com>

### DIFF
--- a/media-libs/lilv/lilv-0.24.10-r1.ebuild
+++ b/media-libs/lilv/lilv-0.24.10-r1.ebuild
@@ -1,0 +1,81 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+PYTHON_COMPAT=( python3_{6,7,8,9} )
+PYTHON_REQ_USE='threads(+)'
+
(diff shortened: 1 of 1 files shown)


redneb:
The ebuild for `media-libs/lilv-0.24.10` installs python bindings for that library. Specifically, it installs files in `/usr/lib/pythonX.Y/site-packages`. But this is not reflected in its `USE` flags: there is no `PYTHON_TARGETS` or `PYTHON_SINGLE_TARGET`.

As a result, the package does not behave nicely with respect to python upgrades: When this package was first installed in my system, python 3.6 was the default, so it installed files for python 3.6. When my profile's `PYTHON_TARGETS` & `PYTHON_SINGLE_TARGET` were updated, that didn't cause a rebuild of the package, so it was stuck in python 3.6.

I think the solution is to update the ebuild so that it uses the machinery from the `python-single-r1`/`python-r1` eclasses.

So this PR contains a revbumped ebuild that uses `python-single-r1`. It could probably rewritten to use `python-r1`, but this is not as straightforward and using `python-single-r1` is already an improvement. As a side bonus, it also enables support for python-3.9 (I tested it against that version and it seems to work).

fordfrog:
@redneb thank you for this report and for the fix!

PR #18690, opened 2020-12-17
Haskell: miscellaneous package bumps and drops

solpeth:
NOT YET READY ~~Tested in an `~amd64` container. Commits pulled from my commits in `::haskell`.~~

Significant changes:

- Agda has been bumped up to 2.6.1.2
- `sci-mathematics/agda-lib-ffi` has been masked and the `ffi` USE flag removed from `agda-stdlib`. Bumping the `agda-stdlib` version to the latest available in the `agda-lib-ffi` ebuild, I could not get past `src_prepare()`. But I am also not familiar with the ebuild and the Agda ecosystem.

**Outstanding issues:**
- Emerging `sci-mathematics/agda` reports that it has fixed insecure RUNPATHs but that a bug should be filed against the package. I am unsure how to proceed on this front as regards actually fixing the cause of the insecure RUNPATH.


trofi:
Pulled. Thank you!

PR #18691, opened 2020-12-17
x11-misc/polybar: add 3.5.3 (fixes build related to documentation)

ajakk:
Adds USE=doc and adds sphinx and git as use dependencies to fix build failure. Git will be
dropped by 3.5.3

Closes: https://bugs.gentoo.org/760150


x11-misc/polybar/polybar-3.5.2.ebuild
```diff
@@ -31,6 +31,10 @@ DEPEND="
 	x11-libs/xcb-util-xrm
 	alsa? ( media-libs/alsa-lib )
 	curl? ( net-misc/curl )
+	doc? (
+		dev-python/sphinx
+		dev-vcs/git
```
thesamesam, line 36: We actually *need* git?

ajakk: Not anymore! Upstream patch added (only affects build time).

ajakk:
In case this is in anyone's queue for today: 3.5.3 is released, and that should be added rather than the fix currently in this PR. I will update this today.

PR #18692, opened 2020-12-17
net-fs/samba: enable support for glusterfs

crabbedhaloablution:
Stop automagic enabling of glusterfs, wrt. bug 703466.
glusterfs use flag controls building /usr/$(get_libdir)/samba/vfs/glusterfs.so,
not /usr/$(get_libdir)/samba/vfs/glusterfs_fuse.so, which is built
unconditionally.

Closes: https://bugs.gentoo.org/703466


ConiKost:
Seems in the meantime got already commited into tree, so closing.

PR #18693, opened 2020-12-17
net-irc/quassel: Remove 'ssl' IUSE, fix dependencies for live ebuild

Sput42:
A new release of Quassel IRC is imminent, so bring the live ebuild up to speed:

- SSL in Qt is now mandatory, so the USE flag can be removed
- dev-libs/boost was added as a new build dependency

### DIFF
--- a/net-irc/quassel/quassel-9999.ebuild
+++ b/net-irc/quassel/quassel-9999.ebuild
@@ -58,7 +58,7 @@ GUI_DEPEND="
 	urlpreview? ( dev-qt/qtwebengine:5[widgets] )
 "
 
-DEPEND="
+RDEPEND="
 	dev-qt/qtcore:5
 	dev-qt/qtnetwork:5[ssl]
 	sys-libs/zlib
@@ -71,7 +71,10 @@ DEPEND="
 		X? ( ${GUI_DEPEND} )
 	)
 "
-RDEPEND="${DEPEND}"
+
+DEPEND="${RDEPEND}
+	dev-libs/boost
+"
 BDEPEND="
 	dev-qt/linguist-tools:5
 	kde-frameworks/extra-cmake-modules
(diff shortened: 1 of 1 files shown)


net-irc/quassel/quassel-9999.ebuild
```diff
@@ -58,9 +58,9 @@ GUI_DEPEND="
 	urlpreview? ( dev-qt/qtwebengine:5[widgets] )
 "
 
-DEPEND="
+COMMON_DEPEND="
```
a17r, line 61: Introduction of `COMMON_DEPEND` is unnecessary. Just reverse `RDEPEND`/`DEPEND` inheritance.

a17r (changes requested):
Also, snorenotify was declared deprecated recently:

https://invent.kde.org/libraries/snorenotify/-/commit/8b64f8a1cdc8cb6abe3a59ffa52f108342433c63

Any plans to port away from it? Does it still do anything for us?

Sput42:
> Also, snorenotify was declared deprecated recently:
> 
> https://invent.kde.org/libraries/snorenotify/-/commit/8b64f8a1cdc8cb6abe3a59ffa52f108342433c63
> 
> Any plans to port away from it? Does it still do anything for us?

Snorenotify is an optional dependency as an alternative notification backend. If Quassel is the only thing still using it, we could probably just remove the USE flag and the package. If there are other users, I guess it doesn't hurt to keep support for it around?

net-irc/quassel/quassel-9999.ebuild
Sput42: OK, I was following the pattern I saw in other ebuilds with boost dependency, and assumed that that's the way to do this ;-) I'll adapt to your suggestion and push a new patchset in a bit.

a17r:
> Snorenotify is an optional dependency as an alternative notification backend. If Quassel is the only thing still using it, we could probably just remove the USE flag and the package. If there are other users, I guess it doesn't hurt to keep support for it around?

`net-irc/quassel` is indeed the only revdep of `x11-libs/snorenotify`. If maybe @johu still has any use for it with quassel, he could take over the package?

a17r (approved):

PR #18701, opened 2020-12-18
dev-python/argcomplete: Update importlib_metadata dependency

nertpinx:
This is based on upstream commit ecac7b02ffe6 which is already part of
`setup.py` in this version (v1.12.2) and is actually the main change from
v1.12.1.

Signed-off-by: Martin Kletzander <nert.pinx@gmail.com>

### DIFF
--- a/dev-python/argcomplete/argcomplete-1.12.2-r1.ebuild
+++ b/dev-python/argcomplete/argcomplete-1.12.2-r1.ebuild
@@ -19,7 +19,7 @@ RESTRICT="!test? ( test )"
 
 RDEPEND="
 	$(python_gen_cond_dep '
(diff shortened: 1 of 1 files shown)


mgorny:
Thank you. Since this changes RDEPEND in a significant way, could you bump ebuild revision too (`git mv` it to `-r1`)?

nertpinx:
@mgorny Sure thing, done now. I thought of copying it into an `-r1`, but did not realize I do not need to keep the older one around. Thanks.

thesamesam:
Oh, sorry. I actually just did this [independently](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=441a4255e603a658d8879356f8e7bd623388830f)! Thank you for the initiative.

nertpinx:
Cool, no problem, thank **you**!

PR #18703, opened 2020-12-18
net-dns/knot: bump to 2.9.8 and 3.0.3

nemunaire:
Hi,

2 minors versions have been released! Included in this PR.

Nothing change, this is just copy/paste :)


thesamesam:
2.9.8 says:
```
 *
 *   /var/lib/knot
 *
 *
```

PR #18704, opened 2020-12-18
net-misc/gns3-{gui,server}: 2.2.17 version bump, net-misc/ubridge: remove unused patch

mm1ke:
Hi,

This is a version bump for net-misc/gns3-{gui,server}.
However, gns3 is broken right now in gentoo because of bug https://bugs.gentoo.org/755899
I've addressed this bug in PR: https://github.com/gentoo/gentoo/pull/18598 which is why this PR should only be pushed after fixing the given bug.


net-misc/gns3-server/gns3-server-2.2.17.ebuild
```diff
@@ -0,0 +1,65 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+PYTHON_COMPAT=( python3_{7,8} )
```
juippis, line 5: One empty line below EAPI please.

net-misc/gns3-server/gns3-server-2.2.17.ebuild
```diff
@@ -0,0 +1,65 @@
...
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="
+	>=app-emulation/dynamips-0.2.18
+	~dev-python/aiofiles-0.5.0[${PYTHON_USEDEP}]
+	~dev-python/aiohttp-3.6.2[${PYTHON_USEDEP}]
+	~dev-python/aiohttp-cors-0.7.0[${PYTHON_USEDEP}]
+	~dev-python/async_timeout-3.0.1[${PYTHON_USEDEP}]
```
juippis, line 22: These fixed dependencies are frowned upon, but not much you can do about I guess. They probably work (see psutil example below) but would require much patching then.

net-misc/gns3-server/gns3-server-2.2.17.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	default
+
+	# newer psutils is fine
+	sed -i -e '/psutil==5.6.7/d' requirements.txt || die "fixing requirements failed"
+
+	# Remove Pre-built busybox binary
+	rm gns3server/compute/docker/resources/bin/busybox || die
+
+	# Package installs 'tests' package which is forbidden
+	rm -rf tests || die
```
juippis, line 43: Drop the `-f`.

net-misc/gns3-server/gns3-server-2.2.17.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	eapply_user
```
juippis, line 44: You already call this via `default`, this'd make user patching fail as it tries to apply the same patch twice.

net-misc/gns3-server/gns3-server-2.2.17.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	mkdir -p "${D}$(python_get_sitedir)/gns3server/compute/docker/resources/bin" || die
+	ln -s /bin/busybox "${D}$(python_get_sitedir)/gns3server/compute/docker/resources/bin/busybox" || die
+}
+
+pkg_postinst() {
+	elog "net-misc/gns3-server has several optional packages that must be merged manually for additional functionality."
+	elog ""
+	elog "The following is a list of packages that can be added:"
+	elog "app-emulation/qemu, app-emulation/virtualbox"
+	elog "app-emulation/docker and net-analyzer/wireshark"
```
juippis, line 61: We prefer to use `optfeature` with these since it only prints deps that are not installed.

net-misc/gns3-gui/gns3-gui-2.2.17.ebuild
```diff
@@ -0,0 +1,56 @@
...
+RDEPEND="
+	>=dev-python/distro-1.3.0[${PYTHON_USEDEP}]
+	~dev-python/jsonschema-3.2.0[${PYTHON_USEDEP}]
+	>=dev-python/psutil-5.7.0[${PYTHON_USEDEP}]
+	>=dev-python/sentry-sdk-0.14.4[${PYTHON_USEDEP}]
+	~net-misc/gns3-server-${PV}[${PYTHON_USEDEP}]
+	dev-python/PyQt5[gui,network,svg,websockets,widgets,${PYTHON_USEDEP}]
+"
+
+DISTUTILS_USE_SETUPTOOLS=bdepend
```
juippis, line 28: This needs to be called before inheriting *distutils-r1.eclass*.

net-misc/gns3-gui/gns3-gui-2.2.17.ebuild
```diff
@@ -0,0 +1,56 @@
...
+
+src_prepare() {
+	default
+
+	# newer psutils is fine
+	sed -i -e '/psutil==5.6.7/d' requirements.txt || die "fixing requirements failed"
+
+	eapply_user
```
juippis, line 36: Same here, you already call this via `default`.

net-misc/gns3-gui/gns3-gui-2.2.17.ebuild
```diff
@@ -0,0 +1,56 @@
...
+	xdg_desktop_database_update
+	xdg_mimeinfo_database_update
+	xdg_icon_cache_update
+}
+
+pkg_postrm() {
+	xdg_desktop_database_update
+	xdg_mimeinfo_database_update
+	xdg_icon_cache_update
+}
```
juippis, line 56: Since you inherit *xdg*, the default behaviour should call all these. So you should be able to just remove these lines from the ebuild.

mm1ke:
@juippis 
I've addressed your issues and also changed the dependencies and removed the pinned versions in the `requirements.txt`. (as also suggested in https://github.com/gentoo/gentoo/pull/18598)
There is still a issue with `dev-python/aiohttp` which i've addressed in the PR mentioned.
However, with the change in the PR mentioned, and the latest version of `aiohttp`, gns3 seems to work fine. This should also close bug: https://bugs.gentoo.org/755899

juippis (approved):

PR #18713, opened 2020-12-18
binutils-apple: fix prefix bootstrap stage3

cognifloyd:
In the version of private header (dyld_priv.h) copied from DYLD sources,
It uses TARGET_OS_BRIDGE which is not defined in the latest MacOSX.sdk.
We don't care about BridgeOS, so we just disable that.

This only hits during stage3 once we've re-compiled our own clang so
that clang starts injecting -Werror,-Wundef-prefix=TARGET_OS_ (Apple
must have that disabled in their version of clang).

Also uses newer patches so that dsymutil can be found even though it is
not named llvm-dsymutil, and it also bumps the max llvm slots searched
from 10 to 12.

Bug: https://bugs.gentoo.org/758167
Bug: https://bugs.gentoo.org/631862

Signed-off-by: Jacob Floyd <cognifloyd@gmail.com>

### DIFF
--- a/sys-devel/binutils-apple/binutils-apple-11.3.1.ebuild
+++ b/sys-devel/binutils-apple/binutils-apple-11.3.1.ebuild
@@ -79,6 +79,10 @@ src_prepare() {
 	mkdir -p include/mach-o || die
 	# never present because it's private
 	cp ../../${DYLD}/include/mach-o/dyld_priv.h include/mach-o || die
+	# TARGET_OS_BRIDGE is undefined in TargetConditionals.h of newer MacOSX.sdk.
+	# We don't target BridgeOS. Disable it to avoid errors when clang adds:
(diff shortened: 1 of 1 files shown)


cognifloyd:
This will be broken until the patch tarball version gets bumped. Right now it can't find the distfile because it doesn't exist yet.

grobian:
I've applied the patch, what about the patch tarball?

grobian:
oh, you mean snapshot for stage3 bootstrap, you can use LATEST_TREE from tomorrow, I'll look into bumping it preferably when we got some more fixes in there to make it complete.

cognifloyd:
Cool. I will file a new PR to highlight what I the llvm-shim fix I added.

cognifloyd:
K. Check out https://github.com/gentoo/gentoo/pull/18728#pullrequestreview-555963887 for more on what patches I was referring to.

Thanks for working on this!

PR #18715, opened 2020-12-18
media-libs/libheif: Bump to 1.10.0

jsmolic:


media-libs/libheif/libheif-1.10.0.ebuild
```diff
@@ -0,0 +1,67 @@
...
+RESTRICT="!test? ( test )"
+
+BDEPEND="test? ( dev-lang/go )"
+DEPEND="
+	media-libs/dav1d:=[${MULTILIB_USEDEP}]
+	>=media-libs/libaom-2.0.0:=[${MULTILIB_USEDEP}]
+	media-libs/libde265:=[${MULTILIB_USEDEP}]
+	media-libs/libpng:0=[${MULTILIB_USEDEP}]
+	media-libs/x265:=[${MULTILIB_USEDEP}]
+	media-video/rav1e:=
```
Whissi, line 33: I am not sure about this, see https://github.com/strukturag/libheif/blob/v1.10.0/CMakeLists.txt#L132 -- libheif will only need an AVIF encoder. But it can be libaom or rav1e.

Also dav1d and libaom are interchangeable.

libde265 is also optional like x265.

Whissi (changes requested):

media-libs/libheif/libheif-1.10.0.ebuild
jsmolic: Yes, you're right. The thing is, build system doesn't have configure switches to choose between all encoders/decoders. I've added some USE flags to chose between different libraries, and adjusted the mask for `rav1e` support

media-libs/libheif/libheif-1.10.0.ebuild
```diff
@@ -0,0 +1,71 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools xdg multilib-minimal
+
+if [[ ${PV} == "9999" ]] ; then
```
SoapGentoo, line 8: `if [[ ${PV} == *9999 ]] ; then`

media-libs/libheif/libheif-1.10.0.ebuild
```diff
@@ -0,0 +1,71 @@
...
+	eautoreconf
+
+	# prevent "stat heif-test.go: no such file or directory"
+	multilib_copy_sources
+}
+
+multilib_src_configure() {
+	local econf_args=(
+		--disable-static
+		$(multilib_is_native_abi && use go || echo --disable-go)
```
SoapGentoo, line 53: tried `$(multilib_native_use_enable go)`?

media-libs/libheif/libheif-1.10.0.ebuild
```diff
@@ -0,0 +1,71 @@
...
+	ECONF_SOURCE="${S}" econf "${econf_args[@]}"
+}
+
+multilib_src_test() {
+	default
+	emake -C go test
+}
+
+multilib_src_install_all() {
+	find "${ED}" -name '*.la' -delete || die
```
SoapGentoo, line 70: `einstalldocs`?

SoapGentoo (changes requested):

media-libs/libheif/libheif-1.10.0.ebuild
jsmolic: Done

media-libs/libheif/libheif-1.10.0.ebuild
jsmolic: Done

media-libs/libheif/libheif-1.10.0.ebuild
jsmolic: Yeah but it doesn't really work, I haven't managed to run tests if we pass `--enable go` so I went with this

PR #18720, opened 2020-12-19
Fix Darwin prefix bootstrap for llvm-11+ projects (sys-libs/libcxx*, sys-devel/clang, sys-libs/llvm-libunwind, sys-libs/compiler-rt*)

cognifloyd:
This adjusts these llvm-11+ packages to support darwin prefix:
- sys-libs/libcxx
- sys-libs/libcxxabi
- sys-devel/clang
- sys-libs/llvm-libunwind (just keyword for `~x64-macos`)
- sys-libs/compiler-rt
- sys-libs/compiler-rt-sanitizers

### For sys-libs/libcxx*:

Add bootstrap useflag to libcxx and libcxxabi to allow the prefix
bootstrap phase to skip llvm_pkg_setup which checks for an installed
llvm. The bootstrap script takes care of setting sufficient variables so
that the correct llvm will be found on the PATH. Plus, during bootstrap
llvm might not be installed yet (depending on the stage), so allow
skipping that.

Also, eprefrixify some dylib paths in CMakeLists.txt (dylib is only used
on darwin, so this is not an issue when compiling on any other platform.

Plus, re-add a couple of darwin specific fixes to set things up
correctly on darwin, and then keyword ~x64-macos

### For sys-devel/clang:

Adjust the clang internal header/framework search paths when building on
darwin prefix. We symlink the selected MacOSX.sdk to EPREFIX/MacOSX.sdk
during bootstrap, so that is the correct place to get system headers,
such as those for libc, or for system framework headers, that we do not
replace in prefix.

### For sys-libs/compiler-rt*:

Gentoo Prefix does not follow standard Apple practice of using --sysroot
or --isysroot on everything because we have to account for two "root"s.
1) EPREFIX is "root"
2) EPREFIX/MacOSX.sdk is also sysroot as it provides system headers.
    
So, adjust sys-libs/compiler-rt and sys-libs/compiler-rt-sanitizers to
prevent them from adding the sysroot flags on Darwin.
    
The ebuilds also adjust some CMAKE args to ensure it is using
EPREFIX/MacOSX.sdk instead of looking things up with xcrun.

### DIFF
--- a/sys-devel/clang/clang-11.0.0.ebuild
+++ b/sys-devel/clang/clang-11.0.0.ebuild
@@ -5,7 +5,7 @@ EAPI=7
 
 PYTHON_COMPAT=( python3_{6..9} )
 inherit cmake llvm llvm.org multilib-minimal pax-utils \
-	python-single-r1 toolchain-funcs
+	prefix python-single-r1 toolchain-funcs
 
 DESCRIPTION="C language family frontend for LLVM"
 HOMEPAGE="https://llvm.org/"
@@ -30,7 +30,7 @@ ALL_LLVM_TARGETS=( "${ALL_LLVM_TARGETS[@]/#/llvm_targets_}" )
 
 LICENSE="Apache-2.0-with-LLVM-exceptions UoI-NCSA MIT"
 SLOT="$(ver_cut 1)"
-KEYWORDS="amd64 arm arm64 ppc64 ~riscv x86 ~amd64-linux"
+KEYWORDS="amd64 arm arm64 ppc64 ~riscv x86 ~amd64-linux ~x64-macos"
 IUSE="debug default-compiler-rt default-libcxx default-lld
 	doc +static-analyzer test xml kernel_FreeBSD ${ALL_LLVM_TARGETS[*]}"
 REQUIRED_USE="${PYTHON_REQUIRED_USE}
@@ -75,6 +75,10 @@ PDEPEND="
 # Therefore: use sys-devel/clang[${MULTILIB_USEDEP}] only if you need
 # multilib clang* libraries (not runtime, not wrappers).
 
+PATCHES=(
+	"${FILESDIR}"/9999/prefix-dirs.patch
+)
+
 pkg_setup() {
 	LLVM_MAX_SLOT=${SLOT} llvm_pkg_setup
 	python-single-r1_pkg_setup
@@ -88,6 +92,11 @@ src_prepare() {
 	llvm.org_src_prepare
 
 	mv ../clang-tools-extra tools/extra || die
+
+	# add Gentoo Portage Prefix for Darwin (see prefix-dirs.patch)
+	eprefixify \
+		lib/Frontend/InitHeaderSearch.cpp \
+		lib/Driver/ToolChains/Darwin.cpp || die
 }
 
 check_distribution_components() {
--- a/sys-libs/compiler-rt-sanitizers/compiler-rt-sanitizers-11.0.0.ebuild
+++ b/sys-libs/compiler-rt-sanitizers/compiler-rt-sanitizers-11.0.0.ebuild
@@ -35,6 +35,10 @@ BDEPEND="
 		sys-libs/compiler-rt:${SLOT} )
 	${PYTHON_DEPS}"
 
+PATCHES=(
+	"${FILESDIR}/9999/compiler-rt-prefix-paths.patch"
+)
+
 python_check_deps() {
 	use test || return 0
 	has_version "dev-python/lit[${PYTHON_USEDEP}]"
@@ -110,8 +114,15 @@ src_configure() {
 
 	if use prefix && [[ "${CHOST}" == *-darwin* ]] ; then
 		mycmakeargs+=(
-			# disable use of SDK for the system itself
-			-DDARWIN_macosx_CACHED_SYSROOT=/
+			# setting -isysroot is disabled with compiler-rt-prefix-paths.patch
+			# this allows adding arm64 support using SDK in EPREFIX
+			-DDARWIN_macosx_CACHED_SYSROOT="${EPREFIX}/MacOSX.sdk"
+			# Set version based on the SDK in EPREFIX
+			# This disables i386 for SDK >= 10.15
+			# Will error while building tsan if SDK < 10.12
+			-DDARWIN_macosx_OVERRIDE_SDK_VERSION="$(realpath ${EPREFIX}/MacOSX.sdk | sed -e 's/.*MacOSX\(.*\)\.sdk/\1/')"
+			# Use our libtool instead of looking it up with xcrun
+			-DCMAKE_LIBTOOL="${EPREFIX}/usr/bin/${CHOST}-libtool"
 		)
 	fi
 
--- a/sys-libs/compiler-rt/compiler-rt-11.0.0.ebuild
+++ b/sys-libs/compiler-rt/compiler-rt-11.0.0.ebuild
@@ -29,6 +29,10 @@ BDEPEND="
 		=sys-devel/clang-${PV%_*}*:${CLANG_SLOT} )
 	${PYTHON_DEPS}"
 
+PATCHES=(
+	"${FILESDIR}/9999/${PN}-prefix-paths.patch"
+)
+
 python_check_deps() {
 	use test || return 0
 	has_version "dev-python/lit[${PYTHON_USEDEP}]"
@@ -83,8 +87,14 @@ src_configure() {
 
 	if use prefix && [[ "${CHOST}" == *-darwin* ]] ; then
 		mycmakeargs+=(
-			# disable use of SDK for the system itself
-			-DDARWIN_macosx_CACHED_SYSROOT=/
+			# setting -isysroot is disabled with compiler-rt-prefix-paths.patch
+			# this allows adding arm64 support using SDK in EPREFIX
+			-DDARWIN_macosx_CACHED_SYSROOT="${EPREFIX}/MacOSX.sdk"
+			# Set version based on the SDK in EPREFIX.
+			# This disables i386 for SDK >= 10.15
+			-DDARWIN_macosx_OVERRIDE_SDK_VERSION="$(realpath ${EPREFIX}/MacOSX.sdk | sed -e 's/.*MacOSX\(.*\)\.sdk/\1/')"
+			# Use our libtool instead of looking it up with xcrun
+			-DCMAKE_LIBTOOL="${EPREFIX}/usr/bin/${CHOST}-libtool"
 		)
 	fi
 
--- a/sys-libs/libcxx/libcxx-11.0.0.ebuild
+++ b/sys-libs/libcxx/libcxx-11.0.0.ebuild
@@ -14,7 +14,7 @@ llvm.org_set_globals
 
 LICENSE="Apache-2.0-with-LLVM-exceptions || ( UoI-NCSA MIT )"
 SLOT="0"
-KEYWORDS="amd64 arm arm64 ~riscv x86"
+KEYWORDS="amd64 arm arm64 ~riscv x86 ~x64-macos"
 IUSE="elibc_glibc elibc_musl +libcxxabi +libunwind +static-libs test"
 REQUIRED_USE="libunwind? ( libcxxabi )"
 RESTRICT="!test? ( test )"
@@ -41,7 +41,11 @@ python_check_deps() {
 }
 
 pkg_setup() {
-	llvm_pkg_setup
+	# darwin prefix builds do not have llvm installed yet, so rely on bootstrap-prefix
+	# to set the appropriate path vars to LLVM instead of using llvm_pkg_setup.
+	if [[ ${CHOST} != *-darwin* ]] || has_version dev-lang/llvm; then
+		llvm_pkg_setup
+	fi
 	use test && python-any-r1_pkg_setup
 
 	if ! use libcxxabi && ! tc-is-gcc ; then
@@ -58,6 +62,10 @@ src_prepare() {
 	eapply "${FILESDIR}/${PN}-3.9-cmake-link-flags.patch"
 
 	llvm.org_src_prepare
+
+	# eprefixify static path references to libc++abi for symbol re-export to
+	# avoid linking against it twice in both /usr/lib and ${EPREFIX}/usr/lib
+	sed -i -e "s^/usr/lib/libc++abi.dylib^${EPREFIX}&^g" lib/CMakeLists.txt || die
 }
 
 test_compiler() {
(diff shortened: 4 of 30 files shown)


sys-devel/clang/clang-11.0.0.ebuild
```diff
@@ -88,6 +92,11 @@ src_prepare() {
 	llvm.org_src_prepare
 
 	mv ../clang-tools-extra tools/extra || die
+
+	# add Gentoo Portage Prefix for Darwin (see prefix-dirs.patch)
+	sed -i -e "s|@GENTOO_PORTAGE_EPREFIX@|${EPREFIX}/|g" \
+		lib/Frontend/InitHeaderSearch.cpp \
+		lib/Driver/ToolChains/Darwin.cpp
```
mgorny, line 99: `||die`

sys-devel/clang/files/11.0.0/prefix-dirs.patch
```diff
@@ -0,0 +1,76 @@
+This mirrors cmake-*-prefix-dirs.patch
```
mgorny, line 1: Patches that need to be kept indefinitely go to 9999.

sys-libs/compiler-rt-sanitizers/compiler-rt-sanitizers-11.0.0.ebuild
```diff
@@ -35,6 +35,10 @@ BDEPEND="
 		sys-libs/compiler-rt:${SLOT} )
 	${PYTHON_DEPS}"
 
+PATCHES=(
+	"${FILESDIR}/compiler-rt-prefix-paths.patch"
```
mgorny, line 39: Please follow suit with other LLVM projects and use versioned subdirectories.

sys-libs/libcxx/libcxx-11.0.0.ebuild
```diff
@@ -14,9 +14,10 @@ llvm.org_set_globals
 
 LICENSE="Apache-2.0-with-LLVM-exceptions || ( UoI-NCSA MIT )"
 SLOT="0"
-KEYWORDS="amd64 arm arm64 ~riscv x86"
-IUSE="elibc_glibc elibc_musl +libcxxabi +libunwind +static-libs test"
-REQUIRED_USE="libunwind? ( libcxxabi )"
+KEYWORDS="amd64 arm arm64 ~riscv x86 ~x64-macos"
+IUSE="elibc_Darwin elibc_glibc elibc_musl +libcxxabi +libunwind +static-libs test bootstrap"
```
mgorny, line 18: Keep sorted.

sys-libs/libcxx/libcxx-11.0.0.ebuild
```diff
@@ -41,7 +42,7 @@ python_check_deps() {
 }
 
 pkg_setup() {
-	llvm_pkg_setup
+	bootstrap || llvm_pkg_setup
```
mgorny, line 45: This is not a valid command.

cognifloyd:
Thanks for the review @mgorny  :) I just resolved your comments and rebased on master.

cognifloyd:
rebased on master to account for the _rc2 bumps

sys-libs/libcxx/libcxx-11.0.0.ebuild
```diff
@@ -58,6 +59,11 @@ src_prepare() {
 	eapply "${FILESDIR}/${PN}-3.9-cmake-link-flags.patch"
 
 	llvm.org_src_prepare
+
+	# eprefixify static path references to libc++abi for symbol re-export to
+	# avoid linking against it twice in both /usr/lib and ${EPREFIX}/usr/lib
+	local lcpa=/usr/lib/libc++abi.dylib
+	sed -i -e "s,${lcpa},${EPREFIX}${lcpa},g" lib/CMakeLists.txt
```
mgorny, line 66:
```suggestion
	sed -i -e "s^/usr/lib/libc++abi.dylib^${EPREFIX}&^g" lib/CMakeLists.txt || die
```

Please really add `|| die` to all sed calls. Also `,` seems remotely possible in EPREFIX.

sys-libs/libcxx/libcxx-11.0.0.ebuild
```diff
@@ -100,6 +106,8 @@ multilib_src_configure() {
 			fi
 		fi
 	fi
+	# Darwin has libunwind in its system
+	[[ ${CHOST} == *-darwin* ]] && want_gcc_s=OFF
```
mgorny, line 110: I am somewhat doubtful about overriding only one of the three variables that are set above. I suppose the above logic should yield reproducible results on Darwin, correct?

sys-libs/compiler-rt-sanitizers/compiler-rt-sanitizers-11.0.0.ebuild
```diff
@@ -35,6 +35,10 @@ BDEPEND="
 		sys-libs/compiler-rt:${SLOT} )
```
mgorny, line 35: This one needs splitting too.

mgorny (changes requested):

sys-libs/libcxx/libcxx-11.0.0.ebuild
```diff
@@ -14,9 +14,10 @@ llvm.org_set_globals
 
 LICENSE="Apache-2.0-with-LLVM-exceptions || ( UoI-NCSA MIT )"
 SLOT="0"
-KEYWORDS="amd64 arm arm64 ~riscv x86"
-IUSE="elibc_glibc elibc_musl +libcxxabi +libunwind +static-libs test"
-REQUIRED_USE="libunwind? ( libcxxabi )"
+KEYWORDS="amd64 arm arm64 ~riscv x86 ~x64-macos"
+IUSE="bootstrap elibc_Darwin elibc_glibc elibc_musl +libcxxabi +libunwind +static-libs test"
+REQUIRED_USE="elibc_Darwin? ( libcxxabi )
```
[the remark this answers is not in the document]
cognifloyd: done

sys-libs/libcxx/libcxx-11.0.0.ebuild
```diff
@@ -14,9 +14,10 @@ llvm.org_set_globals
```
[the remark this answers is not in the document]
cognifloyd: done

sys-libs/libcxx/libcxx-11.0.0.ebuild
cognifloyd: done

sys-libs/compiler-rt-sanitizers/compiler-rt-sanitizers-11.0.0.ebuild
cognifloyd: done

sys-libs/libcxx/libcxx-11.0.0.ebuild
```diff
@@ -41,7 +42,7 @@ python_check_deps() {
 }
 
 pkg_setup() {
-	llvm_pkg_setup
+	use bootstrap || llvm_pkg_setup
```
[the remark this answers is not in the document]
cognifloyd: Yes, using the `bootstrap` useflag is an established pattern, as `bootstrap-prefix.sh` unconditionally sets it during stage2 and stage3. After stage3, however, all the packages get rebuilt during `emerge -e system`, so this does get "undone".

I clarified in the commit messages why I'm using the `bootstrap` useflag.

If you would prefer, I can switch the bootstrap useflag for checking for darwin CHOST + `! has_version sys-devel/llvm`.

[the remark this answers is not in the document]
cognifloyd: K. I added the package.use.mask entries.

Would you prefer the darwin-conditional guesswork?

sys-libs/libcxx/libcxx-11.0.0.ebuild
cognifloyd: Yes. I am concerned about that as well. I'm reintroducing some things from older prefix ebuilds.
This might not be needed. I'll revisit this to see.

sys-libs/libcxx/libcxx-11.0.0.ebuild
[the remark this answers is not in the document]
cognifloyd: I switched to the darwin-conditional guesswork, dropping the bootstrap useflag. Is this more like what you were hoping for?

sys-libs/libcxx/libcxx-11.0.0.ebuild
cognifloyd: I reworked this so that it is hopefully clearer.

cognifloyd:
@mgorny, I split up the commits, dropped the bootstrap useflag, and adjusted according to your other comments. How does this look?

mgorny (approved):
Thank you, this looks much cleaner.

@gentoo/prefix, any comments? If you're okay with it, feel free to merge straight.

sys-devel/clang/clang-11.0.0.ebuild
```diff
@@ -88,6 +92,11 @@ src_prepare() {
 	llvm.org_src_prepare
 
 	mv ../clang-tools-extra tools/extra || die
+
+	# add Gentoo Portage Prefix for Darwin (see prefix-dirs.patch)
+	sed -i -e "s|@GENTOO_PORTAGE_EPREFIX@|${EPREFIX}/|g" \
```
grobian, line 97: please use eprefixify from prefix.eclass to do this:

eprefixify lib/Frontend/InitHeaderSearch.cpp lib/Driver/ToolChains/Darwin.cpp || die

sys-libs/compiler-rt/compiler-rt-11.0.0.ebuild
```diff
@@ -83,8 +87,13 @@ src_configure() {
 
 	if use prefix && [[ "${CHOST}" == *-darwin* ]] ; then
 		mycmakeargs+=(
-			# disable use of SDK for the system itself
-			-DDARWIN_macosx_CACHED_SYSROOT=/
+			# setting -isysroot is disabled with compiler-rt-prefix-paths.patch
+			# this allows adding arm64 support using SDK in EPREFIX
+			-DDARWIN_macosx_CACHED_SYSROOT="${EPREFIX}/MacOSX.sdk"
+			# Set this based on the SDK in EPREFIX
+			-DDARWIN_macosx_OVERRIDE_SDK_VERSION="$(realpath ${EPREFIX}/MacOSX.sdk | sed -e 's/.*MacOSX\(.*\)\.sdk/\1/')"
```
grobian, line 94: Is it necessary to hardcode the symlink target here?  This way, we cannot switch the symlink (e.g. using eselect, or manually after Xcode update).  Is this used only during build, or also from the library itself during runtime?

sys-libs/compiler-rt-sanitizers/compiler-rt-sanitizers-11.0.0.ebuild
```diff
@@ -110,8 +114,13 @@ src_configure() {
 
 	if use prefix && [[ "${CHOST}" == *-darwin* ]] ; then
 		mycmakeargs+=(
-			# disable use of SDK for the system itself
-			-DDARWIN_macosx_CACHED_SYSROOT=/
+			# setting -isysroot is disabled with compiler-rt-prefix-paths.patch
+			# this allows adding arm64 support using SDK in EPREFIX
+			-DDARWIN_macosx_CACHED_SYSROOT="${EPREFIX}/MacOSX.sdk"
+			# Set this based on the SDK in EPREFIX
+			-DDARWIN_macosx_OVERRIDE_SDK_VERSION="$(realpath ${EPREFIX}/MacOSX.sdk | sed -e 's/.*MacOSX\(.*\)\.sdk/\1/')"
```
grobian, line 121: ditto

sys-libs/compiler-rt/compiler-rt-11.0.0.ebuild
cognifloyd: The SDK version number is not baked into the build.
It does however disable i386 for SDK >= 10.15, and issue a fatal error building the TSan lib with SDK < 10.12.

in `compiler-rt/cmake/Modules/CompilerRTDarwinUtils.cmake`:
- function `find_darwin_sdk_version` is the only thing that uses this var
- in `darwin_test_archs`, it has this comment where the `find_darwin_sdk_version` function is used:
```
# Disable building for i386 for macOS SDK >= 10.15. The SDK doesn't support
# linking for i386 and the corresponding OS doesn't allow running macOS i386
# binaries.
```

in `lib/tsan/CMakeLists.txt`, it has this comment where `find_darwin_sdk_version` is used:
```
# Ideally we would check the SDK version for the actual platform we are
# building for here.  To make our lifes easier we assume the host SDK setup is
# sane and use the macOS SDK version as a proxy for aligned SDKs.
```
and then it throws this FATAL ERROR message for `VERSION_LESS 10.12`
```
message(FATAL_ERROR "Building the TSan runtime requires at least macOS SDK 10.12 (or aligned SDK on other platforms)")
```

cognifloyd: Also, I want to clarify the realpath usage: Here we get the realpath only to extract the version number from it with sed.
```
realpath ${EPREFIX}/MacOSX.sdk | sed -e 's/.*MacOSX\(.*\)\.sdk/\1/'
```

I considered reading `MacOSX.sdk/SDKSettings.json` but then I'd have to have a bdep on `jq` or use string magic and hope it doesn't break or something.

sys-libs/compiler-rt-sanitizers/compiler-rt-sanitizers-11.0.0.ebuild
cognifloyd: The same sources are used here, so ditto for my comments :)

sys-libs/compiler-rt/compiler-rt-11.0.0.ebuild
cognifloyd: I updated the comment to clarify what this is used for in the build.

sys-libs/compiler-rt-sanitizers/compiler-rt-sanitizers-11.0.0.ebuild
cognifloyd: I updated the comment to clarify what this is used for in the build.

sys-devel/clang/clang-11.0.0.ebuild
cognifloyd: done

cognifloyd:
@grobian I think I've addressed your comments.

grobian:
merged, thanks!

PR #18722, opened 2020-12-19
dev-embedded/openocd: Fix compilation with gcc 10

kripton:
Closes: https://bugs.gentoo.org/722640

Patch taken from upstream: http://openocd.zylin.com/gitweb?p=openocd.git;a=commitdiff;h=c60252ac2b636c4d99b766a574b9df0966151696

Ebuild copied from 0.10.0-r1, updated copyright header and added the patch. No further changes.


thesamesam:
I've dropped the revbump as it's not [needed](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/) for build failures. Thank you!

PR #18724, opened 2020-12-19
Dev-python/spyder: version bump 4.2.1

AndrewAmmerlaan:
Bumps spyder and adds some new dependencies


dev-python/spyder/spyder-4.2.1.ebuild
```diff
@@ -0,0 +1,230 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8} )
+
+inherit optfeature xdg distutils-r1
```
juippis, line 8:
```
 * DISTUTILS_USE_SETUPTOOLS value is probably incorrect
 *   have:     DISTUTILS_USE_SETUPTOOLS=bdepend (or unset)
 *   expected: DISTUTILS_USE_SETUPTOOLS=no
```
Could you verify?

juippis (commented):
Well tests for spyder just hang for me after running for 2 hours. :I 

Still spyder seems to run fine.

dev-python/oset/oset-0.1.3.ebuild
```diff
@@ -0,0 +1,16 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8,9} )
+
+inherit distutils-r1
```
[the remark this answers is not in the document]
AndrewAmmerlaan: Oops I missed this one, thanks.

dev-python/oset/oset-0.1.3.ebuild
```diff
@@ -0,0 +1,16 @@
...
+
+DESCRIPTION="Ordered Set"
+HOMEPAGE="https://pypi.org/project/oset/"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="OSET"
```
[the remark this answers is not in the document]
AndrewAmmerlaan: Ooh right, I was confused about the different copyright header. And that the license file in ::gentoo has:
```
Copyright (c) <YEAR>, <OWNER>
All rights reserved.
```
which to me seems strange considering that all these packages have different `<OWNER>`s. But I am not a lawyer, so if `LICENSE="BSD"` is fine in these cases I'll just do that.

dev-python/spyder/spyder-4.2.1.ebuild
AndrewAmmerlaan: This warning has been confusing me for quite some time now, it is clearly a false positive since `setup.py` has:
```
# Third party imports
from setuptools import setup, find_packages
from setuptools.command.install import install
```
but I do not understand why the eclass's QA check does not work correctly here.

AndrewAmmerlaan:
> Well tests for spyder just hang for me after running for 2 hours. :I
> 
> Still spyder seems to run fine.

Yeah the test suite does that, for me it wasn't 2 hours but it still took quite long for some tests to pass (easily 30mins+). I suspect the reason that the test suite is being so difficult is because the version of PyQt5 in ::gentoo is more recent than the version explicitly specified by spyder upstream as a dependency. I kinda feel like RESTRICTing the test phase, because I have been spending several hours every version bump trying to find out which tests fail/hang in that version, and I don't think it is worth the time I spend on it because even when I do get it working it is still a very slow and picky test suite, and I will just have to do it all over again in the next version bump.

I fixed oset's `LICENSE` entry.

dev-python/spyder/spyder-4.2.1.ebuild
```diff
@@ -0,0 +1,241 @@
...
+
+# The test suite often hangs or does not work.
+# Technically spyder requires pyqt5<13, which
+# we do not have in ::gentoo any more. Likely
+# this is the reason many of the tests fail
+# or hang. RESTRICTing because IMO it is
+# not worth the several hours I spend every
+# single version bump checking which tests
+# do and do not work. Spyder itself works
+# fine with pyqt5>13.
```
juippis, line 34: I'd rather keep this short and technical.

dev-python/spyder/spyder-4.2.1.ebuild
juippis: https://blogs.gentoo.org/mgorny/2020/10/21/distutils_use_setuptools-qa-spam-and-more-qa-spam/ Yeah, might be due to python-3.8 changes regarding setuptools. I wouldn't know any better.

juippis (approved):
Yep, agree there. It's a bit of a shame since you've worked really hard adding tens of deps to make the tests work too :P

AndrewAmmerlaan:
Thanks @juippis :smile:

PR #18728, opened 2020-12-19
sys-devel/binutils-apple: fix llvm-shim

cognifloyd:
This uses newer patches so that dsymutil can be found even though it is
not named llvm-dsymutil, and it also bumps the max llvm slots searched
from 10 to 12.

Bug: https://bugs.gentoo.org/631862
Signed-off-by: Jacob Floyd <cognifloyd@gmail.com>


sys-devel/binutils-apple/binutils-apple-11.3.1.ebuild
```diff
@@ -30,7 +30,7 @@ SRC_URI="http://www.opensource.apple.com/tarballs/ld64/${LD64}.tar.gz
 	https://dev.gentoo.org/~grobian/distfiles/${PN}-patches-5.1-r2.tar.bz2
 	https://dev.gentoo.org/~grobian/distfiles/${PN}-patches-7.3-r2.tar.bz2
 	https://dev.gentoo.org/~grobian/distfiles/${PN}-patches-8.2-r1.tar.bz2
-	https://dev.gentoo.org/~grobian/distfiles/${PN}-patches-11.3.tar.bz2"
+	https://dev.gentoo.org/~grobian/distfiles/${PN}-patches-11.3-r1.tar.bz2"
```
cognifloyd, line 33: @grobian The updated patches are in https://bugs.gentoo.org/631862

cognifloyd:
Sweet. I see this is fixed now :)

PR #18730, opened 2020-12-19
app-misc/g810-led: bump to 0.4.2 + various fixes

crabbedhaloablution:
0.4.2 brings support for g815, builds with clang, and various minor fixes.

Ebuild fixes:
Create and populate /etc/g810-led/ as in upstream makefile so systemd
.services work.
Change udev rules to use uaccess instead of MODE="666".
Set CXX to system compiler
Build and use "bin-linked" g810-led. Same features, less space.
Include headers in install
Add pkg_postinst with hints about use

Closes: https://bugs.gentoo.org/696468
Closes: https://bugs.gentoo.org/729288


app-misc/g810-led/g810-led-0.4.2.ebuild
```diff
@@ -0,0 +1,77 @@
...
+	emake LIB="$(usex hidapi hidapi libusb)" CXX="$(tc-getCXX)" bin-linked
+}
+
+src_install() {
+	dolib.so "lib/libg810-led.so.${PV}"
+	dosym "libg810-led.so.${PV}" "/usr/$(get_libdir)/libg810-led.so"
+
+	insinto /etc/g810-led/
+	newins sample_profiles/group_keys profile
+	newins sample_profiles/all_off reboot
```
lanodan, line 46: Those are installed as documentation, so you should either remove this new block or remove `sample_profiles` from `DOCS`.

app-misc/g810-led/g810-led-0.4.2.ebuild
```diff
@@ -0,0 +1,77 @@
...
+
+	dobin bin/g810-led
+	local boards=(213 410 413 512 513 610 815 910 pro)
+	local x
+	for x in "${boards[@]}"; do
+		dosym g810-led "/usr/bin/g${x}-led"
+	done
+
+	insinto /usr/include/g810-led
+	doins src/classes/*.h
```
lanodan, line 56: Just pure curiosity to see if I should fix 0.3.6: Is there other programs out there using g810-led as a library?

lanodan (changes requested):
Thanks for this version bump, I'm still using the 0.3.6 version and didn't try the 0.4.x branch but I guess it still breaks parts of my keyboard so that old version should be kept for now.

app-misc/g810-led/g810-led-0.4.2.ebuild
crabbedhaloablution: Not to my knowledge

app-misc/g810-led/g810-led-0.4.2.ebuild
crabbedhaloablution: I disagree.
The intention is that the files installed in /etc should be edited.
The sample_profiles provide examples of how to make a custom profile. Both are needed.

juippis: @lanodan what do you think? LGTM.

lanodan: Well, yeah, why not. LGTM

PR #18738, opened 2020-12-20
add zst support ( closing https://bugs.gentoo.org/760905 )

halcon74:
Hi!

Please add zst [1] support to unpacker.eclass.

I am going also to e-mail this patch to gentoo-dev.

[1] - https://en.wikipedia.org/wiki/Zstandard

### DIFF
--- a/eclass/unpacker.eclass
+++ b/eclass/unpacker.eclass
@@ -1,4 +1,4 @@
-# Copyright 1999-2017 Gentoo Foundation
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 # @ECLASS: unpacker.eclass
@@ -356,6 +356,8 @@ _unpacker() {
 	*.lz)
 		: ${UNPACKER_LZIP:=$(type -P plzip || type -P pdlzip || type -P lzip)}
 		comp="${UNPACKER_LZIP} -dc" ;;
+	*.zst)
+		comp="zstd -dfc" ;;
 	esac
 
 	# then figure out if there are any archiving aspects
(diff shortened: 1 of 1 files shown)


halcon74:
I am sorry for force-push; I have read about other ways to achieve the same result (git revert) already after having run `git reset HEAD^'...

eclass/unpacker.eclass
```diff
@@ -356,6 +356,9 @@ _unpacker() {
 	*.lz)
 		: ${UNPACKER_LZIP:=$(type -P plzip || type -P pdlzip || type -P lzip)}
 		comp="${UNPACKER_LZIP} -dc" ;;
+	*.zst)
+		zstd -df "${zst}" -o "${P}.tar"
+		tar -xpf "${P}.tar" -C "${WORKDIR}"
```
junghans, line 361:
```suggestion
		comp="zstd -dc"
```

halcon74: Thanks a lot! But I add also the "-f option" (zstd -dfc) because here is a symbolic link. Preparing a new commit/patch...

junghans: We aren't do that for any of the other decompressors, but it cannot hurt I guess.

halcon74:
Squashed 2 commits in 1 (this commit signature is not shown up because I had to rename the message after squashing).

eclass/unpacker.eclass
halcon74: Okay.

Otherwise (without -f) zstd was giving an error:

>>> Unpacking source...
>>> Unpacking worker-4.6.1.tar.zst to /var/tmp/portage/app-misc/worker-4.6.1/work
Warning : /var/tmp/portage/app-misc/worker-4.6.1/distdir/worker-4.6.1.tar.zst is a symbolic link, ignoring
tar: This does not look like a tar archive
tar: Exiting with failure status due to previous errors
 * ERROR: app-misc/worker-4.6.1::localrepo failed (unpack phase):
 *   unpacking /var/tmp/portage/app-misc/worker-4.6.1/distdir/worker-4.6.1.tar.zst failed (comp=zstd -dc arch=tar --no-same-owner -xof)

PR #18745, opened 2020-12-21
dev-libs/rocm-opencl-runtime: Version bump to 4.0.0

justxi:
@candrews 
Here you should find the new ebuilds for ROCm 4.0.0.
Please keep the previous ebuilds.
I think I have found a problem in llvm-roc regarding HIP when installling to "/usr". 
I currently create patches and need the "old" ebuilds for testing.


justxi:
... and would you add me as a second maintainer to the other ebuilds?

PR #18747, opened 2020-12-21
dev-python/rq, mail-filter/postgrey, net-irc/anope, net-libs/libpcap, sys-boot/raspberrypi-firmware: remove unused patches/files

mm1ke:


thesamesam:
Thanks a bunch Michael! Apologies for missing these - it is always really appreciated when you pick these up.

PR #18750, opened 2020-12-21
x11-misc/albert: version bump to 0.17.2

CaseOf:
Hello,
Here is an ebuild of albert 0.17.1 with some various changes from one for previous versions.
I also committed removal of old versions.


CaseOf:
Hi,
Updated pull request for v0.17.2

juippis (approved):

juippis:
We already discussed this in IRC but I want to state it here in record that I believe those python plugins should be installed with `python` IUSE.

juippis:
I also want to state it in here that you dropped the last ebuild with stable keywords, but I think we can suspect no one's using that anyway since it's ancient. You can stabilize 0.17.2 if you wish to have stable keywords for albert.

PR #18751, opened 2020-12-21
Wayland 1.18.91 version bump and new dev-util/wayland-scanner

mattst88:
`dev-util/wayland-scanner` is a new package with `/usr/bin/wayland-scanner` split out of `dev-libs/wayland` for easier cross-compiling.

Unfortunately there's a problem...

`dev-util/wayland-scanner` has an appropriate-looking blocker:

    RDEPEND="
    	!<dev-libs/wayland-${PV}
    	...
    "

but despite that `dev-libs/wayland-1.18.91` fails to configure with `ABI_X86=32` because... `<dev-libs/wayland-1.18.91` installs

    /usr/lib64/pkgconfig/wayland-scanner.pc
    /usr/lib/pkgconfig/wayland-scanner.pc

and `dev-util/wayland-scanner` only installs and overwrites `/usr/lib64/pkgconfig/wayland-scanner.pc`.

So, what can I do, short of a hard `!!<` blocker which really sucks for users?


mattst88:
I've made `dev-util/wayland-scanner` multilib so that it will install `.pc` files that overwrite the old wayland's `.pc` files.

dev-util/wayland-scanner/wayland-scanner-1.18.91.ebuild
```diff
@@ -0,0 +1,49 @@
...
+
+DESCRIPTION="wayland-scanner tool"
+HOMEPAGE="https://wayland.freedesktop.org/ https://gitlab.freedesktop.org/wayland/wayland"
+
+LICENSE="MIT"
+SLOT="0"
+
+BDEPEND="virtual/pkgconfig"
+RDEPEND="
+	!<dev-libs/wayland-${PV}
```
Arfrever, line 24: File collision was one-time event, so you can hardcode static version here:
```
	!<dev-libs/wayland-1.18.91
```

Arfrever:
Instead of multilib, maybe this would work:
```
src_install() {
	meson_src_install

	dodir /usr/share		# If necessary
	mv "${ED}/usr/$(get_libdir)/pkgconfig" "${ED}/usr/share/pkgconfig" || die
	rmdir "${ED}/usr/$(get_libdir)" || die
```

mattst88:
> Instead of multilib, maybe this would work in `dev-util/wayland-scanner`:
> 
> ```
> src_install() {
> 	meson_src_install
> 
> 	dodir /usr/share		# If necessary
> 	mv "${ED}/usr/$(get_libdir)/pkgconfig" "${ED}/usr/share/pkgconfig" || die
> 	rmdir "${ED}/usr/$(get_libdir)" || die
> }
> ```

I don't believe this helps, since `/usr/lib/pkgconfig/wayland-scanner.pc` will still be on the system (provided by `dev-libs/wayland-1.18.0`) at the time `dev-libs/wayland-1.18.91` is installed.

PR #18756, opened 2020-12-22
Update SELinux userspace dependencies

dm0-:
@perfinion


perfinion:
Merged with a few small changes and revbumps. Thanks!

PR #18768, opened 2020-12-22
[CI] Unmask slotted Lua + revdeps

mkszuba:
DO NOT MERGE.


band-a-prend:
@mkszuba 
Great thanks for gimp ebuilds updates for lua-slotting. 
I tested that `gimp-9999` and `gimp-2.99.2-r100` successfully merge with `USE="lua"` after unmasked some dependencies from `package.mask` too. 

The only additional changes are required for `gimp-9999`:
`>=media-libs/babl-0.1.84` instead of `0.1.82` version;
`>=media-libs/gegl-0.4.28:0.4` instead of `0.4.27:0.4` version;
`>=x11-libs/pango-1.44.7` instead of `1.42.4` ( Closes: https://bugs.gentoo.org/755962 )

P.S.
Gimp-2.99.2-r* could be incompatible with `gegl-9999` after last upstream changes failing to start because not found `gegl:introspect` but it's still works with `gegl-0.4.26`.

PR #18769, opened 2020-12-23
sci-libs/openblas: version bump 0.3.13

epsilon-0:


epsilon-0:
My changes have been merged upstream which now allow us to create separate BLAS64/LAPACK64 libraries but I have not enabled it here so far as there are other things that should be done first for getting better BLAS64/LAPACK64 library support.

Even reference lapack has merged my changes but we have not added/enabled them in gentoo yet. We should probably do that soon.

thesamesam:
Thanks!

PR #18772, opened 2020-12-23
games-strategy/augustus: new package

TheChymera:

@chewi as discussed in IRC. Upstream fixed this and there are no longer any collisions with julius.


TheChymera:
Updated PR with newest package version and with improvements analogous to those recommended by @ionenwks on the `games-strategy/julius` PR.

games-strategy/augustus/metadata.xml
```diff
@@ -0,0 +1,24 @@
...
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>games@gentoo.org</email>
+		<name>Gentoo Games Project</name>
+	</maintainer>
+	<maintainer type="person">
+		<email>gentoo@chymera.eu</email>
+		<name>Horea Christian</name>
+	</maintainer>
```
ionenwks, line 11: You can move this above games@, it'll make you as the primary maintainer.

games@ would preferably only touch this if there's an urgent issue or if you can't maintain it anymore.

games-strategy/augustus/metadata.xml
```diff
@@ -0,0 +1,24 @@
...
+		The aim of this project is to provide enhanced, customizable gameplay
+		to Caesar 3 using project Julius UI enhancements. Augustus is able to
+		load Caesar 3 and Julius saves, however saves made with Augustus will
+		not work outside Augustus.
+	</longdescription>
+	<upstream>
+		<remote-id type="github">Keriew/augustus</remote-id>
+	</upstream>
+	<use>
+		<flag name='uncapped'>Uncap grand temple count</flag>
```
ionenwks, line 22: Use `""`, for uniformity we avoid `''` in these files.

games-strategy/augustus/augustus-3.0.1.ebuild
```diff
@@ -0,0 +1,47 @@
...
+src_install() {
+	cmake_src_install
+	dodir /usr/libexec
+	mv "${ED}/usr/bin/augustus"  "${ED}/usr/libexec/augustus-game" ||
+		die "Failed to rename executable (required to set default resources location)."
+	newbin - augustus-game <<-EOF
+		#!/usr/bin/env sh
+		exec "${EPREFIX}/usr/libexec/augustus-game" "\${1:-\${HOME}/.local/share/julius/app}"
+	EOF
+	readme.gentoo_create_doc
```
ionenwks, line 41: I don't see the README.gentoo in this PR.

games-strategy/augustus/augustus-3.0.1.ebuild
```diff
@@ -0,0 +1,47 @@
...
+	media-libs/libpng:0=
+	media-libs/libsdl2[joystick,video,sound]
+	media-libs/sdl2-mixer
+"
+RDEPEND="
+	${DEPEND}
+"
+
+src_prepare() {
+	use uncapped && eapply "${FILESDIR}"/${PN}-3.0.1-uncapped.patch
```
ionenwks, line 27: Where's the patch? I guess omitted the entire files/ dir :)

Haven't seen it so I can't really tell yet, but what does this do? From the sound of the USE, is this some kind of game cheat? Is it really needed? Users can always use `/etc/portage/patches` for these kind of things -- nevermind if I'm misinterpreting what this does (unfortunately haven't played this game yet, so description doesn't say much to me).

On a side-note, USE-based conditional patches are generally frowned upon (notably given they're not upstream-able, and adding a autoconf/cmake option is preferable). Not that changing is *essential* to get this merged, but well. Depending on the above, may prefer the patch just removed if that's possible.

ionenwks (commented):
Haven't looked in-depth yet given some files are missing, but I wouldn't mind adding this from a quick look. Fork looks meaningful and ebuild is almost identical to julius making this straight forward.

TheChymera:
@ionenwks done :)

games-strategy/augustus/files/augustus-3.0.1-uncapped.patch
```diff
@@ -0,0 +1,16 @@
...
+Subject: [PATCH] Limiting grand temple number to 6.
+
+--- a/src/building/monument.h	2021-05-17 09:54:41.000000000 -0400
++++ a/src/building/monument.h	17:16:52.327542581 -0400
+@@ -6,7 +6,7 @@
+
+ #define MONUMENT_FINISHED -1
+ #define MONUMENT_START 1
+-#define MAX_GRAND_TEMPLES_PER_CITY 2
++#define MAX_GRAND_TEMPLES_PER_CITY 99
```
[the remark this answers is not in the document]
TheChymera: It's a common request that was discussed on a few occasions, e.g. here: https://github.com/Keriew/augustus/discussions/443

The patch is the fix recommended by upstream at this moment.

games-strategy/augustus/augustus-3.0.1.ebuild
```diff
@@ -0,0 +1,47 @@
...
+	cmake_src_prepare
+}
+
+src_install() {
+	cmake_src_install
+	dodir /usr/libexec
+	mv "${ED}/usr/bin/augustus"  "${ED}/usr/libexec/augustus-game" ||
+		die "Failed to rename executable (required to set default resources location)."
+	newbin - augustus-game <<-EOF
```
ionenwks, line 37: Noticed one last problem while doing final tests, the .desktop is still calling `augustus` rather than `augustus-game`. Fix however you see prefer (sed or patch in prepare works for me).

Other than this, everything seems ready.

games-strategy/augustus/files/augustus-3.0.1-uncapped.patch
```diff
@@ -0,0 +1,16 @@
+From 1d3a24d74bbea85198c2c5d739a16dcc139af18b Mon Sep 17 00:00:00 2001
+From: Horea Christian <chr@chymera.eu>
+Date: Sat, 26 Dec 2020 12:59:15 -0500
+Subject: [PATCH] Unlimited number of Grand Temples.
+
+--- a/src/building/monument.h	2021-05-17 09:54:41.000000000 -0400
++++ a/src/building/monument.h	17:16:52.327542581 -0400
```
ionenwks, line 7: On a side-note, it's unimportant but you can remove the useless timestamp cruft after the .h

Devmanual has a handy sed to clean patches when need to do this often:
https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#clean-patch-howto

TheChymera:
@ionenwks looking good now? :)

games-strategy/augustus/augustus-3.0.1.ebuild
```diff
@@ -0,0 +1,49 @@
...
+	dev-libs/expat
+	media-libs/libpng:0=
+	media-libs/libsdl2[joystick,video,sound]
+	media-libs/sdl2-mixer
+"
+RDEPEND="
+	${DEPEND}
+"
+
+PATCHES="${FILESDIR}/${PN}-3.0.1-desktop_rename.patch"
```
ionenwks, line 27: We normally never use a flat string for patches, it should be done similarly to julius ebuild, aka an array and patch on its own line (this generally makes it cleaner to add/remove patches as needed).

ionenwks (commented):
> @ionenwks looking good now? :)

I already get the notification that it changed, was going to look at this later and fix PATCHES on merge not to overly nag :)

But if you want to fix yourself meanwhile...

games-strategy/augustus/augustus-3.0.1.ebuild
TheChymera: Ok :)

ionenwks (approved):
Thanks.

Alright, time to get this in.

PR #18777, opened 2020-12-23
dev-util/rt-tests: Update to v1.10

shifty91:
Update rt-tests to version v1.10 and drop old. Thanks!


dev-util/rt-tests/rt-tests-1.10.ebuild
```diff
@@ -32,7 +32,7 @@ src_compile() {
 }
 
 src_install() {
-	emake prefix=/usr DESTDIR="${D}" MAN_COMPRESSION=none install
+	emake prefix=/usr DESTDIR="${D}" install
```
thesamesam, line 35: I think we need ${EPREFIX} _or_ ${ED} here.

thesamesam:
Thanks!

dev-util/rt-tests/rt-tests-1.10.ebuild
shifty91: Yeah. It seems like that should be `emake prefix="${EPREFIX}"/usr DESTDIR="${D}" install`.

PR #18778, opened 2020-12-23
dev-haskell/texmath: increase lower bound up to ghc 8.8

solpeth:
Closes: https://bugs.gentoo.org/735218


trofi:
Pulled as is. Thank you!

PR #18779, opened 2020-12-23
sys-libs/compiler-rt: build with atomic support

walker27460:
This build compiler-rt with libatomic support


walker27460:
[please reassign]

PR #18780, opened 2020-12-23
sys-libs/compiler-rt: build with atomic support

walker27460:
Signed-off-by: Yui Tanabe <yuitanabe@aol.com>


mgorny (changes requested):
You've got the tab indentation wrong.

Also, the option desc says:
```
option(COMPILER_RT_EXCLUDE_ATOMIC_BUILTIN
  "Skip the atomic builtin (these should normally be provided by a shared library)"
  On)
```

Could you elaborate on why you're enabling it?

walker27460:
I want to build LLVM-based Linux system because the binary code generated by LLVM is faster than gcc. Clang uses libclang_rt.builtins-x86_64.a as gcc's libgcc_s.so and building LLVM require atomic support in it.

mgorny:
```
    [compiler-rt] [builtins] Don't build __atomic_* by default.
    
    The locks need to be implemented in a shared library to work correctly,
    so they shouldn't be part of libclang_rt.builtins.a, except in
    specialized scenarios where the user can prove it will only be linked
    once.
```

So I'm pretty convinced it won't work. We probably need a standalone libatomic.

PR #18787, opened 2020-12-23
RFC: landing HIP to ::gentoo

heroxbd:
Hi @justxi @littlewu2508 @candrews,

I intend to land HIP to ::gentoo together with patches to llvm-roc to support Prefix.  Could you please let know your thoughts?

Thanks,
Benda


candrews:
It looks pretty good!

* Can you please add me as a co-maintainer?
* In the commit message for the commit that adds hip, the line shouldn't just be "initial commit" it should be a description of the package, something like `sys-devel/hip: C++ Heterogeneous-Compute Interface for Portability`
* Have all the patches been submitted as PRs to upstream? If not, can we please do that?
* I _think_ the `eapply_user` calls should come before the `sed` calls. That order would make it easier for users to get patches from upstream and apply them. Do you agree?

mgorny:
Please no more random packages to `sys-devel`.

justxi:
Cool, as mentioned before I will support it.

I have created "hip-4.0.0-r2" which fixes problems with other ebuilds and got some cleanups. 
(https://github.com/justxi/rocm/blob/master/sys-devel/hip/hip-4.0.0-r2.ebuild)

We should think about setting an environment variable "DEVICE_LIB_PATH" to use it in other ebuilds(?).
But I think best location for that would be "dev-libs/rocm-device-libs".

Furthermore, I encountered problems when compiling packages with hip based on llvm-roc. On my system that seems to be a wrong order of system include directories (isystem) when installing to "/usr". I created a patch for llvm-roc 4.0.0 (https://github.com/justxi/rocm/blob/master/sys-devel/llvm-roc/files/llvm-roc-4.0.0-remove-isystem-usr-include.patch). With this modification it works on my system. It seems that previous version like 3.9 and 3.10 were also affected.

heroxbd:
> It looks pretty good!
> 
>     * Can you please add me as a co-maintainer?

Added.  I would also like to invite you to the Science Project.  What do you say?

>     * In the commit message for the commit that adds hip, the line shouldn't just be "initial commit" it should be a description of the package, something like `sys-devel/hip: C++ Heterogeneous-Compute Interface for Portability`

Modified.

>     * Have all the patches been submitted as PRs to upstream? If not, can we please do that?

That takes more time for us to decide which to submit to upstream. Because most of the patches result from difference of conventions of Gentoo and upstream.  For example, `/opt/rocm` vs `/usr/lib/hip`.
 
>     * I _think_ the `eapply_user` calls should come before the `sed` calls. That order would make it easier for users to get patches from upstream and apply them. Do you agree?

Yes, indeed.

heroxbd:
> Please no more random packages to `sys-devel`.

Moved to `dev-util` after `dev-util/nvidia-cuda-toolkit`.

heroxbd:
> Cool, as mentioned before I will support it.
> 
> I have created "hip-4.0.0-r2" which fixes problems with other ebuilds and got some cleanups.
> (https://github.com/justxi/rocm/blob/master/sys-devel/hip/hip-4.0.0-r2.ebuild)

Incorporated.
 
> We should think about setting an environment variable "DEVICE_LIB_PATH" to use it in other ebuilds(?).
> But I think best location for that would be "dev-libs/rocm-device-libs".

I am not fan of setting a ton of environment variables.  Is it possible to make it read from a configuration file, or at least patch the corresponding files to look from the correct locations? 
 
> Furthermore, I encountered problems when compiling packages with hip based on llvm-roc. On my system that seems to be a wrong order of system include directories (isystem) when installing to "/usr". I created a patch for llvm-roc 4.0.0 (https://github.com/justxi/rocm/blob/master/sys-devel/llvm-roc/files/llvm-roc-4.0.0-remove-isystem-usr-include.patch). With this modification it works on my system. It seems that previous version like 3.9 and 3.10 were also affected.

I cannot reproduce the bug.  I think it needs further investigate to understand what -internal-system does and how it failed in your case.

justxi:
> > We should think about setting an environment variable "DEVICE_LIB_PATH" to use it in other ebuilds(?).
> > But I think best location for that would be "dev-libs/rocm-device-libs".
> 
> I am not fan of setting a ton of environment variables. Is it possible to make it read from a configuration file, or at least patch the corresponding files to look from the correct locations?

Ok, I will think about it.
 
> > Furthermore, I encountered problems when compiling packages with hip based on llvm-roc. On my system that seems to be a wrong order of system include directories (isystem) when installing to "/usr". I created a patch for llvm-roc 4.0.0 (https://github.com/justxi/rocm/blob/master/sys-devel/llvm-roc/files/llvm-roc-4.0.0-remove-isystem-usr-include.patch). With this modification it works on my system. It seems that previous version like 3.9 and 3.10 were also affected.
> 
> I cannot reproduce the bug. I think it needs further investigate to understand what -internal-system does and how it failed in your case.

Did you build ebuilds of my repository, e.g. "rccl" and "sci-libs"?

heroxbd:
> Did you build ebuilds of my repository, e.g. "rccl" and "sci-libs"?

I see.  I have not gone beyond `hip` this round.  Let's focus this PR on getting hip to ::gentoo and discuss about llvm-roc `-internal-system` hacks in a different issue or PR.

heroxbd:
Guys, thank you for your inputs.  Please give another round of review.

heroxbd:
> > > We should think about setting an environment variable "DEVICE_LIB_PATH" to use it in other ebuilds(?).
> > > But I think best location for that would be "dev-libs/rocm-device-libs".
> > 
> > 
> > I am not fan of setting a ton of environment variables. Is it possible to make it read from a configuration file, or at least patch the corresponding files to look from the correct locations?
> 
> Ok, I will think about it.

I replaced the `$ENV{}` in the perl script.  I don't find any value of the flexibility to set alternative locations of `amdgcn/bitcode`.  With gentoo, they are in a predictable place.

justxi:
> > > > We should think about setting an environment variable "DEVICE_LIB_PATH" to use it in other ebuilds(?).
> > > > But I think best location for that would be "dev-libs/rocm-device-libs".
> > > 
> > > 
> > > I am not fan of setting a ton of environment variables. Is it possible to make it read from a configuration file, or at least patch the corresponding files to look from the correct locations?
> > 
> > 
> > Ok, I will think about it.
> 
> I replaced the `$ENV{}` in the perl script. I don't find any value of the flexibility to set alternative locations of `amdgcn/bitcode`. With gentoo, they are in a predictable place.

Yes, we keep the bitcode files in "/usr/lib/amdgcn/bitcode" and then we have to adjust only "dev-libs/rocm-opencl-runtime".

justxi:
> > Did you build ebuilds of my repository, e.g. "rccl" and "sci-libs"?
> 
> I see. I have not gone beyond `hip` this round. Let's focus this PR on getting hip to ::gentoo and discuss about llvm-roc `-internal-system` hacks in a different issue or PR.

Ok. But you can confirm the problem?

heroxbd:
> Ok. But you can confirm the problem?

Yes I confirmed the problem.  That's a crucial patch.

justxi:
> > Ok. But you can confirm the problem?
> 
> Yes I confirmed the problem. That's a crucial patch.
 
A last comment on this: with this https://github.com/justxi/rocm/issues/174#issuecomment-749479598 you can see the difference. Before you add the patch, I will create a new one, which includes the if-statement, which is  needless now.

heroxbd:
> > > Ok. But you can confirm the problem?
> > 
> > 
> > Yes I confirmed the problem. That's a crucial patch.
> 
> A last comment on this here: with this [justxi/rocm#174 (comment)](https://github.com/justxi/rocm/issues/174#issuecomment-749479598) you can see the difference in the include path order, if you compare the results with and without the patch. Before you add the patch, I will create a new one, which includes the if-statement, which is needless now.
> 
> The same problem arises with "hipcc" if "${ROCM_PATH}/include" is not removed (it is already done in the ebuild).

Understood.  Will incorporate it to our llvm-roc::gentoo ebuild.

justxi:
Due to the fact, that HIP is now in Gentoo portage, I will adapt my ebuilds and remove the hip ebuild from my repository.

PR #18796, opened 2020-12-23
package.mask: last rite media-gfx/openexr_viewers

waebbl:


waebbl:
Will get cleaned-up with openexr cleanup, see https://bugs.gentoo.org/770229

PR #18767, opened 2020-12-22
sci-libs/kissfft: new package

stkw0:
Moved from guru because sci-libs/vtk needs it (https://github.com/gentoo/guru/issues/23)

Bug: https://bugs.gentoo.org/754684
Package-Manager: Portage-3.0.12, Repoman-3.0.2
Signed-off-by: David Roman <davidroman96@gmail.com>

PR #15966, opened 2020-05-25
sci-libs/vtk: bump to 9.0.1

waebbl:
Closes: https://bugs.gentoo.org/723744
Package-Manager: Portage-3.0.19, Repoman-3.0.3
Signed-off-by: Bernd Waibel <waebbl-gentoo@posteo.net>

Edit: update portage and repoman versions.
Edit: update email

Everything re-keyworded. Thanks to arm{,64} team for checking the dependencies for this.

PR #18799, opened 2020-12-24
sci-libs/vtk: version bump to 9.0.1 + dependencies

epsilon-0:
Still a WIP...

---

Depends on https://github.com/gentoo/gentoo/pull/18767
Depends on https://bugs.gentoo.org/761439
Depends on https://bugs.gentoo.org/761460

---

Even better, all `VTK_USE_SYSTEM_<PKG>` have been removed and are called (I think) `VTK_MODULE_USE_EXTERNAL_<PKG>`


a17r:
See also https://github.com/gentoo/gentoo/pull/15966

sci-libs/vtk/vtk-9.0.1.ebuild
```diff
@@ -0,0 +1,352 @@
...
+		-DBUILD_EXAMPLES=$(usex examples)
+		-DVTK_MODULE_USE_EXTERNAL_utf8cpp=ON
+		-DVTK_USE_SYSTEM_AUTOBAHN=ON
+		-DVTK_USE_SYSTEM_EXPAT=ON
+		-DVTK_USE_SYSTEM_FREETYPE=ON
+		-DVTK_USE_SYSTEM_FreeType=ON
+		-DVTK_USE_SYSTEM_GL2PS=ON
+		-DVTK_USE_SYSTEM_HDF5=ON
+		-DVTK_USE_SYSTEM_JPEG=ON
+		-DVTK_USE_SYSTEM_LIBPROJ=OFF
```
Alessandro-Barbieri, line 190: Why depend on proj if isn't used? Is it possible to enable?

sci-libs/vtk/vtk-9.0.1.ebuild
```diff
@@ -0,0 +1,352 @@
...
+		-DVTK_USE_COCOA=$(usex aqua)
+	)
+
+	if use java; then
+		local javacargs=$(java-pkg_javac-args)
+		mycmakeargs+=( -DJAVAC_OPTIONS=${javacargs// /;} )
+	fi
+
+	if use mpi; then
+		mycmakeargs+=( -DVTK_USE_SYSTEM_MPI4PY=ON )
```
Alessandro-Barbieri, line 252: I think this should be nested into the if use python below
See also line 92

epsilon-0: There are so many changes to make, reviewing and commenting right now is not that useful.
Basically, most options and workings are going to be replaced.

thesamesam: Yeah, I take non-draft to mean "ready for comments". Let her work for now. I use GitHub PRs like this for CI and to keep the work safe.

PR #18802, opened 2020-12-24
Cinnamon 4.6 stable prep

SparkyBluefang:
Doing some clean-up prep work before submitting a stabilization request for Cinnamon 4.6 packages and starting on Cinnamon 4.8 packages.

* Remove old Cinnamon 4.4 versions
* Fix a few miscellaneous ebuild bugs

For future reference, should bulk removals like this still be broken out by package? Or can it be squashed into a single commit?


thesamesam:
One PR with multiple commits is preferred if that's ok?

thesamesam:
Thanks! 💯

PR #18807, opened 2020-12-24
app-emulation/firecracker-bin: Bump to 0.24.1

s-hamann:
Version bump for the release of Firecracker 0.24.0.
Upstream distributes the binaries in a slightly different way now, so this PR does a little more than renaming the ebuild.


s-hamann:
Firecracker 0.24.1 has been released, so I repurposed this PR to get that in the tree instead of the original 0.24.0.

juippis (approved):

PR #18810, opened 2020-12-25
qmake-utils.eclass: Drop obsolete EAPI-6 support, ban qt4* functions

a17r:
Last consumer in Gentoo ebuild repository was dropped.

### DIFF
--- a/eclass/qmake-utils.eclass
+++ b/eclass/qmake-utils.eclass
@@ -22,7 +22,7 @@ case ${EAPI} in
 	*) die "EAPI=${EAPI:-0} is not supported" ;;
 esac
 
-inherit estack toolchain-funcs
+inherit toolchain-funcs
 
 # @FUNCTION: _qmake-utils_banned_func
 # @INTERNAL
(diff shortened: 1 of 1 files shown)


eclass/qmake-utils.eclass
```diff
@@ -1,71 +1,75 @@
...
+esac
 
 inherit estack toolchain-funcs
 
+# @FUNCTION: _qmake-utils_banned_func
+# @INTERNAL
+# @DESCRIPTION:
+# Banned functions are banned.
+_qmake-utils_banned_func() {
+	die "${FUNCNAME[1]} is banned in EAPI 7 and later"
```
Pesa, line 32: [very minor, feel free to ignore] "in EAPI 7 and later" may be misleading... does it imply it's _not_ banned in EAPI 6 and earlier?

Pesa (approved):

eclass/qmake-utils.eclass
a17r: right, it's just a continuation of the existing message, when one earlier EAPI had still existed.

PR #18811, opened 2020-12-25
sys-cluster/glusterfs: Added versions 7.9 and 8.3

jpds:
Tested on an `amd64` system.


jkroonza:
Would you mind adding yourself as co-maintainer please?  Generally I'm happy to bump and patch as needed, but I'm not always available, so it would be great to have some help on this.

jkroonza (commented):
7.9 looks good.  Adding the python_optimize requirement.

Move to 8.3 drops infiniband support, and I note changes to georeplication too - could you please motivate this?  Dropping of infiniband upstream seems unlikely (but I won't say impossible).  I do know some people that used infiniband at some point, so don't want to drop this unless we don't have a choice.

You're also dropping the flock patch - which I'm assuming is applied upstream?  Are you able to confirm that this is in fact the case and that this patch is really no longer required?

jkroonza (approved):
Discussed with jpds, happy with the changes for 8.

sys-cluster/glusterfs/glusterfs-8.3.ebuild
```diff
@@ -0,0 +1,200 @@
...
+	python_setup "python3*"
+	python-single-r1_pkg_setup
+}
+
+src_prepare() {
+	default
+
+	# build rpc-transport and xlators only once as shared libs
+	find rpc/rpc-transport xlators -name Makefile.am |
+		xargs sed -i 's|.*$(top_srcdir).*\.sym|\0 -shared|' || die
```
juippis, line 87: `find -exec sed`

sys-cluster/glusterfs/glusterfs-8.3.ebuild
```diff
@@ -0,0 +1,200 @@
...
+
+	if use emacs ; then
+		elisp-install ${PN} extras/glusterfs-mode.el*
+		elisp-site-file-install "${FILESDIR}/${SITEFILE}"
+	fi
+
+	if use vim-syntax ; then
+		insinto /usr/share/vim/vimfiles/ftdetect; doins "${FILESDIR}"/${PN}.vim
+		insinto /usr/share/vim/vimfiles/syntax; doins extras/${PN}.vim
+	fi
```
juippis, line 140: This should not be a USE flag and you should just install them unconditionally.

sys-cluster/glusterfs/glusterfs-7.9.ebuild
```diff
@@ -0,0 +1,213 @@
...
+if [[ ${PV#9999} != ${PV} ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/gluster/glusterfs.git"
+else
+	SRC_URI="https://download.gluster.org/pub/gluster/${PN}/$(ver_cut 1)/${PV}/${P}.tar.gz"
+	KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~x86"
+fi
+
+DESCRIPTION="GlusterFS is a powerful network/cluster filesystem"
+HOMEPAGE="https://www.gluster.org/"
```
juippis, line 19: Please add the Github page as 2ndary HOMEPAGE here.

juippis (commented):
Can you drop the commit for 7.9, since if you also introduce 8.3 with same KEYWORDS no one will install 7.9?

jkroonza:
@juippis there may be need for the moment to keep both 7.X and 8.X.  I suggest we subslot here too.  The problem again is that useful features has been dropped from 8 and someone may require it.

@jpds:  SLOT="0/${PV%%.*}"

Again a case where I'd really, really LOVE to have some installation statistics.  I reckon once we get to 8.5 we mask <8.0 and give it 90 days and see if we get comebacks.  90 days because from what I've seen a migration for users of georeplication could be difficult and may require sensible planning.

Agree with all other proposed changed, @jpds - you happy to action?  Would be simpler than me fetching the PR and having to figure out how to push back into the same PR (if that's even possible).

jpds:
@juippis Thanks for the review, I've made the changes.

As for the version number, it's a bit odd as https://www.gluster.org/install/ calls the 7.x series as the current stable release, hence me including both. I've just noticed 9.0 is available now so I've added that too.

@jkroonza Sorry, I thought I mentioned it on IRC, only the infiniband RDMA support has been dropped. Georeplication is still there, but the helper script is no longer installed into `/usr/share/`, just into `/usr/libexec/` per https://bugzilla.redhat.com/show_bug.cgi?id=1387404 - which I've linked in the 8.3 commit.

jkroonza:
> @juippis Thanks for the review, I've made the changes.
> 
> As for the version number, it's a bit odd as https://www.gluster.org/install/ calls the 7.x series as the current stable release, hence me including both. I've just noticed 9.0 is available now so I've added that too.

What I'm getting from that is that the world is confused!  What's the official gluster master currently?  Because https://www.gluster.org/install/ doesn't even mention 9.  Having said that, if 9.0 is released, then either 9.0 is continuation of 8.X branch (which doesn't make sense), or 8.X really should become our stable target.  As such, let's go with adding all three (for now).

> @jkroonza Sorry, I thought I mentioned it on IRC, only the infiniband RDMA support has been dropped. Georeplication is still there, but the helper script is no longer installed into `/usr/share/`, just into `/usr/libexec/` per https://bugzilla.redhat.com/show_bug.cgi?id=1387404 - which I've linked in the 8.3 commit (I've also tested that this is properly installed or not with the USE flag).

You quite possibly did.  I recall something like this now that you mention infiniband.  Let's hope none of our users use infiniband.  From discussion it sounds like the majority of large scale consumers of gluster is facebook and google (based on who redhat insinuated contributed certain features) and they're likely to use 40G or even 100G (again, I know that Facebook utilizes some of the gear from Netronome which will for example plug in on the PCIe's of a couple of servers at 40G and then aggregate a 100G uplink between up to four hosts or something, including various pieces of network offloading.

> Should I still add the subslotting?

Please.  Especially if we're going down the route of masking eg 7, which I think we should do as soon as we mark the newly pushed 8 as stable.  The added advantage of sub slots here are w.r.t. stable candidates listing etc ... for example, with asterisk, the tooling actually starts displaying things differently (I recommend you run this on your machine as the colour output helps too:

```jkroon@plastiekpoot ~$ eshowkw net-misc/asterisk
Keywords for net-misc/asterisk:
              |                             |   u      |  
              | a   a     p s     a   r     |   n      |  
              | m   r h   p p   s l i i m m | e u s    | r
              | d a m p p c a x 3 p a s 6 i | a s l    | e
              | 6 r 6 p p 6 r 8 9 h 6 c 8 p | p e o    | p
              | 4 m 4 a c 4 c 6 0 a 4 v k s | i d t    | o
--------------+-----------------------------+----------+-------
   13.38.1    | + ~ ~ o ~ ~ o + o o o o o o | 7 o 0    | gentoo
   13.38.1-r1 | ~ ~ ~ o ~ ~ o ~ o o o o o o | 7 #      | gentoo
--------------+-----------------------------+----------+-------
[I]13.38.1-r2 | ~ ~ ~ o ~ ~ o ~ o o o o o o | 7 # 0/13 | gentoo
--------------+-----------------------------+----------+-------
   16.15.1    | ~ ~ ~ o ~ ~ o ~ o o o o o o | 7 o 0    | gentoo
   16.15.1-r1 | ~ ~ ~ o ~ ~ o ~ o o o o o o | 7 o      | gentoo
--------------+-----------------------------+----------+-------
   16.15.1-r2 | ~ ~ ~ o ~ ~ o ~ o o o o o o | 7 o 0/16 | gentoo
```

jpds:
@jkroonza Subslotting added. https://github.com/gluster/glusterfs/releases seems to be tracking releases of 7.x all the way to 10.dev releases (and at the same time too).

sys-cluster/glusterfs/glusterfs-9.0.ebuild
```diff
@@ -0,0 +1,198 @@
...
+	python_setup "python3*"
+	python-single-r1_pkg_setup
+}
+
+src_prepare() {
+	default
+
+	# build rpc-transport and xlators only once as shared libs
+	find rpc/rpc-transport xlators -name Makefile.am -exec \
+		sed -i 's|.*$(top_srcdir).*\.sym|\0 -shared|' || die
```
juippis, line 87:
```
find: missing argument to `-exec'
 * ERROR: sys-cluster/glusterfs-9.0::testworld failed (prepare phase):
 *   (no error message)
 * 
 * Call stack:
 *     ebuild.sh, line 125:  Called src_prepare
 *   environment, line 2833:  Called die
 *       find rpc/rpc-transport xlators -name Makefile.am -exec sed -i 's|.*$(top_srcdir).*\.sym| -shared|' || die;
```

sys-cluster/glusterfs/glusterfs-9.0.ebuild
```diff
@@ -0,0 +1,198 @@
...
+fi
+
+DESCRIPTION="GlusterFS is a powerful network/cluster filesystem"
+HOMEPAGE="https://www.gluster.org/ https://github.com/gluster/glusterfs/"
+
+LICENSE="|| ( GPL-2 LGPL-3+ )"
+SLOT="0"
+IUSE="debug emacs +fuse +georeplication ipv6 libressl +libtirpc rsyslog static-libs +syslog test +xml"
+
+REQUIRED_USE="georeplication? ( ${PYTHON_REQUIRED_USE} )
```
juippis, line 25:
```
configure: error: geo-replication requires xml output

!!! Please attach the following file when seeking support:
!!! /var/tmp/portage/sys-cluster/glusterfs-9.0/work/glusterfs-9.0/config.log
 * ERROR: sys-cluster/glusterfs-9.0::testworld failed (configure phase):
 *   econf failed
```

juippis (approved):
Will apply fixes for whatever I commented, but please do check that the sed works like it's supposed to.

sys-cluster/glusterfs/glusterfs-7.9.ebuild
```diff
@@ -0,0 +1,211 @@
...
+	sys-devel/flex
+	virtual/acl
+	test? ( >=dev-util/cmocka-1.0.1
+		app-benchmarks/dbench
+		dev-vcs/git
+		net-fs/nfs-utils
+		virtual/perl-Test-Harness
+		dev-libs/yajl
+		sys-fs/xfsprogs
+		sys-apps/attr )
```
juippis, line 63: Also please sort these test deps alphabetically when you have time / are touching these ebuilds next time.

PR #18601, opened 2020-12-10
sys-auth/fprintd: version bump to 1.90.8, switch to meson buildsystem

bearsh:
Closes: https://bugs.gentoo.org/700282
Supersedes: https://github.com/gentoo/gentoo/pull/15248
Package-Manager: Portage-3.0.12, Repoman-3.0.2
Signed-off-by: Martin Gysel <me@bearsh.org>

PR #18815, opened 2020-12-26
Define slots for libfprint

dm0-:
@floppym This is part of #18813 and #18814.


floppym:
I think we need @Zlogene  and possibly @maksbotan to review the libfprint and fprintd changes.

sys-auth/libfprint/libfprint-1.90.1.ebuild
```diff
@@ -10,23 +10,24 @@ HOMEPAGE="https://cgit.freedesktop.org/libfprint/libfprint/ https://github.com/f
 SRC_URI="https://gitlab.freedesktop.org/libfprint/libfprint/-/archive/v${PV}/${PN}-v${PV}.tar.gz -> ${P}.tar.gz"
 
 LICENSE="LGPL-2.1"
-SLOT="0"
+SLOT="2"
```
SoapGentoo, line 13: requires a revbump

sys-auth/libfprint/libfprint-1.90.2.ebuild
```diff
@@ -10,23 +10,24 @@ HOMEPAGE="https://cgit.freedesktop.org/libfprint/libfprint/ https://github.com/f
 SRC_URI="https://gitlab.freedesktop.org/libfprint/libfprint/-/archive/v${PV}/${PN}-v${PV}.tar.gz -> ${P}.tar.gz"
 
 LICENSE="LGPL-2.1"
-SLOT="0"
+SLOT="2"
```
SoapGentoo, line 13: ditto

SoapGentoo (changes requested):

sys-auth/libfprint/libfprint-1.90.1.ebuild
mgorny: ...and a blocker on the previous version since the new slot's going to collide with it. Or maybe a slotmove but I'm not sure about partial moves.

sys-auth/fprintd/fprintd-0.8.1-r1.ebuild
```diff
@@ -18,11 +18,13 @@ IUSE="doc pam static-libs"
 RDEPEND="
```
mgorny, line 18: Increase revision when changing deps (just `git mv` it), so that people get it rebuilt. Otherwise, people with the old version installed will get libfprint upgraded and old slot removed.

sys-auth/libfprint/libfprint-1.90.1.ebuild
dm0-: Okay, all slot changes got a revbump and blockers are added.

sys-auth/libfprint/libfprint-1.90.1-r1.ebuild
```diff
@@ -10,23 +10,25 @@ HOMEPAGE="https://cgit.freedesktop.org/libfprint/libfprint/ https://github.com/f
...
-	dev-libs/nss
 	>=dev-libs/libgusb-0.3.1
+	dev-libs/nss
 	virtual/libusb:1=
 	x11-libs/gtk+:3
 	x11-libs/pixman
 	x11-libs/libX11
-	x11-libs/libXv"
+	x11-libs/libXv
+	introspection? ( dev-libs/gobject-introspection:= )
```
Arfrever, line 25: According to what I see in [libfprint/meson.build](https://gitlab.freedesktop.org/libfprint/libfprint/-/blob/2d10d864d8bc4b952cf412f6e18e44412dba759a/libfprint/meson.build#L320-349), there seems to be no linking against any library provided by `dev-libs/gobject-introspection`.
For generation of `*.gir` and `*.typelib` files, `dev-libs/gobject-introspection` should be in `BDEPEND`.

dm0-: I put it in DEPEND+RDEPEND because that's where basically every GNOME package puts introspection.  There is not much hope of cross-compiling introspection in Gentoo (see for example https://maxice8.github.io/8-cross-the-gir/), so it shouldn't make a difference, but I guess I can move this one.  It's on my TODO list to hack it together at some point (maybe after #18813).

mgorny:
Could you try doing a slotmove here after all? I've asked around and it seems to work correctly with version ranges.

dm0-:
I pushed a slotmove and moved introspection deps.

sys-auth/fprintd/fprintd-0.8.1-r2.ebuild
```diff
@@ -55,10 +55,10 @@ src_install() {
 	dodoc AUTHORS NEWS README{,.transifex} TODO
 	newdoc pam/README README.pam_fprintd
 	if use doc ; then
-		insinto /usr/share/doc/${PF}/html
-		doins doc/{fprintd-docs,version}.xml
-		insinto /usr/share/doc/${PF}/html/dbus
-		doins doc/dbus/net.reactivated.Fprint.{Device,Manager}.ref.xml
+		docinto /usr/share/doc/${PF}/html
```
mgorny, line 58: `docinto` takes a relative path by design. Otherwise replacing `insinto` wouldn't make any sense.

sys-auth/libfprint/libfprint-1.90.1-r1.ebuild
```diff
@@ -10,23 +10,25 @@ HOMEPAGE="https://cgit.freedesktop.org/libfprint/libfprint/ https://github.com/f
...
+	!>=${CATEGORY}/${PN}-1.90:0"
 
-DEPEND="${RDEPEND}
-	dev-util/gtk-doc"
+DEPEND="${RDEPEND}"
 
-BDEPEND="virtual/pkgconfig"
+BDEPEND="dev-util/gtk-doc
+	virtual/pkgconfig
+	introspection? ( dev-libs/gobject-introspection:= )"
```
Arfrever, line 31: `:=` is not needed in `BDEPEND`.

mattst88:
Thanks!

bearsh:
just for your information, I've updated #18601 (frpintd version bump) to depend on slot 2

by the way, @dm0- libfprint still depends on dev-util/gtk-doc which I believe is wrong, see also https://bugs.gentoo.org/755167

bearsh:
> by the way, @dm0- libfprint still depends on dev-util/gtk-doc which I believe is wrong, see also https://bugs.gentoo.org/755167

I've created #18962 to address that

PR #18818, opened 2020-12-26
www-server/lighttpd: Support wasm files

d-e-s-o:
`.wasm` files should be server with the `application/wasm` MIME type or some
features such as proper streamed loading and compilation may not be used
by browsers.
This change adjusts the lighttpd ebuild to include such a setting in its MIME
type configuration.

Signed-off-by: Daniel Müller <deso@posteo.net>


juippis (approved):
Hmm need to think whether lighttpd should be revbumped after this.

PR #18819, opened 2020-12-26
app-portage/emlop: New ebuild

vincentdephily:
Upstream author here, happy to proxy-maintain.
Prior ebuild development happens at https://github.com/vincentdephily/moltonel-ebuilds

Closes: https://bugs.gentoo.org/649904


vincentdephily:
Force-pushed commit that includes the signoff.

telans:
Append `[please reassign]` to the title. Do you mind if I add this to GURU while waiting for merge?

vincentdephily:
Changed the title :+1: 

I don't know what GURU is, but whatever helps with merging/testing/distributing I'm happy with. Note that the `moltonel` overlay (available from `eselect repository`) has this ebuild.

vincentdephily:
Found the [GURU info](https://wiki.gentoo.org/wiki/Project:GURU), by all means add this ebuild there, though it's already available in my overlay and from crates.io (run `cargo install emlop`). So in my mind getting this into the official gentoo repo is the main todo.

app-portage/emlop/emlop-0.4.1.ebuild
```diff
@@ -0,0 +1,108 @@
...
+DESCRIPTION="A fast, accurate, ergonomic emerge.log parser"
+HOMEPAGE="https://github.com/vincentdephily/emlop"
+SRC_URI="$(cargo_crate_uris ${CRATES})"
+RESTRICT="mirror"
+# License set may be more restrictive as OR is not respected
+# use cargo-license for a more accurate license picture
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="bash-completion zsh-completion fish-completion"
```
Flowdalic, line 86: We do not install shell completion files via a USE conditional. Please remove those use flags.

app-portage/emlop/emlop-0.4.1.ebuild
```diff
@@ -0,0 +1,108 @@
...
+
+inherit cargo bash-completion-r1
+
+DESCRIPTION="A fast, accurate, ergonomic emerge.log parser"
+HOMEPAGE="https://github.com/vincentdephily/emlop"
+SRC_URI="$(cargo_crate_uris ${CRATES})"
+RESTRICT="mirror"
+# License set may be more restrictive as OR is not respected
+# use cargo-license for a more accurate license picture
+LICENSE="GPL-3"
```
Flowdalic, line 83: The license declaration is incomplete.

app-portage/emlop/emlop-0.4.1.ebuild
```diff
@@ -0,0 +1,108 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="bash-completion zsh-completion fish-completion"
+
+DEPEND=">=virtual/rust-1.41.1"
+RDEPEND=""
```
Flowdalic, line 89: Remove this empty assignment.

app-portage/emlop/emlop-0.4.1.ebuild
```diff
@@ -0,0 +1,108 @@
...
+winapi-x86_64-pc-windows-gnu-0.4.0
+"
+
+inherit cargo bash-completion-r1
+
+DESCRIPTION="A fast, accurate, ergonomic emerge.log parser"
+HOMEPAGE="https://github.com/vincentdephily/emlop"
+SRC_URI="$(cargo_crate_uris ${CRATES})"
+RESTRICT="mirror"
+# License set may be more restrictive as OR is not respected
```
Flowdalic, line 81: Remove this comment.

app-portage/emlop/emlop-0.4.1.ebuild
```diff
@@ -0,0 +1,108 @@
...
+winapi-i686-pc-windows-gnu-0.4.0
+winapi-util-0.1.5
+winapi-x86_64-pc-windows-gnu-0.4.0
+"
+
+inherit cargo bash-completion-r1
+
+DESCRIPTION="A fast, accurate, ergonomic emerge.log parser"
+HOMEPAGE="https://github.com/vincentdephily/emlop"
+SRC_URI="$(cargo_crate_uris ${CRATES})"
```
Flowdalic, line 79:
```suggestion
SRC_URI="$(cargo_crate_uris)"
```

app-portage/emlop/emlop-0.4.1.ebuild
```diff
@@ -0,0 +1,108 @@
+# Copyright 2017-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# Auto-Generated by cargo-ebuild 0.3.1
+
+EAPI=7
```
Flowdalic, line 6:
```suggestion
EAPI=8
```

app-portage/emlop/emlop-0.4.1.ebuild
```diff
@@ -0,0 +1,108 @@
...
+
+CRATES="
+aho-corasick-0.7.15
```
Flowdalic, line 9: IIRC newer vesions of cargo-ebuild intent the CRATES with a tab.
```suggestion
	aho-corasick-0.7.15
```
(and so on)

Flowdalic (changes requested):
Thanks for your contribution. Sorry that it took so long.

I would be willing to maintain emlop in ::gentoo and add you as proxied maintainer. If you still want to work on this, together with me, then please update this PR to add the latest version of emlop. Which appears to be 0.5.0 at the time of writing this.

app-portage/emlop/emlop-0.4.1.ebuild
vincentdephily: Install completions for all shells unconditionally ?

app-portage/emlop/emlop-0.4.1.ebuild
vincentdephily: Emlop itself is `GPL-3.0-only`, but its deps are a mix of `Apache-2.0 OR BSL-1.0`, `Apache-2.0 OR MIT`, `MIT` and `MIT OR Unlicense`.

What would be a good license string in Gentoo parlance ? "GPL-3 Apache-2.0 MIT" ? Or should `BSL` and `Unlicense` be included ? Note that AFAICT the gentoo repo only knows `BSL-1.1`, not `BSL-1.0`. Not sure if they're compatible.

app-portage/emlop/emlop-0.4.1.ebuild
Flowdalic: Yes, if people do not want them, then it is consensus that they should use `INSTALL_MASK`. It is the same rule of systemd unit files.

app-portage/emlop/emlop-0.4.1.ebuild
vincentdephily: I missed `(MIT OR Apache-2.0) AND Unicode-DFS-2016`, so the unicode one should be included as well. The latest cargo-ebuild with the current git emlop suggests `LICENSE="Apache-2.0 Boost-1.0 GPL-3 MIT Unicode-DFS-2016 Unlicense"`, I'll check that this is also valid for the 0.5.0 release and go with that.

app-portage/emlop/emlop-0.4.1.ebuild
vincentdephily: Indeed. I'll regenerate with latest cargo-ebuild.

vincentdephily:
Thanks for this, I'll update the PR during the week. The 0.6.0 release has been pending for months, I'll try to do that too.

vincentdephily:
Made the requested changes AFAICT. I'm currently having trouble signing the commit, I might need to register a new key. I still need to release/package version 0.6.

app-portage/emlop/emlop-0.5.0.ebuild
```diff
@@ -0,0 +1,107 @@
...
+	winapi-util-0.1.5
+	winapi-x86_64-pc-windows-gnu-0.4.0
+"
+
+inherit cargo bash-completion-r1
+
+DESCRIPTION="A fast, accurate, ergonomic emerge.log parser"
+HOMEPAGE="https://github.com/vincentdephily/emlop"
+SRC_URI="$(cargo_crate_uris)"
+RESTRICT="mirror"
```
juippis, line 76: This is probably not needed?

app-portage/emlop/emlop-0.5.0.ebuild
```diff
@@ -0,0 +1,107 @@
...
+
+LICENSE="GPL-3"
+# Dependent crate licenses
+LICENSE+="
+	   MIT Unicode-DFS-2016
+	   || ( Apache-2.0 Boost-1.0 )
+"
+
+SLOT="0"
+KEYWORDS="~amd64 ~arm64 ~x86"
```
juippis, line 86: Please only add `~amd64` and let the arch teams test it for the rest.

app-portage/emlop/emlop-0.5.0.ebuild
```diff
@@ -0,0 +1,107 @@
...
+
+DEPEND=">=virtual/rust-1.58.1"
```
juippis, line 88: I'd personally drop this version requirement as it's always going to be satisfied in ::gentoo repo. Cargo eclass will depend on rust so you could remove this line.
```suggestion
```

juippis (commented):
@Flowdalic ping :)

app-portage/emlop/emlop-0.5.0.ebuild
vincentdephily: It's not just about depending on rust, but which version of rust is required. Rust crates have their own MSRV (minimum supported rust version), which could be higher than what the eclass requires. See https://github.com/gentoo/cargo-ebuild/issues/34 for some discussion.

To be fair, in the case of emlop it's probably always going to support the oldest virtual/rust available in `::gentoo`, but I feel that being explicit is better. No strong opinion though, if other people want to remove that DEPEND I'm fine with it.

thesamesam:
Done & big thanks for the patience!

thesamesam:
@vincentdephily Are you planning on doing a PR for the new version? I promise it won't take anywhere near as long :)

thesamesam:
@vincentdephily No worries! FWIW, as a proxied maintainer, you don't actually need a gpg key or to sign your commits. The issue is that `pkgdev` didn't really understand this until recently. See https://github.com/pkgcore/pkgdev/issues/146 and https://github.com/pkgcore/pkgdev/issues/146.

(Until that ends up in a release, you may want to just edit `metadata/layout.conf` to disable signing.)

(You _do_ need to include `Signed-off-by` in commit messages but that's totally different.)

PR #18821, opened 2020-12-26
www-apps/grafana-bin: allow multiple service instances

mookie-:
With these changes it's easier to add additional instances of the grafana-bin service.

### DIFF
--- a/www-apps/grafana-bin/files/grafana.initd.3
+++ b/www-apps/grafana-bin/files/grafana.initd.3
@@ -2,6 +2,11 @@
 # Copyright 1999-2017 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
+user="${GRAFANA_USER:-grafana}"
+group="${GRAFANA_GROUP:-grafana}"
+datadir="${GRAFANA_DATADIR:-/var/lib/grafana}"
+logdir="${GRAFANA_LOGDIR:-/var/log/grafana}"
+
 depend() {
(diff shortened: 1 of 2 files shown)


www-apps/grafana-bin/files/grafana.initd.3
```diff
@@ -9,19 +9,21 @@ depend() {
 start() {
 	ebegin "Starting grafana server"
 
-    checkpath -d -o "${GRAFANA_USER}:${GRAFANA_GROUP}" -m750 "/var/lib/grafana/dashboards"
-    checkpath -d -o "${GRAFANA_USER}:${GRAFANA_GROUP}" -m750 "/var/lib/grafana/plugins"
-    checkpath -d -o "${GRAFANA_USER}:${GRAFANA_GROUP}" -m750 "/var/lib/grafana/sessions"
+    checkpath -d -o "${GRAFANA_USER}:${GRAFANA_GROUP}" -m750 "${GRAFANA_DATADIR}"
```
doctaweeks, line 12: In my opinion, grafana.initd should provide the original defaults for GRAFANA_DATADIR and GRAFANA_LOGDIR in case they are unset in grafana.confd. This would also allow them to be commented by default since changing them usually won't be necessary.

mookie-: @doctaweeks Thank you very much for your suggestion! I've added default values and commented the values in confd in a second commit.

PR #18832, opened 2020-12-27
x11-terms/guake: Add python 3.8 compat

Mord3rca:
Closes: https://bugs.gentoo.org/759529
Signed-off-by: Mord3rca <morderca@morderca.net>

### DIFF
--- a/x11-terms/guake/guake-3.7.0.ebuild
+++ b/x11-terms/guake/guake-3.7.0.ebuild
@@ -2,7 +2,7 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
-PYTHON_COMPAT=( python3_{6,7} )
+PYTHON_COMPAT=( python3_{6..8} )
 DISTUTILS_SINGLE_IMPL=1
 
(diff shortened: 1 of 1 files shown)


telans:
Why not bump it all the way to `python3_9`?

Mord3rca:
I've done this modification to use PYTHON_SINGLE_TARGET="python3_8"

So ... I did not tested the python3_9 compatibility, if it's safe, I can change that

telans:
You can check for any dependency issues with `repoman full -dx --if-modified=y`

thesamesam:
@telans Right, although testing with Python 3.9 is still important

Mord3rca:
So, I've installed gentoo on a VM with Gnome / XFCE

The installation with python 3.8 was fine. Nothing was wrong, everything worked out of the box.

However, for the python 3.9 targets .... A lot of packages block the installation of guake. (because they want python3.8)
+ depending on the graphical program used, blocking packages are different.

So, I won't change the compat to python3.9 for now. It seems risky. + This is supposed to be a small PR ^^"

PR #18835, opened 2020-12-27
sys-fs/tmsu: version bump 0.7.5

TheChymera:


juippis (approved):
Will add few missing `|| die`s.

PR #18845, opened 2020-12-27
dev-util/cmake: fix cmake bootstrap on Darwin

cognifloyd:
Two remaining cmake issues need to be fixed in ::gentoo for prefix bootstraps on darwin.

- Reintroduce the system-jsoncpp use flag
- Explicitly add the sdk framework search paths during cmake bootstrap

Bug: https://bugs.gentoo.org/757513

### DIFF
--- a/dev-util/cmake/cmake-3.19.2.ebuild
+++ b/dev-util/cmake/cmake-3.19.2.ebuild
@@ -78,6 +78,14 @@ cmake_src_bootstrap() {
 		-e '/"${cmake_bootstrap_dir}\/cmake"/s/^/#DONOTRUN /' \
 		bootstrap || die "sed failed"
 
+	# For Darwin prefix, system cmake does not have our prefix paths.
+	# So, ensure system frameworks are found during cmake bootstrap.
+	if [[ ${CHOST} == *-darwin* && -d "${EPREFIX}"/MacOSX.sdk ]] ; then
+		local sdk_frameworks="${EPREFIX}"/MacOSX.sdk/System/Library/Frameworks
+		append-cflags $(test-flags-CC "-F${sdk_frameworks}")
+		append-cxxflags $(test-flags-CXX "-F${sdk_frameworks}")
+	fi
+
 	# execinfo.h on Solaris isn't quite what it is on Darwin
 	if [[ ${CHOST} == *-solaris* ]] ; then
 		sed -i -e 's/execinfo\.h/blablabla.h/' \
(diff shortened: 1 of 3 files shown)


dev-util/cmake/cmake-3.19.2.ebuild
```diff
@@ -18,14 +18,15 @@ LICENSE="CMake"
 SLOT="0"
 [[ "${PV}" = *_rc* ]] || \
 KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~x86-macos ~sparc-solaris ~x64-solaris ~x86-solaris"
-IUSE="doc emacs ncurses qt5 test"
+# system-jsoncpp is only meant to be disabled during prefix bootstrap
+IUSE="doc emacs ncurses qt5 +system-jsoncpp test"
```
thesamesam, line 22: We should probably force it in profiles then?

cognifloyd: ok. I can add that to the profiles.
I believe that means I should force it on in `profiles/base/package.use.force` and then drop the force in `profiles/prefix/darwin/macos/package.use.force`, right?

cognifloyd: Hmm. If I add `dev-util/cmake -system-jsoncpp` to `darwin/macos/package.use.force` that would mean it would force it to be disabled even after bootstrap. Is there a way to force it on, and then just disable the forced useflag for a subsequent profile?

cognifloyd:
This PR should pave the way for dropping cmake-3.13.5 which has been hanging around for the darwin prefix.

dev-util/cmake/cmake-3.19.2.ebuild
cognifloyd: So, I don't think there's a good way to force this useflag in the profiles. I think the default `+system-jsoncpp` should be sufficient. If needed, we can add a warning to the useflag description saying it is for prefix bootstrap use only.

cognifloyd: K. I added the profile changes.

cognifloyd:
I think this is ready to merge :)

thesamesam:
FWIW, this seems fine, but I'd like the others' opinion while I find my feet still.

grobian:
I don't think we need this, for we now bootstrap cmake in stage1

PR #18848, opened 2020-12-28
dev-util/rocm-smi: New ebuild

justxi:


candrews:
I don't believe this actually uses `dev-libs/rocr-runtime` so I'm removing that dependency...

candrews:
* https://github.com/RadeonOpenCompute/ROC-smi says the command is `rocm-smi` so I changed to do that.
* since this is not a library, I used `python-single-r1` instead of `python-r1`, see https://dev.gentoo.org/~mgorny/python-guide/
* the python executable should be installed using `python_newscript`, not `doexe`, so the correct shebang is used. See https://dev.gentoo.org/~mgorny/python-guide/helper.html

justxi:
Ok, thanks for the hints.
I will keep them in mind for future PR.

PR #18850, opened 2020-12-28
sci-libs/libgeodecomp: Fix build with OpenCV enabled

shifty91:
Fix build with OpenCV enabled. Thanks!


sci-libs/libgeodecomp/files/libgeodecomp-0.5.0_pre20201009-opencv.patch
```diff
@@ -0,0 +1,186 @@
+From 3a20ca303ccc8d9d7cfd9787b392f1436d266243 Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: Actually, can we just rm the file to save the space? It's a large-ish patch which it _looks_ like we can avoid.

PR #18855, opened 2020-12-28
media-sound/lollypop: Bump to 1.4.7

Leo3418:
Add `ebuild` for `media-sound/lollypop` version 1.4.7.

This update requires two new dependencies:
- `dev-python/gst-python`
- `gui-libs/libhandy:1`

Because `gui-libs/libhandy:1` is masked on `amd64` as of now, I thought it would be appropriate to mask the new `ebuild` for Lollypop too, but I am not certain about this since this is my first ebuild contribution directly to Gentoo. Please feel free to unmask this `ebuild` if appropriate.

A patch that lets Lollypop use the Python interpreter corresponding to Portage's Python target version instead of `python3` under `PATH` is also required for Lollypop 1.3.1 and later versions. It is included in this pull request.


Leo3418:
I've just pushed a new commit signed with my GnuPG key, whose commit message was revised for consistency with older commits for this package, replacing the original one.

Leo3418:
Superseded by #19368

PR #18857, opened 2020-12-28
media-libs/freeimage Fixes Bug 734724

pyeatt:


pyeatt:
Well, that was not what I wanted to do.

chewi:
I don't know what happened there but you can force push to the same branch rather than create a new pull request.

PR #18866, opened 2020-12-29
media-gfx/prusaslicer: depend on boost subslot

doctaweeks:
Bug: https://bugs.gentoo.org/762430


expeditioneer:
Good catch!

Thank you for your contribution.

PR #18869, opened 2020-12-29
app-emulation/libpod: Add zsh and fish completion

pastalian:
Since v2.2.0, upstream supports fish completion. 
zsh has been supported for a while.

https://github.com/containers/podman/releases/tag/v2.2.0


zmedico:
Thanks!

PR #18879, opened 2020-12-30
Removing because of upcoming removal of dev-python/pycryptodomex.

misanthropos:


mgorny:
I've fixed it to work with pycryptodome.

misanthropos:
Very nice patch!
I tried to do the same by changing the code but the modules Crypto and pwquality would not have been found by meson.

mgorny:
Thanks. I'd prefer if we haven't had to do it but I just can't manage to get CFFI to work with symlinks or any other hacks making `Crypto` and `Cryptodome` modules equivalent.

PR #18883, opened 2020-12-30
media-video/obs-studio: add version 26.1.2

Chiitoo:
The ALSA patch feels at least a bit hacky, hence the conditionals on the release version, and the Python patch may cause issues where there were no issues before, so I only included it in the 9999 in case more people would like to test it out.


Chiitoo:
I've reduced this to a simple version bump with only patching up the crash as needed.

a17r (commented):
Thanks!

PR #18890, opened 2020-12-31
www-misc/buku: bump, cleanup

ajakk:
Diff of newest current buku and 4.5 below. Adds python3_9, SRC_URI switched to Github, dropped old patch, sorted ignored tests in appropriate sections, and ignored new bukuserver test (we don't install bukuserver because its deps aren't all packaged, and therefore do not run tests for it).

```
6c6
< PYTHON_COMPAT=( python3_{6,7,8} )
---
> PYTHON_COMPAT=( python3_{6,7,8,9} )
14c14
< SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
---
> SRC_URI="https://github.com/jarun/${PN}/archive/v$(ver_cut 1-2).tar.gz -> ${P}.tar.gz"
22,23d21
< PATCHES=( "${FILESDIR}/${P}-fix-tests.patch" )
<
65a64,66
>               tests/test_setup.py::test_bukuserver_requirement
>               tests/test_views.py::test_bmv_create_form
>               tests/test_views.py::test_bookmark_model_view
67d67
<               tests/test_views.py::test_tag_model_view_get_list_empty_db
69,70c69
<               tests/test_views.py::test_bookmark_model_view
<               tests/test_setup.py::test_bukuserver_requirement
---
>               tests/test_views.py::test_tag_model_view_get_list_empty_db
73,74d71
<               tests/test_bukuDb.py::test_load_firefox
<               tests/test_bukuDb.py::test_add_rec_exec_arg
77c74,75
<               tests/test_bukuDb.py::test_refreshdb
---
>               tests/test_bukuDb.py::test_add_rec_exec_arg
>               tests/test_bukuDb.py::test_load_firefox
78a77
>               tests/test_bukuDb.py::test_refreshdb
```


stkw0:
Why the switch of the SRC_URI?

ajakk:
Generally Github release tarballs are better than PyPi tarballs because they contain everything in the repository. There are several instances of PyPi tarballs not including unit tests, for example, so switching to a more "pure" tarball will be safer for the future in case some issue like this is ever introduced.

stkw0:
In the past I filled an issue for projects which released incomplete PyPi tarballs. I thought that the Gentoo policy was to prefer pypi mirror over direct links to Github tarballs, this way instead of having a single point of failure (Github) we can have a list of different mirrors of the package.  I will appreciate if some dev could clarify it as I didn't found any official statement about this.

thesamesam:
@stkw0 The [mirroring system](https://devmanual.gentoo.org/general-concepts/mirrors/index.html#automatic-mirroring) ensures that we're not vulnerable to a single point of failure.

GitHub tarballs are completely fine when it's needed for tests or other extra files - but in fact, I just go straight to GitHub to save time. It's even been considered (not enacted) to just make this an official recommendation.

TL;DR: It's fine to use GitHub.

juippis (approved):
Tests fail but the program seems to work.
```
E                               hypothesis.errors.FailedHealthCheck: tests/test_bukuDb.py::test_delete_rec_range_and_delay_commit uses the 'setup' fixture, which is reset between function calls but not between test cases generated by `@given(...)`.  You can change it to a module- or session-scoped fixture if it is safe to reuse; if not we recommend using a context manager inside your test function.  See https://docs.pytest.org/en/latest/fixture.html#sharing-test-data for details on fixture scope.
E                               See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.function_scoped_fixture to the suppress_health_check settings for this test.

/usr/lib/python3.7/site-packages/hypothesis/extra/pytestplugin.py:188: FailedHealthCheck
```

PR #18891, opened 2021-01-01
media-video/shotcut: bump to 21.01.29

xdch47:
Closes: https://bugs.gentoo.org/762826


media-video/shotcut/shotcut-20.11.28-r1.ebuild
```diff
@@ -41,11 +47,6 @@ RDEPEND="${COMMON_DEPEND}
 	virtual/jack
 "
 
-src_prepare() {
-	default
-	sed -i -e '/QT.*private/d' src/src.pro || die
```
a17r, line 46: This is not going to re-introduce the bogus warnings?

xdch47: Yes, but only for this ebuild,  since I brought it upstream. I thought it's nicer to align it with the live ebuild.

xdch47: I think there is no hurry, maybe it would be better to wait for the next release (end of january - https://github.com/mltframework/shotcut/milestone/65) and also avoid needless recompling to fix the bug [762826](https://bugs.gentoo.org/762826) due to the revbump.

juippis (approved):

PR #18893, opened 2021-01-01
dev-cpp/asio: 1.18.1

DarthGandalf:


dev-cpp/asio/asio-1.18.1.ebuild
```diff
@@ -0,0 +1,52 @@
...
+LICENSE="Boost-1.0"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~ppc ~ppc64 ~sparc ~x86"
+IUSE="doc examples libressl ssl test"
+RESTRICT="!test? ( test )"
+# test searches for libssl during ./configure, and REQUIRED_USE is easier than
+# patching configure to not search for it with USE=-ssl
+REQUIRED_USE="test? ( ssl )"
+
+RDEPEND="dev-libs/boost
```
SoapGentoo, line 19: so it's used an runtime, but doesnt link to boost? :thinking:

SoapGentoo (changes requested):

dev-cpp/asio/asio-1.18.1.ebuild
DarthGandalf: Yep!

It's a header-only library, but it #includes boost headers. If your program #includes asio, it will transitively include boost as well, even if you don't use boost directly, so you might not even be aware that you're using boost.

DarthGandalf: And without boost installed, such include would fail

SoapGentoo: in that case, I would still add `:=` to avoid ABI divergence

SoapGentoo (changes requested):

juippis (approved):

PR #18894, opened 2021-01-01
dev-libs/openssl[bindist] update per EC patents expiring

mgorny:


Whissi:
Thank you for your PR but @gentoo/base-system will handle OpenSSL on our own.

For the records, SM2 got disabled in last bump (2915b998b618e03e5c0fa120ae528be64209ea85) because the hobble patch is breaking SM2. However, Fedora (hobble patch upstream) don't care about SM2 because they have it disabled for political reasons (China!). Gentoo had to decide to fork and fix hobble patch or just disable SM2 when applying hobble patch allowing us to keep using upstream patch without any changes and we chose the latter to make maintenance easier.

floppym (approved):
The changes look good to me.

mattst88:
> Thank you for your PR but @gentoo/base-system will handle OpenSSL on our own.

This is really territorial. You should stop that.

floppym:
After reading Whissi's notes more carefully, I guess he is saying we don't need to keep "sm2" behind bindist when we drop the EC-related logic.

Whissi:
This wasn't meant territorial. But this PR is also the perfect example why people not maintaining a package shouldn't come up with a PR making a change they don't understand. It's a waste of time doing things twice. So I thanked the author and told him the project (probably me because I did most of the 1.1.x work in the past) will take care of this.

floppym:
> But this PR is also the perfect example why people not maintaining a package shouldn't come up with a PR making a change they don't understand.

This seems more like the perfect example of why you should document things like this.

mattst88:
So a week later, since you've asked others to stay off your lawn, can we expect you to mow it?

Whissi:
Thank you for your friendly reminder! I am sure you are following the bug behind this PR and have noticed no progress. So based on my little research we are still not ready to proceed here:

- USE=bindist was added around commit 214b83b4e311c017258d450aae81c04652d8bce1.
- Original commit said USE=bindist was added because of patents and mentioned 2015 as latest expiration date; This date was changed later to '??, ??, 2020' and it is not clear why it changed, if patents are really expired or if we need to adjust date again
- It's still not 100% clear what is covered by USE=bindist at all because additional ciphers/algorithms were added/removed to USE=bindist restriction over time without proper documentation
- So if we want to remove the restriction, only the foundation can give approval for this change as it was added because the foundation had legal liability concerns.

floppym:
> So if we want to remove the restriction, only the foundation can give approval for this change as it was added because the foundation had legal liability concerns.

I would really like to see evidence of this claim. By the "foundation", I assume you mean its trustees. Where/when did they express these concerns?

Whissi:
From commit bdd5c9e7d6a375e99b3ae89afd4517a3a5786df2:
> As resolved in the Foundation Trustees meeting 2017/10/22 [...]

Anyway, thank for your suggestion. I am now re-assigning the bug to get their approval to make progress.

dev-libs/openssl/openssl-1.1.1l-r1.ebuild
```diff
@@ -0,0 +1,290 @@
...
+	# Make sure user flags don't get added *yet* to avoid duplicated
+	# flags.
+	CFLAGS= LDFLAGS= echoit \
+	./${config} \
+		${sslout} \
+		$(use cpu_flags_x86_sse2 || echo "no-sse2") \
+		enable-camellia \
+		enable-ec \
+		enable-ec2m \
+		$(use_ssl !bindist sm2) \
```
floppym, line 176: We don't need to keep this. According to Whissi, sm2 is only disabled because the Fedora patches break it.

mgorny: Thanks! This means we can remove USE=bindist altogether. FWICS packages have already been updated to depend on `bindist(-)`, so we should be good here.

thesamesam (approved):
- Debian don't seem to be disabling things
- Fedora still are with their patch
- I've not seen any specific references to which patents we're supposedly protected from with USE=bindist, or whether it is truly exhaustive.
- (While not understanding doesn't mean it doesn't apply, I don't see how binary redistribution is different to the source code anyway.)

Unless the Gentoo Foundation gets specific legal advice, I really don't see the need to apply this to one package. **Especially** now that we ended up dropping it (and updating all revdeps...) for OpenSSL 3.x anyway, even though Fedora have a patch available now.

SoapGentoo (approved):

floppym (approved):

thesamesam:
Note that an ML discussion began [here](https://marc.info/?l=gentoo-dev&m=163215637525390&w=2) and Robin commented on the bindist situation [here](https://marc.info/?l=gentoo-dev&m=163216172229772&w=2).

thesamesam:
Robin did this as https://gitweb.gentoo.org/repo/gentoo.git/commit/dev-libs/openssl?id=c308ac7ab8756cef0d918e5aa0bc211308dc14a7.

thesamesam:
Feel free to reopen for the RC5 bit.

PR #18895, opened 2021-01-01
net-misc/openntpd: support dev-libs/libretls

mgorny:


pbhenson:
This looks good to me; thanks for taking care of it!

PR #15805, opened 2020-05-15
New Package: sndio, aucatctl, gst-plugins-sndio

lanodan:
Been using them on my machines for few weeks, pushed it on the guru repository but needs to be in `::gentoo` for support of other programs like ffmpeg and mpd, which I'll address later on (need to change my patches from stable ebuilds to live ones)

PR #18898, opened 2021-01-01
gui-apps/waybar: version bump to 0.9.5

mrabe89:
A version bump to 0.9.5. I also added a use flag for wifi support, which will enable rfkill features in waybar.

@JonasToth Please have a look at my pull request and forward it, if you see fit.


JonasToth (commented):
Thank you for doing this work!
I added my notes on sndio, that is newly supported by waybar. What do you think about this?

I am myself not a very active or experienced maintainer, but there seems to be missing sign-off. Please fix this. I did not find any other issue.

(Just in case: https://wiki.gentoo.org/wiki/GitHub_Pull_Requests and https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin)

mrabe89:
@JonasToth Thanks for the reply, I added the flag and dependency for sndio as you suggested.
It compiles with sndio enabled, but I can't say anything about the functionality since I don't use sndio.
The commit should be signed-off as well now.

gui-apps/waybar/waybar-0.9.5.ebuild
```diff
@@ -0,0 +1,66 @@
...
+	dev-libs/date:=
+	dev-libs/wayland
+	dev-libs/wayland-protocols
+	gui-libs/wlroots
+	x11-libs/gtk+:3[wayland]
+	mpd? ( media-libs/libmpdclient )
+	network? ( dev-libs/libnl:3 )
+	popups? ( gui-libs/gtk-layer-shell )
+	pulseaudio? ( media-sound/pulseaudio )
+	sndio? ( >=media-sound/sndio-1.7 )
```
JonasToth, line 44: this is currently not possible. `sndio` is not available in gentoo.
It lives in guru only. Maybe i was wrong with asking for this option _now_. Maybe it should be done in a later revision.
This would block your update for until this is somehow resolved (which is unclear it would be done).

JonasToth:
[please reassign]

JonasToth:
@mrabe89 You can remove the label for `no-sign-off`. I think the `sndio` should only be added if we are able to move that from GURU to gentoo. i have no idea if this is possible in a timely manner.
It probably better to do this in a follow up.

I would then try to assign the PR to the proxy maintainers who need to approve the changes before it can be merged.

mrabe89:
@JonasToth Ok, I reverted the sndio change. - Please forward it, if there isn't another problem.

JonasToth:
> @JonasToth Ok, I reverted the sndio change. - Please forward it, if there isn't another problem.

The PRs are addressed in chronological order by the gentoo team. This probably takes a while tho :/

thesamesam:
Yes, I really need to finally do https://github.com/gentoo/gentoo/pull/15805 too..

thesamesam:
Thanks!

PR #18899, opened 2021-01-01
dev-embedded/openocd: Add version 0.11.0_rc2

kripton:
As a result of https://bugs.gentoo.org/722640, dev-embedded/openocd compiles with gcc-10 (commit 45de522eb99a6a1cc619ffb1f69b748133da347b). However, the upstream developer that is set as maintainer in the package's `metadata.xml` asked us to package the current snapshot of 0.11.0 instead. This PR does just that.


paulfertser:
Hello,
I've checked this ebuild after changing to 0.11.0-rc2 (and removing the patch as it's included in -rc2 already).

The issue I noticed: some dependencies got enabled in automagic way:
```
Bitbang mode of FT232R based devices    yes (auto)
TI XDS110 Debug Probe                   yes (auto)
CMSIS-DAP v2 Compliant Debugger         yes (auto)
Nu-Link Programmer                      yes (auto)
Cypress KitProg Programmer              yes (auto)
```
The dependency on libjaylink doesn't have a version number and in fact at least 0.2.0 is needed.
I'm also not quite sure that even if all of the libusb dependencies are properly disabled the resulting openocd executable won't be linking to libusb shared library (even though it wouldn't need it for anything).
I runtime tested it with a jlink clone, seems to be doing the right thing.

dev-embedded/openocd/openocd-0.11.0_rc1.ebuild
```diff
@@ -0,0 +1,102 @@
...
+		virtual/libusb:0
+		virtual/libusb:1
+	)
+	ftdi? ( dev-embedded/libftdi:= )"
+
+DEPEND="${RDEPEND}"
+BDEPEND="virtual/pkgconfig"
+
+PATCHES=(
+	"${FILESDIR}"/${P}-configCapstone.patch
```
jsmolic, line 43: If we have only one patch, you squash the PATCHES array into a single line
i.e. `PATCHES=( "${FILESDIR}"/${P}-configCapstone.patch )`

dev-embedded/openocd/files/openocd-0.11.0_rc1-configCapstone.patch
```diff
@@ -0,0 +1,83 @@
+From 7ff53e77ee4f08dd8a70ac2be30d37d377b44a45 Mon Sep 17 00:00:00 2001
+From: Paul Fertser <fercerpav@gmail.com>
+Date: Wed, 23 Dec 2020 20:49:56 +0300
+Subject: [PATCH] configure: do not make Capstone dependency automagic
+
```
jsmolic, line 5: Is this patch upstreamed?

paulfertser:
> +From 7ff53e77ee4f08dd8a70ac2be30d37d377b44a45 Mon Sep 17 00:00:00 2001
> +From: Paul Fertser <fercerpav@gmail.com>
> +Date: Wed, 23 Dec 2020 20:49:56 +0300
> +Subject: [PATCH] configure: do not make Capstone dependency automagic
> +
> 
> Is this patch upstreamed?

Now yes, part of rc2, and we'll be tagging the final in the coming
weeks most likely.

kripton:
Hi, thanks for your comments. I've updated this PR to add 0.11.0_rc2 instead of 0.11.0_rc1. The patch has been dropped and the dependency to libjaylink now requires v0.2.0 or newer :)

thesamesam:
Sorry for the delay, thanks a bunch guys!

PR #18900, opened 2021-01-01
Qt 5.15.2: Various backports

a17r:
Although the most important bugfix would seem to be https://bugreports.qt.io/browse/QTBUG-88248 but it does not have a 5.15 branch pick yet.


Pesa (approved):
Thanks

PR #18914, opened 2021-01-02
app-emulation/cloud-init: 20.4: DEPENDS: Added pytest and pytest-cov.

jpds:
Closes: https://bugs.gentoo.org/762172


thesamesam:
Any chance of sedding out the pytest-cov stuff? coverage tests (like linters) aren't that useful in Gentoo, so if we can avoid the dep, it helps, but it's not critical if it's a huge pain to do.

holmanb:
This PR appears abandoned, I propose we close it?

This would need to be updated to include more recent releases and no response to a question in >1 year indicates we are unlikely to get one.

thesamesam:
Thanks!

PR #18917, opened 2021-01-02
media-tv/mythtv: Add check for /dev/null home directory

thebitpit:
Closes: https://bugs.gentoo.org/763132


thebitpit:
Fix bug by checking for /dev/null as home directory. Please review and commit.

mgorny:
This doesn't really fix the bug. The thing is, this shouldn't happen in the first place.

thebitpit:
> This doesn't really fix the bug. The thing is, this shouldn't happen in the first place.

I agree.  I do not know how the @Frank's mythtv user was first created. He said he was upgrading from an old version 29 of `mythtv`.  Earlier versions of `mythtv` work without a user home directory for the mythtv user. Some folks [myself included at one time] use the mythtv user for both the backend and frontend as is suggested in obsolete documentation. This fix is to warn old upgrades.

mgorny:
Yes but installing acct-user should have updated the home directory.

thebitpit:
I looked at the source code of the eclass functions involved in setting the home directory and didn't find anything.
The relevant code in the ebuild is:

```
                insinto $(egethome mythtv)
                newins "${FILESDIR}"/bash_profile .bash_profile
                newins "${FILESDIR}"/xinitrc-r1 .xinitrc

```

PR #18919, opened 2021-01-02
media-video/vdr: version bump to 2.4.6

madmartin:
Closes: https://bugs.gentoo.org/739774


media-video/vdr/vdr-2.4.4.ebuild
```diff
@@ -0,0 +1,203 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit eutils flag-o-matic multilib toolchain-funcs
+
```
madmartin, line 7: @jonasstein you asked if "eutils" is necessary - indeed not, I remove it

media-video/vdr/vdr-2.4.4.ebuild
```diff
@@ -0,0 +1,203 @@
...
+		INCDIR			= \$(PREFIX)/include
+
+		DEFINES			+= -DCONFDIR=\"\$(CONFDIR)\"
+		INCLUDES		+= -I\$(DVBDIR)
+
+		# >=vdr-1.7.36-r1; parameter only used for compiletime on vdr
+		# PLUGINLIBDIR (plugin Makefile old) = LIBDIR (plugin Makefile new)
+		LIBDIR			= ${PLUGIN_LIBDIR}
+		PCDIR			= /usr/$(get_libdir)/pkgconfig
+
```
madmartin, line 127: @jonasstein You advised I "should not create files with cat in the make file" - I understand. 
What would be the alternative way? Put a template in $FILESDIR and then do a bunch of "sed...." to fill the desired values in? There are 6 variables to replace, which I think is quite a lot...

ajakk: You wouldn't sed involving a file in FILESDIR. You would create a patch file and then apply the patch with eapply or via the PATCHES array.

madmartin: > You wouldn't sed involving a file in FILESDIR. You would create a patch file and then apply the patch with eapply or via the PATCHES array.
Hello @ajakk , I cant follow your proposal.
As you see in the text above, the file "Make.config" is created with input from some variables as a HERE document. Please take note that some of the $ signs are escaped (  \$ ) and some are not.
@jonasstein proposed to have the "Make.config" in FILESDIR, but then I have to 1) make a copy to WORKDIR 2) fiddle the variable content with "sed" into it. 
So now please explain what you want to say?!?

media-video/vdr/vdr-2.4.6.ebuild
```diff
@@ -0,0 +1,184 @@
...
+
+	einfo "\t Please set one of this values in your sytem make.conf"
+	einfo "\t LINGUAS=\"${LING_PO}\"\n"
+
+	if [[ -z ${LINGUAS} ]]; then
+		einfo "\n \t No values in LINGUAS="
+		einfo "\t You will get only english text on OSD \n"
+	fi
+
+	strip-linguas ${LING_PO} en
```
juippis, line 146: We are trying to get rid of these, just install them all unconditionally.

media-video/vdr/vdr-2.4.6.ebuild
```diff
@@ -0,0 +1,184 @@
...
+LICENSE="GPL-2+"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~x86"
+IUSE="bidi debug demoplugins html keyboard mainmenuhooks menuorg naludump pinplugin permashift systemd ttxtsubs verbose"
+
+COMMON_DEPEND="
+	>=media-libs/fontconfig-2.4.2
+	>=media-libs/freetype-2
+	sys-libs/libcap
+	virtual/jpeg:*
```
juippis, line 25: I see there's `fowners` call in `src_install`, so you need to add the proper acct-* packages here.

media-video/vdr/vdr-2.4.6.ebuild
```diff
@@ -0,0 +1,184 @@
...
+	emake VIDEODIR="/" DESTDIR="${ED}" install
+
+	keepdir "${PLUGIN_LIBDIR}"
+
+	# backup for plugins they don't be able to create this dir
+	keepdir "${CONF_DIR}/plugins"
+
+	if use html; then
+		local HTML_DOCS=( *.html )
+	fi
```
juippis, line 163: It doesn't look like you're building these separately. You should just install them unconditionally, not like html take a lot of space.

media-video/vdr/vdr-2.4.4.ebuild
```diff
@@ -0,0 +1,203 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit eutils flag-o-matic multilib toolchain-funcs
```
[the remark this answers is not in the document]
madmartin: @juippis yes, you're right, I hope you noticed that the PR now updates to vdr-2.4.6 - where eutils is no longer inherited.

media-video/vdr/vdr-2.4.4.ebuild
```diff
@@ -0,0 +1,203 @@
...
+	if use html; then
+		local HTML_DOCS=( *.html )
+	fi
+	local DOCS=( MANUAL INSTALL README* HISTORY CONTRIBUTORS UPDATE-2* )
+	einstalldocs
+
+	insinto /usr/share/vdr
+	doins "${CAP_FILE}"
+
+	fowners vdr:vdr "${CONF_DIR}" -R
```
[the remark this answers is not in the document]
madmartin: Hmm... I would like to add acct-*/vdr but.... this is a longer story.
I reserved the UID+GID for vdr long time ago through the dev-ML and prepared acct-*/vdr for it. vdr traditionally uses HOME=/var/vdr which is not FHS conform, but it is tradition. I had long discussions with the last vdr maintainter "hd_brummy" (who is absent again and doesnt even give a short "leave me alone" reply) about this years ago. But on his last "work phase" he already prepared the acct-user/vdr in his overlay with a different HOME=/var/lib/vdr  (you may look [here](https://gitweb.gentoo.org/proj/vdr/devel.git/tree/acct-user/vdr/vdr-0.ebuild) which is fine for me too. You see code reflecting the upcoming change in querying current $HOME of user vdr.

Unfortunately, hd_brummy invented a USE logic for group memberships  as a user convenience solution for some vdr plugins. I have not completed thinking about that, but I expect  that will need new revisions of the plugins which may need these extra groups.

Long story short, I definitely want acct-* packages, but I hope we can commit this one here without them.

Or - if you agree - I add simplified versions of acct-*/vdr into this version bump (in separate commits).
Let me know....

media-video/vdr/vdr-2.4.4.ebuild
```diff
@@ -0,0 +1,203 @@
...
+	emake VIDEODIR="/" DESTDIR="${ED}" install
+
+	keepdir "${PLUGIN_LIBDIR}"
+
+	# backup for plugins they don't be able to create this dir
+	keepdir "${CONF_DIR}/plugins"
+
+	if use html; then
+		local HTML_DOCS=( *.html )
+	fi
```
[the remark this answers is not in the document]
madmartin: @juippis yes, globbing works here, but the result is one 100k file "PLUGINS.html". maybe there have been more html files in earlier versions. 
If you agree we install this with USE=doc and remove the USE=html  - let me know
Let me mention that `euse -i` reports it is a local useflag and most packages say something like "install html documentation". So IMHO this is technically correct

media-video/vdr/vdr-2.4.6.ebuild
madmartin: **_Very_** interesting!

I like removing this and install **all** language files. BTW this policy can also apply when thinking about the single html documentation file in the previous comment.

media-video/vdr/vdr-2.4.6.ebuild
madmartin: see previous comment about acct-* packages

media-video/vdr/vdr-2.4.6.ebuild
madmartin: yes, see previous comment

media-video/vdr/vdr-2.4.4.ebuild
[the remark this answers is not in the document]
juippis: I'd really like to have them as separate commits in here, because it looks like vdr doesnt' launch unless I **manually** add the group and add vdr to video group. But then again I wouldn't wish to make this too huge with all those plugin revbumps. Can you show me an example of a current plugin ebuild that handles this? You maybe could be able to do bash magic in the acct ebuild, which sets `/var/vdr` or `/var/lib/vdr` depending if that directory exists in the live filesystem or not. Not something I'd encourage though.

media-video/vdr/vdr-2.4.4.ebuild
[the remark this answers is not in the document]
juippis: Yea just install it unconditionally.

media-video/vdr/vdr-2.4.4.ebuild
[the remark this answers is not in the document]
madmartin: As a prerequisite, you have to know that user+group "vdr" are created in media-tv/gentoo-vdr-scripts. media-video/vdr depends on media-tv/gentoo-vdr-scripts. So if you did not install media-tv/gentoo-vdr-scripts, it may come to problems. 

And, a known issue for me is that on systemd installations there are not all file+dir owners are set correctly. an updated media-tv/gentoo-vdr-scripts will fix this, but when I revbump media-tv/gentoo-vdr-scripts without acct-*/vdr then we have again the discussion with acct-*/vdr. So please ignore this for now.

IMHO when installing acct-user/something, a previously existing user and his setting is untouched. So IMHO we need no magic when current ~vdr is different from /var/lib/vdr. But all subseqent packages must respect ~vdr like `VIDEO_DIR="$(egethome vdr)/video"`. Agree?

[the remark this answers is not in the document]
juippis: So the plan would be to revbump all vdr-related packages at once, using the new acct- method? I'm fine with that, as long as it's resolved somehow (sorry this is very confusing trying to read). We can also write a warning to users utilizing `REPLACING_VERSION` variable in ebuild.

I'm fine with pushing this if you're up to resolving the vdr issue. But please remove the `html` USE flag and install it unconditionally.

madmartin:
@juippis thanks for reviewing, I close PR here because it has been overruled

PR #18920, opened 2021-01-02
dev-lang/zig: Drop to maintainer-needed@

mattst88:
Multiple open bugs on the package; maintainer has not committed in 15
months and has no bugzilla activity in a year.

Closes: https://bugs.gentoo.org/691582
Signed-off-by: Matt Turner <mattst88@gentoo.org>

@nrdmn


mgorny:
Actually, I'm pretty sure I had some contact with the maintainer for this. However, the package is a close last rite candidate due to LLD incompatibility. Apparently Zig upstream relies on using LLD libraries, while LLD upstream doesn't seem interested in actually exposing them properly, as they claim they're for LLD internal use only.

thesamesam:
It's a shame because it's an interesting packaging but if that's true then I don't think I have the motivation to look after it.

PR #18921, opened 2021-01-03
dist-kernel module rebuilding logic [WIP]

mgorny:
CC @gentoo/dist-kernel 

Note that I haven't tested it fully yet, waiting for the next kernel bump for that.

I'm wondering if I should name the flag/virtual `prebuilt-kernel` or `dist-kernel`.


mgorny:
Running last final check (locally) and then merging.

PR #18928, opened 2021-01-03
net-libs/libmicrohttpd: bump to 0.9.72

Karlson2k:
Added USE flag 'thread-names'
Added missing virtual/pkgconfig as build dependency.
Added suggestion of myself as proxied maintainer. I'm the upstream co-maintainer.
Added more configure parameters to avoid any "automatic" enable/disable.


blueness:
pushed.  i also check the tests and they work now.

Karlson2k:
Yep, forgot to mention that tests were enabled as they work fine.

PR #18930, opened 2021-01-03
sys-apps/smartmontools: Fix double "/" in ewarn

shuber2:
The variable ${EROOT} ends with "/", so building paths as ${EROOT}/
leads to double slash "//". Fix this.


Whissi:
Let's keep current code and better [migrate to EAPI 7](https://mgorny.pl/articles/the-ultimate-guide-to-eapi-7.html#d-ed-root-eroot-no-longer-have-a-trailing-slash).

PR #18932, opened 2021-01-03
Gedit 3.38 bump

mattst88:


mattst88:
gedit upstream removed the option of disabling introspection support, so the USE flag is now removed.

PR #18933, opened 2021-01-03
x11-libs/libva-intel-media-driver: added X USE flag

Karlson2k:
This patch fixes ambiguous usage of X11 in the package.
Before this patch the package tries to autodetect presence of X11 in the system and tries to use it if found. However, presence of X11 headers doesn't grantee that required packages (libva) were installed with X11 support.
Now result of build depends on USE flags instead of presence/absence of X11 headers on the build system.

Closes: https://bugs.gentoo.org/743625


thesamesam:
Thanks!

Karlson2k:
@thesamesam You are welcome. The more to come! :)

PR #18935, opened 2021-01-03
x11-misc/sddm: 0.19.0_p20210117 snapshot bump, GLEP-81

a17r:
- Drop l10n.eclass, user.eclass
- Drop obsolete consolekit quirks
- Set IUSE="elogind" as default
- Add IUSE="openrc-init" to install init.d/conf.d files for SDDM
  Without the inittab hack they are not compatible with openrc+sysvinit.
- Add some upstream pending, drop some upstreamed patches thx to openSUSE
- Check for CONFIG_DRM using linux-info.eclass
- Use tmpfiles.eclass
- Add pkg_postinst info for fixing entropy
- Fix some typos in metadata.xml
- Switch SDDM configuration to /etc/sddm.conf.d/
  - New Gentoo defaults file: 01gentoo.conf
  - For IUSE="elogind", fix path to /bin/loginctl
  - Move old /etc/sddm.conf to docdir and ewarn about config reset


x11-misc/sddm/files/sddm.initd
```diff
@@ -0,0 +1,107 @@
...
+    ebegin "Setting up SDDM"
+
+    if [ -n "${CHECKVT-y}" ] ; then
+        if vtstatic "${CHECKVT:-7}" ; then
+            if [ -x /sbin/telinit ] && [ "${SOFTLEVEL}" != "BOOT" ] && [ "${RC_SOFTLEVEL}" != "BOOT" ]; then
+                ewarn "Something is already defined on VT ${CHECKVT:-7}, will start X later"
+                telinit a >/dev/null 2>&1
+                return 0
+            else
+                eerror "Something is already defined on VT ${CHECKVT:-7}, not starting"
```
floppym, line 54: We are in an `ebegin` context, so maybe do this instead to end it.
```
eend $? "Something is already defined on VT ${CHECKVT:-7}, not starting"
return
```

x11-misc/sddm/sddm-0.19.0_p20201230.ebuild
```diff
@@ -0,0 +1,172 @@
...
+	systemd? ( sys-apps/systemd:= )
+	!systemd? ( sys-power/upower )
+"
+DEPEND="${COMMON_DEPEND}
+	test? ( dev-qt/qttest:5 )
+"
+RDEPEND="${COMMON_DEPEND}
+	acct-group/sddm
+	acct-user/sddm
+	!kernel-randomness? ( || (
```
floppym, line 47: I don't like this "kernel-randomness" USE flag. Pulling in entropy daemons as optional runtime deps is a violation of QA policy, and really just kind of weird. Please drop this.

x11-misc/sddm/sddm-0.19.0_p20201230.ebuild
```diff
@@ -0,0 +1,172 @@
...
+	fi
+}
+
+src_prepare() {
+	touch "${S}"/01gentoo.conf || die
+
+cat <<-EOF >> "${S}"/01gentoo.conf
+[General]
+# Halt/Reboot command
+HaltCommand=$(usex elogind "/bin/loginctl" "/usr/bin/systemctl") poweroff
```
floppym, line 81: `systemctl` might live in `/bin` depending on `USE=split-usr`.

Do these directives accept bare command names for a `PATH` lookup? If so, can you just do this instead?
```
HaltCommand=$(usex loginctl systemctl) poweroff
```

floppym (changes requested):

x11-misc/sddm/sddm-0.19.0_p20201230.ebuild
a17r: While I agree with that in principle, there's a real issue tried to be solved here, where without one of those daemons sddm will hang for some time (minus e.g. random keyboard presses) but the better part of ~50% systems will work fine. The only alternative then is to hard-depend on one-of entropy daemons.

x11-misc/sddm/files/sddm.initd
```diff
@@ -0,0 +1,107 @@
...
+            fi
+        fi
+    fi
+
+    # We need to source /etc/profile for stuff like $LANG to work, bug #10190.
+    . /etc/profile
+
+    start-stop-daemon --start --exec /usr/bin/sddm --pidfile /run/sddm.pid \
+        -m --background || eerror "ERROR: could not start SDDM"
+    eend 0
```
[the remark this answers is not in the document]
a17r: Thanks a lot for your init script review. Much of it comes from the xdm script in xorg-server, btw, this is essentially copied and slimmed-down, so could be worth revisiting over there too.

x11-misc/sddm/sddm-0.19.0_p20201230.ebuild
floppym: If you want to advertise to users that they need good entropy on boot, write a news item or something. Halfway-enforcing this via a dependency in a random X11 package is the wrong way to go about this.

jospezial:
This snapshot bump is also needed to compile against qt-5.15 git and qt-6 git .
```
/var/tmp/portage/x11-misc/sddm-0.19.0_pre/work/sddm-0.19.0/src/daemon/XorgDisplayServer.cpp:68:42: error: use of deleted function ‘QCharRef& QCharRef::operator=(char)’
   68 |             m_cookie[i] = digits[dis(gen)];
      |                                          ^
```


fixed in https://github.com/sddm/sddm/commit/e93bf95c54ad8c2a1604f8d7be05339164b19308

Chiitoo (approved):
Didn't get to testing yet, but aside from that, I can't see anything I'd change.

Thanks!

x11-misc/sddm/files/sddm.initd
```diff
@@ -0,0 +1,107 @@
...
+    # Bail out early if on a non-OpenRC system:
+    if [ ! -d /run/openrc ]; then
+        eerror "$0 should only be used on OpenRC systems"
+    fi
+
+    ebegin "Setting up SDDM"
+
+    if [ -n "${CHECKVT-y}" ] ; then
+        if vtstatic "${CHECKVT:-7}" ; then
+            if [ -x /sbin/telinit ] && [ "${SOFTLEVEL}" != "BOOT" ] && [ "${RC_SOFTLEVEL}" != "BOOT" ]; then
```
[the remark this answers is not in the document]
a17r: There is no indication it was ever in a different path, it is hardcoded to `/sbin/telinit` in multiple instances across portage. However, it seems `sys-apps/openrc[sysv-utils]` does not even provide it. So we should just check `/sbin/init` instead

x11-misc/sddm/files/sddm.initd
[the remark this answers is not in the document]
a17r: @mattst88 maybe you find some improvements here worthy to also change in `xdm.initd-11`.

x11-misc/sddm/sddm-0.19.0_p20201230.ebuild
a17r: A news item is insufficient. It has been a longstanding problem, and it will probably not go away soon, so unless we make that QA exception here we are down to `ewarn`/`elog` that no one reads (but at least we can say "told you so").

floppym: Depending on entropy daemons that few people will configure to start at boot does nothing to "solve" the issue either.

Also, the lack of an entropy daemon does not imply that the kernel will be entropy starved at boot. For example, I'm running an older AMD processor that does not support the RDRAND instructions, and the kernel is still able to gather sufficient entropy within a few seconds on boot (without an entropy daemon).

A news item on the subject in general would be much more effective I think.

x11-misc/sddm/files/sddm.initd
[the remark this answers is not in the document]
floppym: I guess I just find it a little weird to check for `/sbin/telinit` here, and then call `telinit` on line 51. The asymmetry bothers me, but I guess it's no big deal.

telinit is normally provided by sysvinit. I suspect openrc-init does not support sysv runlevels, and probably requires a different method of launching sddm entirely.

Also, I will note that sys-apps/systemd[sysv-utils] installs both `/sbin/init` and `/sbin/telinit`, and these commands behave very differently from each other. However, I doubt this init script will ever be called on such a system.

x11-misc/sddm/sddm-0.19.0_p20201230.ebuild
a17r: > Depending on entropy daemons that few people will configure to start at boot does nothing to "solve" the issue either.

That I can agree on, which is why it is covered by `pkg_postinst` message.

jospezial:
@a17r Do you want to update to
https://github.com/sddm/sddm/commit/bc5a18f34c0881929a6b2e5d3993971c4f692f4f
?

epsilon-0:
Is there a reason that sddm has its own init script?

a17r:
> Is there a reason that sddm has its own init script?

This one is for use only with openrc-init so it can drop the startDM hackery completely.

epsilon-0:
> > Is there a reason that sddm has its own init script?
> 
> This one is for use only with openrc-init so it can drop the startDM hackery completely.

I see. But the current one does work with openrc-init as well.
I don't have strong feelings, but imo this is a tiny bit superfluous. But it doesn't hurt anything so whatever is ok.

a17r:
> I see. But the current one does work with openrc-init as well.
> I don't have strong feelings, but imo this is a tiny bit superfluous. But it doesn't hurt anything so whatever is ok.

It is certainly not something we must have here, but the discussion around display-manager-init did get me wondering how a future with openrc-init could look like, of course how we could avoid users having multiple dm-inits enabled or if we even need to care about that.

epsilon-0:
We're actually in the process of cleaning up the startDM script and other things. So maybe wait and see with that?

x11-misc/sddm/files/sddm.initd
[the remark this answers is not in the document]
a17r: I mean, on the other hand, why check for `init` at all. It is reasonable to assume it will be in $PATH regardless of sysvinit or openrc-init.

jospezial:
I am using right now sddm-0.19.0_p20210117,acct-group/sddm-0,acct-user/sddm-0 from your sddm-glep81 branch.
Seems to work so far on openrc with gui-libs/display-manager-init-1.0-r2 and x11-base/xorg-server-9999.
I was surprised it is the first displaymanager that starts plasma (xwayland) correctly out of the box. Did not have so much luck with lightdm, others were not useable or to old. Now I can use opengl-3.1 rendering with 10bit HDR in plasma (xwayland) which does not work in normal plasma (black screen).
The layout country flag shows "us" until I type something into the password field. Then it shows "de" as it should.
emerge --info sddm only shows x11-misc/sddm-0.19.0_p20210117 and I have to type emerge --info acct-user/sddm to show that.
The same for "equery g sddm".

Please merge if you are ready.

epsilon-0:
This reminded me, someone was mentioning (in #gentoo-chat) that plasma/ssdm does some things with cgroups, maybe that is something which may have been causing said previous problems.

x11-misc/sddm/files/sddm.initd
```diff
@@ -0,0 +1,95 @@
...
+    else
+        return 1
+    fi
+
+    return
+}
+
+start_pre() {
+    # need eerror
+    . /lib/gentoo/functions.sh
```
Whissi, line 42: No! This is not needed.

x11-misc/sddm/files/sddm.initd
```diff
@@ -0,0 +1,95 @@
...
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License, v2
+
+command="/usr/bin/sddm"
+command_background="yes"
+pidfile="/run/sddm.pid"
+retry="TERM/5/TERM/5"
+
+depend() {
+    need localmount elogind
```
Whissi, line 11: This can cause problems. On my system I hit
```
# /etc/init.d/sddm restart
 * Starting elogind ...
 * start-stop-daemon: /lib64/elogind/elogind is already running                                                                 [ !! ]
 * ERROR: elogind failed to start
 * ERROR: cannot start sddm as elogind would not start
```
That's because I elogind was started manually by some other services...

Whissi:
I am also getting the following error now:
```
# /etc/init.d/sddm restart
 * Stopping sddm ...                                                                                                            [ ok ]
 * Starting sddm ...                                                                                                            [ ok ]
/lib/rc/sh/start-stop-daemon.sh: line 63: service_set_value: command not found
/lib/rc/sh/start-stop-daemon.sh: line 65: service_set_value: command not found
```
Not knowing where this is coming from at the moment.

a17r:
This, minus the last two commits, will likely be pushed to the ebuild repo soon, with minimal changes.

jospezial:
I have copied sddm-0.19.0_p20210418.ebuild to sddm-0.19.0_p20210609.ebuild .
sddm-0.19.0_p20210418-redesign-Xauth.patch has to be renamed or patches line changed in the ebuild.
All patches apply but with fuzz and offset.
Run tested with plasma on X11 and wayland.

Please update and merge.

acct-user/sddm/sddm-0.ebuild
```diff
@@ -0,0 +1,13 @@
+# Copyright 2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit acct-user
+
+DESCRIPTION="User for sddm"
+ACCT_USER_ID=219
+ACCT_USER_GROUPS=( sddm )
```
ConiKost, line 10: Should be ```video``` added as secondary group? Current ebuilds do that.

a17r: yes. only because of nvidia.

ionenwks: I did look into alternatives, but at this point I can safely say this requirement is not going away in a foreseeable future, so yes indeed.

x11-misc/sddm/sddm-0.19.0_p20210418.ebuild
```diff
@@ -0,0 +1,173 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
jospezial, line 4: Maybe you want to change to EAPI=8 and to a newer snapshot.
This is almost 4 months old.

a17r: EAPI-8 is not possible at this point, new snapshot will be much more experimental with recent changes.

jospezial:
Do you have any plans about
https://gitweb.gentoo.org/proj/qt.git/tree/x11-misc/sddm/sddm-9999.ebuild
That has not been updated from l10n.eclass to plocale.eclass yet. And l10n.eclass is gone.

jospezial:
> Do you have any plans about
> https://gitweb.gentoo.org/proj/qt.git/tree/x11-misc/sddm/sddm-9999.ebuild
> That has not been updated from l10n.eclass to plocale.eclass yet. And l10n.eclass is gone.

@Chiitoo did it.
https://github.com/gentoo/qt/commit/59171e27454155763cb77554dd80e6f5474b45a1

jospezial:
With this my ~/.local/share/sddm/wayland-session.log and ~/.local/share/sddm/xorg-session.log are created at login but are empty.
No .xsession-errors file.
Where can I find the logs of plasma-x11?

scardracs:
Hi :) any news?

a17r:
Not really. 0.19.0 release continues to be broken. Some long-term pending patches are still not merged upstream and no longer apply to git master.

Most changes in this PR have made it to 0.18.1-rX meanwhile. What will end up finally in ::gentoo will have to look at what other distributions are doing wrt patching.

tamiko:
@a17r Debian ships 0.19 with the following set of fairly minimal patches: https://sources.debian.org/patches/sddm/0.19.0-4/

jospezial:
@a17r Do you also think this PR can be closed or do you still want to use the init parts somehow?

Chiitoo:
Still relevant/up to date?

ConiKost:
Since 0.20.0 is in tree, I suspect we can close here?

PR #18938, opened 2021-01-04
media-gfx/iscan-2.30.4.2-r2: fix build due to `bool` mismatch

marcan:
Use <stdbool.h> instead of a custom bool type when available, to avoid
build failures if other packages include <stdbool.h> later, which
redefines it to _Bool. This causes build failures due to mismatch
between declarations and definitions.

Signed-off-by: Hector Martin <marcan@marcan.st>

### DIFF
--- a/media-gfx/iscan/iscan-2.30.4.2-r2.ebuild
+++ b/media-gfx/iscan/iscan-2.30.4.2-r2.ebuild
@@ -1,0 +1,153 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+MY_DOC="userg_revQ"
+inherit autotools desktop flag-o-matic
+
+# HINTS:
+# -> non-free modules are x86 and amd64 only
+# -> iscan frontend needs non-free modules
+# -> sane-epkowa should be usable on every arch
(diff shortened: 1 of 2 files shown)


media-gfx/iscan/iscan-2.30.4.2-r2.ebuild
```diff
@@ -0,0 +1,153 @@
...
+		https://dev.gentoo.org/~flameeyes/avasys/${MY_DOC}_e.pdf
+		l10n_ja? ( https://dev.gentoo.org/~flameeyes/avasys/${MY_DOC}_j.pdf )
+	)"
+
+LICENSE="GPL-2 AVASYS"
+SLOT="0"
+KEYWORDS="amd64 x86"
+IUSE="doc gimp l10n_ja nls X"
+
+REQUIRED_USE="gimp? ( X )"
```
a17r, line 35: Should `X` really be `gui` instead?

gringus: To be consistent with iscan-3* `X` should become `gui`.

gringus (commented):
Could you also add:
`Closes: https://bugs.gentoo.org/755884`
to commit comment.

a17r:
I'm going to change this commit to non-revbump as this is only a buildfix. No other changes necessary then.

Thanks for your contribution!

PR #18939, opened 2021-01-04
app-arch/xarchiver-0.5.4.16 version bump

kuraga:
app-arch/xarchiver-0.5.4.16 version bump: https://github.com/ib/xarchiver/releases/tag/0.5.4.16


mgorny:
So apparently this one is obsolete in favor of #19581.

PR #18941, opened 2021-01-04
Gnuradio fixes

doctaweeks:


thesamesam:
Thanks! Unrevbumped gnuradio.

doctaweeks:
> Thanks! Unrevbumped gnuradio.

I always include a revbump when a subslot dependency changes for those who use binary packages. If a binary package was built before the subslot dependency, then it will be selected over actually rebuilding the package.

PR #18943, opened 2021-01-04
media-video/libva-utils: fixed dependencies, minor other fixes

Karlson2k:
1. While it is declared that libva-utils are backward compatible with older version of libva, it is not truth. Let's require similar version of libva as dependency. Closes https://bugs.gentoo.org/743835
2. Xext, Xfixes and libdrm are not required anymore starting from version 2.7.1 of libva-utils. Dropped. See https://github.com/intel/libva-utils/releases/tag/2.7.1
3. Sometimes the upstream publishes tarballs with pre-generated 'configure'. Let's use it.


Karlson2k:
It looks like @aballier hasn't been active at all in the last two and half months. Can anybody else check/merge this PR?

thesamesam:
Thank you!

Karlson2k:
> Thank you! It may be worth dropping some of the logic re autoconf given it complicates the ebuilds, but we can leave that for now given how these have been broken for a while.
@thesamesam, I've published more PRs with similar autoreconf logic. Upstream publishes two types of releases: one is with pre-generated `configure`, another one without `configure` (basically just a copy of git tree).
It worth to support both types of releases as it gives us more flexibility (sometimes quick bug-fixes are published without `configure`)
Maybe just add more comments about this logic to make it easier to maintain future changes?

PR #18944, opened 2021-01-04
 sys-devel/elftoolchain: development tools and libraries for building ELF based program images

jsmolic:


dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
```diff
@@ -0,0 +1,62 @@
...
+		RANLIB="$(tc-getRANLIB)" \
+		-C libelf install || die
+
+	bmake \
+		DESTDIR="${D}" \
+		-C common install || die
+
+	find "${ED}" -name '*.la' -o -name '*.a' -delete || die
+
+	if [ "$(get_libdir)" = "lib64" ] ; then
```
thesamesam, line 59: Bash tests please! It's too hard to get things right with POSIX tests and we're guaranteed to have Bash anyway (see PMS).

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
```diff
@@ -0,0 +1,62 @@
...
+DESCRIPTION="Libraries/utilities to handle ELF objects (BSD drop in replacement for libelf)"
+HOMEPAGE="https://wiki.freebsd.org/LibElf"
+SRC_URI="https://netcologne.dl.sourceforge.net/project/elftoolchain/Sources/${P}/${P}.tar.bz2"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="
+	app-arch/libarchive
```
thesamesam, line 17: If you're linking against libraries and they expose a subslot, we need :=

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
```diff
@@ -0,0 +1,62 @@
...
+		-C libelf install || die
+
+	bmake \
+		DESTDIR="${D}" \
+		-C common install || die
+
+	find "${ED}" -name '*.la' -o -name '*.a' -delete || die
+
+	if [ "$(get_libdir)" = "lib64" ] ; then
+		mv "${D}"/usr/lib "${D}"/usr/$(get_libdir) || die
```
thesamesam, line 60: You probably need "${ED}" here

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
```diff
@@ -0,0 +1,62 @@
...
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="
+	app-arch/libarchive
+	dev-libs/expat
+	dev-libs/libbsd
+	!dev-libs/elfutils
+	!dev-libs/libelf"
+DEPEND="${RDEPEND}
+	virtual/yacc"
```
SoapGentoo, line 23: this is used on `CHOST`?

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
```diff
@@ -0,0 +1,62 @@
...
+src_prepare() {
+	default
+	rm -r test || die
+
+	sed -i -e "s@cc@$(tc-getCC)@" common/native-elf-format || die
+	sed -i -e "s@readelf@$(tc-getREADELF)@" common/native-elf-format || die
+}
+
+src_compile() {
+	export MAKESYSPATH="${EPREFIX}/usr/share/mk/bmake"
```
SoapGentoo, line 41: (not a blocker, but) does this rely on bmake'isms? :thinking:

SoapGentoo (changes requested):

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
jsmolic: Yeah, we do need to set the flag manually, at least I couldn't get it working without

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
jsmolic: Oh good point, this was a quick draft of the ebuild, I'll make small fixups, but I'm fairly certain we don't need yacc on CHOST

virtual/libelf/libelf-4.ebuild
```diff
@@ -0,0 +1,10 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DESCRIPTION="Virtual for libelf.so.1 provider dev-libs/elfoolchain"
+SLOT="0/1"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="dev-libs/elftoolchain"
```
thesamesam, line 10: Why is this the only provider now?

We should revbump the existing one and add || ( this ... existing ), right?

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
```diff
@@ -0,0 +1,62 @@
...
+DESCRIPTION="Libraries/utilities to handle ELF objects (BSD drop in replacement for libelf)"
+HOMEPAGE="https://wiki.freebsd.org/LibElf"
+SRC_URI="https://netcologne.dl.sourceforge.net/project/elftoolchain/Sources/${P}/${P}.tar.bz2"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="
+	app-arch/libarchive:*
```
thesamesam, line 17: Wait, really? (I may need to read up on the previous discussion)

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
```diff
@@ -0,0 +1,62 @@
...
+	!dev-libs/libelf"
+DEPEND="${RDEPEND}"
+BDEPEND="
+	dev-vcs/subversion
+	sys-apps/lsb-release
+	sys-devel/bmake
+	virtual/yacc"
+
+# Requires network access and otherwise fails
+RESTRICT="test"
```
thesamesam, line 30: Any chance of patching those out or pre-downloading the files, or too much work?

virtual/libelf/libelf-4.ebuild
jsmolic: Hm, `-2` has only libelf, and `-3` only elfutils so I was following that

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
```diff
@@ -0,0 +1,62 @@
...
+		RANLIB="$(tc-getRANLIB)" \
+		-C libelf install || die
+
+	bmake \
+		DESTDIR="${D}" \
+		-C common install || die
+
+	find "${ED}" -name '*.la' -o -name '*.a' -delete || die
+
+	if [[ "$(get_libdir)" = "lib64" ]] ; then
```
thesamesam, line 59: I think this is wrong for riscv which uses subdirs.

virtual/libelf/libelf-4.ebuild
jsmolic: Ah I found some discussion in https://bugs.gentoo.org/643802 about this

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
jsmolic: Actually libelf doesn't need libarchive, but I see it's needed for other toolchain binaries in elftoolchain, but since we're not using anything else I think we can skip building other stuff and drop libarchive completely. I'll look into it further

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
jsmolic: I just patched the tests out from main Makefile now, we still need `RESTRICT` clause, is that correct?

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
jsmolic: I don't have riscv hardware to test it, but I'll see if this can be patched in the source

jsmolic: Default install rule is broken, I think it's better to just do it manually

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
jsmolic: I've removed libarchive as we don't need it at all for libelf, other tools from the source use it

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
thesamesam: That completely depends on what happens when it tries to run the test phase. If it skips because there's no target, then no need for RESTRICT (although in some sense it's nice as a note to actually fix tests, I'd just add "# NOTE: Need to fix tests ...")

dev-libs/elftoolchain/elftoolchain-0.7.1.ebuild
thesamesam: Sure.

virtual/libelf/libelf-4.ebuild
thesamesam: I need to read that bug properly but need @SoapGentoo's input here too.

Alessandro-Barbieri:
Please close this https://bugs.gentoo.org/762364

virtual/libelf/libelf-4.ebuild
jsmolic: Yeah, I've removed the virtual commit for now, we can do it later as far as I'm concerned

jsmolic:
@Alessandro-Barbieri Sure, thanks! Please note that I've only picked libelf for now as I don't need other stuff, but feel free to join in in case you want to help out with maintenance or add new tools

sys-devel/elftoolchain/elftoolchain-0.7.1.ebuild
```diff
@@ -0,0 +1,55 @@
...
+	bmake \
+		AR="$(tc-getAR)" \
+		CC="$(tc-getCC)" \
+		LD="$(tc-getLD)" \
+		RANLIB="$(tc-getRANLIB)" \
+		"$@" || die
+}
+
+src_compile() {
+	export MAKESYSPATH="${EPREFIX}/usr/share/mk/bmake"
```
SoapGentoo, line 42: `EPREFIX` correct here? shouldn't this be `BROOT`?

sys-devel/elftoolchain/elftoolchain-0.7.1.ebuild
```diff
@@ -0,0 +1,55 @@
...
+	_bmake -C common
+	_bmake -C libelf
+}
+
+src_install() {
+	doheader common/elfdefinitions.h
+	doheader libelf/{gelf,libelf}.h
+
+	dolib.so libelf/libelf.so.1
+	dosym libelf.so.1 /usr/"$(get_libdir)"/libelf.so
```
SoapGentoo, line 52: quotes not necessary

SoapGentoo (changes requested):

sys-devel/elftoolchain/elftoolchain-0.7.1.ebuild
jsmolic: oops, yea it should be BROOT, fixed

sys-devel/elftoolchain/elftoolchain-0.7.1.ebuild
jsmolic: done

SoapGentoo (approved):

sys-devel/elftoolchain/elftoolchain-0.7.1.ebuild
```diff
@@ -0,0 +1,55 @@
...
+	virtual/yacc"
+
+src_prepare() {
+	default
+	sed -i -e "s@cc@$(tc-getCC)@" common/native-elf-format || die
+	sed -i -e "s@readelf@$(tc-getREADELF)@" common/native-elf-format || die
+}
+
+_bmake() {
+	bmake \
```
thesamesam, line 33: I wish we didn't have to do this -- exporting MAKE instead, but it is what it is for [now](778191)

thesamesam (approved):

PR #18948, opened 2021-01-05
x11-libs/libva: bump version, sync with libva-utils

Karlson2k:
* Added two new versions: 2.9.1 and 2.10.0
* .ebuild code is unified with libva-utils to allow builds for releases without prepared `configure` in tarball, like release 2.9.1
* Fixed DOCS installation (missing `default` in `multilib_src_install_all()`)


x11-libs/libva/libva-2.10.0.ebuild
```diff
@@ -0,0 +1,99 @@
...
+"
+
+REQUIRED_USE="|| ( drm wayland X )
+	opengl? ( X )"
+
+DOCS=( NEWS )
+${AUTOCONFIGURED} || DOCS+=( CONTRIBUTING.md README.md )
+
+MULTILIB_WRAPPED_HEADERS=(
+/usr/include/va/va_backend_glx.h
```
thesamesam, line 72: Indentation, maybe?

x11-libs/libva/libva-2.10.0.ebuild
```diff
@@ -0,0 +1,99 @@
...
+	video_cards_intel? ( >=x11-libs/libva-intel-driver-2.0.0[${MULTILIB_USEDEP}] )
+	video_cards_i965? ( >=x11-libs/libva-intel-driver-2.0.0[${MULTILIB_USEDEP}] )
+	utils? ( media-video/libva-utils )
+"
+
+REQUIRED_USE="|| ( drm wayland X )
+	opengl? ( X )"
+
+DOCS=( NEWS )
+${AUTOCONFIGURED} || DOCS+=( CONTRIBUTING.md README.md )
```
thesamesam, line 69: I'm still really wondering if we'd be better off just always re-generating autotools files if upstream aren't consistent?

thesamesam:
One commit per version please if possible?

x11-libs/libva/libva-2.10.0.ebuild
Karlson2k: Sure we can do it at expense of some flexibility.
* pre-generated `configure` is tested by upstream (but we can hardly expect any errors here)
* skipping of re-generation will speed-up build (but saving is mostly neglectable) 
* no need for autotools to be installed to build (however they are usually installed anyway)

Up to you @thesamesam. I can unify and always re-generate or leave it as is.

Karlson2k:
@thesamesam, please tell me what do you think is the best (always re-generate/conditionally re-generate) and I'll update the PR.

x11-libs/libva/libva-2.10.0.ebuild
thesamesam: Let's try regeneration all the time for now. Like you say, most of the time, these tools are present, and it's not a noticeable slow down mostly.

It'd be better if it was non-autotools, but it is what it is. Is that okay with you?

Karlson2k: OK, will update the PR.

Karlson2k:
@thesamesam, the PR has been updated.

Karlson2k:
Thanks for merging, @thesamesam.

When you have some free time, please take a look on libva-utils PR https://github.com/gentoo/gentoo/pull/18949. Not really urgent as libva-utils are forward-compatible with libva.

PR #18951, opened 2021-01-05
sci-libs/lrslib: Bump to 071

jsmolic:
Closes: https://bugs.gentoo.org/707612


sci-libs/lrslib/lrslib-071.ebuild
```diff
@@ -0,0 +1,51 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit toolchain-funcs eutils
```
SoapGentoo, line 6: where is `eutils` being used?

SoapGentoo (changes requested):

sci-libs/lrslib/lrslib-071.ebuild
jsmolic: Nowhere, sorry :)

PR #18963, opened 2021-01-06
www-client/qutebrowser: Fix build failure with python3.9

pastalian:
In python3.9, `dev-python/importlib_resources` is not needed because built-in module is used instead.

https://github.com/qutebrowser/qutebrowser/blob/master/README.asciidoc
> Requirements
> importlib_resources (on Python 3.8 or older)


steils:
Thanks.

PR #18965, opened 2021-01-06
sys-auth/fprintd: Fix build with gcc-10

jsmolic:
* Disable static-libs

Closes: https://bugs.gentoo.org/764077


sys-auth/fprintd/fprintd-0.8.1-r3.ebuild
```diff
@@ -13,25 +13,28 @@ SRC_URI="https://cgit.freedesktop.org/libfprint/${PN}/snapshot/${MY_PV}.tar.bz2
...
-"
+	doc? (
+		dev-libs/libxml2
+		dev-libs/libxslt
+	)"
 
-S=${WORKDIR}/${MY_PV}
+PATCHES=( "${FILESDIR}"/${P}-fno-common.patch )
+
+S="${WORKDIR}"/${MY_PV}
```
SoapGentoo, line 37: swap these, PATCHES definition should always come after $S

SoapGentoo (changes requested):

sys-auth/fprintd/fprintd-0.8.1-r3.ebuild
jsmolic: Oh, that's right. Done

PR #18967, opened 2021-01-06
Fix building with gcc-10

jsmolic:


net-ftp/pureadmin/files/pureadmin-0.4-gold.patch
```diff
@@ -13,8 +7,6 @@ index 94f61e0..0ebb561 100644
 -pureadmin_LDADD = @GTK_LIBS@ $(INTLLIBS) @BINRELOC_LIBS@
 +pureadmin_LDADD = @GTK_LIBS@ $(INTLLIBS) @BINRELOC_LIBS@ -lm -lX11
```
SoapGentoo, line 8: no `{R,}DEPEND` on libX11?

SoapGentoo (changes requested):

net-ftp/pureadmin/files/pureadmin-0.4-gold.patch
jsmolic: Oh, good catch, I guess that was forgotten by whoever added the patch. Done, and also revbumped

PR #18969, opened 2021-01-06
media-plugins/vdr-xineliboutput: fix QA notice

madmartin:
fix "QA Notice: shell script appears to use non-POSIX feature(s):"
possible bashism in /etc/init.d/vdr-frontend
- line 7 (should be '.', not 'source')

### DIFF
--- a/media-plugins/vdr-xineliboutput/metadata.xml
+++ b/media-plugins/vdr-xineliboutput/metadata.xml
@@ -1,6 +1,18 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
 <pkgmetadata>
+	<maintainer type="person">
+		<email>hd_brummy@gentoo.org</email>
+		<name>Joerg Bornkessel</name>
+	</maintainer>
+	<maintainer type="person">
+		<email>martin.dummer@gmx.net</email>
+		<name>Martin Dummer</name>
+	</maintainer>
+	<maintainer type="project">
+		<email>proxy-maint@gentoo.org</email>
+		<name>Proxy Maintainers</name>
+	</maintainer>
 	<maintainer type="project">
 		<email>vdr@gentoo.org</email>
 		<name>Gentoo VDR Project</name>
--- a/media-plugins/vdr-xineliboutput/files/vdr-frontend
+++ b/media-plugins/vdr-xineliboutput/files/vdr-frontend
@@ -1,10 +1,10 @@
 #!/sbin/openrc-run
-# Copyright 1999-2015 Gentoo Foundation
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 description="Start xineliboutput sxfe/sxfb remote frontend"
 
-source /etc/conf.d/vdr.xineliboutput
+. /etc/conf.d/vdr.xineliboutput
 
 command="/usr/bin/vdr-sxfe"
 command_args="${REMOTE_FRONTEND}"


ConiKost (changes requested):
Please bump ebuild and file to a new revision. You can directly drop the old one after that in a new commit.

media-plugins/vdr-xineliboutput/metadata.xml
```diff
@@ -1,6 +1,18 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
 <pkgmetadata>
+	<maintainer type="person">
+		<email>hd_brummy@gentoo.org</email>
+		<name>Joerg Bornkessel</name>
```
ConiKost, line 6: @hd-brummy Are you fine with adding yourself as primary maint?

madmartin:
> Please bump ebuild and file to a new revision. You can directly drop the old one after that in a new commit.

@ConiKost so you mean 2 commits in one (this) PR?

media-plugins/vdr-xineliboutput/metadata.xml
madmartin: @ConiKost I just copied maintainer order from media-video/vdr where metadata.xml has been written by @hd-brummy  itselt.  So I assume he's fine with that here, too.

hd-brummy: This make no sense to add me, as the only member in the Gentoo VDR Project its me. All bugs around  to vdr@g.o reach me direct.
A had some request in the past, why iam and vdr@g.o is listed in some metadata.xml.  ;)

ConiKost:
> @ConiKost so you mean 2 commits in one (this) PR?

Yes.

media-plugins/vdr-xineliboutput/metadata.xml
ConiKost: Thanks. So please drop hd_brummy and leave only the project :-)

ConiKost:
> @ConiKost so you mean 2 commits in one (this) PR?

Basically yes. This is fine, if you have multiple commits in one PR, as long they make sense :-)

hd-brummy:
> Please bump ebuild and file to a new revision. You can directly drop the old one after that in a new commit.
we could this fix on the fly on next vdr*-r* update, let me explain why:
on vdr update, the user is forced to rebuild all enabled plugins.
in this case he get always any changes, no matter which plugin version he has installed.
on your comment, all existing version (the are 5) has to be with a newer -rx version rebumped. Makes no sense for me, as we have here a easier way. The result is finaly the same, the user get this changes on the existing versions.

ConiKost:
> we could this fix on the fly on next vdr*-r* update, let me explain why:
> on vdr update, the user is forced to rebuild all enabled plugins.
> in this case he get always any changes, no matter which plugin version he has installed.
> on your comment, all existing version (the are 5) has to be with a newer -rx version rebumped. Makes no sense for me, > as we have here a easier way. The result is finaly the same, the user get this changes on the existing versions.

Policy says, if installed files are changed, you should revbump. But I see, what you are up-to. Does the vdr update trigger a rebuild of all modules?

hd-brummy:
> > we could this fix on the fly on next vdr*-r* update, let me explain why:
> > on vdr update, the user is forced to rebuild all enabled plugins.
> > in this case he get always any changes, no matter which plugin version he has installed.
> > on your comment, all existing version (the are 5) has to be with a newer -rx version rebumped. Makes no sense for me, > as we have here a easier way. The result is finaly the same, the user get this changes on the existing versions.
> 
> Policy says, if installed files are changed, you should revbump. But I see, what you are up-to. Does the vdr update trigger a rebuild of all modules?

vdrplugin-rebuild is always needed after any changes in vdr update, this reflects also vdr -rx version updates or the plugins will skip on next vdr start.

hd-brummy:
i am commited martins suggestion bundeld with an update to vdr-2.4.6-r1, as i suggested
yes, i am ignored here the policy in this case, shame on me ;)
anybody knows, how mark the PR as obsolet?

PR #18978, opened 2021-01-07
Qt 5.15.2: Two more backports

a17r:
Both are also carried in major distributions, and esp. w/ dev-qt/qtwayland I don't feel bad about including it already in upcoming stabilisation.


Pesa (approved):
sounds good, thanks!

PR #18979, opened 2021-01-07
dev-db/freetds: security bump to 1.2.18

ajakk:
Drop dependencies which are in the system set, drop libressl support as
the package is on its way out and it already seems to be broken on this
package (bug 685086), drop static-libs use flag (no consumers).

Bug: https://bugs.gentoo.org/718950
Closes: https://bugs.gentoo.org/685086
Closes: https://bugs.gentoo.org/747265

```
freetds $ diff freetds-1.*
1c1
< # Copyright 1999-2020 Gentoo Authors
---
> # Copyright 1999-2021 Gentoo Authors
16c16
< KEYWORDS="~alpha amd64 arm arm64 ~hppa ~ia64 ~mips ppc ppc64 s390 sparc x86 ~ppc-macos"
---
> KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~mips ~ppc ~ppc64 ~s390 ~sparc ~x86 ~ppc-macos"
26,30c26
< # sed, grep, and awk are used by the build system and the osql script.
< COMMON_DEPEND="
<       sys-apps/sed
<       sys-apps/grep
<       virtual/awk
---
> COMMON_DEPEND="virtual/awk
36,40c32
<       ssl? (
<               !libressl? ( dev-libs/openssl:0= )
<               libressl? ( dev-libs/libressl:0= )
<       )
< "
---
>       ssl? ( dev-libs/openssl:0= )"
44d35
< # binutils is for "strings".
46,48c37
<       sys-devel/binutils
<       net-dns/bind-tools
< "
---
>       net-dns/bind-tools"
54a44
>               --disable-static \
59d48
<               $(use_enable static-libs static) \
70,72c59
<       if ! use static-libs; then
<               find "${D}" -name '*.la' -delete || die
<       fi
---
>       find "${D}" -name '*.la' -delete || die
```


thesamesam:
- Dropped static-libs from IUSE
- Patch wasn't being applied?
- Tests still don't pass :(

PR #18981, opened 2021-01-07
net-misc/x11-ssh-askpass: wrt 725200

SpiderX:
Closes: https://bugs.gentoo.org/725200

Signed-off-by: Vladimir Pavljuchenkov <spiderx@spiderx.dp.ua>

### DIFF
--- a/net-misc/x11-ssh-askpass/x11-ssh-askpass-1.2.4.1-r3.ebuild
+++ b/net-misc/x11-ssh-askpass/x11-ssh-askpass-1.2.4.1-r3.ebuild
@@ -1,0 +1,51 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit toolchain-funcs
+
+DESCRIPTION="X11-based passphrase dialog for use with OpenSSH"
+HOMEPAGE="https://github.com/sigmavirus24/x11-ssh-askpass"
+SRC_URI="https://github.com/sigmavirus24/${PN}/archive/${PV}.tar.gz -> ${P}.gh.tar.gz"
+
+LICENSE="HPND"
+SLOT="0"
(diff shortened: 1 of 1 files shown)


net-misc/x11-ssh-askpass/x11-ssh-askpass-1.2.4.1-r3.ebuild
```diff
@@ -0,0 +1,51 @@
...
+	x11-misc/imake"
+
+src_prepare() {
+	default
+
+	# wrt #725200
+	printf '#!/bin/sh\n%s ${*}\n' "$(tc-getLD)" > "${T}"/ld \
+		|| die "printf failed"
+	chmod +x "${T}"/ld || die "chmod failed"
+	export PATH="${T}:${PATH}" || die "export failed"
```
juippis, line 33: Yeeaahh, this seems weird. This is usually solvable with a single `sed` line, but rather just patch it then if not.

ionenwks: I've seen this in a few ebuilds and it needs to stop (don't copy these).

If user happens to set LD=ld you have an infinite self-call, like [bug 764506](https://bugs.gentoo.org/764506) with CC=gcc

That aside, you also shouldn't fix this at all. The problem is with xmkmf, not this package. See [bug 729630#c7](https://bugs.gentoo.org/729630#c7)

ionenwks: > That aside, you also shouldn't fix this at all. The problem is with xmkmf, not this package.

I might add that while it'll fail trying to find ld/gcc during xmkmf (if -native-symlinks), during building it's actually using the proper tools. I can even build this with clang and ld.lld

juippis:
Yep, it should be fixed in imake. Resolved https://bugs.gentoo.org/725200 marking it as duplicate for https://bugs.gentoo.org/729630

PR #18985, opened 2021-01-07
net-misc/modemmanager: 1.14.8 version bump

candrews:
* net-libs/libmbim: 1.24.4 version bump
* net-libs/libqmi: 1.26.6 version bump
* net-misc/modemmanager: 1.14.8 version bump


net-misc/modemmanager/modemmanager-1.14.8.ebuild
```diff
@@ -0,0 +1,130 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
```
dm0-, line 4: There is also an EAPI 7 update for this in #18813.

candrews: Thanks! I've incorporated the changes in https://github.com/gentoo/gentoo/pull/18813/commits/b941b20490fe11a27e41b6d5dfd504bd6767cf6b

mattst88:
I fixed a handful of things here: dropping IUSE=static-libs, disabling -Werror, etc.

candrews:
Thank you!

PR #18988, opened 2021-01-08
Start cross-compiling Xfce

dm0-:
@mgorny Here are some fixes to support cross-compiling core Xfce packages.

### DIFF
--- a/xfce-base/exo/exo-4.16.0.ebuild
+++ b/xfce-base/exo/exo-4.16.0.ebuild
@@ -1,9 +1,9 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-inherit xdg-utils
+inherit autotools xdg-utils
 
 DESCRIPTION="Extensions, widgets and framework library with session support for Xfce"
 HOMEPAGE="https://www.xfce.org/projects/"
@@ -21,8 +21,18 @@ DEPEND="${RDEPEND}"
 BDEPEND="
 	dev-util/gtk-doc-am
 	dev-util/intltool
+	dev-util/xfce4-dev-tools
 	sys-devel/gettext
 	virtual/pkgconfig"
+
+PATCHES=(
+	"${FILESDIR}/${PN}-4.16.0-allow-cross.patch"
+)
+
+src_prepare() {
+	default
+	eautoreconf
+}
 
 src_install() {
 	default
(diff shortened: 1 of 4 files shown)


xfce-base/exo/files/exo-4.16.0-allow-cross.patch
```diff
@@ -0,0 +1,10 @@
+--- a/configure.ac
++++ b/configure.ac
+@@ -151,6 +151,7 @@
+   }
+ ], [
+   AC_DEFINE([HAVE_STRFTIME_EXTENSION], 1, [Define if strftime supports %E and %O modifiers.])
++],,[AC_MSG_WARN([skipping strftime test due to cross-compiling])
```
mgorny, line 7: 1. Please send this patch upstream and gimme the link.
2. I think you should default to `1`. According to `strftime(3)` both `%E` and `%O` are SUS things, so most systems should really have them. Ideally, exo would test this at runtime and run appropriate code.

xfce-base/xfwm4/xfwm4-4.16.0.ebuild
```diff
@@ -29,6 +29,7 @@ RDEPEND=">=dev-libs/glib-2.20
 	>=xfce-base/xfconf-4.13:=
 	opengl? ( media-libs/libepoxy:=[X(+)] )
 	startup-notification? ( x11-libs/startup-notification )
+	xi? ( x11-libs/libXi )
```
mgorny, line 32: Is there actually much point in having it conditional? Maybe enable it unconditionally. It seems that a lot of packages require it on my system. The same probably goes for xcomposite.

mgorny (changes requested):

xfce-base/xfwm4/xfwm4-4.16.0.ebuild
dm0-: No, I just did it to be consistent with the other options.  Updated.

xfce-base/exo/files/exo-4.16.0-allow-cross.patch
dm0-: I opened https://gitlab.xfce.org/xfce/exo/-/issues/52 and changed the patch to default-enable for glibc.

xfce-base/exo/exo-4.16.0.ebuild
```diff
@@ -21,9 +21,27 @@ DEPEND="${RDEPEND}"
...
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	local -a myeconfargs
+	tc-is-cross-compiler && myeconfargs+=(
+		ac_cv_have_strftime_extension=$(usex elibc_glibc)
```
mgorny, line 40: Make it yes. We can revisit when we actually find a system that doesn't support it.

dm0-: Okay, I just dropped `src_configure` since it defaults to being enabled when cross-compiling.

mgorny (approved):
Thank you. Lemme just do a test run and then I'll merge.

PR #18991, opened 2021-01-08
net-misc/dahdi-tools:  general bug fixes.

jkroonza:
Build should not build with -Werror
Build should not install .a files unless USE=static-libs

Closes: https://bugs.gentoo.org/751853
Closes: https://bugs.gentoo.org/751850

### DIFF
--- a/net-misc/dahdi-tools/dahdi-tools-3.1.0-r1.ebuild
+++ b/net-misc/dahdi-tools/dahdi-tools-3.1.0-r1.ebuild
@@ -1,0 +1,64 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools bash-completion-r1 perl-functions
+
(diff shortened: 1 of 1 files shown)


net-misc/dahdi-tools/dahdi-tools-3.1.0-r1.ebuild
```diff
@@ -0,0 +1,64 @@
...
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	econf $(use_with ppp) --with-perllib="$(perl_get_vendorlib)"
+	sed -re 's/ -Werror($|[[:space:]])//' -i xpp/oct612x/Makefile.in || die "sed to eliminate -Werror failed."
+	sed -re '/[[:space:]]*-Werror[[:space:]]*\\$/ d' -i xpp/xtalk/Makefile || die "sed to eliminate -Werror failed."
```
juippis, line 43: Shouldn't these be in `src_prepare`?

juippis (approved):
Not like I can test this when dahdi doesn't compile but looks good :)

Will update my containers before your next PRs about dahdi.

PR #18992, opened 2021-01-08
net-vpn/i2pd: 2.35.0 version bump

khumarahn:
Hi!

This is a regular version bump. Compared to 2.34.0, I added a patch for AVX detection, see https://github.com/PurpleI2P/i2pd/issues/1602 and removed the `WITH_AVX` cmake option. The rest is without change.

I was going to add a proper `src_test` stage function, but found that tests do not work out of the box and need more attention.


thesamesam:
Glad you didn't forget about tests! ;)

Thank you!

PR #18993, opened 2021-01-08
sys-devel/bmake: Bump to 20201222

jsmolic:


sys-devel/bmake/bmake-20201222.ebuild
```diff
@@ -0,0 +1,50 @@
...
+	http://void.crufty.net/ftp/pub/sjg/mk-${MK_VER}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc ~x86"
+
+S="${WORKDIR}/${PN}"
+
+# Skip failing test (sandbox and csh)
+PATCHES=( "${FILESDIR}"/${P}-tests.patch )
```
SoapGentoo, line 21: have you checked whether these are still needed?

SoapGentoo (approved):

sys-devel/bmake/bmake-20201222.ebuild
jsmolic: Yeah, couple of tests still fail

PR #18994, opened 2021-01-08
Asterisk updates

jkroonza:


net-misc/asterisk/asterisk-13.38.1-r2.ebuild
```diff
@@ -0,0 +1,320 @@
...
+SLOT="0/${PV%%.*}"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~x86"
+
+IUSE_VOICEMAIL_STORAGE="
+	+voicemail_storage_file
+	voicemail_storage_odbc
+	voicemail_storage_imap
+"
+IUSE="${IUSE_VOICEMAIL_STORAGE} alsa blocks bluetooth calendar +caps cluster curl dahdi debug doc freetds gtalk http iconv ilbc ldap libressl lua mysql newt odbc oss pjproject portaudio postgres radius selinux snmp span speex srtp +ssl static statsd syslog vorbis xmpp"
+IUSE_EXPAND="VOICEMAIL_STORAGE"
```
mgorny, line 24: What's this? It doesn't seem to be used anywhere.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
```diff
@@ -0,0 +1,320 @@
...
+
+DESCRIPTION="Asterisk: A Modular Open Source PBX System"
+HOMEPAGE="https://www.asterisk.org/"
+SRC_URI="https://downloads.asterisk.org/pub/telephony/asterisk/releases/${P}.tar.gz
+	https://downloads.uls.co.za/gentoo/asterisk/gentoo-asterisk-patchset-4.08.tar.bz2"
+LICENSE="GPL-2"
+SLOT="0/${PV%%.*}"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~x86"
+
+IUSE_VOICEMAIL_STORAGE="
```
mgorny, line 18: We generally prefer arrays for helper vars like this. This makes further operations on them a bit cleaner.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
```diff
@@ -0,0 +1,320 @@
...
+	net-misc/asterisk-moh-opsound
+	selinux? ( sec-policy/selinux-asterisk )
+	syslog? ( virtual/logger )"
+
+BDEPEND="virtual/pkgconfig"
+
+QA_DT_NEEDED="/usr/lib.*/libasteriskssl[.]so[.][0-9]\+"
+
+pkg_pretend() {
+	[ "${MERGE_TYPE}" = "binary" ] && return
```
mgorny, line 109: Use `[[ ... ]]` in ebuilds, and then drop unnecessary quoting, and use `==`.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
```diff
@@ -0,0 +1,320 @@
...
+
+	if tc-is-clang; then
+		! use blocks && die "CC=clang requires USE=blocks"
```
mgorny, line 112:
```suggestion
		use blocks || die "CC=clang requires USE=blocks"
```

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
```diff
@@ -0,0 +1,320 @@
...
+		use blocks && die "USE=blocks can only be used with CC=clang"
+	fi
+}
+
+pkg_setup() {
+	CONFIG_CHECK="~!NF_CONNTRACK_SIP"
+	local WARNING_NF_CONNTRACK_SIP="SIP (NAT) connection tracking is enabled. Some users
+	have reported that this module dropped critical SIP packets in their deployments. You
+	may want to disable it if you see such problems."
+	check_extra_config
```
mgorny, line 123: Is there a reason this isn't done in `pkg_pretend` as well?

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
```diff
@@ -0,0 +1,320 @@
...
+}
+
+src_compile() {
+	emake ASTCFLAGS="${CFLAGS}" ASTLDFLAGS="${LDFLAGS}" NOISY_BUILD=yes
+}
+
+src_install() {
+	local d
+
+	mkdir -p "${ED}/usr/$(get_libdir)/pkgconfig" || die
```
mgorny, line 255: Any reason not to:

```suggestion
	dodir "/usr/$(get_libdir)/pkgconfig"
```

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
```diff
@@ -0,0 +1,320 @@
...
+	emake DESTDIR="${ED}" NOISY_BUILD=yes install
```
mgorny, line 256: Are you sure this doesn't cause duplicate EPREFIX? `econf` should already be injecting EPREFIX.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
```diff
@@ -0,0 +1,320 @@
...
+
+	if use radius; then
+		insinto /etc/radiusclient/
+		doins contrib/dictionary.digium
+	fi
+	diropts -m 0750 -o root -g asterisk
+	keepdir	/etc/asterisk
+	emake NOISY_BUILD=yes DESTDIR="${ED}" CONFIG_SRC=configs/samples CONFIG_EXTEN=.sample install-configs
+	for conffile in "${ED}/etc/asterisk/"*
```
mgorny, line 265: Just call `chmod`/`chown` directly. The helpers are there for convenience, so that you don't have to type ED every time. When you need to strip it, they outlive their usefulness.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
```diff
@@ -0,0 +1,320 @@
...
+	diropts -m0755
+
+	# install the upgrade documentation
+	dodoc UPGRADE* BUGS CREDITS
+
+	# install extra documentation
+	if use doc; then
+		dodoc doc/*.txt
+		dodoc doc/*.pdf
+	fi
```
mgorny, line 296:
```suggestion
	use doc && dodoc doc/*.{txt,pdf}
```

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
```diff
@@ -0,0 +1,320 @@
...
+
+	# install logrotate snippet; bug #329281
+	#
+	insinto /etc/logrotate.d
+	newins "${FILESDIR}/1.6.2/asterisk.logrotate4" asterisk
+
+	# Asterisk installs a few folders that's empty by design,
+	# but still required.  This finds them, and marks them for
+	# portage.
+	for d in $(find "${ED}"/var -type d -empty || die "Find failed."); do
```
mgorny, line 306: Grok a bit for the proper way AKA `while read ... < <(find ...)` loop.

mgorny (changes requested):

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
jkroonza: IIRC this used to be used a long long time ago for indicate that an ebuild made use an USE_EXPAND variable.  Removed.  Not documented this way, and no other similar in tree, nor documented, or used, as you rightfully say.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
jkroonza: converted.  at a point (again, many years back, if I recall, could just have been around the system shell too) there was debate about bash vs sh, bash won out, so it means these features are available now.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
jkroonza: Done.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
jkroonza: Done locally (makes it easier to with a bit of diff | patch magic apply to the 16 subslot too.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
jkroonza: No.  It was just in pkg_setup historically, moved.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
jkroonza: None.  Done.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
jkroonza: OK great, that makes it simpler.

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
jkroonza: Incorporated locally (diff ... | patch).

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
jkroonza: Actually famliar with <() ... think the original variant still came from elsewhere, switched as it's definitely the more correct way to do it (if for some reason somehow it happens that a matching folder has whitespace in the name).

net-misc/asterisk/asterisk-13.38.1-r2.ebuild
jkroonza: There are indirect references to EPREFIX in the build system, so assuming you're right.  Based on what I've read re --prefix and --eprefix around ./configure ... I've updated other uses of ED to D as well as it seems more correct.  Please assist with verifying.

jkroonza: So after much reading, I think I finally "get" this.  You're right, updated.

PR #18995, opened 2021-01-08
x11-misc/screenkey: new proxy-maintained pkg; add 1.2, 1.3 and 9999

xgqt:


xgqt:
https://bugs.gentoo.org/724986#c15

xgqt:
moved from ::guru

https://gitweb.gentoo.org/repo/proj/guru.git/tree/x11-misc/screenkey

x11-misc/screenkey/metadata.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+    <maintainer type="person">
+        <email>xgqt@protonmail.com</email>
+        <name>Maciej Barć</name>
+    </maintainer>
+    <maintainer type="project">
```
stha09, line 8: Just add myself here (sultan@gentoo.org), no need for proxy-maint project.

x11-misc/screenkey/screenkey-1.3.ebuild
```diff
@@ -0,0 +1,43 @@
...
+	dev-python/python-distutils-extra[${PYTHON_USEDEP}]
+"
+RDEPEND="
+	dev-python/pycairo[${PYTHON_USEDEP}]
+	dev-python/pygobject[${PYTHON_USEDEP}]
+	sys-devel/gettext
+"
+
+src_prepare() {
+	xdg_environment_reset
```
stha09, line 37: Not needed.

x11-misc/screenkey/screenkey-1.3.ebuild
```diff
@@ -0,0 +1,43 @@
...
+
+	# Change the doc install path
+	sed -i "s/share\/doc\/screenkey/share\/doc\/${P}/g" setup.py || die
```
stha09, line 40: Just use | as delimiter to avoid escaping of / and change ${P} to ${PF}.

x11-misc/screenkey/screenkey-1.3.ebuild
```diff
@@ -0,0 +1,43 @@
...
+LICENSE="GPL-3"
+SLOT="0"
+
+BDEPEND="
+	dev-python/python-distutils-extra[${PYTHON_USEDEP}]
+"
+RDEPEND="
+	dev-python/pycairo[${PYTHON_USEDEP}]
+	dev-python/pygobject[${PYTHON_USEDEP}]
+	sys-devel/gettext
```
stha09, line 33: This is probably a build-time dependency only.

x11-misc/screenkey/screenkey-1.3.ebuild
```diff
@@ -0,0 +1,43 @@
...
+"
```
stha09, line 34: There are some missing dependencies here like gtk+, slop, fontawesome and libappindicator.

stha09 (changes requested):
Let's start with version 1.3 only. Not sure why you want to version 1.2 as well. I'm not a big fan of live ebuilds and upstream seems to change some deps at the moment, but if you want you can keep it, just double check the dependencies.

x11-misc/screenkey/metadata.xml
xgqt: Without me?

stha09: No, only remove the project and add myself instead. I will push the PRs for you.

x11-misc/screenkey/screenkey-1.3.ebuild
xgqt: > libappindicator

or kde-plasma/xembed-sni-proxy in KDE cause I use KDE w/o dev-libs/libappindicator and have tray icon

stha09: Yes, I think we can hide libappindicator behind USE=appindicator, only needed for GNOME shell.

xgqt: > Yes, I think we can hide libappindicator behind USE=appindicator, only needed for GNOME shell.

sure

xgqt:
Cairo is also needed at runtime


/usr/lib/python3.9/site-packages/Screenkey/screenkey.py

23   │ from gi.repository import Gtk, Gdk, GdkPixbuf, Pango, GObject
24   │ import cairo

stha09:
> Cairo is also needed at runtime
> 
> File: /usr/lib/python3.9/site-packages/Screenkey/screenkey.py
> 
> 23 │ from gi.repository import Gtk, Gdk, GdkPixbuf, Pango, GObject
> 24 │ import cairo
This is part of pycairo :)

xgqt:
> > Cairo is also needed at runtime
> > File: /usr/lib/python3.9/site-packages/Screenkey/screenkey.py
> > 23 │ from gi.repository import Gtk, Gdk, GdkPixbuf, Pango, GObject
> > 24 │ import cairo
> > This is part of pycairo :)

yes, thats what I meant by this 

you said before that "This is probably a build-time dependency only."

telans:
> you said before that "This is probably a build-time dependency only."

I believe they were referring to `sys-devel/gettext`

stha09:
> > you said before that "This is probably a build-time dependency only."
> 
> I believe they were referring to `sys-devel/gettext`

Exactly.

xgqt:
k

stha09:
Something went wrong with your update. Now all versions depend on `dev-python/Babel`, but only live ebuild should depend on it, everything else still uses `sys-devel/gettext`. I'm also not sure why you splitted `DEPEND` and `RDEPEND`. Additionally I checked the sources again and it seems the `COPYING.txt` is not up to date. All source files have GPLv3+ in the header, so you can update `LICENSE` variable as well.

stha09:
Oops, pressed wrong button, sorry :)

xgqt:
> Something went wrong with your update. Now all versions depend on `dev-python/Babel`, but only live ebuild should depend on it, everything else still uses `sys-devel/gettext`. I'm also not sure why you splitted `DEPEND` and `RDEPEND`. Additionally I checked the sources again and it seems the `COPYING.txt` is not up to date. All source files have GPLv3+ in the header, so you can update `LICENSE` variable as well.


setup.py:
> setup_requires=['setuptools', 'babel'],     
> install_requires=['PyGObject', 'pycairo'],


So we need this at buildtime
and then we need that runtime + other stuff like slop & fontawesome

xgqt:
@telans @stha09 Thanks for cooperation.
(https://github.com/gentoo/gentoo/commit/d55a67027a410ede37787b0b7b6d37abf0aa0ff3)

PR #18999, opened 2021-01-09
kicad bump to 5.1.9

l29ah:


zpuskas:
I've built and tested the proposed PR and it seems to be in order. Thank you for your contribution @l29ah.

PR #19011, opened 2021-01-09
dev-cpp/benchmark: Add support for debug USE flag

vilhelmgray:
Closes: https://bugs.gentoo.org/764653


trofi:
Pulled as is. Thank you!

PR #19014, opened 2021-01-10
dev-lang/hy: bump version to 0.20.0, enable python 3.8 and 3.9

pkulev:
0.17.0 version bumped to 0.20.0, python3.8 and 3.9 enabled for both.
Enabled tests (but I had to xfail one test, can't figure out how to fix it).

Closes: https://bugs.gentoo.org/758335
Signed-off-by: Pavel Kulyov <kulyov.pavel@gmail.com>


dev-lang/hy/hy-0.19.0.ebuild
```diff
@@ -1,10 +1,10 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
 RESTRICT="test" # needs some pointy sticks. Seriously.
-PYTHON_COMPAT=(python3_7)
+PYTHON_COMPAT=(python{3_7,3_8})
```
thesamesam, line 7: Can you add Python 3.9 too?

Also, style:
```suggestion
PYTHON_COMPAT=( python3_{7,8} )
```

dev-lang/hy/hy-0.19.0.ebuild
```diff
@@ -1,10 +1,10 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
 RESTRICT="test" # needs some pointy sticks. Seriously.
```
thesamesam, line 6: Can we fix tests?

dev-lang/hy/hy-0.19.0.ebuild
```diff
@@ -1,10 +1,10 @@
...
-PYTHON_COMPAT=(python3_7)
+PYTHON_COMPAT=(python{3_7,3_8})
 
 inherit distutils-r1 eutils
```
thesamesam, line 9: I bet we don't need eutils here

dev-lang/hy/hy-0.19.0.ebuild
```diff
@@ -17,7 +17,7 @@ KEYWORDS="~amd64 ~x86"
 IUSE="test doc"
```
thesamesam, line 17: New line after SRC_URI above, so it's LICENSE/SLOT/KEYWORDS/IUSE all together

thesamesam (commented):
We should be able to replace the nose dep and python_test definition with:
distutils_enable_tests nose

dev-lang/hy/hy-0.19.0.ebuild
pkulev: Python3_9 is blocked by funcparserlib, which works with 3.9 itself, but has no 3_9 in ebuild's PYTHON_COMPAT. Waiting https://github.com/vlasovskikh/funcparserlib/pull/63 to be merged.

thesamesam: If it works (and tests pass with FEATURES=test), feel free to add it in this PR! We're _trying_ to be more ahead of the curve this time around. I think we're getting there.

If tests pass for you, that's good enough for us. I'll check too when merging.

dev-lang/hy/hy-0.19.0.ebuild
thesamesam: Looks like you did, thank you!

dev-lang/hy/hy-0.19.0.ebuild
pkulev: Done

dev-lang/hy/hy-0.19.0.ebuild
pkulev: Fixed, now they run wih "test" use flag. But single test is failing, figuring out why. Just running them locally works well, but when they run by portage - one macro case fails.

thesamesam: Network or similar? We can skip fragile tests like that, but if we can figure it out, that's obviously nicer.

dev-lang/hy/hy-0.19.0.ebuild
pkulev: Ah, opened funcparserlib's ebuild and it is already bumped :smile:. Ok, now we can just add 3.9 and test.

dev-lang/hy/hy-0.19.0.ebuild
pkulev: Looks like something is misplaced within build directory during tests run. I commented out failing assert and next failed too. Maybe there are some docs I didn't notice that can help me with debugging tests? I call `ebuild <file> test` but cannot run single case. Is there the way to pass args to pytests via `ebuild` command or I have to run tests somehow from portage work directory?

dev-lang/hy/hy-0.19.0.ebuild
pkulev: I've also added hy-0.20.0 ebuild and it's working too, but have same issues with tests.

pkulev:
@thesamesam seems like that's it, worked for me on python 3.8 and 3.9 versions 19 and 20.
I see many potential improvements, but maybe next time. It's already much better now.

Things to fix later:
* patch get_version.py with current version, removing file writes
* skipped tests, seems like something with bytecode persistense between tests
* slots, install multiple versions + eselect module

pkulev:
@thesamesam ping

scardracs:
Is really necessary maintain both 0.19.0 and 0.20.0?

pkulev:
> Is it necessary to maintain both 0.19.0 and 0.20.0?

When I stared there was only 0.19.0. Issue was renamed so I've added new version too. It's not big deal to keep or to remove 0.19.0.

pkulev:
@scardracs, removed 0.19.0 ebuild.

PR #19015, opened 2021-01-10
app-emulation/docker-20.10.2 "USE Flag 'selinux' not in IUSE" and  "c…

mt190502:
Closes: https://bugs.gentoo.org/764758
Signed-off-by: Muhammed Taha mtaha1905@yandex.com
Package-Manager: Portage-3.0.12


AndrewAmmerlaan:

PR #19022, opened 2021-01-10
net-p2p/bitcoin{d,-cli,-qt}: cleanup and add cpe strings

ajakk:
Bug: https://bugs.gentoo.org/show_bug.cgi?id=711198


net-p2p/bitcoin-cli/metadata.xml
```diff
@@ -15,5 +15,6 @@
 	<upstream>
 		<remote-id type="github">bitcoin/bitcoin</remote-id>
 		<remote-id type="github">bitcoinknots/bitcoin</remote-id>
+		<remote-id type="cpe">cpe:/a:bitcoin:bitcoin_core</remote-id>
```
luke-jr, line 18: Should probably add one for Knots too?

net-p2p/bitcoin-cli/metadata.xml
```diff
@@ -15,5 +15,7 @@
...
+		<remote-id type="cpe">cpe:/a:bitcoin_knots_project:bitcoin_knots</remote-id>
```
luke-jr, line 19: Where are these CPEs coming from? Why is `bitcoin_core` under `bitcoin` and `bitcoin_knots` under `bitcoin_knots_project`?

ajakk: Bitcoin and Bitcoin Knots are different upstreams, right? That seems to be implied by the HOMEPAGE and SRC_URI values. From there I search NIST's CPE dictionary and modify the CPE format into what is used elsewhere in the tree.

https://nvd.nist.gov/products/cpe/search/results?namingFormat=2.3&keyword=bitcoin
https://nvd.nist.gov/products/cpe/search/results?namingFormat=2.3&keyword=bitcoin+knot

luke-jr: Bitcoin Core and Bitcoin Knots are different software releases of the same development process.

Neither one of them is Bitcoin. Neither one of them is the original Bitcoin project.

ajakk: Well, you told me I should add one for Knots. if the one I added is inaccurate then please offer a correction.

Are you saying the CPE tags on the CVEs in the security bug are wrong?

https://nvd.nist.gov/vuln/detail/CVE-2019-15947
https://nvd.nist.gov/vuln/detail/CVE-2020-14198

luke-jr: No idea how they got on the CVEs either.

ajakk: Well, since that's what is being added to the CVEs I'd say it's important that we match what they use (and not what's technically "right", because we're just trying to match what gets attached to the CVE data). The NVD links above have a link to an email you can contact if you think there's a missing CPE - cpe_dictionary at nist.gov.

PR #19027, opened 2021-01-11
media-libs/libavif: add svt-av1 use flag to live ebuild

novomesk:
media-libs/svt-av1 was added to ::gentoo recently. libavif can optionally use this AV1 encoder.


telans:
Instead of just doing the live ebuild you might want to wait for potential keywording on media-libs/svt-av1: https://bugs.gentoo.org/764761

novomesk:
I know about that. I updated the live ebuild first because it is the easiest part and the rest will follow at the right time later.

juippis (approved):

PR #19032, opened 2021-01-11
www-servers/lighttpd: Version bump (v1.4.58)

herbmillerjr:
Did some housekeeping as well as bump the version, summary follows. Also adding myself at proxied maintainer.

* removed mysql configuration item; replaced by USE flags for dbi
* adjusted descriptions of compression USE flags to more accurately describe what the compression is used for
* added a warning about file system side-effect of output compression
* adjusted behavior of `minimal` USE flag
  * corrected a bug where mod_compress (mod_deflate) was being deleted when using `minimal` USE flag even if compression use flags were specified
* confirmed compatible with up to Lua 5.4; pulled in slotted Lua approach from 1.4.55-r100 ebuild
* switched from geoip to maxminddb due to deprecation
* added additional SSL options (gnutls, nss, mbedtls)
* removed libressl USE flag since removing libressl is coming down the pipe for Gentoo

Upstream introduced a number of workflow changes that I tried to reflect in the ebuild, so please be brutal with the review. I want to make sure I'm moving the package forward without breaking users (too much).


herbmillerjr:
> ## Missing Bugzilla accounts
> **WARNING**: The following maintainers do not match any Bugzilla accounts:

Fixed e-mail on my bugzilla account.

juippis (approved):

PR #19034, opened 2021-01-11
autotools.eclass: Introduce AT_NOEAUTOHEADER variable

Polynomial-C:
Sometimes packages get their headers being broken by autoheader (see
media-libs/freetype for example) or get unwanted defines added (see
media-libs/libsdl2). In that case we want to run eautoreconf without
call to eautoheader.

Bug: https://bugs.gentoo.org/764959

### DIFF
--- a/eclass/autotools.eclass
+++ b/eclass/autotools.eclass
@@ -54,6 +54,8 @@ inherit libtool
 # CONSTANT!
 # The latest major unstable and stable version/slot of automake available
 # on each arch.
+# Only add unstable version if it is in a different slot than latest stable
+# version.
 # List latest unstable version first to boost testing adoption rate because
 # most package manager dependency resolver will pick the first suitable
 # version.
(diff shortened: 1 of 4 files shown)


eclass/autotools.eclass
```diff
@@ -136,6 +136,12 @@ unset _automake_atom _autoconf_atom
 # Additional options to pass to automake during
 # eautoreconf call.
 
+# @ECLASS-VARIABLE: AT_NOEAUTOHEADER
+# @DEFAULT_UNSET
+# @DESCRIPTION:
+# Don't run eautoheader command if set to 'yes'; only used to workaround
```
Arfrever, line 142: That noun can be written as `workaround`, but verb is `work around`.

Polynomial-C: Thanks. Fixed.

eclass/autotools.eclass
```diff
@@ -65,7 +65,7 @@ inherit libtool
 # Do NOT change this variable in your ebuilds!
 # If you want to force a newer minor version, you can specify the correct
 # WANT value by using a colon:  <PV>:<WANT_AUTOMAKE>
-_LATEST_AUTOMAKE=( 1.16.1:1.16 1.15.1:1.15 )
+_LATEST_AUTOMAKE=( 1.16.3:1.16 1.16.1:1.16 )
```
Arfrever, line 68: This does not make sense.
If you want that latest mean only `1.16` slot, then use:
```
_LATEST_AUTOMAKE=( 1.16.3:1.16 )
```

floppym (approved):

Arfrever:
Also copyright 2021 :) .

PR #19007, opened 2021-01-09
dev-lang/zig: Bump to version 0.7.1

atomsymbol:
This patch is performing the following actions:
- Adds new "shared" USE flag to `>=sys-devel/lld-11*`
- Drops `zig-0.5.0.ebuild`
- Adds `zig-0.7.1.ebuild`, which depends on `sys-devel/lld[shared]`
- Removes support for the "experimental" USE flag from `zig-9999`
- Adds a proxy maintainer to `dev-lang/zig/metadata.xml`

@mgorny Please specify (in sufficient detail) how to split this patch into smaller sub-patches, if required. Thanks.

Closes: https://bugs.gentoo.org/719736

PR #19038, opened 2021-01-11
sys-devel/lld: Add "shared" USE flag

atomsymbol:
`lld`'s shared libraries and include files are required by `dev-lang/zig`.

The patched `lld` ebuilds are equivalent to the original `lld` ebuilds if the `shared` USE flag is disabled. The flag is disabled by default, so nothing changes for current users of the `lld` package who don't intend to install `dev-lang/zig`.

Please merge. Thanks.

See also: https://bugs.gentoo.org/719736
See also: https://github.com/gentoo/gentoo/pull/19007

### DIFF
--- a/sys-devel/lld/lld-11.0.0.ebuild
+++ b/sys-devel/lld/lld-11.0.0.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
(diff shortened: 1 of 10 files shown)


sys-devel/lld/lld-11.0.0.ebuild
```diff
@@ -36,7 +36,7 @@ pkg_setup() {
 
 src_configure() {
 	local mycmakeargs=(
-		-DBUILD_SHARED_LIBS=OFF
+		-DBUILD_SHARED_LIBS=$(use shared && echo ON || echo OFF)
```
telans, line 39:
```suggestion
		-DBUILD_SHARED_LIBS=$(usex shared)
```

atomsymbol:
There is no interest in this patch.

PR #19041, opened 2021-01-12
media-libs/gmmlib: fixed build with custom cflags

Karlson2k:
Fixes https://bugs.gentoo.org/show_bug.cgi?id=765055
`-msse4.1` compiler flag is required to inline function, which is used only if SSE4.1 is detected at runtime.

### DIFF
--- a/media-libs/gmmlib/files/gmmlib-20.4.1_custom_cflags_v2.patch
+++ b/media-libs/gmmlib/files/gmmlib-20.4.1_custom_cflags_v2.patch
@@ -1,0 +1,98 @@
+    Prevent overriding of user-define CFLAGS, including -march flag.
+    The flag -msse4.1 is required otherwise compile will not be able to inline sse4.1 code.
+    Note: code should detect CPU features at runtime and use sse4.1 paths only if supported by CPU.
+
+--- a/Source/GmmLib/CMakeLists.txt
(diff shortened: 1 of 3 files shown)


thesamesam:
Can you look at the libva test failures too? I'm wondering if they need Intel hardware to pass or something.

Karlson2k:
> Can you look at the libva test failures too? I'm wondering if they need Intel hardware to pass or something.

Can't find any reported bugs for libva. Do you mean [libva-utils](https://bugs.gentoo.org/758620)?

thesamesam:
sorry, yes!

Karlson2k:
I'll check it after fixing x11-libs/libva-intel-media-driver-20.4.5

thesamesam:
btw, no need for [revbump](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/) for a failed build (users who couldn't upgrade will already have it queued on their nexe world update)

Karlson2k:
Updated without revbump. Should I modify .patch file in-place as well?

thesamesam:
Yeah, patch file in place would've been fine, but we can just clean it up later too post-stabilisation.

FWIW: re stabilisation, the reason nothing has happened is failing tests, so hopefully merging these PRs will help.

PR #19045, opened 2021-01-12
x11-wm/herbstluftwm: fix python-exec[-native-symlinks], add 0.9.1

Flowdalic:


juippis (approved):
Can't test the native-symlinks but looks good enough.

PR #19048, opened 2021-01-13
libva-utils: fixed test in portage sandbox

Karlson2k:
livba-utils contains more than 17 000 internal tests, but almost all of them require installed VA driver and Intel video hardware.
Tests inside portage sandbox cannot access VA driver and cannot rely on presence of Intel video hardware.

There are two tests that still can be used inside portage, so let's use these tests as this is the best possible option for portage and better than no tests at all.


Karlson2k:
@thesamesam As you requested.

PR #19052, opened 2021-01-13
dev-libs/hyperscan: bump to 5.4.0

arkamar:


dev-libs/hyperscan/hyperscan-5.4.0.ebuild
```diff
@@ -0,0 +1,54 @@
...
+inherit cmake flag-o-matic python-any-r1
+
+DESCRIPTION="High-performance regular expression matching library"
+SRC_URI="https://github.com/intel/hyperscan/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+HOMEPAGE="https://www.hyperscan.io/ https://github.com/intel/hyperscan"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="cpu_flags_x86_avx2 +cpu_flags_x86_ssse3 static-libs"
```
thesamesam, line 17: Can we [drop](https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0302) static-libs?

dev-libs/hyperscan/hyperscan-5.4.0.ebuild
```diff
@@ -0,0 +1,54 @@
...
+
+RDEPEND="dev-libs/boost"
```
thesamesam, line 19: := for anything we link against which exposes a subslot, but especially Boost

dev-libs/hyperscan/hyperscan-5.4.0.ebuild
```diff
@@ -0,0 +1,54 @@
...
+	# Respect user -O flags
+	sed -i '/set(OPT_CX*_FLAG/d' CMakeLists.txt || die
+
+	# upstream workaround
+	append-cxxflags -Wno-redundant-move
+	cmake_src_prepare
+}
+
+src_configure() {
+	CMAKE_BUILD_TYPE=Release
```
thesamesam, line 38: We shouldn't need to set this - the cmake eclass does it.

dev-libs/hyperscan/hyperscan-5.4.0.ebuild
arkamar: In my opinion yes, we can.

dev-libs/hyperscan/hyperscan-5.4.0.ebuild
arkamar: Ok, thanks.

dev-libs/hyperscan/hyperscan-5.4.0.ebuild
arkamar: The eclass sets `-DCMAKE_BUILD_TYPE=Gentoo`, but `hyperscan` cmake matches it with `Release`. This varibale is necessary, see https://github.com/gentoo/gentoo/pull/15681#discussion_r436924060, where I introduce it. Anyway, maybe I should add some comment to it.

thesamesam: Oh, great. Thanks. I'll restore it with a comment.

arkamar: `-Werror` is not the only thing that is triggered by `Release` build type. Maybe, the best will be to patch `CMakeList.txt` this way:
```diff
@@ -30,7 +30,7 @@ else()
     message(STATUS "Build type ${CMAKE_BUILD_TYPE}")
 endif()

-if(CMAKE_BUILD_TYPE MATCHES RELEASE|RELWITHDEBINFO|MINSIZEREL)
+if(CMAKE_BUILD_TYPE MATCHES RELEASE|RELWITHDEBINFO|MINSIZEREL|Gentoo)
     message(STATUS "using release build")
     set(RELEASE_BUILD TRUE)
 else()
```

arkamar: Btw, I just have checked that Gentoo does same `CMakeFile.txt` fixes with `CMAKE_BUILD_TYPE` set to `Gentoo` or `Release`.

PR #19054, opened 2021-01-13
dist-kernel: dracut `uefi=yes` support

mgorny:
Includes `mount-boot.eclass` patches for nonfatal support.


eclass/dist-kernel-utils.eclass
```diff
@@ -85,11 +97,28 @@ dist-kernel_install_kernel() {
 	local image=${2}
 	local map=${3}
 
+	# if dracut is used in eufi=yes mode, initrd will actually
```
Dessa, line 100: uefi=yes

PR #19055, opened 2021-01-13
net-vpn/tor: Version updated to 0.4.5.3-rc.

jpds:
Simple version bump, tested on `amd64`.


blueness:
pushed.  i also bumped to 0.4.5.4-rc

PR #19056, opened 2021-01-13
app-misc/ckb: add 0.4.3

ajakk:

Quazip has no use flags so drop the use in its dep and its version constraint has long been pointless so drop it too. Add a couple new dependencies, drop GCC10 patch (compile failure fixed in this release).

```
1c1
< # Copyright 1999-2020 Gentoo Authors
---
> # Copyright 1999-2021 Gentoo Authors
16,17c16,17
< RDEPEND="
<       >=dev-libs/quazip-0.7.2[qt5(+)]
---
> RDEPEND="dev-libs/libdbusmenu-qt
>       dev-libs/quazip
18a19
>       dev-qt/qtdbus:5
21a23,24
>       dev-qt/qtx11extras:5
>       media-sound/pulseaudio
23c26,28
<       x11-libs/libX11"
---
>       x11-libs/libX11
>       x11-libs/libxcb:=
>       x11-libs/xcb-util-wm"
27,28d31
<
< PATCHES=( "${FILESDIR}"/${P}-gcc10.patch )
```


Dwosky:
I have two suggestions:

1. Since this is an alternate version of the original ckb project ([https://github.com/ccMSC/ckb](https://github.com/ccMSC/ckb)), shouldn't it make more sense to have it as **app-misc/ckb-next** instead?
2. Upon starting the application, there is an error window telling the user to run the daemon first:
image
I think we can avoid this popup by setting a message in the ebuild itself, telling the user to enable/start the daemon.

ajakk:
> I have two suggestions:
> 
>     1. Since this is an alternate version of the original ckb project (https://github.com/ccMSC/ckb), shouldn't it make more sense to have it as **app-misc/ckb-next** instead?

In my opinion, yes, but this isn't a package I created.

>     2. Upon starting the application, there is an error window telling the user to run the daemon first:
>        image
>        I think we can avoid this popup by setting a message in the ebuild itself, telling the user to enable/start the daemon.

I agree.


I should also note that the combo of ckb and systemd-247 is breaking input in graphical environments: https://github.com/systemd/systemd/issues/17922. This update fixes that.

PR #19057, opened 2021-01-13
app-i18n/transifex-client: drop dependency to dev-python/mock

robert7k:
Bug: https://bugs.gentoo.org/765349


juippis (commented):
```
error: Could not find suitable distribution for Requirement.parse('smmap<4,>=3.0.1')
 * ERROR: app-i18n/transifex-client-0.14.2-r1::testworld failed (test phase):
 *   Tests fail with python3.7
```
It also has a strict version requirement...

robert7k:
> ```
> error: Could not find suitable distribution for Requirement.parse('smmap<4,>=3.0.1')
>  * ERROR: app-i18n/transifex-client-0.14.2-r1::testworld failed (test phase):
>  *   Tests fail with python3.7
> ```
> 
> It also has a strict version requirement...

That's strange – smmap isn't even a direct dependency of transifex-client. I also couldn't reproduce this after installing smmap-4.0.0 (both with Python 3.7 and 3.9). Do you have more information on where exactly this error message comes from?

Rebased the PR on the most recent master.

ionenwks:
> > ```
> > error: Could not find suitable distribution for Requirement.parse('smmap<4,>=3.0.1')
> > ```
> > It also has a strict version requirement...
>
> That's strange – smmap isn't even a direct dependency of transifex-client. I also couldn't reproduce this after installing smmap-4.0.0 (both with Python 3.7 and 3.9). Do you have more information on where exactly this error message comes from?

Was looking at this earlier because of https://bugs.gentoo.org/767970, this strict dependency was set nearly a year ago and 4.0.0 was released only 7 days ago. I haven't tested but like @thesamesam I question its validity.

ionenwks:
>https://bugs.gentoo.org/767970

Been resolved, I don't think this package need to worry about this smmap issue.

app-i18n/transifex-client/transifex-client-0.14.2-r1.ebuild
```diff
@@ -31,4 +30,5 @@ src_prepare() {
 	eapply_user
 	sed -i -e 's:test_fetch_timestamp_from_git_tree:_&:' \
 		tests/test_utils.py || die
+	sed -i ':tests_require=["mock>=3.0.5,<4.0"]:d' setup.py || die
```
juippis, line 33:
```
>>> Source compiled.
>>> Test phase: app-i18n/transifex-client-0.14.2-r1
 * python3_7: running distutils-r1_run_phase python_test
python3.7 setup.py test --verbose
Warning: 'keywords' should be a list, got type 'tuple'
running test
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
WARNING: The wheel package is not available.
/usr/bin/python3.7: No module named pip
error: Command '['/usr/bin/python3.7', '-m', 'pip', '--disable-pip-version-check', 'wheel', '--no-deps', '-w', '/var/tmp/portage/app-i18n/transifex-client-0.14.2-r1/temp/tmpfl1rd2d3', '--quiet', 'mock<4.0,>=3.0.5']' returned non-zero exit status 1.
 * ERROR: app-i18n/transifex-client-0.14.2-r1::testworld failed (test phase):
 *   Tests fail with python3.7
```
```
>: grep -i mock /var/tmp/portage/app-i18n/transifex-client-0.14.2-r1/work/transifex-client-0.14.2/setup.py 
    tests_require=["mock>=3.0.5,<4.0"],
```
I don't know why, but the `:` don't seem to work very well as delimiters. Switch to `/` and it works.

robert7k: Strange, worked fine for me. Replaced anyway.

app-i18n/transifex-client/transifex-client-0.14.2-r1.ebuild
```diff
@@ -0,0 +1,34 @@
...
+	<dev-python/six-2.0.0[${PYTHON_USEDEP}]
+	dev-python/urllib3[${PYTHON_USEDEP}]"
+
+distutils_enable_tests setup.py
+
+src_prepare() {
+	eapply_user
+	sed -i -e 's:test_fetch_timestamp_from_git_tree:_&:' \
+		tests/test_utils.py || die
+	sed -i '/tests_require=["mock>=3.0.5,<4.0"]/d' setup.py || die
```
juippis, line 33:
```
>: cd /var/tmp/portage/app-i18n/transifex-client-0.14.2-r1/work/transifex-client-0.14.2/
>: grep -i mock setup.py 
    tests_require=["mock>=3.0.5,<4.0"],
>: sed -i '/tests_require=["mock>=3.0.5,<4.0"]/d' setup.py
>: grep -i mock setup.py 
    tests_require=["mock>=3.0.5,<4.0"],
>: sed -i '/tests_require=/d' setup.py
>: grep -i mock setup.py 
>: 
Seriously what is going on? Can someone explain? Is it **just** me?

Comment (ionenwks) [reply app-i18n/transifex-client/transifex-client-0.14.2-r1.ebuild:33]: That sed doesn't work for me either, fine if brackets are escaped.
`sed -i '/tests_require=\["mock>=3.0.5,<4.0"\]/d' setup.py`

Comment (robert7k) [reply app-i18n/transifex-client/transifex-client-0.14.2-r1.ebuild]: Makes sense to escape the brackets, so I added that. Still wondering why it worked for me anyway...

**Review by juippis — COMMENTED** [review]
So now the sed works, but the test cases are actually using mock.
```
ERROR: test_api (unittest.loader._FailedTest)
ImportError: Failed to import test module: test_api
  File "/usr/lib/python3.7/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
  File "/var/tmp/portage/app-i18n/transifex-client-0.14.2-r1/work/transifex-client-0.14.2/tests/test_api.py", line 4, in <module>
    from mock import MagicMock, patch
ModuleNotFoundError: No module named 'mock'


Ran 10 tests in 0.001s

FAILED (errors=6)
Test failed: <unittest.runner.TextTestResult run=10 errors=6 failures=0>
error: Test failed: <unittest.runner.TextTestResult run=10 errors=6 failures=0>
 * ERROR: app-i18n/transifex-client-0.14.2-r1::testworld failed (test phase):
 *   Tests fail with python3.7
```
https://github.com/transifex/transifex-client/blob/master/tests/test_project.py#L16
https://github.com/transifex/transifex-client/blob/master/tests/test_api.py#L4

What are we doing in here? It does look like removing mock needs a bit more, ie a full patch.

juippis (approved):
Only now I get the point of this PR, the tests work with later version of mock, so the restriction is redundant in setup.py. We still need to BDEPEND on mock though. 

Thanks for sticking it through!

PR #19059, opened 2021-01-14
media-gfx/prusaslicer: 2.3.0 version bump + fixes

ionenwks:
Should hopefully be everything.


media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
```diff
@@ -0,0 +1,102 @@
...
+DEPEND="${RDEPEND}"
+
+S="${WORKDIR}/${MY_PN}-version_${PV}"
+
+PATCHES=(
+	"${FILESDIR}/${P}-miniz-zip-header.patch"
+)
+
+src_prepare() {
+	use gui && setup-wxwidgets
```
SoapGentoo, line 63: put this in `src_configure`

SoapGentoo (changes requested):

expeditioneer (approved):
Thank you for your contribution @ionenwks !
The PR looks good to me and i would like to merge that.

@thesamesam i'm not part of the mathematics projects, so somebody has to decide if the changes to qhull are correct and desired.

The changes to 3dprinting affected files are fine. After a review of the qhull changes fell free to merge this PR.

ionenwks:
And thanks for the review.

For qhull, I did end up using meh paths while trying to "hopefully" fix all of its issues (two additional seds for get_libdir, re-add static-libs, a patch), which may be fragile in updates.

static-libs wasn't re-added on a whim, upstream set up some non-deprecated libs to be static-only intentionally which prusaslicer needs (currently fails to detect qhull 2015.2 and falls back to bundled, for 2020.2 it half-detect it but then fail given static library is missing).

Changes don't seem to break consumers I've tried (gr, matplotlib, octave, pcl, plplot), all of which are using the deprecated lib that was half re-added in a rush when it broke matplotlib in ~testing.

I tried meshlab as well but seems that, even with current qhull-2020.2-r2, it uses its bundled qhull-2003.1 instead, also ran into unrelated issues given missing deps on gmp[cxx] and qtdeclarative ([bug #763837](https://bugs.gentoo.org/763837))

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
```diff
@@ -0,0 +1,99 @@
...
+	cmake_src_configure
+}
+
+src_install() {
+	cmake_src_install
+
+	if use gui; then
+		newicon -s 128 resources/icons/PrusaSlicer_128px.png PrusaSlicer.png || die
+		newicon -s 128 resources/icons/PrusaSlicer-gcodeviewer_128px.png PrusaSlicer-gcodeviewer.png || die
+		domenu src/platform/unix/Prusa{Slicer,Gcodeviewer}.desktop || die
```
SoapGentoo, line 83: all these `|| die`s aren't needed, they all come from `desktop.eclass`

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
```diff
@@ -0,0 +1,99 @@
...
+		xdg_desktop_database_update
+	fi
+}
+
+pkg_postrm() {
+	if use gui; then
+		xdg_icon_cache_update
+		xdg_desktop_database_update
+	fi
+}
```
SoapGentoo, line 99: personally, I would remove these and just rely on `xdg.eclass` instead to do the right thing

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
```diff
@@ -0,0 +1,99 @@
...
+MY_PN="PrusaSlicer"
+
+DESCRIPTION="A mesh slicer to generate G-code for fused-filament-fabrication (3D printers)"
+HOMEPAGE="https://www.prusa3d.com/prusaslicer/"
+SRC_URI="https://github.com/prusa3d/${MY_PN}/archive/version_${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="AGPL-3 Boost-1.0 GPL-2 LGPL-3 MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
```
SoapGentoo, line 19: remove empty line

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
```diff
@@ -0,0 +1,99 @@
...
+IUSE="gui test"
+RESTRICT="!test? ( test )"
+
+# tests fail to link with USE=-gui, bug #760096
+REQUIRED_USE="test? ( gui )"
```
SoapGentoo, line 24: put this before `RESTRICT`

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
```diff
@@ -0,0 +1,99 @@
...
+	>=dev-libs/boost-1.73.0:=[nls,threads]
+	dev-libs/cereal
+	dev-libs/expat
+	dev-libs/gmp:=
+	dev-libs/miniz
+	dev-libs/mpfr:=
+	>=media-gfx/openvdb-5.0.0
+	media-libs/ilmbase:=
+	media-libs/libpng:0=
+	media-libs/qhull:=[static-libs]
```
SoapGentoo, line 38: and there's no way to avoid the static libs?

SoapGentoo (changes requested):

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
ionenwks: Done. I should've noticed that myself.

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
ionenwks: It's trivial but xdg.eclass adds two unused DEPEND with -gui that would be allowed to be missing if glib isn't installed.

But don't mind changing anyway if believe ebuild simplicity is more important.

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
ionenwks: Done, left a empty line before the next comment for readability though.

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
ionenwks: qhull upstream states that libqhullcpp.a is intended to be static-only, and prusaslicer needs both libqhull_r.so and libqhullcpp.a

Before it was using the bundled copy, and I don't think that's better.

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
ionenwks: Done.

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
ionenwks: Well, upon closer look, it gets overriden later with DEPEND= anyway, which makes me wonder why it's there (it's setup to fail gracefully if deps are missing). So I'll change it in a moment.

ionenwks: Done.

media-gfx/prusaslicer/prusaslicer-2.3.0.ebuild
ionenwks: Doesn't get rid of it but moved [static-libs] to DEPEND only.

ionenwks:
ping, is there anything else? Considering the previous in-tree version is almost completely broken as-is it'd be nice to get this finished so users can actually use this without workarounds.

PR #19072, opened 2021-01-15
net-mail/muchsync: Bump to version 6

rseichter:
Closes: https://bugs.gentoo.org/765445


thesamesam:
Thank you! I added virtual/pkgconfig as a BDEPEND too.

PR #19075, opened 2021-01-16
sys-apps/bat: Bump dev-libs/libgit2 version and +threads

toshokan:
This is the version recommended by upstream at
https://github.com/rust-lang/git2-rs/blob/master/README.md#version-of-libgit2

Closes: https://bugs.gentoo.org/762811
Signed-off-by: J. Pz <toshokan@shojigate.net>

### DIFF
--- a/sys-apps/bat/bat-0.17.1-r1.ebuild
+++ b/sys-apps/bat/bat-0.17.1-r1.ebuild
@@ -1,0 +1,200 @@
+# Copyright 2019-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+CRATES="
+adler-0.2.3
+aho-corasick-0.7.13
+ansi_colours-1.0.1
+ansi_term-0.11.0
+ansi_term-0.12.1
+arrayref-0.3.6
+arrayvec-0.5.1
+assert_cmd-1.0.1
+atty-0.2.14
+autocfg-1.0.1
+base64-0.12.3
+bincode-1.3.1
+bitflags-1.2.1
+bit-set-0.5.2
+bit-vec-0.6.2
+blake2b_simd-0.5.10
+block-buffer-0.7.3
+block-padding-0.1.5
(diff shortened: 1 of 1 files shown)


sys-apps/bat/bat-0.17.1-r1.ebuild
```diff
@@ -0,0 +1,200 @@
...
+SRC_URI="https://github.com/sharkdp/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+SRC_URI+=" $(cargo_crate_uris ${CRATES})"
+
+LICENSE="Apache-2.0 Boost-1.0 BSD BSD-2 CC0-1.0 ISC LGPL-3+ MIT Apache-2.0 Unlicense ZLIB"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64 ~ppc64 ~x86"
+
+BDEPEND="virtual/pkgconfig"
+DEPEND="
+	>=dev-libs/libgit2-1.1.0[threads]
```
thesamesam, line 169: We need to keep the := which got dropped here to force rebuilds when libgit2's ABI changes

sys-apps/bat/bat-0.17.1-r1.ebuild
```diff
@@ -0,0 +1,200 @@
...
+	dev-libs/oniguruma
```
thesamesam, line 170: oniguruma has a subslot defined (SLOT="X/N" where N is the subslot part), so we should subscribe to that too with :=

toshokan:
@thesamesam Good catch! It looks like sys-libs/zlib also has a subslot defined

thesamesam:
@toshokan You're 100% right but we added a subslot to zlib and we spend our time arguing over whether it was a good idea or not, as it breaks ABI irregularly. It DOES happen, just not often, so really, adding new ones isn't a problem - it's just not worth adding := in existing ebuilds for zlib when not doing other changes. We'll leave it for now anyway.

PR #19077, opened 2021-01-16
media-plugins/vdr-vaapidevice: new package

madmartin:
This is an output device plugin for media-video/vdr

Closes: https://bugs.gentoo.org/765610


media-plugins/vdr-vaapidevice/vdr-vaapidevice-1.0.0_p20190526.ebuild
```diff
@@ -0,0 +1,51 @@
...
+"
+S="${WORKDIR}/vdr-plugin-vaapidevice-${GIT_COMMIT}"
+
+src_prepare() {
+	vdr-plugin-2_src_prepare
+
+	use debug && append-cppflags -DDEBUG
+
+	local GIT_COMMIT_SHORT=${GIT_COMMIT:0:7}
+	sed -i -e "s:GIT_REV =.*:GIT_REV=-${GIT_COMMIT_SHORT}:" Makefile || die "Failed to modify Makefile"
```
telans, line 50:
```suggestion
	sed -i "s|GIT_REV =.*|GIT_REV=-${GIT_COMMIT:0:7}|" Makefile || die
```
might be good to use a different delimiter: https://bugs.gentoo.org/685160

thesamesam: But a git commit won't contain a :, right? It's more of an issue for CFLAGS where that's a valid character.

telans: true, I suppose it doesn't matter

thesamesam: Being in the habit of thinking about it definitely matters though. I keep meaning to make a list of general gotchas like this...

madmartin: > But a git commit won't contain a :, right? It's more of an issue for CFLAGS where that's a valid character.

Yes, a git commit hash never contains ":". And of course I ALWAYS think of the delimiter when using sed search&replace.

media-plugins/vdr-vaapidevice/vdr-vaapidevice-1.0.0_p20190526.ebuild
```diff
@@ -0,0 +1,51 @@
...
+	x11-libs/xcb-util-wm
+	x11-libs/xcb-util-keysyms
+	media-libs/alsa-lib
+"
+DEPEND="${RDEPEND}"
+BDEPEND="
+	virtual/pkgconfig
+"
+
+VDR_CONFD_FILE="${FILESDIR}/confd"
```
hd-brummy, line 37: VDR_CONFD_FILE to define here is not needed, as files/confd exist. The vdr-plugin-2.eclass will pickup this then automaticly.
VDR_CONFD_FILE is used to overwrite this handling, see media-plugins/vdr-live for a good example.
I noticed, the description in the eclass is not exact, i will fix this later in the near future.

madmartin: @hd-brummy  Got it. Ebuild is adjusted.

PR #19078, opened 2021-01-16
Docs and distutils eclasses: make doc building compatible with python-single-r1

AndrewAmmerlaan:
python-single-r1 does not have the python_gen_any_dep function, in this case we should use the python_gen_cond_dep function instead to generate the correct dependencies. This makes doc building using docs.eclass or distutils-r1's `distutils_enable_sphinx` function compatible with python-single-r1.eclass and thus with `DISTUTILS_SINGLE_IMPL` and closes bug https://bugs.gentoo.org/704520.

I will also mail this to the mailing list ASAP

### DIFF
--- a/eclass/distutils-r1.eclass
+++ b/eclass/distutils-r1.eclass
@@ -316,22 +316,25 @@ distutils_enable_sphinx() {
 	_DISTUTILS_SPHINX_PLUGINS=( "${@}" )
 
 	local deps autodoc=1 d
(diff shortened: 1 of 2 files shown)


AndrewAmmerlaan:
> I will also mail this to the mailing list ASAP

Since there have been no further comments/suggestions/objections on the mailing list, could we please get this merged? @juippis @mgorny

juippis (approved):
Thanks, seems to work!

AndrewAmmerlaan:
Thank You @juippis !

PR #19079, opened 2021-01-16
Fix dependencies for Xfce pulseaudio plugins

dm0-:
@mgorny

### DIFF
--- a/dev-libs/keybinder/keybinder-0.3.2-r300.ebuild
+++ b/dev-libs/keybinder/keybinder-0.3.2-r300.ebuild
@@ -1,12 +1,13 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
 
 MY_P=${PN}-3.0-${PV}
 DESCRIPTION="A library for registering global keyboard shortcuts"
 HOMEPAGE="https://github.com/kupferlauncher/keybinder"
 SRC_URI="https://github.com/kupferlauncher/keybinder/releases/download/${PN}-3.0-v${PV}/${MY_P}.tar.gz"
+S=${WORKDIR}/${MY_P}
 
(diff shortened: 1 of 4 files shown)


dev-libs/keybinder/keybinder-0.3.2-r300.ebuild
```diff
@@ -18,8 +18,8 @@ RDEPEND="x11-libs/gtk+:3[X]
 	x11-libs/libXext
 	x11-libs/libXrender
 	introspection? ( dev-libs/gobject-introspection )"
```
mgorny, line 20: Are you sure that gobject-introspection is a DEPEND here? Also, I'm pretty sure it should go to BDEPEND too because of `g-ir-compiler`.

dev-libs/keybinder/keybinder-0.3.2-r300.ebuild
```diff
@@ -18,8 +18,8 @@ RDEPEND="x11-libs/gtk+:3[X]
...
-DEPEND="${RDEPEND}
-	virtual/pkgconfig"
+DEPEND="${RDEPEND}"
+BDEPEND="virtual/pkgconfig"
 
 S=${WORKDIR}/${MY_P}
```
mgorny, line 24: While touching it, could you move S directly below SRC_URI?

mgorny (changes requested):

dev-libs/keybinder/keybinder-0.3.2-r300.ebuild
dm0-: I moved it.  This was the same as what I wrote somewhere in #18815, where I just left it in `DEPEND` like most GNOME packages did, and it shouldn't make a difference in most cases since introspection can't be cross-compiled.

mgorny: Heh, and I thought their elaborate system did account for cross-compilation.

mgorny (approved):
Thanks a lot!

PR #19081, opened 2021-01-16
dev-db/pgcli: Fix test dependencies

jsmolic:
Closes: https://bugs.gentoo.org/765622

### DIFF
--- a/dev-db/pgcli/pgcli-3.1.0.ebuild
+++ b/dev-db/pgcli/pgcli-3.1.0.ebuild
@@ -31,7 +31,6 @@ BDEPEND="
 	test? (
 		dev-db/postgresql
 		dev-python/mock[${PYTHON_USEDEP}]
-		dev-python/python-dateutil[${PYTHON_USEDEP}]
 	)"
 
 distutils_enable_tests pytest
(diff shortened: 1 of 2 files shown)


dev-db/pgcli/pgcli-3.1.0.ebuild
```diff
@@ -31,7 +31,6 @@ BDEPEND="
 	test? (
 		dev-db/postgresql
 		dev-python/mock[${PYTHON_USEDEP}]
-		dev-python/python-dateutil[${PYTHON_USEDEP}]
```
thesamesam, line 34: If it directly needs dateutil, keep it in

dev-python/pendulum/pendulum-2.1.2.ebuild
```diff
@@ -21,7 +21,7 @@ KEYWORDS="~amd64 ~x86"
 # commented for future tests fixups
 RESTRICT="test"
 
-RDEPEND="
+DEPEND="
```
thesamesam, line 24: I'll revbump anyway because it's cheap for a Python package

jsmolic: Thanks, I forgot to revbump

dev-db/pgcli/pgcli-3.1.0.ebuild
jsmolic: Actually `pendulum` needs it, but didn't pull it in due to DEPEND issue

PR #19082, opened 2021-01-16
app-accessibility/caribou: Fix crash due to xserver CVE fix

SparkyBluefang:
xorg-server 1.20.10 contains a fix for CVE-2020-25712.

This causes a previous work-around in caribou to result in a BadLength error, when interacting with XKB, and crashes the hosting application.

Fixes provided by Cinnamon upstream from Fedora RPM:
https://src.fedoraproject.org/rpms/caribou/tree/master

This also includes an upstreamed antler fix and bumps PYTHON_COMPAT for 3.9 support.

Bug: https://bugs.gentoo.org/765661
linuxmint/cinnamon-screensaver#354

Since this is security related, should this be committed as stable and clean-up the old version? Or would the normal stabilization/clean-up process just be expedited?

### DIFF
--- a/app-accessibility/caribou/caribou-0.4.21-r2.ebuild
+++ b/app-accessibility/caribou/caribou-0.4.21-r2.ebuild
@@ -1,0 +1,103 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+GNOME2_LA_PUNT="yes"
+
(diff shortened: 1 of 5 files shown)


thesamesam:
We'll stable it quickly. Thanks! Testing now...

leio:
Please take over the maintenance of this package if maintainer is skipped anyways. It's a good thing (we don't want this package anymore), but please make it official by maintaining caribou along with cinnamon or whatnot :)

PR #19085, opened 2021-01-17
app-misc/rpick: Replace 0.7.1 with 0.7.2

bowlofeggs:
0.7.1 has a testing issue. Since it is not marked
stable, this commit replaces it with 0.7.2 to address
that issue.

Closes: https://bugs.gentoo.org/765604


juippis (approved):
Thanks, remember that even for small quick fixes like the these, we usually do 2 commits. 1 for adding a new ebuild, 1 for removing the old. Usually `git mv` is acceptable on a minor revbump so it's easier to see what exactly was changed.

bowlofeggs:
Thanks @juippis, I'll keep that in mind for next time.

PR #19089, opened 2021-01-17
dev-lang/python-exec*: python-exec.conf generated from PYTHON_TARGETS

mgorny:


SoapGentoo (approved):
Thanks for doing this, because I agree that the previous behaviour violated the principle of least surprise.

PR #19090, opened 2021-01-17
sci-libs/amd: only depend on virtual/fortran if USE=fortran

746B61:
Add FORTRAN_NEEDED=fortran before inheriting the fortran-2 eclass so that the ebuild does not depend on virtual/fortran if USE=-fortran.

Closes: https://bugs.gentoo.org/765739
Signed-off-by: Timo Kamph <tka@kamph.org>

### DIFF
--- a/sci-libs/amd/amd-2.4.6-r1.ebuild
+++ b/sci-libs/amd/amd-2.4.6-r1.ebuild
@@ -3,6 +3,7 @@
 
 EAPI=7
 
+FORTRAN_NEEDED=fortran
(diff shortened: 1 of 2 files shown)


pinicarus:
Hi,

I'm interested in this PR too, in order to install `dev-ml/opam` without dragging a rebuild of gcc with fortran enabled (which does not seem needed).

Any chance to see this merged soon?

AndrewAmmerlaan:
Thanks :+1:

PR #19091, opened 2021-01-17
dev-db/pgcli: Remove unneded test dep

jsmolic:
* Pgcli doesn't explicitly need python-dateutil, it is pulled in by
  dev-python/pendulum since 25758bbe13f42b824713b0d52661e26d8c6f954d

### DIFF
--- a/dev-db/pgcli/pgcli-3.1.0.ebuild
+++ b/dev-db/pgcli/pgcli-3.1.0.ebuild
@@ -31,7 +31,6 @@ BDEPEND="
 	test? (
 		dev-db/postgresql
 		dev-python/mock[${PYTHON_USEDEP}]
-		dev-python/python-dateutil[${PYTHON_USEDEP}]
 	)"
 
 distutils_enable_tests pytest


juippis:
I on the other hand think it explicitly does need it. What if the next pendulum version bump dropped python-dateutil dep, it'd leave these test cases broken again? We should try to list **all** necessary deps in the \*DEPEND lists.

jsmolic:
> I on the other hand think it explicitly does need it. What if the next pendulum version bump dropped python-dateutil dep, it'd leave these test cases broken again? We should try to list **all** necessary deps in the *DEPEND lists.

Hmm, why would the next pendulum version bump drop python-dateutil? I would remove it only if it's not used anymore by pendulum, but it won't matter for pgcli because it doesn't import python-dateutil :smiley:  OTOH it's a cheap python package so it hardly makes a big difference, but the same thing could happen with e.g. dev-python/mock, if it somehow messes up its deps in the ebuild

juippis:
> for pgcli because it doesn't import python-dateutil 

Ah you are correct, I must've interpreted (and remembered) the error message wrong. This was it: 

```
ImportError while loading conftest '/var/tmp/portage/portage/dev-db/pgcli-3.1.0/work/pgcli-3.1.0/tests/conftest.py'.
tests/conftest.py:3: in <module>
    from utils import (
tests/utils.py:4: in <module>
    from pgcli.main import format_output, OutputSettings
../pgcli-3.1.0-python3_7/lib/pgcli/main.py:18: in <module>
    import pendulum
/usr/lib/python3.7/site-packages/pendulum/__init__.py:26: in <module>
    from .date import Date
/usr/lib/python3.7/site-packages/pendulum/date.py:22: in <module>
    from .exceptions import PendulumException
/usr/lib/python3.7/site-packages/pendulum/exceptions.py:1: in <module>
    from .parsing.exceptions import ParserError  # noqa
/usr/lib/python3.7/site-packages/pendulum/parsing/__init__.py:10: in <module>
    from dateutil import parser
E   ModuleNotFoundError: No module named 'dateutil'
```

jsmolic:
@juippis Yup, thanks :)

PR #19094, opened 2021-01-17
app-emulation/lxd: add systemd-cgroups init file

epsilon-0:
needed for handling systemd hosts on openrc system


epsilon-0:
Unnecessary, this can be properly enforced with `/etc/rc.conf`. 
Information has been now documented and updated in the wiki - https://wiki.gentoo.org/wiki/LXD#Running_systemd_based_containers_on_OpenRC_hosts

PR #19097, opened 2021-01-18
app-emulation/crun: version bump to 0.16

epsilon-0:
also removes dependency on md2man, upstream
provides pregenerated man page.


epsilon-0:
Hmm, seems I forgot to check if there were pending PRs - https://github.com/gentoo/gentoo/pull/18415
In any case, this one at least drops dependency on md2man, so maybe that change can be incorporated.

PR #19098, opened 2021-01-18
sys-boot/efibootmgr: EAPI 7, version bump

telans:
Closes: https://bugs.gentoo.org/587094
Closes: https://bugs.gentoo.org/673536
Closes: https://bugs.gentoo.org/557248


sys-boot/efibootmgr/efibootmgr-17.ebuild
```diff
@@ -0,0 +1,39 @@
...
+	sed -i '/extern int efi_set_verbose/d' src/efibootmgr.c || die
+}
+
+src_configure() {
+	tc-export CC
+	export EFIDIR="Gentoo"
+}
+
+src_compile() {
+	PKG_CONFIG="$(tc-getPKG_CONFIG)" emake
```
floppym, line 38: It looks like you swapped the order of emake and PKG_CONFIG here. This has a negative effect on the build process.

In Make.defaults, PKG_CONFIG is set using the `=` operator. This means that environment variables will not override it, but command line arguments will.

sys-boot/efibootmgr/efibootmgr-17.ebuild
```diff
@@ -0,0 +1,39 @@
...
+DEPEND="${RDEPEND}"
+
+PATCHES=( "${FILESDIR}/${P}-clang.patch" )
+
+src_prepare() {
+	default
+	sed -i 's/-Werror //' Make.defaults || die
+	# Remove conflict with =sys-libs/efivar-37.
+	# See: https://github.com/rhboot/efibootmgr/issues/128
+	sed -i '/extern int efi_set_verbose/d' src/efibootmgr.c || die
```
floppym, line 29: A patch would be preferable here.

sys-boot/efibootmgr/efibootmgr-17.ebuild
telans: Sorry, I meant to put that back.

sys-boot/efibootmgr/efibootmgr-17.ebuild
telans: Done

PR #19102, opened 2021-01-18
sci-libs/lapack: add deprecated functions also to 3.9.0

efferre79:
as commented in https://bugs.gentoo.org/702254#c6

### DIFF
--- a/sci-libs/lapack/lapack-3.9.0-r1.ebuild
+++ b/sci-libs/lapack/lapack-3.9.0-r1.ebuild
@@ -6,7 +6,7 @@ EAPI=7
 inherit cmake
 
 DESCRIPTION="BLAS,CBLAS,LAPACK,LAPACKE reference implementations"
-HOMEPAGE="http://www.netlib.org/lapack/"
+HOMEPAGE="https://www.netlib.org/lapack/"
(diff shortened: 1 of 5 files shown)


sci-libs/lapack/lapack-3.9.0-r1.ebuild
```diff
@@ -0,0 +1,103 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake
+
+DESCRIPTION="BLAS,CBLAS,LAPACK,LAPACKE reference implementations"
+HOMEPAGE="http://www.netlib.org/lapack/"
```
vaukai, line 9: https://netlib.org/lapack/

efferre79: updated

thesamesam:
Thanks! Merged in 681412c77f8c1d58962f8159eb0cd0f648888d70

juippis:
Please note that in future you need to sign-off using your legal name, not a nickname.
https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin

(I see you use it in Bugzilla so should be no problem)

thesamesam:
Yeah, I wasn't that bothered because the changes are actually from Debian. This is just about adding the patch. I could've just dropped the signed-off-by line but I thought I'd give some credit for doing the PR.

PR #19104, opened 2021-01-18
dev-libs/granite: become proxied maintainer; bump to version 5.5.0

xgqt:


dev-libs/granite/granite-5.5.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+			"lib/Widgets/OverlayBar.vala"
+			"lib/Widgets/SeekBar.vala"
+			"lib/Widgets/StorageBar.vala"
+			"lib/Widgets/Toast.vala"
+			"lib/Widgets/Welcome.vala"
+		)
+		for src_file in "${doc_sed_list[@]}"; do
+			sed -ie "s@{{../doc@{{${BUILD_DIR}/doc@g" \
+				"./${src_file}" || die "Failed to fix docs for ./${src_file}"
+		done
```
juippis, line 60: Could this be replaced with a simple `find -exec sed` line?

dev-libs/granite/granite-5.5.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+inherit meson vala xdg
+
+DESCRIPTION="Elementary OS library that extends GTK+"
+HOMEPAGE="https://github.com/elementary/granite"
+SRC_URI="https://github.com/elementary/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="LGPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="doc +introspection"
```
juippis, line 18: I don't understand what this USE flag does. How does granite change depending on the `deps[introspection]` state?

xgqt: @juippis well... `option('introspection', type: 'boolean', value: true, description: 'Whether to build introspection files')`

dev-libs/granite/granite-5.5.0.ebuild
xgqt: > Could this be replaced with a simple `find -exec sed` line?

will do

dev-libs/granite/granite-5.5.0.ebuild
juippis: But you don't have it defined in `src_configure` :)

xgqt: @juippis Blame the last maintainer. >:-)

xgqt: @juippis Oh! Optional introspection support is only on master!

juippis (approved):

xgqt:
Thanks for cooperation.

PR #19106, opened 2021-01-18
app-admin/webapp-config: python3_9 and EAPI 7

vaukai:


ajakk:
Done in 2b88148 and b98629e8, but thanks!

PR #19108, opened 2021-01-18
app-misc/radeontop: bump to 1.3

mercuriete:
Closes: https://bugs.gentoo.org/765529
Signed-off-by: Abel Garcia Dorta <mercuriete@yahoo.es>


mercuriete:
I think that CI error is a false positive because I didn't touch that package.

Can somebody have a look on that? I don't have the knowledge to understand what is happening.

thesamesam:
@mercuriete Please rebase on master, this was an unrelated issue. Thanks!

mercuriete:
@thesamesam thanks I already did it 1 minute ago.

app-misc/radeontop/radeontop-1.3.ebuild
```diff
@@ -0,0 +1,54 @@
+# Copyright 1999-2019 Gentoo Authors
```
mm1ke, line 1: You need to update the header (or use repoman commit)

mm1ke (commented):
With the updated header i'm fine with the bump :)

mercuriete:
I will do It when I have time.
Probably tonight. Thanks for the feedback

app-misc/radeontop/radeontop-1.3.ebuild
mercuriete: @mm1ke 
I did a change on the header by hand...
I am not sure if the change is ok.
Thank you for your review.

mercuriete:
@mm1ke gentle ping.

mm1ke:
Thanks, i'm fine now with the bump.

juippis (approved):

PR #19116, opened 2021-01-18
eclass/toolchain-funcs: new function tc-ld-force-bfd()

telans:
```
No functions currently force ld.bfd usage when ld.lld is active.
This function forces ld.bfd when either ld.gold or ld.lld is active.
tc-ld-disable-gold() now calls this new function
 only if ld.gold is active.
```

### DIFF
--- a/eclass/toolchain-funcs.eclass
+++ b/eclass/toolchain-funcs.eclass
@@ -1,4 +1,4 @@
-# Copyright 2002-2019 Gentoo Authors
+# Copyright 2002-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 # @ECLASS: toolchain-funcs.eclass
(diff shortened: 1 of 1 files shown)


telans:
> New issues caused by PR:

Pretty sure this is wrong

floppym:
> Although I wonder if these two functions should be migrated to something like tc-ld-enable-bfd?

Yes, this seems like a much nicer solution. I would probably call it `tc-ld-force-bfd` instead.

We could also retain `tc-ld-disable-gold` and just have it call `tc-ld-force-bfd` until all ebuilds are migrated to the new function.

telans:
> I would probably call it `tc-ld-force-bfd` instead.

Done, this is definitely better.

floppym (changes requested):
I think you should replace the body of tc-ld-disable-gold with a call to tc-ld-force-bfd. It should also be marked as deprecated.

Please remove "eclass/" from the summary line in the commit message.

This patch should then be sent to the gentoo-dev mailing list for official review.

trofi (approved):

trofi:
Pulled as is. Thank you!

PR #19121, opened 2021-01-19
games-fps/gzdoom: Add '-DNDEBUG' to CXXFLAGS

TheGreatMcPain:
Allows certain mods, like Project Brutality's git master, to work.

See: https://github.com/coelckers/gzdoom/issues/1279

I've also added a 'debug' useflag that skips 'append-cppflags -DNDEBUG' just incase.

### DIFF
--- a/games-fps/gzdoom/gzdoom-4.5.0.ebuild
+++ b/games-fps/gzdoom/gzdoom-4.5.0.ebuild
@@ -1,9 +1,9 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-inherit cmake desktop xdg
+inherit cmake desktop xdg flag-o-matic
 
 DESCRIPTION="A modder-friendly OpenGL source port based on the DOOM engine"
 HOMEPAGE="https://zdoom.org"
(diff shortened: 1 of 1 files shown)


games-fps/gzdoom/metadata.xml
```diff
@@ -14,6 +14,7 @@
 		<name>Gentoo Games Project</name>
 	</maintainer>
 	<use>
+		<flag name="debug">Enable assert() functions for debugging</flag>
```
vilhelmgray, line 17: This description matches the global description of `debug`, so we probably don't need this line added here.

chewi:
Heh, so I've been working on a map with GZDoom for the past year and now you're telling me it was running much slower than it should have been? Can't say I noticed. :wink:

Remove the flag description and I'll get this merged.

TheGreatMcPain:
Done

chewi:
Oh, sorry, I didn't notice your contribution is missing the sign-off. We can't accept it until you add that. It can be done by adding `DCO_SIGNED_OFF_BY='Your Real Name <someone@example.com>'` to your `make.conf`. Please amend your commit and force push.

chewi:
Merged, thanks! I realised afterwards that it's `CPPFLAGS`, not `CXXFLAGS` (these are different things), but it does the right thing anyway. I fired up the game and it might be my imagination but it does really feel that bit smoother now.

PR #19123, opened 2021-01-19
dev-java/openjdk: pass --with-toolchain-type to configure (fix configure with clang)

telans:
`--with-toolchain-type` is needed for ./configure to correctly determine version info from CC. Otherwise configure bails out with:
```
configure: Will use user supplied compiler CC=clang
checking for clang... /usr/lib/llvm/11/bin/clang
checking resolved symbolic links for CC... /usr/lib/llvm/11/bin/clang-11
configure: The C compiler (located as /usr/lib/llvm/11/bin/clang) does not seem to be the required gcc compiler.
configure: The result from running with --version was: ""
configure: error: A gcc compiler is required. Try setting --with-tools-dir.
configure exiting with result code 1
```

"gcc" within "A **gcc** compiler is required" comes from `$TOOLCHAIN_TYPE` which in turn is derived from `$DEFAULT_TOOLCHAIN` when `--with-toolchain-type` is not set.

See: https://github.com/openjdk/jdk/blob/master/doc/building.md#clang

### DIFF
--- a/dev-java/openjdk/openjdk-11.0.9_p11.ebuild
+++ b/dev-java/openjdk/openjdk-11.0.9_p11.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=6
@@ -168,6 +168,7 @@ src_configure() {
 		--with-zlib=system
 		--enable-dtrace=$(usex systemtap yes no)
 		--enable-headless-only=$(usex headless-awt yes no)
+		$(tc-is-clang && echo "--with-toolchain-type=clang")
 	)
 
 	if use javafx; then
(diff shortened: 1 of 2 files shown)


gyakovlev:
it needs a bit more logic.

`configure: Toolchain type powerpc64le-unknown-linux-gnu-gcc is not valid on this platform.`

`tc-get-CC()` will return full triple by default.

so your patch only works if users set `CC=clang/gcc`

if someone sets full triple prefixed, it'll break, gentoo by default sets full form, so patch is broken by default.

I'd do something like
```
$(tc-is-clang && echo "--with-toolchain-type=clang")
```

no need extra condition for gcc, leave it do autofailure/detection.

please test if it works for you and update PR accordingly, so we can proceed.

gyakovlev (changes requested):
see above ^

telans:
Thanks for the suggestion, updated.

gyakovlev:
thanks, merging.
I've extended the PR to new versions as well, forgot about this PR on bump.

gyakovlev:
https://github.com/gentoo/gentoo/commit/9f79570a7ea547821762120fe8fe7eace69ca3c0

PR #19135, opened 2021-01-20
sys-fs/compsize: bump to 1.4, fix LICENSE

telans:
```
sys-fs/compsize: bump to 1.4
    
    ebuild changes:
     - use einstalldocs instead of dodoc
     - update PREFIX instead of sedding "/usr" to Makefile
     - remove append-cflags -std=gnu90
     - add -Wall to debug flags
```

```
    sys-fs/compsize: fix LICENSE
    
    LICENSE in source is GPL-2+, but files such as 'radix-tree.c' are GPL-2
```
See: https://github.com/kilobyte/compsize/blob/master/radix-tree.c

Closed bug is unrelated to changes here but was fixed in sys-fs/compsize-1.3

Closes: https://bugs.gentoo.org/716200


ulm:
>     LICENSE in source is GPL-2+, but files such as 'radix-tree.c' are GPL-2

So the ebuild should have `LICENSE="GPL-2+ GPL-2"`.

telans:
Fixed license

juippis (approved):

PR #19136, opened 2021-01-20
dev-java/openjdk(-jre)(-bin): Set JKS as default trust store format

Flowdalic:
Only a JKS formated trust store is created by
sys-apps/baselayout-java: namely /etc/ssl/certs/java/cacerts. This
file is the symlink target for all JDK/JREs, even though, since Java 9
the default key store format become PKCS12.

Hence we switch the default key store format back to JSK, until
baselayout-java also takes care of creating a trust store in PCS12
format (e.g. /etc/ssl/certs/java/cacerts.p12).

Bug: https://bugs.gentoo.org/712290
RepoMan-Options: --force


Flowdalic:
Obsolete, closing.

PR #19139, opened 2021-01-20
net-analyzer/linkchecker: bump to 10.0.0

hydrapolic:


net-analyzer/linkchecker/linkchecker-10.0.0.ebuild
```diff
@@ -0,0 +1,58 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{6..9} )
```
telans, line 6:
```suggestion
PYTHON_COMPAT=( python3_{7..9} )
```

python 3.6 support has been dropped

net-analyzer/linkchecker/linkchecker-10.0.0.ebuild
```diff
@@ -0,0 +1,58 @@
...
+
+RDEPEND="
+	dev-python/beautifulsoup[${PYTHON_USEDEP}]
+	dev-python/dnspython[${PYTHON_USEDEP}]
+	dev-python/pyxdg[${PYTHON_USEDEP}]
+	dev-python/requests[${PYTHON_USEDEP}]
+"
+
+python_prepare_all() {
+	local PATCHES=(
```
telans, line 36: can `local PATCHES` & `local DOCS` be moved into global scope? would allow removing `python_prepare_all()`

net-analyzer/linkchecker/linkchecker-10.0.0.ebuild
hydrapolic: Python 3.6 is still in portage, if anyone wishes to use it, it's possible. Otherwise if you run a current system then you'll only have python3.8 anyway.

net-analyzer/linkchecker/linkchecker-10.0.0.ebuild
hydrapolic: Previously it was present in the global scope, but it has been noted to move it to local scope instead.

thesamesam: Really? Why?

hydrapolic: To avoid polluting the global scope. However, it can be changed, I don't see a problem in that.

@thesamesam are you too suggesting to switching to global scope?

hydrapolic: Moved to global scope.

hydrapolic:
Thank you @Whissi

hydrapolic:
And thanks @telans for the review

PR #19144, opened 2021-01-21
app-emulation/punes: update to 0.107

winterheart:
Update to 0.107, remove old.


app-emulation/punes/punes-0.107.ebuild
```diff
@@ -0,0 +1,64 @@
...
+	xdg_icon_cache_update
+	xdg_desktop_database_update
+	xdg_mimeinfo_database_update
+}
+
+pkg_postrm() {
+	xdg_icon_cache_update
+	xdg_desktop_database_update
+	xdg_mimeinfo_database_update
+}
```
juippis: In this case it'd be smarter to just inherit *xdg.eclass* and either remove these phases or simply call `xdg_pkg_postinst` etc.

juippis (approved):

PR #19156, opened 2021-01-22
net-misc/icaclient: create basic version info file

henning-schild:
The tarball actually contains a rather complicated script to fill in all
sorts of fields in the file. It is called "hinst".
Instead of trying to execute it simply create a file that is good enough
for "wfica -version".

Bug: https://bugs.gentoo.org/766504

### DIFF
--- a/net-misc/icaclient/icaclient-20.12.0.12-r2.ebuild
+++ b/net-misc/icaclient/icaclient-20.12.0.12-r2.ebuild
@@ -1,0 +1,252 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+# eutils inherit required for make_wrapper call
+inherit desktop eutils multilib xdg-utils
+
+DESCRIPTION="ICA Client for Citrix Presentation servers"
+HOMEPAGE="https://www.citrix.com/"
+SRC_URI="amd64? ( linuxx64-${PV}.tar.gz )
+	x86? ( linuxx86-${PV}.tar.gz )"
+
+LICENSE="icaclient"
+SLOT="0"
+KEYWORDS="-* ~amd64 ~x86"
+IUSE="l10n_de l10n_es l10n_fr l10n_ja l10n_zh-CN"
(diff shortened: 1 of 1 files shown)


henning-schild:
A full version info file would look like this, for 20.12 installed in a container
```
ID_VERSION=20.12.0.12
DISP_VERSION=20.12.0
CHANGE_NO=12
ID_PRODUCT_NAME=CitrixICAClient
DISP_PRODUCT_NAME=Citrix Workspace app for Linux
uname=Linux e2084ece1154 5.4.80-gentoo-r1 #2 SMP PREEMPT Mon Dec 21 17:34:27 CET 2020 x86_64 x86_64 x86_64 GNU/Linux
PORT=linuxx64
LANGUAGE=*
date=20210122101847 CET
INFO_FILE=Debian_installation
DATA_FILE=Debian_package
```

But let us hope a simple one will also do, or we have to try and execute "hinst" during build, or "rewrite" its `put_version_info`

In fact an easier to read source for such a file is the postinst script from the deb packages, there it is called `set_version_info`

net-misc/icaclient/icaclient-20.12.0.12-r2.ebuild
```diff
@@ -0,0 +1,256 @@
...
+	echo "SEARCH_DIRS_MASK=\"${ICAROOT}\"" \
+		> "${ED}"/etc/revdep-rebuild/70icaclient
+
+	ICAPORT="linuxx64"
+	if use x86; then
+		ICAPORT="linuxx86"
+	fi
+	dodir "${ICAROOT}"/pkginf
+	insinto "${ICAROOT}"/pkginf
+	newins "${WORKDIR}"/PkgId Ver.core."${ICAPORT}"
```
Polynomial-C, line 238: Instead of that newly introduced `ICAPORT` variable did you try to use the `ICAARCH` variable that is also being used in `pkg_setup()`? That would avoid duplicating code that is already present.

henning-schild:
I guess this should also be done to 20.10, let me know if that guess is correct and i will do so

PR #19157, opened 2021-01-22
net-proxy/mitmproxy: Bump to 6.0.2

00-matt:


00-matt:
Rebased after ~arm64 keyword

PR #19159, opened 2021-01-22
dev-php/pecl-event: stabilized version 2.5.7

rosmanov:
Closes: https://bugs.gentoo.org/721122
Signed-off-by: Ruslan Osmanov <rrosmanov@gmail.com>


thesamesam:
Maintainers can self-stabilise on amd64/x86 but usually for proxy maintainers we do it via the arch teams instead. So, those bugs are assigned to you, but you don't need to do anything other than object/give consent (maybe the version isn't ready, too buggy, etc) on the bug.

Thank you!

PR #19164, opened 2021-01-22
app-emulation/crun: bump to 0.17

robertgzr:
Signed-off-by: Robert Günzler <r@gnzler.io>

---

I feel like I'm spamming you with PRs lol


app-emulation/crun/crun-0.17.ebuild
```diff
@@ -14,20 +14,19 @@ SRC_URI="https://github.com/containers/${PN}/releases/download/${PV}/${P}.tar.gz
...
 	sys-kernel/linux-headers
 	>=dev-libs/yajl-2.0.0
 	caps? ( sys-libs/libcap )
-	criu? ( >=sys-process/criu-3.13 )
+	criu? ( >=sys-process/criu-3.15 )
 	seccomp? ( sys-libs/libseccomp )
 	systemd? ( sys-apps/systemd:= )
 "
 RDEPEND="${DEPEND}"
 BDEPEND="
```
telans, line 28: I would just remove BDEPEND

robertgzr: do I not need PYTHON_DEPS there? according to https://dev.gentoo.org/~mgorny/python-guide/any.html I do...

the build system depends on it:
https://github.com/containers/crun/blob/0.17/configure.ac#L21

juippis (approved):

PR #19165, opened 2021-01-22
sys-apps/openrc: allow devfs in lxc

epsilon-0:

---

allows running development containers with pkgcheck, which currently fails due to lack of `/dev/shm`


floppym:
You should submit the change upstream first. Also, I left a comment on https://github.com/OpenRC/openrc/issues/272#issuecomment-768036295.

PR #19172, opened 2021-01-23
app-admin/cdist: Bump to 6.9.4 

jsmolic:


app-admin/cdist/files/cdist-6.9.4-tests.patch
```diff
@@ -0,0 +1,44 @@
+--- a/cdist/preos.py
++++ b/cdist/preos.py
+@@ -7,7 +7,7 @@ import cdist
+ import logging
+ import cdist.argparse
+ import cdist.configuration
+-import cdist.exec.util as util
++from cdist.exec import util
```
SoapGentoo: check whether this is still necessary, seeing that python3.6 is gone now

SoapGentoo (changes requested):

app-admin/cdist/files/cdist-6.9.4-tests.patch
jsmolic: We can drop this patch along with 3.6

jsmolic: Dropped

PR #19174, opened 2021-01-23
dev-cpp/cpp-taskflow: Bump to 3.0.0 

jsmolic:


dev-cpp/cpp-taskflow/cpp-taskflow-3.0.0.ebuild
```diff
@@ -12,42 +12,30 @@ SRC_URI="https://github.com/${PN}/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
...
 }
 
 src_install() {
-	cmake-utils_src_install
-
-	if $(use doc); then
-		einstalldocs
-	fi
+	cmake_src_install
+	use doc && einstalldocs
```
SoapGentoo, line 40: this is dead code, as `cmake_src_install` calls `einstalldocs` internally

SoapGentoo (changes requested):

dev-cpp/cpp-taskflow/cpp-taskflow-3.0.0.ebuild
```diff
@@ -12,42 +12,30 @@ SRC_URI="https://github.com/${PN}/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
 LICENSE="MIT"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE="doc"
-
-RDEPEND=""
-DEPEND=""
+IUSE="examples doc test"
```
SoapGentoo, line 15: where is `USE=test` used in the ebuild?

SoapGentoo (changes requested):

dev-cpp/cpp-taskflow/cpp-taskflow-3.0.0.ebuild
jsmolic: It got lost during the process :)

dev-cpp/cpp-taskflow/cpp-taskflow-3.0.0.ebuild
jsmolic: Ah that's right. Removed

dev-cpp/cpp-taskflow/cpp-taskflow-3.0.0.ebuild
```diff
@@ -12,42 +12,28 @@ SRC_URI="https://github.com/${PN}/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
...
 
 src_prepare() {
-	default
-
 	# fix library directoy
 	sed -i "s#/lib#/$(get_libdir)#g" CMakeLists.txt || die "sed failed"
 
-	cmake-utils_src_prepare
+	cmake_src_prepare
 }
```
SoapGentoo, line 28: do you still need to override this whole phase function?

SoapGentoo (changes requested):

dev-cpp/cpp-taskflow/cpp-taskflow-3.0.0.ebuild
jsmolic: Actually no since we patched this out, I'll remove it

jsmolic: Done

PR #19180, opened 2021-01-23
sci-libs/libecpint: Bump to 1.0.4

jsmolic:


sci-libs/libecpint/libecpint-1.0.4.ebuild
```diff
@@ -15,12 +15,12 @@ KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
 IUSE="test"
 RESTRICT="!test? ( test )"
```
SoapGentoo, line 16: order of this block

SoapGentoo (changes requested):

sci-libs/libecpint/libecpint-1.0.4.ebuild
jsmolic: Done

SoapGentoo (approved):

SoapGentoo (approved):

PR #19183, opened 2021-01-23
www-misc/monitorix: Version bump to 3.13.1

Dwosky:
Signed-off-by: Pedro Arizmendi <dwosky@pm.me>


Dwosky:
Version 3.13.1 was just published with some bugfix, so I've updated the ebuild to use that version instead.

juippis (approved):

PR #19189, opened 2021-01-24
gui-apps/waybar: sync live

myrvogna:
Syncing the live ebuild with the latest (0.9.5) ebuild.

### DIFF
--- a/gui-apps/waybar/waybar-9999.ebuild
+++ b/gui-apps/waybar/waybar-9999.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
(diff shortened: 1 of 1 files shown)


gui-apps/waybar/waybar-9999.ebuild
```diff
@@ -56,6 +57,8 @@ src_configure() {
 		$(meson_feature pulseaudio)
 		$(meson_feature tray dbusmenu-gtk)
 		$(meson_feature udev libudev)
+		$(meson_feature wifi rfkill)
+		-Dsndio=disabled
```
thesamesam, line 61: @JonasToth FYI sndiod is now in tree, so we may want to test and expose this?

thesamesam:
Thanks!

gui-apps/waybar/waybar-9999.ebuild
JonasToth: Thank you very much!
I will try to take a look at it this weekend. But my time is currently low :/

PR #19194, opened 2021-01-24
app-admin/newsyslog: Fix build with gcc-10, bump sys-devel/bmake

jsmolic:


app-admin/newsyslog/newsyslog-1.1.0.81-r1.ebuild
```diff
@@ -11,22 +11,23 @@ LICENSE="MIT"
 SLOT="0"
 KEYWORDS="~alpha ~amd64 ~hppa ~mips ~ppc ~ppc64 ~sparc ~x86"
 
-DEPEND="
-	sys-apps/groff"
-
+DEPEND="sys-apps/groff"
```
thesamesam, line 14: Not BDEPEND?

SoapGentoo, line 14: sure `groff` is not a `BDEPEND`?

app-admin/newsyslog/newsyslog-1.1.0.81-r1.ebuild
```diff
@@ -11,22 +11,23 @@ LICENSE="MIT"
...
 
 DOCS=( newsyslog.conf AUTHORS ChangeLog INSTALL NEWS ToDo )
 
 src_configure() {
-	local myconf="--with-syslogd_pid=/var/run/syslog.pid"
-
 	if has_version 'app-admin/syslog-ng'; then
 		myconf="--with-syslogd_pid=/var/run/syslog-ng.pid"
+	else
+		myconf="--with-syslogd_pid=/var/run/syslog.pid"
```
thesamesam, line 30: I wonder if these need EPREFIXifying. I really dislike using has_version because of how fragile it is, but I guess we can't do anything about that here.

app-admin/newsyslog/newsyslog-1.1.0.81-r1.ebuild
```diff
@@ -11,22 +11,23 @@ LICENSE="MIT"
...
 	fi
```
SoapGentoo, line 31: can't you do this as a bash array instead?

SoapGentoo (changes requested):

app-admin/newsyslog/newsyslog-1.1.0.81-r1.ebuild
SoapGentoo: I agree, most `has_version` patterns I see promote subtle "config desyncing" where replacing what it depends on gets changed later, and then everything breaks. The correct solution IMO would be to do this via USE flags, but that's too involved for the fixes @jsmolic is pushing.

app-admin/newsyslog/newsyslog-1.1.0.81-r1.ebuild
jsmolic: Sure, it's a BDEPEND

app-admin/newsyslog/newsyslog-1.1.0.81-r1.ebuild
jsmolic: Sure

PR #19203, opened 2021-01-24
games-util/xpadneo: new package

ionenwks:
xpadneo is an "Advanced Linux Driver for Xbox One Wireless Controller", also includes other derived gamepads.

Upstream will do a better job at explaining reasons to use it:
https://atar-axis.github.io/xpadneo/
https://github.com/atar-axis/xpadneo

Upstream is very active, has a good release cycle, and driver is fairly mature.

Been using for half a year in my local overlay and thought it'd make a fine addition to Gentoo where I'll take care to maintain.

For naming, overlays have used both games-util/xpadneo and sys-kernel/xpadneo but using the former (ancient xboxdrv uses it as well, albeit it's uinput).


ionenwks:
Just fixing minor nitpicks happened to spot, no real changes.

PR #19204, opened 2021-01-25
eclass: Fixing user/group creation when using different ROOT

Mord3rca:
When creating a user for another environement, user is created on the HOST and not the ROOT dir.
Adding "-R <CHROOT_DIR>" for all user* / group* commands fix the issue.

Signed-off-by: Mord3rca <morderca@morderca.net>

### DIFF
--- a/eclass/acct-user.eclass
+++ b/eclass/acct-user.eclass
@@ -195,8 +195,15 @@ eislocked() {
 	*)
 		# NB: 'no password' and 'locked' are indistinguishable
 		# but we also expire the account which is more clear
-		[[ $(getent shadow "$1" | cut -d: -f2) == '!'* ]] &&
-			[[ $(getent shadow "$1" | cut -d: -f8) == 1 ]]
+		local shadow
+		if [[ -n "${ROOT}" ]]; then
+			shadow=$(grep "^$1:" "${ROOT}/etc/shadow")
+		else
+			shadow=$(getent shadow "$1")
+		fi
+
+		[[ $( echo ${shadow} | cut -d: -f2) == '!'* ]] &&
+			[[ $(echo ${shadow} | cut -d: -f8) == 1 ]]
 		;;
 	esac
 }
@@ -223,14 +230,22 @@ elockuser() {
 	eislocked "$1"
 	[[ $? -eq 0 ]] && return 0
 
+	local opts
+	[[ -n ${ROOT} ]] && opts=( --prefix "${ROOT}" )
+
 	case ${CHOST} in
 	*-freebsd*|*-dragonfly*)
-		pw lock "$1" || die "Locking account $1 failed"
-		pw user mod "$1" -e 1 || die "Expiring account $1 failed"
+		pw lock "${opts[@]}" "$1" || die "Locking account $1 failed"
+		pw user mod "${opts[@]}" "$1" -e 1 || die "Expiring account $1 failed"
 		;;
 
 	*-netbsd*)
-		usermod -e 1 -C yes "$1" || die "Locking account $1 failed"
+		if [[ -n "${ROOT}" ]]; then
+			ewarn "NetBSD's usermod does not support --prefix <dir> option."
+			ewarn "Please use: usermod ${opts[@]} -e 1 -C yes \"$1\" in a chroot"
+		else
+			usermod "${opts[@]}" -e 1 -C yes "$1" || die "Locking account $1 failed"
+		fi
 		;;
 
 	*-openbsd*)
@@ -238,7 +253,7 @@ elockuser() {
 		;;
 
 	*)
-		usermod -e 1 -L "$1" || die "Locking account $1 failed"
+		usermod "${opts[@]}" -e 1 -L "$1" || die "Locking account $1 failed"
 		;;
 	esac
 
@@ -266,14 +281,22 @@ eunlockuser() {
 	eislocked "$1"
 	[[ $? -eq 1 ]] && return 0
 
+	local opts
+	[[ -n ${ROOT} ]] && opts=( --prefix "${ROOT}" )
+
 	case ${CHOST} in
 	*-freebsd*|*-dragonfly*)
-		pw user mod "$1" -e 0 || die "Unexpiring account $1 failed"
-		pw unlock "$1" || die "Unlocking account $1 failed"
+		pw user mod "${opts[@]}" "$1" -e 0 || die "Unexpiring account $1 failed"
+		pw unlock "${opts[@]}" "$1" || die "Unlocking account $1 failed"
 		;;
 
 	*-netbsd*)
-		usermod -e 0 -C no "$1" || die "Unlocking account $1 failed"
+		if [[ -n "${ROOT}" ]]; then
+			ewarn "NetBSD's usermod does not support --prefix <dir> option."
+			ewarn "Please use: \"usermod ${opts[@]} -e 0 -C no $1\" in a chroot"
+		else
+			usermod "${opts[@]}" -e 0 -C no "$1" || die "Unlocking account $1 failed"
+		fi
 		;;
 
 	*-openbsd*)
@@ -282,7 +305,7 @@ eunlockuser() {
 
 	*)
 		# silence warning if account does not have a password
-		usermod -e "" -U "$1" 2>/dev/null || die "Unlocking account $1 failed"
+		usermod "${opts[@]}" -e "" -U "$1" 2>/dev/null || die "Unlocking account $1 failed"
 		;;
 	esac
 
@@ -418,7 +441,13 @@ acct-user_pkg_preinst() {
 		# default ownership to user:group
 		if [[ -z ${_ACCT_USER_HOME_OWNER} ]]; then
 			local group_array=( ${_ACCT_USER_GROUPS} )
-			_ACCT_USER_HOME_OWNER=${ACCT_USER_NAME}:${group_array[0]}
+			if [[ -n "${ROOT}" ]]; then
+				local euid=$(egetent passwd ${ACCT_USER_NAME} | cut -d: -f3)
+				local egid=$(egetent passwd ${ACCT_USER_NAME} | cut -d: -f4)
+				_ACCT_USER_HOME_OWNER=${euid}:${egid}
+			else
+				_ACCT_USER_HOME_OWNER=${ACCT_USER_NAME}:${group_array[0]}
+			fi
 		fi
 		# Path might be missing due to INSTALL_MASK, etc.
 		# https://bugs.gentoo.org/691478
--- a/eclass/user-info.eclass
+++ b/eclass/user-info.eclass
@@ -23,6 +23,7 @@ _USER_INFO_ECLASS=1
 # dscl (Mac OS X 10.5), and pw (FreeBSD) used in enewuser()/enewgroup().
 #
 # Supported databases: group passwd
+# Warning: This function can be used only in pkg_* phases when ROOT is valid.
 egetent() {
 	local db=$1 key=$2
 
@@ -43,18 +44,31 @@ egetent() {
 		# lookup by uid/gid
 		local opts
 		if [[ ${key} == [[:digit:]]* ]] ; then
-			[[ ${db} == "user" ]] && opts="-u" || opts="-g"
+			[[ ${db} == "user" ]] && opts=( -u ) || opts=( -g )
 		fi
+
+		# Handle different ROOT
+		[[ -n ${ROOT} ]] && opts+=( -R "${ROOT}" )
 
 		pw show ${db} ${opts} "${key}" -q
 		;;
 	*-openbsd*)
-		grep "${key}:\*:" /etc/${db}
+		grep "${key}:\*:" "${EROOT}/etc/${db}"
 		;;
 	*)
-		# ignore nscd output if we're not running as root
-		type -p nscd >/dev/null && nscd -i "${db}" 2>/dev/null
-		getent "${db}" "${key}"
+		# getent does not support -R option, if we are working on a different
+		# ROOT than /, fallback to grep technique.
+		if [[ -z ${ROOT} ]]; then
+			# ignore nscd output if we're not running as root
+			type -p nscd >/dev/null && nscd -i "${db}" 2>/dev/null
+			getent "${db}" "${key}"
+		else
+			if [[ ${key} =~ ^[[:digit:]]+$ ]]; then
+				grep -E "^([^:]*:){2}${key}" "${ROOT}/etc/${db}"
+			else
+				grep "^${key}:" "${ROOT}/etc/${db}"
+			fi
+		fi
 		;;
 	esac
 }
@@ -151,7 +165,16 @@ egetgroups() {
 	[[ $# -eq 1 ]] || die "usage: egetgroups <user>"
 
 	local egroups_arr
-	read -r -a egroups_arr < <(id -G -n "$1")
+	if [[ -n "${ROOT}" ]]; then
+		local pgroup=$(egetent passwd "$1" | cut -d: -f1)
+		local sgroups=( $(grep -E ":([^:]*,)?$1(,[^:]*)?$" "${ROOT}/etc/group" | cut -d: -f1) )
+
+		# Remove primary group from list
+		sgroups=${sgroups#${pgroup}}
+		egroups_arr=( ${pgroup} ${sgroups[@]} )
+	else
+		read -r -a egroups_arr < <(id -G -n "$1")
+	fi
 
 	local g groups=${egroups_arr[0]}
 	# sort supplementary groups to make comparison possible
--- a/eclass/user.eclass
+++ b/eclass/user.eclass
@@ -117,6 +117,9 @@ enewuser() {
 	# options to pass to useradd
 	local opts=()
 
+	# handle for ROOT != /
+	[[ -n ${ROOT} ]] && opts+=( --prefix "${ROOT}" )
+
 	# handle uid
 	local euid=${1}; shift
 	if [[ -n ${euid} && ${euid} != -1 ]] ; then
@@ -207,13 +210,24 @@ enewuser() {
 		;;
 
 	*-netbsd*)
-		useradd "${opts[@]}" "${euser}" || die
+		if [[ -n "${ROOT}" ]]; then
+			ewarn "NetBSD's usermod does not support --prefix option."
+			ewarn "Please use: \"useradd ${opts[@]} ${euser}\" in a chroot"
+		else
+			useradd "${opts[@]}" "${euser}" || die
+		fi
 		;;
 
 	*-openbsd*)
-		# all ops the same, except the -g vs -g/-G ...
-		useradd -u ${euid} -s "${eshell}" \
-			-d "${ehome}" -g "${egroups}" "${euser}" || die
+		if [[ -n "${ROOT}" ]]; then
+			ewarn "OpenBSD's usermod does not support --prefix option."
+			ewarn "Please use: \"useradd ${opts[@]} ${euser}\" in a chroot"
+		else
+			# all ops the same, except the -g vs -g/-G ...
+			useradd -u ${euid} -s "${eshell}" \
+				-d "${ehome}" -g "${egroups}" "${euser}" || die
+		fi
+
 		;;
 
 	*)
@@ -224,6 +238,10 @@ enewuser() {
 	if [[ -n ${create_home} && ! -e ${ROOT}/${ehome} ]] ; then
 		elog " - Creating ${ehome} in ${ROOT}"
 		mkdir -p "${ROOT}/${ehome}"
+		# Use UID if we are in another ROOT than /
+		if [[ -n "${ROOT}" ]]; then
+			euser=$(egetent passwd ${euser} | cut -d: -f3)
+		fi
 		chown "${euser}" "${ROOT}/${ehome}"
 		chmod 755 "${ROOT}/${ehome}"
 	fi
@@ -286,6 +304,10 @@ enewgroup() {
 	fi
 	elog " - Groupid: ${egid}"
 
+	# handle different ROOT
+	local opts
+	[[ -n ${ROOT} ]] && opts=( --prefix "${ROOT}" )
+
 	# handle extra
 	if [[ $# -gt 0 ]] ; then
 		die "extra arguments no longer supported; please file a bug"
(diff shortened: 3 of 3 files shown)


Mord3rca:
To test it:
* Create a new rootfs:
```sh
export ROOT=/tmp/gentoo
mkdir -p ${ROOT}
emerge baselayout coreutils
```
 * Install a new user which is not already in the base layout:
 ```sh
 emerge acct-user/sshd
 ```
 
 Test with and without the modification:
 * Without: User is installed in /
 * With: User is installed in ${ROOT}

eclass/user-info.eclass
```diff
@@ -37,18 +37,27 @@ egetent() {
...
 		fi
 
+		# Handle different ROOT
+		[[ $ROOT == "/" ]] || opts+=( -R "${ROOT}" )
+
 		pw show ${db} ${opts} "${key}" -q
 		;;
 	*-openbsd*)
-		grep "${key}:\*:" /etc/${db}
+		grep "${key}:\*:" ${ROOT}/etc/${db}
```
telans, line 49:
```suggestion
		grep "${key}:\*:" "${ROOT}"/etc/${db}
```

should this be quoted?

mgorny:
First of all, we're gonna need your real legal name on the signoff. Without that, I won't even look at the code since that could cause copyright problems if I had to reimplement it independently.

eclass/user-info.eclass
Mord3rca: Good catch, I will quote the whole string

eclass/acct-user.eclass
```diff
@@ -222,22 +222,25 @@ elockuser() {
 	eislocked "$1"
 	[[ $? -eq 0 ]] && return 0
 
+	local opts
+	[[ ${ROOT} == "/" ]] ||	opts=( -R "${ROOT}" )
```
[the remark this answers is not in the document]
Mord3rca: Just tested that and it doesn't seems to work as expected

Mord3rca:
> First of all, we're gonna need your real legal name on the signoff. Without that, I won't even look at the code since that could cause copyright problems if I had to reimplement it independently.

Sure, I'll update my commit message then.

eclass/acct-user.eclass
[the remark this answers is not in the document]
mgorny: In any case, please use positive logic, e.g. `condition && ...`. Otherwise, it's going to be quite confusing.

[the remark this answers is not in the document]
mgorny: Also, it's weird it doesn't work for you. Could you use a `die "${ROOT}"` to debug this? ;-)

eclass/acct-user.eclass
```diff
@@ -222,22 +222,25 @@ elockuser() {
...
+
 	case ${CHOST} in
 	*-freebsd*|*-dragonfly*)
-		pw lock "$1" || die "Locking account $1 failed"
-		pw user mod "$1" -e 1 || die "Expiring account $1 failed"
+		pw lock "${opts[@]}" "$1" || die "Locking account $1 failed"
```
mgorny, line 230: Have you verified (e.g. by reading manpages) that all these tools support `-R`?

Mord3rca: Yes, https://www.freebsd.org/cgi/man.cgi?pw(8)
First option in the description section.

But I did not test that, unfortunately :/

eclass/acct-user.eclass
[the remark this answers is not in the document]
Mord3rca: @mgorny My bad, it is working ^^" Probably missed something while fixing that quickly.

eclass/acct-user.eclass
mgorny: FreeBSD seems to work fine but NetBSD doesn't have `-R` option. I guess you have to make non-/ ROOT support OS-conditional then.

eclass/acct-user.eclass
```diff
@@ -222,22 +222,25 @@ elockuser() {
 	eislocked "$1"
 	[[ $? -eq 0 ]] && return 0
 
+	local opts
+	[[ ! -z ${ROOT} ]] && opts=( -R "${ROOT}" )
```
mgorny, line 226:
```suggestion
	[[ -n ${ROOT} ]] && opts=( -R "${ROOT}" )
```

and below. Really easier to understand stuff without unnecessary negations.

eclass/user-info.eclass
```diff
@@ -37,18 +37,27 @@ egetent() {
 		# lookup by uid/gid
 		local opts
 		if [[ ${key} == [[:digit:]]* ]] ; then
-			[[ ${db} == "user" ]] && opts="-u" || opts="-g"
+			[[ ${db} == "user" ]] && opts=( "-u" ) || opts=( "-g" )
```
mgorny, line 40:
```suggestion
			[[ ${db} == "user" ]] && opts=( -u ) || opts=( -g )
```

eclass/user-info.eclass
```diff
@@ -37,18 +37,27 @@ egetent() {
...
 		fi
 
+		# Handle different ROOT
+		[[ ! -z ${ROOT} ]] && opts+=( -R "${ROOT}" )
+
 		pw show ${db} ${opts} "${key}" -q
 		;;
 	*-openbsd*)
-		grep "${key}:\*:" /etc/${db}
+		grep "${key}:\*:" "${ROOT}/etc/${db}"
```
mgorny, line 49:
```suggestion
		grep "${key}:\*:" "${EROOT}/etc/${db}"
```

I guess.

eclass/user-info.eclass
```diff
@@ -37,18 +37,27 @@ egetent() {
 		# lookup by uid/gid
 		local opts
 		if [[ ${key} == [[:digit:]]* ]] ; then
-			[[ ${db} == "user" ]] && opts="-u" || opts="-g"
+			[[ ${db} == "user" ]] && opts=( "-u" ) || opts=( "-g" )
 		fi
 
+		# Handle different ROOT
+		[[ ! -z ${ROOT} ]] && opts+=( -R "${ROOT}" )
```
mgorny, line 44: Since ROOT is only valid in `pkg_*` phases, I wonder if we should formally ban the function in other phases.

eclass/user-info.eclass
```diff
@@ -37,18 +37,27 @@ egetent() {
...
-		type -p nscd >/dev/null && nscd -i "${db}" 2>/dev/null
-		getent "${db}" "${key}"
+		# getent does not support -R option, if we are working on a different
+		# ROOT than /, fallback to grep technique.
+		if [[ -z ${ROOT} ]]; then
+			# ignore nscd output if we're not running as root
+			type -p nscd >/dev/null && nscd -i "${db}" 2>/dev/null
+			getent "${db}" "${key}"
+		else
+			grep "^${key}" "${ROOT}/etc/${db}"
```
mgorny, line 59: This regex is definitely too weak — `egetent passwd foo` would match `foobar`.

eclass/user.eclass
```diff
@@ -282,6 +285,10 @@ enewgroup() {
 	fi
 	elog " - Groupid: ${egid}"
 
+	# handle different ROOT
+	local chroot_cmd
+	[[ ! -z ${ROOT} ]] && chroot_cmd=( -R "${ROOT}" )
```
mgorny, line 290: That var name is weird, tbh.

mgorny (changes requested):
I'm sorry that it took me this long to review it.

eclass/user-info.eclass
Mord3rca: I guess that: "^${key}:.*:" should do the trick.
+ I need to replace ROOT by EROOT like in the previous comment ?

eclass/user-info.eclass
Mord3rca: How can I do that ? Just by notify it in the doc string ? or with a check ?

eclass/acct-user.eclass
Mord3rca: For the netBSD stuff, this is actually processed by the *) condition right ?
In that case, do I need to process the user/group creation with awk ? Or just printing a warning telling that the user/group is actually not created in the target and should be added manually ? (with `chroot <command>` assuming ARCH is compatible with the host system)

eclass/user-info.eclass
mgorny: Yes, I guess so. Maybe even `^${key}:`

eclass/user-info.eclass
mgorny: Truth is, I'm not sure. Maybe docstring will suffice for now.

eclass/acct-user.eclass
mgorny: Just issue a warning, like the one we're already issuing if we can't do stuff or it fails.

Mord3rca:
@mgorny PR updated following your comments. If you're OK with the modification, I'll cleanup my git history by rebasing/squashing the commits (keeping only the first message)

mgorny:
I think it's good enough to run through official review now. Please squash into one commit, split into one commit per eclass and then send to gentoo-dev ml for review.

thesamesam:
Let's get this sent to the gentoo-dev mailing list for full review & get this in.

dkjii-g (changes requested):
I cannot seem to add comments to parts you didn't change, but any function using usernames in these files will not work correctly using $EROOT. This includes fowners and chown, which will use the host's uid/gid instead of the $EROOT one

in user-info.eclass, egetgroups will not work with different ROOT

eclass/acct-user.eclass
```diff
@@ -222,22 +222,30 @@ elockuser() {
 	eislocked "$1"
 	[[ $? -eq 0 ]] && return 0
 
+	local opts
+	[[ -n ${ROOT} ]] && opts=( -R "${ROOT}" )
```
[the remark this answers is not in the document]
Mord3rca: According to shadow code: https://github.com/shadow-maint/shadow/blob/master/libmisc/root_flag.c#L92
`-R` only drop privilege / chdir / chroot, so if the sandbox is bypassed, it was already the case.

eclass/user.eclass
```diff
@@ -203,7 +206,12 @@ enewuser() {
...
-		useradd "${opts[@]}" "${euser}" || die
+		if [[ -n "${ROOT}" ]]; then
+			ewarn "NetBSD's usermod does not support -R option."
+			ewarn "Please use: \"useradd ${opts[@]} ${euser}\" in a chroot"
+		else
+			useradd "${opts[@]}" "${euser}" || die
+		fi
 		;;
 
 	*-openbsd*)
```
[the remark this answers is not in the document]
Mord3rca: Like netbsd, openbsd does not support the `-R` flag, I'll add the ewarn message since I've forgotten to add it

eclass/user-info.eclass
```diff
@@ -37,18 +37,27 @@ egetent() {
...
-		type -p nscd >/dev/null && nscd -i "${db}" 2>/dev/null
-		getent "${db}" "${key}"
+		# getent does not support -R option, if we are working on a different
+		# ROOT than /, fallback to grep technique.
+		if [[ -z ${ROOT} ]]; then
+			# ignore nscd output if we're not running as root
+			type -p nscd >/dev/null && nscd -i "${db}" 2>/dev/null
+			getent "${db}" "${key}"
+		else
+			grep "^${key}:" "${EROOT}/etc/${db}"
```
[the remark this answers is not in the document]
Mord3rca: If it is a UID, I can do something like this:
```sh
grep "^${key}:" "${EROOT}/etc/${db}" || \
  grep "^*:x:${key}:" "${EROOT}/etc/${db}"
```
It should fail if the key isn't found because it is a UID and so, try again by searching by UID this time.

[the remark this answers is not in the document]
dkjii-g: Could do if [[ ... =~ [[:digit:]* ]] like above for the freebsd

I don't think the x is guaranteed though, you could use awk to extract the uid like in that patch in the bug tracker

eclass/acct-user.eclass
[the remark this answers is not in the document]
dkjii-g: I meant that chroot'ing executes code from the $ROOT which you might not expect. Plus this is in an unsandboxed function (since {pre,post}{inst,rm} can change / even if $ROOT != /)

Also, if we're gonna chroot using individual tools we might as well run all of {pre,post}{inst,rm} in a chroot and this would fix all of these kinds of issues (and make these kind of changes unnecessary)

I'm not a gentoo-dev though so idk what's the concensus on this

We use --prefix in the patches from the bugtracker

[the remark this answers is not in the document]
Mord3rca: Agreed on the fact that all of this depend of what is installed on $ROOT (see the case separation with BSD)

The chroot idea is not bad but is limited to a set of compatible architecture between the host and the target (AMD64 -> X86 / AMD64 for example).
If you want to compile another architecture which is not natively compatible with your host, like AARCH64 for AMD64 host, you will have to use a Qemu/KVM trick which can works but, on my side, can have some heavy performance issue (+ you need to have a stage3 already decompressed).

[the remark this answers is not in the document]
dkjii-g: Yeah that's fair, realized that after posting. But that applies to shadow too, have you tested this with non-compatible targets? usermod -R is basically a glorified chroot so I don't think this is different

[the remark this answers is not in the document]
dkjii-g: Aka the difference between -R and --prefix seems to be NIS/LDAP/pam support, but those require the TARGET's binaries to be supported I assume (pam library/niscd/sssd)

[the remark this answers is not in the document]
Mord3rca: My test are actually done with: `aarch64-linux-gnu-emerge`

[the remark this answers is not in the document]
Mord3rca: I did further testing by decompressing a aarch64 stage3 and adding a user.
You're right, `-R` fails where `--prefix` works. After that, the home creation failed since chown failed to get the user (if it is not installed on the host).

eclass/user-info.eclass
dkjii-g: From talking to people on IRC, I think you need $ROOT here and not $EROOT, since the target system (under $ROOT) will not even check $EPREFIX for users. $EPREFIX != / should output an error for usermod operations IMO and user-info should use $ROOT and not $EROOT

eclass/acct-user.eclass
dkjii-g: If you want to be compatibly with both eapi == 7 and eapi != 7, I think you need `[[ -n "${ROOT}" ]] && [[ "${ROOT}" != "/" ]]`

eclass/user-info.eclass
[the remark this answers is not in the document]
Mord3rca:
```sh
if [[ ${key} =~ ^[[:digit:]]+$ ]]; then
	grep -E "^([^:]*:){2}${key}" "${ROOT}/etc/${db}"
else
	grep "^${key}:" "${ROOT}/etc/${db}"
fi
```

This should do the trick.

eclass/acct-user.eclass
```diff
@@ -222,22 +222,30 @@ elockuser() {
 	eislocked "$1"
```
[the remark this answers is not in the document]
Mord3rca: Since egetent does not support shadow (yet), is it safe to read directly `${ROOT}/etc/shadow` and search for a '!' in the second field ?

Mord3rca:
@dkjii-g PR rebased & updated.

I've tested on my side for AARCH64 / AMD64.
The user creation is working but there is an error for `esetgroups` which does not seems to be critical (home created, user is fine, ....). Maybe if a user have multiple groups, it can be an issue.

eclass/user-info.eclass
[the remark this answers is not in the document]
Mord3rca: btw: `[[ ${key} =~ [[:digit:]]*  ]]` will watch if the key contain a digit which is probably not what we want.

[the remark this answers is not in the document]
dkjii-g: You're right, I meant like freebsd is doing [[ ${key} == [[:digit:]]* ]]

eclass/acct-user.eclass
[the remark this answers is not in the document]
dkjii-g: I personally added shadow support to egetent in my patches, I'm not sure reading /etc/shadow is safe in most platforms

dkjii-g:
> I cannot seem to add comments to parts you didn't change, but any function using usernames in these files will not work correctly using $EROOT. This includes fowners and chown, which will use the host's uid/gid instead of the $EROOT one
> 
> in user-info.eclass, egetgroups will not work with different ROOT
> 
> EDIT: you may want to see this bug report which has a similar patchset https://bugs.gentoo.org/541406, idk how that affects copyright though

You mightve missed this comment for esetgroups/egetgroups. egetgroups uses id which isn't root aware, I use egetent + regexes in my patches to fix this

I also added comments above for EROOT usage and [[ -z $ROOT ]] idiom, I'm not too familiar with github

eclass/acct-user.eclass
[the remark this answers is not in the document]
Mord3rca: It is done with this commit: a001c02fe4064984e47bc1a7df8991c8514a9b06 (I did not test that unfortunatly)

Mord3rca:
> > I cannot seem to add comments to parts you didn't change, but any function using usernames in these files will not work correctly using $EROOT. This includes fowners and chown, which will use the host's uid/gid instead of the $EROOT one
> > in user-info.eclass, egetgroups will not work with different ROOT
> > EDIT: you may want to see this bug report which has a similar patchset https://bugs.gentoo.org/541406, idk how that affects copyright though
> 
> You mightve missed this comment for esetgroups/egetgroups. egetgroups uses id which isn't root aware, I use egetent + regexes in the patches to fix this
> 
> I also added comments above for EROOT usage and [[ -z $ROOT ]] idiom, I'm not too familiar with github
> 
> @Mord3rca

Mord3rca:
Sorry wrong manipulation for the close/reopen :/

@dkjii-g Regarding your patch, I've tested that and actually, something is missing.
`id -G -n` put the first group ahead, your regex list groups according to their position in `/etc/group`

```sh
$ id -G -n morderca | wc -w
11
$ cat /etc/group | grep --color=never -E ":([^:]*,)?morderca(,[^:]*)?$" | cut -d':' -f1 | wc -w
10
# The primary group is missing + not ahead of the other (user group: morderca)
```

I'll see what I can do for that after my work time :)

dkjii-g:
> Sorry wrong manipulation for the close/reopen :/
> 
> @dkjii-g Regarding your patch, I've tested that and actually, something is missing.
> `id -G -n` put the first group ahead, your regex list groups according to their position in `/etc/group`
> 
> ```shell
> $ id -G -n morderca | wc -w
> 11
> $ cat /etc/group | grep --color=never -E ":([^:]*,)?morderca(,[^:]*)?$" | cut -d':' -f1 | wc -w
> 10
> # The primary group is missing + not ahead of the other (user group: morderca)
> ```
> 
> I'll see what I can do for that after my work time :)

You're right, didn't test this thouroughly, you can probably append egetgroupname (e.g. egetgroupname "$1" && grep)

Mord3rca:
@dkjii-g I've fixed (I think) the function `egetgroups`, however, esetgroup is still not good. I've an error and changing the command to use the GID instead of the group name doesn't fix it.

```sh
# usermod -g opendkim -R /tmp/aarch64-root opendkim
usermod: group 'opendkim' does not exist

# grep opendkim /tmp/aarch64-root/etc/passwd       
opendkim:x:334:334:User for OpenDKIM:/dev/null:/sbin/nologin
```

However, the user installed correctly (with: `SYSROOT=/tmp/aarch64-root ROOT=/tmp/aarch64-root aarch64-linux-gnu-emerge -avt acct-user/opendkim`)

I don't know why the group is not found ...

dkjii-g:
> @dkjii-g I've fixed (I think) the function `egetgroups`, however, esetgroup is still not good. I've an error and changing the command to use the GID instead of the group name doesn't fix it.
> 
> ```shell
> # usermod -g opendkim -R /tmp/aarch64-root opendkim
> usermod: group 'opendkim' does not exist
> 
> # grep opendkim /tmp/aarch64-root/etc/passwd       
> opendkim:x:334:334:User for OpenDKIM:/dev/null:/sbin/nologin
> ```
> 
> However, the user installed correctly (with: `SYSROOT=/tmp/aarch64-root ROOT=/tmp/aarch64-root aarch64-linux-gnu-emerge -avt acct-user/opendkim`)
> 
> I don't know why the group is not found ...

Whats `grep opendkim /tmp/aarch64-root/etc/group`?

eclass/acct-user.eclass
dkjii-g: @Mord3rca

eclass/user-info.eclass
dkjii-g: @Mord3rca

eclass/user-info.eclass
```diff
@@ -145,7 +158,16 @@ egetgroups() {
 	[[ $# -eq 1 ]] || die "usage: egetgroups <user>"
 
 	local egroups_arr
-	read -r -a egroups_arr < <(id -G -n "$1")
+	if [[ -n "${ROOT}" ]]; then
+		local pgroup=$(grep "^$1:" ${ROOT}/etc/passwd | cut -d: -f1)
```
dkjii-g, line 162: you should use `pgroup="$(egetent passwd $1)"`

dkjii-g: also i don't think this works, the -f is wrong here, it should be -f4, which gets the gid, then use egetgroupname $gid

Mord3rca:
> > @dkjii-g I've fixed (I think) the function `egetgroups`, however, esetgroup is still not good. I've an error and changing the command to use the GID instead of the group name doesn't fix it.
> > ```shell
> > # usermod -g opendkim -R /tmp/aarch64-root opendkim
> > usermod: group 'opendkim' does not exist
> > 
> > # grep opendkim /tmp/aarch64-root/etc/passwd       
> > opendkim:x:334:334:User for OpenDKIM:/dev/null:/sbin/nologin
> > ```
> > 
> > 
> > However, the user installed correctly (with: `SYSROOT=/tmp/aarch64-root ROOT=/tmp/aarch64-root aarch64-linux-gnu-emerge -avt acct-user/opendkim`)
> > I don't know why the group is not found ...
> 
> Whats `grep opendkim /tmp/aarch64-root/etc/group`? (also you used -R instead of --prefix)

My mistake ... It was the --prefix :/
I've retested that in a clean root directory (still in AARCH64) and it's working great

eclass/user-info.eclass
Mord3rca: -f1 get the group name, -f4 will get the gid.
Why can't I use -f1 then ? Is it different for BSD systems ?

Mord3rca:
Cleaning up history

thesamesam:
ping, shall we get this done? just needs sending to gentoo-dev?

Mord3rca:
@thesamesam Kinda forgot about it ... Thanks for the reminder. PR rebased and patch sent :)

thesamesam:
Huge thanks!

thesamesam:
See https://bugs.gentoo.org/958022.

PR #19205, opened 2021-01-25
x11-misc/xfractint: remove unused patches

mm1ke:


janko-j1:
Right. Seems like these were not removed with 20.04_p14 in commit 4d9c6e448feeb4f6f0bbf5ad516f655e27407bcd.

Maybe there could be a section in p.g.o with "old package" that (proxy-)maintainers could/should remove?

PR #19213, opened 2021-01-25
net-fs/samba EAPI-7 bump, missing dep fix/cleanup

a17r:


net-fs/samba/samba-4.13.3-r1.ebuild
```diff
@@ -1,32 +1,47 @@
...
-EAPI=6
+EAPI=7
 
 PYTHON_COMPAT=( python3_{6..9} )
-PYTHON_REQ_USE='threads(+),xml(+)'
+PYTHON_REQ_USE="threads(+),xml(+)"
 inherit python-single-r1 waf-utils multilib-minimal linux-info systemd pam
 
-MY_PV="${PV/_rc/rc}"
-MY_P="${PN}-${MY_PV}"
```
Polynomial-C, line 11: Please keep these

net-fs/samba/samba-4.13.3-r1.ebuild
```diff
@@ -1,32 +1,47 @@
...
+DESCRIPTION="Samba Suite Version 4"
+HOMEPAGE="https://samba.org/"
 
+MY_P="${PN}-${PV/_rc/rc}"
 SRC_PATH="stable"
-[[ ${PV} = *_rc* ]] && SRC_PATH="rc"
-
+if [[ ${PV} = *_rc* ]]; then
+	SRC_PATH="rc"
+	KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~ppc ~ppc64 ~sparc ~x86"
```
Polynomial-C, line 17: You just inverted the logig here. If `${PV}` contains `_rc` we don't want to have any `KEYWORDS`.

Polynomial-C (changes requested):

net-fs/samba/samba-4.13.3-r1.ebuild
a17r: ah, right.

net-fs/samba/samba-4.13.3-r1.ebuild
a17r: even if MY_PV is unused?

sys-libs/tevent/tevent-0.10.2.ebuild
```diff
@@ -1,43 +1,48 @@
...
 	elibc_glibc? (
 		net-libs/libtirpc[${MULTILIB_USEDEP}]
 		|| (
 			net-libs/rpcsvc-proto
 			<sys-libs/glibc-2.26[rpc(+)]
 		)
 	)
-	${PYTHON_DEPS}
 "
-# build system does not work with python3
```
a17r, line 34: that comment is probably also outdated ;)

Polynomial-C: Yeah, drop it please.

Polynomial-C (approved):

PR #19232, opened 2021-01-26
net-im/prosody: Added selinux USE flag and dependency.

jpds:
Simple dependency addition.

### DIFF
--- a/net-im/prosody/prosody-0.11.7-r101.ebuild
+++ b/net-im/prosody/prosody-0.11.7-r101.ebuild
@@ -15,14 +15,14 @@ SRC_URI="https://prosody.im/downloads/source/${P}.tar.gz"
 LICENSE="MIT"
 SLOT="0"
 KEYWORDS="amd64 arm arm64 x86"
-IUSE="icu +idn +libevent libressl mysql postgres +sqlite +ssl test +zlib"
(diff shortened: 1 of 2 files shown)


net-im/prosody/prosody-0.11.7-r101.ebuild
```diff
@@ -35,6 +35,7 @@ RDEPEND="
 	lua_single_target_lua5-1? ( $(lua_gen_cond_dep 'dev-lua/lua-bit32[lua_targets_lua5-1(-)]') )
 	mysql? ( $(lua_gen_cond_dep 'dev-lua/luadbi[mysql,${LUA_USEDEP}]') )
 	postgres? ( $(lua_gen_cond_dep 'dev-lua/luadbi[postgres,${LUA_USEDEP}]') )
+	selinux? ( sec-policy/selinux-jabber )
```
ConiKost, line 38: Is there a reason, why you are adding that package also into `DEPEND`? I don't think, this one will be used for compilation, as this will be some sort of runtime only rules for selinux?

Since `sec-policy/selinux-jabber` is not available on `arm`, you must mask `selinux` on arm profile.

ConiKost (changes requested):

ConiKost (approved):

PR #19235, opened 2021-01-27
dev-util/bitcoin-tx: remove unused patch(es)

mm1ke:


luke-jr:
Annoyingly this wasn't fixed in 0.19.2, but even if we do bother with that bump, it will need a new patch file anyway due to conflicts.

So this looks fine to me.

ConiKost (approved):

PR #19242, opened 2021-01-27
x11-terms/mlterm: allow wayland without X

vaukai:

### DIFF
--- a/x11-terms/mlterm/mlterm-3.9.1.ebuild
+++ b/x11-terms/mlterm/mlterm-3.9.1.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI="7"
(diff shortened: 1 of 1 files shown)


x11-terms/mlterm/mlterm-3.9.1.ebuild
```diff
@@ -12,12 +12,11 @@ SRC_URI="mirror://sourceforge/${PN}/${P}.tar.gz"
 LICENSE="BSD"
 SLOT="0"
 KEYWORDS="~amd64 ~ppc ~ppc64 ~x86"
-IUSE="bidi brltty cairo canna debug fbcon fcitx freewnn gtk gtk2 harfbuzz ibus libssh2 m17n-lib nls regis scim skk static-libs uim utempter wayland xft"
+IUSE="X bidi brltty cairo canna debug fbcon fcitx freewnn gtk gtk2 harfbuzz ibus libssh2 m17n-lib nls regis scim skk static-libs uim utempter wayland xft"
 REQUIRED_USE="gtk2? ( gtk )"
```
ionenwks, line 16: Need `|| ( X fbcon wayland )` and a default (`+X` would make sense to preserve old behavior), if all are unset it tries to use automagic.

ionenwks (commented):
Being conditional will have some implications, for now I see:
- USE=gtk's mlimgloader has automagic on libX11, can build without but will use if found.
- USE=scim fails to build if Xlib.h doesn't exist

ionenwks:
>     * USE=gtk's mlimgloader has automagic on libX11, can build without but will use if found.
Solution to this one seems to be to pass --without-x when USE=-X

x11-terms/mlterm/mlterm-3.9.1.ebuild
vaukai: Done

vaukai:
> > ```
> > * USE=gtk's mlimgloader has automagic on libX11, can build without but will use if found.
> > ```
> 
> Solution to this one seems to be to pass --without-x when USE=-X

Oops, how could I have missed this? 
Done.

vaukai:
Ping

PR #19243, opened 2021-01-28
app-text/csvfix: update ebuild

winterheart:
Updated HOMEPAGE and SRC_URI, migrated to EAPI=7, remove from package.mask.


app-text/csvfix/csvfix-1.6-r1.ebuild
```diff
@@ -0,0 +1,45 @@
...
+	edos2unix $(find csvfix/tests -type f)
+}
+
+src_compile() {
+	emake CC="$(tc-getCXX)" AR="$(tc-getAR)" lin
+}
+
+src_test() {
+	cd ${PN}/tests
+	chmod +x run1 runtests
```
telans, line 38: `die`, and maybe `fperms`

winterheart: Yes, good catch, thanks. Fixed.

app-text/csvfix/csvfix-1.6-r1.ebuild
```diff
@@ -0,0 +1,45 @@
...
+	default
+	edos2unix $(find csvfix/tests -type f)
+}
+
+src_compile() {
+	emake CC="$(tc-getCXX)" AR="$(tc-getAR)" lin
+}
+
+src_test() {
+	cd ${PN}/tests || die
```
ConiKost, line 37: You could that combine with `./runtests``` and ```&&```, so you only need one die.

app-text/csvfix/csvfix-1.6-r1.ebuild
```diff
@@ -0,0 +1,45 @@
...
+	chmod +x run1 runtests || die
```
ConiKost, line 38: I would do that in ```src_prepare```.

app-text/csvfix/csvfix-1.6-r1.ebuild
```diff
@@ -0,0 +1,45 @@
...
+SRC_URI="https://github.com/jheusser/csvfix/archive/version-${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="dev-libs/expat"
+DEPEND="${RDEPEND}"
+
+S="${WORKDIR}/${PN}-version-${PV}"
```
ConiKost, line 19: Just a small nit pick. If you have a look into ```skel.ebuild```, S is after SRC_URI :-)

app-text/csvfix/metadata.xml
```diff
@@ -3,6 +3,6 @@
 <pkgmetadata>
   <!-- maintainer-needed -->
```
ConiKost, line 4: Don't you want to maintain that?

ConiKost (changes requested):

ConiKost:
Ping. Could you also please rebase?

app-text/csvfix/csvfix-1.6-r1.ebuild
winterheart: All ebuilds that I saw, places S after *DEPEND section.

app-text/csvfix/csvfix-1.6-r1.ebuild
winterheart: done.

app-text/csvfix/csvfix-1.6-r1.ebuild
winterheart: done

app-text/csvfix/metadata.xml
winterheart: Not interested.
My original intention was tree-clean package, but after some feedback some devs pointed that main issues of package can be fixed. So why there this PR.
There still two options - delete package or let it remain in tree. Since no one even care about keeping it in tree, I prefer to delete.

winterheart:
@ConiKost rebased and updated.

PR #19246, opened 2021-01-28
dev-db/sqlitebrowser: version bump and adoption

beatussum:
- update the `metadata.xml` file
- update dependencies: remove `dev-cpp/antlr-cpp:2` (see https://github.com/sqlitebrowser/sqlitebrowser/commit/ce66c1da20d15e8211a14db13f63cd42c2ede770), update minimum required version of `x11-libs/qscintilla`
- add the live version

Closes: https://bugs.gentoo.org/735716

### Addendum

I'm a bit confused by the minimum required version of one dependency set in the previous ebuild: I can't find any version restriction on `dev-libs/qcustomplot`.


dev-db/sqlitebrowser/metadata.xml
```diff
@@ -1,15 +1,37 @@
...
+		compatible avec SQLite. Celui-ci est à déstination des utilisateurs et
+		des développeurs qui veulent créer des base de données et interragir
+		avec ces dernières, tout en utilisant une interface semblable à celle
+		d'un tableur, et, de ce fait, ne pas être contraint à apprendre des
+		commandes SQL complexes.
+	</longdescription>
+	<upstream>
+		<maintainer status="active">
+			<name>DB Browser for SQLite (DB4S) project</name>
+		</maintainer>
```
mgorny, line 33: I feel like this doesn't add any value. It's like saying "sqlitebrowser's maintainer is: sqlitebrowser's maintainers".

dev-db/sqlitebrowser/sqlitebrowser-3.11.1-r1.ebuild
```diff
@@ -1,67 +0,0 @@
-# Copyright 1999-2020 Gentoo Authors
```
mgorny, line 1: Our preferred policy is to make one commit adding the new version, and then another removing old. Possibly include some delay between the two in case it turned out the new version is buggy and people'd like to downgrade.

mgorny (changes requested):

beatussum:
@mgorny requested changes resolved in the last commit.

juippis (approved):

PR #19248, opened 2021-01-28
net-dns/knot: minor bump

nemunaire:


thesamesam:
Can you look at:
```
 *
 *   /var/lib/knot
 *
 ```

PR #19279, opened 2021-01-31
Kodi ebuild fixes and improvments

Karlson2k:
While waiting for Kodi Matrix release (expected at February, 14) I decided to improve ebuild.
This PR integrate several fixes and improvements for Kodi 19.0_rc1. Each change in separate commit.

Overall:
* Honor cpu_flags_* USE expand flags instead of detecting CPU features **at compile-time** on a **build machine**.
* dav1d library is optional for those who don't need to play AV1 files
* Other minor improvements.


candrews:
Thank you very much!

Karlson2k:
@candrews, you are welcome.
Commit 11fc99b1d7aa84e6a73c6a7f84b40cad9a131563 was missed. Is it intentional?
Is I understand, global USE flags can be described locally to clarify specific usage in the package.

Before I checked the Kodi code if was very unclear why Kodi may need libcap. I think it is much better with description.

candrews:
You're right - it should be there. I've now added it: 26fda8d37acae8a75b935bf3d5f63335e2694af3

PR #19282, opened 2021-02-01
Bitcoin 0.21.0

luke-jr:


luke-jr:

dev-libs/libsecp256k1/libsecp256k1-0.1_pre20201028.ebuild
```diff
@@ -0,0 +1,76 @@
+# Copyright 1999-2020 Gentoo Authors
```
juippis, line 1: Bump the copyright year.

dev-libs/libsecp256k1/libsecp256k1-0.1_pre20201028.ebuild
```diff
@@ -0,0 +1,76 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools eutils
```
juippis, line 6: *eutils.eclass* is deprecated and shouldn't have any use with ebuilds anymore.

dev-libs/libsecp256k1/libsecp256k1-0.1_pre20201028.ebuild
```diff
@@ -0,0 +1,76 @@
...
+
+MyPN=secp256k1
+DESCRIPTION="Optimized C library for EC operations on curve secp256k1"
+HOMEPAGE="https://github.com/bitcoin-core/secp256k1"
+COMMITHASH="3967d96bf184519eb98b766af665b4d4b072563e"
+SRC_URI="${HOMEPAGE}/archive/${COMMITHASH}.tar.gz -> ${PN}-v${PV}.tgz"
```
juippis, line 12: Plaintext the `${HOMEPAGE}` as also instructed by the CI check.

dev-libs/libsecp256k1/libsecp256k1-0.1_pre20201028.ebuild
```diff
@@ -0,0 +1,76 @@
...
+		$(use_enable schnorr module-schnorrsig) \
+		$(usex lowmem '--with-ecmult-window=2 --with-ecmult-gen-precision=2' '') \
+		$(usex precompute-ecmult '--with-ecmult-window=24 --with-ecmult-gen-precision=8' '') \
+		$(use_with valgrind) \
+		--disable-static
+}
+
+src_install() {
+	dodoc README.md
+	emake DESTDIR="${D}" install
```
juippis, line 74: You can just call `default`

net-p2p/bitcoin-qt/bitcoin-qt-0.21.0.ebuild
```diff
@@ -0,0 +1,187 @@
...
+		--with-system-libsecp256k1
+		--with-system-univalue
+	)
+	econf "${my_econf[@]}"
+}
+
+src_install() {
+	default
+
+	rm -f "${ED}/usr/bin/test_bitcoin" || die
```
juippis, line 147: We generally don't use `rm -f` because they don't error out if the file is missing, so these lines will be left to ebuilds for eternity.

net-p2p/bitcoind/bitcoind-0.21.0.ebuild
```diff
@@ -0,0 +1,171 @@
...
+		--with-system-libsecp256k1
+		--with-system-univalue
+	)
+	econf "${my_econf[@]}"
+}
+
+src_install() {
+	default
+
+	rm -f "${ED}/usr/bin/test_bitcoin" || die
```
juippis, line 132: Same here. Is it somehow predictable when this file will exist and be installed? Via test phase?

juippis (commented):
Please also fix whatever the CI found.

dev-libs/libsecp256k1/libsecp256k1-0.1_pre20201028.ebuild
```diff
@@ -0,0 +1,76 @@
...
+			asm_opt=arm
+		else
+			asm_opt=auto
+		fi
+	else
+		asm_opt=no
+	fi
+	econf \
+		--disable-benchmark \
+		--enable-experimental \
```
thesamesam, line 56: Shouldn't this be $(use experimental)?

luke-jr: It only gates the experimental modules, but it can be (`use_enable`, not `use`).

(This had been removed in the overlay due to the ECDH module being retroactively de-experimentalised in older versions.)

luke-jr:
Done

juippis (approved):

PR #19285, opened 2021-02-01
package.mask: Last rite app-misc/slmon

jsmolic:

### DIFF
--- a/profiles/package.mask
+++ b/profiles/package.mask
@@ -32,6 +32,18 @@
 
 #--- END OF EXAMPLES ---
 
+# Jakov Smolic <jakov.smolic@sartura.hr> (2021-02-12)
+# Fails to build with gcc-10, stuck on EAPI 5, dead upstream
+# multiple reported bugs, not touched by maintainer in years.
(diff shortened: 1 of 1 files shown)


profiles/package.mask
```diff
@@ -32,6 +32,12 @@
 
 #--- END OF EXAMPLES ---
 
+# Jakov Smolic <jakov.smolic@sartura.hr> (2021-02-01)
+# Fails to build with gcc-10, no maintainer, last release in 2004,
+# no reverse dependencies.
+# Removal on 2021-03-03.  Bug #707504.
```
SoapGentoo, line 38: prefer to always write `Removal in 30 days`

SoapGentoo (changes requested):

profiles/package.mask
jsmolic: Sure, I'm using mgorny's script, it switched to `Removal on date` recently, but I'll update it

jsmolic: Done

SoapGentoo (approved):

PR #19297, opened 2021-02-01
dev-embedded/jlink: Add JLink 6.94a

thughes:
This ebuild is based on 6.46h from:
https://cgit.gentoo.org/user/palmer.git/plain/dev-embedded/jlink/jlink-6.46h.ebuild?id=3e93cda2f058ee2a1ca26e0435a29505836d26d1
which was created by Palmer Dabbelt <palmer@dabbelt.com>.

Bug: https://bugs.gentoo.org/703160


thesamesam:
I think ideally we'd get permission from them. Could you ask them to post the GCO sign off here?

thughes:
> I think ideally we'd get permission from the original author. Could you ask them to post the GCO sign off here?

@palmer-dabbelt

PR #19301, opened 2021-02-02
x11-wm/i3-gaps: bump to 4.19.1

ajakk:

Only trivial differences from 4.19.


juippis (approved):

x11-wm/i3-gaps/i3-gaps-4.19.1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+		dev-perl/IPC-Run
+		dev-perl/Inline
+		dev-perl/Inline-C
+		dev-perl/X11-XCB
+		dev-perl/XS-Object-Magic
+		x11-apps/xhost
+		x11-base/xorg-server[xephyr,xvfb]
+		x11-misc/xvfb-run
+	)"
+BDEPEND="app-text/asciidoc
```
juippis, line 44: Oh yea I wanted to ask if asciidoc is required everytime, or just when `USE="doc"` is wanted?

ajakk: Sorry it took so long to respond! `i3-gaps` doesn't have USE=doc, but `i3` itself does. I don't see much of a need for it in any case, given `qsize` reports `x11-wm/i3: 64 files, 15 non-files, 2243 KiB` with USE=-doc and `x11-wm/i3: 66 files, 16 non-files, 2272 KiB` with USE=doc. A difference of a couple dozen kilobytes doesn't necessitate adding the flag, I think.

PR #19304, opened 2021-02-02
app-admin/sysklogd: Make logrotate USE flag pull in required dep.

jpds:


jpds:
@Polynomial-C Can you please review this when you have a moment?

PR #19311, opened 2021-02-02
media-libs/lsp-plugins: Add missing libglvnd[X] USE dependendency

simonvanderveldt:
Closes: https://bugs.gentoo.org/737536

### DIFF
--- a/media-libs/lsp-plugins/lsp-plugins-1.1.24.ebuild
+++ b/media-libs/lsp-plugins/lsp-plugins-1.1.24.ebuild
@@ -1,7 +1,9 @@
-# Copyright 2019-2020 Gentoo Authors
+# Copyright 2019-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
+
+inherit xdg
 
 DESCRIPTION="Linux Studio Plugins"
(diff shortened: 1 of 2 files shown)


fordfrog:
@simonvanderveldt thank you for the fix. though as this is a compile-time fix, you should do the fix without revbump. could you please update the pr? or i can fix it directly, your choice. it also needs to inherit xdm as the icon cache is not updated correctly. that's the issue of the original ebuild and not of your patch, but this should be fixed aswell.

fordfrog:
also, looking at the upstream site, 1.1.29 is already out.

simonvanderveldt:
> @simonvanderveldt thank you for the fix. though as this is a compile-time fix, you should do the fix without revbump. could you please update the pr? or i can fix it directly, your choice. it also needs to inherit xdm as the icon cache is not updated correctly. that's the issue of the original ebuild and not of your patch, but this should be fixed aswell.

I can do this, no problem. Will do so tonight.

> also, looking at the upstream site, 1.1.29 is already out.

I'll have a look at doing a revbump in a follow-up PR

simonvanderveldt:
@fordfrog Should be fixed now :)

fordfrog:
@simonvanderveldt cool, thanks! :-)

PR #19312, opened 2021-02-03
media-sound/qjackctl: Switch to cmake, add missing DEPEND and RDEPEND

a17r:
Bug: https://bugs.gentoo.org/766300


fordfrog:
thank you Andreas!

PR #19314, opened 2021-02-03
sys-devel/bmake: Disable failing test

jsmolic:


Whissi:
Maybe include a comment **why** the tests were disabled. If we are disabling tests just because they fail, we can just go RESTRICT=test ;)

jsmolic:
> Maybe include a comment **why** the tests were disabled. If we are disabling tests just because they fail, we can just go RESTRICT=test ;)

I think we already have comment in the ebuild about the tests. bmake has sensitive test suite and it changes from version to version, couple of tests were already disabled, and I can't reproduce this failure on `amd64` so I just figured it's easier to disable one more test.

PR #19315, opened 2021-02-03
media-video/libva-utils-2.8.0: fixed testing in portage

Karlson2k:
Closes: https://bugs.gentoo.org/768486


Karlson2k:
@aballier @thesamesam Could you check it? Thanks in advance.

Karlson2k:
@thesamesam Could you check it?

PR #19319, opened 2021-02-04
media-tv/kodi: fix build with libcdio[-cxx]

pks-t:
When building Kodi with ISO9660PP support, then it requires
libiso9660++.so to be installed. This library is provided by
dev-libs/libcdio, but only when building with C++ libraries. We thus
need to depend on libcdio[cxx] to not fail the build.

Signed-off-by: Patrick Steinhardt <ps@pks.im>


candrews:
Thank you!

PR #19320, opened 2021-02-04
sys-devec/bc: fix building with locked account

pks-t:
The build of bc fails when building it with a user whose shell is set to
`/sbin/nologin`. The root cause is that the "fix-libmath_h" script does
not have a shebang, so it'll instead just use the user's configured
shell.

Fix this issue by patching the script to have a shebang.

Signed-off-by: Patrick Steinhardt <ps@pks.im>


floppym:
I'm unable to reproduce the build failure, but I guess the patch seems reasonable.

PR #19322, opened 2021-02-04
eclass/xdg{,-utils}: fix RDEPENDS

epsilon-0:

---

The dependencies should be RDEPENDs and are needed in the proper eclass `xdg-utils`, which `xdg` inherits.

### DIFF
--- a/eclass/xdg-utils.eclass
+++ b/eclass/xdg-utils.eclass
@@ -22,6 +22,15 @@ case "${EAPI:-0}" in
 	*) die "EAPI=${EAPI} is not supported" ;;
 esac
 
+# Avoid dependency loop as both depend on glib-2
+if [[ ${CATEGORY}/${P} != dev-libs/glib-2.* ]] ; then
+RDEPEND="
+	dev-util/desktop-file-utils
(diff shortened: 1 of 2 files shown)


ionenwks:
xdg-utils is setup to do nothing if the commands are not found rather than die, which is mostly fine, should get updated by related dependencies when available.

And then some packages use this eclass in an optional way and may not want extra depends.

Besides, odds are DEPEND/RDEPEND is overwritten by most consumers, it'd make more sense to remove entirely if anything (but unsure if some particular ebuilds need this, so not touching this may be better).

ionenwks:
> And then some packages use this eclass in an optional way and may not want extra depends.

This is actually something that been bothering me in a package where I use USE=gui with xdg.eclass, this gives me two depends that do nothing when -gui (no icons/desktop installed), while the commands can gracefully do nothing if either nothing to update or commands are missing.

Either way, I'd like the situation to not get worse so I oppose adding this RDEPEND to xdg-utils

Arfrever:
I also oppose this addition to `xdg-utils.eclass`.
I selectively use only what is needed from this eclass and I do not want unnecessary dependencies.

PR #19335, opened 2021-02-04
dev-db/redis: bump to 6.0.10

hydrapolic:
- Config moved to /etc/redis directory
- TLS support from 6.0.9-r1
- lua-single from 6.0.9-r100


hydrapolic:
Thanks @Whissi

PR #19340, opened 2021-02-05
dev-cpp/prometheus-cpp: Version bump to 0.12.1

vilhelmgray:
Closes: https://bugs.gentoo.org/768771


dev-cpp/prometheus-cpp/prometheus-cpp-0.12.1.ebuild
```diff
@@ -0,0 +1,38 @@
+# Copyright 1999-2020 Gentoo Authors
```
juippis, line 1: Please remember this or use repoman to commit in future :)

juippis (approved):

PR #19345, opened 2021-02-05
media-libs/lsp-plugins: Bump to 1.1.29

simonvanderveldt:

The tag schema was changed to just contain the version, so that's different compared to the previous ebuilds. I've also updated the 9999 ebuild with this. See https://github.com/sadko4u/lsp-plugins/issues/139 for more details.

I'm not sure if any and if which old versions should be dropped, so didn't drop any in this PR.


simonvanderveldt:
@fordfrog As discussed in #19311 here's the follow-up PR with the revbump :)

fordfrog:
@simonvanderveldt thanks! i also did the cleanup and will do a stabilization request for 1.2.24.

fordfrog:
well, it still has some open issues which i do not have time to resolve now...

PR #19350, opened 2021-02-06
app-misc/radeontop: add dependency on amd radeon gpus, remove old

mm1ke:
Hi,

In order to fix bug #768549 I've added dependencies on amd radeon gpu's. Please review.


ConiKost (changes requested):
Lgtm, but please add with one commit new `-r1` version and drop with a seperate commit old non `-r1` one.

mm1ke:
@ConiKost updated the commits :)

ConiKost:
FYI: You dropped `-9999`, so I readded it.

PR #19354, opened 2021-02-06
media-plugins/hexter: drop gtk2 support

epsilon-0:
Closes: https://bugs.gentoo.org/769035


fordfrog:
@epsilon-0 is there any gui replacement? or there's no need for gui and the text interface is sufficient?

epsilon-0:
@fordfrog unsure, I don't use hexter.
I had looked around a bit upstream in the options and it *seems* like textui should be enough.

fordfrog:
@epsilon-0 i'd give it some more time if you don't mind. still not sure whether to drop it already to the textui only interface.

epsilon-0:
yes, no problem, theres no hurry.

fordfrog:
@epsilon-0 thanks!

PR #19358, opened 2021-02-07
media-sound/mpd: Add sndio support

lanodan:
Depends on https://bugs.gentoo.org/show_bug.cgi?id=769122 to avoid keyword regressions.


lanodan:
Depends on https://bugs.gentoo.org/show_bug.cgi?id=769122 for alpha and arm keywording.

lanodan:
@strikersh Fixed this PR so it's mergeable, I think it's a better one than ours.

Zlogene:
I am sorry, I kinda forgot about this PR at all :/

PR #19359, opened 2021-02-07
net-misc/mstpd: Multiple spanning tree protocol daemon

jsmolic:
* mstpd is an user-space daemon that implements RSTP and MSTP protocols


net-misc/mstpd/mstpd-0.0.8.ebuild
```diff
@@ -0,0 +1,36 @@
...
+SRC_URI="https://github.com/mstpd/mstpd/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+src_prepare() {
+	default
+
+	sed -i -e 's:-Werror::g' Makefile.am || die
```
SoapGentoo, line 19: can you tell upstream to make this an option? saves us the (slow) autoreconf

net-misc/mstpd/mstpd-0.0.8.ebuild
```diff
@@ -0,0 +1,36 @@
...
+		--localstatedir="${EPREFIX}"/var
+}
+
+src_install() {
+	default
+
+	# install systemd service
+	systemd_dounit "${FILESDIR}"/${PN}.service
+
+	docompress -x /usr/share/man/man{5,8}
```
SoapGentoo, line 35: is the compression a problem?

jsmolic: It's just a QA warning

net-misc/mstpd/mstpd-0.0.8.ebuild
jsmolic: Sure, I'll forward it

jsmolic: Forwarded in https://github.com/mstpd/mstpd/issues/105

PR #19365, opened 2021-02-07
Kodi fixes and improvements

Karlson2k:
Fixed Kodi dependencies, required combinations of use flags and other minor improvements.

Looks like many dependencies were just taken from Kodi documentation, while some of them are used only to build internal version of some system libraries. As Kodi on Gentoo uses system libraries (except FFmpeg), there are no need for many dependencies.


Karlson2k:
Looks like BDEPEND/RDEPEND need some more work. The rest is fine.

Karlson2k:
Added `metadata.xml` information from previous PRs (seems accidentally missing), added some simplification.

@candrews Could you have a look?

Karlson2k:
Added 'eventclients' USE flag.

I think this is a final commit as currently all Kodi parameters are handled. Not more commits are expected before release of v19.

Karlson2k:
@gentoo-repo-qa-bot Second time reports already fixed error and referred non-existing strings in ebuild. 
What's going on?

Karlson2k:
Thanks, @candrews!
The Kodi release is postponed, but probably will be made in the next few days. Just need to make a final polishing.

candrews:
I'm super psyched for the release!

And thank you for your work on the ebuilds in Gentoo :clap:

PR #19366, opened 2021-02-07
profiles: amd64: x32: Add dev-qt/qtwebengine to package.mask

a17r:


mgorny:
I suppose this makes somewhat sense *but* if we were to take x32 really seriously, we'd have to make these chromiumish packages multilib ;-/.

a17r:
> I suppose this makes somewhat sense _but_ if we were to take x32 really seriously, we'd have to make these chromiumish packages multilib ;-/.

too big to fail (TM)...

mgorny:
I somewhat feel like you're going to mask everything™ ;-).

PR #19374, opened 2021-02-08
dev-util/goland: Fix desktop entry name typo

Neirpyc:
Previously, the desktop entry being create was entitled "gogland", which is a typo (Should be "goland"), and is fixed with this PR.


zx2c4:
Applied, thanks.

PR #19376, opened 2021-02-08
net-analyzer/nrpe: rev bump to add USE=tcpd

jkroonza:
This includes a patch for ./configure that has also been submitted
upstream:

https://github.com/NagiosEnterprises/nrpe/pull/247

Closes:  https://bugs.gentoo.org/698794


juippis:
Ideally we'd like to see what upstream says about it first. If there's no action in ~30 days, please ping in this PR and we'll process it.

jkroonza:
> Ideally we'd like to see what upstream says about it first. If there's no action in ~30 days, please ping in this PR and we'll process it.

Agreed, no rush.

net-analyzer/nrpe/nrpe-4.0.2-r1.ebuild
```diff
@@ -0,0 +1,105 @@
...
+	tcpd? (
+		sys-apps/tcp-wrappers
+	)
+	ssl? (
+		!libressl? ( dev-libs/openssl:0= )
+		libressl? ( dev-libs/libressl:0= )
+	)"
+RDEPEND="${DEPEND}
+	|| ( net-analyzer/nagios-plugins net-analyzer/monitoring-plugins )
+	selinux? ( sec-policy/selinux-nagios )"
```
juippis, line 27: I'm not really happy seeing this USE flag having no other function in this ebuild. Could you move it to `optfeature`?

net-analyzer/nrpe/nrpe-4.0.2-r1.ebuild
```diff
@@ -0,0 +1,105 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+inherit autotools systemd
```
juippis, line 5: We like to write one empty line below EAPI but it's not as important. Just do it if you edit this ebuild due to selinux comment.

net-analyzer/nrpe/nrpe-4.0.2-r1.ebuild
```diff
@@ -0,0 +1,105 @@
...
+	eautoreconf
+}
+
+src_configure() {
+	# The configure script tries to detect what OS, distribution, and
+	# init system you're running and changes the build/install process
+	# depending on what it comes up with. We specify fixed values
+	# because we don't want it guessing, for example, whether or not
+	# to install the tmpfiles.d entry based on whether or not systemd
+	# is currently running (OpenRC uses them too).
```
juippis, line 45: I forgot to check, but this does install tmpfiles.d correctly right? There's also an eclass to handle those files.

net-analyzer/nrpe/nrpe-4.0.2-r1.ebuild
```diff
@@ -0,0 +1,105 @@
...
+	newins "${FILESDIR}/nrpe.xinetd.2" nrpe
+
+	rm "${ED}/usr/bin/nrpe-uninstall" || die 'failed to remove uninstall tool'
+	rm -r "${ED}/run" || die 'failed to remove /run'
+}
+
+pkg_postinst() {
+	elog 'Some users have reported incompatibilities between nrpe-2.x and'
+	elog 'nrpe-3.x. We recommend that you use the same major version for'
+	elog 'both your server and clients.'
```
juippis, line 96: `There are known incompabilities`

I'm not really aware, but can these version restrictions be somehow made in ebuilds? Or are there tons of servers/clients capable of using nrpe, outside ::gentoo repo?

net-analyzer/nrpe/nrpe-4.0.2-r1.ebuild
jkroonza: Did a quick grep for selinux on the tree, not seeing optfeature used for this at all, all selinux USE flag uses the structure here.  But I think I get why you're saying this ... if the USE flag change, we get a pointless recompile.  Fortunately selinux is one of those that should ALMOST NEVER change.

net-analyzer/nrpe/nrpe-4.0.3.ebuild
```diff
@@ -0,0 +1,106 @@
...
+
+	insinto /etc/xinetd.d/
+	newins "${FILESDIR}/nrpe.xinetd.2" nrpe
+
+	rm "${ED}/usr/bin/nrpe-uninstall" || die 'failed to remove uninstall tool'
+	rm -r "${ED}/run" || die 'failed to remove /run'
+}
+
+pkg_postinst() {
+	elog 'Some users have reported incompabilities between nrpe-2.x and'
```
ajakk, line 95: Is this message still necessary now that we only have 4.x in the tree for almost a year now?

net-analyzer/nrpe/nrpe-4.0.3.ebuild
```diff
@@ -0,0 +1,106 @@
...
+	elog 'nrpe-3.x. We recommend that you use the same major version for'
+	elog 'both your server and clients.'
+
+	if use command-args ; then
+		ewarn ''
+		ewarn 'You have enabled command-args for NRPE. That lets clients'
+		ewarn 'supply arguments to the commands that are run, and IS A'
+		ewarn 'SECURITY RISK!'
+		ewarn''
```
ajakk, line 104: Missing space in between ewarn and ''

net-analyzer/nrpe/nrpe-4.0.3.ebuild
jkroonza: killed.  agree.

net-analyzer/nrpe/nrpe-4.0.2-r1.ebuild
juippis: You are correct, I was kind of 50-50 about that comment since selinux is a bit special, and straight out controlled by profiles. You can safely presume it's globally either enabled or disabled.

PR #19378, opened 2021-02-08
x11-misc/py3status: add minimal USE flag

ferki:
This PR is an attempt to add a new `minimal` USE flag to the x11-misc/py3status ebuild that prevents pulling in x11-misc/i3status as a dependency.

### Context

I've recently ran into a use case where I ended up migrating my config to use only py3status modules, and at that point, I realized I don't necessariliy need i3status itself anymore. So I've patched the ebuild in my overlay, and by now I'm using py3status for about 2 months without i3status without any issues. In hopes of this being useful for more than just myself, I've decided to propose this change for the official tree as well.

### DIFF
--- a/x11-misc/py3status/metadata.xml
+++ b/x11-misc/py3status/metadata.xml
@@ -11,6 +11,7 @@
 	</upstream>
 	<use>
 		<flag name="gevent">Install extra requirement <pkg>dev-python/gevent</pkg> to enable green threads.</flag>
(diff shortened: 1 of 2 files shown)


ultrabug:
Thank you @ferki I never expected to get a PR on this ebuild :+1:

ferki:
Thanks @ultrabug, I'm glad you found it useful!

PR #19380, opened 2021-02-08
media-sound/qjackctl: fix cmake flags

waebbl:
Remove passing of CONFIG_JACK_VERSION to enable switching
between virtual/jack slots.

Closes: https://bugs.gentoo.org/769563

### DIFF
--- a/media-sound/qjackctl/qjackctl-0.9.1.ebuild
+++ b/media-sound/qjackctl/qjackctl-0.9.1.ebuild
@@ -41,7 +41,6 @@ src_configure() {
 		-DCONFIG_DBUS=$(usex dbus 1 0)
 		-DCONFIG_DEBUG=$(usex debug 1 0)
 		-DCONFIG_PORTAUDIO=$(usex portaudio 1 0)
-		-DCONFIG_JACK_VERSION=1
 	)
 	cmake_src_configure
 }


ionenwks:
Not familiar with this package but is it necessary to pass this at all? It's an unlikely event but if build with jack2 and go back to jack-audio-connection-kit:
```
$ qjackctl --version
qjackctl: symbol lookup error: qjackctl: undefined symbol: jack_get_version_string
```
While if unset can switch at will and doesn't get automagic enabled (but --version won't show jack's version).

fordfrog:
> Not familiar with this package but is it necessary to pass this at all? It's an unlikely event but if build with jack2 and go back to jack-audio-connection-kit:
> 
> ```
> $ qjackctl --version
> qjackctl: symbol lookup error: qjackctl: undefined symbol: jack_get_version_string
> ```
> 
> While if unset can switch at will and doesn't get automagic enabled (but --version won't show jack's version).

this sounds like a better approach to me.

waebbl:
I'm gonna look at this approach.
Don't know the package either, but after @ionenwks comment at b.g.o., I was thinking this looks like a quick fix.

fordfrog:
@waebbl imo both would work but i have no idea how to recompile `qjackctl` when the `jack` main version changes so i would prefer the solution that does not require recompiling `qjackctl` on the change.

waebbl:
The both jack virtual seem anyway not easily exchangeable. After switching back from jack2 to jack-audio-connection-kit, I get a bunch of preserved rebuilds. However it looks like, it's not any package installed with `jack` support, as, for example, qjackctl is missing on the list.

For my desktop machine these are: media-libs/portaudio, media-plugins/gst-plugins-jack, media-plugins/libvisual-plugins, media-libs/libsdl2, media-video/ffmpeg and media-plugins/alsa-plugins

waebbl:
> @waebbl imo both would work but i have no idea how to recompile `qjackctl` when the `jack` main version changes so i would prefer the solution that does not require recompiling `qjackctl` on the change.

@fordfrog Agreed, I would prefer this way too.

fordfrog:
thank you both! :-)

PR #19382, opened 2021-02-08
dev-libs/tvision: fix linker flags

waebbl:
Enables linking with LLD

Closes: https://bugs.gentoo.org/737202

### DIFF
--- a/dev-libs/tvision/files/tvision-2.2.1.4-Gentoo-specific-fix-linker-paths.patch
+++ b/dev-libs/tvision/files/tvision-2.2.1.4-Gentoo-specific-fix-linker-paths.patch
@@ -1,0 +1,33 @@
+From 62fce1e63e92ae71e2ce061c40ba736f4b22f71d Mon Sep 17 00:00:00 2001
+From: Bernd Waibel <waebbl-gentoo@posteo.net>
+Date: Mon, 8 Feb 2021 23:43:47 +0100
+Subject: [PATCH] [Gentoo-specific] fix linker paths
+
+Signed-off-by: Bernd Waibel <waebbl-gentoo@posteo.net>
+---
+ config.pl | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/config.pl b/config.pl
+index 4664baa..9e2cb96 100644
+--- a/config.pl
++++ b/config.pl
+@@ -179,13 +179,13 @@ if ($OS eq 'UNIX')
+    # QNX 6.2 beta 3 workaround
+    $MakeDefsRHIDE[3].='/lib ' if ($OSf eq 'QNXRtP');
+    # Link with installed libraries
+-   $MakeDefsRHIDE[3].=$realPrefix.'/lib';
++   #$MakeDefsRHIDE[3].=$realPrefix.'/lib';
+    $MakeDefsRHIDE[3].='/'.$conf{'libs-subdir'} if $conf{'libs-subdir'};
+    $MakeDefsRHIDE[3].=' ';
+    $MakeDefsRHIDE[3].='../../makes ' unless $conf{'libs-here'} eq 'no';
+    $MakeDefsRHIDE[3].=$here.'/makes ' unless $conf{'libs-here'} eq 'no';
+    $MakeDefsRHIDE[3].='../../intl/dummy ' if $UseDummyIntl;
+-   $MakeDefsRHIDE[3].=$conf{'X11LibPath'}.' ' if ($conf{'HAVE_X11'} eq 'yes');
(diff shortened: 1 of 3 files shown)


waebbl:
What's the reason for the gentoo-bot to remove the self-assigned tag?

thesamesam:
Have you added that email to your GitHub account?

waebbl:
Yes it's added there and set as primary email.

thesamesam:
Looks like we need to respect CXX:
```
g++  -I../include -I/usr/X11R6/include       -fPIC -O2 -pipe -march=native -fdiagnostics-color=always -frecord-gcc-switches  -c ../stream/schdirdi.cc -o obj/schdirdi.lo
```

thesamesam:
Thanks!

waebbl:
> Looks like we need to respect CXX:
> 
> ```
> g++  -I../include -I/usr/X11R6/include       -fPIC -O2 -pipe -march=native -fdiagnostics-color=always -frecord-gcc-switches  -c ../stream/schdirdi.cc -o obj/schdirdi.lo
> ```

@thesamesam I noticed it this morning too. There's also a bug https://bugs.gentoo.org/730022 where ar is called directly, and I noticed that also ranlib is called directly. Wanted to address these issues as a separate commit to this PR, but can open a new PR as well.

The problem here is, the code is hard-checked in config.pl and then environment files, a bunch of make include files and the makefile are generated from this. It checks for various systems, like qnx, windows, osx, etc. but not for different architectures. The package is becoming more and more unmaintainable.

The options that I see are
- add all the possible triplets for the build tools into the config.pl file, which looks like a maintenance nightmare to me (having no knowledge of coding in perl) with respect to supported arches.
- contribute to upstream and create a proper modern build system
- There's https://github.com/magiblot/tvision which started as a fork with source code level compatibility, but this is AFAIK no longer valid. But it should be mostly API compatible. We could use this fork as a code base, possibly in a new slot and let the other versions go away. This one has a new code base and a modern build system.
- move the package to ::guru or my own overlay ::waebbl and remove it from ::gentoo

The package itself has to my knowledge no revdeps yet, but the modern fork from the 3rd item above could be used as an alternative to ncurses for cli apps.

magiblot:
> * There's https://github.com/magiblot/tvision which started as a fork with source code level compatibility, but this is AFAIK no longer valid. But it should be mostly API compatible. We could use this fork as a code base, possibly in a new slot and let the other versions go away. This one has a new code base and a modern build system.

It is source-code compatible with applications written against Borland's original Turbo Vision 2.0 from the early 1990's. It does not implement the extensions of SET's port (http://tvision.sourceforge.net/), the one you are dealing with here. The original author (https://github.com/set-soft) may be able to lend you a hand.

waebbl:
> It does not implement the extensions of SET's port (http://tvision.sourceforge.net/), the one you are dealing with here. The author (https://github.com/set-soft) may be able to lend you a hand.

@magiblot Thanks for stepping into this topic. Do you know what are those extensions? I'm not so much into the code base.
The question is, how many users are having the package actually installed. If there's only a few using it, possibly with some old 90's code it might worth to try your fork instead, from which I got a much better impression and an easier time maintaining it.

Sure the author of SET's port might lend me a hand, but to me it looks like a lot of effort. Currently `g{cc,++}` are hard-coded as compilers for linux in config.pl, as well as other build tools, and adding several compiler-linker-platform-arch combinations seems like a lot of work.

I don't have any old tools left. I once had a program from the mid-90's from when I was working as a student at the university, but I can't find it anymore, so unfortunately I don't have a real test case, but I'd like to keep to project alive.

magiblot:
> Do you know what are those extensions? I'm not so much into the code base.

I'm not very familiar with the code base either, but I know it has at least the following features (at the API level):

* A "compatibility layer" that defines macros for cross-compiler compatibility. E.g. `CLY_cin` expands to `cin` or `std::cin` depending on the compiler.
* A system for codepage and font setting.
* Integration with the operating system clipboard. 

But anyway, the general idea is that both ports are not compatible at the source code level.

> The question is, how many users are having the package actually installed. If there's only a few using it, possibly with some old 90's code it might worth to try your fork instead, from which I got a much better impression and an easier time maintaining it.

Although it's not what you asked for, you can find some projects using this fork on GitHub: [setedit](https://github.com/set-soft/setedit), [tvterm](https://github.com/rkd77/tvterm), [elinks-tv](https://github.com/rkd77/elinks-tv). So there definitely are some projects out there which depend on SET's Turbo Vision. More, probably, than projects using my port.

I want to make it clear that it was never my intention to displace other ports of Turbo Vision. In this sense, it may be unfortunate that I kept the name `tvision`, although this doesn't conflict with SET's port, which uses `rhtvision` for the include directory and the shared library name. So they are not exclusive libraries, they only need different package names.

PR #19383, opened 2021-02-09
Drop obsolete SSL deps in dev-python/urllib3 and dev-python/requests

floppym:


mgorny:
Have you managed to test this somehow? trustme pulls in cryptography for me (and uses it heavily).

dev-python/urllib3/urllib3-1.26.3-r1.ebuild
```diff
@@ -21,10 +21,6 @@ RESTRICT="!test? ( test )"
 RDEPEND="
 	>=dev-python/PySocks-1.5.8[${PYTHON_USEDEP}]
 	<dev-python/PySocks-2.0[${PYTHON_USEDEP}]
-	dev-python/certifi[${PYTHON_USEDEP}]
-	>=dev-python/cryptography-1.3.4[${PYTHON_USEDEP}]
-	>=dev-python/pyopenssl-0.14[${PYTHON_USEDEP}]
-	>=dev-python/idna-2.0.0[${PYTHON_USEDEP}]
```
mgorny, line 27: Maybe you could remove `contrib/pyopenssl.py` as well, while at it.

floppym: I thought about that, but importing the module will fail either way so I'm not sure there is any point in removing it. Also, the comment at the top of the file might be useful to have as documentation.

I do anticipate some breakage if packages import urllib3.compat.pyopenssl regardless of python version. Hopefully some time in ~arch will expose any such packages.

floppym:
> Have you managed to test this somehow? trustme pulls in cryptography for me (and uses it heavily).

I ran into the same issue. I'll do some basic testing with requests before pushing.

dev-python/urllib3/urllib3-1.26.3-r1.ebuild
mgorny: My thought is that if we removed the file, we make a clear signal that something is wrong and people need to investigate why it was removed. If we leave it, they'll get a 'add dep on cryptography' signal. Which might or might not be what we want.

mgorny (approved):
Either way, this LGTM.

dev-python/urllib3/urllib3-1.26.3-r1.ebuild
floppym: If we leave it, pyopenssl essentially becomes an optional runtime dependency, and users can install pyopenssl manually if they run into packages that utilize urllib3.compat.pyopenssl unconditionally. I think that's better than making it impossible for the user to resolve themselves.

However, this probably does warrant a comment in the ebuild so that we don't have random developers re-adding these deps by accident.

mgorny: Well, I more meant that people would file bugs if they actually need it, so we'll know.

floppym:
Pushed as deacc5bbcf72a2357bb35bb077acf65b4e6df13c.

PR #19385, opened 2021-02-09
Update GNOME C++ bindings

dm0-:
@leio @mattst88 This updates most of the mm packages based on https://github.com/mattst88/gentoo/commits/mm.  (I skipped pangomm since it can't be updated due to [bug #698922](https://bugs.gentoo.org/698922).)  I propagated the `doc` USE flag through all of the dependencies for consistency since it was used somewhere in the original ebuilds, but it is optional and can be dropped to simplify things.

They all have new API versions, and they don't overlap with existing versions, so the ebuilds need new slots.  This also requires explicitly setting the slot in package dependencies, since they won't build with newer versions.


dm0-:
I added GLEP 81 packages to silence some warnings.  Here is the `api.git` patch: https://dpaste.com/83U6YDJBZ.txt

dm0-:
@SoapGentoo I've updated this to fix merge conflicts from recent changes.  Can at least the second half of this be merged so I don't have to maintain dozens of package changes?

I've also opened gentoo/api-gentoo-org#368 for the UID/GID reservation.  (I didn't realize it was on GitHub for PRs to that repo.)

PR #19392, opened 2021-02-09
mail-client/neomutt: Add autocrypt use flag

shuber2:
NeoMutt supports autocrypt since version 2019-10-25. It requires gpgme
of version 1.8.0 or higher and sqlite. It is recommended that idn or
idn2 is enabled according to the manual.

The minimum requirement of gpgme of neomutt-20201127.ebuild is increased
from 0.9.0 to 1.8.0. This change has no consquence as the minimum
available version in portage is 1.14.0.

Closes: https://bugs.gentoo.org/769698
Signed-off-by: Stefan Huber <shuber@sthu.org>

### DIFF
--- a/mail-client/neomutt/metadata.xml
+++ b/mail-client/neomutt/metadata.xml
@@ -6,6 +6,7 @@
 		<name>Nicolas Bock</name>
 	</maintainer>
 	<use>
+		<flag name="autocrypt">Enable autocrypt.org support</flag>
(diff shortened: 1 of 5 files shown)


j-licht:
Please describe the new use flag in the metadata.xml.

shuber2:
> Please describe the new use flag in the metadata.xml.

Done.

nicolasbock (approved):
Thanks!

PR #19393, opened 2021-02-09
net-p2p/mldonkey-3.1.7-r1 revision bump - improvements in init.d script

chusopr:
* Implements the ability to send shutdown and bandwidth-toggle commands via telnet interface in addition to HTTP in case the HTTP interface is disabled as requested in [#338017](https://bugs.gentoo.org/338017).
* Refactors some repeated code to make it more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).
* Fixes small issue when stopping the service that made it always print `start-stop-daemon: no matching processes found` because the `start-stop-daemon` test was missing a second `--quiet` to suppress the expected error: https://github.com/gentoo/gentoo/blob/71eef902e2b5e371fca6f9e81b6d69e995990977/net-p2p/mldonkey/files/mldonkey.initd#L86
* In a separate commit, removes some stray patches in `files/` that are no longer used.


juippis (approved):

juippis:
https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=30cf999cd2ab9c4d29751305c986d4e692721d2c

PR #19396, opened 2021-02-09
app-crypt/pinentry: add USE=efl

bjacquin:
As per https://git.gnupg.org/cgi-bin/gitweb.cgi?p=pinentry.git;a=blob;f=NEWS, pinentry 1.1.1 adds an EFL based pinentry.

Bug: https://bugs.gentoo.org/767181

### DIFF
--- a/app-crypt/pinentry/pinentry-1.1.1-r1.ebuild
+++ b/app-crypt/pinentry/pinentry-1.1.1-r1.ebuild
@@ -1,0 +1,90 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools flag-o-matic qmake-utils toolchain-funcs
+
+DESCRIPTION="Simple passphrase entry dialogs which utilize the Assuan protocol"
+HOMEPAGE="https://gnupg.org/aegypten2"
+SRC_URI="mirror://gnupg/${PN}/${P}.tar.bz2"
+
+LICENSE="GPL-2"
+SLOT="0"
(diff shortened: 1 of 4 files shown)


app-crypt/pinentry/pinentry-1.1.1-r1.ebuild
```diff
@@ -0,0 +1,90 @@
...
+}
+
+src_configure() {
+	[[ "$(gcc-major-version)" -ge 5 ]] && append-cxxflags -std=gnu++11
+
+	export QTLIB="$(qt5_get_libdir)"
+
+	econf \
+		$(use_enable emacs pinentry-emacs) \
+		$(use_enable efl pinentry-efl) \
```
juippis, line 61: If you wanted to make this appear in alphabetical order, you missed it juust by a bit ;)

juippis (approved):
I didn't try this PR, but made similar changes locally and can confirm that pinentry-efl does work. At least when running enlightenment WM.

app-crypt/pinentry/pinentry-1.1.1-r1.ebuild
```diff
@@ -0,0 +1,90 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools flag-o-matic qmake-utils toolchain-funcs
+
+DESCRIPTION="Simple passphrase entry dialogs which utilize the Assuan protocol"
+HOMEPAGE="https://gnupg.org/aegypten2/index.html"
```
vaukai, line 9: "index.html" not needed.
HOMEPAGE="https://gnupg.org/aegypten2/"

bjacquin: fixed in next revision

app-crypt/pinentry/pinentry-1.1.1-r1.ebuild
bjacquin: fixed in next revision

PR #19404, opened 2021-02-10
sci@ dev-python/* cleanup

jsmolic:


epsilon-0:
ok for dropping plotly

PR #19411, opened 2021-02-11
dev-go/act: new package

waebbl:
Closes: https://bugs.gentoo.org/770043

Package is needed to bump media-libs/lib3mf-2.1.0. This package has pre-built binary blobs available for x86_64, win64 and darwin, but not for x86_32 and arm, hence the keywords had to be dropped temporarily. This package should be able to re-enable those keywords.


waebbl:
The package does not use the standard go layout, hence the functions from golang-build.eclass can IMO not be used.
I'm having not much experience in building go packages. If the eclass functions can actually be used, and I'm missing something, I'd be happy for some pointers to fix the ebuild.

dev-go/act/act-1.6.0.ebuild
```diff
@@ -0,0 +1,82 @@
...
+SRC_URI="https://github.com/Autodesk/AutomaticComponentToolkit/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/AutomaticComponentToolkit-${PV}"
+
+LICENSE="BSD-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RESTRICT="strip test"
+
+DOCS=( CODE_OF_CONDUCT.md CONTRIBUTING.md README.md )
```
juippis, line 19: I don't really think users need to be aware of the `CODE_OF_CONDUCT.md`, it's more of a contributor-thing. (Same goes with your freecad).

I'd just match the default docs, and remove this line altogether from this ebuild.

dev-go/act/act-1.6.0.ebuild
```diff
@@ -0,0 +1,82 @@
...
+				;;
+		*)
+				die "unknown GOOS for ${1:-${CHOST}}"
+				;;
+	esac
+}
+
+go-tuple() {
+	echo "$(go_os $@)_$(go_arch $@)"
+}
```
juippis, line 65: Do we really need to care about all this? I see *lib3mf* is only keyworded for `amd64 arm64 x86`.

dev-go/act/act-1.6.0.ebuild
```diff
@@ -0,0 +1,82 @@
...
+
+src_compile() {
+	export GOARCH=$(go_arch)
+	export GOOS=$(go_os)
+	if [[ ${ARCH} == arm ]]; then
+		export GOARM=$(go_arm)
+	fi
+
+	cd "${S}"/Source
```
juippis, line 74: `|| die`. You can use `pushd -> popd` if you like it better.

dev-go/act/act-1.6.0.ebuild
waebbl: I added them in case someone wants to do the step from a user into actually helping with a project. But I'm ok, to drop these files.

dev-go/act/act-1.6.0.ebuild
waebbl: Good catch. Forgot about `die` here

dev-go/act/act-1.6.0.ebuild
waebbl: I'm not sure about what we can remove here. As far as I understand this, go needs a tuple set before being called. We can probably cleanup the go_os (all non-linux) and go_arch (ppc and s390) until lib3mf gets keyworded for any of those.
We could also simplify it by hardcoding linux for go-os in go-tuple

waebbl: What do you think about the last changes?

juippis (commented):
Just a status update, I tested the previous commit and it worked fine. Give me until tomorrow to think about this, but so far LGTM, even though I believe the arm checks in here are irrelevant for arm64.

waebbl:
Sadly, I don't have a arm machine to test this. I don't actually know exactly how go is working, found those functions in the dev-lang/go ebuild and thought the env vars are needed to properly run go. I can be wrong on this. Yet, those env vars control the way go is building packages in some way, see `go env` and `go help environment`.

dev-go/act/act-1.6.0.ebuild
```diff
@@ -0,0 +1,57 @@
...
+			;;
+	esac
+}
+
+src_compile() {
+	export GOARCH=$(go_arch)
+	export GOOS=linux
+	if [[ ${ARCH} == arm ]]; then
+		export GOARM=$(go_arm)
+	fi
```
juippis, line 47: @thesamesam what do you think? Are these relevant for arm64? I don't think they are. AFAIK this doesn't need to be keyworded with arm at all, just arm64.

waebbl: > AFAIK this doesn't need to be keyworded with arm at all, just arm64.

That's correct, lib3mf-1 has only support for arm64.

thesamesam (approved):

dev-go/act/act-1.6.0.ebuild
thesamesam: This'll do for now, but please speak to William about possibly including this in golang-base. It's fine to keep this in because there's no sense in gatekeeping future support.

dev-go/act/act-1.6.0.ebuild
```diff
@@ -0,0 +1,57 @@
...
+DESCRIPTION="Automated Component Toolkit code generator"
+HOMEPAGE="https://github.com/Autodesk/AutomaticComponentToolkit"
+SRC_URI="https://github.com/Autodesk/AutomaticComponentToolkit/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/AutomaticComponentToolkit-${PV}"
+
+LICENSE="BSD-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RESTRICT="strip test"
```
thesamesam, line 17: No tests? :(

thesamesam:
Added https://github.com/gentoo/gentoo/commit/bb5b84a2390610c4db06fc852fb1df33fef34f17 for a verbose build.

dev-go/act/act-1.6.0.ebuild
waebbl: Sadly no. There's an `Examples/UnitTest` directory but it looks like it's not updated for this version.

waebbl:
Thanks!

PR #19412, opened 2021-02-11
media-libs/lib3mf 2.1.0: version bump

waebbl:
Closes: https://bugs.gentoo.org/769275

Current versions of media-gfx/openscad don't work with this API version and need to be restricted to <lib3mf-2. The upcoming openscad version bump does work with it. See
Bug: https://bugs.gentoo.org/769278


waebbl:
Switching openscad live ebuild to use cmake instead of qmake. Works with both, lib3mf-1.8.1 and lib3mf-2.1.0. Latter needs a patch due to a case change from lib3MF.pc to lib3mf.pc.

waebbl:
For the act dependency needed to re-add keywords for x86 and arm see https://github.com/gentoo/gentoo/pull/19411

juippis:
Is the act dependency still needed? Can we somehow combine this PR with #19714?

juippis:
Actually could you combine all the related work to a single PR. I'm at loss, and am fearing there'll be tree breakage or unnecessary rebuilds for users. Openscad is not a little package.

waebbl:
@juippis Sure, I can combine them in a single PR.

The act package is 'only' needed to get x86 and arm keywords back to lib3mf. For amd64 the included binary does the job. Should I add it to the PR too?

juippis:
Up to you, I quickly glanced at the ebuild and it might need .. a lot of fixes before it can be pushed. So maybe best to leave it out for now.

waebbl:
Yes it might nieed some rework. The point with the package is, that upstream doesn't follow the usual way Go packages are developed, so the Go eclasses don't seem to be useful. On the other hand the build is rather trivial, so I decided for the way it's done.
I agree, it's better to leave it out for now and wait for a dedicated review of the package.

juippis:
I've nearly caught up with our PR work queue, please ping me in ~a week from now for the act package and I'll take a look.

media-gfx/openscad/openscad-2019.05-r4.ebuild
```diff
@@ -46,7 +46,7 @@ RDEPEND="
 	media-libs/freetype
 	>=media-libs/glew-2.0.0:0=
 	media-libs/harfbuzz:=
-	media-libs/lib3mf
+	<media-libs/lib3mf-2
```
juippis, line 49: Will the update cause runtime issues? Ie should this be revbumped for this...

media-gfx/openscad/openscad-2021.01.ebuild
```diff
@@ -0,0 +1,110 @@
...
+	dev-qt/qtmultimedia:5
+	dev-qt/qtnetwork:5
+	dev-qt/qtopengl:5
+	dev-qt/qtwidgets:5
+	media-gfx/opencsg
+	media-libs/fontconfig
+	media-libs/freetype
+	>=media-libs/glew-2.0.0:0=
+	media-libs/harfbuzz:=
+	media-libs/lib3mf
```
juippis, line 46: Does this work with both 1 and 2? Is some restriction needed? That's also why I asked to combine the PRs so that no modification is needed for this version after merging.

juippis (commented):
Let me know the answers to my questions, there's no need for you to do anything, I can fix those minor issues while merging.

media-gfx/openscad/openscad-2019.05-r4.ebuild
waebbl: 2019.05 will not build with lib3mf-2. Lib3mf-2 has been released after 2019.05 and support for it has not been backported

media-gfx/openscad/openscad-2021.01.ebuild
waebbl: The new version works with both lib3mf versions.

media-gfx/openscad/openscad-2019.05-r4.ebuild
juippis: The idea is, people who have 2019.05-r4 installed now will update to lib3mf-2, because the restriction won't hit their package database before a rebuild. So, if lib3mf-2 will cause **runtime** issues to openscad-2019.05, this ebuilds needs t obe revbumped so that won't happen.

waebbl: As the package won't build with lib3mf-2 I suspect it will cause runtime issues, whenever someone tries to im-/export a 3mf format file.
I'm gonna revbump the package then.

juippis (approved):

PR #19415, opened 2021-02-11
kde-plasma/kde-gtk-config: Don't print debug message

zippy2:
I've noticed that whenever a GTK application is started the
stdout is polluted with the following message:

WINDOW DECORATIONS RELOADED

After some research I came across KDE bug:

  https://bugs.kde.org/show_bug.cgi?id=431365

which says the message is a debug message and could be dropped.
In the upstream repository, the following commit was merged which
does exactly that:

https://invent.kde.org/plasma/kde-gtk-config/-/commit/62e272df848d6848482ceb534d14bf36c9fde241.patch

Backport the commit until the version of kde-gtk-config is
bumped.

Closes: https://bugs.gentoo.org/770097


zippy2:
cc @a17r

a17r (commented):
Thanks!

PR #19419, opened 2021-02-11
*/*:  gcc-10 fixes (and EAPI-7 porting)

jsmolic:
* All packages are dead upstream for many years, and patching requires significant code rewrite, so we can just use `flag-o-matic` to enable building with gcc-10


jsmolic:
merged

PR #15017, opened 2020-03-19
app-misc/ola: Add to tree in version 0.10.7

kripton:
This PR adds the new, proxy-maintained package app-misc/ola to the tree. Latest upstream version 0.10.7 is used. However, since that version is a bit older, three patches are needed to compile against the packages currently in the Gentoo Portage tree. I'm also working with upstream to have those changes included in the next versions (see https://github.com/OpenLightingProject/ola/pull/1621 and https://github.com/OpenLightingProject/ola/pull/1620).

The ebuild was tested on amd64 with python2_7, python3_6 and python3_7 and the tests included with the package all succeeded there.

The ebuild also works as live ebuild (version ola-9999.ebuild). In that case, the patches are not applied. I don't commit that live version with this PR, however. I hope it's okay to have it prepared already so it could be symlinked when needed.

Looking forward to your comments :)

Closes: https://bugs.gentoo.org/429576

PR #19420, opened 2021-02-11
app-misc/ola: Add in version 0.10.8

kripton:
This PR supersedes #15017 since there was a new version released upstream that includes all patches that were needed for 0.10.7.

Description copied from the old PR:
This PR adds the new, proxy-maintained package app-misc/ola to the tree. Latest upstream version 0.10.8 is used.

The ebuild was tested on amd64 with python3_7, python3_8 and python3_9 and the tests included with the package all succeeded there.

The ebuild also works as live ebuild (version ola-9999.ebuild) I don't commit that live version with this PR, however. I hope it's okay to have it prepared already so it could be symlinked when needed. The live ebuild can be found in my overlay: https://github.com/kripton/kripton-overlay/tree/master/app-misc/ola

Looking forward to your comments :)

Closes: https://bugs.gentoo.org/429576


kripton:
Closing this since ola was added in a875aa576c8ac0578f63e5a24e190129dad2733b

PR #19423, opened 2021-02-12
net-misc/sipp: Version 3.6.1

jkroonza:


net-misc/sipp/sipp-3.6.1.ebuild
```diff
@@ -0,0 +1,52 @@
...
+
+	cmake_src_prepare
+}
+
+src_configure() {
+	local u
+	local mycmakeargs=()
+
+	for u in ${IUSE//[+-]/}; do
+		mycmakeargs+=(-DUSE_${u^^}=$(usex ${u} 1 0))
```
thesamesam, line 41: I think this hurts readability and is going to make it a pain if we have to adjust any of them. Could we lay them out explicitly?

jkroonza: Sure ... I get why this seems nasty :).

PR #19438, opened 2021-02-13
gui-libs/display-manager-init: revbump packages with init files referencing xdm

epsilon-0:
for stable packages added a new ebuild with revision bump, for unstable packages moved latest unstable with revision bump.


app-misc/g15stats/g15stats-1.9.7-r2.ebuild
```diff
@@ -12,7 +12,6 @@ SRC_URI="mirror://sourceforge/g15daemon/${P}.tar.bz2"
 LICENSE="GPL-2"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE=""
```
a17r, line 15: packages with maintainers should not get unrelated changes made to them, even trivial ones, on such bulk changes.

Polynomial-C (approved):

PR #19440, opened 2021-02-13
retext 7.2.0

holgersson32644:
- app-editors/retext: Bump to 7.2.0
- app-editors/retext: Update the live ebuild


app-editors/retext/retext-7.2.0.ebuild
```diff
@@ -0,0 +1,69 @@
...
+
+EAPI=7
+
+# Please don't add pypy support before testing if it's actually supported. The
+# old compat matrix is no longer accessible as of 2021-02-13 but stated back
+# in 2020-07-05 that PyQt5 was explicitly not supported.
+PYTHON_COMPAT=( python3_{8,9} )
+
+inherit distutils-r1 optfeature virtualx xdg-utils
+DISTUTILS_USE_SETUPTOOLS=rdepend
```
juippis, line 12: This needs to be set before inheriting *distutils-r1.eclass*. But it looks like false anyway.

app-editors/retext/retext-7.2.0.ebuild
```diff
@@ -0,0 +1,69 @@
...
+"
+DEPEND="${RDEPEND}"
+
+src_test() {
+	virtx distutils-r1_src_test
+}
+
+python_test() {
+	esetup.py test
+}
```
juippis, line 52: Can be replaced with `distutils_enable_tests setup.py`. But weirdly that ends up in a modulenotfound error, 
```
ERROR: test_editor (unittest.loader._FailedTest)
ImportError: Failed to import test module: test_editor
  File "/usr/lib/python3.8/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
  File "/var/tmp/portage/app-editors/retext-7.2.0/work/ReText-7.2.0/tests/test_editor.py", line 28, in <module>
    from PyQt5.QtTest import QTest
ModuleNotFoundError: No module named 'PyQt5.QtTest'
```
So it might need `dev-python/PyQt5[testlib]` as a test-dep? Could you check?

app-editors/retext/retext-7.2.0.ebuild
holgersson32644: Right, needs to be bdepend - I fixed it locally.

juippis (approved):

PR #19444, opened 2021-02-13
kde-apps/kalgebra: fix LICENSE

mm1ke:

Regarding LICENSE: I've checked some sourcefiles (for example: https://github.com/KDE/kalgebra/blob/master/src/main.cpp) and they all say: `...as published by the Free Software Foundation; either version 2 of the License, or (at your option) any later version.`

### DIFF
--- a/kde-apps/kalgebra/kalgebra-20.08.3.ebuild
+++ b/kde-apps/kalgebra/kalgebra-20.08.3.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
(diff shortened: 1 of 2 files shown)


a17r:
Please split it in two separate commits. I started an inquiry upstream about the whereabouts of kalgebra website, I don't think it is correct to use kalgebramobile instead.

ConiKost (approved):

ConiKost:
Ok, won't commit that ;-)

ConiKost:
> 
> 
> Please split it in two separate commits. I started an inquiry upstream about the whereabouts of kalgebra website, I don't think it is correct to use kalgebramobile instead.

The mobile homepage at least links to the normal application, but also provides a mobile application. The upstream Repo seems to have both included.

a17r:
I think it is an upstream mistake and after reporting upstream you might look at a broken website on Monday already.

ConiKost:
> 
> 
> I think it is an upstream mistake and after reporting upstream you might look at a broken website uri on Monday already.

Let's see, what they say about your inquery :-)

But https://apps.kde.org/ lists that as mobile, normal seems to be gone.
According to archive.org, this seems to be longer the case its not there anymore.

a17r:
If anything, the normal page should link to some mobile bits, not the other way round.

mm1ke:
I've updated this PR to only fix the LICENSE.
Lets see how upstream decides, we still can fix the website later on if needed.

PR #10017, opened 2018-09-29
sci-libs/cantera: New package

band-a-prend:
The first commit adds new package **sci-libs/cantera** of version 2.4.0
(cantera-2.4.0.ebuild):

Description:
Cantera is an open-source suite of object-oriented software tools
for problems involving chemical kinetics, thermodynamics,
and/or transport processes. Cantera can be used from Python 
and Matlab*, or in applications written in C/C++ and Fortran 90.

\* Build of Matlab interface is not implemented  within this ebuild currently.

Homepage and documentation: http://cantera.org
GitHub page: https://github.com/Cantera/cantera

IUSE="+cti fortran pch +python test"

Local USE flags:
cti - Install CTI tools (ck2cti, ctml_writer) for conversion of Chemkin files
to Cantera format.

Closes: https://bugs.gentoo.org/200425

After long break since 2006 when version 1.7.0 mentioned in this issue
was released the development was resumed in 2015 with release
of version 2.0.0 with new SCons-based build system.
So this is entirely new ebuild from scratch of this package.

@proxy-maint

-------
The second commit adds new supplementary package **app-doc/cantera-docs**
that installs the API documentation for 'sci-libs/cantera' package.

Description:
Cantera Doxygen API Documentation for C++ library
and Sphinx API Documentation for CTI tool, Python module and Matlab interface.

Unfortunately there is no official documentation tarball for latest version
of Cantera. So this package download the tarball build by myself from sources.
This tarball includes some modifications of 'Menubar' to navigate local sections

cantera_sphinx_menu_crop

and patch for async loading remote javascript and css files that allow
to render pages without long delays in some cases of offline reading.

PR #19448, opened 2021-02-13
sci-libs/cantera: 2.5.0 version bump

band-a-prend:
Cantera now uses yaml as input format and thus depends on
`dev-cpp/yaml-cpp` and `dev-python/ruamel-yaml`.

The following dependences in DEPEND are actually not required to build
and if absence the appropriate 2 tests just will be skipped:
```
test? (
	...
	python? (
	$(python_gen_cond_dep '
		dev-python/h5py[${PYTHON_MULTI_USEDEP}]
		dev-python/pandas[${PYTHON_MULTI_USEDEP}]
        ')
)
```

Is it worth then to drop it from DEPEND?


thesamesam:
We may as well keep maximal coverage for tests if possible, so don't drop.

thesamesam:
Also:
```
   /usr/lib64/libcantera.so.2.5.0
   /usr/lib/python3.8/site-packages/cantera/_cantera.cpython-38-x86_64-linux-gnu.so
   /usr/lib64/libcantera.a
```

Can we make the static libs (.a) conditional or drop them [entirely](https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0302) (preferred)? Thanks!

band-a-prend:
The build of C++ static library it's not controlled by build script and fortran interface is provided only via static library. So do not drop already build static library for C++ it was decided to left it that was discussed with at initially cantera-2 4.0 new package ebuild: https://github.com/gentoo/gentoo/pull/10017#discussion_r229210565

PR #19449, opened 2021-02-13
dev-java/junit: add version 4.13.1

fee1-dead:
Many methods added/made public in [4.13](https://github.com/junit-team/junit4/blob/384d804930e054a331a0be32826a99e6fd8a35a4/doc/ReleaseNotes4.13.md) and [4.13.1](https://github.com/junit-team/junit4/blob/384d804930e054a331a0be32826a99e6fd8a35a4/doc/ReleaseNotes4.13.1.md).


fordfrog:
@fee1-dead could you please put your real name into Signed-off-by? and i would also suggest to lift up the min jdk/jre versions to 1.8 as this is the lowest java version we support. thank you!

fee1-dead:
looks like dev-java/hamcrest-core needs a bump with this as well. I will look in to this later

PR #19450, opened 2021-02-13
app-arch/engrampa: add missing optfeature inherit

mm1ke:
Installing this package doesn't show me the optional packages. 
This is because the `optfeature` doesn't get inherited, so lets fix it.


ConiKost (approved):

NP-Hardass:
@ConiKost Please don't make changes to another dev's packages without consulting them.  That goes against Gentoo policy.

PR #19451, opened 2021-02-13
sys-cluster/nova: add missing udev inherit

mm1ke:
Hi,

All ebuilds are calling `udev_newrules`
for example: https://github.com/gentoo/gentoo/blob/b3675bae9e25428c81cf9d3eba2ba6e433f4ad4c/sys-cluster/nova/nova-22.1.0.ebuild#L201
but doesn't inherit the `udev` eclass which is why i think this cannot work right now. (i haven't tested it since it pulls alot of deps for me). Never the less the ebuilds should inherit the eclass anyway.
On a side note: As far as i can see the `eutils` eclass could be removed - it doesn't get used at all :)

Please review.

### DIFF
--- a/sys-cluster/nova/nova-2020.1.9999.ebuild
+++ b/sys-cluster/nova/nova-2020.1.9999.ebuild
@@ -4,7 +4,7 @@
 EAPI=7
 
 PYTHON_COMPAT=( python3_7 )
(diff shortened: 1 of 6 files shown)


prometheanfire:
merged, nice find

PR #19455, opened 2021-02-13
profiles/use.desc: add global man flag

epsilon-0:


juippis (approved):
@gentoo/qa check please.

Dessa:
sys-auth/sssd is missing from this list.
i changed manpages to man with my pr but the oldest stable still uses manpages useflag

epsilon-0:
> sys-auth/sssd is missing from this list.
> i changed manpages to man with my pr but the oldest stable still uses manpages useflag

added :D 
keeping the manpages flag until the package is removed.

PR #19473, opened 2021-02-15
dev-java/asm{,-*}: new packages and bumps

kma06:
dev-java/asm bumped to the latest versions, because upstream doesn't
publish asm-all anymore, we should also split it into components.

The dependencies are as follows:
 - asm: no dependency
 - asm-tree: asm
 - asm-analysis: asm, asm-tree
 - asm-commons and asm-util: asm, asm-tree, asm-analysis

Testing is still restricted because it uses JUnit 5 for testing.


fordfrog:
@kma06 nice addition, thanks!

PR #19474, opened 2021-02-15
sys-fs/compsize: bump to 1.5

telans:


thesamesam:
Thanks, added respect CC fix too.

PR #19482, opened 2021-02-15
media-fonts/cascadia-code: Version bump 2102.03

janhenke:


media-fonts/cascadia-code/cascadia-code-2102.03.ebuild
```diff
@@ -0,0 +1,18 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+FONT_SUFFIX="ttf"
```
ConiKost, line 6: Please move it after BDEPEND.

ConiKost (changes requested):

media-fonts/cascadia-code/cascadia-code-2102.03.ebuild
thesamesam: I normally like to put eclass variables before the inherit in case they use them outside of a specific phase but at inherit-time.

media-fonts/cascadia-code/cascadia-code-2102.03.ebuild
```diff
@@ -0,0 +1,18 @@
...
+inherit font
+
+DESCRIPTION="A monospaced font developed for Windows Terminal"
+HOMEPAGE="https://github.com/microsoft/cascadia-code"
+SRC_URI="https://github.com/microsoft/cascadia-code/releases/download/v${PV}/CascadiaCode-${PV}.zip -> ${P}.zip"
```
thesamesam, line 11: Do they supply non-zips, allowing us to drip the BDEPEND?

media-fonts/cascadia-code/cascadia-code-2102.03.ebuild
ConiKost: Yeah, but AFAIK, with font class, it's supposted to be but after all stuff.

media-fonts/cascadia-code/cascadia-code-2102.03.ebuild
janhenke: Unfortunately not for the compiled font files. There is a source tarball, but I already looked into compiling this font some time ago and it would require quite some dependencies.

juippis (approved):

PR #19483, opened 2021-02-16
media-libs/libprojectm: fix deps&USEs with revbump

Fat-Zer:
Fix jack support: it requires qt enabled
Make pulseaudio support optional

Signed-off-by: Alexander Golubev <fatzer2@gmail.com>


Fat-Zer:
Note RESTRICT_USE are required because without `--enabled-qt`  `jack` and `pulseaudio` support just gets silently disabled.

PR #19485, opened 2021-02-16
sys-cluster/keepalived: bump to 2.2.1

hydrapolic:
Tested on ~amd64, thanks!


Whissi (changes requested):
2.2.x series added explicit systemd support (notify). Please add systemd USE flag.

hydrapolic:
Thanks @Whissi, I don't use systemd yet so haven't noticed. Pushed an update.

sys-cluster/keepalived/keepalived-2.2.1.ebuild
```diff
@@ -0,0 +1,89 @@
...
+
+	eautoreconf
+}
+
+src_configure() {
+	local init
+	if use systemd; then
+		init=systemd
+	else
+		init=custom
```
Whissi, line 45: I think `init=custom` is wrong. It's an unknown value in configure.ac, better leave the ELSE clause out.

sys-cluster/keepalived/keepalived-2.2.1.ebuild
```diff
@@ -0,0 +1,89 @@
...
+DESCRIPTION="A strong & robust keepalive facility to the Linux Virtual Server project"
+HOMEPAGE="https://www.keepalived.org/"
+SRC_URI="https://www.keepalived.org/software/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~hppa ~ia64 ~ppc ~ppc64 ~s390 ~sparc ~x86"
+IUSE="+bfd dbus -json regex snmp systemd"
+
+RDEPEND="dev-libs/libnl:=
```
Whissi, line 17: We need to pull in systemd for USE=systemd like it is required at build time, https://github.com/acassen/keepalived/blob/v2.2.1/keepalived/core/Makefile.am#L47-L50

sys-cluster/keepalived/keepalived-2.2.1.ebuild
```diff
@@ -0,0 +1,89 @@
...
+		$(use_enable systemd)
+}
+
+src_install() {
+	default
+
+	newinitd "${FILESDIR}"/keepalived.init-r1 keepalived
+	newconfd "${FILESDIR}"/keepalived.confd-r1 keepalived
+
+	systemd_newunit "${FILESDIR}"/${PN}.service ${PN}.service
```
Whissi: It looks like enabling systemd will install upstream service files: https://github.com/acassen/keepalived/blob/v2.2.1/keepalived/Makefile.am#L68-L70

I am also not a systemd user but setting correct `Type` is important now that keepalived-2.2.x supports notify, https://github.com/acassen/keepalived/commit/027157a0617bea626723af5e32d8fe6ca53855b5

@gentoo/systemd Can you help? What do you suggest? Upstream would install https://github.com/acassen/keepalived/blob/v2.2.1/keepalived/keepalived.service.in and we are currently using https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-cluster/keepalived/files/keepalived.service?id=2c77e25783da444692e66820b72472c21bbeab32

Whissi (changes requested):

sys-cluster/keepalived/keepalived-2.2.1.ebuild
hydrapolic: It means no init is chosen and we ship our own openrc scripts.

sys-cluster/keepalived/keepalived-2.2.1.ebuild
hydrapolic: I thought users of systemd will implicitly have this set and nobody with openrc will just set USE=systemd. But yes, you're right, I'll make it explicit.

sys-cluster/keepalived/keepalived-2.2.1.ebuild
hydrapolic: I've updated ours, that way ``/etc/systemd/system/keepalived.service.d/00gentoo.conf`` is present for customization and this is omitted: ``EnvironmentFile=-/etc/sysconfig/keepalived``

hydrapolic:
@Whissi, compile tested with stage3-systemd docker image, it now installs with notify support.

hydrapolic:
And sorry it took so long :(

ultrabug:
Thanks!

hydrapolic:
Thanks @Whissi and @ultrabug

PR #19486, opened 2021-02-16
dev-python/lark-parser: bump to 0.11.2

evdenis:
Signed-off-by: Denis Efremov <efremov@linux.com>


dev-python/lark-parser/lark-parser-0.11.1.ebuild
```diff
@@ -0,0 +1,16 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7..9} )
+
+inherit distutils-r1
+
+DESCRIPTION="Python module to propose a modern general-purpose parsing library for Python"
+HOMEPAGE="https://github.com/lark-parser/lark"
```
thesamesam, line 11: Looks like upstream has tests. Could you add them please?

[This](https://dev.gentoo.org/~mgorny/python-guide/distutils.html#enabling-tests) might be useful.

evdenis: tests require additional dependency Js2Py which is not presented in the packages.

thesamesam: Let's add it then! ;)

ionenwks: It's available in ::guru so maybe it should graduate :) I ran those tests while trying py3.9 compat (which passed on the previous version I might add, but did have to fetch a module). [bug #769596](https://bugs.gentoo.org/769596)

xgqt: > It's available in ::guru so maybe it should graduate :) I ran those tests while trying py3.9 compat (which passed on the previous version I might add, but did have to fetch a module). [bug #769596](https://bugs.gentoo.org/769596)

@thesamesam @evdenis @ionenwks 
https://github.com/gentoo/gentoo/pull/19498

xgqt:
js2py and pyjsparser merged into ::gentoo

https://github.com/gentoo/gentoo/pull/19498

xgqt:
js2py and pyjsparser keywording request: https://bugs.gentoo.org/772377

dev-python/lark-parser/lark-parser-0.11.1.ebuild
thesamesam: Beaten to it!

juippis:
`~x86` keyword added, please rebase against fresh ::gentoo.

mattst88:
What's the status here? Test deps are in the tree I think. Do lark's tests now pass? Looks like there's a new version as well.

evdenis:
I've tried my best to take commands from tutorial to enable testing. But the best that I achieved is still:
```
# with USE=test
...
>>> Source compiled.
>>> Test phase [not enabled]: dev-python/lark-parser-0.11.2
...
```

dev-python/lark-parser/lark-parser-0.11.2.ebuild
```diff
@@ -0,0 +1,30 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7..9} )
+DISTUTILS_USE_SETUPTOOLS=bdepend
```
juippis, line 7: This is the default and can be removed.

dev-python/lark-parser/lark-parser-0.11.2.ebuild
```diff
@@ -0,0 +1,30 @@
...
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="test"
+RESTRICT="!test? ( test )"
+
+BDEPEND="
+	dev-python/setuptools[${PYTHON_USEDEP}]
```
juippis, line 22: Already defined by `DISTUTILS_USE_SETUPTOOLS`.

juippis (commented):
If I had to guess, the pypi tarball doesn't contain test files while Github archive does. So please switch to Github release then.

juippis:
Well disregard that, 
```
>>> Source compiled.
>>> Test phase: dev-python/lark-parser-0.11.2
...
...
...
Ran 972 tests in 29.334s

OK (skipped=97)
 * python3_7: running distutils-r1_run_phase _distutils-r1_clean_egg_info
 * python3_8: running distutils-r1_run_phase _distutils-r1_clean_egg_info
 * python3_9: running distutils-r1_run_phase _distutils-r1_clean_egg_info
>>> Completed testing dev-python/lark-parser-0.11.2
```

PR #19488, opened 2021-02-16
Bumps for media-libs/kcolorpicker and media-libs/kimageannotator

dflogeras:
As discussed with @a17r, single consumer in tree kde/spectacle should be bumped to depend on >=kimageannotator-0.4.1 to mitigate SONAME change, and old versions removed afterwards.

https://bugs.gentoo.org/770949


media-libs/kimageannotator/kimageannotator-0.4.1.ebuild
```diff
@@ -0,0 +1,47 @@
...
+
+MY_PN=kImageAnnotator
+MY_P="${MY_PN}-${PV}"
+
+DESCRIPTION="Tool for annotating images"
+HOMEPAGE="https://github.com/ksnip/kImageAnnotator"
+SRC_URI="https://github.com/ksnip/${MY_PN}/archive/v${PV}.tar.gz -> ${MY_P}.tar.gz"
+
+LICENSE="LGPL-3+"
+SLOT="0"
```
juippis, line 17: If the SONAME change is breaking revdeps, this should specify a subslot. 
https://devmanual.gentoo.org/general-concepts/slotting/index.html#sub-slots

@a17r please take it from here? :)

a17r: no need to, as it is a fix, and the only revdep can be bumped in lockstep with a new version requirement.

a17r (approved):
Thanks!

PR #19489, opened 2021-02-16
dev-python/clade: add 3.3.7

evdenis:
Add clade util for building a compilation database.
https://github.com/17451k/clade


dev-python/cchardet/metadata.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+<maintainer type="project">
```
thesamesam, line 4: We [shouldn't](https://wiki.gentoo.org/wiki/Project:Python#Adding_packages) add python@ to projects unless you've agreed it with them (or any project).

dev-python/cchardet/metadata.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
thesamesam, line 1: Indent with tabs? Also, add proxy-maint.

dev-python/clade/clade-3.3.7.ebuild
```diff
@@ -0,0 +1,26 @@
...
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7..9} )
+DISTUTILS_USE_SETUPTOOLS=rdepend
+
+inherit distutils-r1
+
+DESCRIPTION="Tool for extracting information about software build process and source code"
+HOMEPAGE="https://pypi.org/project/clade/"
```
thesamesam, line 12: Add GitHub to [metadata](https://devmanual.gentoo.org/ebuild-writing/misc-files/metadata/index.html) (<remote type="github-id">).

dev-python/clade/clade-3.3.7.ebuild
```diff
@@ -0,0 +1,26 @@
...
+KEYWORDS="~amd64"
+
+RDEPEND="
+	dev-python/ujson[${PYTHON_USEDEP}]
+	dev-python/chardet[${PYTHON_USEDEP}]
+	dev-python/cchardet[${PYTHON_USEDEP}]
+	dev-python/graphviz[${PYTHON_USEDEP}]
+	dev-python/ply[${PYTHON_USEDEP}]
+"
+BDEPEND="dev-util/cmake"
```
thesamesam, line 26: Tests too, [please](https://dev.gentoo.org/~mgorny/python-guide/distutils.html#enabling-tests)! Also, if we're using cmake, do we need to be using the eclass to pick up the goodness required (it does some special stuff to adapt to Gentoo)?

dev-python/clade/metadata.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+<maintainer type="project">
```
thesamesam, line 4: See earlier comments re Python project.

thesamesam (changes requested):

dev-python/clade/clade-3.3.7.ebuild
evdenis: > Tests too, please!

Tests will not work. They require additional external components to download.

> Also, if we're using cmake, do we need to be using the eclass to pick up the goodness required

I think no. setup.py script calls cmake internally (https://github.com/17451k/clade/blob/master/setup.py#L61-L67) to build C-libs for intercepting.

thesamesam: In that case, we normally ask upstream to make downloading optional / allow local copies, and try to download manually in SRC_URI conditionally based on test?. But if it's too much work to do that, we just ask them to make it optional/allow local copies.

evdenis: I already asked the author to make tests that depends on external components conditional.

PR #19500, opened 2021-02-17
net-mail/mu: Add live ebuild

00-matt:


juippis (commented):
Do we really want a live ebuild? Is there any need for it?
https://devmanual.gentoo.org/ebuild-writing/functions/src_unpack/vcs-sources/index.html#disadvantages-of-vcs-sources

You can't reach fast enough to it's faults through proxy-maint.

00-matt:
I thought it could be nice since there's been quite a few changes in master since the last release, but yes it might not be the most useful if it takes a couple of days for fixes to get merged in

juippis:
You can always pick a good commit from upstream and make a snapshot release from that, calling it something like `1.4.15_p20210218` (depending which date the commit was made).
Check the idea from https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/qtcompress/qtcompress-0_p20180513.ebuild

PR #19507, opened 2021-02-17
www-servers/lighttpd: Update mime types, #399401

alanswanson:
Update with commonly used file types without the bloat from
generation using entire /etc/mime.types source.

Closes: https://bugs.gentoo.org/399401


juippis (approved):
@herbmillerjr no objections I presume?

PR #19511, opened 2021-02-17
x11-misc/rss-glx: Rescue from last rites and fix building with clang

miller-alex:


x11-misc/rss-glx/rss-glx-0.9.1-r2.ebuild
```diff
@@ -0,0 +1,137 @@
...
+		$(use_enable bzip2) \
+		$(use_enable openal sound) \
+		$(use_with quesoglc) \
+		--bindir="${EPREFIX}/usr/$(get_libdir)/misc/xscreensaver" \
+		--enable-shared --disable-static \
+		--with-configdir="${EPREFIX}/usr/share/xscreensaver/config"
+}
+
+src_install() {
+	default
```
SoapGentoo, line 71:
```
libtool: link: x86_64-pc-linux-gnu-g++ -O2 -pipe -march=native -fdiagnostics-color=always -frecord-gcc-switches -Wall -Wl,-O1 -Wl,--hash-style=gnu -Wl,--sort-common -Wl,--defsym=__gentoo_check_ldflags__=1 -o fieldlines driver.o fieldlines.o  -lGLEW -Wl,--as-needed -lXext -lGLU -lGL -lSM -lICE -lX11 -lm
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: /var/tmp/portage/x11-misc/rss-glx-0.9.1-r2/work/rss-glx_0.9.1/src/PixelCity/.libs/libPixelCity.a(Deco.o): undefined reference to symbol '_ZTVN10__cxxabiv121__vmi_class_type_infoE@@CXXABI_1.3'
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: /usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/libstdc++.so.6: error adding symbols: DSO missing from command line
```

SoapGentoo (changes requested):

x11-misc/rss-glx/rss-glx-0.9.1-r2.ebuild
SoapGentoo: you're linking a C and C++ code using the C compiler, this can't possibly work.

SoapGentoo (approved):

miller-alex:
@SoapGentoo did you intentionally drop the stable ebuild?

SoapGentoo:
I have indeed, that was not a mistake. The idea is: we're unmasking it, so we're effectively starting from scratch again. This is also the procedure defined in the manual. We can shorten the time to calling stabilizations to 14 days.

PR #19513, opened 2021-02-18
net-misc/apt-cacher-ng: bump to 3.6_p1, properly make systemd optional

ajakk:
Fix broken symlink by installing the symlink's target instead. Re-add
systemd dependency/IUSE because an installed binary links to it when
available. Drop debian patches because they don't exist for this
version, and appear to be necessary anyway.

Closes: https://bugs.gentoo.org/770046

```
diff --git a/net-misc/apt-cacher-ng/apt-cacher-ng-3.5_p3.ebuild b/net-misc/apt-cacher-ng/apt-cacher-ng-3.6_p1.ebuild
index 56294639238..907ea8a00d4 100644
--- a/net-misc/apt-cacher-ng/apt-cacher-ng-3.5_p3.ebuild
+++ b/net-misc/apt-cacher-ng/apt-cacher-ng-3.6_p1.ebuild
@@ -8,13 +8,12 @@ inherit cmake
 DESCRIPTION="Yet another caching HTTP proxy for Debian/Ubuntu software packages"
 HOMEPAGE="https://www.unix-ag.uni-kl.de/~bloch/acng/
        https://packages.qa.debian.org/a/apt-cacher-ng.html"
-SRC_URI="mirror://debian/pool/main/a/${PN}/${PN}_${PV/_*}.orig.tar.xz
-       mirror://debian/pool/main/a/${PN}/${PN}_${PV/_p/-}.debian.tar.xz"
+SRC_URI="mirror://debian/pool/main/a/${PN}/${PN}_${PV/_*}.orig.tar.xz"

 LICENSE="BSD-4 ZLIB public-domain"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE="doc fuse tcpd"
+IUSE="doc fuse systemd tcpd"

 DEPEND="acct-user/apt-cacher-ng
        acct-group/apt-cacher-ng
@@ -23,6 +22,7 @@ DEPEND="acct-user/apt-cacher-ng
        dev-libs/openssl:0=
        sys-libs/zlib
        fuse? ( sys-fs/fuse:0 )
+       systemd? ( sys-apps/systemd )
        tcpd? ( sys-apps/tcp-wrappers )"
 BDEPEND="virtual/pkgconfig"
 RDEPEND="${DEPEND}
@@ -31,7 +31,6 @@ RDEPEND="${DEPEND}
 PATCHES=(
        "${FILESDIR}/${PN}-3.3.1-flags.patch"
        "${FILESDIR}/${PN}-3.5-perl-syntax.patch"
-       "${WORKDIR}/debian/patches/debian-changes"
 )

 S="${WORKDIR}/${P/_*}"
@@ -58,18 +57,18 @@ src_configure() {
                "-DSDINSTALL=1"
        )

-       if tc-ld-is-gold; then
-               mycmakeargs+=( "-DUSE_GOLD=yes" )
-       else
-               mycmakeargs+=( "-DUSE_GOLD=no" )
-       fi
-
        cmake_src_configure

        sed -ie '/LogDir/s|/var/tmp|/var/log/'"${PN}"'|g' "${BUILD_DIR}"/conf/acng.conf || die
 }

 src_install() {
+       # README is a symlink to doc/README and README automatically gets
+       # installed, leading to a broken symlink installed. Fix this by removing
+       # the symlink then installing the actual README. https://bugs.gentoo.org/770046
+       rm README || die
+       dodoc doc/README
+
        newinitd "${FILESDIR}/initd-r3" "${PN}"
        newconfd "${FILESDIR}/confd-r2" "${PN}"
```


juippis (commented):
You should patch the `systemd` IUSE to be optional in the build system. As apt-cacher-ng uses CMake, that should be easy. This should also give you extra pointers wrt recruiting as it's connected to one of the questions ;)

Have an example here: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/check/files/check-0.14.0-r2-disable-automagic-dep.patch

ajakk:
> You should patch the `systemd` IUSE to be optional in the build system. As apt-cacher-ng uses CMake, that should be easy. This should also give you extra points wrt recruiting as it's connected to one of the questions ;)
> 
> https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies
> 
> Have an example here: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/check/files/check-0.14.0-r2-disable-automagic-dep.patch

Well, I understand why this is an issue, I just didn't notice it, so thank you. This push should rectify it.

PR #19514, opened 2021-02-18
sys-power/uhubctl: amend for chromium os toolchain

coconutruben:
The two commits address two failures in the chromium os toolchain.
1. using tc-export for pkg-config
2. reverting to the * keyword

### DIFF
--- a/sys-power/uhubctl/uhubctl-2.4.0.ebuild
+++ b/sys-power/uhubctl/uhubctl-2.4.0.ebuild
@@ -24,6 +24,8 @@ src_prepare() {
 		-e "s/^GIT_VERSION/#&/" \
 		Makefile \
 		|| die
+
+	tc-export PKG_CONFIG
 }
 
 src_compile() {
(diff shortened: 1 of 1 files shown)


ZeroChaos-:
yeah, we don't set keywords like that in gentoo, and I have no idea what purpose setting PKG_CONFIG at the end of the phase serves.  I won't accept the keywords change, but if you can explain the PKG_CONFIG one I'll accept something like that.

coconutruben:
Sorry to drag this on forever, I hadn't noticed you left a note. Thank you for the feedback.
The setting of PKG_CONFIG is to allow us to feed a toolchain specific pkg-config when building. This is to leverage the work the maintainer did to allow this pipeline at https://github.com/mvp/uhubctl/commit/92736a860431da6c658f07b51cad2c293421fd77.
If you agree, please let me know if I should just do another pr with just the PKG_CONFIG?

coconutruben:
added at https://github.com/gentoo/gentoo/pull/20561

PR #19517, opened 2021-02-18
app-text/asciidoc: switch to more stable distfiles

marcecj:
Switch away from the automatically generated source archives to those
provided by upstream.  This should prevent any future checksum errors.

Fixes: https://bugs.gentoo.org/770841
RepoMan-Options: --force


marcecj:
Is this fine, or should I limit this PR to updating the manifest hashes to the changed tarballs and switch distfile sources at the next version bump?

thesamesam:
Given the hash has changed, we need to [temporarily](https://devmanual.gentoo.org/general-concepts/mirrors/) use a new filename on the mirrors, so e.g. -> ${P}.new.tar.gz.

(See the Replacing Automatically Mirrored Files section).

marcecj:
> Given the hash has changed, we need to [temporarily](https://devmanual.gentoo.org/general-concepts/mirrors/) use a new filename on the mirrors, so e.g. -> ${P}.new.tar.gz.
> 
> (See the Replacing Automatically Mirrored Files section).

Should I create a separate commit to rename the files back (which could be merged at a later point in time), or do I not need to worry about that?

juippis (commented):
As far as I can see, there's no need to revbump these ebuilds. And to answer your question, you don't have to "move back" to regular file names, just update them on version bumps.

thesamesam:
I'll keep the revbump for safety even though it's not strictly needed because git archives shouldn't necessarily be trusted.

PR #19537, opened 2021-02-19
dev-java/myfaces-api:  bump to 2.2.14 (CVE-2021-26296)

vaukai:
Bug: https://bugs.gentoo.org/771543


fordfrog:
thanks! just please note it's a good habit to lift up the jdk/jre version to the lowest supported on gentoo to prevent issues with future jdk releases. for example, jdk 11 now reports this for packages with dep on >=1.6:

```
warning: [options] source value 6 is obsolete and will be removed in a future release
warning: [options] target value 1.6 is obsolete and will be removed in a future release
```
iirc lower versions are not supported anymore.

PR #19544, opened 2021-02-19
games-util/lutris: Add dev-python/lxml and x11-libs/gdk-pixbuf[jpeg] to RDEPEND

SethDusek:
Hi, currently lutris-9999 breaks without dev-python/lxml, since it is now required. Furthermore, a gdk-pixbuf with jpeg is required to load thumbnails. Lutris errors otherwise and displays garbage/undefined image outputs. This change adds both lxml and requires gdk-pixbuf be built with JPEG support to the lutris-9999 ebuild

### DIFF
--- a/games-util/lutris/lutris-9999.ebuild
+++ b/games-util/lutris/lutris-9999.ebuild
@@ -45,6 +45,7 @@ RDEPEND="
 		dev-python/python-magic[${PYTHON_USEDEP}]
 		dev-python/pyyaml[${PYTHON_USEDEP}]
 		dev-python/requests[${PYTHON_USEDEP}]
+		dev-python/lxml[${PYTHON_USEDEP}]
 	')
 	gnome-base/gnome-desktop:3[introspection]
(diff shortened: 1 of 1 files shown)


SethDusek:
Hello, How can I fix this "unnecessary files in manifest" problem? This is my first PR to Gentoo

thesamesam:
Hi!

For the Gentoo repo, we use 'thin-manifests' in the metadata folder of the repository. Your local overlay probably has this off which means it includes unnecessary files like EBUILD, MISC, and so on whose integrity is already guaranteed by git.

So, you need to adjust thin-manifests to off in your overlay, then run 'repoman manifest' again and git push -f here.

Come to #gentoo-proxy-maint on Freenode (IRC) if you need any help too!

SethDusek:
Alright, should be fixed up now. Thanks for the guidance

thesamesam:
Done, thank you! (sorted deps in f157b4b8764)

PR #19561, opened 2021-02-20
sys-apps/ethq: wrt 770943 and cleanup

SpiderX:


juippis (commented):
If you are just fixing -Werror build issue in 0.6.1 then please don't revbump it. People who managed to build it before don't need to rebuild it for this.

 Rest looks good...

PR #19565, opened 2021-02-20
dev-libs/quazip: Move version 1.1 from SLOT=1 to SLOT=0/1, fix all revdeps

a17r:
Upstream QuaZip migration guide: https://github.com/stachenov/quazip/blob/master/QuaZip-1.x-migration.md

They will tell us that they really really want us to install it in parallel to 0.x but in reality there is no sense doing it just for the fixed build system and changed include/lib paths.

See also https://github.com/stachenov/quazip/issues/92#issuecomment-706573280, ABI is not the problem.

Maintainers, please ACK.


chewi:
A bit more context behind the change would be nice. I take it upstream have a new pkg-config name? That aside, the rbutil and chessx changes look good. Thanks for also fixing an unrelated rbutil bug!

chewi (approved):

a17r:
> A bit more context behind the change would be nice. 

Updated top post, thanks for ack!

net-im/zoom/zoom-5.5.7011.0206-r2.ebuild
```diff
@@ -91,7 +91,13 @@ src_install() {
...
 	dosym8 -r {"/usr/$(get_libdir)",/opt/zoom}/libmpg123.so
-	dosym8 -r {"/usr/$(get_libdir)",/opt/zoom}/libquazip.so
+
+	local quazip_so="libquazip1-qt5.so"
+	if has_version "<dev-libs/quazip-1.0"; then
+		quazip_so="libquazip5.so"
+	fi
+	dosym8 -r "/usr/$(get_libdir)/${quazip_so}" /opt/zoom/libquazip.so
+
 	dosym8 -r "/usr/$(get_libdir)/libfdk-aac.so.2" /opt/zoom/libfdkaac2.so
```
ulm, line 101: Minor nitpick: Keep the unconditional links together, then the conditional ones.
```suggestion
	dosym8 -r "/usr/$(get_libdir)/libfdk-aac.so.2" /opt/zoom/libfdkaac2.so

	local quazip_so="libquazip1-qt5.so"
	if has_version "<dev-libs/quazip-1.0"; then
		quazip_so="libquazip5.so"
	fi
	dosym8 -r "/usr/$(get_libdir)/${quazip_so}" /opt/zoom/libquazip.so
```

The dependency should be updated to `dev-libs/quazip:0=` I guess.

Also, the change needs to be applied to zoom-5.4.53391.1108-r1.ebuild because that's the latest available version for x86.

ConiKost (approved):
lgtm. Thanks!

net-im/zoom/zoom-5.5.7011.0206-r2.ebuild
a17r: thanks for the feedback.

ulm: Now you've applied the changes to 5.5.6981.0202 which isn't strictly necessary (removal candidate after next release).
We need them for zoom-5.4.53391.1108-r1.ebuild which is the latest (and only) x86 version.

Polynomial-C (approved):

ulm (approved):

thesamesam (approved):

PR #19566, opened 2021-02-20
media-libs/opencv: drop dependency on virtual/lapacke

epsilon-0:


rossbridger:
I haven't gotten a chance a to test this PR myself yet, but did you find it successful compiling opencv with lapack USE flag on any machines whence lapacke.h doesn't exist?

epsilon-0:
> I haven't gotten a chance to test this PR myself yet, but did you find it successful compiling opencv with lapack USE flag on any machines whence lapacke.h didn't exist?

Yes, I have managed to get it to compile.
The lapacke.h workaround was only needed on old debian where there was a cmake file error. It is not needed on gentoo.

PR #19567, opened 2021-02-20
media-plugins/vdr-* plugins from Rolf Ahrenberg: update SRC_URI and EAPI

madmartin:
Rolf Ahrenberg moved his packages to github
inplace changes of SRC_URI and EAPI if necessary
without version bump

media-plugins/vdr-skinsoppalusikka
media-plugins/vdr-iptv
media-plugins/vdr-femon


juippis (commented):
Hey, 

not comfortable enough to do an EAPI bump with stable keywords. Please revbump and reset KEYWORDS to ~ where applicable, in your first commit, no EAPI was touched.

madmartin:
> Hey,
> 
> not comfortable enough to do an EAPI bump with stable keywords. Please revbump and reset KEYWORDS to ~ where applicable, in your first commit, no EAPI was touched.

Hello @juippis ,
I have changed all ebuilds with EAPI update to -r1 versions with ~ keywords.

juippis (commented):
Do any of these compile with gcc-11 for you?

\*sigh\*.

madmartin:
> Do any of these compile with gcc-11 for you?
> 
> *sigh*.

Unfortunately not. Its the same min -> std::min and max->std::max  thing like in media-video/vdr itself.
Do you want to point me to add gcc-11 patches here?

juippis:
If you have those available, then yes please! As these'd go to ~ and gcc-11 is ~, it's essentially broken.

And if there aren't those patches available, then, \*sigh\* I guess I'll just close my eyes and let's commit these EAPI bumps straight to stable. I'll be mad to receive any bug report related to that though...

madmartin:
> If you have those available, then yes please! As these'd go to ~ and gcc-11 is ~, it's essentially broken.

I dont have patches available, but it should not be too complicated to fix. Like on  vdr itself, it is the different behaviour of gcc-11 about https://en.cppreference.com/w/cpp/algorithm/min and max - instead of just write min(x,y) you now have to write std::min(x,y).
So please hold another few days, I will test&fix and notify upstream. On my other PRs, I wrote the gcc-11 state into comments.

juippis (approved):
Brilliant, thanks a lot! I had some emerge errors because vdr-2.2.0 and 2.4.0 mixed, but manually I got everything to work. This project's a mess...

madmartin:
> Brilliant, thanks a lot! I had some emerge errors because vdr-2.2.0 and 2.4.0 mixed, but manually I got everything to work. This project's a mess...

@juippis If you have time I'd like to hear your meaning of options to improve the vdr project. Yes, we have two main versions - 2.2 and 2.4 (difference is: vdr-2.4 supports HD-TV, vdr-2.2 only SD-TV, some people like to stay on 2.2 because there are more plugins supported)
How to divide Plugins and their versions to the two vdr versions? Slotting is no real option, putting higest versions in DEPEND is fobidden (would force downgrades but satisfies the user).
Extra USE flags like "vdr-22" and "vdr-24" and combining this with the depends? 
And then there is finally the personal coding style from Klaus .- the VDR author - he has some ugly habits like this templating of standard functions, naming c++ source files ".c" instead ".cpp"  which is criticized by clang. And his VDR project is not a co-working project, he is the "highlander".

juippis:
The simplest solution coming to my mind right now is subslotting vdr. But only if the plugins are **really** incompatible. 

But in general, the packages (at least vdr) could use some old version removal, general cleanup here and there.

PR #19575, opened 2021-02-21
sci-libs/opencascade: restrict to <vtk-9

waebbl:
This version is not compatible with vtk-9. Because I got
several notifications on my overlay and via email, to make
this clear, the package needs to be restricted to <vtk-9,
although it's not yet available in ::gentoo.


waebbl:
I'm gonna handle this together with an upcoming fix for bug https://bugs.gentoo.org/771300

PR #19584, opened 2021-02-21
dev-util/vulkan-headers: bump to 1.2.170

myrvogna:
Signed-off-by: Octiabrina Terrien-Puig octiabrina@myrvogna.net


dev-util/vulkan-headers/vulkan-headers-1.2.170.ebuild
```diff
@@ -0,0 +1,24 @@
+# Copyright 1999-2021 Gentoo Authors
```
sarnex, line 1: We need to bump everything in the stack at once, including
glslang
spirv-headers
spirv-tools
vulkan-headers
vulkan-tools
vulkan-loader
vulkan-layers
shaderc

sarnex (changes requested):

dev-util/vulkan-headers/vulkan-headers-1.2.170.ebuild
sarnex: If you're not interested in bumping the whole stack feel free to drop the PR

myrvogna: The vulkan-* packages have the same versions to be bumped to, that I can figure out on my own. Is there anything I need to know about the others packages, or would bumping them to their latest upstream version be sufficient?

sarnex: https://vulkan.lunarg.com/sdk/config/latest/linux/config.json

The vulkan team made that API for us to know what versions of everything we need to use. They're still on 1.6.162 so we might want to wait until they list 1.6.170. For everything else, you can just snapshot if it there's not a new release

myrvogna: Ah, I was not yet aware of the technicalities of Vulkan packaging and got overeager. With that in mind, I agree that waiting until they list a higher version sounds like the better choice here.

I shall drop the PR, thank you for your time and the information provided.

sarnex: Thanks for the interest! There might be some bugs to address if you still want to contribute

PR #19588, opened 2021-02-22
app-misc/beep: capabilities

Alessandro-Barbieri:

### DIFF
--- a/app-misc/beep/beep-1.4.9-r1.ebuild
+++ b/app-misc/beep/beep-1.4.9-r1.ebuild
@@ -1,0 +1,57 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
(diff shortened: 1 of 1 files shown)


Alessandro-Barbieri:
ping

mgorny:
https://bugs.gentoo.org/716734 pretty correctly says `USE=suid` doesn't work anymore, so at the very least you can remove that.

It works just fine for me, though. I don't recall if I set some special rules for it but I doubt it. Is there really a need to play with caps these days?

mgorny (approved):
Thanks for this, and I'm sorry that it took this long to merge it. Please remember to leave a comment after pushing changes as we otherwise find it hard to notice that something has happened.

ionenwks:
Note that if USE=-filecaps, this will fallback to installing beep suid root without the former :audio group limitation.

PR #19593, opened 2021-02-22
app-text/xmlto: EAPI 7

vaukai:
Closes: https://bugs.gentoo.org/757576

Adding www-client/links as an alternative to virtual/w3m
Closes  https://bugs.gentoo.org/726156

### DIFF
--- a/app-text/xmlto/xmlto-0.0.28-r4.ebuild
+++ b/app-text/xmlto/xmlto-0.0.28-r4.ebuild
@@ -1,0 +1,42 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DESCRIPTION="script for converting XML and DocBook documents to a variety of output formats"
+HOMEPAGE="https://pagure.io/xmlto"
+SRC_URI="https://releases.pagure.org/${PN}/${P}.tar.bz2"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~sparc-solaris ~sparc64-solaris ~x64-solaris ~x86-solaris"
+IUSE="latex text"
+
+RDEPEND="app-text/docbook-xsl-stylesheets
+	app-text/docbook-xml-dtd:4.2
+	dev-libs/libxslt
+	|| ( sys-apps/util-linux app-misc/getopt )
+	text? ( || ( virtual/w3m www-client/elinks www-client/links www-client/lynx ) )
+	latex? ( dev-texlive/texlive-formatsextra )"
+# We only depend on flex when we patch the input lexer.
+DEPEND="${RDEPEND}"
+
+DOCS="AUTHORS ChangeLog FAQ NEWS README THANKS"
+
+PATCHES=( "${FILESDIR}"/${PN}-0.0.22-format_fo_passivetex_check.patch )
+
+src_prepare() {
+	default
+
+	# fix symbol clash on Solaris
+	if [[ ${CHOST} == *-solaris* ]] ; then
+		sed -i -e 's/\(attrib\|val\)/XMLTO\1/g' xmlif/xmlif.l || die
+	fi
+}
+
+src_configure() {
+	# We don't want the script to detect /bin/sh if it is bash.
+	export ac_cv_path_BASH=${BASH}
+	has_version sys-apps/util-linux || export GETOPT=getopt-long
+	econf
+}


app-text/xmlto/xmlto-0.0.28-r4.ebuild
```diff
@@ -0,0 +1,47 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+inherit eutils
```
thesamesam, line 5: New line after EAPI

thesamesam, line 5: Also, is eutils needed?

app-text/xmlto/xmlto-0.0.28-r4.ebuild
```diff
@@ -0,0 +1,47 @@
...
+	app-text/docbook-xml-dtd:4.2
+	app-shells/bash:0
+	dev-libs/libxslt
+	sys-apps/sed
+	>=sys-apps/coreutils-6.10-r1
+	|| ( sys-apps/util-linux app-misc/getopt )
+	sys-apps/which
+	text? ( || ( virtual/w3m www-client/elinks www-client/links www-client/lynx ) )
+	latex? ( dev-texlive/texlive-formatsextra )"
+# We only depend on flex when we patch the imput lexer.
```
thesamesam, line 26:
```suggestion
# We only depend on flex when we patch the input lexer.
```

app-text/xmlto/xmlto-0.0.28-r4.ebuild
```diff
@@ -0,0 +1,47 @@
...
+DEPEND="${RDEPEND}"
```
thesamesam, line 27: A lot/all of these are probably actually BDEPEND?

app-text/xmlto/xmlto-0.0.28-r4.ebuild
vaukai: Adjusted

app-text/xmlto/xmlto-0.0.28-r4.ebuild
vaukai: Thanks for pointing me here. Line removed.

vaukai: No longer inheriting

app-text/xmlto/xmlto-0.0.28-r4.ebuild
vaukai: Yes.

thesamesam: Are we changing it then? :)

ionenwks: For BDEPEND, from a quick strace and outside base set (e.g. excuding sed/which), flex, m4, util-linux, and pkgconfig were called regardless of USE.

Not that I really checked this package in-depth, but it been bothering me for a while and glad someone is looking at it.

ionenwks: Actually scratch pkg-config, that one's not needed.

vaukai: Is there another way (except searching the source code) for checking the BDEPENDs?  I am somewhat lost here

ionenwks: I use strace to find them, but it's hard to tell if really needed to be honest. It seems to build fine without flex and m4 but it calls them if they exist, unsure how this affects things.

ionenwks: Needs more testing but not convinced even really need BDEPEND for this package after all. Doesn't need all those DEPEND either way given those are for just what the runtime script uses (and exist in base set for many).

vaukai: When running
ebuild xmlto-0.0.28-r4.ebuild clean configure | grep 'checking for'
would all that output be BDEPEND?

ionenwks: Not all of them are actually ran, it just checks if they exist.. which I guess mean they may need to be in normal DEPEND to be detected :| But again, most of those are base set and sed/bash/which seem weird to define.

vaukai: > Are we changing it then? (i.e. do not define DEPEND or define it to be a small subset?)

Sorry here.  I am still lost with the {B,R,}dependencies stuff.  Only removed some Implicit System Dependencies for now according to [devmanual](https://devmanual.gentoo.org/general-concepts/dependencies/#implicit-system-dependency).
Need to find a reliable method to distinguish the runtime and build dependencies.

thesamesam: We actually should keep @system stuff but no need for sed/Bash/anything specified by [PMS](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-11800012.1.1). I'll make a note to update the devmanual to be more explicit.

Having @system elements in there means we can parallelise emerges better.

thesamesam (approved):
Thanks! Minor style changes made in b3a2f4f792b6348909d636452d35ca1ac099de0a.

PR #19594, opened 2021-02-22
Stalonetray pm bump

jkroonza:


x11-misc/stalonetray/metadata.xml
```diff
@@ -1,7 +1,14 @@
...
+	<maintainer type="person">
+		<email>jaco@uls.co.za</email>
+		<name>Jaco Kroon</name>
+	</maintainer>
+	<maintainer type="project">
+		<email>proxy-maint@gentoo.org</email>
+		<name>Proxy Maintainers</name>
+	</maintainer>
 	<use>
 		<flag name="graceful-exit">use non-portable hack to exit gracefuly on signal</flag>
```
thesamesam, line 13:
```suggestion
		<flag name="graceful-exit">use non-portable hack to exit gracefully on signal</flag>
```

x11-misc/stalonetray/stalonetray-0.8.4.ebuild
```diff
@@ -0,0 +1,47 @@
...
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	econf $(use_enable debug) \
+		$(use_enable graceful-exit) \
+		$(use_enable kde native-kde)
```
thesamesam, line 35: Interesting - does it need any extra deps?

x11-misc/stalonetray/stalonetray-0.8.4.ebuild
```diff
@@ -0,0 +1,47 @@
...
+KEYWORDS="~amd64 ~x86"
+IUSE="debug +graceful-exit kde"
+
+RDEPEND="x11-libs/libX11
+	x11-libs/libICE
+	x11-libs/libSM
+	x11-libs/libXpm"
+DEPEND="${RDEPEND}
+	x11-base/xorg-proto"
+BDEPEND="${DEPEND}
```
thesamesam, line 23: Definitely need to include ${DEPEND} HERE?

x11-misc/stalonetray/stalonetray-0.8.4.ebuild
```diff
@@ -0,0 +1,46 @@
...
+src_compile() {
+	emake
+	emake dist
+}
+
+src_install() {
+	emake DESTDIR="${D}" install
+
+	dodoc AUTHORS BUGS COPYING NEWS README.md TODO stalonetrayrc.sample stalonetray.html
+}
```
juippis, line 46: I'd prefer if you defined global `DOCS` array, and you could just remove `src_install` altogether.

x11-misc/stalonetray/stalonetray-0.8.4.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	econf $(use_enable debug) \
+		$(use_enable graceful-exit) \
+		--enable-native-kde
```
juippis, line 34: Can you please tell me what this does, and why apparently no KDE deps are needed?
I have a feeling it's got something to do with appindicator support? Does this then break support for older xembed system trays, ie, should this be made optional?

x11-misc/stalonetray/stalonetray-0.8.4.ebuild
jkroonza: Done thanks.

x11-misc/stalonetray/stalonetray-0.8.4.ebuild
jkroonza: Oddly enough, it's some KDE integration (I don't use KDE so can't really tell you what it does).  There was a USE flag on this, but Sam indicated to just enable it unconditionally since it doesn't actually require additional dependencies.  I'm running as per this on evilwm and can't see that anything broke ... so doesn't look like it needs to be optional.

jkroonza: Looking at the code it looks like it allows stalonetray to be used as the tray manager for KDE, so instead of running the native KDE tray you can run stalonetray.  I'm afraid that's as much as I can figure out from looking at the source code.  It enables additional code, but doesn't cause *alternative* paths to be followed as far as I can tell.

juippis: Thanks, just wanted to confirm it doesn't break for people not either on appindicator-supported or xembed-supported environments (as that's the reason to use stalonetray in the first place)

PR #19595, opened 2021-02-22
dev-util/packer: bump version to 1.7.0

xp-vit:


dev-util/packer/packer-1.7.0.ebuild
```diff
@@ -9,14 +9,17 @@ DESCRIPTION="A tool to create identical machine images for multiple platforms"
...
 SLOT="0"
 KEYWORDS="~amd64"
 
 DOCS=( {README,CHANGELOG}.md )
 
+RESTRICT+=" test"
+
 src_compile() {
 	go build \
+		-ldflags "${go_ldflags}" \
```
juippis, line 22: Where does this variable come from?

juippis (commented):
Please split adding new version and removing old ones to separate commits.

PR #19597, opened 2021-02-22
app-text/jabref-bin: explicitly mark as a live ebuild with live property

thesamesam:
All of our eclasses for live ebuilds set PROPERTIES="live" as per PMS
to indicate their live nature, but manual live ebuilds obviously
don't have this automatic property.

Let's set it manually to correctly identify this as a live ebuild.


nicolasbock (approved):
Thanks!

PR #19602, opened 2021-02-22
Revert "x11-misc/rss-glx: Remove old"

miller-alex:
The commit removed the stable revision for no good reason.


SoapGentoo:
It did remove it for a good reason: we've changed a lot, and we'll start from scratch. See https://github.com/gentoo/gentoo/pull/19511#issuecomment-783565199

PR #19606, opened 2021-02-22
dev-lang/zig: version bump to 0.7.1

epsilon-0:
also readds `shared` flag to `sys-devel/lld` as zig needs lld libraries.


juippis (approved):
I've just finished build-testing every ebuild in this PR and they work. I tried calling `zig` and it printed the help message, and also compiled `firefox` with the modified `lld`. It worked. So to me, it looks like this works, but I'm not knowledgeable enough to say for sure.

https://dev.gentoo.org/~juippis/logs/zig-gh-pr-19606.txt

mgorny (approved):
Feel free to merge.

juippis:
And if needed, we can mask the new versions and run a tinderbox through revdeps.

dev-lang/zig/zig-0.7.1.ebuild
```diff
@@ -0,0 +1,52 @@
...
+)
+ALL_LLVM_TARGETS=( "${ALL_LLVM_TARGETS[@]/#/llvm_targets_}" )
+LLVM_TARGET_USEDEPS="${ALL_LLVM_TARGETS[@]}"
+
+LLVM_MAX_SLOT=11
+
+RDEPEND="
+	sys-devel/clang:${LLVM_MAX_SLOT}
+	>=sys-devel/lld-11.0.0
+	<sys-devel/lld-12.0.0
```
mjsir911, line 36: Not to be super annoying, but this doesn't actually work. A valid package (and the one resolving for me!) is `sys-devel/lld-11.0.1`, which hasn't been touched my this PR. Would a solution be to just backport the fix to existing ebuilds? unmasking the new revisions? adding a `shared` flag to existing ebuilds? I can file something if you give me a good idea and you want @mgorny

mjsir911: See: https://bugs.gentoo.org/778398

PR #19607, opened 2021-02-22
dev-python/pychromecast: Mark setuptools as BDEPEND

jsmolic:
`DISTUTILS_USE_SETUPTOOLS=bdepend` is declared by default in `distutils-r1`, so no need to specify it manually


jsmolic:
Pushed as https://github.com/gentoo/gentoo/commit/6b2fd11867be39a2612315c9ae59829b70e0c098

PR #19610, opened 2021-02-23
media-sound/moc: Add sndio support

strikersh:

### DIFF
--- a/media-sound/moc/metadata.xml
+++ b/media-sound/moc/metadata.xml
@@ -8,6 +8,7 @@
 	<use>
 		<flag name="cache">Enable tags caching support</flag>
 		<flag name="sid">Build with SID (Commodore 64 Audio) support</flag>
+		<flag name="sndio">Enable support for the <pkg>media-sound/sndio</pkg> backend</flag>
 		<flag name="tremor">Build vorbis support using an integer implementation
 			of the vorbis library (<pkg>media-libs/tremor</pkg>)</flag>
 	</use>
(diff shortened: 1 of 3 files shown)


fordfrog:
you should keep alphabetical order at least in the use flags and the conditional deps. and there is an issue with the dependency.

strikersh:
okay, I changed the order.
I don't understand the issue with `media-sound/sndio`
it is part of `::gentoo` https://packages.gentoo.org/packages/media-sound/sndio

fordfrog:
@strikersh i did some testing and noticed that the problem is in the fact that you are modifying stable ebuilds and depending on an unstable package. that's the cause of the issues. basically, before you commit anything, you should run `repoman full` and `pkgcheck scan` in the package directory to verify none of these programs has any complaints. there are two solutions to this issue. either do a revbump instead of modifying the stable ebuilds, or first ask for stabilization of sndio and then this pr would work. or you can do a revbump, and in the meantime you can ask for sndio stabilization and once sndio is stable we can stabilize your revbumps.

fordfrog:
@strikersh i did some testing and noticed that the problem is in the fact that you are modifying stable ebuilds and depending on an unstable package. that's the cause of the issues. basically, before you commit anything, you should run `repoman full` and `pkgcheck scan` in the package directory to verify none of these programs has any complaints. there are two solutions to this issue. either do a revbump instead of modifying the stable ebuilds, or first ask for stabilization of sndio and then this pr would work. or you can do a revbump, and in the meantime you can ask for sndio stabilization and once sndio is stable we can stabilize your revbumps.

fordfrog:
@strikersh thanks!

PR #19613, opened 2021-02-23
media-video/mplayer: Add sndio support

strikersh:


profiles/arch/alpha/package.use.mask
```diff
@@ -11,6 +11,7 @@ sci-libs/umfpack cholmod
 media-libs/openal sndio
 media-video/ffmpeg sndio
 media-sound/moc sndio
+media-video/mplayer sndio
```
lanodan, line 14: You don't need to add mplayer in this case, mplayer-9999 doesn't have ~alpha in it's keywords.

PR #19616, opened 2021-02-23
sci-biology/bedtools: bump to 2.30.0

ezzieyguywuf:
The samtools patch is no longer needed, as upstream has fixed this. Test
suite passes without it now.


ezzieyguywuf:
See https://bugs.gentoo.org/696188

ezzieyguywuf:
Is the CI failing solely b/c of the DuplicateFiles warning?

If so, how should I resolve this?

1. rename patch to newer version, and update older ebuild?
2. reference older versioned patch in this new ebuild?
3. remove version from patch filename?

SoapGentoo:
@ezzieyguywuf yup, why duplicate the patch? Just refer to the patch using `${PN}`

ezzieyguywuf:
> 
> 
> @ezzieyguywuf yup, why duplicate the patch? Just refer to the patch using `${PN}`

@SoapGentoo I was worried that referring to `${PN}-2.29.2...` in `bedtools-2.30.0.ebuild` would be misleading/"wrong"

Is this a non-issue?

SoapGentoo:
@ezzieyguywuf yup, complete non-issue, so much that it's encouraged if it avoids duplicating patches. If you fix this, I'll merge it.

ezzieyguywuf:
> 
> 
> @ezzieyguywuf yup, complete non-issue, so much that it's encouraged if it avoids duplicating patches. If you fix this, I'll merge it.

already fixed before your comment even posted :)

PR #19620, opened 2021-02-23
media-fonts/polarsys-b612-fonts: Update license

janhenke:
Upstream did not react yet. So add the other licenses too.
While at it, bump to EAPI 7.

Bug: https://bugs.gentoo.org/746725

### DIFF
--- a/media-fonts/polarsys-b612-fonts/polarsys-b612-fonts-1.003.ebuild
+++ b/media-fonts/polarsys-b612-fonts/polarsys-b612-fonts-1.003.ebuild
@@ -1,7 +1,7 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
 
 inherit font vcs-snapshot
 
@@ -9,7 +9,7 @@ DESCRIPTION="Font designed for aircraft cockpit displays"
 HOMEPAGE="https://b612-font.com/"
 SRC_URI="https://git.polarsys.org/c/b612/b612.git/snapshot/b612-bd14fde2544566e620eab106eb8d6f2b7fb1347e.tar.bz2 -> ${P}.tar.bz2"
 
-LICENSE="EPL-1.0"
+LICENSE="EPL-1.0 BSD OFL-1.1"   # to be clarified #746725
 SLOT="0"
 KEYWORDS="amd64 arm arm64 ppc ppc64 x86"
 IUSE=""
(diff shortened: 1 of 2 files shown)


media-fonts/polarsys-b612-fonts/polarsys-b612-fonts-1.003.ebuild
```diff
@@ -1,15 +1,15 @@
...
+EAPI=7
 
 inherit font vcs-snapshot
 
 DESCRIPTION="Font designed for aircraft cockpit displays"
 HOMEPAGE="https://b612-font.com/"
 SRC_URI="https://git.polarsys.org/c/b612/b612.git/snapshot/b612-bd14fde2544566e620eab106eb8d6f2b7fb1347e.tar.bz2 -> ${P}.tar.bz2"
 
-LICENSE="EPL-1.0"
+LICENSE="EPL-1.0 BSD OFL-1.1"   # to be clarified #746725
```
luke-jr, line 12: Should be wrapped in "|| ( ... )"

janhenke: In the bug, ulm suggested to not wrap it. I would like to stick to his suggestion.

media-fonts/polarsys-b612-fonts/polarsys-b612-fonts-1.008.ebuild
```diff
@@ -1,15 +1,15 @@
...
 
 DESCRIPTION="Font designed for aircraft cockpit displays"
 HOMEPAGE="https://b612-font.com/"
 SRC_URI="https://github.com/polarsys/b612/archive/${PV}.tar.gz -> ${P}.tar.gz"
 
-LICENSE="OFL-1.1"
+LICENSE="EPL-1.0 BSD OFL-1.1"   # to be clarified #746725
 SLOT="0"
 KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~x86"
 IUSE=""
```
ulm, line 15: `IUSE` is redundant and could be dropped. Otherwise LGTM.

@luke-jr: I'm aware that this may not be the last word about the license, but we need some interim solution. In any case, the "and" leans on the safe side.

ulm (approved):

media-fonts/polarsys-b612-fonts/polarsys-b612-fonts-1.003.ebuild
luke-jr: Sounded to me like he doesn't care either way.

Wrapping it is correct, and I for one do actually care because OFL really isn't a free license.

media-fonts/polarsys-b612-fonts/polarsys-b612-fonts-1.008.ebuild
luke-jr: If there was any real question, I would agree - but the intent is unambiguously clear IMO.

media-fonts/polarsys-b612-fonts/polarsys-b612-fonts-1.003.ebuild
ulm: > [...] OFL really isn't a free license.

Neither the [FSF](https://www.gnu.org/licenses/license-list.html.en#SILOFL) nor the [OSI](https://opensource.org/licenses/OFL-1.1) share that opinion.

PR #19621, opened 2021-02-23
x11-misc/*: adopt + cleanup five small packages

ionenwks:
Notes:
- numlockx was EAPI-7 bumped in stable and would rather not revbump to ~arch (but will if asked)
- wmctrl also received changes in stable but very trivial (hardly needed attention, but picking up since using in some old scripts).
- rest already had ~arch versions, so only changed those


x11-misc/dzen/dzen-0.9.5_p2.ebuild
```diff
@@ -0,0 +1,67 @@
...
+	emake -C gadgets \
+		CC="$(tc-getCC)" CFLAGS="${cflags}" \
+		LDFLAGS="${LDFLAGS}" X11LIBS="${libs}"
+
+	local flag
+	for flag in xft xinerama xpm; do
+		# xft always-enabled wrt bug #477656
+		[[ ${flag} == xft ]] || use ${flag} || continue
+		cflags+=" $($(tc-getPKG_CONFIG) --cflags ${flag}) -DDZEN_${flag^^}"
+		libs+=" $($(tc-getPKG_CONFIG) --libs ${flag})"
```
SoapGentoo, line 53: let's say `USE="-xinerama"`, this will still unconditionally try to add the libs from `xinerama.pc`?

SoapGentoo: scratch that, I for the logic wrong. Could you maybe consider de-tersifying the continue?
```
if [[ ${flag} == xft ]] || use ${flag}; then
    continue
fi

File: x11-misc/dzen/dzen-0.9.5_p2.ebuild [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+	done
```
SoapGentoo, line 54:
```
	local flag
	# xft always-enabled wrt bug #477656
	for flag in xft $(usev xinerama) $(usev xpm); do
		cflags+=" $($(tc-getPKG_CONFIG) --cflags ${flag}) -DDZEN_${flag^^}"
		libs+=" $($(tc-getPKG_CONFIG) --libs ${flag})"
	done
```

ionenwks: Done, much better.

SoapGentoo (approved):

PR #19624, opened 2021-02-24
app-text/pdftk: version bump to 3.2.2

myrvogna:


fordfrog:
@myrvogna thank you for your contributions! please increase the min jdk/jre version to 1.8. it is the lowest version supported on gentoo and there might be issues with future java releases not supporting older versions. if you are interested, you can find more info at these links:
https://wiki.gentoo.org/wiki/Java_Developer_Guide
https://wiki.gentoo.org/wiki/Gentoo_Java_Packing_Policy

myrvogna:
Thank you for the information! I shall fix this as soon as possible.

fordfrog:
@myrvogna could you please merge (squash) the commits into one? it's better to do a bump in one commit rather than in two. (in your other contribution i did exactly that - i updated your bump, but that was because i did not want to bother you with the changes i wanted to do and did not want to modify your commit that you authored.) here's how to do it in case you don't know: https://stackoverflow.com/questions/14534397/squash-all-my-commits-into-one-for-github-pull-request

myrvogna:
I have yet to master all the intricacies of git, but I believe this should be fine now.?

fordfrog:
@myrvogna excellent! thanks! :-)

PR #19625, opened 2021-02-24
dev-java/bcprov: version bump to 1.68

myrvogna:


fordfrog:
@myrvogna thank you for the bump! please also note that i removed src_compile and src_install as they were redundant and the default work fine (i know this is how it is in 1.66).

PR #19628, opened 2021-02-24
media-libs/libavif: 0.9.0 version bump

novomesk:
ebuild use `svt-av1` use flag (depends on `media-libs/svt-av1` which is unavailable on x86).

added mask to profiles/arch/x86/package.use.mask


juippis (commented):
There is something weird in this version.
```
/usr/lib64/libavif.so -> libavif.so.10
/usr/lib64/libavif.so.10 -> libavif.so.10.0.0
/usr/lib64/libavif.so.10.0.0
/usr/lib64/libavif.so.6 -> libavif.so.6.0.0
/usr/lib64/libavif.so.6.0.0
```
Why does this install two libraries with different SONAMEs?

I'm also getting 
```
!!! existing preserved libs:
>>> package: media-libs/libavif-0.9.0
 *  - /usr/lib64/libavif.so.6
 *  - /usr/lib64/libavif.so.6.0.0
```
Which to me seems to indicate the symlinking `/usr/lib64/libavif.so.6 -> libavif.so.6.0.0` is not always done?

novomesk:
On my system it installs only:
```
libavif.so -> libavif.so.10
libavif.so.10 -> libavif.so.10.0.0
libavif.so.10.0.0
```

libavif.so.6 is from 0.8.1 version. You have some application which still use/need the old version with different API.

juippis:
Indeed, silly me, I tested against unmodified efl which still pulls `avif? ( ~media-libs/libavif-0.8.1 )`

Gonna do another testing round later today but it looks good.

juippis (approved):
Let's request keywording for x86. Can you do it?

novomesk:
> Let's request keywording for x86. Can you do it?

Sorry, keywording for x86 of which package/dependency?

juippis:
Nevermind, I see svt-av1 can't have x86 support for now.

PR #19629, opened 2021-02-24
app-editors/retext: Depend optionally on qtwebengine

holgersson32644:
To avoid breakage for existing installations the deps are not made optional
but moved into USE flags (bug 772197).

Also there is added a missing dbus dep (bug 772197, too).

Closes: https://bugs.gentoo.org/772197

### DIFF
--- a/app-editors/retext/retext-7.2.0-r1.ebuild
+++ b/app-editors/retext/retext-7.2.0-r1.ebuild
@@ -37,10 +37,10 @@ RDEPEND="
 	dev-python/markups[${PYTHON_USEDEP}]
 	dev-python/pygments[${PYTHON_USEDEP}]
 	dev-python/python-markdown-math[${PYTHON_USEDEP}]
-	dev-python/PyQt5[gui,network,printsupport,widgets,${PYTHON_USEDEP}]
-	dev-python/PyQtWebEngine[${PYTHON_USEDEP}]
+	dev-python/PyQt5[dbus,gui,printsupport,widgets,${PYTHON_USEDEP}]
 "
 DEPEND="${RDEPEND}"
+BDEPEND="test? ( dev-python/PyQt5[testlib,${PYTHON_USEDEP}] )"
 
 src_test() {
 	virtx distutils-r1_src_test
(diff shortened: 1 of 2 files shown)


app-editors/retext/retext-7.2.0-r1.ebuild
```diff
@@ -7,8 +7,10 @@ EAPI=7
 # old compat matrix is no longer accessible as of 2021-02-13 but stated back
 # in 2020-07-05 that PyQt5 was explicitly not supported.
 PYTHON_COMPAT=( python3_{8,9} )
+# See https://dev.gentoo.org/~mgorny/python-guide/distutils.html
+DISTUTILS_USE_SETUPTOOLS=bdepend
```
juippis, line 11: This is the default-defined value, it doesn't need to be put into the ebuild.

app-editors/retext/retext-7.2.0-r1.ebuild
```diff
@@ -28,17 +30,29 @@ fi
...
+	dev-python/PyQt5[dbus,gui,printsupport,widgets,${PYTHON_USEDEP}]
+	enchant? ( dev-python/pyenchant[${PYTHON_USEDEP}] )
+	markdown? (
+		dev-python/markdown[${PYTHON_USEDEP}]
+		dev-python/python-markdown-math[${PYTHON_USEDEP}]
+	)
+	test? (
+		dev-python/PyQt5[testlib,${PYTHON_USEDEP}]
+		x11-base/xorg-server[xvfb]
+	)
```
juippis, line 51: Test deps should be in `BDEPEND` and the `x11-base/xorg-server[xvfb]` is handled via *virtualx.eclass* that needs it.

juippis (commented):
```
To avoid breakage for existing installations the deps are not made optional
but moved into USE flags (bug 772197).
```
If they are **optional**, then there is no breakage. Just put them into `optfeature`.

app-editors/retext/retext-7.2.0-r1.ebuild
holgersson32644: fixed locally. Had the idea from https://bugs.gentoo.org/746071.

> ```
> To avoid breakage for existing installations the deps are not made optional
> but moved into USE flags (bug 772197).
> ```
> 
> If they are **optional**, then there is no breakage. Just put them into `optfeature`.
OK. I actually had to put markdown support back into the package unconditionally - sorry for that noise.

app-editors/retext/retext-7.2.0-r1.ebuild
holgersson32644: fixed locally

juippis (approved):

juippis:
```
app-editors/retext
  WhitespaceFound: version 7.2.0-r1: ebuild has trailing whitespace on line: 64
  WhitespaceFound: version 9999: ebuild has trailing whitespace on line: 64
```

holgersson32644:
> ```
> app-editors/retext
>   WhitespaceFound: version 7.2.0-r1: ebuild has trailing whitespace on line: 64
>   WhitespaceFound: version 9999: ebuild has trailing whitespace on line: 64
> ```
Yes, sorry, I missed these because of the shortly broken repoman-portage combination, which is fixed now. Won't happen with repoman next time.

PR #19630, opened 2021-02-24
app-crypt/libtpms: Bump to 0.8.2

salahcoronya:
The fact RSA keys are not as strong as they should be is disclosed in this version:

Note: The TPM 2 implementation returns 2048 bit keys with ~1984 bit
strength due to a bug in the TPM 2 key creation algo that cannot
easily be fixed. The bug is in RsaAjustPrimeCandidate, which is
called before the prime number check.

See https://github.com/stefanberger/libtpms/issues/183


thesamesam:
That's probably a candidate for a security bug. Could you file one?

salahcoronya:
Upstream has release version 0.8.2 with the following note: 

    CryptSym: fix AES output IV
    A CVE has been filed for this bugfix. Unfortunately multi-step encrypted
    data won't decrypt anymore but are now compatible with other TPM 2 devices.

dev-libs/libtpms/files/libtpms-0.8.0-Remove-WError.patch
```diff
@@ -0,0 +1,13 @@
...
++++ b/configure.ac
+@@ -304,7 +304,7 @@ if test "x$enable_hardening" != "xno"; then
+ 	AC_SUBST([HARDENING_LDFLAGS])
+ fi
+ 
+-CFLAGS="$CFLAGS $COVERAGE_CFLAGS -Wall -Werror -Wreturn-type -Wsign-compare -Wno-self-assign"
++CFLAGS="$CFLAGS $COVERAGE_CFLAGS -Wall -Wreturn-type -Wsign-compare -Wno-self-assign"
+ CFLAGS="$CFLAGS -Wmissing-prototypes"
+ LDFLAGS="$LDFLAGS $COVERAGE_LDFLAGS"
+
```
juippis: Any chance to get this upstreamed? It's annoying having to carry this kind of patch around version after version.

juippis (approved):

PR #19634, opened 2021-02-24
net-im/signal-desktop-bin: add new version 1.40.1

robert7k:
Closes: https://bugs.gentoo.org/766006


juippis (approved):
Everything seems to be in right condition now (wrt the postinst bug). Thanks!

PR #19639, opened 2021-02-25
Pum cleanup

NeddySeagoon:
Start cleaning up package.use.mask on arm64


thesamesam:
breakage fixed in https://github.com/gentoo/gentoo/commit/b522aef84d6df05c3ff7b1571a0bfc067fb39ca7

PR #19648, opened 2021-02-25
dev-util/distro-info: version bump to 1.0

nE0sIghT:
Closes: https://bugs.gentoo.org/771939


dev-util/distro-info/distro-info-1.0.ebuild
```diff
@@ -0,0 +1,94 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8} )
```
thesamesam, line 6: No Python 3.9?

dev-util/distro-info/distro-info-1.0.ebuild
```diff
@@ -0,0 +1,94 @@
...
+
+REQUIRED_USE="python? ( ${PYTHON_REQUIRED_USE} )"
+
+COMMON_DEPEND="dev-lang/perl:=
+	python? ( ${PYTHON_DEPS} )"
+DEPEND="${COMMON_DEPEND}
+	python? ( dev-python/setuptools[${PYTHON_USEDEP}] )
+	test? (
+		dev-util/shunit2
+		dev-python/pylint[${PYTHON_USEDEP}]
```
thesamesam, line 29: Can we drop the pylint dep? It's not really useful for users

nE0sIghT: I can't get your idea. How we can (and why we should) drop test-time dependency?

dev-util/distro-info/distro-info-1.0.ebuild
nE0sIghT: No Python 3.9 in pylint

juippis (approved):

dev-util/distro-info/distro-info-1.0.ebuild
thesamesam: It is [now](0e51dbc4000a3aa4dddb84ec3025640d22cd8f32)!

dev-util/distro-info/distro-info-1.0.ebuild
thesamesam: If it's possible, dropping pylint from the test suite is helpful. We don't really care if formatting standards aren't met -- that's an upstream problem and doesn't indicate if the application works as expected.

pylint is an unnecessary dependency for our case - testing if the application works properly. It's similar to [this](https://dev.gentoo.org/~mgorny/python-guide/pytest.html?highlight=tests#avoiding-dependencies-on-other-pytest-plugins) where coverage tests aren't helpful for end-users.

It's not a big deal, just a suggestion.

juippis: @thesamesam could you please open a bug for it? I'm all for removing useless deps if it's possible.

PR #19650, opened 2021-02-25
net-fs/samba: new USE flags spotlight, regedit, glusterfs, ntvfs

dongsupark:
Introduce a USE flag `spotlight`, to be able to disable the spotlight backend by default, as it is not needed by Linux.

Introduce a USE flag `rededit`, to be able to disable the rededit tool if needed.

Introduce a USE flag `glusterfs`, to be able to disable the glusterfs by default.

Introduce a USE flag `ntvfs`, to be able to disable the ntvfs-fileserver by default.

Since the docbook-xsl-stylesheets and libxslt are needed only at build time, we should move those deps to BDEPEND.

Pulls in `dev-libs/icu` only if spotlight is enabled.


thesamesam:
@Polynomial-C

PR #19653, opened 2021-02-25
net-analyzer/zabbix: added agent2 useflag

VTimofeenko:
Implements the useflag to build zabbix-agent2, a go-based Zabbix agent.

The agent2 service files provided in files/ directory were tested on
openrc and systemd-based systems.

Cleaned up some minor formatting in the original ebuild.

Closes: https://bugs.gentoo.org/749351


fordfrog:
@VTimofeenko sorry, it took me some time to get to this. could you please make downloading of the agent2 related stuff conditional based on whether the use flag is on? the list is pretty long and there is no point to wait for all the stuff to get downloaded if one does not have agent2 use flag enabled. and also, could you please also port agent2 support to 5.0? it's an lts version so i think it makes sense. thanks!

VTimofeenko:
@fordfrog,
> could you please make downloading of the agent2 related stuff conditional based on whether the use flag is on? the list is pretty long and there is no point to wait for all the stuff to get downloaded if one does not have agent2 use flag enabled

Makes perfect sense, I'll change this.

> could you please also port agent2 support to 5.0? it's an lts version so i think it makes sense. thanks!

Definitely. A couple of questions: 
* Should I add -r1 to 5.0.8, 5.0.9 or both? 
* Looking at [ebuild revisions devmanual section](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/), I think I should add -r1 with agent2 alongside the 5.0 current versions. Is that correct?

fordfrog:
@VTimofeenko it's enough if we have 5.2.5-r1 and 5.0.9-r1. thanks!

VTimofeenko:
@fordfrog, done

fordfrog:
@VTimofeenko thanks for your contribution!

VTimofeenko:
@fordfrog, sure. Taking a look now.

PR #19654, opened 2021-02-25
sys-devel/gcc: remove unused patches

mm1ke:

As far as i can see the only two remaining patches which might get used are `gcc-configure-LANG.patch` and `gcc-configure-LANG.patch` (which are used via the `toolchain.eclass`)
I hope I didn't oversaw something.


trofi:
Pulled as is. Thank you!

PR #19657, opened 2021-02-25
media-gfx/gscan2pdf: Mark ALLARCHES

cjmayo:
Signed-off-by: Chris Mayo <aklhfex@gmail.com>

---

gscan2pdf is written in Perl.

### DIFF
--- a/media-gfx/gscan2pdf/metadata.xml
+++ b/media-gfx/gscan2pdf/metadata.xml
@@ -9,6 +9,7 @@
 		<email>proxy-maint@gentoo.org</email>
 		<name>Proxy Maintainers</name>
 	</maintainer>
+	<stabilize-allarches/>
 	<upstream>
 		<remote-id type="sourceforge">gscan2pdf</remote-id>
 	</upstream>


juippis:
@thesamesam ping, is this okay? I can understand it with perl, but I have no idea since I don't do AT work :)

thesamesam:
@juippis @cjmayo Unfortunately, probably not. This is really a mixed bag but a lot of Perl tends to contain hidden calls to things which are highly platform-dependent, like unpack. 

We _could_ audit to figure out if all of it is sane and looks OK, but it's not worth the time given we'd need to do this on each bump. Let's revisit this if we find stabilisations are taking too long for this package though.

(I was personally learning pro-ALLARCHES for some Perl until I spoke with the Perl Project about it a few months ago, who were quite strongly against.)

cjmayo:
Yes, wouldn't help to be explicit - I had read that paragraph. But it seemed worth putting it up for review, thanks! There a couple of uses of unpack. Although, I do have a motive in trying to get amd64 stabilization on https://bugs.gentoo.org/751391 to allow us to get rid of the old GTK 2 dependent ebuild.

PR #19658, opened 2021-02-25
app-i18n/transifex-client: fix failing test on 32 bit systems

robert7k:
Closes: https://bugs.gentoo.org/771660

The patch has already been submitted upstream: https://github.com/transifex/transifex-client/pull/316


app-i18n/transifex-client/files/transifex-client-timestamp.patch
```diff
@@ -0,0 +1,14 @@
+diff -Naur ../transifex-client-0.14.2.orig/tests/test_project.py ./tests/test_project.py
```
thesamesam, line 1: I've tidied up the patch a bit - see https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#patch-descriptions and ab3ec2595d0398dae2a2856122229b4454fe6a54. Thank you!

thesamesam (approved):

PR #19659, opened 2021-02-25
dev-libs/igraph: version bump to 0.9.0

epsilon-0:
moves to cmake based build system


dev-libs/igraph/igraph-0.9.0.ebuild
```diff
@@ -0,0 +1,52 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+CMAKE_MAKEFILE_GENERATOR="emake"
```
SoapGentoo, line 6: why is emake required?

dev-libs/igraph/igraph-0.9.0.ebuild
```diff
@@ -0,0 +1,52 @@
...
+inherit cmake
+
+DESCRIPTION="Creating and manipulating undirected and directed graphs"
+HOMEPAGE="http://www.igraph.org/"
+SRC_URI="https://github.com/igraph/igraph/releases/download/${PV}/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0/0"
```
SoapGentoo, line 14: so the SONAME is `0`?

dev-libs/igraph/files/igraph-0.9.0-cmakedirs.patch
```diff
@@ -0,0 +1,15 @@
...
+--- a/src/CMakeLists.txt
++++ b/src/CMakeLists.txt
+@@ -440,8 +440,8 @@ generate_export_header(igraph
+ include(GNUInstallDirs)
+ install(
+   TARGETS igraph
+-  RUNTIME DESTINATION bin
+-  LIBRARY DESTINATION lib
++  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
++  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
```
SoapGentoo, line 12: has this been submitted upstream?

dev-libs/igraph/igraph-0.9.0.ebuild
```diff
@@ -0,0 +1,52 @@
...
+		-DIGRAPH_ENABLE_TLS=$(usex threads)
+		-DBUILD_TESTING=$(usex test)
+	)
+	cmake_src_configure
+}
+
+src_test(){
+	cd "${BUILD_DIR}"
+	emake check
+}
```
SoapGentoo, line 52: nope, wrong, `cmake_src_test check`. Does the default test phase function not work?

SoapGentoo (changes requested):

dev-libs/igraph/files/igraph-0.9.0-cmakedirs.patch
epsilon-0: Am going to send it, they are pretty receptive and chill.

dev-libs/igraph/igraph-0.9.0.ebuild
epsilon-0: Yes, the default doesn't work, is also why I had to use `emake` as generator.
What you mentioned here, `cmake_src_test check`, also does not work.

dev-libs/igraph/igraph-0.9.0.ebuild
epsilon-0: Yes, the library name installed is `libigraph.so.0`

dev-libs/igraph/igraph-0.9.0.ebuild
epsilon-0: For the testing phase, as the default testing functions don't work.

dev-libs/igraph/igraph-0.9.0.ebuild
epsilon-0: Using `cmake_src_compile check` in src_test seems to work, would that be more acceptable?

SoapGentoo: yes, use that please

SoapGentoo: and work with upstream on just using CTest so we don't have to override it

dev-libs/igraph/igraph-0.9.0.ebuild
epsilon-0: removed

dev-libs/igraph/igraph-0.9.0.ebuild
epsilon-0: done

dev-libs/igraph/files/igraph-0.9.0-cmakedirs.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
+index 723d06d..bb6e041 100644
```
SoapGentoo, line 2: clean out the git metadata

dev-libs/igraph/igraph-0.9.0.ebuild
```diff
@@ -0,0 +1,51 @@
...
+		-DIGRAPH_USE_INTERNAL_GLPK=OFF
+		-DIGRAPH_USE_INTERNAL_GMP=OFF
+		-DIGRAPH_USE_INTERNAL_LAPACK=OFF
+		-DIGRAPH_ENABLE_TLS=$(usex threads)
+		-DBUILD_TESTING=$(usex test)
+	)
+	cmake_src_configure
+}
+
+src_test(){
```
SoapGentoo, line 49: `src_test() {`

SoapGentoo (changes requested):

dev-libs/igraph/files/igraph-0.9.0-cmakedirs.patch
epsilon-0: done

dev-libs/igraph/igraph-0.9.0.ebuild
epsilon-0: done

dev-libs/igraph/files/igraph-0.9.0-cmakedirs.patch
szhorvat: igraph developer here. This is fixed in 0.9.1 now, and do let us know if there are any other problems. :-)

dev-libs/igraph/igraph-0.9.0.ebuild
ntamas: fyi, upstream has recently committed a [patch](https://github.com/igraph/igraph/commit/d829842c97cba752c4f1999a832a29ec2c174fbf) that makes it possible to use `make test` similarly to `make check`

szhorvat: To be clear, this is not yet in 0.9.1, but it will be in 0.9.2.

PR #14668, opened 2020-02-15
sci-mathematics/topcom: new version

Alessandro-Barbieri:
* EAPI 7
* build dynamic libraries instead of static
* respect `*FLAGS`
* do not build with debug symbols

Package-Manager: Portage-2.3.87, Repoman-2.3.20
Signed-off-by: Alessandro Barbieri <lssndrbarbieri@gmail.com>

PR #19661, opened 2021-02-26
sci-mathematics/gfan: adapt to newer cddlib

kiwifb:
Bug: https://bugs.gentoo.org/show_bug.cgi?id=762817


kiwifb:
@orlitzky it would be good if we could have that or something like that merged. Same with PR #14668 which is not from me.

PR #19665, opened 2021-02-26
dev-libs/libclangformat-ide: Re-adding package

hansfn:
Supporting Clang 9.0.1, 10.0.1 and 11.0.{0,1}.
Support for Clang 10 was added in bug 734732.

I pledge to maintain this package for all stable versions of Clang as long as this patch is needed to get a working Clang format plugin in Qt Creator.

The goal is to get a new version of Qt Creator into the package tree.

Related bugs: 734732, 757756, 750527


hansfn:
I notice that I got a "Missing GCO sign-off" warning from the gentoo-bot. How do I best fix that. Can I just amend the commit and force push?

I did follow https://wiki.gentoo.org/wiki/GitHub_Pull_Requests and the commit is GPG signed, but it seems "repoman commit" doesn't add the signoff like I suspected. (The text before is "Note that if you intent to open a pull request and get your contribution in to the main tree, you need to properly sign-off all your commits as configured in step 3." **Which step 3?**)

Pesa:
Thanks a lot for volunteering! However, given [this upstream commit](https://code.qt.io/cgit/qt-creator/qt-creator.git/commit/?id=f7d8f8b7048d631e369b55db24df0a14e2437354), I wouldn't bother re-adding the patched version of libclangformat. (that commit will be in 4.15 which should be out in a couple of months, but I'll try to add a beta/rc package before then)

hansfn:
That is great news. I agree that we shouldn't waste any more energy on this. However, could you please release a 4.14.1 ebuild for Qt Creator with clangformat just removed / disabled. Then we can request removal of clang mask (from /usr/portage/profiles/base/package.use.mask).

PS! I'm happy to help you update or test ebuilds for new releases of Qt Creator - if time is the reason that we still only have 4.12.3 in the Gentoo tree.

hansfn:
I was hoping that @Pesa would make a quick release, but it seems he is busy. I guess your best bet is to use the latest patch from https://bugs.gentoo.org/750527 (and "dev-qt/qt-creator -clang" in /etc/portage/profile/package.use.mask). It's a bit sad that we have had this situation since late November.

lisyarus:
Thank you @hansfn, I can here from the linked bug & I'm already creating a local overlay with fixed qt-creator, thank you :) Completely agree that the situation is really sad.

PR #19674, opened 2021-02-26
sys-apps/lshw: bump to B.02.19.2

vaukai:
Closes: https://bugs.gentoo.org/747901

EAPI 7 bump

Not use static with sqlite
Bug: https://bugs.gentoo.org/720370

Replacing most of *-gentoo.patch with sed
Dropping older patches
   lshw-02.18b-gettext-array.patch
   lshw-02.18b-sgx.patch


a17r (changes requested):
Please consider what will happen here:
1) Two ebuilds are added with ~arch keywords.
2) Every user accepting ~arch keywords will get the latest version, no questions asked
3) 02.19.2 will likely receive no testing at all and won't be eligible for stabilisation

Best course of action would be to inquire with @gentoo/base-system what they prefer.

vaukai:
> Please consider what will happen here:
> 
>     1. Two ebuilds are added with ~arch keywords.

Even this is gonna go in two steps.  Let's see what happens next

PR #19675, opened 2021-02-26
Various cmake-utils -> cmake EAPI-7 bumps and cleanup

a17r:
Please don't use `cmake-utils.eclass` anymore.


media-sound/pnmixer/pnmixer-0.7.2-r1.ebuild
```diff
@@ -0,0 +1,59 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+MY_P="${PN}-v${PV}"
+inherit cmake xdg-utils
```
thesamesam, line 7: You not joining the xdg bandwagon?

media-sound/pamix/pamix-1.6-r1.ebuild
```diff
@@ -0,0 +1,55 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake
+
+if [[ ${PV} == 9999 ]] ; then
```
thesamesam, line 8: It'd be nicer to merge this with the existing block but this isn't your package, so I get the reluctance to meddle too much with style

media-sound/pamix/pamix-1.6-r1.ebuild
```diff
@@ -0,0 +1,55 @@
...
+	EGIT_REPO_URI="https://github.com/patroclos/PAmix.git"
+else
+	SRC_URI="https://github.com/patroclos/PAmix/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64 ~x86"
+	S="${WORKDIR}/PAmix-${PV}"
+fi
+
+RDEPEND="
+	media-sound/pulseaudio
+	sys-libs/ncurses:0=[unicode?]
```
thesamesam, line 28: unicode is forced on in profiles now, so there's no extra dep/rebuild involved. Should we just enable it unconditionally?

thesamesam (approved):
Made some misc notes but I completely get if you don't want to interfere with existing style, given this needs doing anyway.

ack on behalf of sci@.

games-arcade/performous/files/performous-1.1-boost-1.73.patch
```diff
@@ -0,0 +1,26 @@
+diff --git a/game/engine.cc b/game/engine.cc
+index 9f0f7ef..560ce71 100644
+--- a/game/engine.cc
++++ b/game/engine.cc
+@@ -8,6 +8,8 @@
+ #include <iostream>
+ #include <list>
+ 
++using namespace boost::placeholders;
```
SoapGentoo, line 9: honestly, is full qualification a problem? `using namespace boost::placeholders;` is a really brittle fixup, and I'd prefer fully qualifying arguments and not relying on ADL at all.

games-arcade/performous/performous-1.1-r2.ebuild
```diff
@@ -1,18 +1,15 @@
...
+EAPI=7
 
-CMAKE_REMOVE_MODULES="yes"
-CMAKE_REMOVE_MODULES_LIST="FindALSA FindBoost FindFreetype FindGettext FindJpeg FindPng FindTiff FindZ"
-inherit cmake-utils desktop gnome2-utils
-
-MY_PN="Performous"
-MY_P="${MY_PN}-${PV}"
-SONGS_PN="ultrastar-songs"
+CMAKE_REMOVE_MODULES_LIST=( FindALSA FindBoost FindFreetype FindGettext FindJpeg FindPng FindTiff FindZ )
```
SoapGentoo, line 6: not that it's super important: but does this have to be here? This doesn't affect metadata generation or other global scope thingies, so it could just be with `PATCHES` and friends?

games-arcade/performous/performous-1.1-r2.ebuild
```diff
@@ -1,18 +1,15 @@
...
+inherit cmake desktop xdg-utils
```
SoapGentoo, line 7: may I suggest `xdg`? or are you against that?

games-engines/openxcom/openxcom-1.0.0_p20180317-r1.ebuild
```diff
@@ -1,61 +1,69 @@
...
+BDEPEND="doc? ( app-doc/doxygen )"
 
-S="${WORKDIR}/OpenXcom-${COMMIT}"
 DOCS=( README.md )
 
 src_compile() {
-	use doc && cmake-utils_src_compile doxygen
-	cmake-utils_src_compile
+	use doc && cmake_build doxygen
+	cmake_src_compile
```
SoapGentoo, line 38: does the doc stuff have to be before the main stuff? Also, I generally prefer putting the `HTML_DOCS` definition in the same block as the `use doc`, but again, your call

media-sound/pamix/pamix-1.6-r1.ebuild
SoapGentoo: agreed, also `if [[ ${PV} == *9999 ]] ; then` please

media-sound/pamix/pamix-1.6-r1.ebuild
```diff
@@ -0,0 +1,55 @@
...
+EAPI=7
+
+inherit cmake
+
+if [[ ${PV} == 9999 ]] ; then
+	inherit git-r3
+fi
+
+DESCRIPTION="A PulseAudio NCurses mixer"
+HOMEPAGE="https://github.com/patroclos/PAmix"
```
SoapGentoo, line 13: definitely needs an empty line after this

sci-libs/votca-tools/votca-tools-1.6.4.ebuild
```diff
@@ -3,11 +3,9 @@
 
 EAPI=7
 
-CMAKE_MAKEFILE_GENERATOR="ninja"
+inherit cmake
 
-inherit cmake eutils multilib
-
-if [ "${PV}" != "9999" ]; then
+if [[ ${PV} != "9999" ]]; then
```
SoapGentoo, line 8: please switch this to positive logic

SoapGentoo (changes requested):

games-arcade/performous/files/performous-1.1-boost-1.73.patch
a17r: I'm not an expert; the patch is from Arch, but the Gentoo bug would have an alternative.

media-sound/pamix/pamix-1.6-r1.ebuild
a17r: yes, that's how I would do it, but rather leave it up to @Polynomial-C

games-arcade/performous/performous-1.1-r2.ebuild
a17r: I prefer listing of cmake meta variables pre-inherit

media-sound/pnmixer/pnmixer-0.7.2-r1.ebuild
a17r: `xdg.eclass` always contains the potential for future trouble with `cmake.eclass` because both export `src_prepare` and inherit order then suddenly becomes a factor. When it's not my own ebuilds I don't enforce one or the other method.

media-sound/pamix/pamix-1.6-r1.ebuild
Polynomial-C: I have no strong opinion. Leave it like this or change it but please don't break it :)

media-sound/pamix/pamix-1.6-r1.ebuild
Polynomial-C: No, that I want to keep. Don't change it.

diogocp (approved):

junghans (approved):

amadio:
I'd like to keep ROOT 6.20 for a bit longer. I will convert it to cmake.eclass, please leave that to me. Thanks.

a17r:
> I will convert it to cmake.eclass, please leave that to me. 

@amadio I'm not going to stop you, but you don't have to convert the old one to cmake.eclass. The main concern is the dependency restriction on old xrootd.

amadio:
> I'd like to keep ROOT 6.20 for a bit longer. I will convert it to cmake.eclass, please leave that to me. Thanks.

Nevermind, I see this is about XRootD, in that case you can go ahead with the removal.

hartwork:
@a17r I cherry-picked and pushed both your uriparser commits just now, good work, thank you! :+1: :pray:

candrews (approved):

amadio:
Ok for dev-libs/vc.

benkohler:
Go ahead on dev-libs/libtsm

a17r:
@benkohler cleanup old version as well?

benkohler:
Yeah go ahead, I had been keeping the old one since it is the original upstream but the new fork has proven to work everywhere

sbraz:
@a17r go ahead with unittest++.

PR #19677, opened 2021-02-26
*/*: Fix test failures

jsmolic:


net-im/err/err-6.1.7.ebuild
```diff
@@ -44,9 +45,8 @@ RDEPEND="${DEPEND}
 		dev-python/pyasn1[${PYTHON_USEDEP}]
 		dev-python/pyasn1-modules[${PYTHON_USEDEP}]
 		dev-python/slixmpp[${PYTHON_USEDEP}]
-	)
-"
-S="${WORKDIR}/${MY_P}"
+	)"
+BDEPEND="test? ( dev-python/mock[${PYTHON_USEDEP}] )"
 
 distutils_enable_tests pytest
```
thesamesam, line 51: btw, I suspect the IRC comment below is obsolete now

jsmolic: Oh, yeah, it's a leftover from old version, I'll remove it

PR #19683, opened 2021-02-27
app-misc/goobook: Add missing dev-python/xdg dep 

jsmolic:


dev-python/xdg/metadata.xml
```diff
@@ -0,0 +1,16 @@
...
+		<name>Kalin KOZHUHAROV</name>
+		<email>kalin@thinrope.net</email>
+	</maintainer>
+	<maintainer type="project">
+		<email>proxy-maint@gentoo.org</email>
+		<name>Proxy Maintainers</name>
+	</maintainer>
+	<stabilize-allarches/>
+	<upstream>
+		<remote-id type="pypi">xdg</remote-id>
```
thesamesam, line 14: No GitHub remote-id?

dev-python/xdg/metadata.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
```
thesamesam, line 4: Also, I think I agree with the outcome here (giving new deps to the maintainers) but let's check Soap is happy. I think he is though.

dev-python/xdg/xdg-5.0.1.ebuild
```diff
@@ -0,0 +1,15 @@
...
+PYTHON_COMPAT=( python3_{7..9} )
+inherit distutils-r1
+
+DESCRIPTION="Implementation of the XDG Base Directory Specification in Python"
+HOMEPAGE="https://pypi.org/project/xdg"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="ISC"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
```
thesamesam, line 15: Tests are a single file right, so not worth bothering, or?

dev-python/xdg/metadata.xml
jsmolic: Yes, done now :smiley:

dev-python/xdg/xdg-5.0.1.ebuild
jsmolic: Yeah, it's a single file, also, I tested github tarball, had to use pyproject.toml which apparantly doesn't work? (`file xdg.py (for module xdg) not found` so I just did a build from pypi tarball for now (it has a setup.py)

dev-python/xdg/metadata.xml
jsmolic: Sure, I do it like that for all new packages so it should be good to go

SoapGentoo (approved):

PR #19686, opened 2021-02-27
dev-lang/erlang: remove unused patch

mm1ke:


trofi:
Pulled as is. Thank you!

PR #19688, opened 2021-02-27
media-gfx/zbar: fix musl build

arkamar:
`media-gfx/zbar` fails to build with `musl` libc with following error due to a missing `locale.h` header file.
```
/var/tmp/portage/media-gfx/zbar-0.23.1/work/zbar-0.23.1/zbarimg/zbarimg.c: In function 'main':
/var/tmp/portage/media-gfx/zbar-0.23.1/work/zbar-0.23.1/zbarimg/zbarimg.c:317:5: warning: implicit declaration of function 'setlocale' [-Wimplicit-function-declaration]
  317 |     setlocale (LC_ALL, "");
      |     ^~~~~~~~~
/var/tmp/portage/media-gfx/zbar-0.23.1/work/zbar-0.23.1/zbarimg/zbarimg.c:317:16: error: 'LC_ALL' undeclared (first use in this function)
  317 |     setlocale (LC_ALL, "");
      |                ^~~~~~
/var/tmp/portage/media-gfx/zbar-0.23.1/work/zbar-0.23.1/zbarimg/zbarimg.c:317:16: note: each undeclared identifier is reported only once for each function it appears in
```

The patch is taken from upstream https://github.com/mchehab/zbar/pull/115. I am not entirely sure if this can be fixed without revbump, but I think it can. Anyway, let me know if I should change it somehow or if this should be moved to a musl overlay.

### DIFF
--- a/media-gfx/zbar/files/zbar-0.23.1_musl_include_locale_h.patch
+++ b/media-gfx/zbar/files/zbar-0.23.1_musl_include_locale_h.patch
@@ -1,0 +1,26 @@
+diff --git a/zbarcam/zbarcam.c b/zbarcam/zbarcam.c
+index f7ea281..2452613 100644
+--- a/zbarcam/zbarcam.c
++++ b/zbarcam/zbarcam.c
+@@ -37,6 +37,7 @@
+ #ifdef ENABLE_NLS
+ #include "../zbar/gettext.h"
+ # include <libintl.h>
++# include <locale.h>
+ # define _(string) gettext(string)
+ #else
+ # define _(string) string
+diff --git a/zbarimg/zbarimg.c b/zbarimg/zbarimg.c
+index 0796fd8..3192ec1 100644
+--- a/zbarimg/zbarimg.c
++++ b/zbarimg/zbarimg.c
+@@ -42,6 +42,7 @@
+ #ifdef ENABLE_NLS
+ #include "../zbar/gettext.h"
+ # include <libintl.h>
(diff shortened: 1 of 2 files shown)


thesamesam:
No [need](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html) for revbump -- but it's good to ask whenever you're not sure. In this case, it's a build-only fix (if it builds, you're fine, no need for the fix) so we don't need to force people who successfully built it to rebuild.

arkamar:
> No [need](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html) for revbump -- but it's good to ask whenever you're not sure. In this case, it's a build-only fix (if it builds, you're fine, no need for the fix) so we don't need to force people who successfully built it to rebuild.

Thanks! I have read this documentation but I wasn't sure if it applies for stable packages as well.

thesamesam:
Yeah, all good for stable too if it's a build fix like this. But being cautious is good. Pushed an extra change (adding a reference to the patch itself) in 39a498488dcbed644cdf0cfa10c8ca9d70e7066a too (in line with the [new](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#patch-descriptions) fancy Patches guide!)

arkamar:
> [new](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#patch-descriptions) fancy Patches guide!

Ok, I'll read this one as well :)

PR #19694, opened 2021-02-27
games-engines/gargoyle: Fix typo in BDEPEND

monoidic:
s/BDEPEBD/BDEPEND/
Makes sure build deps are actually pulled in.


thesamesam:
Nice catch, thanks!

PR #19695, opened 2021-02-27
xfce-extra/xfce4-kbdleds-plugin-0.2.3: bump

Hawk777:


xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
```diff
@@ -0,0 +1,43 @@
...
+
+EAPI=7
+
+inherit xdg-utils
+
+DESCRIPTION="A panel plug-in to show state of Caps, Num and Scroll Lock keys"
+HOMEPAGE="https://github.com/oco2000/xfce4-kbdleds-plugin"
+SRC_URI="https://github.com/oco2000/${PN}/archive/${P}.tar.gz"
+S="${WORKDIR}/${PN}-${P}"
+PATCHES=( "${FILESDIR}/fix-build-0.2.3.patch" )
```
mgorny, line 12: Please move it below all other vars, into a separate group. Also, please start the patch name with `${P}-`.

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
```diff
@@ -0,0 +1,43 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit xdg-utils
+
+DESCRIPTION="A panel plug-in to show state of Caps, Num and Scroll Lock keys"
+HOMEPAGE="https://github.com/oco2000/xfce4-kbdleds-plugin"
```
mgorny, line 9: Please also update or add remote-id in metadata.xml.

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
```diff
@@ -0,0 +1,43 @@
...
+SRC_URI="https://github.com/oco2000/${PN}/archive/${P}.tar.gz"
+S="${WORKDIR}/${PN}-${P}"
+PATCHES=( "${FILESDIR}/fix-build-0.2.3.patch" )
+
+LICENSE="GPL-2 LGPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
```
mgorny, line 17: Drop empty IUSE.

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
```diff
@@ -0,0 +1,43 @@
...
+
+RDEPEND="xfce-base/libxfce4ui
```
mgorny, line 19: `:=` for libxfce*

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
```diff
@@ -0,0 +1,43 @@
...
+	xfce-base/libxfce4util
+	xfce-base/xfce4-panel"
+DEPEND="${RDEPEND}
+	dev-util/intltool
+	sys-devel/gettext
+	virtual/pkgconfig"
```
mgorny, line 25:
```suggestion
DEPEND=${RDEPEND}
BDEPEND="
	dev-util/intltool
	sys-devel/gettext
	virtual/pkgconfig"
```

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
```diff
@@ -0,0 +1,43 @@
...
+
+src_configure() {
+	./autogen.sh || die
```
mgorny, line 28: Nononono. There's autotools.eclass for that, and I'm pretty sure you're going to need a dep on xfce4-dev-tools.

Also, could you ask upstream to make `make distcheck` or make a ready dist tarball yourself? It'd be better than requiring all the users to do regen.

mgorny (changes requested):

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
Hawk777: Moved and file renamed.

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
Hawk777: Done.

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
Hawk777: Done.

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
Hawk777: Added. I thought it wasn’t necessary because I looked at the `libxfce4ui` ebuild and it just had `SLOT="0"` and nothing else, so I removed it (it used to be `:=` in the old version). Sorry about that.

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
Hawk777: Moved into `BDEPEND`.

mgorny:
FYI, unless you find that helpful to keep track of your work, no need to reply to individual comments. Just ping me when it's ready for another round.

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
Hawk777: `./autogen.sh` is what the upstream README told me to do, that’s why I did it. I tried calling `eautoreconf` in `src_prepare` (I think that’s where it’s supposed to be used?) but just got six copies of `/bin/grep: configure.??: No such file or directory` instead, followed by bailing out. The `autogen.sh` shipped by upstream runs `/usr/bin/xdt-autogen`, which is part of `xfce4-dev-tools` (yes I added the dep, thanks for noticing that).

I don’t really know much about autotools. Can you give any guidance about what I should be doing here?

Hawk777:
OK, will do that next time @mgorny ; I had some trouble with the autotools part if you can help.

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
mgorny: Just run `make distcheck` to make a good tarball with pregenerated files and use that instead.

Hawk777: OK, where should I put the tarball though? It’s not my project, and since I’m not a Gentoo dev I can’t upload it to a Gentoo-infra-hosted home directory.

mgorny: Try `https://dev.gentoo.org/~mgorny/dist/${P}.tar.bz2` -- and diff the unpacked directory against your archive to be sure.

Hawk777: Yep, that worked (and once I upgraded to unstable autoconf and automake I was able to generate a tarball with identical contents, albeit not byte-for-byte identical, I guess maybe it had different file timestamps). At this point I’m not sure whether `xfce4-dev-tools` is still needed (now that `autogen` no longer needs to run); any thoughts on that? The previous version didn’t depend on it.

Hawk777: It compiled fine without it installed, so I’m guessing not (and have removed it accordingly). If you disagree, let me know and I’ll re-add it.

Hawk777: @mgorny AFAIK that was all the outstanding items.

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+EAPI=7
+
+inherit xdg-utils
+
+DESCRIPTION="A panel plug-in to show state of Caps, Num and Scroll Lock keys"
+HOMEPAGE="https://github.com/oco2000/xfce4-kbdleds-plugin"
+SRC_URI="https://dev.gentoo.org/~mgorny/dist/${P}.tar.bz2"
+
+LICENSE="GPL-2 LGPL-2"
```
mgorny, line 12: Could you check if it isn't 'or newer'? i.e. GPL-2+ LGPL-2+?

xfce-extra/xfce4-kbdleds-plugin/xfce4-kbdleds-plugin-0.2.3.ebuild
```diff
@@ -0,0 +1,39 @@
...
+	xfce-base/libxfce4ui:=
+	xfce-base/libxfce4util:=
+	xfce-base/xfce4-panel"
+DEPEND=${RDEPEND}
+BDEPEND="
+	dev-util/intltool
+	sys-devel/gettext
+	virtual/pkgconfig"
+
+PATCHES=( "${FILESDIR}/${P}-xfce-4.16.patch" )
```
mgorny, line 26: Link to the upstream PR would be nice.

mgorny (commented):
Thanks. It looks quite good already. If you could also do a quick license check, that'd be great.

Also, commit message should indicate the new version, e.g. 'bump to X.Y.Z'.

Hawk777:
Addressed your two points. I am not 100% certain what the right thing to do with the license is. Here is the license snippet at the top of the source files:
```
 *  Copyright (c) 2011-2019 OCo <oco2000@gmail.com>
 *
 *  This program is free software; you can redistribute it and/or modify
 *  it under the terms of the GNU General Public License as published by
 *  the Free Software Foundation; either version 2 of the License, or
 *  (at your option) any later version.
 *
 *  This program is distributed in the hope that it will be useful,
 *  but WITHOUT ANY WARRANTY; without even the implied warranty of
 *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 *  GNU Library General Public License for more details.
 *
 *  You should have received a copy of the GNU General Public License
 *  along with this program; if not, write to the Free Software
 *  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
```

So, it says the program is released under GPL-2+, but the second paragraph, regarding warranty, says to look at the LGPL for details, not the GPL. Since the warranty paragraphs of the GPL-2 and LGPL-2 are identical, I thought it was reasonable to just write `LICENSE="GPL-2+"`, but let me know if it should really have been `LICENSE="GPL-2+ LGPL-2"`, or maybe `LICENSE="GPL-2+ LGPL-2+"`, but since the *library* general public license (as opposed to the *lesser* general public license) only existed at V2, I don’t see how one could justify writing `LGPL-2+` based on this text.

mgorny:
Based on this para it's GPL-2+. But you may need to look into files inside the repo, maybe part of the sources (e.g. a library) are LGPL-2.1+ or alike.

Hawk777:
That’s what I thought made the most sense. Every source file has the same header (the one I pasted above); I did check that already.

mgorny (approved):
Thanks!

Hawk777:
@mgorny Thanks for the review. BTW I did not remove the `package.mask` entry; I wasn’t sure whether that was supposed to be done here or by a separate process.

PR #19706, opened 2021-02-28
sci-mathematics/rw: new upstream version

kiwifb:


kiwifb:
@orlitzky upstream fixed the links so I cleaned that part of the ebuild. What was the story about igraph that was so bad as to disable the executable?

orlitzky:
> @orlitzky What was the story about igraph that was so bad as to disable the executable?

We were stuck on an old version of igraph for a long time because of the bundled libraries. A Gentoo dev had written a patch to unbundle most things for igraph-0.7.x, but the patch didn't work for igraph-0.8.x, and igraph-0.8.x bundles EVEN MORE libraries -- many of which already have corresponding gentoo packages. That was blocking any igraph upgrade, and so igraph was scheduled to be removed. The `rw` executable was the only thing using igraph, so I disabled the executable to make removal possible.

While I wasn't looking, someone else upgraded igraph without noticing that it bundled 10+ libraries. So now we have igraph-0.8.x in the tree, but it's still not something I'd want installed.

kiwifb:
igraph is an optional dependency of sage. I may have to take custody then.

kiwifb:
igraph doesn't look too bad, it is just a bunch of recursive dependencies that are annoying. But one of the vendored package doesn't actually have a proper release as a C-library https://github.com/ntamas/plfit. So, that bit is annoying.

orlitzky:
Last time I checked I came up with this list of bundled libs:

  * AMD
  * arpack
  * bliss
  * CHOLMOD
  * cliquer
  * COLAMD
  * cxsparse
  * plfit
  * prpack

And best of all, some of them have been modified from the upstream version to make them work with igraph. I tried to talk the devs out of adding some of those, but they're happy with the way they're doing things. It's only going to be a liability down the road, one of those things we can never upgrade because no one feels like updating the unbundling patch after every release.

kiwifb:
OK, modification is just plain bad form. But at least the list shrinked a bit
* glpk
* * AMD (shipped in glpk)
* * COLAMD (shipped in glpk)
* f2c (for "lapack/arpack" below)
* arpack (disguised in the lapack folder because it combines the bits of blas/lapack it needs with it)
* cxsparse
* plfit
* "mini" gmp
 
I'll admit it is probably a moving target depending on what the devs have found the day before the release.

kiwifb:
OK, I officially do not know what I am talking about. I thought vendored stuff was all in the "vendor" folder. But there is a lot mixed in the src folder. At least there seem to be movement to put the stuff in the vendor folder which would makes things easier.

PR #19714, opened 2021-02-28
media-gfx/openscad: bump to 2021.01

waebbl:
Bug: https://bugs.gentoo.org/773217
Closes: https://bugs.gentoo.org/769278


media-gfx/openscad/openscad-2021.01.ebuild
```diff
@@ -0,0 +1,105 @@
...
+SITEFILE="50${PN}-gentoo.el"
+
+DESCRIPTION="The Programmers Solid 3D CAD Modeller"
+HOMEPAGE="https://www.openscad.org/"
+SRC_URI="https://github.com/${PN}/${PN}/releases/download/${P}/${P}.src.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64 ~x86"
+IUSE="ccache emacs"
```
ajakk, line 17: Do we really need a USE=ccache? Is FEATURES=ccache broken?

media-gfx/openscad/openscad-2021.01.ebuild
```diff
@@ -0,0 +1,105 @@
...
+RESTRICT="test"
```
ajakk, line 18: Why?

ajakk (commented):
You can also close https://bugs.gentoo.org/769278 with your commit message.

media-gfx/openscad/openscad-2021.01.ebuild
waebbl: Don't use it, so I don't know whether it's broken. It's to avoid an automagic dep and use if the binary is found.
See https://github.com/gentoo/gentoo/pull/19714/files#diff-b734e3660f655b6e0ef09b77e8ebe066d32ce4cb5edd7f38a95a5ea7934d3d05

media-gfx/openscad/openscad-2021.01.ebuild
waebbl: They are not ready yet. Most of them work, if you build them with cmake, which isn't yet officially supported. See https://github.com/gentoo/gentoo/pull/19412/files#diff-245dbc481dcedeb832abdc08edc62a2277fa08d9f234768a3ce8e0f0d4a2a187

waebbl:
> You can also close https://bugs.gentoo.org/769278 with your commit message.

Yeah sure, forgot about this bug.

media-gfx/openscad/openscad-2021.01.ebuild
waebbl: Will take some time to test this out on a clean machine. The package has lots of dependencies.

ajakk: What I mean is the patch can probably be unconditionally applied. If the build respects CC/CXX then FEATURES=ccache should do everything necessary to use it - explicit build system support shouldn't be necessary.

media-gfx/openscad/openscad-2021.01.ebuild
ajakk: Can you add a comment to this effect in the ebuild?

media-gfx/openscad/openscad-2021.01.ebuild
waebbl: I see.
FEATURES=ccache won't pull in the ccache package though, so isn't the USE flag necessary to pull the package in if users have set the feature?

waebbl: It's working if I leave the USE flag to pull in the package.

ajakk: If a user wants FEATURES=ccache they can install it themselves. Ebuilds don't need a dependency on it.

thesamesam: Note that this is sometimes misleading because upstreams expose a ccache option but we usually want to force it off. When we're using ccache with Portage, they don't need to know about it.

waebbl: > If a user wants FEATURES=ccache they can install it themselves. Ebuilds don't need a dependency on it.

Understood. Although it's IMO counter-intuitive. If I, as a user, add ccache to FEATURES, I expect the package manager and / or the ebuild to pull in any necessary dependencies, so the package in question can immediately use it. Like with FEATURES="test", which activates a USE flag and pulls in any needed deps.

> Note that this is sometimes misleading because upstreams expose a ccache option but we usually want to force it off. When we're using ccache with Portage, they don't need to know about it.

Thanks for the concise explanation!

waebbl:
Closing. Will be added to PR #19412 as per request from @juippis

PR #19720, opened 2021-03-01
dev-libs/libebml: revbump for gcc-11 fix

kobboi:
Patch source: https://github.com/Matroska-Org/libebml/commit/f0bfd53
(I dropped the ChangeLog part as that depends on other commits to apply nicely)

Closes: https://bugs.gentoo.org/773538


thesamesam:
Note that we [don't need](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html) to revbump for build fixes

dev-libs/libebml/libebml-1.4.2.ebuild
```diff
@@ -13,6 +13,11 @@ LICENSE="LGPL-2.1"
 SLOT="0/5" # subslot = soname major version
 KEYWORDS="~alpha amd64 arm arm64 ~hppa ~ia64 ppc ppc64 sparc x86 ~amd64-linux ~x86-linux ~ppc-macos"
 
+src_prepare() {
```
thesamesam, line 16: I'm going to move this into PATCHES=() instead

thesamesam:
Thank you!

PR #19721, opened 2021-03-01
net-firewall/xtables-addons: Version updated to 3.17

jpds:
Tested on `amd64` system.


blueness:
pushed.  thanks for the bump.  please test.

PR #19729, opened 2021-03-02
app-laptop/i8kutils: update to 1.43 and adoption

myrvogna:
Updated the package to the latest version, as well as several other improvements to the ebuild:
- adding missing runtime dependency on tcl
- removing tk USE flag, the tk-based applet having been removed in 1.43
- adding a non-fatal check for the necessary `dell-smm-hwmon` kernel module
- passing all toolchain flags properly to the Makefile
- now using the upstream-provided systemd unit file
- adding a new OpenRC init script, the old one referencing i8kbuttons features already removed from version 1.42

Alongside adopting the package as a proxied maintainer.


app-laptop/i8kutils/files/i8kmon.init
```diff
@@ -0,0 +1,19 @@
...
+	after modules
+}
+
+command="/usr/bin/i8kmon"
+command_args="--nouserconfig"
+command_background=true
+pidfile="/run/${RC_SVCNAME}.pid"
+
+start_pre() {
+	if [ ! -e /proc/i8k ]; then
```
juippis, line 15: You could use `[[ ]]`, otherwise LGTM!

juippis (approved):

app-laptop/i8kutils/files/i8kmon.init
thesamesam: I thought we needed OpenRC scripts to be POSIX-compliant, but maybe it's different with openrc-init or something? (I've never really poked at it that much)

ionenwks: Yeah no [[ ]] in there, openrc claims compatibility with busybox sh, ash, dash, bash, mksh, and zsh, and then our scripts break that.

juippis: Well damn, someone should tell lxd's maintainer that so people don't learn bad examples from there :P 
https://gitweb.gentoo.org/repo/gentoo.git/tree/app-emulation/lxd/files/lxd-4.0.0.initd#n31

ionenwks: We have a tracker too, no that it seen much activity [bug #609070](https://bugs.gentoo.org/609070)

PR #19735, opened 2021-03-02
dev-python/pytest-salt: bug fix

janko-j1:
When pytest-salt is involved in other package's testing suite, it can
cause the corresponding package suite to fail.

Closes:https://bugs.gentoo.org/740400

Signed-off-by: David Denoncin <ddenoncin@gmail.com>


chutz:
Please revbump the ebuild when making this sort of change (it's a simple rename the ebuild, replace the `-r1` with `-r2` in the file name)

janko-j1:
Thanks for the reminder. Do I keep -r1 or is it fine as it is?

PR #19736, opened 2021-03-02
dev-util/bazel: enable statically linked libstdc++ for gentoo prefix

littlewu2508:
Bug: https://bugs.gentoo.org/773982

### DIFF
--- a/dev-util/bazel/bazel-3.2.0.ebuild
+++ b/dev-util/bazel/bazel-3.2.0.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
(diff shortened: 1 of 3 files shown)


littlewu2508:
In this pull request I added two USE flags -- 'prefix' and 'static-libs'. I want to prefer and force static-libs when prefix is used, and '-static-libs' as the default value when prefix is not spcefied. The current syntax using REQUIRED_USE only guarantees static-libs must come with prefix. Is there a better syntax?

zmedico:
> In this pull request I added two USE flags -- 'prefix' and 'static-libs'. I want to prefer and force static-libs when prefix is used, and '-static-libs' as the default value when prefix is not spcefied. The current syntax using REQUIRED_USE only guarantees static-libs must come with prefix. Is there a better syntax?

REQUIRED_USE is for requirements, not preferences. You can set a preference in the corresponding prefix profile's package.use. For automatic solving of REQUIRED_USE, we have https://www.gentoo.org/glep/glep-0073.html.

zmedico (changes requested):
The `REQUIRED_USE="prefix? ( static-libs )"` setting only makes sense if that's the only workable solution for prefix. You can use package.use.force in the relevant profile if it needs to be forced for a particular profile.

littlewu2508:
> 
> 
> The `REQUIRED_USE="prefix? ( static-libs )"` setting only makes sense if that's the only workable solution for prefix. You can use package.use.force in the relevant profile if it needs to be forced for a particular profile.

Currently it is the only workable solutions for prefix, so I think REQUIRED_USE is necessary. I updated profiles/features/prefix/package.use, which can enable static-libs by default. I suppose this is a resonable solution.

littlewu2508:
@zmedico Any updates on reviewing?

zmedico:
Thanks!

PR #19737, opened 2021-03-02
eclass/meson.eclass: add override for -Werror

epsilon-0:


Whissi:
I think this the wrong attempt: Packages should get patched (and patch should get upstreamed) to not auto-enable `-Werror` but we still want to be able to use `-Werror`, i.e. on explicit user wish. The eclass change would prevent that, not?

epsilon-0:
> I think this the wrong attempt: Packages should get patched (and patch should get upstreamed) to not auto-enable `-Werror` but we still want to be able to use `-Werror`, i.e. on explicit user wish. The eclass change would prevent that, not?

No, a user can still specify an override `-Dwerror=true` in `MYMESONARGS`, like
```
$ MYMSEONARGS="-Dwerror=true" emerge -av <some-pkg>
```

lanodan:
Related bug: https://bugs.gentoo.org/754279

PR #19741, opened 2021-03-02
dev-python/aesara: Another rename of theano (version 2.0.1)

jsmolic:


dev-python/aesara/aesara-2.0.1.ebuild
```diff
@@ -0,0 +1,59 @@
...
+
+MY_P=aesara-rel-${PV}
+DESCRIPTION="Library for operating on mathematical expressions with multi-dimensional arrays"
+HOMEPAGE="https://github.com/pymc-devs/aesara"
+SRC_URI="https://github.com/pymc-devs/aesara/archive/rel-${PV}.tar.gz -> ${MY_P}.tar.gz"
+S="${WORKDIR}/${MY_P}"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~x86"
```
jsmolic, line 18: I think we can avoid another keywording here so I pushed it with `~arm{,64}`

dev-python/aesara/aesara-2.0.1.ebuild
```diff
@@ -0,0 +1,59 @@
...
+distutils_enable_sphinx doc 'dev-python/sphinx_rtd_theme'
+distutils_enable_tests pytest
+
+python_prepare_all() {
+	sed -i -e "s/tests.\*/tests\*/" setup.py || die
+
+	distutils-r1_python_prepare_all
+}
+
+python_test() {
```
jsmolic, line 39: It's a straight copy from previous version but I didn't finish running tests because they're really long and resource-heavy

mgorny:
BTW am I thinking correctly that we can lastrite the 'original' theano?

thesamesam:
@mgorny Yes.

jsmolic:
Yeah, I just tested sympy with theano-pymc, I'm noticing few test failures, I'll look into fixing it and we can last rite it afterwards

PR #19745, opened 2021-03-03
net-p2p/eiskaltdcpp: Bump to 2.4.2

klondi:
Bump to 2.4.2 which fixes an issue with searches.
Add da and ie translations.
Remove FindASPELL module provided by modern Cmake.
Remove old patches that are now in upstream.
Port changes to the live ebuild too.

Closes: https://bugs.gentoo.org/774027


juippis (approved):
We usually split adding and removing ebuilds to their own, separate commits. If 2.4.1 needs to be restored, we can just revert the commit that removes it, while now reverting that commit would also delete 2.4.2. But it works fine for me.

PR #19746, opened 2021-03-03
dev-util/perf: fix libdir

12101111:
Bug: https://bugs.gentoo.org/774006
Signed-off-by: Han PuYu <w12101111@gmail.com>

### DIFF
--- a/dev-util/perf/perf-5.10.16.ebuild
+++ b/dev-util/perf/perf-5.10.16.ebuild
@@ -198,6 +198,7 @@ perf_make() {
 		NO_ZLIB= \
 		WERROR=0 \
 		LIBDIR="/usr/libexec/perf-core" \
+		libdir="${EPREFIX}/usr/$(get_libdir)" \
 		"$@"
 }
 
--- a/dev-util/perf/perf-5.11.ebuild
+++ b/dev-util/perf/perf-5.11.ebuild
@@ -198,6 +198,7 @@ perf_make() {
 		NO_ZLIB= \
 		WERROR=0 \
 		LIBDIR="/usr/libexec/perf-core" \
+		libdir="${EPREFIX}/usr/$(get_libdir)" \
 		"$@"
 }
 
--- a/dev-util/perf/perf-5.8.ebuild
+++ b/dev-util/perf/perf-5.8.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
@@ -184,6 +184,7 @@ perf_make() {
 		NO_ZLIB= \
 		WERROR=0 \
 		LIBDIR="/usr/libexec/perf-core" \
+		libdir="${EPREFIX}/usr/$(get_libdir)" \
 		"$@"
 }
 
--- a/dev-util/perf/perf-5.9.9.ebuild
+++ b/dev-util/perf/perf-5.9.9.ebuild
@@ -198,6 +198,7 @@ perf_make() {
 		NO_ZLIB= \
 		WERROR=0 \
 		LIBDIR="/usr/libexec/perf-core" \
+		libdir="${EPREFIX}/usr/$(get_libdir)" \
 		"$@"
 }
 


dev-util/perf/perf-5.10.16.ebuild
```diff
@@ -198,6 +198,7 @@ perf_make() {
 		NO_ZLIB= \
 		WERROR=0 \
 		LIBDIR="/usr/libexec/perf-core" \
+		libdir="/usr/$(get_libdir)" \
```
amadio, line 201: Shouldn't this be `lib=$(get_libdir)` as mentioned in the bug? I need to check if this works on prefix or not before merging.

12101111: `lib=$(get_libdir)` will install those so files to `imagelib/traceevent/plugins/`.

ionenwks: Really? I tried lib=lib and I get the expected:
`/usr/lib/traceevent/plugins/plugin_tlb.so`
On prefix with lib=test:
`/home/ionen/prefix/usr/test/traceevent/plugins/plugin_tlb.so`
Not that I tested all that much.

amadio: We do want those in `usr/$(get_libdir)/traceevent/plugins`, right? These are not general libraries, but plugins only used by perf, so they are not meant to go straight into `/usr/lib64` or `/usr/lib` like `libperf-*.so`. The fix is to use `lib` rather than `lib64` with musl as far as I can tell from the bug. Howver, I guess your patch is equivalent as it doesn't change `plugindir` in any case.

12101111:
This is my test.
original ebuild on musl:
```
> ls /tmp/portage/dev-util/perf-5.11/image/usr/lib64/traceevent/plugins
plugin_cfg80211.so  plugin_futex.so    plugin_jbd2.so  plugin_kvm.so       plugin_sched_switch.so  plugin_tlb.so
plugin_function.so  plugin_hrtimer.so  plugin_kmem.so  plugin_mac80211.so  plugin_scsi.so          plugin_xen.so
```

`libdir="$(get_libdir)"` on musl :

```
       for plugin in  plugin_jbd2.so plugin_hrtimer.so plugin_kmem.so plugin_kvm.so plugin_mac80211.so plugin_sched_switch.so plugin_function.so plugin_futex.so plugin_xen.so plugin_scsi.so plugin_cfg80211.so plugin_tlb.so; do 		if [ ! -d ''/tmp/portage/dev-util/perf-5.11.2/image'lib/traceevent/plugins' ]; then install -d -m 755 ''/tmp/portage/dev-util/perf-5.11.2/image'lib/traceevent/plugins'; fi; install  $plugin ''/tmp/portage/dev-util/perf-5.11.2/image'lib/traceevent/plugins'; done
```
```
> ls /tmp/portage/dev-util/perf-5.11/
build-info  distdir  files  homedir  image  imagelib  temp  work
> ls /tmp/portage/dev-util/perf-5.11/imagelib/traceevent/plugins
plugin_cfg80211.so  plugin_futex.so    plugin_jbd2.so  plugin_kvm.so       plugin_sched_switch.so  plugin_tlb.so
plugin_function.so  plugin_hrtimer.so  plugin_kmem.so  plugin_mac80211.so  plugin_scsi.so          plugin_xen.so
```

`libdir="${EPREFIX}/usr/$(get_libdir)"` on musl:

```
> ls /tmp/portage/dev-util/perf-5.11/image/usr/lib/traceevent/plugins
plugin_cfg80211.so  plugin_futex.so    plugin_jbd2.so  plugin_kvm.so       plugin_sched_switch.so  plugin_tlb.so
plugin_function.so  plugin_hrtimer.so  plugin_kmem.so  plugin_mac80211.so  plugin_scsi.so          plugin_xen.so
```

`libdir="${EPREFIX}/usr/$(get_libdir)"` on gentoo prefix:

```
> ls /tmp/portage/dev-util/perf-5.11/image/home/han/.gentoo/usr/lib64/traceevent/plugins
plugin_cfg80211.so  plugin_futex.so    plugin_jbd2.so  plugin_kvm.so       plugin_sched_switch.so  plugin_tlb.so
plugin_function.so  plugin_hrtimer.so  plugin_kmem.so  plugin_mac80211.so  plugin_scsi.so          plugin_xen.so
```

amadio:
Ok, great, that looks correct, but there's a difference with the patch, which is to prepend `${EPREFIX}` to `libdir`. Does it also work without it the same way or do we need the prepend it? That was my original question.

amadio:
Oh, I see you added `${EPREFIX}` to your latest patch, I guess I can merge that now. Thanks a lot!

PR #19755, opened 2021-03-03
dev-python/aesara: Add blocker against dev-python/theano-pymc

jsmolic:
Closes: https://bugs.gentoo.org/774072


thesamesam:
The blocker should be two-way, right? (added to thano-pymc, and I guess theano).

jsmolic:
Yes, yes, we should block against theano as well

jsmolic:
theano and theano-pymc should probably block each other as well

mgorny:
Maybe we should strip `bin/__init__.py` in all packages. It looks namespace-ish.

mgorny (approved):
Looks good, thanks!

PR #19759, opened 2021-03-04
media-gfx/blender: version bumps 2.83.12, 2.92.0

redchillipadi:
A version bump to blender 2.92.0 simply required renaming the 2.91.2
ebuild. The dependencies have not changed.

Python 3.9 deleted PyNoArgsFunction (issue 18091) which is used in the
Blender mathutils. This was fixed in a patch for blender 2.90+ (see
upstream T78089), and should be backported to 2.83 LTS soon.

In the interim I have reproduced the patch and confirm that 2.83.12
also builds with python 3.9 the patch applied.


thesamesam:
ping

DarkDefender:
I think this can be closed now.
I guess we forgot to close this when #20565 was merged.

PR #19761, opened 2021-03-04
Remove old  net-libs/xrootd{,-ceph} and dev-python/nbval

jsmolic:
* Also requires dropping old sci-physics/root


amadio:
Fine for sci-physics/root, please go ahead.

PR #19779, opened 2021-03-05
net-print/cnrdrvcups-lb: remove whitespace

mm1ke:


juippis (approved):
Heh, didn't see this as an issue enough for a commit on its own, but won't reject help. Thanks!

PR #19786, opened 2021-03-05
sys-apps/plocate: new package

AnAverageHuman:
cc @thesamesam


sys-apps/plocate/files/plocate-1.1.5-no-systemd-check.patch
```diff
@@ -0,0 +1,29 @@
+diff -abur a/meson.build plocate-1.1.5/meson.build
+--- a/meson.build	2021-03-05 12:40:39.594922524 -0500
++++ b/meson.build	2021-03-05 12:40:44.654922623 -0500
+@@ -76,17 +76,17 @@
+                configuration: conf_data)
+ install_man(updatedb_conf_man)
+ 
+-if get_option('install_systemd')
```
thesamesam, line 8: Let's drop commented line from patches - we already know what's missing (from the - lines).

sys-apps/plocate/metadata.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
+		<email>jeffrey@icurse.nl</email>
+		<name>Jeffrey Lin</name>
+	</maintainer>
+	<maintainer type="project">
```
thesamesam, line 8: I'll add myself too, If you don't mind. :)

sys-apps/plocate/plocate-1.1.5.ebuild
```diff
@@ -0,0 +1,60 @@
...
+	app-arch/zstd:=
+	!sys-apps/mlocate
+	io-uring? ( sys-libs/liburing:= )
+"
+DEPEND="${RDEPEND}"
+
+BDEPEND="acct-group/locate"
+
+CONFIG_CHECK="~IO_URING"
+ERROR_IO_URING="recommended for USE=io-uring"
```
thesamesam, line 34:
```suggestion
ERROR_IO_URING="required for USE=io-uring"
```

sys-apps/plocate/plocate-1.1.5.ebuild
```diff
@@ -0,0 +1,60 @@
...
+
+PATCHES=(
+	"${FILESDIR}"/plocate-1.1.5-no-systemd-check.patch
+)
+
+src_prepare() {
+	# pretend liburing dep doesn't exist when USE flag off
+	use io-uring || sed -i "s/dependency('liburing/dependency('/" meson.build || die
```
thesamesam, line 42:
```suggestion
	if ! use io-uring ; then
	    sed -i "s/dependency('liburing/dependency('/" meson.build || die
	fi
```
(Just a style thing on my part, not required per se).

sys-apps/plocate/plocate-1.1.5.ebuild
```diff
@@ -0,0 +1,60 @@
...
+
+RDEPEND="
+	acct-group/locate
+	app-arch/zstd:=
+	!sys-apps/mlocate
+	io-uring? ( sys-libs/liburing:= )
+"
+DEPEND="${RDEPEND}"
+
+BDEPEND="acct-group/locate"
```
thesamesam, line 31: DEPEND+RDEPEND, no? (DEPEND given it's needed by the build system, it seems). Don't think we need it for BDEPEND?

sys-apps/plocate/plocate-1.1.5.ebuild
```diff
@@ -0,0 +1,60 @@
...
+
+LICENSE="GPL-2 GPL-2+"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="io-uring"
+
+RDEPEND="
+	acct-group/locate
+	app-arch/zstd:=
+	!sys-apps/mlocate
```
thesamesam, line 26: Let's put the blocker at the end of RDEPEND.

sys-apps/plocate/plocate-1.1.5.ebuild
```diff
@@ -0,0 +1,60 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit linux-info meson systemd
+
+DESCRIPTION="Posting locate is a much faster locate"
+HOMEPAGE="https://plocate.sesse.net/"
+
+if [[ "${PV}" == "9999" ]]; then
```
thesamesam, line 11: But you're not committing a live ebuild. :)

(I'm fine with it if you are, in this case.)

sys-apps/plocate/plocate-1.1.5.ebuild
```diff
@@ -0,0 +1,60 @@
...
+"
+DEPEND="${RDEPEND}"
+
+BDEPEND="acct-group/locate"
+
+CONFIG_CHECK="~IO_URING"
+ERROR_IO_URING="recommended for USE=io-uring"
+
+PATCHES=(
+	"${FILESDIR}"/plocate-1.1.5-no-systemd-check.patch
```
thesamesam, line 37:
```suggestion
	"${FILESDIR}"/${PN}-1.1.5-no-systemd-check.patch
```

thesamesam (changes requested):
Looks good, made some comments.

Could you possibly reach out to the guy who added this to GURU yesterday too?

sys-apps/plocate/metadata.xml
AnAverageHuman: Taken care of that for you now. :)

sys-apps/plocate/plocate-1.1.5.ebuild
AnAverageHuman: I'm actually not sure what happens when liburing is linked in but uring kernel support is not available... that will require some more testing.

sys-apps/plocate/plocate-1.1.5.ebuild
AnAverageHuman: Upstream makes releases often enough that we probably won't need a live ebuild - removed.

AnAverageHuman:
@ArsenArsen: You committed this package to ::guru yesterday, would you like to maintain in ::gentoo?

ArsenArsen:
sure, I'm already watching it for updates anyway.

sys-apps/plocate/plocate-1.1.5.ebuild
thesamesam: I'd say assume it won't work tbh

sys-apps/plocate/plocate-1.1.5.ebuild
```diff
@@ -0,0 +1,54 @@
...
+		sed -i "s/dependency('liburing/dependency('/" meson.build || die
+	fi
+
+	default
+}
+
+src_configure() {
+	local emesonargs=(
+		-Dlocategroup=locate
+		-Dinstall_cron=true
```
ArsenArsen, line 44: this is misleading, out of the box the cron script plocate installs is not very useful, only converting an mlocate-style database into a plocate style one:

```sh
#! /bin/bash

set -e

if [ @PROCESSED_BY_MESON@ = 1 ]; then
        SBINDIR=@sbindir@
        LOCATEGROUP=@groupname@
        DBFILE=@dbfile@
else
        SBINDIR=/usr/local/sbin
        LOCATEGROUP=plocate
        DBFILE=/var/lib/plocate/plocate.db
fi

$SBINDIR/plocate-build /var/lib/mlocate/mlocate.db $DBFILE.new
chgrp $LOCATEGROUP $DBFILE.new
mv $DBFILE.new $DBFILE
```

see: https://gitweb.gentoo.org/repo/proj/guru.git/commit/?id=2aaa3ab6406398c4f531e32b0284dbc7d5c6b5d3

perhaps the cron script in there should utilize is_ac_power or similar, to bring it up to main repo standards

ArsenArsen (changes requested):

sys-apps/plocate/plocate-1.1.5.ebuild
AnAverageHuman: Thanks for pointing that out. I've added most of the files from mlocate's FILESDIR for feature parity, including the default updatedb.conf, and cron script.

Remmy:
Would it be useful to also symlink the manpage from plocate to locate? I notice the man page says locate(1) at the top.

Remmy:
/etc/cron.daily/plocate refers to /usr/bin/updatedb, but the updatedb command gets installed to /usr/sbin/

ArsenArsen:
> Would it be useful to also symlink the manpage from plocate to locate? I notice the man page says locate(1) at the top.

If it's installed under plocate, I'd suggest fixing the man page instead

AnAverageHuman:
Fixed cron, and copied the manpage to locate.1 instead since I'm not sure if symlinks are allowed in the man directory, but `locate` will likely be the popular command.

Remmy:
> If it's installed under plocate, I'd suggest fixing the man page instead

Well, it is installed under both names (plocate and symlinked to it from locate). The current locate most people have installed (mlocate?) also is available under both names if I am not mistaken.

thesamesam:
Let's go for it and see how things are.

thesamesam:
Thanks! I made some touchups:

- https://github.com/gentoo/gentoo/commit/07b37efd20fd2e887b9737a2396fccb4efd190d6
- https://github.com/gentoo/gentoo/commit/7ff6a92ef268c009bb0471b1452bc322bb89b339
- https://github.com/gentoo/gentoo/commit/4486192bc4e88e29325bdd73e0d7b3aa8497f3c2
- https://github.com/gentoo/gentoo/commit/55db4f18ecd2f2060ab6466764c4887c94992838

PR #19788, opened 2021-03-06
sys-libs/compiler-rt-sanitizers: Filter -D_FORTIFY_SOURCE away

lanodan:
Closes: https://bugs.gentoo.org/636604

### DIFF
--- a/sys-libs/compiler-rt-sanitizers/compiler-rt-sanitizers-10.0.1-r1.ebuild
+++ b/sys-libs/compiler-rt-sanitizers/compiler-rt-sanitizers-10.0.1-r1.ebuild
@@ -1,0 +1,152 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7..9} )
+inherit check-reqs cmake flag-o-matic llvm llvm.org python-any-r1
+
+DESCRIPTION="Compiler runtime libraries for clang (sanitizers & xray)"
+HOMEPAGE="https://llvm.org/"
+LLVM_COMPONENTS=( compiler-rt )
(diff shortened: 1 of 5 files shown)


lanodan:
I'm also modifying stabilized ebuilds there, not sure if that's allowed, could remove those.

sys-libs/compiler-rt-sanitizers/compiler-rt-sanitizers-13.0.0.9999.ebuild
```diff
@@ -102,6 +102,7 @@ src_configure() {
 		local -x CC=${CHOST}-clang
 		local -x CXX=${CHOST}-clang++
 		strip-unsupported-flags
+		filter-flags '-D_FORTIFY_SOURCE*'
```
mgorny, line 105: Don't you need it for gcc too, though?

lanodan: After testing that, true, I had that modification for a while, I guess I got confused by `strip-unsupported-flags`.

Fixed it.

mgorny:
@gentoo/toolchain is this a good idea? Sounds a bit like un-hardening to me but it might be necessary anyway for sanitizers.

lanodan:
Any updates on this topic?

mgorny:
```
<@dilfridge> https://github.com/google/sanitizers/issues/247
<@dilfridge> https://reviews.llvm.org/D92521
<@dilfridge> https://reviews.llvm.org/D93877
<@dilfridge> ^ the last one is what you want to backport
```

Could you try that patch instead?

lanodan:
> Could you try that patch instead?

Tried it against 11.1.0 and it works fine. Will change & test the other ebuilds.

mgorny:
Thanks for the testing. I'm going to add the patch into the patchset, test and then bump them.

PR #19790, opened 2021-03-06
app-crypt/tpm2-tools: Silence clang warnings about unsupported flags

salahcoronya:
Although upstream has fixed compilation of this package so it compiles under clang, it produces a ton of warnings about unsupported flags. Thus, if clang is detected, a patch is applied is remove the options clang doesn't understand.


salahcoronya:
This version removes the conditional patch. The build system is patched unconditionally and the clang detection is moved to the build system.

salahcoronya:
Actually, I'm having second thoughts on this. Looking at previous upstream bugs, it looks like they tried to deal with it, but its difficult for a robust solution.. Since it does compile, albeit noisily, I think I'm just going to leave it as. ff someone wants a fix got noise, they can submit a bug/PR upstream

PR #19792, opened 2021-03-06
media-libs/libsdl: BDEPEND

vaukai:
Closes: https://bugs.gentoo.org/617772


chewi:
This actually doesn't play out in your snapshot like it did in the version that was originally reported. The snapshot bundles pkg.m4 but obviously you still need pkg-config installed for it to work. The snapshot just fails to detect PulseAudio and DirectFB, although we don't support the latter any more. It was used for libcaca but strangely isn't any more. So I don't know why you gated it on `X?` here, if anything, it should be `pulseaudio?`.

chewi:
Thanks. I just noticed that nasm/yasm should also be in BDEPEND rather than DEPEND so I fixed that too.

PR #19793, opened 2021-03-06
{profiles,sys-devel/clang}: enable abi_x86_32 clang-runtime & deps; support default unwindlib

telans:
The end goal of these commits is to fix __Unwind linking issues when using `sys-devel/clang[default-compiler-rt]` and not specifying `-unwindlib=libunwind`. See the bug linked to close.
A side effect of `sys-devel/clang` defaulting to using `sys-libs/llvm-libunwind` as the unwindlib is that all `abi_x86_32` multilib builds then require `sys-libs/llvm-libunwind[abi_x86_32]` to build, too.
After speaking to @mgorny it was determined that the best course of action is to enable `abi_x86_32` on all dependencies of `sys-devel/clang-runtime` on multilib profiles.

```
profiles: enable abi_x86_32 clang-runtime & deps
    
    Enable for compatability with -unwindlib=libunwind
    sys-devel/clang[default-compiler-rt] requires these for multilib builds
    
```
```
sys-devel/clang: support default unwindlib
    
    !riscv? ( sys-libs/libunwind ) rather than use masking as it is not keyworded,
     but riscv can still use sys-libs/llvm-libunwind
    
    Closes: https://bugs.gentoo.org/702344
    Closes: https://bugs.gentoo.org/722978
```

### DIFF
--- a/profiles/arch/amd64/package.use
+++ b/profiles/arch/amd64/package.use
@@ -1,5 +1,13 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
+
+# Theo Anderson <telans@posteo.de> (2021-03-06)
+# Enable 32-bit clang runtime to ease -m32 builds with non-GCC runtime
+sys-devel/clang-runtime abi_x86_32
+sys-libs/libcxxabi abi_x86_32
+sys-libs/libcxx abi_x86_32
+sys-libs/libomp abi_x86_32
+sys-libs/llvm-libunwind abi_x86_32
 
 # Mike Gilbert <floppym@gentoo.org> (2016-11-17)
 # Enable both efi and pc by default.
--- a/sys-devel/clang/clang-13.0.0.9999.ebuild
+++ b/sys-devel/clang/clang-13.0.0.9999.ebuild
@@ -24,7 +24,7 @@ LICENSE="Apache-2.0-with-LLVM-exceptions UoI-NCSA MIT"
 SLOT="$(ver_cut 1)"
 KEYWORDS=""
 IUSE="debug default-compiler-rt default-libcxx default-lld
-	doc +static-analyzer test xml kernel_FreeBSD ${ALL_LLVM_TARGETS[*]}"
+	doc llvm-libunwind +static-analyzer test xml kernel_FreeBSD ${ALL_LLVM_TARGETS[*]}"
 REQUIRED_USE="${PYTHON_REQUIRED_USE}
 	|| ( ${ALL_LLVM_TARGETS[*]} )"
 RESTRICT="!test? ( test )"
@@ -52,7 +52,11 @@ RDEPEND="${RDEPEND}
 PDEPEND="
 	sys-devel/clang-common
 	~sys-devel/clang-runtime-${PV}
-	default-compiler-rt? ( =sys-libs/compiler-rt-${PV%_*}* )
+	default-compiler-rt? (
+		=sys-libs/compiler-rt-${PV%_*}*
+		llvm-libunwind? ( sys-libs/llvm-libunwind )
+		!llvm-libunwind? ( sys-libs/libunwind )
+	)
 	default-libcxx? ( >=sys-libs/libcxx-${PV} )
 	default-lld? ( sys-devel/lld )"
 
@@ -262,6 +266,7 @@ multilib_src_configure() {
 		-DCLANG_DEFAULT_CXX_STDLIB=$(usex default-libcxx libc++ "")
 		-DCLANG_DEFAULT_RTLIB=$(usex default-compiler-rt compiler-rt "")
 		-DCLANG_DEFAULT_LINKER=$(usex default-lld lld "")
+		-DCLANG_DEFAULT_UNWINDLIB=$(usex default-compiler-rt libunwind "")
 
 		-DCLANG_ENABLE_ARCMT=$(usex static-analyzer)
 		-DCLANG_ENABLE_STATIC_ANALYZER=$(usex static-analyzer)
--- a/sys-devel/clang/clang-12.0.0-r1.ebuild
+++ b/sys-devel/clang/clang-12.0.0-r1.ebuild
@@ -24,7 +24,7 @@ LICENSE="Apache-2.0-with-LLVM-exceptions UoI-NCSA MIT"
 SLOT="$(ver_cut 1)"
 KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~riscv ~sparc ~x86 ~amd64-linux ~x64-macos"
 IUSE="debug default-compiler-rt default-libcxx default-lld
-	doc +static-analyzer test xml kernel_FreeBSD ${ALL_LLVM_TARGETS[*]}"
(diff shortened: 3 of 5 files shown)


profiles/arch/amd64/package.use.force
```diff
@@ -1,6 +1,14 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Theo Anderson <telans@posteo.de> (2021-03-06)
```
mgorny, line 4: No, not force. Just set them as the default via `package.use`.

telans: I tried that but it doesn't enable them by default:
```
telans@genbox ~ $ emerge -vp1 clang-runtime libomp libcxx libcxxabi llvm-libunwind

Calculating dependencies... done!
[ebuild   R   *] sys-libs/llvm-libunwind-12.0.0_rc2::gentoo  USE="-debug -static-libs -test" ABI_X86="(64) -32* (-x32)" 0 KiB
[ebuild   R   *] sys-libs/libomp-12.0.0_rc2::gentoo  USE="-cuda -hwloc -offload -ompt -test" ABI_X86="(64) -32* (-x32)" 0 KiB
[ebuild   R   *] sys-libs/libcxxabi-12.0.0_rc2::gentoo  USE="libunwind -static-libs -test" ABI_X86="(64) -32* (-x32)" 0 KiB
[ebuild   R   *] sys-libs/libcxx-12.0.0_rc2::gentoo  USE="(libcxxabi) libunwind -static-libs -test" ABI_X86="(64) -32* (-x32)" 0 KiB
[ebuild   R   *] sys-devel/clang-runtime-12.0.0_rc2:12.0.0::gentoo  USE="compiler-rt libcxx openmp sanitize" ABI_X86="(64) -32* (-x32)" 0 KiB

Total: 5 packages (5 reinstalls), Size of downloads: 0 KiB
telans@genbox ~ $ grep -r llvm-libunwind /etc/portage/package.use/
telans@genbox ~ $ grep -r llvm-libunwind /var/db/repos/gentoo/profiles/
/var/db/repos/gentoo/profiles/arch/amd64/no-multilib/package.use.force:sys-libs/llvm-libunwind -abi_x86_32
/var/db/repos/gentoo/profiles/arch/amd64/package.use:sys-libs/llvm-libunwind abi_x86_32
```

mgorny: Show me the code™.

telans:
```
diff --git a/profiles/arch/amd64/package.use b/profiles/arch/amd64/package.use
index d54c8e8baca..81a683e2e0b 100644
--- a/profiles/arch/amd64/package.use
+++ b/profiles/arch/amd64/package.use
@@ -1,6 +1,14 @@
 # Copyright 1999-2019 Gentoo Authors
 
+# Theo Anderson <telans@posteo.de> (2021-03-06)
+# Enable abi_x86_32 Clang libs for compatability with -unwindlib=libunwind and alike
+sys-devel/clang-runtime abi_x86_32
+sys-libs/libcxxabi abi_x86_32
+sys-libs/libcxx abi_x86_32
+sys-libs/libomp abi_x86_32
+sys-libs/llvm-libunwind abi_x86_32
+
 # Mike Gilbert <floppym@gentoo.org> (2016-11-17)
 # Enable both efi and pc by default.
 # https://bugs.gentoo.org/600110
```

mgorny: It works for me. Are you sure you've tested it correctly, in particular that you don't have local overrides?

telans: Yes, in fact x11-misc/nvidia-drivers isn't abi_x86_32 by default for me either (but it's also in that package.use). Maybe it's because of my systemd profile for some reason? `default/linux/amd64/17.1/systemd`

telans: Nevermind, it's because I'm stupid. `-abi_x86_32` in make.conf. Who the hell known why on earth I did that

telans:
I'm going to assume I no longer need the `profiles/arch/amd64/no-multilib/package.use.force` changes now either as I'm not forcing the flags.

profiles/arch/amd64/package.use
```diff
@@ -1,6 +1,14 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Theo Anderson <telans@posteo.de> (2021-03-06)
+# Enable abi_x86_32 Clang libs for compatability with -unwindlib=libunwind and alike
```
mgorny, line 5:
```suggestion
# Enable 32-bit clang runtime to ease -m32 builds with non-GCC runtime
```

sys-devel/clang/clang-13.0.0.9999.ebuild
```diff
@@ -52,7 +52,10 @@ RDEPEND="${RDEPEND}
 PDEPEND="
 	sys-devel/clang-common
 	~sys-devel/clang-runtime-${PV}
-	default-compiler-rt? ( =sys-libs/compiler-rt-${PV%_*}* )
+	default-compiler-rt? (
+		=sys-libs/compiler-rt-${PV%_*}*
+		sys-libs/llvm-libunwind
```
mgorny, line 57:
```suggestion
		|| (
			sys-libs/llvm-libunwind
			sys-libs/libunwind
		)
```

mgorny (commented):
Thanks for your work so far. Given the comments, I'm not sure about the clang part but I think I can commit the profiles patch already since it makes sense to me nevertheless.

telans:
Also depends on `sys-libs/libunwind ~riscv` now with latest changes: https://bugs.gentoo.org/774900

telans:
Seeing as `sys-libs/libunwind` doesn't support riscv: 

- https://bugs.gentoo.org/774900
- https://github.com/libunwind/libunwind/issues/151
- https://github.com/libunwind/libunwind/issues/99

but does support `sys-libs/llvm-libunwind`, instead of introducing another llvm-libunwind USE flag which could be masked in `profiles/arch/riscv` I just put the `sys-libs/libunwind` dep behind `!riscv?`. Is this acceptable?

Otherwise I believe this is ready to be reviewed/merged

Arfrever:
> instead of introducing another llvm-libunwind USE flag which could be masked in `profiles/arch/riscv` I just put the `sys-libs/libunwind` dep behind `!riscv?`. Is this acceptable?

`llvm-libunwind` global USE flag already exists and is forced and masked where it needs to be:
* Forced in RISC-V and SPARC profiles (where only sys-libs/llvm-libunwind is supported)
* Masked in HPPA, IA64 and S390 profiles (where only sys-libs/libunwind is supported)
 (S390 consists of 32-bit S390 and 64-bit S390X. sys-libs/libunwind is supported only in 64-bit S390X.)
* No libunwind implementation is supported in Alpha, M68K and 32-bit S390

telans:
Now using `llvm-libunwind?`

telans:
@mgorny Is it possible to get this merged? (perhaps just the use changes until the rest is reviewed?)

mgorny:
Thanks. I've given it some more thought and I've decided to merge it. It doesn't fix the root of the problem but I suppose it's a good enough hack to keep clang working mostly out of the box with `USE=default-compiler-rt`.

PR #19794, opened 2021-03-06
Prelude-SIEM 5.2: properly set DISTUTILS_USE_SETUPTOOLS variable

ToToL:
Hello

This PR is related to all prelude bugs about "uses a probably incorrect DISTUTILS_USE_SETUPTOOLS value"

I agree with the proposed values so I just update the ebuilds in this way.

Regards


juippis (commented):
Can you please update the copyright headers of these ebuild files? If you commit using `repoman` tool, it'd do this automatically for you.

juippis (approved):
Thanks!

PR #19800, opened 2021-03-06
net-analyzer/netcat: use HTTPs, EAPI 7 bump

mm1ke:
Hi,

This is a simple EAPI7 bump for net-analyzer/netcat. 
The ebuild hast two conditional `sed` statements like `use ipv6 || sed ... ` which is why i can't add `|| die`. What's the prefered way to handle this? Should i create a normal if .. then statement here?
Please review.


ZeroChaos-:
👍

thesamesam:
Yeah, in such cases, I just turn it into a normal if.

thesamesam:
Thank you!

PR #19804, opened 2021-03-07
app-emulation/wine-staging: support llvm-libunwind as unwinder

telans:
See: https://bugs.winehq.org/show_bug.cgi?id=49506


sarnex:
I feel this patch is too much to pull into Gentoo before it's been accepted by wine officially or at least wine-staging. And when they do, we should only update the dependencies.

telans:
It is a bit of a hack but it's not a large patch or anything. I'm not sure if wine-staging takes patch requests but I'll ask to see if it can be put in there. Regardless I think this should be merged as all it does is increase compatability

telans:
I'll take it that this won't be accepted

PR #19810, opened 2021-03-07
app-backup/dar: fix some bugs

ceamac:
USE flags "static" and "static-libs" are already masked for almost 6 months now.  They were blocking adding the "rsync" USE flag because `net-libs/librsync` does not support "static-libs" currently.

The bugs are:
- https://bugs.gentoo.org/739698
- https://bugs.gentoo.org/729150
- https://bugs.gentoo.org/767454
- https://bugs.gentoo.org/686278


thesamesam (changes requested):
Good work. Can you split 'app-backup/dar: add support for threads' into two commits, one which adds the new package, and another which adds the support to dar?

ceamac:
Thank you.  I split it into two commits (that's how I wrote them at the beginning, but I squashed them to have one commit per bug).

About the remaining bug in `dar`:
- https://bugs.gentoo.org/771696 - this one was reported by me and was caused by linking with gold, ld.bfd solved the issue.  I grepped portage for ld.bfd and haven't found ebuilds forcing ld.bfd, except a compiler and the kernel, so I think this bug can be closed, not fixed.

dev-libs/libthreadar/libthreadar-1.3.0.ebuild
```diff
@@ -0,0 +1,42 @@
...
+
+EAPI=7
+
+DESCRIPTION="Threading library used by dar archiver"
+HOMEPAGE="https://sourceforge.net/projects/libthreadar/"
+SRC_URI="mirror://sourceforge/libthreadar/${P}.tar.gz"
+
+LICENSE="LGPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc ~sparc ~x86 ~amd64-linux"
```
thesamesam, line 12: We need to drop to just ~amd64 and request rekeywording of the new version of dar.

app-backup/dar/metadata.xml
```diff
@@ -15,6 +22,7 @@ operations, including with ssh.
 		by 64 bit integers</flag>
 		<flag name="gcrypt">Enables strong encryption support</flag>
 		<flag name="gpg">Enables usage of asymetric crypto algorithms</flag>
+		<flag name="rsync">Enables linking with net-libs/librsync for delta compression support</flag>
```
thesamesam, line 25:
```suggestion
		<flag name="rsync">Enables linking with <pkg>net-libs/librsync</pkg> for delta compression support</flag>
```
(I'll do it)

app-backup/dar/dar-2.6.13-r1.ebuild
```diff
@@ -0,0 +1,90 @@
...
+		$(usex nls '' --disable-nls)
+		$(usex rsync '' --disable-librsync-linking)
+		$(usex threads '' --disable-threadar)
+		$(usex xattr '' --disable-ea-support)
+	)
+
+	# Bug 103741
+	filter-flags -fomit-frame-pointer
+
+	myconf+=( --disable-dar-static --disable-static )
```
thesamesam, line 74: I don't think this belongs on its own.

app-backup/dar/dar-2.6.13-r1.ebuild
```diff
@@ -0,0 +1,90 @@
...
+	# Do _not_ use $(use_enable) until you have verified that the
+	# logic has been fixed by upstream.
+	local myconf=(
+		--disable-python-binding
+		--disable-upx
+		$(usex curl '' --disable-libcurl-linking)
+		$(usex dar32 --enable-mode=32 '')
+		$(usex dar64 --enable-mode=64 '')
+		$(usex doc '' --disable-build-html)
+		#$(usex examples --enable-examples '')
```
thesamesam, line 61: Why?

app-backup/dar/dar-2.6.13-r1.ebuild
```diff
@@ -0,0 +1,90 @@
...
+
+DEPEND="
+	${RDEPEND}
+"
+BDEPEND="
+	nls? (
+		sys-devel/gettext
+		virtual/libintl
+	)
+	doc? ( app-doc/doxygen )
```
thesamesam, line 41: I'll sort BDEPEND.

app-backup/dar/dar-2.6.13-r1.ebuild
```diff
@@ -0,0 +1,90 @@
...
+SRC_URI="mirror://sourceforge/dar/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc ~sparc ~x86 ~amd64-linux"
+IUSE="curl dar32 dar64 doc gcrypt gpg lzo nls rsync threads xattr"
+
+RESTRICT="test" # need to be run as root
+
+RDEPEND="
```
thesamesam, line 19: I'll sort RDEPEND.

app-backup/dar/dar-2.6.13-r1.ebuild
ceamac: The ebuild for 2.6.13 had this:
```
	if ! use static ; then
		myconf+=( --disable-dar-static )
		if ! use static-libs ; then
			myconf+=( --disable-static )
		fi
	fi
```
So when I removed use static and use static-libs, I simply resolved the if.  Maybe put it on two lines?  I think the way configure is written, all not disabled is enabled by default

dev-libs/libthreadar/libthreadar-1.3.0.ebuild
```diff
@@ -0,0 +1,42 @@
...
+
+DEPEND=""
```
thesamesam, line 14: Not needed - see [here](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-620007.3)

dev-libs/libthreadar/metadata.xml
```diff
@@ -0,0 +1,19 @@
...
+		<email>ceamac.paragon@gmail.com</email>
+		<name>Viorel Munteanu</name>
+	</maintainer>
+	<maintainer type="project">
+		<email>proxy-maint@gentoo.org</email>
+		<name>Proxy Maintainers</name>
+	</maintainer>
+	<longdescription>
+Libthreadar is a C++ library providing an abstracted set of C++ *classes* to manipulate threads in a very simple and efficient way from your C++ code.
+It also handles exceptions thrown from a thread and propagated to another one, when the later is calling the thread::join() method. This let one manage exceptions as simply as it is in C++ single threaded context.
```
dwfreed, line 14:
```suggestion
It also handles exceptions thrown from a thread and propagated to another one, when the latter is calling the thread::join() method. This lets one manage exceptions as simply as it is in C++ single threaded context.
```

app-backup/dar/metadata.xml
```diff
@@ -15,6 +22,7 @@ operations, including with ssh.
 		by 64 bit integers</flag>
 		<flag name="gcrypt">Enables strong encryption support</flag>
 		<flag name="gpg">Enables usage of asymetric crypto algorithms</flag>
```
dwfreed, line 24:
```suggestion
		<flag name="gpg">Enables usage of asymmetric encryption algorithms</flag>
```

dwfreed (changes requested):

app-backup/dar/dar-2.6.13-r1.ebuild
```diff
@@ -0,0 +1,90 @@
...
+}
+
+src_install() {
+	emake DESTDIR="${D}" pkgdatadir="${EPREFIX}"/usr/share/doc/${PF}/html install
+
+	local DOCS=( AUTHORS ChangeLog NEWS README THANKS TODO )
+	einstalldocs
+
+	find "${ED}" -name '*.la' -delete || die
+	find "${ED}" -name '*.a' -delete || die
```
juippis, line 86:
```suggestion
	find "${ED}" \( -name "*.a" -o -name "*.la" \) -delete || die
```

app-backup/dar/dar-2.6.13-r1.ebuild
```diff
@@ -0,0 +1,90 @@
...
+
+	# Bug 729150
+	(cd "${ED}/usr/share/doc/${PF}/html/samples" && mkdir MyBackup && tar zxf MyBackup.sh.tar.gz -C MyBackup && rm MyBackup.sh.tar.gz) || die
```
juippis, line 89: What's up with this monster? How can you tell which part fails now? Actually can you walk me through what's happening here, because there's most likely a better way to deal with it. 

If you look at *Makefile.in* file and grep for *gzip* you most likely start to see where this is coming from. We'd prefer a build-system patch. I believe simply emptying out `DIST_TARGETS` will fix it, but please do test with 
https://wiki.gentoo.org/wiki/Package_testing#make.conf

app-backup/dar/dar-2.6.13-r1.ebuild
```diff
@@ -0,0 +1,90 @@
...
+
+	myconf+=( --disable-dar-static --disable-static )
+
+	econf "${myconf[@]}"
+}
+
+src_install() {
+	emake DESTDIR="${D}" pkgdatadir="${EPREFIX}"/usr/share/doc/${PF}/html install
+
+	local DOCS=( AUTHORS ChangeLog NEWS README THANKS TODO )
```
juippis, line 82: Please move this to global scope, between `REQUIRED_USE` and `src_configure` for example.

app-backup/dar/dar-2.6.13-r1.ebuild
ceamac: It doesn't come from there.  There is a tar.gz file included in the distribution, please check `dar-2.6.13/doc/samples/MyBackup.sh.tar.gz`, it is not built by the build system, only copied as is.  `DIST_TARGETS` applies to `make dist`, not to `make install`.

I unpacked it because the bug is an automated check.  Other ways to fix it would be:
- patch the build system to unpack it at `make install` time
- or simply remove it from installation

How do you suggest I fix this bug?

juippis: Hmm I see. Can you try calling `docompress -x` on it in `src_install`, which ignores the package-manager compression for it?

Is the package any relevant for the user, does it make sense installing it?

ceamac: After a bit of searching it seems unpacking it is not a solution, because that directory contains this page: http://dar.linux.free.fr/doc/samples/index.html

Unpacking `MyBackup.sh.tar.gz` would break the link in `index.html` anyway.  So if I remove it and someone really needs it, they can download it.  Is this acceptable?

ceamac: Also I'll check if this directory is related to that "examples" USE flag from earlier.

juippis: Yes please, sounds good.

app-backup/dar/dar-2.6.13-r1.ebuild
ceamac: I checked.  `--disable-examples` stops with an error, `--enable-examples` does nothing (same files are installed in docs).  So I removed it.

ceamac:
I think (I hope) I've solved everything.
I masked "threads" USE flag on *ppc*, *sparc* and *x86*, and I've got a conflict with master after pushing.  I hope I fixed it ok.
Please check and review again.

dev-libs/libthreadar/libthreadar-1.3.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+	# configure.ac is totally funked up regarding the AC_ARG_ENABLE
+	# logic.
+	# For example "--enable-dar-static" causes configure to DISABLE
+	# static builds of dar.
+	# Do _not_ use $(use_enable) until you have verified that the
+	# logic has been fixed by upstream.
+
+	local myconf=( --disable-static )
+
+	econf "${myconf[@]}"
```
juippis, line 28: You could just call `econf --disable-static` here, but I don't mind this.

juippis (approved):
Thanks! 

Regarding the co-authors, I had to read https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin if it's legal to include them. I'd say yes for now.

PR #19812, opened 2021-03-07
nvidia-drivers

ionenwks:
~WIP mostly in the sense that it needs more runtime testing before being added, but PR now so it can be reviewed.~

See ~460.56-r1~ 460.67's commit message for details.

See https://github.com/gentoo/api-gentoo-org/pull/379 for acct-user/nvpd uid

This drops IUSE=uvm/compat so the following will be needed as well:
- ~dev-util/nvidia-cuda-sdk: uvm(+),~ or remove check entirely, not a big issue.
- ~dev-util/nvidia-cuda-toolkit: uvm(+),~ see also [bug #663834](https://bugs.gentoo.org/663834) that should be possible with the new ebuilds
- ~x11-misc/primus-0.2-r0: remove old, [bug #774111](https://bugs.gentoo.org/774111)~
Thankfully ffmpeg didn't go forward with depending on [uvm].

There's also a mask for uvm on x86, but that's now handled internally (clearly it's never coming back).

After old ebuilds gone, most files/* will be removable (renamed for changes rather than -r1's, hoping for filesdir to be a bit more consistent).


x11-drivers/nvidia-drivers/nvidia-drivers-390.141-r1.ebuild
```diff
@@ -0,0 +1,377 @@
...
+	BUILD_PARAMS='V=1 NV_VERBOSE=1 SYSSRC="${KERNEL_DIR}" IGNORE_CC_MISMATCH=yes'
+	use x86 && BUILD_PARAMS+=' ARCH=i386' # needed for recognition
+	MODULE_NAMES="
+		nvidia(video:kernel)
+		nvidia-drm(video:kernel)
+		nvidia-modeset(video:kernel)"
+	use amd64 && MODULE_NAMES+=" nvidia-uvm(video:kernel)" # no x86 support
+	linux-mod_pkg_setup
+
+	if [[ ${MERGE_TYPE} == source ]] && kernel_is gt ${NV_KERNEL_MAX/./ }; then
```
SoapGentoo, line 109: `[[ ${MERGE_TYPE} != binary ]]` is the idiomatic check we use in eclasses and everywhere else

ionenwks:
added sed to avoid misdetecting DRM support wrt [bug #603818](https://bugs.gentoo.org/603818)

ionenwks:
Removed IUSE=wayland, this library was never useful on its own and supporting EGLStreams sound better handled in a separate egl-wayland as the x11-wm/mutter ebuild commented on.

x11-drivers/nvidia-drivers/files/nvidia-430.conf
```diff
@@ -17,4 +17,4 @@ options nvidia \
 	NVreg_DeviceFileUID=0 \
 	NVreg_ModifyDeviceFiles=1 \
 	NVreg_PreserveVideoMemoryAllocations=1 \
-	NVreg_TemporaryFilePath=/tmp
+	NVreg_TemporaryFilePath=/var/tmp
```
f0o, line 20: why was the temporary directory moved for it?

x11-drivers/nvidia-drivers/metadata.xml
```diff
@@ -4,21 +4,23 @@
...
+		<name>Ionen Wolkens</name>
+	</maintainer>
 	<use>
-		<flag name="X">Install the X.org driver, OpenGL libraries, XvMC libraries, and VDPAU libraries</flag>
 		<flag name="compat">Install non-GLVND libGL for backwards compatibility</flag>
-		<flag name="driver">Install the kernel driver module</flag>
+		<flag name="driver">Install kernel driver modules</flag>
 		<flag name="kms">Enable support for kernel mode setting (KMS)</flag>
 		<flag name="tools">Install additional tools such as nvidia-settings</flag>
 		<flag name="uvm">Install the Unified Memory kernel module (nvidia-uvm) for sharing memory between CPU and GPU in CUDA programs</flag>
```
f0o, line 16: if UVM is now always included, shouldn't this flag be removed?

ionenwks: Can't remove until cleanup old, those versions are stable right now.

x11-drivers/nvidia-drivers/files/nvidia-430.conf
ionenwks: It's used for hibernation storage, /tmp typically being tmpfs felt like an unreasonable default.

x11-drivers/nvidia-drivers/metadata.xml
ionenwks: Yes, I mentioned this in my opening comment. The CI is failing because of this too.

ionenwks:
Added old primus removal and nvidia-cuda-* bare minimum updates as I'd rather not mess with those beyond, i.e. only uvm(+) with revbump, can't say I like the idea of revbumping here but I've little faith in dynamic deps. Just tell me if anything would rather have handled differently.

Beside that I think these drivers changes are mostly ready.

ionenwks:
Replaced 460.56-r1 by 460.67 version bump (trivial bugfix release that required no changes).

PR #19815, opened 2021-03-07
dev-lang/rust: patchelf & prefixify in stage0, for gentoo prefix

littlewu2508:
Also delete the dependency of llvm-9 for rust-1.46, to resolve the
dependency issue discovered by repoman

Bug: https://bugs.gentoo.org/739574
Closes: https://bugs.gentoo.org/682370
Signed-off-by: YiyangWu <xgreenlandforwyy@gmail.com>

### DIFF
--- a/dev-lang/rust/rust-1.46.0.ebuild
+++ b/dev-lang/rust/rust-1.46.0.ebuild
@@ -1,11 +1,11 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
 PYTHON_COMPAT=( python3_{7..9} )
 
-inherit bash-completion-r1 check-reqs estack flag-o-matic llvm multiprocessing multilib-build python-any-r1 rust-toolchain toolchain-funcs
+inherit prefix bash-completion-r1 check-reqs estack flag-o-matic llvm multiprocessing multilib-build python-any-r1 rust-toolchain toolchain-funcs
 
 if [[ ${PV} = *beta* ]]; then
 	betaver=${PV//*beta}
@@ -49,10 +49,7 @@ IUSE="clippy cpu_flags_x86_sse2 debug doc libressl miri nightly parallel-compile
 # 2. Update the := to specify *max* version, e.g. < 11.
 # 3. Specify LLVM_MAX_SLOT, e.g. 10.
 LLVM_DEPEND="
-	|| (
-		sys-devel/llvm:10[${LLVM_TARGET_USEDEPS// /,}]
-		sys-devel/llvm:9[${LLVM_TARGET_USEDEPS// /,}]
-	)
+	sys-devel/llvm:10[${LLVM_TARGET_USEDEPS// /,}]
 	<sys-devel/llvm-11:=
 	wasm? ( sys-devel/lld )
 "
@@ -61,6 +58,7 @@ LLVM_MAX_SLOT=10
 BOOTSTRAP_DEPEND="|| ( >=dev-lang/rust-1.$(($(ver_cut 2) - 1)) >=dev-lang/rust-bin-1.$(($(ver_cut 2) - 1)) )"
 
 BDEPEND="${PYTHON_DEPS}
+	prefix? ( !system-bootstrap? ( dev-util/patchelf ) )
 	app-eselect/eselect-rust
 	|| (
 		>=sys-devel/gcc-4.7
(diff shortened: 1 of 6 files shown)


littlewu2508:
@gyakovlev @heroxbd Any progress in reviewing this PR? I heard other pepole talking about this bug in gentoo prefix.

In bugzila https://bugs.gentoo.org/attachment.cgi?id=689850&action=diff provides a similar fix, we can combine mine with this together.

dev-lang/rust/rust-1.46.0.ebuild
```diff
@@ -49,10 +49,7 @@ IUSE="clippy cpu_flags_x86_sse2 debug doc libressl miri nightly parallel-compile
 # 2. Update the := to specify *max* version, e.g. < 11.
 # 3. Specify LLVM_MAX_SLOT, e.g. 10.
 LLVM_DEPEND="
-	|| (
-		sys-devel/llvm:10[${LLVM_TARGET_USEDEPS// /,}]
-		sys-devel/llvm:9[${LLVM_TARGET_USEDEPS// /,}]
-	)
```
gyakovlev, line 55: please don't touch dependencies in stable ebuilds (ignore warning)

dev-lang/rust/rust-1.46.0.ebuild
```diff
@@ -198,6 +196,31 @@ src_prepare() {
 
 		"${WORKDIR}/${rust_stage0}"/install.sh --disable-ldconfig \
 			--destdir="${rust_stage0_root}" --prefix=/ || die
+
+		if use prefix; then
+			interpreter=`ldconfig -p | grep ld-linux | awk '{print $NF}' || die "Cannot find your ld.so, why?"`
```
gyakovlev, line 201: we don't use backticks in ebuilds, please use '$(...)' form.
also that pipe does not do what you thing it does, so split it into several steps and don't pipe more than once in single substitution.

also make variables `local`

gyakovlev: and remove, `why?` part, it's confusing and does not help at all.

dev-lang/rust/rust-1.46.0.ebuild
```diff
@@ -198,6 +196,31 @@ src_prepare() {
...
+			ebegin "Changing interpreter to $interpreter for Gentoo prefix"
+			for f in `ls -Ud ${rust_stage0_root}/bin/*  || die "Cannot find binary of rust_stage0, why?"`; do
```
gyakovlev, line 203: don't use ls as input, use `find -exec` instead

also make variables local. and ditto for backticks.

gyakovlev (changes requested):
this needs code cleanup.
I've left some comments, but more is needed, will get back later

dev-lang/rust/rust-1.46.0.ebuild
```diff
@@ -1,11 +1,11 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
 PYTHON_COMPAT=( python3_{7..9} )
 
-inherit bash-completion-r1 check-reqs estack flag-o-matic llvm multiprocessing multilib-build python-any-r1 rust-toolchain toolchain-funcs
+inherit prefix bash-completion-r1 check-reqs estack flag-o-matic llvm multiprocessing multilib-build python-any-r1 rust-toolchain toolchain-funcs
```
gyakovlev, line 8: prefix eclass should come in alphabetical order, so add it right before python-any-r1

dev-lang/rust/rust-1.46.0.ebuild
gyakovlev: `ld-linux` is interpreter name only on amd64 and arm64, on other platforms in may be different.

for instance, on ppc64 it's `ld64.so.2`, so that solution is not portable at all.

littlewu2508:
@gyakovlev Thanks for pointing out those non-standard syntax. I've corrected them in a new commit. However I did not figure out a way to use `find -exec` to apply a bash function or a series of command, so I use for loop on `find`'s output instead.

gyakovlev:
> @gyakovlev Thanks for pointing out those non-standard syntax. I've corrected them in a new commit. However I did not figure out a way to use `find -exec` to apply a bash function or a series of command, so I use for loop on `find`'s output instead.

thanks, that looks better.
I'll test it in a while and will merge and fix or maybe ask for more fixes.

as for find, there are plenty of examples in `eclass/*`, it'll need to use `-print0` or `-exec`, so you don't iterate over non-NULL terminated args.

it should start with something like 
`while IFS= read -r -d '' f; do`
and end with 
`done < <(find -H "${path}" -type f -print0 || die)`

also all variables should be in `${}` form, even positional args like ${1}.

heroxbd:
It is easier to patchelf `dev-lang/rust-bin` as in https://bugs.gentoo.org/739574#c8, and enforce USE=system-boostrap in dev-lang/rust.

Hacking the stage0 compiler looks more fragile to me.

littlewu2508:
> 
> 
> It is easier to patchelf `dev-lang/rust-bin` as in https://bugs.gentoo.org/739574#c8, and enforce `USE=system-boostrap` in Prefix profiles for `dev-lang/rust`.
> 
> Hacking the stage0 compiler looks more fragile to me.

Agreed. Although patchelf on stage0 compiler works, it would be more beneficial hacking rust-bin, because there are people directly use rust-bin.

gyakovlev:
Indeed, let's go that way instead. fixing 2 issues with 1 change is always nice =)

littlewu2508:
I reverted commits on patching stage0 in rust, and add a new commit about system-bootstrap.

For rust-bin I created a new PR : https://github.com/gentoo/gentoo/pull/20073

gyakovlev:
I think REQUIRED_USE is not required ( sorry for the pun ).
we usually do such changes via package.use.force in profile.
I'll merge this or similar one later today with https://github.com/gentoo/gentoo/pull/20073

gyakovlev:
done =) lmk  if something needs more attention or works not as expected. and thanks for working on it and pushing it. rust is a beast to maintain and help is appreciated.

heroxbd:
> done =) lmk if something needs more attention or works not as expected. and thanks for working on it and pushing it. rust is a beast to maintain and help is appreciated.

Thanks. I will take care of the prefix profiles when necessary.

gyakovlev:
> > done =) lmk if something needs more attention or works not as expected. and thanks for working on it and pushing it. rust is a beast to maintain and help is appreciated.
> 
> Thanks. I will take care of the prefix profiles when necessary.

I've already added `use.force` entry to prefix feature in https://github.com/gentoo/gentoo/commit/18b519726110646e816e2b45e73ac39e3c9ee2bc so it should be ok.

heroxbd:
Georgy Yakovlev ***@***.***> writes:

> I've already added use.force entry to prefix feature in 18b5197 so it should be ok.

Got you.  Thanks again.

PR #19818, opened 2021-03-07
net-p2p/eiskaltdcpp: Strip ie locale when USE=gtk

klondi:
Eiskaltdc++ is planning to drop the GTK+ based GUI in the next version and has asked translators not to consider it when creating new translations.

As a result of this, the ie translation does not includes the locale for the GTK library which results in an error when the ebuild tries to compile it.

To address this, we filter the ie locale from the LINGUAS variable whenever the gtk USE flag is set, to address the issue and still allow using GTK if needed.

### DIFF
--- a/net-p2p/eiskaltdcpp/eiskaltdcpp-2.4.2.ebuild
+++ b/net-p2p/eiskaltdcpp/eiskaltdcpp-2.4.2.ebuild
@@ -104,6 +104,7 @@ src_prepare() {
 }
 
 src_configure() {
+	use gtk && strip-linguas -i eiskaltdcpp-gtk/po/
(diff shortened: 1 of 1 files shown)


vaukai:
Could you pls move the [SRC_URI stuff](https://github.com/gentoo/gentoo/blob/f00b96687b7cc0d478fd32e11d21a3a6a9a28480/net-p2p/eiskaltdcpp/eiskaltdcpp-2.4.2.ebuild#L30-L36) further up to where it belongs? And remove lines with =""

juippis (approved):
Seems to work for me now with `USE="gtk"`.

PR #19824, opened 2021-03-08
Net irc/atheme services 7.2.11

clinew:


clinew:
I plan to request stabilization in 5 weeks, so future upgrades will keep the old version by default while waiting out the pre-stabilization period.

PR #19828, opened 2021-03-08
*/*: directly depend on sys-apps/util-linux, remove virtual/eject from tree

mm1ke:
Hi,

This PR adapts all remaining ebuilds which depend on `virtual/eject` to switch to `sys-apps/util-linux`
As already mentioned in the bug, `sys-block/eject-bsd` got removed in 2015 which leaves `virtual/eject` with only one dependency:
```
RDEPEND="|| ( >=sys-apps/util-linux-2.22 sys-block/eject-bsd )"
```

I'm not sure if it's ok to remove the virtual directly. Please let me know if I should only mask it first.


SoapGentoo:
@mm1ke can you do the following:

1. every ebuild with a changed `RDEPEND`, revbump it, keeping the stable keywords
2. check that new ebuilds are also caught
3. just remove the virtual

once you've rebased and checked it, I will merge it. I'm aware this is against the guidance in the devmanual, but I don't want to wait years before we remove the virtual.

mm1ke:
@SoapGentoo I did as you wished:

* revbump all ebuilds with changed `RDEPEND` (with `git mv`) and kept the stable keywords. (actually most of ebuilds were affected..).
* I've checked the tree again but couldn't find any new package (checked with `grep -R "virtual/eject" *`)

Please review

thesamesam (approved):

SoapGentoo (approved):

PR #19829, opened 2021-03-08
add arm64 support for terraform

dme86:
enable arm64 support for terraform.
tested.


Zlogene:
please add Signed-off-by and do not add stable arm64 keywords yet, they all must be just `~arm64`

Zlogene:
@dme86 also please squash your two commits into the single one with `git rebase -i`

PR #19831, opened 2021-03-08
dev-lang/ocaml: fixing patching, remove unused eutils eclass

mm1ke:
Closes: https://bugs.gentoo.org/774867

`epatch` doesn't work in EAPI7 and the eclass is banned, which is why we get following error: 
```
/var/tmp/portage/dev-lang/ocaml-4.05.0-r3/temp/environment: line 1639: epatch: command not found
```
I've updated the ebuild using `eapply` and removed the unused `eutils` eclass.


thesamesam:
Good catch, thanks!

PR #19835, opened 2021-03-08
dev-libs/libxml2: Add upstream patch for xmllint

yetamrra:
This fixes an out-of-bounds read in xmllint when built with icu.  See
CVE-2020-24977 and https://gitlab.gnome.org/GNOME/libxml2/-/issues/178
for more info.

Signed-off-by: Benjamin Gordon <bmgordon@chromium.org>


yetamrra:
Thanks!

thesamesam:
Thanks! It'd fallen off my radar given the maintainer's response upstream, but we should have it patched.

PR #19839, opened 2021-03-09
Update metadata.xml

ScardracS:


thesamesam:
Hi, can you use the same PR for all your changes? Thanks!

Note that we normally wait to let people make a few contributions before adopting a package, btw.

PR #19842, opened 2021-03-09
net-analyzer/zabbix: fix eclass usage

mm1ke:
Closes: https://bugs.gentoo.org/775032

As mentioned in the bug the ebuilds have following code:

```
 zabbix_homedir=$(egethome zabbix)
 if [ -n "${zabbix_homedir}" ] && \
    [ "${zabbix_homedir}" != "/var/lib/zabbix/home" ]; then
       ewarn
       ewarn "The user 'zabbix' should have his homedir changed"
       ...
```

`egethome` is a function from the `user-info.eclass` which doesn't get directly or implicit inherited by the ebuild. I've fixed this and also removed the `flag-o-matic` eclass since this one isn't used at all.


fordfrog:
@mm1ke nice catch, thanks!

PR #19849, opened 2021-03-09
dev-cpp/antlr-cpp: depend on app-arch/unzip

mm1ke:

I haven't found any reason why this package should depend on `app-arch/zip` so i think this should be `app-arch/unzip` - which is required to unzip the src files.


fordfrog:
@mm1ke thanks! you might also want to use `pkgcheck scan` to see other issues for ebuilds working on. this ebuild also needed switch from `cmake-utils` to `cmake` (d378d3a2af7ca08a915eb85b45d33c42eab1b814).

PR #19856, opened 2021-03-10
sys-fs/loop-aes: Version bump, clean up old

hlein:
Per upstream, this version adds support for 5.11 kernels.


juippis (commented):
Please split adding a new version and removing old into 2 separate commits.

hlein:
> Please split adding a new version and removing old into 2 separate commits.

Done, thanks.

juippis (approved):
Thanks!

PR #19858, opened 2021-03-10
dev-libs/opensc: ship config snippet for app-crypt/p11-kit

henning-schild:
In order to use smartcards with app-crypt/p11-kit, a module needs to be
loaded by it. Doing this makes sure "p11tool --list-tokens" will find
smartcards without further manual steps.
The method is inspired by Fedora and OpenSuse.

### DIFF
--- a/dev-libs/opensc/files/opensc.module
+++ b/dev-libs/opensc/files/opensc.module
@@ -1,0 +1,8 @@
+# This file describes how to load the opensc module
+# See: http://p11-glue.freedesktop.org/doc/p11-kit/config.html
+
+# This is a relative path, which means it will be loaded from
+# the p11-kit default path which is usually $(libdir)/pkcs11.
+# Doing it this way allows for packagers to package opensc for
+# 32-bit and 64-bit and make them parallel installable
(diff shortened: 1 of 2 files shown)


henning-schild:
Please let me know if this needs a bug

henning-schild:
OpenSuse
https://build.opensuse.org/package/view_file/home:pzb:smartcard/opensc/opensc.module?expand=1

Fedora
https://build.opensuse.org/package/view_file/home:pzb:smartcard/opensc/opensc.module?expand=1

I took the OpenSuse way, placing it in `/etc`, so if we are really wrong with that default, users can disable it again. Fedora goes of /usr. Then there is the question of which module, i find the "onepin" to do the job just fine and being a good default. Since we are in /etc people can remove "onepin-" for whatever reason ... not sure i ever really understood the difference between the two modules.

henning-schild:
In fact this one is worth a try with upstream opensc. But since i had some trouble with them it is probably best to do it here first. Once that is done i will try once again, at which point we will have at least 3 distros doing downstream fixing, which might be convincing.

PR #19862, opened 2021-03-10
media-video/ushare: remove enewuser, update LICENSE

mm1ke:

Hi,

Since the packages already uses new GLEP81 users, `enewuser` isn't required anymore. In fact it doesn't even work right now because the ebuilds doesn't inherit the `user` eclass.
I've also checked some source files and updated the License to `GPL-2+`.


benkohler:
Thanks for catching this

PR #19864, opened 2021-03-10
sci-libs/tensorflow-estimator: cleanup eclasses, add missing app-arch/unzip dep

mm1ke:
Hi,

I've removed some unused eclasses from both ebuilds and added the missing `app-arch/unzip` dependency to 2.3.0.

### DIFF
--- a/sci-libs/tensorflow-estimator/tensorflow-estimator-2.3.0.ebuild
+++ b/sci-libs/tensorflow-estimator/tensorflow-estimator-2.3.0.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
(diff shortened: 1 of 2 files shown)


sci-libs/tensorflow-estimator/tensorflow-estimator-2.3.0.ebuild
```diff
@@ -8,15 +8,14 @@ MY_PN="estimator"
...
-DESCRIPTION="A high-level TensorFlow API that greatly simplifies machine learning programming"
+DESCRIPTION="High-level TensorFlow API that greatly simplifies machine learning programming"
 HOMEPAGE="https://www.tensorflow.org/"
 
 LICENSE="Apache-2.0"
 SLOT="0"
 KEYWORDS="~amd64"
-IUSE=""
 
 bazel_external_uris="
```
vaukai, line 20: $ pkgcheck scan --net -c SrcUriCheck  sci-libs/tensorflow-estimator
sci-libs/tensorflow-estimator
  TarballAvailable:

mm1ke: thanks for the hint. 
However, do i need to make an `r1` if i change the source files? Because in that case i think it's not worth for the older ebuild.

AndrewAmmerlaan: > However, do i need to make an `r1` if i change the source files? Because in that case i think it's not worth for the older ebuild.

Only if this change causes a change in things that are installed to the system (which is most likely not the case here)

PR #19865, opened 2021-03-10
[DRAFT] rstudio bump to 1.4.1106

ezzieyguywuf:
This will remain in `DRAFT` until it is ready.

For now, it is far enough along that code review is welcome, and also, it
compiles!

Here's what's still left to do:

- [ ] Check what features from previous ebuilds I'm still missing that are
  needed, notably `USE=dedicated`. Consider bug #736758
- [ ] Try to install it, I have a feeling I'll need some of those missing
  dependencies from the previous ebuild
- [ ] Try to add `USE=pandoc`, per #531996


dev-db/soci/soci-4.0.1.ebuild
```diff
@@ -0,0 +1,31 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
mm1ke, line 4: empty line after `EAPI`

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,80 @@
...
+EAPI=7
+
+inherit cmake qmake-utils
+
+QT_VER=5.12.6
+QT_SLOT=5
+
+DESCRIPTION="IDE for the R language"
+HOMEPAGE="
+	http://www.rstudio.org
```
mm1ke, line 13: i suggest here: `https://rstudio.com`

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,80 @@
...
+	>=dev-qt/qtsql-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtsvg-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtwebchannel-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtwebengine-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtxml-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtxmlpatterns-${QT_VER}:${QT_SLOT}
+"
+DEPEND="
+	${RDEPEND}
+	app-arch/unzip
```
mm1ke, line 53: i think this should go into `BDEPEND`

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,80 @@
...
+	"${FILESDIR}/${PN}-1.4.1106-add-missing-boost-namespace-import.patch"
+	"${FILESDIR}/${PN}-1.4.1106-add-missing-boost-bind-import.patch"
+	"${FILESDIR}/${PN}-1.4.1106-remove-hardcoded-internal-mathjax-and-pandoc-deps.patch"
+	"${FILESDIR}/${PN}-1.4.1106-soci-cmake-find_library.patch"
+)
+
+src_unpack() {
+	unpack ${P}.tar.gz
+	cd "${S}" || die
+	mkdir -p dependencies/common/dictionaries
```
mm1ke, line 67: missing `|| die`

ezzieyguywuf: Hey man, I don't know if we've met but thank you so so much for all of your code reviews! They are greatly appreciated!

mm1ke: no problem :) glad i can help.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+
+# src_configure...-DRSTUDIO_DISABLE_CRASHPAD
+src_configure() {
+	export RSTUDIO_VERSION_MAJOR=$(ver_cut 1)
+	export RSTUDIO_VERSION_MINOR=$(ver_cut 2)
+	export RSTUDIO_VERSION_PATCH=$(ver_cut 3)
+
+	local mycmakeargs=(
+		-DQT_QMAKE_EXECUTABLE="$(qt5_get_bindir)/qmake"
+		-DRSTUDIO_INSTALL_SUPPORTING=${EPREFIX}/usr/share/${PN}
```
thesamesam, line 91:
```suggestion
		-DRSTUDIO_INSTALL_SUPPORTING="${EPREFIX}"/usr/share/${PN}
```

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+		-DRSTUDIO_UNIT_TESTS_DISABLED=On
```
thesamesam, line 92:
```suggestion
		-DRSTUDIO_UNIT_TESTS_DISABLED=ON
```
or
```suggestion
		-DRSTUDIO_UNIT_TESTS_DISABLED=YES
```
if it works, for consistency.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+		-DRSTUDIO_USE_SYSTEM_BOOST=Yes
```
thesamesam, line 93: See above re ON/YES.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+	"${FILESDIR}/${PN}-1.4.1106-remove-hardcoded-internal-mathjax-and-pandoc-deps.patch"
+	"${FILESDIR}/${PN}-1.4.1106-remove-mathjax-internal-dep-check.patch"
+	"${FILESDIR}/${PN}-1.4.1106-soci-cmake-find_library.patch"
+)
+
+src_unpack() {
+	unpack ${P}.tar.gz
+	cd "${S}" || die
+	mkdir -p dependencies/common/dictionaries || die
+	unzip -qd dependencies/common/dictionaries \
```
thesamesam, line 70: Can we call unpack?

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+	>=dev-qt/qtsvg-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtwebchannel-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtwebengine-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtxml-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtxmlpatterns-${QT_VER}:${QT_SLOT}
+"
+DEPEND="${RDEPEND}"
+BDEPEND="
+	app-arch/unzip
+	dev-vcs/git
```
thesamesam, line 54: We really need git to build?

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+# IUSE="dedicated libressl server"
+
+# ADD?
+# libsqlite3-dev (this is the ubuntu thingy) → dev-db/postgres
+# libpq-dev → dev-db/sqlite
+# yaml-cpp something?
+#
+# check docker/jenkins
+RDEPEND="
+	dev-db/postgresql
```
thesamesam, line 32: Missing slot op?

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+QT_VER=5.12.6
+QT_SLOT=5
+
+DESCRIPTION="IDE for the R language"
+HOMEPAGE="
+	https://rstudio.com
+	https://github.com/rstudio/rstudio/"
+SRC_URI="
+	https://github.com/rstudio/rstudio/archive/v${PV}.tar.gz -> ${P}.tar.gz
+	https://s3.amazonaws.com/rstudio-dictionaries/core-dictionaries.zip
```
thesamesam, line 17: I'd suggest versioning this somehow in case it gets changed in future, in case we have a few versions in tree (and so the file on user machines is labelled), e.g.

```suggestion
	https://s3.amazonaws.com/rstudio-dictionaries/core-dictionaries -> core-dictionaries-1.zip
```

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+"
+
+LICENSE="AGPL-3"
+SLOT="0"
+KEYWORDS="~amd64"
+# IUSE="dedicated libressl server"
+
+# ADD?
+# libsqlite3-dev (this is the ubuntu thingy) → dev-db/postgres
```
thesamesam, line 26: You mixed up the two!

```suggestion
# libsqlite3-dev (this is the ubuntu thingy) → dev-db/sqlite
```

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+# libpq-dev → dev-db/sqlite
```
thesamesam, line 27:
```suggestion
# libpq-dev → dev-db/postgresql
```

(the 'pq' is 'postgresql')

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+# yaml-cpp something?
```
thesamesam, line 28: eix yaml-cpp? :p

```suggestion
# yaml-cpp -> dev-cpp/yaml-cpp
```

sci-mathematics/rstudio/files/rstudio-1.4.1106-soci-cmake-find_library.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/src/cpp/CMakeLists.txt b/src/cpp/CMakeLists.txt
```
thesamesam, line 1: I'd consider dropping the first two lines but it's not needed either.

sci-mathematics/rstudio/files/rstudio-1.4.1106-soci-cmake-find_library.patch
```diff
@@ -0,0 +1,17 @@
...
+index df54994..eda8c8f 100644
+--- a/src/cpp/CMakeLists.txt
++++ b/src/cpp/CMakeLists.txt
+@@ -409,9 +409,9 @@ if(UNIX)
+    if(NOT APPLE AND RSTUDIO_USE_SYSTEM_SOCI)
+       set(SOCI_LIBRARY_DIR "/usr/lib")
```
thesamesam, line 7: I wonder if this is going to cause issues for us, not respecting libdir.

sci-mathematics/rstudio/files/rstudio-1.4.1106-add-missing-boost-bind-import.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/src/cpp/core/HtmlUtils.cpp b/src/cpp/core/HtmlUtils.cpp
```
thesamesam, line 1: Is any of this upstreamable btw?

thesamesam (commented):
Looking good so far!

dev-libs/soci/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
+		<email>WolfgangESanyer@Gmail.com</email>
+		<name>Wolfgang E. Sanyer</name>
+	</maintainer>
+</pkgmetadata>
```
juippis, line 8: You'll need a project or a person with `@gentoo.org` address to proxy commits for you.

dev-libs/soci/soci-4.0.1.ebuild
```diff
@@ -0,0 +1,32 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake
+
+DESCRIPTION="Simple Open (Database) Call Interface - write SQL in C++"
+HOMEPAGE="https://github.com/SOCI/soci"
+SRC_URI="https://github.com/${PN}/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
```
juippis, line 10: We generally dislike having `${PN}` in here, as it's not a changeable variable, and makes copy-pasting/grepping harder. Also imagine Github had case-sensitive URL matching.

dev-libs/soci/soci-4.0.1.ebuild
```diff
@@ -0,0 +1,32 @@
...
+IUSE="firebird mysql odbc oracle postgres sqlite"
+
+RDEPEND="
+	>=dev-libs/boost-1.33.1:=
+	firebird? ( dev-db/firebird )
+	mysql? ( dev-db/mysql )
+	odbc? ( dev-db/unixODBC )
+	oracle? ( dev-db/oracle-instantclient )
+	postgres? ( dev-db/postgresql:* )
+	sqlite? ( dev-db/sqlite )
```
juippis, line 24: These USE flags have no function in this ebuild, and are banned by QA policy.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+		"${DISTDIR}"/core-dictionaries.zip || die
+}
+
+src_prepare() {
+	cmake_src_prepare
+
+	# This will set the target install directory for "support" items
+	sed -i \
+		-e "s|set(RSTUDIO_INSTALL_SUPPORTING .)|set(RSTUDIO_INSTALL_SUPPORTING ${EPREFIX}/usr/share/${PN})|g" \
+		CMakeGlobals.txt
```
juippis, line 80: Missing `|| die`.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,97 @@
...
+	>=dev-qt/qtnetwork-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtopengl-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtpositioning-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtsensors-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtsql-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtsvg-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtwebchannel-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtwebengine-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtxml-${QT_VER}:${QT_SLOT}
+	>=dev-qt/qtxmlpatterns-${QT_VER}:${QT_SLOT}
```
juippis, line 49: I don't do qt so don't quote me on this... but AFAIK Qt doesn't break its ABI on a major slot. So you should be able to just depend on `:5`. But ask @a17r to be sure. This just looks weird to me :)

dev-libs/soci/metadata.xml
ezzieyguywuf: Where do i solicit such a thing?

dev-libs/soci/soci-4.0.1.ebuild
ezzieyguywuf: Ah hah, well noted!

So how do I handle, e.g. `rstudio` depending on this AND `postgresql`.....nvm just answered my own question

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: I'll be interested to hear what @a17r has to say.

For what it's worth this is the same approach taken in the previous version of this ebuild.

dev-libs/soci/metadata.xml
thesamesam: [Add](https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide#Proxied_maintainer_in_metadata.xml) proxy-maint as a project

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Should work, cmake is very forgiving in this regard.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: I'll check! This is how the original ebuild author did it, so I assumed it was necessary

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Yes but probably no.

The build system as is required it for some reason, but probably i can patch the cmake configs with no negative consequences.

I'll look into this (i added it at first to just get the thing to build at least)

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: I had a similar though actually, just didn't let the thought fully play out in my head 🤣

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Lol, mixed them up in the comment, but i knew in my head which was which 

These notes are purely temporary and only intended for me: don't peak behind the curtain!!

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Hah, indeed. I did find it, but ended up compiling fine without it.

I decided to leave the note to myself in case some issue pops up later during testing.

sci-mathematics/rstudio/files/rstudio-1.4.1106-add-missing-boost-bind-import.patch
ezzieyguywuf: Very good question - probably yes, though the codebase is so vast that getting involved with upstream is a tad daunting.

Perhaps once I get the ebuild buttoned up i can work on a single megaissue for upstream.

ezzieyguywuf:
Closes #19885 
Closes #19893 

I originally thought that splitting out these other two PR's made sense, in order to try to better manage the size/scope of this rstudio version bump.

However, complications arose - in order to bump the pandoc version (which is needed for the rstudio bump), we needed to mask for removal `pandoc-citeproc`.

Why? Because `pandoc-citeproc` is deprecated and not compatible with the newer pandoc, and due to the specifics of how ghc handles things (see [here][1] for more info), it's rather challenging to allow multiple versions of haskell packages using the tried-and-true slot system.

Ok, so the original idea was to mask `pandoc-citeproc` for removal and start the 30-day countdown, but it turns out that `rstudio-1.3.1093 (the only other version in the tree other than the one we're trying to add) has a direct dependency on `pandoc-citeproc`! So this older rstudio needs to be masked at the same time as `pandoc-citeproc`, but if that's done outside of this PR, we're left with zero un-masked rstudio ebuilds, which it seems is a bad thing.

Therefore, I'm merging the three PR's into one.

TL;DR these three PR's probably should have always just been a single PR anyway. Feel free to skip the novel above and be about your business.

[1]: https://github.com/gentoo-haskell/gentoo-haskell/blob/master/projects/doc/maintaining-tips.rst#on-problems-with-slotting-haskell-libraries

sci-mathematics/rstudio/files/rstudio-1.4.1106-soci-cmake-find_library.patch
ezzieyguywuf: Can you elaborate on what you mean by this?

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: I opted for `"${EPREFIX}/usr/share/${PN}"`

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Ok, I posted a new attempt using `unpack` let me know your thoughts.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Added `${P}` to the filename.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Added `:*` slot op.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Removed the comments.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: so far it doesn't seem like it's needed...

sci-mathematics/rstudio/files/rstudio-1.4.1106-soci-cmake-find_library.patch
ezzieyguywuf: For better or worse, I use the raw `git diff` as described [here on the wiki](https://wiki.gentoo.org/wiki/Patches#Creating_a_patch). Do you think the wiki should be updated?

ezzieyguywuf:
Anyone interested, this is ready for testing (though not ready for merging yet).

I'm unfamiliar with rstudio so testers are welcome

dev-libs/soci/soci-4.0.1.ebuild
ezzieyguywuf: @juippis it seems like `firebird` is masked, and has a bunch of masked  dependencies as well.

Would it be appropriate to place at least this one behind a `USE`-flag so that a user can decide whether or not to unmask all that stuff?

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Fixed

dev-libs/soci/soci-4.0.1.ebuild
juippis: Masked by...? At least the `firebird::gentoo` looks fine to me, no weird deps and it's only stable.masked, so wouldn't stop this at all.

Just use `optfeature`.

sci-mathematics/rstudio/files/rstudio-1.4.1106-soci-cmake-find_library.patch
thesamesam: I think I clarified on IRC in the end

sci-mathematics/rstudio/files/rstudio-1.4.1106-soci-cmake-find_library.patch
thesamesam: [devmanual](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#clean-patch-howto) is authoritative but yes

dev-games/goatee/metadata.xml
```diff
@@ -1,21 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
 <pkgmetadata>
-	<maintainer type="project">
```
thesamesam, line 4: Did you ask someone on games@?

dev-haskell/commonmark/commonmark-0.1.1.4.ebuild
```diff
@@ -0,0 +1,29 @@
...
+SLOT="0/${PV}"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="dev-haskell/parsec:=[profile?]
+	dev-haskell/text:=[profile?]
+	>=dev-lang/ghc-8.4.3:=
+"
+DEPEND="${RDEPEND}
+	>=dev-haskell/cabal-2.2.0.1
+	test? ( dev-haskell/quickcheck
```
thesamesam, line 25: I'd really do a new line here and before the last )

dev-libs/soci/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
+		<email>WolfgangESanyer@Gmail.com</email>
+		<name>Wolfgang E. Sanyer</name>
+	</maintainer>
```
thesamesam, line 7: You still need proxy-maint here or a gentoo dev

dev-libs/soci/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
+		<email>WolfgangESanyer@Gmail.com</email>
```
thesamesam, line 5: This isn't your git author email btw

games-board/goatee-gtk/goatee-gtk-0.4.0.ebuild
```diff
@@ -0,0 +1,30 @@
...
+
+DESCRIPTION="A monadic take on a 2,500-year-old board game - GTK+ UI"
+HOMEPAGE="https://khumba.net/projects/goatee"
+SRC_URI="https://hackage.haskell.org/package/${P}/${P}.tar.gz"
+
+LICENSE="AGPL-3"
+SLOT="0/${PV}"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND=">=dev-games/goatee-0.4:=[profile?] <dev-games/goatee-0.5:=[profile?]
```
thesamesam, line 19: I don't love doing these on the same line _but_ if it's the Haskell style, that's fine. We let e.g. the Perl project have their own style wrt sorting in a specific way too.

profiles/package.mask
```diff
@@ -32,6 +32,12 @@
 
 #--- END OF EXAMPLES ---
 
+# Wolfgang E. Sanyer <WolfgangESanyer@Gmail.com> (2021-03-11)
+# Per upstream, this has been deprecated in favour if citeproc. bug #775545
```
thesamesam, line 36:
```suggestion
# Per upstream, this has been deprecated in favour of citeproc. bug #775545
```

profiles/package.mask
```diff
@@ -32,6 +32,12 @@
...
+# opened per the examples - Masked for removal in 30 days
+dev-haskell/pandoc-citeproc
+=sci-mathematics/rstudio-1.3.1093
```
thesamesam, line 39: What I would've preferred is:
- Add new versions of everything
- Remove old after a while

But I understand that's a mess sometimes with huge things like Haskell, ML, ..., so if we have to, it works I suppose.

sci-mathematics/rstudio/files/rstudio-1.4.1106-add-missing-boost-bind-import.patch
thesamesam: Sure.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,112 @@
...
+	https://github.com/rstudio/rstudio/"
+SRC_URI="
+	https://github.com/rstudio/rstudio/archive/v${PV}.tar.gz -> ${P}.tar.gz
+	https://s3.amazonaws.com/rstudio-dictionaries/core-dictionaries.zip -> ${P}-core-dictionaries.zip
+"
+
+LICENSE="AGPL-3"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="-headless test"
```
thesamesam, line 23: We [shouldn't](https://devmanual.gentoo.org/general-concepts/use-flags/index.html#iuse-defaults) do "-foo" in IUSE, it's pointless

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,112 @@
...
+	for ARCHIVE in ${A}
+	do
+		if [[ ${ARCHIVE:-${#dict_archive_name}} == "${dict_archive_name}" ]]
+		then
+			mkdir "${WORKDIR}/dictionaries" || die
+			pushd "${WORKDIR}/dictionaries" > /dev/null || die
+			unpack ${ARCHIVE}
+			popd > /dev/null || die
+		else
+			unpack ${ARCHIVE}
```
thesamesam, line 77: Happy!

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,112 @@
...
+	"${FILESDIR}/${PN}-1.4.1106-remove-mathjax-internal-dep-check.patch"
+	"${FILESDIR}/${PN}-1.4.1106-soci-cmake-find_library.patch"
+)
+
+src_unpack() {
+	local dict_archive_name="${P}-core-dictionaries.zip"
+
+	for ARCHIVE in ${A}
+	do
+		if [[ ${ARCHIVE:-${#dict_archive_name}} == "${dict_archive_name}" ]]
```
thesamesam, line 70: A comment or two might be worthwhile

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,112 @@
...
+	"${FILESDIR}/${PN}-1.4.1106-remove-gwt-build.patch"
+	"${FILESDIR}/${PN}-1.4.1106-remove-hardcoded-internal-mathjax-and-pandoc-deps.patch"
+	"${FILESDIR}/${PN}-1.4.1106-remove-mathjax-internal-dep-check.patch"
+	"${FILESDIR}/${PN}-1.4.1106-soci-cmake-find_library.patch"
+)
+
+src_unpack() {
+	local dict_archive_name="${P}-core-dictionaries.zip"
+
+	for ARCHIVE in ${A}
```
thesamesam, line 68: This is fine, although unusual style for ebuilds.

Normally:
```suggestion
	for ARCHIVE in ${A} ; then
```
or
```suggestion
	for ARCHIVE in ${A}; then
```

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,112 @@
...
+	export RSTUDIO_VERSION_MINOR=$(ver_cut 2)
+	export RSTUDIO_VERSION_PATCH=$(ver_cut 3)
+
+	local mycmakeargs=(
+		-DRSTUDIO_TARGET=$(usex headless Server Desktop)
+		-DRSTUDIO_INSTALL_SUPPORTING="${EPREFIX}/usr/share/${PN}"
+		-DRSTUDIO_UNIT_TESTS_DISABLED=ON
+		-DRSTUDIO_USE_SYSTEM_BOOST=ON
+	)
+	if ! use headless
```
thesamesam, line 106:
```suggestion
	if ! use headless ; then
```

sci-mathematics/rstudio/files/rstudio-1.4.1106-soci-cmake-find_library.patch
ezzieyguywuf: ok, duly noted, I'll rework all the patches to match the criteria laid out in the devmanual.

sci-mathematics/rstudio/files/rstudio-1.4.1106-soci-cmake-find_library.patch
ezzieyguywuf: Yes you did, I will fix this.

sci-mathematics/rstudio/files/rstudio-1.4.1106-add-missing-boost-bind-import.patch
ezzieyguywuf: 👍 I'll close this as resolved and add a task to the `TODO` list at the top of the PR

dev-libs/soci/soci-4.0.1.ebuild
ezzieyguywuf: Yea, I meant "masked by unstable keyword".

I'll do as you've suggested and use `optfeature`.

dev-games/goatee/metadata.xml
ezzieyguywuf: I did not - should I?

Copied this over from `::haskell`, so I thought it was appropriate to leave it with `haskell@`

dev-haskell/commonmark/commonmark-0.1.1.4.ebuild
ezzieyguywuf: This ebuild is generated using `hackport` - I can manually make these stylistic changes, but more appropriate would probably be to update the `hackport` thing so it outputs the stylings we want.

For example, I recently updated `hackport` to exclude `IUSE=""` from the ebuild.

dev-libs/soci/metadata.xml
ezzieyguywuf: See [here][1], I  think there may be some github shenanigans going on, or maybe I just hadn't pushed the update when you did this review.

[1]: https://github.com/gentoo/gentoo/pull/19865/files?file-filters%5B%5D=.xml#diff-f27b045bf2a2451cb68ad2bce7d4fdfda159683536d887ee4a2c7007ec509525R8

dev-libs/soci/metadata.xml
ezzieyguywuf: Hrm, that's right, I had to locally change my git author email to `ezzieyguywuf@gmail.com` to match my bugzilla thing.

I guess I should use the same email address here for consistency?

games-board/goatee-gtk/goatee-gtk-0.4.0.ebuild
ezzieyguywuf: I've had some conversations with @trofi about this, see [here][1] and [here][2], and I think I have some IRC logs as well. Long story short, it's more-or-less as you've said - it's the "Haskell style".

Personally, I prefer one per line as well, but rather that fight that battle I prefer to focus my energies trying to revitalize [bug 4315][3].  I think @mgorny does a good job [laying out the problem][4] and discussing why such a feature is necessary.

[1]: https://github.com/gentoo-haskell/gentoo-haskell/pull/1097
[2]: https://github.com/gentoo-haskell/gentoo-haskell/pull/1099#discussion_r546277694
[3]: https://bugs.gentoo.org/4315
[4]: https://wiki.gentoo.org/wiki/Future_EAPI/Version_syntax_changes#Version_ranges

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Yes, I noticed the preference for the `; then` on the same line - why is that? I think it looks cleaner and is easier to read with `then` on its own line, plus it gets rid of the ugly `;`.

I can change it if absolutely necessary.

dev-libs/soci/metadata.xml
ezzieyguywuf: Understood, I'll update this to my bugzilla email.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
thesamesam: Happens to all of us!

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
thesamesam: I suppose it should be okay. I think there's value in using the same-ish style as everybody else as it eases readability though.

games-board/goatee-gtk/goatee-gtk-0.4.0.ebuild
thesamesam: 💯

dev-libs/soci/metadata.xml
thesamesam: yes, I don't really mind which, but they should be consistent

dev-haskell/commonmark/commonmark-0.1.1.4.ebuild
thesamesam: Yeah, let's not worry for now then (and yay!)

dev-games/goatee/metadata.xml
thesamesam: You're dropping them as a maintainer, so yes ;)

dev-haskell/commonmark/commonmark-0.1.1.4.ebuild
ezzieyguywuf: Opened [an issue][https://github.com/gentoo-haskell/hackport/issues/92] on the hackport project so I don't forget about this.

dev-libs/soci/metadata.xml
ezzieyguywuf: Fixed, changed it to the bugzilla email.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Fixed

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: per `#gentoo-dev` convo, I went ahead and just changed it.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: Done, see new comments.

ezzieyguywuf:
Hah, progress.

imagen

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
ezzieyguywuf: I believe you stated in a different comment that you were happy with the latest approach. Closing this as resolved.

sci-mathematics/rstudio/files/rstudio-1.4.1106-soci-cmake-find_library.patch
ezzieyguywuf: Patches are fixed and should conform to the layout suggested in the devmanual. Closing as resolved.

dev-games/goatee/metadata.xml
ezzieyguywuf: This is moot now, as @trofi has bumped the package already on `master`. Closing this, and removing from this PR.

profiles/package.mask
ezzieyguywuf: fixed

juippis (commented):
I haven't read the ebuilds yet, just launched a build-test to give you something to do ;) well I didn't get to rstudio because gwt failed. I'd really appreciate if you fixed gwt in here, or in a separate PR (we can process that as the maintainer doesn't seem to care - please ping me)

One thing that did happen to, when emerging *soci* I always got:
```
QA: other
The ebuild is installing to one or more unexpected paths:

  /usr/cmake

Please fix the ebuild to use correct FHS/Gentoo policy paths.
```
```
# ls /usr/cmake
SOCI-gentoo.cmake  SOCI.cmake
```
An easy fix, but do check why these get installed, and is the purpose to use these files somehow.

ezzieyguywuf:
@juippis thanks for the ping, i had stepped away from this recently to work on the quiz and also "work work" had been calling.

I do plan to get back to it though, perhaps i can look tonight.

`gwt` sounds familiar, i think there may be a bug report open with a workaround. I'll take a look at this and the `soci` thing you mentioned. The `soci` one has been on my list but i wanted to get `rstudio` further along first.

dev-libs/soci/soci-4.0.1.ebuild
ezzieyguywuf: This is fixed now.

ezzieyguywuf:
> I'd really appreciate if you fixed gwt in here, or in a separate PR

Done, see #20222 

> One thing that did happen to, when emerging _soci_<snip>
> 
> Please fix the ebuild to use correct FHS/Gentoo policy paths.

Done, and I've also updated my make.conf so that I should hopefully catch these myself in the future.

dev-libs/soci/soci-4.0.1.ebuild
ezzieyguywuf: @juippis i think optfeature is the wrong tool here - unless the user is installing a binpkg, they won't see any message due to the RDEPEND.

I think I'll use einfo instead

juippis (commented):
```
[ebuild   R    ] sci-mathematics/rstudio-1.4.1106::gentoo  USE="server* -headless -test" 0 KiB

Total: 1 package (1 reinstall), Size of downloads: 0 KiB

>>> Verifying ebuild manifests
>>> Emerging (1 of 1) sci-mathematics/rstudio-1.4.1106::gentoo
>>> Installing (1 of 1) sci-mathematics/rstudio-1.4.1106::gentoo
>>> Jobs: 1 of 1 complete                           Load avg: 8.87, 8.81, 5.99

 * Messages for package sci-mathematics/rstudio-1.4.1106:
 * Log file: /var/tmp/portage/vbslogs/build/sci-mathematics/rstudio-1.4.1106:20210407-132719.log

 * 
 *   /usr/db
 *   /usr/extras
 * 
```

Then soci's *test phase* fails, 
```
/var/tmp/portage/dev-libs/soci-4.0.1/work/soci-4.0.1/tests/common-tests.h:547: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  I/O error during "open" operation for file "/tmp/test.fdb"
  Error while trying to open file
  No such file or directory
```
(getting tons of these, sandbox-related for sure)

Then I can execute rstudio, but the program is blank and no menu entry work. 
```
# rstudio
"Uncaught TypeError: Cannot read property 'notifyRCrashed' of undefined"
```

I don't mind the wrong directories too much, but the soci ebuild should be checked.

dev-libs/soci/soci-4.0.1.ebuild
```diff
@@ -0,0 +1,48 @@
...
+	>=dev-libs/boost-1.33.1:=
+"
+DEPEND="
+	${RDEPEND}
+	dev-db/firebird
+	dev-db/mysql
+	dev-db/unixODBC
+	dev-db/oracle-instantclient
+	dev-db/postgresql:*
+	dev-db/sqlite
```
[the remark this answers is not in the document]
ezzieyguywuf: so, I had these behind `USE` flags at first, but per [the devmanual](https://devmanual.gentoo.org/general-concepts/use-flags/index.html#when-not-to-use-use-flags) (specifically "The usage of a USE flag should not control runtime dependencies") as well as some conversation in this PR, I removed the `USE` flags and just unconditionally build the package against all it's supported backends.

> Maybe just one out of them all?

But how would I choose which one?

> What's the interaction with `DEPEND` here?

I think soci needs the headers and stuff at compile time.

> Are they `dlopen`'d

I'm not sure - how do I check this?

dev-libs/soci/files/soci-4.0.1-cmake-config-path.patch
```diff
@@ -0,0 +1,14 @@
...
+@@ -124,7 +124,8 @@ if (SOCI_STATIC)
+     ARCHIVE DESTINATION ${LIBDIR})
+ endif()
+ 
+-install(EXPORT SOCI NAMESPACE SOCI:: DESTINATION cmake)
++install(EXPORT SOCI NAMESPACE SOCI:: DESTINATION
++    ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/cmake)
+ 
+ #
+ # Core configuration summary
```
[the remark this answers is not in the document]
ezzieyguywuf: I think I need to fix the cmake install path on `SOCI`?

Except, I just tried this on a fresh, clean, dev environment and this did not cause an issue.

hrm...

YIKES! `SOCI libraries found under  /opt/rstudio-tools/soci/build/lib` dunno where that came from. I think my "clean" environment is probably not-so-clean.

dev-libs/soci/soci-4.0.1.ebuild
```diff
@@ -0,0 +1,48 @@
...
+pkg_postinst() {
+	local databases=(
+		dev-db/firebird
+		dev-db/mysql
+		dev-db/unixODBC
+		dev-db/oracle-instantclient
+		dev-db/postgresql
+		dev-db/sqlite
+	)
+	optfeature "database integration" ${databases[@]}
```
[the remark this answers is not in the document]
ezzieyguywuf: see my answer above.

dev-libs/soci/soci-4.0.1.ebuild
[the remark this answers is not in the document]
juippis: > 
> > What's the interaction with `DEPEND` here?
> 
> I think soci needs the headers and stuff at compile time.
> 

Okay, I digged through the upstream code,
https://github.com/SOCI/soci/blob/master/src/backends/CMakeLists.txt
and unfortunately it looks like they use what we call "automagic dependency detection" to find those deps.

Well a good thing is they use CMake so patching these automagic deps into configure options, how we do in Gentoo, should be rather easy. But it requires work of course.
https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies#CMake
https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/check/files/check-0.14.0-r2-disable-automagic-dep.patch (and an example)


> > Are they `dlopen`'d
> 
> I'm not sure - how do I check this?

This should be in your quiz, no? ;)
You can use **strace** to see `dlopen`s, but I personally just find unpacking the source tarball and `grep`ping a alot faster. For binary-only packages you need to use **strace** though.

dev-libs/soci/files/soci-4.0.1-cmake-config-path.patch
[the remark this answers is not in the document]
juippis: It did look like those two files were also correctly installed under `/usr/lib64/cmake`. Maybe you can just remove the bogus dir for a fast fix. Maybe.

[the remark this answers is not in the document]
ezzieyguywuf: This is fixed now - turns out upstream had **two** locations where they were calling out `install` for this file (actually more - one for each backend) - I just had to update my patch to capture both (one is a macro) and it's good now! I should probably submit this patch upstream honestly...

dev-libs/soci/soci-4.0.1.ebuild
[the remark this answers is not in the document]
ezzieyguywuf: Thanks for the thorough answer here, I'll get to work dealing with this automagic thing you mention.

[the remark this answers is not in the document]
ezzieyguywuf: I've upated this SOCI ebuild now, let me know what you think @juippis

ezzieyguywuf:
@juippis I think this should finally install (using the other two PR's, they're mentioned in the first post at the very top) and run. Do you mind giving it a test?

ezzieyguywuf:
Note - failing CI is due to #20575 and #20222 not being merged yet - I can squash the three into one large PR if anyone thinks it is worthwhile.

dev-libs/soci/soci-4.0.1.ebuild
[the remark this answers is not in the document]
ezzieyguywuf: marking as resolved since I split this into #20575

dev-libs/soci/soci-4.0.1.ebuild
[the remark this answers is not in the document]
ezzieyguywuf: Marking as resolved since soci ebuild was split into separate PR

sci-mathematics/rstudio/files/rstudio-1.4.1106-soci-cmake-find_library.patch
ezzieyguywuf: @thesamesam  I looked in to this: per the [cmake documentation][1], the `PATHS` option simply adds extra places to search. The main point of my patch is to remove `NO_DEFAULTS` - the cmake `find_library` is smart enough to figure out where the libraries are (usually).

TL;DR this should not be an issue, as it is only used by cmake to find the library.

[1]: https://cmake.org/cmake/help/latest/command/find_library.html

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,217 @@
...
+# [ FAIL 7 | WARN 6 | SKIP 2 | PASS 980 ]
+RESTRICT="test"
+
+RDEPEND="
+	dev-db/postgresql:*
+	dev-db/sqlite
+	dev-java/aopalliance:1
+	dev-java/gin:2.1
+	dev-java/javax-inject
+	=dev-java/validation-api-1.0*:1.0[source]
```
fordfrog, line 46: is still this old version needed? could the version restriction be dropped and keep just the slot?

ezzieyguywuf: @fordfrog  I'll grep around in the cmake files and see what I can find out.

fordfrog: the easiest way to check is to drop the version restriction, emerge the newest `validation-api` and then try to emerge `rstudio` against it. if it compiles then the api used is compatible and the restriction is not needed anymore.

ezzieyguywuf: > What's the outcome for this?

Have not tested yet

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,217 @@
...
+src_configure() {
+	export RSTUDIO_VERSION_MAJOR=$(ver_cut 1)
+	export RSTUDIO_VERSION_MINOR=$(ver_cut 2)
+	export RSTUDIO_VERSION_PATCH=$(ver_cut 3)
+
+	# The cmake configurations allow installing either or both of a Desktop
+	# component (i.e. a graphical front-end written in Qt) or a Server component
+	# (i.e. a background process that is accessed remotely, via a browser or
+	# some such).
+	local RSTUDIO_BUILD_TYPE=""
```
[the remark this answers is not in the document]
ezzieyguywuf: Si you think i should keep this local, or should I make it global so it gets saved with the package metadata?

sci-mathematics/rstudio/files/rstudio-1.4.1106-server-paths.patch
```diff
@@ -0,0 +1,139 @@
+# This path moves the server installation files to a FHS-compliant location.
+# However, there's been limited tested done with the actualy server runtime, so
+# please report any issues you have (this codebase has a tendency to hardcode
+# paths or to try to be clever about determining where things are located).
```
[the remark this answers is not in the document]
ezzieyguywuf: Hrm, the patch is my own.
Maybe i should change the comment here if it makes it seems like the patch came from somewhere else.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,217 @@
...
+
+	# move the dictionary files to the expected place
+	mkdir -p "${S}/dependencies/common" || die
+	mv "${WORKDIR}/dictionaries" "${S}/dependencies/common" || die
+}
+
+src_configure() {
+	export RSTUDIO_VERSION_MAJOR=$(ver_cut 1)
+	export RSTUDIO_VERSION_MINOR=$(ver_cut 2)
+	export RSTUDIO_VERSION_PATCH=$(ver_cut 3)
```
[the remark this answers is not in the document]
ezzieyguywuf: Yes the cmake config uses these variables.

If I'm not mistaken, if you don't set these them it tries to auto detect the version info using git

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
```diff
@@ -0,0 +1,217 @@
...
+src_install() {
+	cmake_src_install
+
+	if use server;then
+		dopamd src/cpp/server/extras/pam/rstudio
+		newinitd "${FILESDIR}/rstudio-server.initd" rstudio-server
+	fi
+
+	# This binary name is much to generic, so we'll change it
+	mv "${D}/${EPREFIX}/usr/bin/diagnostics" "${D}/${EPREFIX}"
```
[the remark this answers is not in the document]
ezzieyguywuf: I think intended to rename diagnostics to rstudio-diagnostics, but i did that in a patch instead. 

This is just an artifact of my original attempt. I'll take it out.

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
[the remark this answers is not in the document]
juippis: Sweet, please add a comment above these lines in the ebuild to state that :)

sci-mathematics/rstudio/rstudio-1.4.1106.ebuild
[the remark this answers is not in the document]
juippis: If the build system uses and expects it, it doesn't matter, but to me it just looked like it was used as a variable in the ebuild, and had no interaction with the build system.

sci-mathematics/rstudio/files/rstudio-1.4.1106-resource-path.patch
```diff
@@ -0,0 +1,35 @@
+# Upstream takes it upon themselves to assume that the resources dir is one
+# directory above the binary, i.e. if /usr/bin/rstudio, then it should be
+# /usr/resources. This patch allows to specify a target resources directory to
+# cmake, and the rstudio runtime will respect this path.
```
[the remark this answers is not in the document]
ezzieyguywuf: I don't think so: usually, I like to use `/opt` whenever the program ships with a bunch of pre-compiled stuff, i.e. dwarf-fortress. In this case, we compile everything, but we just need to put the large hierarchy of expected files in an appropriate location.

I'm open to input on this though.

AndrewAmmerlaan:
What is the status here? It looks mostly fine to me (apart from some small things that @juippis already pointed out).

I noticed that this package is now masked for removal soon, can we save it with this PR? (CC @thesamesam )

Note that the checks are failing because we need to merge https://github.com/gentoo/gentoo/pull/20575 first.

ezzieyguywuf:
> What is the status here? It looks mostly fine to me (apart from some small things that @juippis already pointed out).

I can take another pass at this. If memory serves, the compile time is sort of long, so i don't know if i can commit to getting it across the finish line, but I'll definitely give it another try or two

thesamesam:
I honestly can't keep track of whether the ebuilds in the bug are supposed to work or not. Last time I read through it (and there's a lot of noisy comments), there was still some issue when actually trying to use them. But maybe I'm wrong.

I think the ebuilds in the bug / overlays linked in the bug build upon this PR and so this PR would need refreshing.

But it would be really great if we could keep rstudio.

t0b3:
> I can take another pass at this. If memory serves, the compile time is sort of long, so i don't know if i can commit to getting it across the finish line, but I'll definitely give it another try or two

@ezzieyguywuf if it serves you feel free to profit from 
[rstudio-2022.02.0.443.tar.gz](https://github.com/gentoo/gentoo/files/9299437/rstudio-2022.02.0.443.tar.gz)

notice: with this I got rstudio 2022.02.7 to work for me, though it hasn't been tested it the wild yet :smile: the patches are basically a port of yours IIRC.

PR #19876, opened 2021-03-11
net-libs/pjproject: security rev bump

jkroonza:
Upstream didn't release a new version as one would expect.  Instead
patches are applied locally.

Also add subslot because they are equally good at maintaining ABI
compatibility, and SONAME is never updated, thus we need to be able to
depend on subslots to rebuild (preserved-rebuild is no good).

Bug: https://bugs.gentoo.org/775359
Bug: https://bugs.gentoo.org/775353

### DIFF
--- a/net-libs/pjproject/pjproject-2.10-r1.ebuild
+++ b/net-libs/pjproject/pjproject-2.10-r1.ebuild
@@ -1,0 +1,125 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools flag-o-matic toolchain-funcs
+
+DESCRIPTION="Open source SIP, Media, and NAT Traversal Library"
+HOMEPAGE="https://www.pjsip.org/"
+SRC_URI="https://github.com/pjsip/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~x86"
+
+LICENSE="GPL-2"
+SLOT="0/${PV}"
+
+# g729 not included due to special bcg729 handling.
+CODEC_FLAGS="g711 g722 g7221 gsm ilbc speex l16"
+VIDEO_FLAGS="sdl ffmpeg v4l2 openh264 libyuv vpx"
+SOUND_FLAGS="alsa portaudio"
+IUSE="amr debug epoll examples ipv6 libressl opus resample silk ssl static-libs webrtc
+	${CODEC_FLAGS} g729
+	${VIDEO_FLAGS}
+	${SOUND_FLAGS}"
+
+PATCHES=(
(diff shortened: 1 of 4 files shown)


net-libs/pjproject/pjproject-2.10-r1.ebuild
```diff
@@ -0,0 +1,125 @@
...
+		myconf+=( $(use_enable ${t} ${t}-codec) )
+	done
+	myconf+=( $(use_enable g729 bcg729) )
+
+	for t in ${VIDEO_FLAGS}; do
+		myconf+=( $(use_enable ${t}) )
+		use "${t}" && videnable="--enable-video"
+	done
+
+	[ "${videnable}" = "--enable-video" ] && append-cflags -DPJMEDIA_HAS_VIDEO=1
```
thesamesam, line 84: Bash tests next time around?

net-libs/pjproject/pjproject-2.10-r1.ebuild
```diff
@@ -0,0 +1,125 @@
...
+VIDEO_FLAGS="sdl ffmpeg v4l2 openh264 libyuv vpx"
+SOUND_FLAGS="alsa portaudio"
+IUSE="amr debug epoll examples ipv6 libressl opus resample silk ssl static-libs webrtc
+	${CODEC_FLAGS} g729
+	${VIDEO_FLAGS}
+	${SOUND_FLAGS}"
+
+PATCHES=(
+	"${FILESDIR}/pjproject-2.9-ssl-enable.patch"
+	"${FILESDIR}/pjproject-2.10-race-condition-between-transport-destroy-and-acquire.patch"
```
thesamesam, line 27: ${PN} at least, but we both know I'd do ${P} for upstreamed patches ;)

thesamesam (approved):

net-libs/pjproject/pjproject-2.10-r1.ebuild
jkroonza: that's smart actually, that means it will break on simple copy instead of trying to re-apply a patch.

net-libs/pjproject/pjproject-2.10-r1.ebuild
jkroonza: [[ ... ]] ?

net-libs/pjproject/pjproject-2.10-r1.ebuild
thesamesam: Exactly! It's a hint to "not even bother trying". Sometimes, scarily, they DO apply (wrongly) because things changed with fuzz or we applied a slightly different version.

net-libs/pjproject/pjproject-2.10-r1.ebuild
thesamesam: Yeah, exactly. It's just because we're guaranteed to have Bash and POSIX tests ([]) have some unexpected semantics.

PR #19879, opened 2021-03-11
net-misc/asterisk-*: add <stabilize-allarches/>

jkroonza:
These packages merely install a set of audio files and contain no code.


net-misc/asterisk-core-sounds/metadata.xml
```diff
@@ -9,6 +9,7 @@
 		<email>proxy-maint@gentoo.org</email>
 		<name>Proxy Maintainers</name>
 	</maintainer>
+	<stabilize-allarches />
```
thesamesam, line 12: I changed to drop the space, thanks!

PR #19881, opened 2021-03-11
media-libs/libavif: revbump with multilib support

novomesk:


thesamesam:
Thanks for working so quickly on this.

PR #19883, opened 2021-03-11
dev-vcs/tortoisehg: 5.7 add PYTHON_COMPAT python3_9 

band-a-prend:


dev-vcs/tortoisehg/tortoisehg-9999.ebuild
```diff
@@ -11,13 +11,13 @@ inherit desktop distutils-r1 xdg-utils
 if [[ ${PV} != *9999* ]]; then
 	KEYWORDS="~amd64 ~arm64 ~x86"
 	SRC_URI="https://www.mercurial-scm.org/release/tortoisehg/targz/${P}.tar.gz"
-	HG_DEPEND=">=dev-vcs/mercurial-5.6
-		<dev-vcs/mercurial-5.8"
+	HG_DEPEND=">=dev-vcs/mercurial-5.6[${PYTHON_USEDEP}]
+		<dev-vcs/mercurial-5.8[${PYTHON_USEDEP}]"
```
juippis, line 15: Hmm, this got me thinking. Isn't mercurial just used during build-time to get the sources? What's the interaction so it's still required in RDEPEND? 

If this is BDEPEND-only, it shouldn't need `PYTHON_USEDEP` either. And yeah, put in BDEPEND.

band-a-prend: Mercurial is used as RDEPEND, it checks on thg start if compatible mercurial version is installed.

juippis (approved):

PR #19890, opened 2021-03-12
freedict.eclass: Support EAPI-7, port all ebuilds, drop EAPI-6

a17r:


eclass/freedict.eclass
```diff
@@ -1,37 +1,32 @@
...
+	7)
+		[[ ${FORLANG} ]] && die "FORLANG is banned, set DESCRIPTION instead"
+		[[ ${TOLANG} ]] && die "TOLANG is banned, set DESCRIPTION instead"
+		;;
 	*) die "${ECLASS}.eclass is banned in EAPI=${EAPI}" ;;
 esac
 
-MY_P=${PN/freedict-/}
-
-DESCRIPTION="Freedict for language translation from ${FORLANG} to ${TOLANG}"
```
ulm, line 32: As far as I can see, `DESCRIPTION` is no longer assigned in EAPI 6, and the existing ebuilds don't assign it either.

app-dicts/freedict-deu-eng/freedict-deu-eng-1.0.ebuild
```diff
@@ -1,11 +1,10 @@
...
 
-EAPI=6
+EAPI=7
 
-FORLANG="German"
-TOLANG="English"
 inherit freedict
 
+DESCRIPTION="Freedict for language translation from German to English"
+
```
ulm, line 9: I'd remove that empty line in all ebuilds, because there are only two variables left.
(But of course, this is entirely a matter of personal taste.)

eclass/freedict.eclass
```diff
@@ -1,37 +1,32 @@
...
-# @DESCRIPTION:
-# Please see above for a description.
+# Strips PN of 'freedict' prefix, to be used in SRC_URI and doins
+FREEDICT_P=${PN/freedict-/}
 
 case ${EAPI:-0} in
-	6) ;;
+	7)
+		[[ ${FORLANG} ]] && die "FORLANG is banned, set DESCRIPTION instead"
+		[[ ${TOLANG} ]] && die "TOLANG is banned, set DESCRIPTION instead"
```
ulm, line 23: No need for these two lines to be in the EAPI conditional (and it confused me at first).

a17r: right, can be moved out after the EAPI-6 dropping

ulm (approved):
LGTM

PR #19891, opened 2021-03-12
aspell-dict-r1.eclass: Support EAPI-7, port all ebuilds, drop EAPI-6

a17r:


ulm:
Most (all?) of these packages would profit from `<stabilize-allarches/>` in metadata.xml.

ConiKost:
Well ok :-( I was in preparing all that, but you seem to do some stuff of it:
https://github.com/gentoo/gentoo/pull/19855

PR #19892, opened 2021-03-12
www-client/qutebrowser: verbump to 2.1.0

BouncyMaster:
Signed-off-by: Nicu Borta <nicuborta123@tutanota.com>


www-client/qutebrowser/qutebrowser-2.1.0.ebuild
```diff
@@ -0,0 +1,90 @@
...
+	xdg_desktop_database_update
+	xdg_icon_cache_update
+	xdg_mimeinfo_database_update
+}
+
+pkg_postrm() {
+	xdg_desktop_database_update
+	xdg_icon_cache_update
+	xdg_mimeinfo_database_update
+}
```
juippis, line 90: Since pretty much the whole set of calls is needed here, one could just inherit *xdg.eclass* and call `xdg_pkg_post*`.

juippis (approved):

juippis:
```
www-client/qutebrowser
  PythonCompatUpdate: version 1.14.0: PYTHON_COMPAT update available: python3_9
```

PR #19893, opened 2021-03-12
profiles/package.mask: mask dev-haskell/pandoc-citeproc for removal

ezzieyguywuf:
This package has been deprecated upstream, and will block newer pandoc
if we keep it around.


ezzieyguywuf:
@trofi ping, as discussed on IRC.

profiles/package.mask
```diff
@@ -32,6 +32,11 @@
 
 #--- END OF EXAMPLES ---
 
+# Wolfgang E. Sanyer <WolfgangESanyer@Gmail.com> (2021-03-11)
+# Per upstream, this has been deprecated in favour if citeproc. bug #775545
+# opened per the examples - Masked for removal in 30 days
+dev-haskell/pandas-citeproc
```
trofi, line 38: I think you spelled it wrong.

ezzieyguywuf: Fixed - I was thinking of the python `pandas` library for some reason, I have no idea why :-P

profiles/package.mask
```diff
@@ -32,6 +32,12 @@
 
 #--- END OF EXAMPLES ---
 
+# Wolfgang E. Sanyer <WolfgangESanyer@Gmail.com> (2021-03-11)
```
trofi, line 35: The mask claims it masks deprecated package. But it also masks last available version of softwware maintained by someone else. You will need to sort out the plan with them first.

PR #19895, opened 2021-03-12
 dev-java/jmock: EAPI 7

vaukai:


dev-java/jmock/jmock-1.2.0-r2.ebuild
```diff
@@ -0,0 +1,48 @@
...
+HOMEPAGE="http://jmock.org/"
+SRC_URI="http://jmock.org/downloads/${P}-jars.zip"
+
+LICENSE="BSD"
+SLOT="1.0"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos"
+
+DEPEND="dev-java/junit:0"
+RDEPEND="${DEPEND}
+	>=virtual/jre-1.8:=
```
fordfrog, line 20: `:=` is not needed for java, `:*` is more appropriate as no matter what jdk you use to compile the package, the bytecode should still be able to run on the lowest supported version. this also applies to the other occurrences of `virtual/{jre,jdk}-1.8:=`, i won't tag those specifically.

dev-java/jmock/jmock-2.6.1-r1.ebuild
```diff
@@ -0,0 +1,45 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+JAVA_PKG_IUSE="doc source"
```
fordfrog, line 6: please also add `MAVEN_ID`

dev-java/jmock/jmock-1.2.0-r2.ebuild
vaukai: Done

dev-java/jmock/jmock-2.6.1-r1.ebuild
vaukai: Please re-check.  Missing any information how to determine the exact MAVEN_ID
[Java_Developer_Guide](https://wiki.gentoo.org/wiki/Java_Developer_Guide) is lacking such information :-(

fordfrog: sorry for that. `MAVEN_ID` is used by our tool called `java-ebuilder` to find matching dependencies. `MAVEN_ID` is just what it says, id of the package in maven repository. the template is`group_id:artifact_id:version`. i was just searching for the correct value for this version but i must admit i'm confused with jmock and their maven artifacts. anyway, [here](https://github.com/jmock-developers/jmock-library/tree/master) you can see what it looks like for their currentversion (in the maven section, <dependency>...). i suppose in our case it should be `MAVEN_ID="org.jmock:jmock-junit4:2.6.1"` but it might aswell be `jmock-legacy` so really not sure. so leave it as it is, thanks and sorry for the confusion :-)

dev-java/jmock/jmock-1.2.0-r2.ebuild
```diff
@@ -0,0 +1,48 @@
...
+SLOT="1.0"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos"
+
+DEPEND="dev-java/junit:0"
+RDEPEND="${DEPEND}
+	>=virtual/jre-1.8
+	"
+BDEPEND="${DEPEND}
+	app-arch/unzip
+	>=virtual/jdk-1.8
```
fordfrog, line 24: we don't put jdk in BDEPEND but in DEPEND. from the logic the package uses classes from the jdk and "links" against them so it belongs to DEPEND instead. also, for both jre/jdk it should specify `:*` at the end of the dep, so `>=virtual/jdk-1.8:*` and `>=virtual/jre-1.8:*`.

dev-java/jmock/jmock-2.6.1-r1.ebuild
vaukai: Did you mean to keep what I added here?
And, thanks for the explaining the template. I'll try to use it whenever possible.

fordfrog: please drop the following `MAVEN_ID`, that's not valid (the syntax is `group_id:artifact_id:version` and we use the version of the upstream, so in this case it would be `2.6.1`, without that revision number `-r1`). and as i wrote, i'm not even sure what is the correct `group_id` and `artifact_id` for this version.

vaukai: Dropped.  Sorry for misreading.

fordfrog:
@vaukai thanks! will do some tests tomorrow before pushing it.

fordfrog:
@vaukai pushed, thanks!

vaukai:
@fordfrog Please cc-arches for https://bugs.gentoo.org/775635?

fordfrog:
@vaukai done!

PR #19905, opened 2021-03-12
net-libs/libsrtp-1.6.0-r1: Fix compilation with sys-devel/autoconf-2.71

kripton:
Closes: https://bugs.gentoo.org/775680

Fix found at https://aur.archlinux.org/packages/lib32-libsrtp/#comment-787895
Tested with sys-devel/autoconf-2.71 and latest stable (sys-devel/autoconf-2.69-r5)

### DIFF
--- a/net-libs/libsrtp/libsrtp-1.6.0-r1.ebuild
+++ b/net-libs/libsrtp/libsrtp-1.6.0-r1.ebuild
@@ -45,6 +45,7 @@ src_prepare() {
 
 	mv configure.in configure.ac || die
 	eautoreconf
+	touch ar-lib || die #775680
 }
 
 multilib_src_configure() {


net-libs/libsrtp/libsrtp-1.6.0-r1.ebuild
```diff
@@ -45,6 +45,9 @@ src_prepare() {
 
 	mv configure.in configure.ac || die
 	eautoreconf
+
+	# Compatibility with sys-devel/autoconf:2.71
+	sed -i -e 's|install-sh ar-lib|install-sh|' configure
```
juippis, line 50: Can you use the approach from 4710a539a8237cd8afe2bfcc92c2e0850ce521ff? I want this pushed today, so sorry about the fast notice.

kripton: Sure, I'll check if that also solves the issue and change this PR then. Thanks for the hint

kripton: @juippis : Done. Fix looks better that way

juippis (approved):
Thanks!

PR #19906, opened 2021-03-12
app-office/libreoffice: 7.1 preparations

a17r:
@chewi I don't know why we have box2d slotting but it is complicating things at least for LO here, so this version bump is dropping it. cmake build system seems to work fine for it and also gives us cmake config modules to find it.


jospezial:
-with-jdk-home or --with-jdk-home ?
Both is in the ebuilds. Does the first one work too?

chewi:
Gah, seriously!? Box2D drops CMake in favour of the awful Premake and then 3 months after I commit the package using a custom Meson build script instead, they go back to CMake again!

What you've done is fine. I don't know how stable the ABI is but the package has never had any revdeps before so you may as well drop the older versions.

chewi (approved):

PR #19909, opened 2021-03-13
app-emulation/libvirt: Specify docdir on configure cmd line

zippy2:
Up until very recently libvirt did not allow specifying docdir
via any (meson) configure option. That is why we patched
meson.build and changed docdir to point where we wanted it to
point. Well, this is changed with upstream commit of
v7.1.0-145-gee4a392dda. Therefore, don't patch the meson.build
file anymore and specify docdir like this:

  -Ddocdir="${EPREFIX}/usr/share/doc/${P}"

### DIFF
--- a/app-emulation/libvirt/libvirt-9999.ebuild
+++ b/app-emulation/libvirt/libvirt-9999.ebuild
@@ -128,7 +128,6 @@ DEPEND="${BDEPEND}
 PATCHES=(
 	"${FILESDIR}"/${PN}-6.0.0-fix_paths_in_libvirt-guests_sh.patch
 	"${FILESDIR}"/${PN}-6.7.0-do-not-use-sysconfig.patch
-	"${FILESDIR}"/${PN}-6.7.0-doc-path.patch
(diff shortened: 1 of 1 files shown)


zippy2:
cc @tamiko @trofi

app-emulation/libvirt/libvirt-9999.ebuild
```diff
@@ -276,6 +275,7 @@ src_configure() {
 
 		--localstatedir="${EPREFIX}/var"
 		-Drunstatedir="${EPREFIX}/run"
+		-Ddocdir="${EPREFIX}/usr/share/doc/${P}"
```
juippis, line 278: We use `${PF}` for our docdirs. LGTM otherwise.

tamiko (approved):

app-emulation/libvirt/libvirt-9999.ebuild
zippy2: Pushed fixed version.

juippis (approved):
Thanks for going the extramile and fixing the dir, we could've easily do that while merging. I was waiting for feedback from other maintainer, but all the better now!

PR #19922, opened 2021-03-13
dev-lang/icon: bump version to 9.5.20i

CheyenneWills:
Update to current upstream version (9.5.20i)

Ebuild updated to correct missing LDFLAG when building libcfuncs as
mentioned in #737296

Bug:https://bugs.gentoo.org/775896

Signed-off-by: Cheyenne Wills <cwills@witznd.net>


dev-lang/icon/icon-9.5.20i.ebuild
```diff
@@ -0,0 +1,124 @@
...
+
+DESCRIPTION="very high level language"
+HOMEPAGE="http://www.cs.arizona.edu/icon/"
+
+SRC_URI="https://github.com/gtownsend/icon/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="public-domain HPND"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~hppa ~ia64 ~ppc ~ppc64 ~s390 ~sparc ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos"
+IUSE="X iplsrc"
```
juippis, line 16: Since the `iplsrc` is only used to control what files to install, would it make sense to always install those files unconditionally?

dev-lang/icon/icon-9.5.20i.ebuild
```diff
@@ -0,0 +1,124 @@
...
+	sed -i -e 's/CFLAGS="$(CFLAGS)"/& LDFLAGS="$(LDFLAGS)"/g' ipl/packs/loadfunc/Makefile || die
+	sed -i -e 's/-o $LIBNAME $CFLAGS/& $LDFLAGS/g' ipl/cfuncs/mklib.sh || die
+	sed -i -e 's/$LIBNAME -fPIC/$LIBNAME -fPIC $LDFLAGS/g' ipl/cfuncs/mklib.sh || die
+	sed -i -e 's/$(RLINK)/& $(LDFLAGS)/g' src/runtime/Makefile || die
+
+	# executables in src/common are only used to prep the build
+	sed -i -e 's/$(CC) $(CFLAGS)/& $(LDFLAGS)/g' src/common/Makefile || die
+	# executables in src/rtt are only used to perform the build
+	sed -i -e 's/$(CC) $(CFLAGS)/& $(LDFLAGS)/g' src/rtt/Makefile || die
+
```
juippis, line 43: This is just.. *sigh*. Any chance, at some other time, to patch these and preferably straight in upstream?

dev-lang/icon/icon-9.5.20i.ebuild
```diff
@@ -0,0 +1,124 @@
...
+	# do not prestrip files
+	find src -name 'Makefile' | xargs sed -i -e "/strip/d" || die
```
juippis, line 45: `find -exec sed` please.

dev-lang/icon/icon-9.5.20i.ebuild
```diff
@@ -0,0 +1,124 @@
...
+
+	cd "${S}/doc" || die
+	DOCS=( *.txt ../README )
+
+	HTML_DOCS=( *.{htm,gif,jpg,css} )
+	einstalldocs
+
+	# Clean up items from make Install that get installed elsewhere
+	rm -rf "${ED}"/usr/$(get_libdir)/${PN}/man || die
+	rm -rf "${ED}"/usr/$(get_libdir)/icon/{doc,README} || die
```
juippis, line 112: We don't use `rm -f` because you should know whether a file exists or not.

dev-lang/icon/icon-9.5.20i.ebuild
```diff
@@ -0,0 +1,124 @@
...
+
+	# optional Icon Programming Library
+	if use iplsrc; then
+		cd "${S}" || die
+
+		# Remove unneeded files before copy
+		rm -fv ipl/{BuildBin,BuildExe,CheckAll,Makefile} || die
```
juippis, line 119: Same here, although I'd just always install these unconditionally.

dev-lang/icon/icon-9.5.20i.ebuild
CheyenneWills: Sure.  I'll update the ebuild

dev-lang/icon/icon-9.5.20i.ebuild
CheyenneWills: Those files are only used at build time and are useless as part of the actual library. I'll make a better comment in the ebuild.

dev-lang/icon/icon-9.5.20i.ebuild
CheyenneWills: ACK

dev-lang/icon/icon-9.5.20i.ebuild
CheyenneWills: ACK

dev-lang/icon/icon-9.5.20i.ebuild
CheyenneWills: I've asked in the past.  The upstream maintainer has been reluctant to make any changes to the build process.  Upstream has been stabilized and has typically only been taking updates when there has been a bug found (build failures, or runtime related bugs).


As a quick side note, any new work on the language has been happening in the Unicon dialect (which is a different project)

juippis (approved):

PR #19925, opened 2021-03-14
LXDE GTK3 updates

a17r:
TODO:
`x11-misc/obconf`, @johu, https://bugs.gentoo.org/562346
~~media-gfx/gpicview, @slashbeast, https://bugs.gentoo.org/642294~~
~~lxde-base/lxsession-0.5.5 fails to build~~


a17r:
@Chiitoo re openbox
@hannob for any potential inside knowledge about the gtk3 situation

jospezial:
While you are at the lxde packages you could bump lxterminal to 0.4.0 and remove the optional gtk2 from the ebuild.
I was getting tired in waiting for my lxde gentoo bug reports processed. So I switched to xfce as alternative besides kde.
Thank you for your work so far. I will install it all in the next days and test.

formula-spectre:
I don't know if this is relevant, But I do have an ebuild for lxpolkit if that is needed

a17r:
lxpolkit is part of lxsession, see https://bugs.gentoo.org/623744

PR #19926, opened 2021-03-14
dev-libs/aws-sdk-cpp: Version Bump to 1.8.160 (#761385 #774276)

Yamakuzure:
This pull requests bumps dev-libs/aws-sdk-cpp itself, plus:

* dev-libs/aws-c-common
* dev-libs/s2n
* dev-libs/aws-c-io and
* dev-libs/aws-c-event-stream

The following issues are fixed:
* dev-libs/aws-c-cal: #761385 : Fix building with USE="static-libs"
* dev-libs/s2n: #774276 : Fix testing

Closes: https://bugs.gentoo.org/761385
Closes: https://bugs.gentoo.org/774276


dev-libs/aws-c-cal/aws-c-cal-0.4.5.ebuild
```diff
@@ -31,6 +31,15 @@ PATCHES=(
 	"${FILESDIR}"/${P}-cmake-prefix.patch
 )
 
+src_prepare() {
+	default
+
+	# Bug #761385 : Build fails with USE="static-libs" on final linking without -lz being added.
+	if use static-libs; then
+		eapply "${FILESDIR}"/${P}-add_libz_for_static.patch
+	fi
```
juippis, line 40: Is there any chance you could build this conditional into the cmake build files instead?

dev-libs/aws-c-common/aws-c-common-0.5.2.ebuild
```diff
@@ -0,0 +1,29 @@
+# Copyright 1999-2020 Gentoo Authors
```
juippis, line 1: How did you manage to push a 2020-copyrighted file with repoman??

dev-libs/s2n/s2n-1.0.0.ebuild
```diff
@@ -0,0 +1,56 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake
+
+DESCRIPTION="simple, small, fast and secure C99 implementation of the TLS/SSL protocols"
```
juippis, line 8:
```suggestion
DESCRIPTION="Simple, small, fast and secure C99 implementation of the TLS/SSL protocols"
```

dev-libs/s2n/s2n-1.0.0.ebuild
```diff
@@ -0,0 +1,56 @@
...
+		-e '$a\\n#if defined(__linux__) && defined(__ELF__)\n.section .note.GNU-stack,"",%progbits\n#endif' \
+		-i "${S}"/pq-crypto/sike_r2/sikep434r2_fp_x64_asm.S || die "sed failed"
+
+	# Fix shared library building, needed for USE="test"
+	# See: https://github.com/awslabs/s2n/issues/2401
+	if use test; then
+		sed -i -e 's, -fvisibility=hidden,,' "${S}"/CMakeLists.txt || die "sed failed"
+		# Remove s2n_self_talk_nonblocking_test, it is broken.
+		# See: https://github.com/awslabs/s2n/issues/2051#issuecomment-744543724
+		rm "${S}"/tests/unit/s2n_self_talk_nonblocking_test.c
```
juippis, line 46:
```suggestion
		rm "${S}"/tests/unit/s2n_self_talk_nonblocking_test.c || die
```

dev-libs/s2n/s2n-1.0.0.ebuild
```diff
@@ -0,0 +1,56 @@
+# Copyright 1999-2020 Gentoo Authors
```
juippis, line 1: Oh yeah here too.

dev-libs/aws-c-common/aws-c-common-0.5.2.ebuild
Yamakuzure: Strange... I always do a "repoman full" beforehand, no idea how that slipped...

dev-libs/aws-c-cal/aws-c-cal-0.4.5.ebuild
Yamakuzure: I'll see what I can do. Shouldn't be too hard. But then it would always get patched. Is that better?

Yamakuzure:
@juippis : Thank you for your feedback. Unfortunately two projects at work exploded in my face, so I had no time to address your pointers. I will look into everything asap.

dev-libs/aws-c-cal/aws-c-cal-0.4.5.ebuild
juippis: Yes please, if you can write it into build-system, it shouldn't break in `-static-libs` or `+static-libs`. Also preferably, as always, get it upstreamed ;)

Yamakuzure:
@juippis : Done. Everything should be fixed, now.

dev-libs/aws-c-event-stream/aws-c-event-stream-0.2.7.ebuild
```diff
@@ -0,0 +1,45 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="static-libs test"
+
+RESTRICT="!test? ( test )"
+
+DEPEND="
+	>=dev-libs/aws-c-common-0.5.2:=[static-libs=]
+	>=dev-libs/aws-c-io-0.9.2:=[static-libs=]
+	>=dev-libs/aws-checksums-0.1.10:=[static-libs=]
```
juippis, line 22: I think `[static-libs=]` is unnecessary and you'll want to use `?` instead.

juippis (approved):

dev-libs/aws-c-event-stream/aws-c-event-stream-0.2.7.ebuild
Yamakuzure: I have taken this from some other ebuilds in the tree. How would these lines be better written? I have to admit I do not know what you mean by "use ? instead" exactly.

juippis: `[static-libs?]` will require `+static-libs` on a dep if it's enabled for this ebuild, while 
`[static-libs=]` will need the dep match to match use flag for this ebuild. So to say, In order for me to test, or anyone to install this package with `+static-libs`, it needs to be enabled on all the specified deps (correct), but if I want to test `-static-libs` or install that afterward, all the deps also need to be recompiled with `-static-libs`. I doubt that is **required**.

Yamakuzure: Ooohhh! Alright, I will fix that with the next bump. Which will be this month already as AWS has a really quick release cycle.
Thank you very much for all your help and support!

PR #19932, opened 2021-03-14
dev-java/asm: EAPI 7 (Slots 3+4) 

vaukai:


kma06 (approved):

dev-java/asm/asm-5.1.ebuild
```diff
@@ -1,23 +1,26 @@
...
 LICENSE="BSD"
 SLOT="4"
-IUSE=""
 KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86 ~x64-macos"
 
-DEPEND=">=virtual/jdk-1.6"
-RDEPEND=">=virtual/jre-1.6"
+CDEPEND=""
+DEPEND=">=virtual/jdk-1.8:*"
+RDEPEND=">=virtual/jre-1.8:*"
```
fordfrog, line 21: idk if i noticed that before on other contributions, but generally, this should be done in revbump because it changes the bytecode of the jar files. if we do it without revbump, the package won't get rebuilt which might cause issues in future, because future jdks might not support 1.6 bytecode and hence the jar might not be loadable by the jdk and package rebuild would be the only fix, which would not be obvious though.

vaukai: Revbump done.  Thanks for clear explanation.

kma06: I don't think the eapi bump should be a new revision because it doesn't break the package.

vaukai: Done anyway. Problems with revbump?

fordfrog:
@vaukai both pushed, thanks!

PR #19934, opened 2021-03-15
net-im/mattermost-desktop-bin: new version

ceamac:
New version and one QA fix


net-im/mattermost-desktop-bin/mattermost-desktop-bin-4.6.2.ebuild
```diff
@@ -0,0 +1,95 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+MY_PN="${PN%-*}"
+
+inherit desktop eutils
```
juippis, line 8: *eutils.eclass* shouldn't be used for ebuilds anymore. I don't see any functions in this ebuild requiring it either.

net-im/mattermost-desktop-bin/mattermost-desktop-bin-4.6.2.ebuild
```diff
@@ -0,0 +1,95 @@
...
+	x11-libs/libXcursor:0
+	x11-libs/libXdamage:0
+	x11-libs/libXext:0
+	x11-libs/libXfixes:0
+	x11-libs/libXi:0
+	x11-libs/libXrandr:0
+	x11-libs/libXrender:0
+	x11-libs/libXScrnSaver:0
+	x11-libs/libXtst:0
+	x11-libs/pango:0"
```
juippis, line 53: What weird dep requirements. Is it still gtk-2?  You have a lot of slots defined on packages with no other slots even available. This could use some serious cleaning.

net-im/mattermost-desktop-bin/mattermost-desktop-bin-4.6.2.ebuild
```diff
@@ -0,0 +1,95 @@
...
+	opt/mattermost-desktop/libffmpeg.so
+	opt/mattermost-desktop/libGLESv2.so
+	opt/mattermost-desktop/libEGL.so
+"
+
+DOCS=(
+	NOTICE.txt
+	LICENSES.chromium.html
+	LICENSE.txt
+	LICENSE.electron.txt
```
juippis, line 67: We don't install license files, as they'd be duplicated. But do check the correct `LICENSE` string for this package, then.

net-im/mattermost-desktop-bin/mattermost-desktop-bin-4.6.2.ebuild
```diff
@@ -0,0 +1,95 @@
...
+	README.md
+)
+
+src_unpack() {
+	unpack ${A}
+	use amd64 && mv "${WORKDIR}/mattermost-desktop-${PV}-linux-x64" "${S}"
+	use x86 && mv "${WORKDIR}/mattermost-desktop-${PV}-linux-ia32" "${S}"
+}
```
juippis, line 75: You can do this in `src_install` without having to define `src_unpack` at all. Also please open up the if-statements and add `|| die` after `mv` command. Although in `src_install` it'd be `cd`.

net-im/mattermost-desktop-bin/mattermost-desktop-bin-4.6.2.ebuild
```diff
@@ -0,0 +1,95 @@
...
+
+	insinto "/opt/${MY_PN}/resources"
+	doins resources/*.asar
+
+	insinto "/opt/${MY_PN}"
+	doins *.pak *.bin *.dat
+	exeinto "/opt/${MY_PN}"
+	doexe *.so "${MY_PN}"
+
+	dosym "../..//opt/${MY_PN}/${MY_PN}" "/usr/bin/${MY_PN}"
```
juippis, line 89: You can also use *eapi8-dosym.eclass* if that's easier.

net-im/mattermost-desktop-bin/mattermost-desktop-bin-4.6.2.ebuild
ceamac: This is a binary package.  I tried to obtain a new dependency list with ldd and equery and I've got an even bigger one, but with gtk3.  I'm not happy with it, so I'll leave it alone until I have more time to clean it.

thesamesam: You should use lddtree or similar, not ldd, because ldd includes transitive dependencies (not direct ones).

net-im/mattermost-desktop-bin/mattermost-desktop-bin-4.6.2.ebuild
ceamac: Doesn't work:
```
 * ERROR: net-im/mattermost-desktop-bin-4.6.2::gentoo failed (prepare phase):
 *   The source directory '/mnt/tmp/portage/net-im/mattermost-desktop-bin-4.6.2/work/mattermost-desktop-bin-4.6.2' doesn't exist
```
But I can set `S` to the desired value instead of renaming the directory.

I could do this in `pkg_setup` instead of `src_unpack`, but I see `pkg_setup` doesn't have sandbox, so I'll avoid it

net-im/mattermost-desktop-bin/mattermost-desktop-bin-4.6.2.ebuild
ceamac: Much better, thanks

net-im/mattermost-desktop-bin/mattermost-desktop-bin-4.6.2.ebuild
juippis: `S="${WORKDIR}"`.

ceamac: I set `S` to `WORKDIR` at global scope, then change it again in *src_install*.  Also `cd` into it.  I do not know if this is acceptable, just wanted to make sure anything still using `S` uses the correct value.

net-im/mattermost-desktop-bin/mattermost-desktop-bin-4.6.2.ebuild
```diff
@@ -0,0 +1,75 @@
...
+S="${WORKDIR}"
+
+src_install() {
+	# adjust `S` based on architecture
+	if use amd64; then
+		S="${WORKDIR}/mattermost-desktop-${PV}-linux-x64"
+	elif use x86; then
+		S="${WORKDIR}/mattermost-desktop-${PV}-linux-ia32"
+	fi
+	cd "${S}" || die
```
juippis, line 56: No you can't change `S` in a local phase, but what you can do and what I was after, is `cd` to proper directory based on the `if-else`.

ceamac: Done it, will force push the commit.

juippis (approved):
Thanks! Sorry about the nitpicks, but it's perfect now!

PR #19935, opened 2021-03-15
app-cdr/dvd+rw-tools: fix bugs

ceamac:
Fix problems compiling with clang.  Bump to EAPI 7.

Also, seems not maintained upstream.

### DIFF
--- a/app-cdr/dvd+rw-tools/dvd+rw-tools-7.1-r4.ebuild
+++ b/app-cdr/dvd+rw-tools/dvd+rw-tools-7.1-r4.ebuild
@@ -1,0 +1,58 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+inherit toolchain-funcs
+
+DESCRIPTION="A set of tools for DVD+RW/-RW drives"
+HOMEPAGE="http://fy.chalmers.se/~appro/linux/DVD+RW/"
+SRC_URI="http://fy.chalmers.se/~appro/linux/DVD+RW/tools/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~ppc ~ppc64 ~sparc ~x86 ~amd64-linux ~x86-linux"
+IUSE=""
+
+RDEPEND="app-cdr/cdrtools"
+DEPEND="${RDEPEND}
+	sys-devel/m4"
+
+PATCHES=(
+	"${FILESDIR}"/${PN}-7.0-sysmacros.patch
+	"${FILESDIR}"/${PN}-7.0-wctomb-r1.patch
+	"${FILESDIR}"/${PN}-7.0-glibc2.6.90.patch
+	"${FILESDIR}"/${PN}-7.0-dvddl-r1.patch
+	"${FILESDIR}"/${PN}-7.0-wexit.patch
+	"${FILESDIR}"/${PN}-7.0-reload.patch
+	"${FILESDIR}"/${PN}-7.1-noevent.patch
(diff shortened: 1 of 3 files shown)


app-cdr/dvd+rw-tools/dvd+rw-tools-7.1-r3.ebuild
```diff
@@ -1,7 +1,7 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
```
juippis, line 4: This requires a revbump and with that, resetting KEYWORDS.

app-cdr/dvd+rw-tools/dvd+rw-tools-7.1-r3.ebuild
```diff
@@ -28,8 +28,11 @@ PATCHES=(
 	"${FILESDIR}"/${PN}-7.1-lastshort.patch
 	"${FILESDIR}"/${PN}-7.1-bluray_srm+pow.patch
 	"${FILESDIR}"/${PN}-7.1-bluray_pow_freespace.patch
+	"${FILESDIR}"/${PN}-7.1-clang.patch
 )
 
+DOCS=( index.html )
```
juippis, line 34: Are there no other docs included? Wouldn't this be better to be installed as `HTML_DOCS`?

app-cdr/dvd+rw-tools/files/dvd+rw-tools-7.1-clang.patch
```diff
@@ -0,0 +1,71 @@
+--- a/dvd+rw-booktype.cpp	2006-06-24 12:21:48.000000000 +0300
++++ b/dvd+rw-booktype.cpp	2021-03-14 10:16:02.533741829 +0200
```
juippis, line 2: Did you find this somewhere? We prefer to keep headers in patch files for traceability.

ceamac: No, I tried to compile it with clang and fixed the bugs.  There is a patch attached to one of the bugs, but it is not complete for clang 11.

app-cdr/dvd+rw-tools/dvd+rw-tools-7.1-r3.ebuild
ceamac: This is the only doc.  And it's a text file, not html.  Just has a weird name

ceamac:
I added 7.1-r4 and set all keywords to unstable.  Please review

app-cdr/dvd+rw-tools/dvd+rw-tools-7.1-r4.ebuild
```diff
@@ -0,0 +1,58 @@
...
+	"${FILESDIR}"/${PN}-7.1-bluray_pow_freespace.patch
+	"${FILESDIR}"/${PN}-7.1-clang.patch
+)
+
+# this is a text file, not html
+DOCS=( index.html )
+
+src_prepare() {
+	# Linux compiler flags only include -O2 and are incremental.
+	sed -i '/FLAGS/s:-O2::' Makefile.m4 || die "failed to sed out FLAGS"
```
juippis, line 39: This delimiter has a tendency of breaking gcc, 
https://bugs.gentoo.org/685160

will fix it but just FYI.

juippis (approved):

app-cdr/dvd+rw-tools/dvd+rw-tools-7.1-r4.ebuild
juippis: Wait, there's nothing wrong with it. Sorry...

PR #19936, opened 2021-03-15
sci-mathematics/topcom: define HTML_DOCS only when doc is in USE.

kiwifb:
Bug: https://bugs.gentoo.org/776310

### DIFF
--- a/sci-mathematics/topcom/topcom-0.17.8-r1.ebuild
+++ b/sci-mathematics/topcom/topcom-0.17.8-r1.ebuild
@@ -24,7 +24,6 @@ RDEPEND="${DEPEND}"
 BDEPEND="app-shells/tcsh"
 
 PATCHES=( "${FILESDIR}/${P}-buildsystem.patch" )
-HTML_DOCS=( "${DISTDIR}/TOPCOM-manual.html" )
 
 src_prepare () {
 	default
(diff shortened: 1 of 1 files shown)


orlitzky:
Actually, can you have it download the HTML file unconditionally? That webpage is more akin to a man page that we install unconditionally than it is to API documentation (which is what we try to hide behind `USE=doc`).

sci-mathematics/topcom/topcom-0.17.8-r1.ebuild
```diff
@@ -24,7 +24,6 @@ RDEPEND="${DEPEND}"
 BDEPEND="app-shells/tcsh"
 
 PATCHES=( "${FILESDIR}/${P}-buildsystem.patch" )
-HTML_DOCS=( "${DISTDIR}/TOPCOM-manual.html" )
```
vaukai, line 27: use doc && HTML_DOCS+=( "${DISTDIR}/TOPCOM-manual.html" )

Instead of 3-liner in src-install()

kiwifb: I am not against that per see but I will say I am uncomfortable with using the `use` construct in global scope. It may just be silly from me. That being said I could put the line as is in `src_install` where I have the block. As to block versus one-liner I am also on the fence on the clarity of the style.

kiwifb:
@orlitzky that's actually a very good argument for its systematic installation. Unless I hear otherwise, I think I will go for that.

PR #19940, opened 2021-03-15
dev-lang/ghc: remove unused patches

mm1ke:


trofi:
Pushed as is. Thank you!

PR #19946, opened 2021-03-15
dev-java/xalan-serializer: EAPI 7

vaukai:

### DIFF
--- a/dev-java/xalan-serializer/xalan-serializer-2.7.2-r1.ebuild
+++ b/dev-java/xalan-serializer/xalan-serializer-2.7.2-r1.ebuild
@@ -1,0 +1,44 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+JAVA_PKG_IUSE="doc source"
+
+inherit java-pkg-2 java-ant-2
+
+MY_PV="$(ver_rs 1- '_')"
+MY_P="xalan-j_${MY_PV}"
+
+DESCRIPTION="DOM Level 3 serializer from Apache Xalan, shared by Xalan and Xerces"
+HOMEPAGE="https://xalan.apache.org/"
+SRC_URI="mirror://apache/xalan/xalan-j/source/${MY_P}-src.tar.gz"
+
+LICENSE="Apache-2.0"
+SLOT="0"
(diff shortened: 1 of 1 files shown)


dev-java/xalan-serializer/xalan-serializer-2.7.2-r1.ebuild
```diff
@@ -0,0 +1,44 @@
...
+DESCRIPTION="DOM Level 3 serializer from Apache Xalan, shared by Xalan and Xerces"
+HOMEPAGE="https://xalan.apache.org/"
+SRC_URI="mirror://apache/xalan/xalan-j/source/${MY_P}-src.tar.gz"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~sparc-solaris ~sparc64-solaris ~x64-solaris ~x86-solaris"
+
+RDEPEND=">=virtual/jre-1.8:*"
+DEPEND=">=virtual/jdk-1.8:*"
```
fordfrog, line 22: it does not compile for me with `openjdk:11` so it should be `DEPEND=virtual/jdk:1.8` instead. it still might fail during runtime though, but that depends, the reason why the compilation fails would have to be investigated.

vaukai: Thanks for telling - just switched to testing with openjdk:11
Adjusted to :1.8

fordfrog:
@vaukai 

i was not happy with the idea of xalan-serializer restricting other packages to `jdk:1.8` because of this issue so i tried to migrate the package to `java-pkg-simple` and it compiles fine. i also checked for changes in the new jar against the old one using these commands:
```
japi-compliance-checker --lib=xalan-serializer /var/tmp/portage/dev-java/xalan-serializer-2.7.2/image/usr/share/xalan-serializer/lib/serializer.jar /var/tmp/portage/dev-java/xalan-serializer-2.7.2-r1/image/usr/share/xalan-serializer/lib/xalan-serializer.jar
pkgdiff /var/tmp/portage/dev-java/xalan-serializer-2.7.2/image/usr/share/xalan-serializer/lib/serializer.jar /var/tmp/portage/dev-java/xalan-serializer-2.7.2-r1/image/usr/share/xalan-serializer/lib/xalan-serializer.jar
```
and i did not find any issues with the new jar.

so the final commit is based on your changes but switched to `java-pkg-simple`.

thank you for the contribution!

PR #19950, opened 2021-03-16
app-mobilephone/heimdall: bumping to r1

ScardracS:
The actual version of Heimdall points to a dead gitlab repo
while active one is on github. Also taking ownership.


juippis (commented):
So are you only updating SRC_URI here? That won't need a revbump. And the source files are mirrored anyway, so if the tarball on Github isn't **an exact** match, you shouldn't update the SRC_URI at all.

ScardracS:
> So are you only updating SRC_URI here? That won't need a revbump. And the source files are mirrored anyway, so if the tarball on Github isn't **an exact** match, you shouldn't update the SRC_URI at all.

Yes, was just a link update

ScardracS:
I will release a new version with fixed links for both 1.4.2 and 9999

PR #19951, opened 2021-03-16
Add proxied="" attributes to metadata.xml for William Breathitt Gray

vilhelmgray:


mgorny:
Thanks for your effort but I think we'll just mass add it for everyone. Do you happen to have some script/oneliner you've used for that, or did you just do it by hand?

vilhelmgray:
> Thanks for your effort but I think we'll just mass add it for everyone. Do you happen to have some script/oneliner you've used for that, or did you just do it by hand?

Ah yes you're right mass adding it would be better so we can probably close this PR. I did it by hand since I only had a few packages, so I unfortunately don't have a script provide. :disappointed:

PR #19953, opened 2021-03-16
dev-util/rebar-bin 3.14.4

00-matt:
Also looking to adopt `dev-util/rebar` and bumping that to version 3 too, currently just trying to tame its build system into working in the network sandbox.


dev-util/rebar-bin/rebar-bin-3.14.3.ebuild
```diff
@@ -0,0 +1,31 @@
...
+MY_P=${MY_PN}-${PV}
+
+DESCRIPTION="A sophisticated build-tool for Erlang projects that follows OTP principles"
+HOMEPAGE="https://github.com/erlang/rebar3"
+
+SRC_URI="https://github.com/erlang/rebar3/releases/download/${PV}/rebar3 -> ${P}"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~hppa ~ia64 ~ppc ~ppc64 ~sparc ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~x64-solaris"
```
juippis, line 16: Where did all these keywords come from? Let the arch teams add them if they want.

00-matt: I copied them from `dev-lang/erlang` since it runs anywhere you can run Erlang. Will remove

juippis (approved):

PR #19965, opened 2021-03-17
media-video/pipewire: Bump to 0.3.25 & ebuild overhaul.

pinkflames:
Please accept the updated ebuild which seeks to both provide a newer
version of PipeWire as well as to improve the user experience by
automatically doing the reasonable amount of integration.

Because the changes are relatively extensive, the were split into
multiple commits but the bulk of changes ended up in a single commit
due to difficulty in untangling them.

In addition to things covered in individual commit messages, it may
be worth raising extra attention to the fact that JACK is no longer
a build or a runtime dependency of PipeWire. Furthermore in the
future PW will probably become another provider of virtual/jack.


pinkflames:
In accordance with [New Gentoo copyright policy explained (draft)](https://mgorny.pl/articles/new-gentoo-copyright-policy-explained.html) I, Niklāvs Koļesņikovs <80783143+pinkflames@users.noreply.github.com>, hereby accept the Gentoo Certificate of Origin.

joakim-tjernlund:
Openrc integration: Will not XDG autostart work here(similar as pulseaudio)?

joakim-tjernlund:
Oh, there is libopenaptx in Gentoo as of now too

media-video/pipewire/pipewire-0.3.23.ebuild
```diff
@@ -0,0 +1,207 @@
...
+}
+
+src_install() {
+	meson_src_install
+
+	insinto /etc/security/limits.d/
+	doins ${limitsdfile}
+
+	if use pipewire-alsa; then
+		dosym ../../../usr/share/alsa/alsa.conf.d/50-pipewire.conf /etc/alsa/conf.d/50-pipewire.conf
```
stefson, line 161: I'm not a dev, but are you sure it's a good idea to use an absolute path here?

media-video/pipewire/pipewire-0.3.23.ebuild
```diff
@@ -0,0 +1,207 @@
...
+	SRC_URI="https://github.com/PipeWire/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~x86"
+fi
+
+DESCRIPTION="Multimedia processing graphs"
+HOMEPAGE="https://pipewire.org/"
+
+LICENSE="LGPL-2.1+"
+SLOT="0/0.3"
+IUSE="alsa aac bluetooth debug doc extra ffmpeg-spa gstreamer jack ldac pipewire-alsa +pipewire-jack systemd test udev v4l vulkan-spa"
```
stefson, line 22: again, I'm not a dev, but maybe you don't want to preselect pipewire-jack here for everyone?

media-video/pipewire/pipewire-0.3.23.ebuild
pinkflames: The only absolute path is _/etc/alsa/conf.d/50-pipewire.conf_ which I assume is kosher as every other ebuild I looked at is doing it that way. As for the relative path of _../../../usr/share/alsa/alsa.conf.d/50-pipewire.conf_, I did try replacing it with  _${EROOT}_ but **repoman full -d -x** was not happy about that, so I can only assume that dosym is inherently EPREFIX aware (which would explain why everyone is using plain _/etc/.._).

media-video/pipewire/pipewire-0.3.23.ebuild
pinkflames: Other than building the emulation support of an already very short build, it has absolutely no drawbacks - no dependencies, no surprising change to system behavior, nothing. In fact, per Gentoo guidelines it very well could be always enabled without a USE flag to toggle it (and I was/am half expecting developers to require just that).

pinkflames:
> Oh, there is libopenaptx in Gentoo as of now too

Thank you for the swift action, @joakim-tjernlund . I have made the changes locally and verified that the build system does pick up the new dependency but **repoman full -d -x** is throwing _dependency.bad_ since this ebuild is currently setting _KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~x86"_  but _media-libs/libopenaptx-0.2.0_ only has _KEYWORDS="~amd64"_.

Looking at documentation, it would appear that I'd need to mask _media-video/pipewire-0.3.23_ on those profiles but that seems rather excessive. Guidance on how to proceed sorely needed.

pinkflames:
@joakim-tjernlund wrote:
> Openrc integration: Will not XDG autostart work here(similar as pulseaudio)?

In principle it could be generated and installed in the same fashion as the limits.d file but after looking at this in-depth I have identified multiple issues with XDG autostart:

1. This would do nothing for people who do not use XDG compliant setups but that's the least of its issues.
2. Installation would need to be conditional on !use systemd because otherwise it would prevent users of XDG compliant desktop environments from using the upstream supported and much more robust startup method. This is due to the autostart file having no obvious way of not starting only for systemd users.
3. PulseAudio uses a different approach where each client application can trigger autospawning (if that hasn't been disabled by the user) - with systemd --user we can reserve the socket, tricking the application into thinking that the daemon is already running, while waiting for either pipewire-pulse and pulseaudio to get ready. But with XDG autostart it would be a race between the first PA client application triggering autospawn and pipewire-pulse setting up its socket.
4. The statement above also means that if autospawning is disabled, any PA using application that starts before PipeWire does, will probably either completely error out or have no audio (untested but seems correct). Needless to say XDG does not have a reliable startup order.

It's probably best to let the user figure out startup ordering on their particular setup.

Finally I'd also like to point out that Gentoo, unlike other distributions, has a policy of not starting deamons or background services automatically. Whether this policy should (or does) apply to the user session and not just system services, is beyond this pull request but it would create a discrepancy between systemd and OpenRC where the former enforces manual enabling while the later would have an unreliable autostart. 

P.S. I'm aware that my version of the ebuild is set on OpenRC systems to automatically start pipewire-media-session and pipewire-pulse along with the main pipewire binary but this merely improves the out of the box user experience and only applies when pipewire binary is started, which happens only if and when user intentionally calls for pipewire to launch.

joakim-tjernlund:
Well, I disagree. it is important to offer an "just works" install too, after all systemd tries to do that.
i think 1 and 2 is not an argument against, just  observations.

3. PW is not PA and I don't think PW has auto spawning?

4. Might be a problem but we don't know(yet)

While XDG might not be ideal I think it is important to offer a "just works" option, certainly not  each user having to search around and edit random files.
At least a USE flag for XDG autostart would be minimum at this point.

joakim-tjernlund:
> > Oh, there is libopenaptx in Gentoo as of now too
> 
> Thank you for the swift action, @joakim-tjernlund . I have made the changes locally and verified that the build system does pick up the new dependency but **repoman full -d -x** is throwing _dependency.bad_ since this ebuild is currently setting _KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~x86"_ but _media-libs/libopenaptx-0.2.0_ only has _KEYWORDS="~amd64"_.
> 
> Looking at documentation, it would appear that I'd need to mask _media-video/pipewire-0.3.23_ on those profiles but that seems rather excessive. Guidance on how to proceed sorely needed.

Ahh, will have a look at that. My fault that just amd64 is there now.

media-video/pipewire/pipewire-0.3.23.ebuild
a17r: There are several `IUSE` in this list that do not tick any box inside `src_configure()` which makes them invalid.
Example: If I have `media-libs/fdk-aac` installed, the pipewire build results in linking to this package despite having `IUSE="-aac"` set.

a17r:
If a new dependency is lacking KEYWORDS, then pipewire needs to drop all those arches. It can be re-keyworded in the same KEYWORDREQ bug with the dependency in question.

a17r:
> At least a USE flag for XDG autostart would be minimum at this point.

- We don't do IUSE for small files.
- What is going to happen when both pulseaudio and pipewire are in XDG autostart?

media-video/pipewire/pipewire-0.3.23.ebuild
a17r: Syntax is correct, without having checked actuall linkage here.

joakim-tjernlund:
> > At least a USE flag for XDG autostart would be minimum at this point.
> 
>     * We don't do IUSE for small files.
> 
>     * What is going to happen when both pulseaudio and pipewire are in XDG autostart?

Ok, but if PA USE flag is set in PW it is meant to replace PA so require that PA is not installed in this case ?

joakim-tjernlund:
> > At least a USE flag for XDG autostart would be minimum at this point.
> 
>     * We don't do IUSE for small files.
> 
>     * What is going to happen when both pulseaudio and pipewire are in XDG autostart?

If XDG autostart is inappropriate, maybe  /etc/X11/xinit/xinitrc.d/90-pipewire script would work better?

a17r:
> If XDG autostart is inappropriate, maybe /etc/X11/xinit/xinitrc.d/90-pipewire script would work better?

That's just another small file and doesn't really address the issue at hand, besides being X11 only.

> Ok, but if PA USE flag is set in PW it is meant to replace PA so require that PA is not installed in this case ?

Even if meanwhile we manage to establish pipewire as an alternative pulseaudio provider in reverse dependencies, in reality, many packages depend on libpulse and will continue to do so in the future. But pulseaudio is not yet split into lib and daemon at this point. A blocker therefore will not only make testing more tedious, it will make installing pipewire almost impossible to do for users. So that's a big no-no.

That means, to achieve a seamless user experience, coordination with pulseaudio maintainers is required, and that will take time.

pinkflames:
> Ok, but if PA USE flag is set in PW it is meant to replace PA so require that PA is not installed in this case ?

The pulseaudio USE flag was removed around version v0.3.21 or v0.3.22 because pipewire-pulse is always built without a way to disable it in the build system. Not to mention that for PA using applications to build and work, they need PA headers and libraries which can only be provided by the media-sound/pulseaudio currently.

> If XDG autostart is inappropriate, maybe /etc/X11/xinit/xinitrc.d/90-pipewire script would work better?

Except this will get completely ignored on Wayland, the tty (some people probably listen to music that way) not to mention the risk of starting a second instance for every person having configured any other startup method except for ~/.xinitrc or i3 config. And before anyone asks, starting pipewire a second time creates a process that prints error messages and does nothing (but does not exit).

> i think 1 and 2 is not an argument against, just observations.

Indeed, I would not consider them road-blockers on their own but that's why I put them as the first two out of four.

>     3. PW is not PA and I don't think PW has auto spawning?

The salient point here is that PipeWire relies on unmodified PulseAudio libraries for its emulation as the API was deemed too complex to readily provide an alternative library implementation (unlike JACK libraries which while not bug free or complete are very much on track). And these PA libraries will always autospawn pulseaudio, not pipewire-jack. 

>     4. Might be a problem but we don't know(yet)

Indeed, the few programs I quickly tested with autospawn = no and without pipewire-pulse or pulseaudio running either fell back to another API or gracefully returned an error about PA refusing connections. But this definitely has the potential to break some setups. Would it be worse than letting PulseAudio autospawn? Perhaps not but these are all bad choices that come from the fact that currently PipeWire targets systemd --user and not the legacy autospawn approach.

As an anecdotal side-note, while testing various approaches to pipewire-pulse, I managed to make either libcanberra or something else inside KDE lose audio output, so the risk of breaking UX is very much there.

As I see it, the only three options for PipeWire on OpenRC that have a reliable outcome is to:

1. Mask pulseaudio USE flag on OpenRC profiles. Extreme, I know, but that's the territory we're in now.
2. Develop an autospawn solution - most obvious would be to provide an eselect pulse module to switch between the real pulseaudio and pipewire-pulse. This would be sketchy but a quick test indicates that this could actually work. User experience, though, would be a mess since it would rely on matching eselect configuration with PipeWire config either emulating or not emulating PA. Not to mention dealing with race conditions where a PA client might want to autospawn pipewire-pulse before pipewire gets started (assuming it's even set up to start).
3. Provide an alternative to systemd --user with socket activation and preferably the same unit file syntax - easily the best approach of the many bad ones but I'm not holding my breath for this.

pinkflames:
@a17r  , with this latest forced push to my branch I believe I have addressed all currently raised concerns that could be reasonably addressed. Waiting for further instructions (or a happy merge).

stefson:
I'm not a native speaker of english, but can please someone who is competent enough review the elog message of the ebuild? It seems a bit gibberish to me, but maybe I'm just not very good at reading it?

a17r:
> @a17r , with this latest forced push to my branch, I believe, I have addressed all of the concerns others have raised so far that could be reasonably resolved. Waiting for further instructions (or a happy merge).

I don't see the `IUSE` situation resolved. We can't have flags that do not effectively work on the build system.

media-video/pipewire/pipewire-0.3.23.ebuild
```diff
@@ -11,15 +11,15 @@ if [[ ${PV} == 9999 ]]; then
 	inherit git-r3
 else
 	SRC_URI="https://github.com/PipeWire/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
-	KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~x86"
+	KEYWORDS="~amd64"
```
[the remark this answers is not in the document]
a17r: > There are several `IUSE` in this list that do not tick any box inside `src_configure()` which makes them invalid.
> Example: If I have `media-libs/fdk-aac` installed, the pipewire build results in linking to this package despite having `IUSE="-aac"` set.

media-video/pipewire/pipewire-0.3.23.ebuild
pinkflames: It should be working now with the latest commit. Let me know if I missed anything.

media-video/pipewire/pipewire-0.3.23.ebuild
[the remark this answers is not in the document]
pinkflames: Unresolving this as the original issue got somehow deatched from the changes. Anyway, IUSE should now be up to code.

louregni:
Hello, a little question what is the reason for using the github mirror of the official repository ? It should be better to use this https://gitlab.freedesktop.org/pipewire/pipewire ?

pinkflames:
> Hello, a little question what is the reason for using the github mirror of the official repository ? It should be better to use this https://gitlab.freedesktop.org/pipewire/pipewire ?

I wasn't able to find any guidelines on using GitHub mirrors over upstream's GitLab. I imagine the development might have started on GitHub and then moved to freedesktop.org. This might be further supported by the fact that the EGIT_BRANCH is outdated. But it could also be a load balancing thing, so I'm differing the question to you, @a17r .

BTW, I doubled checked this morning and I think all IUSE flags _now_ have an effect in src_configure().

joakim-tjernlund:
I am happy with the changes too, it is a big step forward

pinkflames:
BTW, v0.3.24 was released a few hours ago. Can I directly add support for that here before merging this PR or must it be a new one?

media-video/pipewire/pipewire-0.3.23.ebuild
```diff
@@ -0,0 +1,229 @@
...
+		elog
+		elog "Rebooting is strongly recommended to avoid surprises from"
+		elog "remnant PulseAudio daemon auto-spawning and surviving logouts."
+		elog
+	else
+		elog "This ebuild tries to do right by OpenRC users by auto-enabling"
+		elog "PulseAudio replacement."
+		elog "For now, users are strongly advised to edit:"
+		elog "${EROOT}/etc/pulse/client.conf"
+		elog "by uncommenting the following line and seting its value to no:"
```
stefson, line 215: suggestion: and to uncomment the following line 
typo: setting

p.s.: why uncommenting and setting value to no?

stefson (commented):
I made some suggestions to the elog message

media-video/pipewire/pipewire-0.3.23.ebuild
```diff
@@ -0,0 +1,229 @@
...
+		elog "Per Gentoo policy installed systemd units must be manually enabled:"
+		elog "systemctl --user disable pulseaudio.service pulseaudio.socket"
+		elog "systemctl --user enable pipewire.socket pipewire-pulse.socket"
+		elog "systemctl --user enable pipewire-media-session.service"
+		elog
+		elog "Rebooting is strongly recommended to avoid surprises from"
+		elog "remnant PulseAudio daemon auto-spawning and surviving logouts."
+		elog
+	else
+		elog "This ebuild tries to do right by OpenRC users by auto-enabling"
```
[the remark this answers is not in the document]
pinkflames: The original is a bit old fashioned but should be correct according to both [Oxford Learners' Dictionaries](https://www.oxfordlearnersdictionaries.com/definition/english/right_3#right_idm_16) & [macmillan dictionary](https://www.macmillandictionary.com/dictionary/british/do-right-by-someone).

Of course if others also feel that it's a bit too out there, then I'm willing to rephrase it, of course.

media-video/pipewire/pipewire-0.3.23.ebuild
pinkflames: Because PulseAudio like systemd and OpenSSH lists its default values commented out and lets people put in the values they want themselves, so to disable auto-spawn'ing of the pulseaudio daemon the following line:
_; autospawn = yes_
must be turned into:
_autospawn = no_

I felt that using a present participle made the text flow better but if I get the feeling or a GLEP that elog messages should be as simple and easy to understand as possible, then I'll look into turning them into something more like simple English.

As for the typo, sure, I'll fix that. Well spotted.

stefson: I think elog messages should be kept as simple as possible, this at least if you want to include non nativ speakers.

media-video/pipewire/pipewire-0.3.23.ebuild
pinkflames: As a quick heads up, I have since then replaced _dosym_ with _dosym8 -r_ which allows to pass in presumably prefix aware absolute paths and get a relative link to the target as a result.

media-video/pipewire/pipewire-0.3.23.ebuild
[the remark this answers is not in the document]
thesamesam: It's fine either way, yeah.

prometheanfire:
`0.3.24` was just released as well

pinkflames:
I have already asked in this thread, if I can include the ebuild for that in this PR (renaming it accordingly) or if it must be a new PR. But no one has answered that yet.

The ebuild for v0.3.24 essentially matches this ebuild but with the required changes made and the new upstream defaults reflected in the resulting configuration. I have been debugging/poking the corner cases for more than a day now and it seems ready for merging as well.

a17r:
No need for anything else than the latest available version, if that is 0.3.24 now, so be it.

pinkflames:
@a17r , thank you for clearing that up and sorry for that last rebase - I did that before noticing your response. I believe all currently brought up issues have been addressed but of course the ebuild now looks quite a bit different from, so I'll understand if something new has popped up.

stefson:
I disagree in terms of the elog messages, still too vague for non native speakers. This is information is vital, please fix it.

pinkflames:
> I disagree in terms of the elog messages, still too vague for non native speakers. This is information is vital, please fix it.

Please at least highlight exact sections explaining what and why is confusing you. For now I can only assume that you want me to include exact commands for users to invoke for each step. Is that it? In principle I'm okay with that but I never thought that Gentoo users would need to be told how to emerge and then enable either an OpenRC service or a systemd --system unit. Instructions for systemd --user and config file contents are only included because it's perfectly reasonable for people not to be well familiar with them (yet).

I am not a native English speaker either, and I suspect that many open source contributors have English as their second or even third language, too. I understand that not everyone was fortunate like me to get exposed to English while still being a child. For the user facing messages, I try to aim for roughly CEFR B2/C1 (late middle school/early high school) which I think is about right for people using a technically challenging Linux distribution.

media-video/pipewire/pipewire-0.3.24.ebuild
```diff
@@ -0,0 +1,228 @@
...
+		elog "Per Gentoo policy installed systemd units must be manually enabled:"
+		elog "systemctl --user disable pulseaudio.service pulseaudio.socket"
+		elog "systemctl --user enable pipewire.socket pipewire-pulse.socket"
+		elog "systemctl --user enable pipewire-media-session.service"
+		elog
+		elog "Rebooting is strongly recommended to avoid surprises from"
+		elog "remnant PulseAudio daemon auto-spawning and surviving logouts."
+		elog
+	else
+		elog "This ebuild tries to do right by OpenRC users by auto-enabling"
```
stefson, line 210: Again, no one understands this construction, even with a dictionary at hand. Suggestion: This ebuild tries do be helpfull to/for (not sure which preposition is correct here) OpenRc users by

pinkflames: A Gentoo maintainer already said it's fine in this same discussion on previous ebuild. Besides even if someone does not understand the idiom, it should not really affect the following sentence with clear instructions on how to proceed.

media-video/pipewire/pipewire-0.3.24.ebuild
```diff
@@ -0,0 +1,228 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit meson eapi8-dosym
+
+if [[ ${PV} == 9999 ]]; then
+	EGIT_REPO_URI="https://github.com/PipeWire/pipewire.git"
```
pinkflames, line 9: @a17r , @thesamesam , this (and the one below, needing an ACK) are the only outstanding issues right now that anyone has brought up - should the ebuild keep using the GitHub mirror (controlled by the upstream, I think) or their [main GitLab instance](https://gitlab.freedesktop.org/pipewire/pipewire). And if the latter, should we prefer to download the tar.gz or the tar.bz2? I did google for an answer but so far found not even a discussion, much less guidance.

media-video/pipewire/pipewire-0.3.24.ebuild
```diff
@@ -0,0 +1,228 @@
...
+	meson_src_install
+
+	insinto /etc/security/limits.d/
+	doins ${limitsdfile}
+
+	if use pipewire-alsa; then
+		keepdir /etc/alsa/conf.d
+
+		dosym8 -r /usr/share/alsa/alsa.conf.d/50-pipewire.conf /etc/alsa/conf.d/50-pipewire.conf
+		dosym8 -r /usr/share/alsa/alsa.conf.d/99-pipewire-default.conf /etc/alsa/conf.d/99-pipewire-default.conf
```
pinkflames, line 169: @a17r , unless I have missed eapi8-dosym.eclass being considered bad, this ought to satisfy everyone.

media-video/pipewire/pipewire-0.3.24.ebuild
thesamesam: I'd prefer the main GitLab instance and I'd say your choice of tar.gz, tar.bz2, although I always end up picking .gz out of habit (probably should break that habit). I'm not the maintainer though.

Thanks for your efforts here. I know it's devolved into a little bit of a bikeshed but you ended up bringing a lot of improvements which we've been waiting to do, so naturally, that brings a lot of discussion. Your next PR will hopefully not be so hard ;)

Both asturm and I are on IRC (#gentoo-proxy-maint on Freenode) if you find that easier to ask any quick questions by the way.

media-video/pipewire/pipewire-0.3.24.ebuild
```diff
@@ -0,0 +1,228 @@
...
+	elog "It's recommended to install and enable sys-auth/rtkit daemon to allow"
+	elog "PipeWire to obtain restricted realtime capabilities at runtime."
+	elog
+		if use bluetooth; then
+		elog "For oFono support install and enable net-misc/ofono daemon, and"
+		elog "pipewire-media-session should automatically contact it at runtime."
+		elog
+	fi
+	if !use pipewire-alsa; then
+		elog "PulseAudio's ALSA plugin is being used to integrate ALSA applications"
```
thesamesam, line 183: Can we make this a bit shorter? Direct advice or link to the wiki for a discussion on it, rather than too even-handed. I get the point being made here, it's just that users will end up glossing over it and it'll fill a lot of lines in the terminal.

thesamesam: I think I'm happy with the rest otherwise.

pinkflames: That's a fair and good point - I probably did overreact, having spent an entire work day's amount of time over two day period on debugging why all of a sudden the reliable PulseAudio plugin is broken with PipeWire even with known good package versions, USE flag combos and kernels.

media-video/pipewire/pipewire-0.3.24.ebuild
a17r: there was no problem previously either.

pinkflames: Thank you for clearing that up.

media-video/pipewire/pipewire-0.3.24.ebuild
pinkflames: Using upstream seems like a solid choice - I was just worried that it might be considered bad practice since GitHub probably has higher capacity than gitlab.freedesktop.org but, I guess, most users will get them from Gentoo mirrors anyway.

I hope so, too, but considering it will likely involve either getting sys-auth/rtkit to allow for higher RT priority by default (PW now prefers to run at 88) or getting the excellently named [hsphfpd](https://github.com/pali/hsphfpd-prototype), which PW can now use instead of oFono, into Portage tree, this may be the easier PR still.

Thank you for bringing that to my attention, I'll look into using IRC (probably via Matrix).

media-video/pipewire/metadata.xml
```diff
@@ -10,11 +10,20 @@
 		<name>Andreas Sturmlechner</name>
 	</maintainer>
 	<upstream>
-		<remote-id type="github">PipeWire/pipewire</remote-id>
```
pinkflames, line 13: While the <remote-id type="gitlab"> does exist, it almost certainly is meant for referencing to gitlab.com hosted projects and not gitlab.freedesktop.org or other hosts using GitLab software. One would imagine that freedesktop should have its own remote-id type but after looking thrice and in various places, I still could not find such a thing.

media-video/pipewire/metadata.xml
```diff
@@ -10,11 +10,20 @@
...
+		<bugs-to>https://gitlab.freedesktop.org/pipewire/pipewire/-/issues</bugs-to>
+		<changelog>https://gitlab.freedesktop.org/pipewire/pipewire/-/releases</changelog>
```
pinkflames, line 14: Must this be a literal file or would a probably auto-generated URL with fancy HTML also count? It's updated on release but I have not verified it has no subtly changing bits.

media-video/pipewire/metadata.xml
pinkflames: I have confirmed with Gentoo devs that that is, indeed, the case. Closing this.

media-video/pipewire/metadata.xml
pinkflames: Apparently nothing is automatically parsing it, so it shouldn't be a problem.

Whissi:
Sorry, you probably have to rebase. I wasn't aware of this work when I did the 0.3.24 bump.

joakim-tjernlund:
> Sorry, you probably have to rebase. I wasn't aware of this work when I did the 0.3.24 bump.

Just do a -9999 ebuild for now, it is needed as pipewire moves forward quickly with fixes.

t-8ch:
@pinkflames 
> 2. Installation would need to be conditional on !use systemd because otherwise it would prevent users of
> XDG compliant desktop environments from using the upstream supported and much more robust startup method.
> This is due to the autostart file having no obvious way of not starting only for systemd users.

This should be doable with [X-systemd-skip=true](https://systemd.io/DESKTOP_ENVIRONMENTS/)
(If I understand the problem correctly and completely)

pinkflames:
It's not like I understand it completely but, I believe, currently KDE by default is still using their old startup scripts - [although there's now an option to use systemd --user including the generator](https://blog.davidedmundson.co.uk/blog/plasma-and-the-systemd-startup/) - _X-systemd-skip=true_ would, as far as I can tell, not apply to it it before the migration. I'm not sure about how things are on GNOME side of things but Gentoo has a huge number of options on the spectrum from full desktop environments to potentially running bare X11. Not to mention Wayland which only adds more users shells to run. This is what I meant when I said there's no reliable way as far as I could tell, because there's no XDG desktop switch that says "if systemd is the init, do not run me" or better yet "if systemd --user is running, refuse starting this desktop file".

joakim-tjernlund:
> Personally I think this PR has suffered enough bike shedding as it is but if someone wants to make a future PR about this, this could be the most viable option so far. Good luck to them - they may need every ounce of it!

Agreed, a 9999 version of this ebuild should in now. Remaining non systemd issues should be in another PR, once they are known.

pinkflames:
No, I'm doing -r1, because the one in tree already [has a bug open](https://bugs.gentoo.org/777837) that this should fix (just waiting for confirmation from the bug reporter).

pinkflames:
With permission from asturm, I have now included the [upstream commit 8c334fa3abddc8b6cf797fc5dbeb71282a3a60fe](https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/8c334fa3abddc8b6cf797fc5dbeb71282a3a60fe) to provide a current version of PipeWire that's free from a rather annoying bug, while not waiting for upstream to release a new version (which I'm sure they'll do soon anyway).

pinkflames:
Sorry for the previously broken push - I became aware of 0.3.25 release right before going for dinner and not wanting someone to get the jump on me, I hurried to rebase for the new version and forgot to do either repoman full -d -x or pkgcheck scan --net.

Whissi:
Thank you very much. I applied some minor changes (versioned files, udev...).

pinkflames:
Thank you for merging this. Changes were very minor, indeed.

media-video/pipewire/pipewire-0.3.25.ebuild
```diff
@@ -0,0 +1,234 @@
...
+	)
+	jack-client? ( >=media-sound/jack2-1.9.10:2[dbus] )
+	pipewire-alsa? (
+		>=media-libs/alsa-lib-1.1.7
+		|| (
+			media-plugins/alsa-plugins[-pulseaudio]
+			!media-plugins/alsa-plugins
+		)
+	)
+	!pipewire-alsa? ( media-plugins/alsa-plugins[pulseaudio] )
```
ostash, line 71: Hello,

I'm wondering why this is needed. I currently have a setup where all application use exclusively pulseaudio, so I don't need or have pulseaudio plugin for ALSA installed. It works perfectly for many years. 

Can this be removed? What if user do not need any plugin for ALSA, neither pusle nor pipewire?

pinkflames: It would break ALSA applications for great many people - if you don't want the pulseaudio's plugin (which does not pull in pulseaudio itself), then you can set the pipewire-alsa plugin and the native one will be installed. ;)

ostash: Hm, media-sound/pulseaudio has recommended 'alsa-plugin' USE flag for this. It should up to the user whether to enable it or no.

I have it disabled and everything works perfectly. I think media-video/pipewire can do the same and install ALSA plugin only if requested by USE flag (be it plugin from PulseAudio or from Pipewire).

joakim-tjernlund:
Pull in a pipewire-9999 to the tree. Then you will find any remaining problems.
This PR has been sitting here far too long

pinkflames:
@joakim-tjernlund if you rename the existing 0.3.25 to 9999, optionally uncomment the EGIT_BRANCH and remove the docdir patch, you will find that it actually functions as one. As for official live ebuild it would require frequent updates that can only be done by a maintainer with commit access.

thesamesam:
@joakim-tjernlund it is already merged, but yes, @pinkflames is right re a live ebuild. Pain to look after.

ostash:
What if I don't need any plugins for ALSA? :)

media-sound/pulseaudio has recommended '+alsa-plugin' USE flag for this. It should up to the user whether to enable it or no. I have it disabled and everything works perfectly.

I think media-video/pipewire can do the same and install ALSA plugin only if requested by USE flag (be it plugin from PulseAudio or from Pipewire).

Something like:

REQUIRED_USE = "
pipewire-alsa ? ( alsa-plugin )
"

RDEPEND = "
alsa-plugin ? (
pipewire-alsa ? ( media-plugins/alsa-plugins[-pulseaudio] )
!pipewire-alsa? ( media-plugins/alsa-plugins[pulseaudio] )
)
"

media-video/pipewire/pipewire-0.3.25.ebuild
pinkflames: But if you try to use an ALSA application, it will not work. Gentoo generally is not in the business of making dysfunctional or known broken packages for the sake of it - there has to be a technical reason for this. The ALSA plugin is generally a good fallback to have for any program that either lacks a newer interface or has been configured to use the ALSA API, which is a perfectly valid choice. If you still disagree, open a formal [Gentoo bug](https://bugs.gentoo.org).

ostash: Done https://bugs.gentoo.org/785076 :)

joakim-tjernlund:
> @joakim-tjernlund if you rename the existing 0.3.25 to 9999, optionally uncomment the EGIT_BRANCH and remove the docdir patch, you will find that it actually functions as one. As for official live ebuild it would require frequent updates that can only be done by a maintainer with commit access.

Good enough forw, I can just copy & rename the ebuild into my own overlay if I need to. Thanks.
BTW, 0.3.26 is out now.

PR #19966, opened 2021-03-17
sys-auth/seatd: Version bump, 0.5.0

lanodan:


sys-auth/seatd/seatd-0.5.0.ebuild
```diff
@@ -0,0 +1,48 @@
...
+	local emesonargs=(
+		-Dman-pages=enabled
+		-Dwerror=false
+	)
+
+	if use elogind || use systemd; then
+		emesonargs+=( -Dlogind=enabled )
+	else
+		emesonargs+=( -Dlogind=disabled )
+	fi
```
juippis, line 39:
```
option('seatd', type: 'feature', value: 'enabled', description: 'seatd support')
option('builtin', type: 'feature', value: 'disabled', description: 'builtin seatd server')
option('server', type: 'feature', value: 'enabled', description: 'seatd server')
```
Any sense to make any of these toggleable for the user?

juippis (approved):

sys-auth/seatd/seatd-0.5.0.ebuild
lanodan: I don't think they make sense in gentoo, IIRC it's something that will depend from one distro/vendor (ie. static-linking) to another.

PR #19968, opened 2021-03-17
app-mobilephone/heimdall: updated links and RDEPEND

ScardracS:


app-mobilephone/heimdall/heimdall-9999.ebuild
```diff
@@ -3,16 +3,9 @@
...
 
-if [[ ${PV} != 9999 ]]; then
-	SRC_URI="https://gitlab.com/BenjaminDobell/Heimdall/-/archive/v${PV}/Heimdall-v${PV}.tar.bz2 -> ${P}.tar.bz2"
-	KEYWORDS="~amd64"
-	S="${WORKDIR}/Heimdall-v${PV}"
-else
-	inherit git-r3
-	EGIT_REPO_URI="https://gitlab.com/BenjaminDobell/Heimdall.git"
-fi
+EGIT_REPO_URI="github.com/Benjamin-Dobell/Heimdall.git"
```
Polynomial-C, line 8: Why are you doing this? By adding this change you will never be able to copy the live ebuild over to a new release. Usually you have required ebuild changes in the live ebuild first before it hits a new release.

ScardracS: I think this was meant only to check is build is actually 9999 or not and on 1.4.2 obviously it isn't, as there is heimdall-9999!for this

Polynomial-C: Well, I am the former heimdall package maintainer and I wrote that ebuild so I know why this is in the live ebuild :smiley: 
And this change is done in the heimdall-9999 ebuild.
If you insist on doing it by all means do it. it's your decision. But that means you will never be able to simply copy over the `heimdall-9999.ebuild` to a release version. You'd have to do each and every change in both live and non-live ebuilds by hand.

ScardracS: Well, I didn't know you were the maintainer 😅 I will put it back asap

Polynomial-C: I **was** the maintainer. I cannot maintain it anymore due to lack of time. So feel free to take over.

PR #19969, opened 2021-03-17
dev-python/pygal: python 3.9 and tests fix

ZeroPointEnergy:
There are various changes upstream to update the lib, however there is no new release yet. I extracted a minimal patch that fixes the tests and added python 3.9 support.

Closes: https://bugs.gentoo.org/738208


ZeroPointEnergy:
Just saw that the ebuild may have some redundant dependencies defined (setuptools and pytest), removed them. There doesn't seem to be a runtime dependency for setup, at least I can't find any import.

Would be nice if this can get merged. I can't update a package in GURU without it.

ZeroPointEnergy:
Anything I can do to help move this along?

Whissi:
Thank you, looks good. Passing all tests for me.

PR #19970, opened 2021-03-17
net-libs/libyang: Bump to 1.0.225

jsmolic:


net-libs/libyang/libyang-1.0.225.ebuild
```diff
@@ -0,0 +1,45 @@
...
+}
+
+multilib_src_compile() {
+	cmake_src_compile
+
+	multilib_is_native_abi && use doc && cmake_src_compile doc
+}
+
+multilib_src_install_all() {
+	use doc && dodoc -r "${S}"/doc/.
```
SoapGentoo, line 44: `"${S}"/` shouldn't be necessary

SoapGentoo (changes requested):

net-libs/libyang/libyang-1.0.225.ebuild
jsmolic: yup, done

PR #19972, opened 2021-03-17
dev-db/sqlite: use correct prefixes for cross-compilation

dongsupark:
In case `ESYSROOT` is located other than `/`, cross compilation fails due to wrong prefixes used by configure.

For example, when the cross toolchain is under `/build/arm64-usr`, build fails like that:

```
./configure --build=x86_64-pc-linux-gnu --host=aarch64-cros-linux-gnu
--with-sysroot=/build/arm64-usr
--with-readline-inc=-I/build/arm64-usr/usr/include/readline
--enable-tcl
./libtool --mode=compile --tag=CC aarch64-cros-linux-gnu-gcc
-O2 -pipe -mtune=generic -g -I.
-I/build/arm64-usr/var/tmp/portage/dev-db/sqlite-3.33.0/work/sqlite-src-3330000-.arm64/src
-I/usr/include -c sqlite3.c
In file included from /usr/include/features.h:489,
                 from /usr/include/bits/libc-header-start.h:33,
                 from /usr/include/stdint.h:26,
                 from sqlite3.c:13484:
/usr/include/gnu/stubs.h:7:11: fatal error: gnu/stubs-32.h: No such file
or directory
```

That happens because `CFLAGS` includes `-I/usr/include`, which is not the correct location of cross toolchains.
The bare prefix is provided automatically by the option `--enable-tcl`.
It should be actually e.g. `-I/build/arm64-usr/usr/include`.

To make configure look up the correct path, we should specify `--enable-tcl --with-tcl="${ESYSROOT}/usr/$(get_libdir)"`, where e.g. `ESYSROOT` is `/build/arm64-usr`.

To avoid similar cross-toolchain issues, let's pass in `--exec-prefix=${ESYSROOT}/usr` to the configure script.

### DIFF
--- a/dev-db/sqlite/sqlite-3.33.0.ebuild
+++ b/dev-db/sqlite/sqlite-3.33.0.ebuild
@@ -115,6 +115,7 @@ multilib_src_configure() {
 	options+=(
 		--enable-load-extension
 		--enable-threadsafe
+		--exec-prefix="${ESYSROOT}/usr"
 	)
 
 	# Support detection of misuse of SQLite API.
@@ -274,7 +275,10 @@ multilib_src_configure() {
 	options+=($(use_enable static-libs static))
 
 	# tcl, test, tools USE flags.
-	options+=(--enable-tcl)
+	options+=(
+		--enable-tcl
+		--with-tcl="${ESYSROOT}/usr/$(get_libdir)"
+	)
 
 	if [[ "${CHOST}" == *-mint* ]]; then
 		# sys/mman.h not available in MiNTLib.
--- a/dev-db/sqlite/sqlite-3.34.0.ebuild
+++ b/dev-db/sqlite/sqlite-3.34.0.ebuild
@@ -152,6 +152,7 @@ multilib_src_configure() {
 	options+=(
 		--enable-load-extension
 		--enable-threadsafe
+		--exec-prefix="${ESYSROOT}/usr"
 	)
 
 	# Support detection of misuse of SQLite API.
@@ -311,7 +312,10 @@ multilib_src_configure() {
 	options+=($(use_enable static-libs static))
 
 	# tcl, test, tools USE flags.
-	options+=(--enable-tcl)
+	options+=(
+		--enable-tcl
+		--with-tcl="${ESYSROOT}/usr/$(get_libdir)"
+	)
 
 	if [[ "${CHOST}" == *-mint* ]]; then
 		# sys/mman.h not available in MiNTLib.
--- a/dev-db/sqlite/sqlite-3.34.1.ebuild
+++ b/dev-db/sqlite/sqlite-3.34.1.ebuild
@@ -152,6 +152,7 @@ multilib_src_configure() {
 	options+=(
 		--enable-load-extension
 		--enable-threadsafe
+		--exec-prefix="${ESYSROOT}/usr"
 	)
 
 	# Support detection of misuse of SQLite API.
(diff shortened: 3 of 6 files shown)


thesamesam:
@Arfrever

Arfrever:
Please file bug and attach full build log and `config.log`.

According to what I see in `configure.ac`, argument of `--enable-tcl` is not really used if different than `yes` or `no`.
Passing `--enable-tcl=somethingelse` disables detection of Tcl configuration.
There is also `--with-tcl=DIR` option.
```
AC_ARG_ENABLE(tcl, AC_HELP_STRING([--disable-tcl],[do not build TCL extension]),
      [use_tcl=$enableval],[use_tcl=yes])
if test "${use_tcl}" = "yes" ; then
  AC_ARG_WITH(tcl, AC_HELP_STRING([--with-tcl=DIR],[directory containing tcl configuration (tclConfig.sh)]), with_tclconfig=${withval})
  AC_MSG_CHECKING([for Tcl configuration])
  AC_CACHE_VAL(ac_cv_c_tclconfig,[
    # First check to see if --with-tcl was specified.
    if test x"${with_tclconfig}" != x ; then
      if test -f "${with_tclconfig}/tclConfig.sh" ; then
        ac_cv_c_tclconfig=`(cd ${with_tclconfig}; pwd)`
      else
        AC_MSG_ERROR([${with_tclconfig} directory doesn't contain tclConfig.sh])
      fi
    fi

  ...

fi
if test "${use_tcl}" = "no" ; then
  HAVE_TCL=""
else
  HAVE_TCL=1
fi
AC_SUBST(HAVE_TCL)
```

Maybe passing `--with-tcl=/usr/$(get_libdir)` would work?

Arfrever:
Also which version of `dev-lang/tcl` do you use?
If not newest (`8.6.11`), then firstly upgrade to newest (`8.6.11`).

dongsupark:
@Arfrever Thanks for the review.
As you suggested, I have fixed the option to `--with-tcl="${ESYSROOT}/usr/$(get_libdir)`.
Rebased, and added the same code to all versions from 3.33.0 to 3.35.2.
Tested also with `dev-lang/tcl` 8.6.11.

Arfrever:
With what USE flags have you tested? (Relevant are `tcl`, `test`, `tools`.)
Do you get any error without `--exec-prefix="${ESYSROOT}/usr"`?

dongsupark:
In our build environments, I have been always building sqlite with USE flags `-tcl`, `-test`, and `-tools`.
So since the beginning I have hoped `USE=-tcl` could somehow prevent such build errors.
However, no, turning off the `tcl` flag did not help.

`--exec-prefix="${ESYSROOT}/usr` alone did not make any difference, at least in my case.
I suspect, it has something to do with hidden issues deep in the sqlite itself.

Note, as written in the description, the issue occurs only when we have distinct SYSROOT for cross builds. If we simply try to build native binary, then obviously the issue is not visible.

Arfrever:
I meant, do you get error if you use ebuild with added `--with-tcl="${ESYSROOT}/usr/$(get_libdir)"` and NOT added `--exec-prefix="${ESYSROOT}/usr"`?

Arfrever:
`tcl-8.6.9-r1` was the first version with fix to not include `-I/usr/include` in `TCL_INCLUDE_SPEC` variable in `tclConfig.sh` (`${BROOT}/usr/lib64/tclConfig.sh` or `${ESYSROOT}/usr/lib64/tclConfig.sh`).

Is it perhaps the case that your `${ESYSROOT}` now has `>=tcl-8.6.9-r1`, but your `${BROOT}` has `<tcl-8.6.9-r1`?

Arfrever:
You can test my already committed fix with both `USE="-tcl"` and `USE="tcl"`.

PR #19983, opened 2021-03-18
selinux userspace: Updated to 3.2

jpds:
Tested on `amd64` system. I also pruned the `MY_P` as upstream removed the release date from the download URL.


perfinion:
Thanks for this! I ended up doing the MY_P thing based on your first commit before i saw your update. The other change was that the fcommon thing was fixed upstream too and also upstream changed how shlib symbols were done and it now needs `-fno-semantic-interposition`

PR #19984, opened 2021-03-18
dev-lang/rust-{bin}: request ARCH=riscv support

dlan17:
anyone from @gentoo/python team can help review them? I'm not python guru..

I just want someone to take a look at them.

_ please don't merge them directly into portage tree _


dev-vcs/git-pw/git-pw-2.0.0.ebuild
```diff
@@ -0,0 +1,35 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{8..9} )
+DISTUTILS_USE_SETUPTOOLS=bdepend
```
thesamesam, line 7: This is the [default](https://dev.gentoo.org/~mgorny/python-guide/distutils.html#distutils-use-setuptools), so we don't need to set it (it's not a problem if you do though)

dev-vcs/git-pw/git-pw-2.0.0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+fi
+
+DESCRIPTION="A tool for integrating Git with Patchwork"
+HOMEPAGE="https://github.com/getpatchwork/git-pw"
+
+LICENSE="MIT"
+SLOT="0"
+IUSE=""
+
+DEPEND="
```
thesamesam, line 25: These are probably all only RDEPEND, usually Python stuff is filled with runtime-only dependencies

dev-vcs/git-pw/git-pw-2.0.0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+	dev-python/arrow[${PYTHON_USEDEP}]
+	dev-python/click[${PYTHON_USEDEP}]
+	dev-python/requests[${PYTHON_USEDEP}]
+	dev-python/tabulate[${PYTHON_USEDEP}]
+"
+
+RDEPEND="${DEPEND}"
+BDEPEND=""
```
thesamesam, line 33:
```suggestion
```

dev-vcs/git-pw/git-pw-2.0.0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+
```
thesamesam, line 34: Try running [tests](https://dev.gentoo.org/~mgorny/python-guide/distutils.html#enabling-tests)?

dev-vcs/git-pw/git-pw-2.0.0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+LICENSE="MIT"
+SLOT="0"
+IUSE=""
+
+DEPEND="
+	dev-python/arrow[${PYTHON_USEDEP}]
+	dev-python/click[${PYTHON_USEDEP}]
+	dev-python/requests[${PYTHON_USEDEP}]
+	dev-python/tabulate[${PYTHON_USEDEP}]
+"
```
thesamesam, line 30: These dependencies are listed correctly btw :)

dev-vcs/git-publish/git-publish-1.7.0.ebuild
```diff
@@ -0,0 +1,39 @@
...
+SRC_URI="https://github.com/stefanha/git-publish/archive/v${PV}.tar.gz -> $P.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="man"
+
+DEPEND=""
+RDEPEND="${PYTHON_DEPS}
+	>=dev-lang/python-3.8:*
```
thesamesam, line 20:
```suggestion
```

The eclass [handles](https://dev.gentoo.org/~mgorny/python-guide/single.html#basic-use-for-unconditional-python) this for us - -we don't want to interfere with it.

dev-vcs/git-publish/git-publish-1.7.0.ebuild
```diff
@@ -0,0 +1,39 @@
...
+DESCRIPTION="Tool for preparing and storing patch revisions as git tag"
+HOMEPAGE="https://github.com/stefanha/git-publish"
+SRC_URI="https://github.com/stefanha/git-publish/archive/v${PV}.tar.gz -> $P.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="man"
+
+DEPEND=""
```
thesamesam, line 18: We don't need to [define](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-620007.3) some empty variables, so we can drop it

dev-vcs/git-publish/git-publish-1.7.0.ebuild
thesamesam: .. and because we're not adding anything custom, we can actually not define RDEPEND or DEPEND at all as it does it for us (it sets it to ${PYTHON_DEPS} automatically)

dev-vcs/git-publish/git-publish-1.7.0.ebuild
```diff
@@ -0,0 +1,39 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{8..9} )
+inherit python-single-r1
+
+DESCRIPTION="Tool for preparing and storing patch revisions as git tag"
+HOMEPAGE="https://github.com/stefanha/git-publish"
+SRC_URI="https://github.com/stefanha/git-publish/archive/v${PV}.tar.gz -> $P.tar.gz"
```
prometheanfire, line 11: the `$P` should be `${P}`

dev-vcs/git-pw/git-pw-2.0.0.ebuild
```diff
@@ -0,0 +1,35 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{8..9} )
+DISTUTILS_USE_SETUPTOOLS=bdepend
+
+inherit distutils-r1
+if [[ "${PV}" == 99999 ]] ; then
```
thesamesam, line 10: You didn't commit a live ebuild though ;)

dev-vcs/git-pw/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
+		<email>dlan@gentoo.org</email>
+		<name>Yixun Lan</name>
+	</maintainer>
+</pkgmetadata>
```
thesamesam, line 8: If they don't compile anything, add ```<stabilize-allarches/>``` here.

dev-vcs/git-publish/git-publish-1.7.0.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="man"
```
prometheanfire, line 16: I want to say we have a policy of installing some docs all the time, not sure if manpages fall into it.

thesamesam (changes requested):
Looks good. Most of my comments are just informative. The only actual issue (the rest are informative) is fixing the dev-lang/python:* depend (remove it).

Re question in commit message: no need to add python@ project, see [here](https://wiki.gentoo.org/wiki/Project:Python#Adding_packages)

dev-vcs/git-pw/git-pw-2.0.0.ebuild
prometheanfire, line 10: I don't think it's standardized, but I generally see (and use) 4 `9`'s.

dev-vcs/git-pw/git-pw-2.0.0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/getpatchwork/git-pw.git"
+else
+	SRC_URI="https://github.com/getpatchwork/git-pw/archive/${PV}.tar.gz -> $P.tar.gz"
```
prometheanfire, line 14: `${P}` here too

dev-vcs/git-pw/git-pw-2.0.0.ebuild
prometheanfire, line 25: the versions in depend here should probably match requirements.txt

prometheanfire: BDEPEND (and I think DEPEND as well) needs pbr added as a dep (see setup.py)

prometheanfire (changes requested):

dev-vcs/git-publish/git-publish-1.7.0.ebuild
Arfrever: Maybe you confused this situation with situations where `distutils-r1.eclass` is used.
`distutils-r1.eclass` sets e.g. `BDEPEND` or `RDEPEND`, but `python-single-r1.eclass` does not.
If this package has only pure-Python modules (no extension modules written in C/C++/Fortran/...), then the following are still needed:
```
BDEPEND="${PYTHON_DEPS}"
RDEPEND="${PYTHON_DEPS}
	dev-vcs/git
"
```

dev-vcs/git-pw/git-pw-2.0.0.ebuild
dlan17: you right, my typo

dlan17: not intend to commit live ebuild, but just want to keep to logic

dev-vcs/git-pw/git-pw-2.0.0.ebuild
dlan17: not sure how to handle this? any example?

thesamesam: This is a nice [example](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-python/python-utils/python-utils-2.5.6.ebuild#n19)

dev-vcs/git-publish/git-publish-1.7.0.ebuild
thesamesam: ah, yes

dev-vcs/git-pw/git-pw-2.0.0.ebuild
thesamesam: np

dev-vcs/git-publish/git-publish-1.7.0.ebuild
```diff
@@ -0,0 +1,42 @@
...
+inherit python-single-r1 python-utils-r1
+
+DESCRIPTION="Tool for preparing and storing patch revisions as git tag"
+HOMEPAGE="https://github.com/stefanha/git-publish"
+SRC_URI="https://github.com/stefanha/git-publish/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="+man"
```
Arfrever, line 16: After whatever `IUSE=...`, add:
```
REQUIRED_USE="${PYTHON_REQUIRED_USE}"
```

dev-vcs/git-publish/git-publish-1.7.0.ebuild
```diff
@@ -0,0 +1,42 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{8..9} )
+inherit python-single-r1 python-utils-r1
```
Arfrever, line 7: Do not inherit `python-utils-r1.eclass` directly.
`python-single-r1.eclass` already inherits `python-utils-r1.eclass`.

PR #20012, opened 2021-03-19
sys-apps/grep: allow building with libsigsegv (fix test failure on sparc)

thesamesam:
This fixes test failures on sparc in grep 3.6, although 3.5 still fails. This seems to be because of an older copy of gnulib in that version.

I'd say it's too risky to backport gnulib changes to 3.5, although we can do it if @gentoo/base-system wants to. I'd prefer to just skip stabilisation for 3.5 on sparc and stabilise 3.6 when the time is right.

Note that I ended up discovering an independent issue because `filename-lineno.pl` failed on sparc too. This ended up being a real bug in that libpcre upstream doesn't support JIT on sparc64 which manifested in both the test failing and e.g. `grep -P foo` dying with SIGBUS. I've therefore masked dev-libs/libpcre[jit] on sparc64 in fd573f430b05c418f74a5700cc182568123c812a.

### DIFF
--- a/profiles/arch/sparc/package.use.mask
+++ b/profiles/arch/sparc/package.use.mask
@@ -1,6 +1,15 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Sam James <sam@gentoo.org> (2021-03-19)
+# On sparc, we have to force sys-apps/grep[libsigsegv]
+# for accurate stack overflow handling. dev-libs/libsigsegv
+# lacks USE=static-libs, so for now, this means we can't
+# build a static grep on sparc.
+# No real pressing need to add it.
+# bug #768135
+sys-apps/grep static
+
 # Miroslav Šulc <fordfrog@gentoo.org> (2021-03-13)
 # virtual/jdk isn't keyworded here
 media-libs/rubberband jni
--- a/sys-apps/grep/grep-3.5.ebuild
+++ b/sys-apps/grep/grep-3.5.ebuild
@@ -15,7 +15,11 @@ SLOT="0"
 KEYWORDS="~alpha amd64 arm arm64 hppa ~ia64 ~m68k ~mips ppc ppc64 ~riscv ~s390 ~sparc x86 ~x64-cygwin ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~sparc-solaris ~sparc64-solaris ~x64-solaris ~x86-solaris"
 IUSE="nls pcre static"
 
-LIB_DEPEND="pcre? ( >=dev-libs/libpcre-7.8-r1[static-libs(+)] )"
+# We lack dev-libs/libsigsegv[static-libs] for now
+REQUIRED_USE="static? ( !sparc )"
+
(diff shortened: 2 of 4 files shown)


sys-apps/grep/grep-3.5.ebuild
```diff
@@ -13,9 +13,12 @@ SRC_URI="mirror://gnu/${PN}/${P}.tar.xz
 LICENSE="GPL-3+"
 SLOT="0"
 KEYWORDS="~alpha amd64 arm arm64 hppa ~ia64 ~m68k ~mips ppc ppc64 ~riscv ~s390 ~sparc x86 ~x64-cygwin ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~sparc-solaris ~sparc64-solaris ~x64-solaris ~x86-solaris"
-IUSE="nls pcre static"
+IUSE="nls libsigsegv pcre static"
```
Arfrever, line 16: Sort it alphabetically.

sys-apps/grep/grep-3.5.ebuild
```diff
@@ -13,9 +13,12 @@ SRC_URI="mirror://gnu/${PN}/${P}.tar.xz
...
 
-LIB_DEPEND="pcre? ( >=dev-libs/libpcre-7.8-r1[static-libs(+)] )"
+REQUIRED_USE="static? ( !libsigsegv )"
```
Arfrever, line 18: You can add `static-libs` USE flag to `dev-libs/libsigsegv`.
Otherwise you would need ` sys-apps/grep static` in SPARC's `package.use.mask`.

thesamesam: Thanks, I'll do that.

sys-apps/grep/grep-3.5.ebuild
thesamesam: This wasn't intended for review yet ;)

But done, thanks!

profiles/arch/sparc/package.use.mask
```diff
@@ -1,6 +1,13 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Sam James <sam@gentoo.org> (2021-03-19)
+# We can't have USE=static on sparc for now because
+# dev-libs/libsigsegv lacks USE=static-libs, and there's
```
Arfrever, line 6: This explanation is not correct...
It should be something like:
```
# USE=static currently requires USE=-libsigsegv (due to dev-libs/libsigsegv
# not supporting USE=static-libs), but USE=libsigsegv is forced on sparc.
```

thesamesam: Again, this isn't intended to be ready for review yet. This is a rough draft. But the explanation isn't wrong - we could however add more information.

thesamesam:
After discussion in #gentoo-base, we concluded that a conditional-arch dependency was better for now, but to be revisited if further arches need support. So, I've marked this as a draft until I've updated this PR.

thesamesam:
@Polynomial-C

PR #20015, opened 2021-03-20
Fix app-shells/mksh license

lanodan:


app-shells/mksh/mksh-59c.ebuild
```diff
@@ -20,7 +20,12 @@ DESCRIPTION="MirBSD Korn Shell"
 # Host is TLSv1.0-only, keep to http for compatibility with modern browsers
 HOMEPAGE="http://mirbsd.de/mksh"
 
-LICENSE="BSD"
+# See http://www.mirbsd.org/TaC-mksh.txt or ${S}/www/files/TaC-mksh.txt
+# MirOS for most of it
+# BSD for when strlcpy(3) is absent, included as libbsd isn't in @system
```
thesamesam, line 25: I'm a bit confused here. Why not just depend on libbsd?

lanodan: Well musl provides `strlcpy(3)` and looking further I think I might have to patch it to make it use `libbsd` instead.

juippis (approved):

PR #20016, opened 2021-03-20
dev-java/core-specs-alpha: bumping to new version

ScardracS:


tgbugs:
Everything seems ok to me.

tgbugs:
I think you still want SLOT="0.2", otherwise there will be a fight between `0.2.44` and `0.2.56` which should not be installed at the same time on the same system. The slotting for `0.1` vs `0.2` has to do with the different requirements for clojure `1.9` vs `1.10`.

scardracs:
> I think you still want SLOT="0.2", otherwise there will be a fight between `0.2.44` and `0.2.56` which should not be installed at the same time on the same system. The slotting for `0.1` vs `0.2` has to do with the different requirements for clojure `1.9` vs `1.10`.

Thanks for the tip, I will fix it asap

dev-java/core-specs-alpha/core-specs-alpha-0.2.56.ebuild
```diff
@@ -0,0 +1,40 @@
...
+EAPI=7
+
+inherit java-pkg-2 java-ant-2
+
+EGIT_REF="053d789"
+
+MY_PN=${PN//-/.}
+MY_P=${MY_PN}-${PV}
+
+DESCRIPTION="A Clojure library with specs to describe Clojure core macros and functions."
```
juippis, line 13:
```suggestion
DESCRIPTION="A Clojure library with specs to describe Clojure core macros and functions"
```

dev-java/core-specs-alpha/core-specs-alpha-0.2.56.ebuild
```diff
@@ -0,0 +1,40 @@
...
+HOMEPAGE="https://clojure.org/ https://github.com/clojure/core.specs.alpha"
+SRC_URI="https://github.com/clojure/${MY_PN}/archive/${MY_P}.tar.gz -> ${P}.tar.gz"
```
juippis, line 15: You could plaintext the `MY_PN` here too.

dev-java/core-specs-alpha/core-specs-alpha-0.2.56.ebuild
```diff
@@ -0,0 +1,40 @@
...
+
+LICENSE="EPL-1.0"
+SLOT="0.2"
+KEYWORDS="~amd64 ~x86 ~x86-linux"
+
+CDEPEND="dev-java/ant-core:0"
+RDEPEND=">=virtual/jre-1.8:*"
+DEPEND="
+	${CDEPEND}
+	>=virtual/jdk-1.8:*"
```
juippis, line 25: It doesn't look like you need `CDEPEND` here at all.

dev-java/core-specs-alpha/core-specs-alpha-0.2.56.ebuild
```diff
@@ -0,0 +1,40 @@
...
+EANT_EXTRA_ARGS="-Dmaven.build.finalName=${MY_P}"
+
+src_prepare() {
+	default
+	cp "${FILESDIR}/build.xml" . || die
+}
+
+src_install() {
+	java-pkg_newjar "target/${MY_P}.jar"
+	dodoc CONTRIBUTING.md README.md
```
juippis, line 39: These should be matched by the default `einstalldocs` call.

juippis:
@tgbugs can you test this, what do you think?

dev-java/core-specs-alpha/core-specs-alpha-0.2.56.ebuild
scardracs: Moved directly on DEPEND and removed CDEPEND at all

dev-java/core-specs-alpha/core-specs-alpha-0.2.56.ebuild
scardracs: Could you change it on merging? Because I have problems with older branches ATM

scardracs: @juippis fixed with einstalldocs

dev-java/core-specs-alpha/core-specs-alpha-0.2.56.ebuild
```diff
@@ -0,0 +1,44 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit java-pkg-2 java-ant-2
+
+EGIT_REF="053d789"
+
+MY_PN=core.specs.alpha
```
juippis, line 10: You can use `ver_rs` here too, something to consider for the next version bump.

scardracs: So core_specs_alpha you mean?

juippis: No, there's a `ver_rs` function that you can use to manipulate the version separators. So you can simply use it to replace `-` with `.` in `${PN}`.

PR #19991, opened 2021-03-18
dev-lang/clojure: bumping to new version

scardracs:
Bumping to clojure-1.10.3

Package-Manager: Portage-3.0.17, Repoman-3.0.2
Signed-off-by: Marco Scardovi <marco@scardovi.com>

PR #20017, opened 2021-03-20
dev-java/spec-alpha: bumping to new version

scardracs:
Bumping to spec-alpha-0.2.194


scardracs:
@tgbugs I need you help to fix a problem with #19991, as you are the original maintainer of spec-alpha... Maybe you can understand why it isn't working (check error provided by @juippis for more informations)

dev-java/spec-alpha/spec-alpha-0.2.194.ebuild
```diff
@@ -0,0 +1,40 @@
...
+
+EAPI=7
+
+inherit java-pkg-2 java-ant-2
+
+EGIT_REF="fb55378"
+
+MY_P=spec.alpha-${PV}
+
+DESCRIPTION="A Clojure library to describe the structure of data and functions."
```
juippis, line 12:
```suggestion
DESCRIPTION="A Clojure library to describe the structure of data and functions"
```

dev-java/spec-alpha/spec-alpha-0.2.194.ebuild
```diff
@@ -0,0 +1,40 @@
...
+LICENSE="EPL-1.0"
+SLOT="0.2"
+KEYWORDS="~amd64 ~x86 ~x86-linux"
+IUSE=""
+
+COMMON_DEPEND="dev-java/ant-core:0"
+RDEPEND=">=virtual/jre-1.8:*"
+DEPEND="
+	${COMMON_DEPEND}
+	>=virtual/jdk-1.8:*"
```
juippis, line 25: It doesn't look like `COMMON_DEPEND` is needed here at all.

dev-java/spec-alpha/spec-alpha-0.2.194.ebuild
```diff
@@ -0,0 +1,40 @@
...
+EANT_EXTRA_ARGS="-Dmaven.build.finalName=${MY_P}"
+
+src_prepare() {
+	default
+	cp "${FILESDIR}/build.xml" . || die
+}
+
+src_install() {
+	java-pkg_newjar "target/${MY_P}.jar"
+	dodoc CONTRIBUTING.md README.md
```
juippis, line 39: These should be matched by default `einstalldocs` call, but I don't mind either way which you call.

scardracs: So einstalldoc ... instead of dodoc ... ?

juippis: Your choice. There's a possibility that `einstalldocs` matches more doc files, but I haven't looked at the source tarball...

dev-java/spec-alpha/spec-alpha-0.2.194.ebuild
scardracs: Moved dev-java/ant-core:0 to DEPEND

dev-java/spec-alpha/spec-alpha-0.2.194.ebuild
scardracs: Fixed

scardracs:
@tgbugs if you have some notes please tell me

tgbugs:
@scardracs looks ok to me.

juippis (approved):

PR #20019, opened 2021-03-20
dev-java/hamcrest-core: Switch :1.3 to java-pkg-simple

vaukai:
Closes: https://bugs.gentoo.org/642534

### DIFF
--- a/dev-java/hamcrest-core/hamcrest-core-1.3-r2.ebuild
+++ b/dev-java/hamcrest-core/hamcrest-core-1.3-r2.ebuild
@@ -5,7 +5,7 @@ EAPI=7
 
 JAVA_PKG_IUSE="doc source"
 
-inherit java-pkg-2 java-ant-2
+inherit java-pkg-2 java-pkg-simple
 
 MY_PN="hamcrest"
 MY_P="${MY_PN}-${PV}"
@@ -19,16 +19,15 @@ LICENSE="BSD-2"
 SLOT="${PV}"
 KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86 ~ppc-macos ~x64-macos"
 
-DEPEND="~dev-java/hamcrest-generator-${PV}:1.3
-	>=virtual/jdk-1.8:*
+DEPEND=">=virtual/jdk-1.8:*
 	userland_GNU? ( sys-apps/findutils )"
 
 RDEPEND=">=virtual/jre-1.8:*"
+BDEPEND=">=dev-java/hamcrest-generator-${PV}:1.3"
 
-JAVA_ANT_REWRITE_CLASSPATH="true"
-JAVA_ANT_CLASSPATH_TAGS="${JAVA_ANT_CLASSPATH_TAGS} java java-to-jar"
+JAVA_SRC_DIR="${PN}/src"
 
-EANT_BUILD_TARGET="core"
+DOCS=( {CHANGES,LICENSE,README}.txt )
 
 PATCHES=(
 	# https://bugs.gentoo.org/751379
@@ -37,24 +36,26 @@ PATCHES=(
 
 src_prepare() {
 	default
-	# Empty out the contents of the generator target; it has already been built.
-	eapply "${FILESDIR}/hamcrest-1.3-empty_generator.patch"
+	java-pkg_clean
+}
 
-	# Fix problems with Javadoc target.
-	eapply "${FILESDIR}/hamcrest-core-1.3-fix_javadoc.patch"
+src_compile() {
+	java-pkg-simple_src_compile
 
-	find -iname "*.jar" -exec rm -v {} + || die "Unable to clean bundled JAR files"
+	# Need to add this in order to generate "CoreMatchers.java" as with java-ant-2 was triggered by "build.xml"
+	"$(java-config -J)" \
+		-cp $(java-config --with-dependencies --classpath hamcrest-generator:1.3):${PN}.jar \
+		org.hamcrest.generator.config.XmlConfigurator \
+		core-matchers.xml \
+		hamcrest-core/src/main/java \
+		org.hamcrest.CoreMatchers \
+		hamcrest-core/src/main/java
 
-	local cp="build/${P}.jar"
-	cp="${cp}:$(java-pkg_getjars --build-only --with-dependencies hamcrest-generator-${SLOT})"
-	EANT_EXTRA_ARGS="-Dversion=${PV} -Dgentoo.classpath=${cp}"
+	# Compile again, this time including the freshly generated "CoreMatchers.java"
+	java-pkg-simple_src_compile
 }
 
 src_install() {
-	java-pkg_newjar build/${P}.jar ${PN}.jar
-
-	dodoc README.txt CHANGES.txt
-
-	use doc && java-pkg_dojavadoc build/temp/hamcrest-all-${PV}-javadoc.jar.contents
-	use source && java-pkg_dosrc ${PN}/src/main/java/org
+	default
+	java-pkg-simple_src_install
 }
--- a/dev-java/hamcrest-core/files/hamcrest-core-1.1-java-11.patch
+++ b/dev-java/hamcrest-core/files/hamcrest-core-1.1-java-11.patch
@@ -1,5 +1,3 @@
-diff --git a/hamcrest-core/src/main/java/org/hamcrest/core/AllOf.java b/hamcrest-core/src/main/java/org/hamcrest/core/AllOf.java
-index d435303..962c945 100644
 --- a/hamcrest-core/src/main/java/org/hamcrest/core/AllOf.java
 +++ b/hamcrest-core/src/main/java/org/hamcrest/core/AllOf.java
 @@ -1,6 +1,7 @@
@@ -19,8 +17,6 @@ index d435303..962c945 100644
      }
  
      /**
-diff --git a/hamcrest-core/src/main/java/org/hamcrest/core/AnyOf.java b/hamcrest-core/src/main/java/org/hamcrest/core/AnyOf.java
-index e003a7f..4ed9030 100644
 --- a/hamcrest-core/src/main/java/org/hamcrest/core/AnyOf.java
 +++ b/hamcrest-core/src/main/java/org/hamcrest/core/AnyOf.java
 @@ -1,6 +1,7 @@
--- a/dev-java/hamcrest-core/files/hamcrest-core-1.3-java-11.patch
+++ b/dev-java/hamcrest-core/files/hamcrest-core-1.3-java-11.patch
@@ -1,5 +1,3 @@
-diff --git a/hamcrest-core/src/main/java/org/hamcrest/core/AllOf.java b/hamcrest-core/src/main/java/org/hamcrest/core/AllOf.java
-index a206b80..d55cf6b 100644
 --- a/hamcrest-core/src/main/java/org/hamcrest/core/AllOf.java
 +++ b/hamcrest-core/src/main/java/org/hamcrest/core/AllOf.java
 @@ -57,7 +57,7 @@ public class AllOf<T> extends DiagnosingMatcher<T> {
@@ -11,8 +9,6 @@ index a206b80..d55cf6b 100644
      }
  
      /**
-diff --git a/hamcrest-core/src/main/java/org/hamcrest/core/AnyOf.java b/hamcrest-core/src/main/java/org/hamcrest/core/AnyOf.java
-index 55d08dd..2a46631 100644
 --- a/hamcrest-core/src/main/java/org/hamcrest/core/AnyOf.java
 +++ b/hamcrest-core/src/main/java/org/hamcrest/core/AnyOf.java
 @@ -47,7 +47,7 @@ public class AnyOf<T> extends ShortcutCombination<T> {
--- a/dev-java/hamcrest-core/metadata.xml
+++ b/dev-java/hamcrest-core/metadata.xml
@@ -6,7 +6,6 @@
 		<name>Java</name>
 	</maintainer>
 	<upstream>
-		<remote-id type="google-code">hamcrest</remote-id>
 		<remote-id type="github">hamcrest/</remote-id>
 	</upstream>
 </pkgmetadata>


vaukai:
[build.log.gz](https://github.com/gentoo/gentoo/files/6175178/build.log.gz)

fordfrog:
@vaukai wrt `CoreMatchers` class, see this fragment of `build.xml`:
```
    <target name="core" depends="generator" description="Build core Hamcrest library">
        <java-to-jar srcdir="hamcrest-core/src/main/java" modulename="hamcrest-core"/>

        <!-- Generate one class with all static imports -->
        <mkdir dir="build/temp/hamcrest-core/generated-code"/>

        <java classname="org.hamcrest.generator.config.XmlConfigurator"
              fork="yes"
              failonerror="yes"
              classpath="
                build/hamcrest-core-${version}.jar;
                build/hamcrest-generator-${version}.jar;
              ">
            <arg value="core-matchers.xml"/>
            <arg value="hamcrest-core/src/main/java"/>
            <arg value="org.hamcrest.CoreMatchers"/>
            <arg value="build/temp/hamcrest-core/generated-code"/>
        </java>
        <!-- Append to core jar -->
        <java-to-jar srcdir="build/temp/hamcrest-core/generated-code"
                     modulename="hamcrest-core"
                     classpath="build/hamcrest-core-${version}.jar"/>
    </target>
```
it should be possible (and imo should be done) to generate `CoreMatchers` using `hamcrest-generator`. in case you need help with that, just drop me a message.

vaukai:
> it should be possible (and imo should be done) to generate `CoreMatchers` using `hamcrest-generator`. in case you need help with that, just drop me a message.

@fordfrog I've already studied the "build.xml" file yesterday.  No clue how to proceed. Please help

fordfrog:
> @fordfrog I've already studied the "build.xml" file yesterday. No clue how to proceed. Please help

@vaukai this is a rough hint and needs to be adjusted to the ebuild paths and even ebuild style, but this is the command to start with:
```
$(java-config -J) -cp /usr/share/hamcrest-generator-1.3/lib/hamcrest-generator.jar:/usr/share/qdox-1.12/lib/qdox.jar org.hamcrest.generator.config.XmlConfigurator core-matchers.xml hamcrest-core/src/main/java org.hamcrest.CoreMatchers build/temp/hamcrest-core/generated-code
```
it mimics what the `build.xml` says. anyway, it fails on freshly unpacked sources because it also needs the path to the compiled `hamcrest-core` classes. so that needs to be added to `-cp` and it needs to be called after the main compilation is finished so that there are the compiled `hamcrest-core` classes available.

vaukai:
> it mimics what the `build.xml` says. anyway, it fails on freshly unpacked sources because it also needs the path to the compiled `hamcrest-core` classes. so that needs to be added to `-cp` and it needs to be called after the main compilation is finished so that there are the compiled `hamcrest-core` classes available.

@fordfrog It wasn't even clear to me that java-pkg-simple does not use the "build.xml".  Now I see I could completely drop the [hamcrest-1.3-empty_generator.patch](https://github.com/gentoo/gentoo/pull/20019/files#diff-ed92538eb983e45d90810c885ece09b842d671d8d991fd330ce7f00cd39637f8L40-L41).
The [Preparing sources](https://wiki.gentoo.org/wiki/Java_Developer_Guide#Preparing_sources) section in the wiki where the [Modify Build System](https://wiki.gentoo.org/wiki/Java_Developer_Guide#Modify_Build_System) subsection is about "build.xml" should mention that "build.xml" is ignored by [java-pkg-simple](https://wiki.gentoo.org/wiki/Java_Developer_Guide#java-pkg-simple).
For your proposal using java-config I need some time to read and to try.

fordfrog:
@vaukai that `java-config -J` is just a command to return current java executable, to which then the parameters are passed. so this way we can simply execute the same command as ant is instructed to, while it reads the instructions from `build.xml`.

vaukai:
@fordfrog 
Am I misreading what you said?

$ java-config -J -cp /usr/share/hamcrest-generator-1.3/lib/hamcrest-generator.jar:/usr/share/qdox-1.12/lib/qdox.jar:/var/tmp/portage/dev-java/hamcrest-core-1.3-r2/image/usr/share/hamcrest-core-1.3/lib/hamcrest-core.jar org.hamcrest.generator.config.XmlConfigurator core-matchers.xml hamcrest-core/src/main/java org.hamcrest.CoreMatchers build/temp/hamcrest-core/generated-code

/usr/lib64/openjdk-11/bin/java
/usr/lib64/openjdk-11/bin/javac
!!! ERROR: Package /usr/share/hamcrest-generator-1.3/lib/hamcrest-generator.jar-/usr/share/qdox-1.12/lib/qdox.jar-/var/tmp/portage/dev-java/hamcrest-core-1.3-r2/image/usr/share/hamcrest-core-1.3/lib/hamcrest-core.jar was not found!

fordfrog:
@vaukai i suppose ant is responsible for creating `build/temp/hamcrest-core/generated-code`. you can aswell generate it elsewhere, like that `hamcrest-core/src/main/java`. then you need to compile that `org.hamcrest.CoreMatcher` and add it to the jar, or recompile the whole package, which would not be that effective as most of it is already compiled, but it might be easier to implement.

fordfrog:
@vaukai i must admit i was not sure how's that possible that the issue with the current ebuild does not appear in the ebuild using `java-pkg-simple`. it just hit me atm. it's because the `org.hamcrest.CoreMatcher` file was not part of the build process. i suppose it will fail again when building docs. anyway, i still prefer the updated ebuild using `java-pkg-simple` instead of the one using `ant`. to fix the issue, try to replace `<p/>` with '<br/>`. if that helps and there are some remaining issues, just paste them there and i will look at it what's the cause. also, you can push the updated ebuild here so i can easily replicate and possibly test the fix if it would not be trivial.

vaukai:
@fordfrog 
> @vaukai i must admit i was not sure how's that possible that the issue with the current ebuild does not appear in the ebuild using `java-pkg-simple`. it just hit me atm. it's because the `org.hamcrest.CoreMatcher` file was not part of the build process. i suppose it will fail again when building docs. 

It doesn't look that bad atm.  I am still missing the step recompiling CoreMasters.java (need help how to put in the ebuild).  But documentation is generated AND I am getting some warnings which I think should go away with compilation of CoreMatches.java:

`hamcrest-core/src/main/java/org/hamcrest/CustomMatcher.java:9: warning - invalid usage of tag &`
`hamcrest-core/src/main/java/org/hamcrest/CustomMatcher.java:9: warning - invalid usage of tag &`
`hamcrest-core/src/main/java/org/hamcrest/Matcher.java:15: warning - Tag @link: reference not found: CoreMatchers`
`hamcrest-core/src/main/java/org/hamcrest/Matcher.java:19: warning - Tag @see: reference not found: CoreMatchers`

> anyway, i still prefer the updated ebuild using `java-pkg-simple` instead of the one using `ant`. to fix the issue, try to replace `<p/>` with `<br/>`. if that helps and there are some remaining issues, just paste them here and i will look at it what's the cause. also, you can push the updated ebuild here so i can easily replicate and possibly test the fix if it would not be trivial.

I still have (beginners) questions:
How could I definitely know what to put into JAVA_GENTOO_CLASSPATH ? Removing hamcrest-generator from it doesn't seem to break things.  But I'm unsure.
 * java-pkg_ensure-dep: The ebuild is attempting to use hamcrest-generator-1.3, which is not declared with a SLOT in [RP]DEPEND and --build-only wasn't given.

How could I use this information from the build log, where to put that "-Xlint:unchecked"?
`Note: Some input files use unchecked or unsafe operations.`
`Note: Recompile with -Xlint:unchecked for `details.`

fordfrog:
> @fordfrog
> 
> It doesn't look that bad atm. I am still missing the step recompiling CoreMasters.java (need help how to put in the ebuild).

it should be something like this:
```
src_compile()  {
    java-pkg-simple_src_compile
    # explanation here why it's done this way, i.e. generating CoreMasters needs compiled classes and then we need to compile generated CoreMasters and put it in the final jar
    # your code generating CoreMasters.java in the directory of the sources
    java-pkg-simple_src_compile # not sure whether the jar will be updated this way or you first need to drop the one generated from the previous call
}
```

> But documentation is generated AND I am getting some warnings which I think should go away with compilation of CoreMatches.java:
> 
> `hamcrest-core/src/main/java/org/hamcrest/CustomMatcher.java:9: warning - invalid usage of tag &`
> `hamcrest-core/src/main/java/org/hamcrest/CustomMatcher.java:9: warning - invalid usage of tag &`

this seems to be an issue with the javadoc syntax, but it's just a warning.

> `hamcrest-core/src/main/java/org/hamcrest/Matcher.java:15: warning - Tag @link: reference not found: CoreMatchers`
> `hamcrest-core/src/main/java/org/hamcrest/Matcher.java:19: warning - Tag @see: reference not found: CoreMatchers`

this really should disappear when `CoreMatchers` source file exists.

> I still have (beginners) questions:
> How could I definitely know what to put into JAVA_GENTOO_CLASSPATH ? Removing hamcrest-generator from it doesn't seem to break things. But I'm unsure.

obviously from what you wrote, `hamcrest-generator` is used only as a build tool to generate `CoreMatchers`, so it belongs to `BDEPEND` and no need to put it on classpath other than when you generate `CoreMatchers`.

> How could I use this information from the build log, where to put that "-Xlint:unchecked"?
> `Note: Some input files use unchecked or unsafe operations.`
> `Note: Recompile with -Xlint:unchecked for `details.`

[here](https://github.com/gentoo/gentoo/blob/master/eclass/java-pkg-simple.eclass#L348) you can see `src_compile` passes `${JAVAC_ARGS}` to `ejavac` so you have to define in the ebuild that variable:
```
JAVAC_ARGS="-Xlint:unchecked"
```
it should be probably enough to put it in `src_compile`.

dev-java/hamcrest-core/hamcrest-core-1.3-r2.ebuild
```diff
@@ -25,10 +25,10 @@ DEPEND="~dev-java/hamcrest-generator-${PV}:1.3
 
 RDEPEND=">=virtual/jre-1.8:*"
 
-JAVA_ANT_REWRITE_CLASSPATH="true"
-JAVA_ANT_CLASSPATH_TAGS="${JAVA_ANT_CLASSPATH_TAGS} java java-to-jar"
+JAVA_SRC_DIR="${PN}/src"
+JAVA_GENTOO_CLASSPATH="hamcrest-generator-1.3"
```
fordfrog: from what we discussed, `hamcrest-generator` does not belong to classpath. also, it needs to be in `BDEPEND` instead of `DEPEND`.

dev-java/hamcrest-core/hamcrest-core-1.3-r2.ebuild
```diff
@@ -37,24 +37,22 @@ PATCHES=(
...
 }
 
-src_install() {
-	java-pkg_newjar build/${P}.jar ${PN}.jar
+src_compile() {
+	java-pkg-simple_src_compile
 
-	dodoc README.txt CHANGES.txt
+	"$(java-config -J)" \
+		-cp /usr/share/hamcrest-generator-1.3/lib/hamcrest-generator.jar:/usr/share/qdox-1.12/lib/qdox.jar:hamcrest-core.jar \
```
fordfrog, line 47: this should be `-cp $(java-config --with-dependencies --classpath hamcrest-generator:1.3):${PN}.jar`. you do not want to hardcode the paths.

dev-java/hamcrest-core/hamcrest-core-1.3-r2.ebuild
```diff
@@ -37,24 +37,22 @@ PATCHES=(
...
+		org.hamcrest.generator.config.XmlConfigurator \
+		core-matchers.xml \
+		hamcrest-core/src/main/java \
+		org.hamcrest.CoreMatchers \
+		hamcrest-core/src/main/java
```
fordfrog, line 52: i think this just generates the source file, now you need to compile it, using `java-pkg-simple_src_compile` again. make sure the final jar contains what it should contain.

vaukai:
@fordfrog I think postponing the generation of javadocs for the second java-pkg-simple_src_compile could get us rid of the warnings but is kind of overkill.

fordfrog (commented):
> @fordfrog I think postponing the generation of javadocs for the second java-pkg-simple_src_compile could get us rid of the warnings but is kind of overkill.

@vaukai you're right that it should be run only on the second run. for the first run there's the missing class.

dev-java/hamcrest-core/hamcrest-core-1.3-r2.ebuild
vaukai: This I need to compare and to check more in detail.  Thanks for the advise

dev-java/hamcrest-core/hamcrest-core-1.3-r2.ebuild
vaukai: Was already added. And, finally also junit:4 compiles with it.

dev-java/hamcrest-core/hamcrest-core-1.3-r2.ebuild
vaukai: This was already changed.  Just adjusted typo from {P,B}DEPEND

dev-java/hamcrest-core/hamcrest-core-1.3-r2.ebuild
vaukai: @fordfrog 
 There were our latest conversations crossing - Sorry for confusing number of commits

dev-java/hamcrest-core/hamcrest-core-1.3-r2.ebuild
```diff
@@ -37,24 +36,26 @@ PATCHES=(
 
 src_prepare() {
 	default
-	# Empty out the contents of the generator target; it has already been built.
-	eapply "${FILESDIR}/hamcrest-1.3-empty_generator.patch"
+	find -iname "*.jar" -exec rm -v {} + || die "Unable to clean bundled JAR files"
```
fordfrog, line 39: [this](https://github.com/gentoo/gentoo/blob/master/eclass/java-utils-2.eclass#L2925) is the standard way to do it.

vaukai: > [this](https://github.com/gentoo/gentoo/blob/master/eclass/java-utils-2.eclass#L2925) is the standard way to do it.

Are you saying that's also to be removed?  Then I could drop the whole src_prepare() phase ...
But later, gonna take a break now

fordfrog: @vaukai i understand, this ebuild is exhausting... :-) but i guess you learned a lot :-) anyway, to your question, you should call `java-pkg_clean` from `src_prepare`.

vaukai: Yes, done.
Also mentioned here: https://wiki.gentoo.org/wiki/Java_Developer_Guide#Removing_bundled_classes_and_jars

vaukai:
@fordfrog metadata's remote-id is outdated.  Should be replaced with "hamcrest.org"?

fordfrog:
@vaukai thanks!

vaukai:
@fordfrog cc-arches for https://bugs.gentoo.org/777927 please

fordfrog:
@vaukai done!

PR #20021, opened 2021-03-20
dev-qt/qtwebengine: depend on net-libs/nodejs[ssl]

amoskvin:
Closes: https://bugs.gentoo.org/777405

### DIFF
--- a/dev-qt/qtwebengine/qtwebengine-5.15.2_p20210224.ebuild
+++ b/dev-qt/qtwebengine/qtwebengine-5.15.2_p20210224.ebuild
@@ -84,7 +84,7 @@ DEPEND="${RDEPEND}
 	dev-util/gperf
 	dev-util/ninja
 	dev-util/re2c
-	net-libs/nodejs
+	net-libs/nodejs[ssl]
 	sys-devel/bison
 "
 


amoskvin:
GCO fixed.

Chiitoo:
The Chromium ebuild depends on 'inspector' instead (which in turn will require 'ssl'), so that may or may not make more sense.  I never found a reason for 'inspector' to be enabled yet though, so I really can't say for sure, and as such I didn't initially depend on it either.

amoskvin:
I'm not familiar with the internals, but qtwebengine builds fine without
the inspector flag, and Falkon seems to work fine as far as I can tell.

[ebuild   R    ] net-libs/nodejs-15.11.0:0/15::gentoo  USE="snapshot ssl system-ssl -debug -doc -icu -inspector -npm -pax_kernel -system-icu -systemtap -test" CPU_FLAGS_X86="sse2" 0 KiB
[ebuild   R    ] dev-qt/qtwebengine-5.15.2_p20210224:5/5.15::gentoo  USE="alsa jumbo-build pulseaudio system-ffmpeg system-icu widgets -bindist -debug -designer -geolocation -kerberos -test" 0 KiB

Pesa (approved):
The patch looks fine to me. @a17r any comments?

a17r:
@stha09 can you advise whether 'ssl' or 'inspector' should be used here?

PR lgtm except that it needs a rebase for recently added new ~arch snapshot. Then maybe https://github.com/gentoo/qt/pull/237 should be merged first, picked for gentoo.git and afterwards this PR.

stha09:
Chromium requires nodejs inspector at build time, but AFAICS only for DevTools. If qtwebengine does not build DevTools (idk), then USE=ssl would be enough. Otherwise USE=inspector shall be added and that enables USE=ssl anyways.

Chiitoo:
I've tried Qt WebEngine with nodejs[ssl -inspector], but didn't notice anything different with the DevTools (may well have missed something).

PR #20024, opened 2021-03-21
Bump app-misc/worker to 4.7.0 with release signatures: requires 1 new package and 2 eclasses patches

halcon74:
Bug: https://bugs.gentoo.org/777336

The main purpose of the linked bugreport was to add a new package app-crypt/openpgp-keys-worker. 

But, as I am creating the PR on my own, I have the audacity to include in it also:
- the ebuild of app-misc/worker-4.7.0 with some of my **modifications**
- **patches for 2 eclasses**, necessary to combine verifying release signatures and unpacking .zst files (the patches make unpacker and verify-sig work together correctly).

The patches are trivial, I also commented the changes.

As for the modifications in app-misc/worker-4.7.0. - I am not sure, maybe, not all of them can be accepted. The diff with the current 4.6.1-r100.ebuild is so:

```
--- app-misc/worker/worker-4.6.1-r100.ebuild	2021-03-21 01:41:38.593133706 +0300
+++ app-misc/worker/worker-4.7.0.ebuild	2021-03-21 02:45:26.498943535 +0300
@@ -3,44 +3,58 @@
 
 EAPI=7
 
-LUA_COMPAT=( lua5-{1..2} )
+LUA_COMPAT=( lua5-{1..4} )
 
-inherit autotools lua-single
+inherit lua-single unpacker verify-sig
 
 DESCRIPTION="Worker Filemanager: Amiga Directory Opus 4 clone"
 HOMEPAGE="http://www.boomerangsworld.de/cms/worker/"
-SRC_URI="http://www.boomerangsworld.de/cms/worker/downloads/${P}.tar.bz2"
+SRC_URI="mirror://sourceforge/workerfm/${P}.tar.zst
+	verify-sig? ( mirror://sourceforge/workerfm/${P}.tar.zst.asc -> ${P}.tar.zst.sig )"
 
 LICENSE="GPL-2"
 SLOT="0"
-KEYWORDS="amd64 ~arm ~hppa ppc ~ppc64 x86"
-IUSE="avfs debug dbus examples libnotify lua +magic xinerama xft"
+KEYWORDS="~amd64 ~arm ~hppa ~ppc ~ppc64 ~x86"
+IUSE="avfs debug dbus examples libnotify lua +magic openssl +verify-sig xinerama xft"
+
+VERIFY_SIG_OPENPGP_KEY_PATH="/usr/share/openpgp-keys/RalfHoffmann.asc"
+VERIFY_SIG_RUN_DEFAULT_SRC_UNPACK="no"
 
 REQUIRED_USE="lua? ( ${LUA_REQUIRED_USE} )"
 
 RDEPEND="x11-libs/libX11
 	avfs? ( >=sys-fs/avfs-0.9.5 )
-	dbus? (	sys-apps/dbus )
+	dbus? ( dev-libs/dbus-glib )
 	lua? ( ${LUA_DEPS} )
 	magic? ( sys-apps/file )
+	openssl? ( dev-libs/openssl )
 	xft? ( x11-libs/libXft )
 	xinerama? ( x11-libs/libXinerama )"
-DEPEND="${RDEPEND}"
+DEPEND="${RDEPEND}
+	verify-sig? ( >=app-crypt/openpgp-keys-worker-20210320 )
+"
 BDEPEND="virtual/pkgconfig"
 
 DOCS=( AUTHORS ChangeLog INSTALL NEWS README README_LARGEFILES THANKS )
 
 src_prepare() {
 	default
+	# https://bugs.gentoo.org/709450
+	sed -i -r 's/appdatadir = \$\(datadir\)\/appdata/appdatadir = \$\(datadir\)\/metainfo/g' contrib/Makefile.in || die
+}
 
-	# Don't use /usr/share/appdata
-	sed -i -e "s:/appdata:/metainfo:" contrib/Makefile.am || die
-	eautoreconf
+src_unpack() {
+	# first verify signatures
+	verify-sig_src_unpack
+	# finally unpack .tar.zst
+	unpacker_src_unpack
 }
 
 src_configure() {
 	# there is no ./configure flag to disable libXinerama support
 	export ac_cv_lib_Xinerama_XineramaQueryScreens=$(usex xinerama)
+	# there is no ./configure flag to disable openssl support
+	export ac_cv_header_openssl_sha_h=$(usex openssl)
 	econf \
 		--without-hal \
 		--enable-utf8 \
```

- **LUA_COMPAT** - according to the new version in Portage;
- **autotools** removed because I don' t use eautoreconf because I am patching Makefile.in instead of Makefile.am;
- USE flag **openssl** - I made it by analogy with xinerama support (because similar consructions are used in the sources for enabling xinerama and openssl) but I haven't tested the program without this new flag;
- **dev-libs/dbus-gli**b instead of sys-apps/dbus because it is explicitly indicated in the sources, in file INSTALL, line 32.


app-crypt/openpgp-keys-worker/openpgp-keys-worker-20210320.ebuild
```diff
@@ -0,0 +1,23 @@
...
+DESCRIPTION="Ralf Hoffmann's openPGP key used to sign app-misc/worker releases"
+HOMEPAGE="http://www.boomerangsworld.de/cms/about.html"
+SRC_URI="
+	https://keys.openpgp.org/vks/v1/by-fingerprint/F9299EE90A729029E71AF26B667132D0FBC52B37
+		-> RalfHoffmann.asc
+"
+
+LICENSE="public-domain"
+SLOT="0"
+KEYWORDS="amd64 ~arm ~hppa ppc ~ppc64 x86"
```
thesamesam, line 15: I need to check what we do for KEYWORDS for verify-sig but I suspect we need to just copy the keywords of app-misc/worker, "ekeyword ~all", then stabilise with it in future.

app-misc/worker/worker-4.7.0.ebuild
```diff
@@ -0,0 +1,77 @@
...
+RDEPEND="x11-libs/libX11
+	avfs? ( >=sys-fs/avfs-0.9.5 )
+	dbus? ( dev-libs/dbus-glib )
+	lua? ( ${LUA_DEPS} )
+	magic? ( sys-apps/file )
+	openssl? ( dev-libs/openssl )
+	xft? ( x11-libs/libXft )
+	xinerama? ( x11-libs/libXinerama )"
+DEPEND="${RDEPEND}
+	verify-sig? ( >=app-crypt/openpgp-keys-worker-20210320 )
```
thesamesam, line 34: Should this be in BDEPEND instead?

eclass/unpacker.eclass
```diff
@@ -340,6 +340,12 @@ _unpacker() {
 	local m=$(echo "${a}" | tr '[:upper:]' '[:lower:]')
 	a=$(find_unpackable_file "${a}")
 
+	# do nothing with release signatures
```
thesamesam, line 343: We're going to need to get review for this from @mgorny and post it to the gentoo-dev mailing list (for eclass changes, unless he feels it's trivial).

thesamesam:
Nice work.

- Can you commit with 'repoman commit' or 'pkgdev commit' to ensure that commit messages are [right](https://www.gentoo.org/glep/glep-0066.html#commit-messages) and manifests are generated too?
- Can you squash commits so it's one-commit-per-change? (Manifest changes should be squashed the commit that bumps an ebuild.)

app-crypt/openpgp-keys-worker/openpgp-keys-worker-20210320.ebuild
halcon74: "amd64 ~arm ~hppa ppc ~ppc64 x86" - I copied it from app-misc/worker, yes.
Ah! Noticed now that in app-misc/worker I made all testing (added "~"), but forgot to do the same for keys-worker.

app-misc/worker/worker-4.7.0.ebuild
halcon74: Maybe. I don't understand completely when which *DEPEND should be used... :(

eclass/unpacker.eclass
halcon74: Sure! Who will post it?

halcon74:
> Nice work.

Thank you!

>     * Can you commit with 'repoman commit' or 'pkgdev commit' to ensure that commit messages are [right](https://www.gentoo.org/glep/glep-0066.html#commit-messages) and manifests are generated too?

I'll try... Sorry that I haven't mastered that ability yet.

>     * Can you squash commits so it's one-commit-per-change? (Manifest changes should be squashed the commit that bumps an ebuild.)

Yes, I can. I have already worked with git history changes.

So... 

- The commit "add Manifest for app-crypt/openpgp-keys-worker-20210320" shouldn't be separate (repoman will add Manifest) - this is clear.
- All the modifications to app-misc/worker should be separated, starting from the copy of the existing 4.6.1-r100.ebuild, right?
- No need to split 2 blocks of verify-sig.eclass patch to 2 commits?
- Whom should I indicate as the maintainer of keys-worker?

> We need @mgorny to review the eclass changes.

Clear!

halcon74:
I rebased/rewrote all the commits using repoman. 

Fixed DEPEND->BDEPEND and forgotten testing arches for keys/worker.

All the manifests are generated automatically by repoman. 

The two commits editing eclasses are made via git but "simulating" repoman by multiline messages (for uniformity reasons - is it OK, or no?)

PR #20025, opened 2021-03-21
eclass/common-lisp-3:  Fix asd suffix substitution

mjsir911:
Filenames containing the string "asd" such as `iolib.asdf.asd`
are improperly modified by the bash substitution (/) -> `iolibf.asd`
Replacing it with suffix remove (%) fixes this: `iolib.asdf


mjsir911:
cc @nimiux

mjsir911:
cc @ulm when you get the chance :)

PR #20026, opened 2021-03-21
sys-boot/refind: version bump to 0.13.2

sveyret:
At last, here is the version bump for refind 0.13.2
I don’t have a computer to fully test it, so if you are using refind on yours, do not hesitate and try this version
@juippis


sys-boot/refind/refind-0.13.2.ebuild
```diff
@@ -0,0 +1,140 @@
...
+		export EFIARCH=x64
+		export BUILDARCH=x86_64
+	fi
+}
+
+src_prepare() {
+	default
+
+	# bug 598647 - PIE not supported
+	sed -e '/^CFLAGS/s:$: -fno-PIE:' -i Make.common || die
```
juippis, line 48: If I'm reading this right, you're using `:` as sed delimiters. It's known to break things, 
https://bugs.gentoo.org/685160

Rest looks good.

sveyret: Correct. Changed, now. Thanks.

juippis (approved):
Thanks!

PR #20035, opened 2021-03-21
sys-auth/fprintd: bumping to version 1.90.9

scardracs:


sys-auth/fprintd/fprintd-1.90.9.ebuild
```diff
@@ -0,0 +1,76 @@
...
+
+EAPI=7
+
+inherit meson pam systemd
+
+DESCRIPTION="D-Bus service to access fingerprint readers"
+HOMEPAGE="https://gitlab.freedesktop.org/libfprint/fprintd"
+SRC_URI="https://gitlab.freedesktop.org/libfprint/${PN}/-/archive/v${PV}/${PN}-v${PV}.tar.bz2"
+
+LICENSE="GPL-2"
```
juippis, line 12: Seems to be `GPL-2+`. At least the 1.90.8 is also gpl-2+, could you check 0.8 versions too, and update all relevant ebuilds with updated LICENSE field in a **separate** commit.

scardracs: @juippis I have requested stabilization for 1.90.8. Can I change the license without break the stabilization request?

juippis: Yes, it doesn't affect runtime or list of installed files.

scardracs: Ok, I change them now

juippis (approved):

PR #20047, opened 2021-03-21
media-video/pipewire: version bump, add use flags

returntoreality:
This bumps the version, and adds several use flags, which would otherwise result in automatically building the features depending on the presence of the dependencies.


a17r:
https://github.com/gentoo/gentoo/pull/19965 is already covering this.

pinkflames:
I'd just like to point out that tools IUSE is a bad choice, since pw-cat is but one tool out of many, such as the neat pw-top, none of which are controlled by the pw-cat feature. Furthermore pw-cat wouldn't actually get reliably enabled, because internally the build system requires libsndfile to be enabled as well (this got me, too).

Oh, and oFono is a pure runtime dependency (i.e. no build time dependency) that also has negligible impact on build time, so it should not be controlled via IUSE. But you can print an elog to the user, so that they know they have the option to install oFono and get automatic integration with PW.

I hope this doesn't put you off, and we'll see more PRs soon. Cheers!

returntoreality:
> #19965 is already covering this.
I missed that, I will close this pull request, the other pull request does a lot more. Should have checked the other pull requests, I only checked the gentoo bug tracker.

> I'd just like to point out that the _tools_ IUSE is not the best pick, since pw-cat is but one tool out of many, such as the neat pw-top, none of which are controlled by the pw-cat feature. A more appropriate IUSE would be _pw-cat_ or _extra_ (I went with the later). Furthermore pw-cat wouldn't actually get reliably enabled, because internally the build system requires libsndfile to be enabled as well (this got me, too).
Good point. I was looking for a more generic name, extra(s) is indeed a better name.
> 
> Oh, and oFono is a pure runtime dependency (i.e. does not pull in any additional dependencies for building and the resulting binaries work correctly with or without oFono available on the host system) that also has negligible impact on build time, so it should not be controlled via IUSE. But you can print a message to the user via elog function, so that they know they have the option to install oFono and get automatic integration with PW.
I see, yeah that makes sense.
> 
> I hope this doesn't put you off, and we'll see more PRs soon. Cheers!
No worries. Thanks for the feedback!

PR #20050, opened 2021-03-22
sys-libs/timezone-data: adjust /etc/localtime logic

floppym:
Stop using the "Factory" timezone. Programs will use UTC if localtime is
missing. Not installing the Factory timezone as /etc/localtime will
allow us to more easily detect if /etc/localtime exists and is valid.

If /etc/localtime exists as a regular file, copy the new timezone file
over it as was done previously. Emit a warning message encouraging the
user to remove the file.

If /etc/localtime does not exist, but we have a timezone configured in
/etc/timezone, create a symlink and remove /etc/timezone.

The end result of this should be that new installs default to UTC.
If users create /etc/timezone according to the handbook, a symlink will
be created at /etc/localtime and /etc/timezone will be removed.

### DIFF
--- a/sys-libs/timezone-data/timezone-data-2021a.ebuild
+++ b/sys-libs/timezone-data/timezone-data-2021a.ebuild
@@ -150,28 +150,35 @@ configure_tz_data() {
 	fi
 
 	if ! tz=$(get_TIMEZONE) ; then
-		einfo "Assuming your empty ${etc_lt} file is what you want; skipping update."
+		einfo "Assuming your empty ${src} file is what you want; skipping update."
 		return 0
 	fi
+
 	if [[ "${tz}" == "FOOKABLOIE" ]] ; then
-		elog "You do not have TIMEZONE set in ${src}."
-
-		if [[ ! -e "${etc_lt}" ]] ; then
-			cp -f "${EROOT}"/usr/share/zoneinfo/Factory "${etc_lt}"
-			elog "Setting ${etc_lt} to Factory."
-		else
-			elog "Skipping auto-update of ${etc_lt}."
-		fi
+		einfo "You do not have a timezone set in ${src}; skipping update."
 		return 0
 	fi
 
-	if [[ ! -e "${EROOT}/usr/share/zoneinfo/${tz}" ]] ; then
-		elog "You have an invalid TIMEZONE setting in ${src}"
-		elog "Your ${etc_lt} has been reset to Factory; enjoy!"
-		tz="Factory"
+	local tzpath
+
+	if [[ -f ${etc_lt} ]]; then
+		# If a regular file already exists, copy over it.
+		tzpath="${EROOT}/usr/share/zoneinfo/${tz}"
+		ewarn "Found a regular file at ${etc_lt}."
+		ewarn "Some software may expect a symlink instead."
+		ewarn "You may convert it to a symlink by removing the file and running:"
+		ewarn "  emerge --config sys-libs/timezone-data"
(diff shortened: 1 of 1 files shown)


sys-libs/timezone-data/timezone-data-2021a.ebuild
```diff
@@ -150,28 +150,29 @@ configure_tz_data() {
...
+	local tzpath="${EROOT}/usr/share/zoneinfo/${tz}"
+
+	if [[ -f ${etc_lt} ]]; then
+		# If a regular file already exists, copy over it.
+		einfo "Copying ${tzpath} to ${etc_lt}."
+		cp -f "${tzpath}" "${etc_lt}"
+	else
+		# Otherwise, create a symlink and remove the timezone file.
+		einfo "Linking ${tzpath} at ${etc_lt}."
+		if ln -snf "${tzpath}" "${etc_lt}"; then
```
ulm, line 171: Shouldn't this be a relative symlink? /etc/localtime -> ../usr/share/zoneinfo/foo/bar

trofi: +1 for symlink.

sys-libs/timezone-data/timezone-data-2021a.ebuild
```diff
@@ -150,28 +150,29 @@ configure_tz_data() {
 	fi
```
trofi, line 150: Can you update a commit message to clarify the fate of `/etc/timezone`? Now it claims it will also be a symlink. While you are deleting it.

trofi (changes requested):

sys-libs/timezone-data/timezone-data-2021a.ebuild
trofi, line 150: For existing users of a symlink and `/etc/timezone`:

```
	if [[ -L "${etc_lt}" ]] ; then
		einfo "Assuming your ${etc_lt} symlink is what you want; skipping update."
		return 0
	fi
```

Should we suggest users to remove `/etc/localtime` manually (or do it for them)?

sys-libs/timezone-data/timezone-data-2021a.ebuild
floppym: Updated.

sys-libs/timezone-data/timezone-data-2021a.ebuild
floppym: I reworded the last paragraph to make this more clear.

floppym: I added an ewarn to the ebuild with instructions.

floppym:
In case it wasn't obvious, this PR is related to a thread on gentoo-dev.

https://archives.gentoo.org/gentoo-dev/message/417bd48a17042665cfdf2e0a1775ca52

sys-libs/timezone-data/timezone-data-2021a.ebuild
floppym: Oh, I guess you meant when /etc/timezone is a symlink and /etc/localtime is a symlink already.

I guess my thought process is that once /etc/localtime is a symlink, we treat it as the authority and ignore /etc/timezone from that point forward. Eventually, we remove references to /etc/timezone from documentation and just have users create the /etc/localtime symlink themselves (or using `timedatectl set-timezone` from systemd).

trofi (approved):

trofi:
LGTM

PR #20058, opened 2021-03-22
games-engines/odamex: Version bump to 0.9.0

vilhelmgray:
Closes: https://bugs.gentoo.org/777651


vilhelmgray:
I need some assistance on this ebuild. I've unbundled jsoncpp so that we use the system-provided one now. Unfortunately, I get some C++11 errors when trying to compile with the system-provided `/usr/include/jsoncpp/json.h` header.

juippis:
@SoapGentoo could you take a look?

ionenwks:
> I need some assistance on this ebuild. I've unbundled jsoncpp so that we use the system-provided one now. Unfortunately, I get some C++11 errors when trying to compile with the system-provided `/usr/include/jsoncpp/json.h` header.

As far as I'm aware jsoncpp-1.x.x hard-requires c++11, but package uses -std=gnu++98 and bundles a 7 years old jsoncpp.

I tried to build it against system jsoncpp-0.11.0 (released last year) to test and it built fine, but that's not available in ::gentoo and would partly conflict (just the pkgconfig file) with jsoncpp-1.x.x if wanted to slot it.

vilhelmgray:
Should we use the bundled jsoncpp in this case for now then?

juippis:
I don't really see any other viable option here. You should try to get upstream update their code though ;)

vilhelmgray:
I've updated this PR so we link to the internal jsoncpp. I've also opened https://github.com/odamex/odamex/issues/261 to request upstream to allow linking to the system-provided libraries, as well as https://github.com/odamex/odamex/pull/262 to add support for the CMake "Gentoo" build type.

juippis (approved):

PR #18307, opened 2020-11-17
media-video/ffmpeg: Add avresample USE flag.

orbea:
It is useful to be able to build ffmpeg without the deprecated libavresample library to ensure programs on the system continue to work without deprecated ffmpeg features.

PR #18300, opened 2020-11-17
app-shells/yash: Install the binary to /bin like other shells.

orbea:
Its better to have `/bin/yash` like other shells such as `bash` or `mksh` than `/usr/bin/yash`.

PR #20069, opened 2021-03-22
app-text/texlive-core: Fix linker path with slibtool.

orbea:
Fixes https://bugs.gentoo.org/775170

rdlibtool shows the libtool command:
```
rdlibtool --tag=CC --mode=link x86_64-pc-linux-gnu-gcc -Wimplicit -Wreturn-type -O2 -pipe
-Wl,-O1 -Wl,--as-needed -o fixwrites fixwrites.o -L. -lweb2c
```
And the actual command:
```
rdlibtool: link: x86_64-pc-linux-gnu-gcc fixwrites.o -Wimplicit -Wreturn-type -O2 -pipe
-Wl,-O1 -Wl,--as-needed -L./.libs -lweb2c -o .libs/fixwrites
```
The problem is that `-L.` is transformed to `-L./.libs` since slibtool has no way of knowing that `libweb2c.a` static archive not created without libtool. Making the linker path more explicit solves the issue.

### DIFF
--- a/app-text/texlive-core/texlive-core-2020-r12.ebuild
+++ b/app-text/texlive-core/texlive-core-2020-r12.ebuild
@@ -167,6 +167,8 @@ src_prepare() {
 		-e "s,/usr/include /usr/local/include.*echo \$KPATHSEA_INCLUDES.*,${EPREFIX}/usr/include\"," \
 		texk/web2c/configure || die
 
+	sed -i 's/-L\. -lweb2c/libweb2c\.a/' texk/web2c/web2c/Makefile.in || die
+
 	eapply "${WORKDIR}"/patches
 
 	default


midipix:
>> The problem is that -L. is transformed to -L./.libs since slibtool has no way of knowing that libweb2c.a static archive not created without libtool. Making the linker path more explicit solves the issue.

For a package that uses libtool, the ideal fix would be to build libweb2c.a with libtool as well, therefore: build (a static) libweb2c.la, and then, in the link command, replace -L. -lweb2c  with libweb2c.la.

midipix:
> For a package that uses libtool, the ideal fix would be to build libweb2c.a with libtool as well, therefore: build (a static) libweb2c.la, and then, in the link command, replace -L. -lweb2c with libweb2c.la.

Otherwise, for a manually built archive that's in the current working directory: just replace -L. -lweb2c with an explicit libweb2c.a :-)

Zlogene:
Could you please rather report it upstream (to the tex-k mailing list)? We maintain enough texlive hacks already.

orbea:
> For a package that uses libtool, the ideal fix would be to build libweb2c.a with libtool as well, therefore: build (a static) libweb2c.la, and then, in the link command, replace -L. -lweb2c with libweb2c.la.

Too much effort for a build with a broken autoreconf.

> Could you please rather report it upstream (to the tex-k mailing list)? We maintain enough texlive hacks already.

This hack would be easy to apply to either the `Makefile.in` or `Makefile.am` files it autoreconf is ever fixed. Where is the upstream mailing list?

orbea:
> Otherwise, for a manually built archive that's in the current working directory: just replace -L. -lweb2c with an explicit libweb2c.a :-)

Thanks for the advice, I made this change.

Zlogene:
> This hack would be easy to apply to either the `Makefile.in` or `Makefile.am` files it autoreconf is ever fixed. Where is the upstream mailing list?

Though it is better to get upstreamed, but they will not include it into the tl '21, codebase is frozen now I bet (less than 2 weeks until the release).

https://tug.org/mailman/listinfo/tex-k

juippis:
Sorry, the GCO requirement still applies. Please rebase with a proper sign-off line, or please stop wasting everyone's time with these.
https://github.com/gentoo/gentoo/pull/18300#issuecomment-729736581
https://github.com/gentoo/gentoo/pull/18307#issuecomment-729332847

PR #20071, opened 2021-03-22
sci-libs/cantera: pass AR env variable. Add lapack support.

band-a-prend:
* Pass AR env variable to be able to use llvm-ar instead of ar.
Closes: https://bugs.gentoo.org/773541

* Add lapack support

@asarubbo 
please check if this PR (especially first commit) fixes the issue.


juippis (approved):
Can't test the ar fix, but at least lapack works!

juippis:
Actually it seems to fail with `-python` IUSE.

band-a-prend:
@juippis 
tests fail with ` USE=-python` ?

band-a-prend:
> Actually it seems to fail with `-python` IUSE.

Could I fix it without revbump?

band-a-prend:
@juippis 
I fixed it in https://github.com/gentoo/gentoo/pull/20317 pull request
As installation was broken anyway I don't make revision bump this time.

juippis:
Yep, looks good, thanks!

PR #20076, opened 2021-03-23
dev-libs/granite: bump to 6.0.0

xgqt:
Signed-off-by: Maciej Barć <xgqt@riseup.net>


juippis (commented):
This doesn't compile with `+doc`. I see an issue reported about it, 
https://github.com/elementary/granite/issues/482

could you try to figure it out, or disable doc for this version? Leave a note with link to upstream issue so it can be tracked and checked for the next version release.

xgqt:
> This doesn't compile with `+doc`. I see an issue reported about it,
> [elementary/granite#482](https://github.com/elementary/granite/issues/482)
> 
> could you try to figure it out, or disable doc for this version? Leave a note with link to upstream issue so it can be tracked and checked for the next version release.

ack

dev-libs/granite/granite-6.0.0.ebuild
```diff
@@ -0,0 +1,41 @@
...
+RDEPEND="${DEPEND}"
+
+src_prepare() {
+	default
+	vala_src_prepare
+}
+
+src_configure() {
+	meson_src_configure
+}
```
juippis, line 37: Could you visibly disable the `documentation` configure option here, while adding a link above it to the upstream issue like I asked?

I also noticed the upstream meson build system has `introspection` option, but I guess we don't need it since nobody's asked for it yet. But `introspection` is quite popular USE flag in Gentoo.

dev-libs/granite/granite-6.0.0.ebuild
```diff
@@ -0,0 +1,41 @@
...
+
+src_install() {
+	meson_src_install
+}
```
juippis, line 41: This shouldn't be needed in this ebuild.

juippis:
Could you please read what I'm typing here. This is the 3rd time I'm asking to provide an upstream link to the issue, in the ebuild, so we can check its status during the next version bump. It's simple as:
```
src_configure() {
	# docs disabled due to https://github.com/elementary/granite/issues/482
	local emesonargs=(
		-Ddocumentation=false
	)
	meson_src_configure
}
```

juippis (approved):
Thanks!

PR #5768, opened 2017-09-22
dev-lang/python: enable profile guided optimization

InBetweenNames:
This solves bug #615412.  Add a USE flag for PGO
that is off by default, but if enabled, will trigger the appropriate
make target to do a PGO build of the Python interpreter.

Ping @gentoo/python

Package-Manager: Portage-2.3.10, Repoman-2.3.3

PR #20077, opened 2021-03-23
dev-lang/python: Added pgo USE flag

jpds:
Tested on `amd64`, the test benchmarks successfully run with:

- 3.8.7-r1
- 3.8.8
- 3.9.2
- 3.10.0a6-r2

### DIFF
--- a/dev-lang/python/metadata.xml
+++ b/dev-lang/python/metadata.xml
@@ -7,6 +7,7 @@
 </maintainer>
 <use>
 	<flag name="bluetooth">Build Bluetooth protocol support in socket module</flag>
+	<flag name="pgo">Build with profile guided optimizations (PGO)</flag>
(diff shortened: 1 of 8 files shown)


mattst88:
There's an existing bug report in bugzilla: https://bugs.gentoo.org/615412

And an earlier PR that was closed for no good reason: https://github.com/gentoo/gentoo/pull/5768

Might want to take a look at both and read the discussions there.

jpds:
@mattst88 I had looked at those, but they just fizzled out without any consensus (and/or the changes went to the gentooLTO overlay).

I think that this is a reasonable thing to toggle with a simple USE flag for those of us that wish to have the option enabled, hence why I filed this PR.

mattst88:
> @mattst88 I had looked at those, but they just fizzled out without any consensus (and/or the changes went to the gentooLTO overlay).
> 
> I think that this is a reasonable thing to toggle with a simple USE flag for those of us that wish to have the option enabled, hence why I filed this PR.

Oh yeah, I totally agree. I just wanted to make sure you had seen those discussions.

PR #20079, opened 2021-03-23
Make ninja verbose option disableable with user NINJAOPTS

kristjanbb:
It is not even clear if verbose should be in the default NINJAOPTS either?

### DIFF
--- a/eclass/ninja-utils.eclass
+++ b/eclass/ninja-utils.eclass
@@ -47,9 +47,9 @@ eninja() {
 	[[ ${EAPI:-0} != [245] ]] && nonfatal_args+=( -n )
 
 	if [[ -z ${NINJAOPTS+set} ]]; then
-		NINJAOPTS="-j$(makeopts_jobs) -l$(makeopts_loadavg "${MAKEOPTS}" 0)"
+		NINJAOPTS="-v -j$(makeopts_jobs) -l$(makeopts_loadavg "${MAKEOPTS}" 0)"
 	fi
-	set -- ninja -v ${NINJAOPTS} "$@"
+	set -- ninja ${NINJAOPTS} "$@"
 	echo "$@" >&2
 	"$@" || die "${nonfatal_args[@]}" "${*} failed"
 }


orbea:
Wouldn't it be better to have a flag specific for verbosity so that it can be enabled or disabled without affecting other options?

kristjanbb:
> Wouldn't it be better to have a variable specific for verbosity so that it can be enabled or disabled without affecting other options?

Possibly, I only looked at emake as reference. It really only offers MAKEOPTS for all option control. On second look cmake eclass does have CMAKE_VERBOSE (default on), but there aren't really many precedents for this?

orbea:
My impression was that gentoo hasn't spent much time considering allowing non-verbose build output. I think something potentially ideal would a setting to control verbosity for everything and then settings to control verbosity for specific things like `ninja`. For example it might be useful to default to `make` being verbose while `ninja` is not verbose.

Maybe this could be done with a feature too?

kristjanbb:
It's very hard to control (gnu)make in generic way. By default it is very noisy (prints out every command in full) or completely silent (-s). This caused many projects to use custom rules to reduce the noise level (see the linux kernel for example), but at the cost of removing all hope standardized control via make options.

Then you have various make file generators add options to toggle verbosity, autotool/cmake uses commonly uses V=1 or VERBOSE=1 for increased verbosity, but it's not a given that those variables control anything or worse don't break anything.

All I want is have eninja not to unconditionally force verbose with no option of disabling it. It drowns out useful warnings and makes errors tedious to find when building with very high core count. It really adds no value to have verbose forced on.

orbea:
> but at the cost of removing all hope standardized control via make options.

There could be a standard setting and for non-standard projects they could set the appropriate variable conditionally in the ebuild.

I do agree that focusing on ninja first would be fine.

floppym (changes requested):
Please update the commit summary to the format specified in [GLEP 66](https://www.gentoo.org/glep/glep-0066.html#commit-messages).

For example:
```
ninja-utils.eclass: move -v option into NINJAOPTS

This allows the user to disable verbose output by setting NINJAOPTS in make.conf.
```

Also, this is a change to a widely used eclass, so you will need to send your patch to the gentoo-dev mailing list for formal review.

thesamesam:
Closing per above (feedback pending).

PR #20081, opened 2021-03-23
dev-java/cglib: bump to 3.3.0 

vaukai:
Version bump.


dev-java/cglib/cglib-3.1-r1.ebuild
```diff
@@ -0,0 +1,66 @@
...
+
+	# Get rid of some parts in XML.
+	eapply "${FILESDIR}"/${P}-build.xml.patch
+
+	java-pkg_clean
+	cd lib || die 'cannot cd'
+	java-pkg_jarfrom ant-core
+	java-pkg_jarfrom asm-4
+	if use test ; then
+		java-pkg_jarfrom junit-4
```
fordfrog, line 44: i get this warning when compiling the package:
```
 * java-pkg_ensure-dep: The ebuild is attempting to use junit-4, which is not declared with a SLOT in [RP]DEPEND and --build-only wasn't given.
```

dev-java/cglib/cglib-3.1-r1.ebuild
```diff
@@ -0,0 +1,66 @@
...
+
+DESCRIPTION="cglib is a powerful, high performance and quality Code Generation Library"
+HOMEPAGE="https://github.com/cglib/cglib"
+SRC_URI="mirror://sourceforge/${PN}/${PN}-src-${PV}.jar"
+
+LICENSE="Apache-2.0"
+SLOT="3"
+KEYWORDS="~amd64 ~arm64 ~ppc64 ~x86"
+
+RESTRICT="!test? ( test )"
```
fordfrog, line 18: i think this should be `RESTRICT="test"` if you want to disable the tests completely. with your syntax, it's still possible to run (and fail) tests for this package. i searched the devmanual but i must admit it does not provide enough info to use this feature in the right way, imo.

dev-java/cglib/cglib-3.1-r1.ebuild
vaukai: Yep, got it.

dev-java/cglib/cglib-3.1-r1.ebuild
vaukai: Other proxy-maintainer advised me to use this in a recent case. Got many of examples searching:
`qgrep 'RESTRICT="!test?'`

Have you any idea how to fix this testing bug? [build.log.gz](https://github.com/gentoo/gentoo/files/6197077/build.log.gz)
Or should we really turn off the tests here?
Surprisingly with EAPI 5 and Java 11 it seems to work ...

fordfrog: i was already looking at it before but i really have no idea what is it trying to say... anyway, it is a generally known problem, you can find some info for example [here](https://stackoverflow.com/questions/15122890/java-lang-verifyerror-expecting-a-stackmap-frame-at-branch-target-jdk-1-7#15122960). you can try to pass that `-noverify` to java for the tests to see if it has any effect.

vaukai: After trying many things now I prefer to use `RESTRICT="test"` as you suggested.  USE flag will go.

fordfrog:
@vaukai 

i found there is already 3.3.0 released: https://github.com/cglib/cglib/releases/tag/RELEASE_3_3_0

also, as this package provides `pom.xml`, i strongly suggest you install and get yourself familiarized with our tool called `java-ebuilder` (install live ebuild, that's the most up-to-date version). it will save you a lot of time to prepare the ebuild, you'll also see what `MAVEN_ID` is really used for. the generated ebuild will need a polishment for sure, but it's really a time-saver.

to save you some time, here are two commands that will come handy in this case (and ofc `java-ebuilder` command itself, without any args, will give you usage help):

```
java-ebuilder --refresh-cache --portage-tree <path_to_your_portage_tree>
```
this one goes through all ebuilds, finds those that use java, and extracts information from the ebuilds including `MAVEN_ID` variable. the path can be either your system repo or your git clone, it's not really important. you can see content of the cache at `~/.java-ebuilder/cache`.

```
java-ebuilder --generate-ebuild --dump-projects --download-uri https://github.com/cglib/cglib/archive/refs/tags/RELEASE_3_3_0.tar.gz --ebuild cglib-3.3.0.ebuild --keywords "~amd64 ~arm64 ~ppc64 ~x86" --pom cglib-RELEASE_3_3_0/cglib/pom.xml --workdir .
```
this will generate a rough ebuild that then needs polishing and testing.

if you need more assistance, i'm here, just ask. i really appreciate your help.

vaukai:
> ```
> java-ebuilder --generate-ebuild --dump-projects --download-uri https://github.com/cglib/cglib/archive/refs/tags/RELEASE_3_3_0.tar.gz --ebuild cglib-3.3.0.ebuild --keywords "~amd64 ~arm64 ~ppc64 ~x86" --pom cglib-RELEASE_3_3_0/cglib/pom.xml --workdir .
> ```
> 
> this will generate a rough ebuild that then needs polishing and testing.

@fordfrog 
Last time I tried using `java-ebuilder`I could not find out how to use that tool.  And this time with your kind explanation I hope to get it working here.

`/var/db/repos/vaukai/dev-java/cglib $ java-ebuilder --generate-ebuild --dump-projects --download-uri https://github.com/cglib/cglib/archive/refs/tags/RELEASE_3_3_0.tar.gz --ebuild cglib-3.3.0.ebuild --keywords "~amd64 ~arm64 ~ppc64 ~x86" --pom cglib-RELEASE_3_3_0/cglib/pom.xml --workdir .`
`ERROR: POM file /var/db/repos/vaukai/dev-java/cglib/cglib-RELEASE_3_3_0/cglib/pom.xml does not exist.`

So it seems the package needs to be downloaded and unpacked and I need to cd into workdir before running that command.
To have it downloaded and unpacked I usually have the basic ebuild and run ```ebuild cglib-3.3.0.ebuild clean unpack```
How are you preparing the workdir for running `java-ebuilder` there? Are you also starting with an ebuild?  
Would that tool overwrite an existing ebuild?
Is workdir the recommended directory from where to run that command?

fordfrog:
@vaukai 
my process is roughly this - i do this outside the portage tree, in a dedicated directory, can be `~/tmp` or anything that suits you:
1. i update the `java-ebuilder` cache
2. i find and download the source tarball
3. i check whether all files in the tarball are under a top directory, as i want to unpack all the files in one sub-directory and not in the current directory
4. i unpack the tarball
5. then i search for the right `pom.xml` - for example in this case there are many so i have to choose the right one
6. then i set up `java-ebuilder` parameters and run the command in the directory where the sub-directory with the tarball files exists - it creates ebuild in the current directory, unless i set a different path
7. then i inspect the output of the command and if there are any deps that are note resolved, i investigate whether i can add `MAVEN_ID` to those - if so, i do that, refresh cache and run generation of the ebuild again until i'm satisfied with the output
8. then i inspect the content of the ebuild, delete stuff that is not needed etc... once done, i copy the ebuild to my gentoo.git repo, run `repoman manifest`, `repoman full` and `pkgcheck scan` - if everything is fine, i test to emerge the ebuild and fine-tune it

vaukai:
> @vaukai
> my process is roughly this - i do this outside the portage tree, in a dedicated directory, can be `~/tmp` or anything that suits you:
> 

@fordfrog 
That's a big step forward. Here's my first result:  [cglib-3.3.0.ebuild.gz](https://github.com/gentoo/gentoo/files/6206073/cglib-3.3.0.ebuild.gz)

According to the [pom.xml](https://github.com/cglib/cglib/blob/9a14a6cc0b2ef50e69839b754d32ba4d8675d3ab/pom.xml#L16) I would have expected:
`MAVEN_ID="cglib:cglib-parent:3.3.0`
But why does it put 
`MAVEN_ID="cglib:cglib:3.3.0"`?
Would I have to read it from [D E P E N D E N C Y  M A N A G E M E N T](https://github.com/cglib/cglib/blob/9a14a6cc0b2ef50e69839b754d32ba4d8675d3ab/pom.xml#L215-L219) ?

>     1. i update the `java-ebuilder` cache
> 
>     2. i find and download the source tarball
> 

Here, as I gave it the download URI, why can't it download and unpack into a `work` directory automatically? Such like portage does?

In the generated ebuild it tells me:
`# org.apache.ant:ant:1.10.3 -> !!!groupId-not-found!!!`
`# org.ow2.asm:asm:7.1 -> !!!groupId-not-found!!!`

Does this mean the `MAVEN_ID` in the `ant-1.10.3` and `asm-7.1` ebuilds need to be adjusted (spoofed)?   Or we have to put dependencies manually because asm uses gradle and so does not provide such maven identifiers?

Another question:
In `SRC_URI` ( -> ${P}-**sources**.tar.gz" ) the `sources` part confuses me.  Why is it there, where does it come from?

I will see how I can use the ebuild, which adjustments it needs.

fordfrog:
> @fordfrog
> That's a big step forward. Here's my first result: [cglib-3.3.0.ebuild.gz](https://github.com/gentoo/gentoo/files/6206073/cglib-3.3.0.ebuild.gz)
> 
> According to the [pom.xml](https://github.com/cglib/cglib/blob/9a14a6cc0b2ef50e69839b754d32ba4d8675d3ab/pom.xml#L16) I would have expected:
> `MAVEN_ID="cglib:cglib-parent:3.3.0`
> But why does it put
> `MAVEN_ID="cglib:cglib:3.3.0"`?
> Would I have to read it from [D E P E N D E N C Y M A N A G E M E N T](https://github.com/cglib/cglib/blob/9a14a6cc0b2ef50e69839b754d32ba4d8675d3ab/pom.xml#L215-L219) ?

each `pom.xml` can reference a parent `pom.xml` but what you need is `<project>/<groupId>` and `<project>/<artifactId>`. `<project>/<groupId>` might also be inherited from the parent, like in this case.

> > ```
> > 1. i update the `java-ebuilder` cache
> > 
> > 2. i find and download the source tarball
> > ```
> 
> Here, as I gave it the download URI, why can't it download and unpack into a `work` directory automatically? Such like portage does?

the download uri is just for putting it into the ebuild, it does not download anything. that's just how it works.

> In the generated ebuild it tells me:
> `# org.apache.ant:ant:1.10.3 -> !!!groupId-not-found!!!`
> `# org.ow2.asm:asm:7.1 -> !!!groupId-not-found!!!`
> 
> Does this mean the `MAVEN_ID` in the `ant-1.10.3` and `asm-7.1` ebuilds need to be adjusted (spoofed)? Or we have to put dependencies manually because asm uses gradle and so does not provide such maven identifiers?

in these cases i usually look up the appropriate `MAVEN_ID` and put it in the relevant ebuild. you can also put the dep in the generated ebuild manually but then you can't benefit from the `java-ebuilder` cache look up feature.

> Another question:
> In `SRC_URI` ( -> ${P}-**sources**.tar.gz" ) the `sources` part confuses me. Why is it there, where does it come from?

`java-ebuilder` supports working both with sources and with binary files downloaded from maven repository, so it is here to distinguish those two files.

> I will see how I can use the ebuild, which adjustments it needs.

vaukai:
Thanks for answering my questions

> in these cases i usually look up the appropriate `MAVEN_ID` and put it in the relevant ebuild. you can also put the dep in the generated ebuild manually but then you can't benefit from the `java-ebuilder` cache look up feature.

Here, what exactly would you do?  How important is an exact version match?
- cglib wants `ant:1.10.3` but Gentoo has `ant-core-1.10.9`
- cglib wants `asm:7.1` but Gentoo has `asm-7.3.1`

I will force-push what I have now. It works but fails tests with 7 errors.  Need help to fix them.
And I will leave it as draft for the moment

fordfrog:
@vaukai minor version difference is not an issue at all, so `>=1.10` is ok, and as we don't have older versions, you can drop the version restriction completely. wrt `asm`, you can try to build against the latest and if there is a compilation issue, you can switch to the `:7` slot in this case. wrt the tests, i'll look at them later.

dev-java/cglib/cglib-3.3.0.ebuild
```diff
@@ -0,0 +1,50 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+JAVA_PKG_IUSE="source test"
```
fordfrog, line 6: should have also `doc` flag

dev-java/cglib/cglib-3.3.0.ebuild
```diff
@@ -0,0 +1,50 @@
...
+	${CDEPEND}
+"
+
+S="${WORKDIR}"
+
+JAVA_GENTOO_CLASSPATH="asm-7 ant-core"
+JAVA_SRC_DIR="${MY_P}/${PN}/src/main/java"
+JAVA_RESOURCE_DIRS=(
+	"${MY_P}/${PN}/src/main/resources"
+	"${MY_P}"
```
fordfrog, line 46: this one might not be needed, might be a bug in `java-ebuilder`

fordfrog:
@vaukai might also want to change the pr subject

fordfrog:
@vaukai

```
1) testSql(net.sf.cglib.proxy.TestEnhancer)
...
Caused by: java.lang.ClassNotFoundException: net.sf.cglib.proxy.Factory
...
```
this class is missing from the tests classpath - try to find out why
```
2) testPackagePrivateMethod_bootstrapClassLoader(net.sf.cglib.reflect.TestFastClass)
java.lang.NoSuchMethodException: java.lang.String.getChars([C, int)
	at java.base/java.lang.Class.getDeclaredMethod(Class.java:2475)
	at net.sf.cglib.reflect.TestFastClass.testPackagePrivateMethod_bootstrapClassLoader(TestFastClass.java:663)
```
i'd patch this test out
```
3) initializationError(net.sf.cglib.TestGenerator)
org.junit.runners.model.InvalidTestClassError: Invalid test class 'net.sf.cglib.TestGenerator':
  1. Test class should have exactly one public zero-argument constructor
  2. No runnable methods
...
```
this is not a test class, exclude it using `JAVA_TEST_EXCLUDES`.
```
4) testSql(net.sf.cglib.proxy.TestEnhancer)
...
Caused by: java.lang.ClassNotFoundException: net.sf.cglib.proxy.Factory
...
```
the same as 1)
```
5) initializationError(net.sf.cglib.proxy.TestInterceptor)
java.lang.IllegalArgumentException: Test class can only have one constructor
...
```
probably again not a test class, same as 3)
```
6) testPackagePrivateMethod_bootstrapClassLoader(net.sf.cglib.reflect.TestFastClass)
java.lang.NoSuchMethodException: java.lang.String.getChars([C, int)
	at java.base/java.lang.Class.getDeclaredMethod(Class.java:2475)
	at net.sf.cglib.reflect.TestFastClass.testPackagePrivateMethod_bootstrapClassLoader(TestFastClass.java:663)
```
same as 2)
```
7) warning(junit.framework.TestSuite$1)
junit.framework.AssertionFailedError: No tests found in net.sf.cglib.CodeGenTestCase
...
```
just a warning, but the same as 3)

dev-java/cglib/cglib-3.3.0.ebuild
vaukai: With USE=doc it would run into errors -> [build.log.gz](https://github.com/gentoo/gentoo/files/6210510/build.log.gz)
How to proceed?  Are there docs at all?  If so I could not find them :-(  
Also an "examples" USE flag seems not to trigger any examples' installation.

Regarding the slot dependencies I successfully raised asm to `:9`.  Is possible to drop the slot restriction or to set a `minimal slot` so it could use any slot >7?  I've never seen such case :-(

> this one might not be needed, might be a bug in `java-ebuilder`

Removal of the `"${MY_P}"`line seems not to break anything.

> @vaukai might also want to change the pr subject

PR subject of course will be synced with latest commit message.

Before I push again, could we please solve the USE flags matter.  Meanwhile I need to try and understand your comments on the tests stuff ....

fordfrog: wrt min slot, it's safer to use `:9` as future version of `asm` does not have to be compatible with `cglib`.

fordfrog: wrt the `doc` use flag, can you try to run following command in the `WORKDIR` after emerge fails?
```
javadoc -Xdoclint:none -d target/api -encoding UTF-8 -docencoding UTF-8 -charset UTF-8 -classpath //usr/share/asm-9/lib/asm.jar://usr/share/ant-core/lib/ant.jar://usr/share/ant-core/lib/ant-bootstrap.jar://usr/share/ant-core/lib/ant-launcher.jar -quiet @sources.lst
```
then you can try to add the generated `cglib.jar` file to the `-classpath` and see whether it helps. because the error says `package net.sf.cglib does not exist` so i suppose it might need to have that on the classpath, though not sure, i'm not that familiar with `javadoc`.

also, the api documentation in java is inlined in the source files so it's generated from the source files. jfyi :-)

vaukai: This is what happens:

```
/var/tmp/portage/dev-java/cglib-3.3.0/work $ javadoc -Xdoclint:none -d target/api -encoding UTF-8 -docencoding UTF-8 -charset UTF-8 -classpath //usr/share/asm-9/lib/asm.jar://usr/share/ant-core/lib/ant.jar://usr/share/ant-core/lib/ant-bootstrap.jar://usr/share/ant-core/lib/ant-launcher.jar -quiet @sources.lst
cglib-RELEASE_3_3_0/cglib/src/main/java/net/sf/cglib/reflect/MethodDelegate.java:20: error: package net.sf.cglib does not exist
import net.sf.cglib.*;
^
1 error
```
Nothing gets generated there.  How is the procedure to compile cglib outside of portage including the javadocs?

vaukai:
> ```
> 1) testSql(net.sf.cglib.proxy.TestEnhancer)
> ...
> Caused by: java.lang.ClassNotFoundException: net.sf.cglib.proxy.Factory
> ...
> ```
> 
> 
> this class is missing from the tests classpath - try to find out why

Tried and failed.  `TEST_EXCLUD`ing it may be more an ugly hack than a workaround, but ...
And I am going to force-push, still without the `doc` USE flag.

> ```
> 7) warning(junit.framework.TestSuite$1)
> junit.framework.AssertionFailedError: No tests found in net.sf.cglib.CodeGenTestCase
> ...
> ```
> 
> 
> just a warning, but the same as 3)

Sometimes it's worth reading the [docs](https://devmanual.gentoo.org/eclass-reference/java-pkg-simple.eclass/index.html) exactly mentioning `net.sf.cglib.CodeGenTestCase`.

Will you please take it as is?

vaukai:
@fordfrog And, thanks again for patiently guiding me to the `java-ebuilder` tool

fordfrog:
@vaukai thank you for the contribution! i did some enhancements in d8be8eb1c8b1b4c26dbeee32758fc26c49383373.

vaukai:
@fordfrog 
Please cc-arches on https://bugs.gentoo.org/778596
Thank you

fordfrog:
@vaukai fot `dev-java/cglib-3.3.0` i'd wait at least few days whether any issue pops up... `dev-java/asm` done.

dev-java/cglib/cglib-3.1-r1.ebuild
vaukai: >  think this should be `RESTRICT="test"` if you want to disable the tests completely. with your syntax, it's still possible to run (and fail) tests for this package. i searched the devmanual but i must admit it does not provide enough info to use this feature in the right way, imo.

@fordfrog 
Not in devmanual but found it mentioned [here](https://projects.gentoo.org/qa/policy-guide/other-metadata.html#pg0703). 
There seems to be another way to completely disable tests:  
From my experience, an ebuild not having the `test` will simply not run tests even with `FEATURES=test` set.
But imho all java ebuilds should have the test flag added to "doc source", except for packages coming without test cases.

PR #20087, opened 2021-03-23
media-libs/libsdl2: Fix --disable-static patch for slibtool.

orbea:
With slibtool both `libSDL2main.a` and `libSDL2_test.a` fail to install. This was fixed in slibtool here.

https://dev.midipix.org/cross/slibtool/c/9a5f9681f88126c47e11ef49e7fb01c604c90986?branch=main

However using it still does not work in gentoo with `-all-static`, but works fine if its changed back to `-static` as is done upstream in SDL2. GNU libtool does not seem to behave any differently here and its unclear why this was ever changed?

Its also possible to entirely drop the patch, I am not sure it does anything useful for gentoo anymore?

Also see this slibtool issue:

https://dev.midipix.org/cross/slibtool/issue/28

### DIFF
--- a/media-libs/libsdl2/files/libsdl2-2.0.14-static-libs.patch
+++ b/media-libs/libsdl2/files/libsdl2-2.0.14-static-libs.patch
@@ -24,11 +24,11 @@ Bug: https://bugzilla.libsdl.org/show_bug.cgi?id=1431
  
  $(objects)/$(SDLMAIN_TARGET): $(SDLMAIN_OBJECTS)
 -	$(RUN_CMD_LTLINK)$(LIBTOOL) --tag=CC --mode=link $(CC) -static -o $@ $(SDLMAIN_OBJECTS) -rpath $(libdir)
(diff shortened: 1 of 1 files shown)


chewi:
I suspect it's because of the following paragraph from libsdl-org/SDL#556 but I'll need to look more closely.

> This won't work on all platforms as libSDL2main contains the main() entry point, at least on Windows. Also, linking the SDL2main dll will fail with undefined references to SDL functions. I got around this by adding -all-static to the SDL2main and SDL2_test link commands so that libtool does not try to link dynamic libraries. I'm no libtool/autoconf expert, but it seems to work fine (always creates static SDLmain), even with --disable-static.

orbea:
GNU libtool docs says:
```
-all-static

If output-file is a program, then do not link it against any shared libraries at all. 
If output-file is a library, then only create a static library. In general, this flag 
cannot be used together with ‘disable-static’ (see LT_INIT). 
```
https://www.gnu.org/software/libtool/manual/html_node/Link-mode.html

I'm not sure about the past, but I am fairly certain that `-all-static` is not needed to create those static libraries with GNU libtool anymore at least in gentoo. Using `-static` like in upstream should be fine.

chewi:
I tried to reproduce the issue by cross-compiling for x86_64-w64-mingw32 but it didn't break. I didn't think the problem would have gone away but evidently it has. We can always revisit this if it causes issues later.

PR #20090, opened 2021-03-24
media-video/shotcut: bump to 21.03.21

xdch47:


juippis (commented):
Please split adding and removing ebuilds into their own commits. So this should be 2 commits instead of 1.

juippis (approved):
Thanks!

PR #20095, opened 2021-03-24
sys-apps/plocate: Do not cron if systemd timer is enabled

Flowdalic:
On systems using systemd with a cron compatibility layer,
e.g. sys-process/systemd-cron, udpatedb would potentially run twice
daily: One time invoked by cron, and a another time invoked by the
systemd timer.

To avoid this, check in the cron script if the plocate-updatedb systemd timer is
enabled, and abort if so.


AnAverageHuman:
Seems reasonable, but why isn't this an issue with mlocate?

Flowdalic:
> Seems reasonable, but why isn't this an issue with mlocate?

I think nobody claimed that `mlocate` doesn't also suffer from this. ☺

I general, it is less an issue and more a waste of resources that we should try to avoid.

thesamesam:
> On 24 Mar 2021, at 15:28, Florian Schmaus ***@***.***> wrote:
> 
> 
> Seems reasonable, but why isn't this an issue with mlocate?
> 
> I think nobody claimed that mlocate doesn't also suffer from this. ☺

Sure, it’s just a natural question ;)

> 
> I general, it is less an issue and more a waste of resources that we should try to avoid.

Agreed. Fancy looking at the others too? (Not a requirement, just an interesting thought).

I wonder how many other packages suffer from this. :|

pinkflames:
The mlocate 0.26-r2 ebuild, that's keyworded as stable, does not yet install any systemd units. This does, however, appear to affect mlocate 0.26-r3.

Alessandro-Barbieri:
I don't think this is necessary.
emerge `sytemd-cron[-runparts]` (why is it default?)
`systemctl mask cron-daily-plocate.timer`

Flowdalic:
> I don't think this is necessary.

The point is that some users may want to have runparts enabled. And yes, you could tell users to manually disable the timer. But first, why require manual intervention when you could simply act accordingly at runtime (i.e., what this PR does)? And furthermore, if you have a dual-setup where /etc/cron.d *and* systemd is effective, then you usually want to use systemd's facilities, as they are far more flexible than cron.

floppym:
I don't like this. It would be better to ensure that the cron job and the systemd timer are not enabled together.

floppym:
If this gets merged, I would ask that you not hard-code the path to systemctl. Just let cron find it in PATH.

sys-apps/plocate/files/plocate.cron
```diff
@@ -1,6 +1,12 @@
 #! /bin/sh
 set -e
 
+# Do not run the cron job if plocate's systemd timer is enabled.
+if which systemctl >/dev/null 2>&1 \
+	&& systemctl is-enabled --quiet plocate-updatedb.timer; then
```
floppym, line 6: There's really no need to call `which` here. The `if` statement could be simplified to this:
```
if systemctl is-enabled --quiet plocate-updatedb.timer >/dev/null 2>&1; then
```
If the `systemctl` command is not found in PATH, the shell will treat it as status 127, and the `exit` command will not be executed.

Flowdalic:
Closed in favor of https://github.com/gentoo/gentoo/pull/20771

PR #20098, opened 2021-03-24
app text/sigil: bump v1.5.1

arthurzam:
- remove old stable version (there is still `1.4.3` stable version)
- bump to version `1.5.1`

I have waited some days, without any urgent bug report upstream, meaning we can be safe to deploy at gentoo on testing.


arthurzam:
Updated to the `v1.5.1` which was posted yesterday.

juippis (approved):

PR #20100, opened 2021-03-24
sys-block/whdd: fix obsolete github SRC_URIs

Flowdalic:


Flowdalic:
Fixed with 17ecd9e96325 ("sys-block/whdd: fix obsolete github SRC_URIs")

PR #20109, opened 2021-03-25
sys-boot/os-prober: Fix doexe warning

Peter-Levine:
**sys-boot/os-prober: Fix doexe warning** 
Fix "doexe: warning, skipping directory os-probes/mounted/x86/efi".

Signed-off-by: Peter Levine <plevine457@gmail.com>

### DIFF
--- a/sys-boot/os-prober/os-prober-1.78-r1.ebuild
+++ b/sys-boot/os-prober/os-prober-1.78-r1.ebuild
@@ -1,0 +1,94 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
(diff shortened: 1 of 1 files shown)


benkohler:
Your commit to silence the doexe warnings seems to have stopped installing /lib/linux-boot-probes/mounted/{40grub,50lilo}

benkohler:
I see, your new loop logic doesn't touch linux-boot-probes at all.  I think this patch can be simplified a bit, I like the for loop but I don't see any reason to replace `if [[ -d ${dir}/${debarch} ]]; then` with `if [[ -d os-probes/mounted/${debarch} ]]; then`

Peter-Levine:
@benkohler
OK.  This should work.

PR #20111, opened 2021-03-25
net-voip/sipvicious: Version 0.3.3

jkroonza:


juippis (commented):
```
 *
 *   /usr/man
 *
 * QA Notice: This ebuild installs into the following deprecated directories:
 *
 *   usr/man
```
We use `/usr/share/man/`. As always, rest looks good :P

net-voip/sipvicious/sipvicious-0.3.3.ebuild
```diff
@@ -0,0 +1,31 @@
...
+
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+RDEPEND="$(python_gen_cond_dep '
+		dev-python/dnspython[${PYTHON_USEDEP}]
+		dev-python/reportlab[${PYTHON_USEDEP}]
+	')"
+
+src_prepare()
+{
```
thesamesam, line 28: Oi!

PR #20116, opened 2021-03-25
dev-db/mydumper: bump to 0.10.3

hydrapolic:
Tested on ~amd64.


dev-db/mydumper/mydumper-0.10.3.ebuild
```diff
@@ -0,0 +1,44 @@
...
+IUSE="doc"
+
+COMMON_DEPEND="dev-db/mysql-connector-c:=
+	dev-libs/glib:=
+	dev-libs/libpcre:=
+	dev-libs/openssl:0=
+	sys-libs/zlib:="
+DEPEND="${COMMON_DEPEND}
+	virtual/pkgconfig
+	doc? ( dev-python/sphinx )"
```
scardracs, line 23: You have actually set a python program without provide PYTHON_COMPAT=( python3_{7..9} ) and ${PYTHON_DEPS}

scardracs: Also, IIRC, you have to inherit python-any-r1 to, but probably @thesamesam or @juippis will know better than me

dev-db/mydumper/mydumper-0.10.3.ebuild
```diff
@@ -0,0 +1,44 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake-utils
```
juippis, line 6: Switch to *cmake.eclass* already good man :)

dev-db/mydumper/mydumper-0.10.3.ebuild
```diff
@@ -0,0 +1,44 @@
...
+
+COMMON_DEPEND="dev-db/mysql-connector-c:=
+	dev-libs/glib:=
+	dev-libs/libpcre:=
+	dev-libs/openssl:0=
+	sys-libs/zlib:="
+DEPEND="${COMMON_DEPEND}
+	virtual/pkgconfig
+	doc? ( dev-python/sphinx )"
+RDEPEND="${COMMON_DEPEND}"
```
juippis, line 24: It doesn't look like you actually need `COMMON_DEPEND`.

dev-db/mydumper/mydumper-0.10.3.ebuild
juippis, line 23: These are `BDEPEND`.

juippis: No it's fine, if you don't have a BDEPEND-executable sphinx, you've got bigger problems in your system.

dev-db/mydumper/mydumper-0.10.3.ebuild
```diff
@@ -0,0 +1,44 @@
...
+DESCRIPTION="A high-performance multi-threaded backup (and restore) toolset for MySQL"
+HOMEPAGE="https://github.com/maxbube/mydumper"
+SRC_URI="https://github.com/maxbube/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="doc"
+
+COMMON_DEPEND="dev-db/mysql-connector-c:=
+	dev-libs/glib:=
```
juippis, line 17: `:2=`

dev-db/mydumper/mydumper-0.10.3.ebuild
```diff
@@ -0,0 +1,44 @@
...
+	doc? ( dev-python/sphinx )"
+RDEPEND="${COMMON_DEPEND}"
+
+PATCHES=(
+	"${FILESDIR}/${PN}-atomic.patch" #654314
+)
+
+src_prepare() {
+	# respect user cflags; do not expand ${CMAKE_C_FLAGS} (!)
+	sed -i -e 's:-Werror -O3 -g:${CMAKE_C_FLAGS}:' CMakeLists.txt || die
```
juippis, line 32: This is error-prone, https://bugs.gentoo.org/685160

dev-db/mydumper/mydumper-0.10.3.ebuild
Whissi: Can we talk about this? This is Gentoo, a rolling distribution. We should only set upper bounds when know for sure it will not work with whatever comes next. But pinning to a slot when this cannot be known yet is against the idea of a rolling distribution. Not?

scardracs:
@hydrapolic you may want to read comments made my @juippis and fix them

dev-db/mydumper/mydumper-0.10.3.ebuild
thesamesam: To an extent, although you're going to have to add the machinery if you ever need a custom theme. Sphinx itself should be okay though, I suppose.

dev-db/mydumper/mydumper-0.10.3.ebuild
```diff
@@ -0,0 +1,44 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="doc"
+
+COMMON_DEPEND="dev-db/mysql-connector-c:=
+	dev-libs/glib:=
+	dev-libs/libpcre:=
+	dev-libs/openssl:0=
+	sys-libs/zlib:="
+DEPEND="${COMMON_DEPEND}
```
thesamesam, line 21: BDEPEND and so on

dev-db/mydumper/mydumper-0.10.3.ebuild
```diff
@@ -0,0 +1,44 @@
...
+
+EAPI=7
+
+inherit cmake-utils
+
+DESCRIPTION="A high-performance multi-threaded backup (and restore) toolset for MySQL"
+HOMEPAGE="https://github.com/maxbube/mydumper"
+SRC_URI="https://github.com/maxbube/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+LICENSE="GPL-3"
+SLOT="0"
```
thesamesam, line 12: new line after SRC_URI

hydrapolic:
Thanks @juippis @thesamesam @Whissi @scardracs for the input, pushed an update.

juippis (approved):

hydrapolic:
Thanks @juippis for the merge.

PR #20117, opened 2021-03-25
net-im/telegram-desktop: Take maintainership and version bump

mid-kid:
Also bumped the depended libraries


gyakovlev:
didn't upstream got rid of pulse/alsa distinction and handle sound via openal only?
I was not following development closely lately.

media-libs/tg_owt/metadata.xml
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
 <pkgmetadata>
-	<!-- maintainer-needed -->
+	<maintainer type="person">
+		<email>esteve.varela@gmail.com</email>
+		<name>Esteve Varela Colominas</name>
+	</maintainer>
```
gyakovlev, line 7: you cannot be only maintainer listed. you need proxy-maint added as well.
also you need `proxied="yes""` attribute for yourself.

look at `app-arch/lzma/metadata.xml` for an example how to do it.

mid-kid:
@gyakovlev 
Yes, but the latest tg_owt currently doesn't work (waiting on https://github.com/desktop-app/tg_owt/pull/57 to be merged), and the version used still requests them. Also libtgvoip still depends on them, and I'm not sure if it's safe to build that without them (I'll probably add USE flags for them once tg_owt is free of them).

media-libs/tg_owt/tg_owt-0_pre20210309.ebuild
```diff
@@ -0,0 +1,76 @@
+# Copyright 2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake
+
+TG_OWT_COMMIT="7f965710b93c4dadd7e6f1ac739e708694df7929"
+LIBVPX_COMMIT="5b63f0f821e94f8072eb483014cfc33b05978bb9"
+LIBYUV_COMMIT="ad890067f661dc747a975bc55ba3767fe30d4452"
```
gyakovlev, line 10: bundling is really bad and one of the reasons I've stepped down as maintainer.

I see you've started unbundling work in https://github.com/desktop-app/tg_owt/pull/55
We can make an exception this time, because tg needs a bump pretty urgently, but I really hope your patch gets merged and we could use system versions of libuv, libvpx and openh264.

gyakovlev (commented):
looks good on first glance, I'll do some testing as time permits (probably on weekend) and provide more feedback.

media-libs/tg_owt/tg_owt-0_pre20210309.ebuild
gyakovlev: I'm less concerned about other libs, like abseil, video/media stuff is a priority due to CPU flag situation and lack of handling it upstream.

mid-kid: I've been working on this for a week and I just want to upgrade telegram at this point. This is *known* to work and it gives me time to work out the kinks with the unbundling PR.
That said, libyuv has no stable versions and probably will never end up in the tree for that reason. Similarly, libvpx is currently using a git release.
I get why bundling is bad, but I can't do much about it in this case, so if it works, it works...

gyakovlev:
I went ahead tested and merged, fixing years in ebuilds. Thanks for doing all that work.

I also took liberty to exclude `desktop_capture` module from `tg_owt` build, which has same effect as https://github.com/desktop-app/tg_owt/pull/57, but allows to build on arches without SSE2  ( arm* , ppc64 )

since you've taken maintenance please take ownership of the bugs as well.
You seem to be in CC on most of open ones.

mid-kid:
@gyakovlev
Thanks a ton, I will be looking into resolving the relevant issues, though some of them (including the CVE) remain open because the stable version doesn't fix them, and I have no idea how the stabilization process is handled.
How do I assign myself to these issues?

gyakovlev:
for security bugs - those are assigned to security team, developers and maintainers remain in CC. that's security team rule.

as for stabilization - general rule of 30 days applies, but can be somewhat ignored if security vuln is serious.

generally I was only stabilizing release versions of telegram, almost never betas or alphas.

general pointers can be found here: https://wiki.gentoo.org/wiki/Stable_request

once you decide it's time for new version to go stable - file a stablereq bug, like described here https://wiki.gentoo.org/wiki/Bugzilla/Bug_report_guide#Request_stabilization
proxied maintainers generally can't just stabilize things via PR, unless special agreement with the proxy is present.

do not stabilize it on ppc64.
I'm still using it on ppc64 and arm64 so I'm still looking and will generally help if I have time or see any issues on non-amd64 arches.

mid-kid:
Thanks, I'lll keep those pointers in mind. Given how telegram releases a new "stable" version roughly every month that's going to be a regular affair, if I want to keep some stable version around, that is.

I only run amd64 so I can only suitably test on this arch. Having someone around to test for ppc64 is very helpful, so thanks.

gyakovlev:
FYI, sent a small follow up on my initial for grabs message =)
https://archives.gentoo.org/gentoo-dev/message/176ce2478d4ccc275dc3235094300962

PR #20118, opened 2021-03-25
sci-geosciences/GeographicLib: improvements

Alessandro-Barbieri:
Closes: https://bugs.gentoo.org/636338
Closes: https://bugs.gentoo.org/777810

### DIFF
--- a/sci-geosciences/GeographicLib/GeographicLib-1.51-r1.ebuild
+++ b/sci-geosciences/GeographicLib/GeographicLib-1.51-r1.ebuild
@@ -1,0 +1,111 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="7"
+
+PYTHON_COMPAT=( python3_{7,8,9} )
+
+inherit cmake distutils-r1
+
+MY_PN="${PN,,}"
+
+DESCRIPTION="C++ library for converting geographic coordinate systems"
+HOMEPAGE="https://sourceforge.net/projects/geographiclib/"
+SRC_URI="mirror://sourceforge/${MY_PN}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="doc examples python"
+# To-Do:find out why this doesn't work
+# 105/167 Test #105: GeodSolve82 ......................***Failed
+# Required regular expression not found. Regex=[20\.0010* 0\.0* 180\.0* 20\.0010* 0\.0* 180\.0* 0\.0* 0\.0* 0\.0* 1\.0* 1\.0* -?0
+RESTRICT="test"
+
+REQUIRED_USE="python? ( ${PYTHON_REQUIRED_USE} )"
+RDEPEND="python? ( ${PYTHON_DEPS} )"
+DEPEND="${RDEPEND}"
+BDEPEND="
+	doc? (
+		>=app-doc/doxygen-1.8.7
+		>=dev-lang/perl-5.26.1-r1
+		>=dev-python/sphinx-1.6.3-r2
+		>=sys-apps/util-linux-2.31
+	)
+"
+
+distutils_enable_tests setup.py
+# there are additional docs in the python dir
+distutils_enable_sphinx python/doc
+
+src_prepare() {
+	cmake_src_prepare
+
+	if use python; then
+		pushd "python" || die
+		distutils-r1_python_prepare_all
+		popd || die
+	fi
+}
+
+src_configure() {
+	export GEODATAPATH="/usr/share/${MY_PN}"
+
+	local mycmakeargs=(
+		-DGEOGRAPHICLIB_DOCUMENTATION=$(usex doc ON OFF)
+		-DGEOGRAPHICLIB_LIB_TYPE="SHARED"
+		-DGEOGRAPHICLIB_DATA="${GEODATAPATH}"
+	)
+	cmake_src_configure
+}
+
+src_compile() {
+	cmake_src_compile
+
+	if use python; then
+		pushd "python" || die
+		python_foreach_impl distutils-r1_python_compile
+		popd || die
+		use doc && build_sphinx python/doc
+	fi
+}
+
+src_test() {
+	# Only 1 failing test in the C code, python passes for me
+	cmake_src_test
+
+	if use python; then
+		pushd "python" || die
+		python_foreach_impl python_test
+		popd || die
+	fi
+}
+
+src_install() {
+	cmake_src_install
+
+	# remove python things added by the cmake_src_install function
+	rm -rf "${D}/usr/$(get_libdir)/python" || die
+	# if use python we re-add these python files for each implementation
+	if use python; then
+		pushd "python" || die
+		python_foreach_impl distutils-r1_python_install
+		popd || die
+	fi
+
+	#TODO: find out if java stuff need something
+
+	# Installs to wrong doc dir for some reason
+	# Also happens with USE="-doc"
+	mv "${D}/usr/share/doc/${PN}" "${D}/usr/share/doc/${PF}" || die
+
+	if use examples; then
+		dodoc -r examples
+		docompress -x "/usr/share/doc/${PF}/examples"
+	fi
+
+	insinto "/usr/share/maxima/${MY_PN}"
+	doins -r maxima/.
+
+	find "${D}" -name "*.la" -delete || die
+}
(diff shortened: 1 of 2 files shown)


Alessandro-Barbieri:
@aballier

scardracs (approved):

scardracs (approved):

aballier:
err 'replace ebuild by the one in guru' doesn't provide much info on what the commit actually does
I would suggest splitting the changes in several relatively atomic commits so that the improvements are clear

sci-geosciences/GeographicLib/GeographicLib-1.51-r1.ebuild
```diff
@@ -0,0 +1,197 @@
...
+IUSE_MAGNETIC_MODELS="
+	magnetic-models-wmm2010
+	magnetic-models-wmm2015v2
+	magnetic-models-wmm2020
+	magnetic-models-igrf11
+	magnetic-models-igrf12
+	magnetic-models-emm2010
+	magnetic-models-emm2015
+	magnetic-models-emm2017
+"
```
aballier, line 46: do we really need a useflag for every single one of those ?
it seems simpler to have a single useflag for all models

plus they don't seem to change anything on the installed programs, so maybe even worth a separate ebuild with a P/RDEPEND from geographiclib since they seem completely independent

sci-geosciences/GeographicLib/GeographicLib-1.51-r1.ebuild
```diff
@@ -0,0 +1,197 @@
...
+gravity-models-wgs84?            ( ${COMMON_URI}/gravity-distrib/wgs84.tar.bz2		-> gravity-wgs84.tar.bz2	)
+
+magnetic-models-wmm2010?         ( ${COMMON_URI}/magnetic-distrib/wmm2010.tar.bz2	-> magnetic-wmm2010.tar.bz2	)
+magnetic-models-wmm2015v2?       ( ${COMMON_URI}/magnetic-distrib/wmm2015v2.tar.bz2	-> magnetic-wmm2015v2.tar.bz2	)
+magnetic-models-wmm2020?         ( ${COMMON_URI}/magnetic-distrib/wmm2020.tar.bz2	-> magnetic-wmm2020.tar.bz2	)
+magnetic-models-igrf11?          ( ${COMMON_URI}/magnetic-distrib/igrf11.tar.bz2	-> magnetic-igrf11.tar.bz2	)
+magnetic-models-igrf12?          ( ${COMMON_URI}/magnetic-distrib/igrf12.tar.bz2	-> magnetic-igrf12.tar.bz2	)
+magnetic-models-emm2010?         ( ${COMMON_URI}/magnetic-distrib/emm2010.tar.bz2	-> magnetic-emm2010.tar.bz2	)
+magnetic-models-emm2015?         ( ${COMMON_URI}/magnetic-distrib/emm2015.tar.bz2	-> magnetic-emm2015.tar.bz2	)
+magnetic-models-emm2017?         ( ${COMMON_URI}/magnetic-distrib/emm2017.tar.bz2	-> magnetic-emm2017.tar.bz2	)
```
aballier, line 89: are we 100% guaranteed those tarballs will never be updated in place ?

sci-geosciences/GeographicLib/GeographicLib-1.51-r1.ebuild
```diff
@@ -0,0 +1,197 @@
...
+"
+LICENSE="
+	MIT
+	geoids? ( public-domain )
+	gravity? ( public-domain )
+	magnetic? ( public-domain )
+"
+
+RDEPEND="
+	>=dev-libs/boost-1.65.0
```
aballier, line 99: := ?

aballier: missing boost conditional here ?

sci-geosciences/GeographicLib/GeographicLib-1.51-r1.ebuild
```diff
@@ -0,0 +1,197 @@
...
+
+src_prepare() {
+	#TODO: strip cflags
+#	sed -i "s|CXXFLAGS = -g -Wall -Wextra -O3 -std=c++0x||" tools/Makefile.mk || die
+#	sed -i "s|CXXFLAGS = -g -Wall -Wextra -O3 -std=c++0x||" src/Makefile.mk || die
+
+	cmake_src_prepare
+
+	if use python; then
+		cd "python" || die
```
aballier, line 123: pushd/popd

Alessandro-Barbieri:
Should I build the examples or I only install the source for them?

aballier:
> Should I build the examples or I only install the source for them?

if they don't do anything useful, IMHO it's better just to install the sources

Alessandro-Barbieri:
I've moved the data to a new package in guru

Alessandro-Barbieri:
@aballier can you re review?

sci-geosciences/GeographicLib/GeographicLib-1.51-r1.ebuild
```diff
@@ -0,0 +1,134 @@
...
+KEYWORDS="~amd64"
+IUSE_PRECISION="
+	precision-arbitrary
+	+precision-double
+	precision-long-double
+	precision-quad
+	precision-single
+"
+IUSE="${IUSE_PRECISION} doc examples python"
+IUSE_EXPAND="PRECISION"
```
aballier, line 32: what is this supposed to achieve ?

sci-geosciences/GeographicLib/GeographicLib-1.51-r1.ebuild
```diff
@@ -0,0 +1,134 @@
...
+RESTRICT="test"
+
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE_PRECISION="
+	precision-arbitrary
+	+precision-double
+	precision-long-double
+	precision-quad
+	precision-single
```
aballier, line 29: I really don't like these useflags: they appear to change the ABI, for example some Math.hpp classes will have different signatures

please run an abi checker tool before and after toggling one of those flags

if they differ, I'm afraid we'll have to pick one precision and stick with it (ideally one that matches other distros so that we can have some sort of binary compatibility)

Alessandro-Barbieri: I've remove the precision useflags

Alessandro-Barbieri:
@aballier ready to be merged

aballier (approved):

sci-geosciences/GeographicLib/GeographicLib-1.51-r1.ebuild
```diff
@@ -0,0 +1,110 @@
...
+
+REQUIRED_USE="python? ( ${PYTHON_REQUIRED_USE} )"
+RDEPEND=""
+DEPEND="
+	${RDEPEND}
+	doc? (
+		>=app-doc/doxygen-1.8.7
+		>=dev-lang/perl-5.26.1-r1
+		>=dev-python/sphinx-1.6.3-r2
+		>=sys-apps/util-linux-2.31
```
AndrewAmmerlaan, line 34: These are probably BDEPEND I think, at least sphinx and doxygen are things that are typically _executed_ during build time.

sci-geosciences/GeographicLib/GeographicLib-1.51-r1.ebuild
```diff
@@ -0,0 +1,110 @@
...
+		popd || die
+	fi
+}
+
+src_install() {
+	cmake_src_install
+
+	# remove python things added by the cmake_src_install function
+	# these are installed in the wrong python dir
+	rm -rf "${D}/usr/$(get_libdir)/python" || die
```
AndrewAmmerlaan, line 90: You might be able to fix this path (and the doc one) in the configure phase with an appropriate `-D....=...` option.

Alessandro-Barbieri: which one? `DOCDIR` doesn't get picked up by cmake

Alessandro-Barbieri: even if I fix the python install path, python modules needs to be installed multiple times in multiple paths

AndrewAmmerlaan: > even if I fix the python install path, python modules needs to be installed multiple times in multiple paths

Good point, never mind then

PR #20121, opened 2021-03-26
app-crypt/nitrocli: Bump version to 0.4.0

d-e-s-o:
This patch adds an ebuild for app-crypt/nitrocli version 0.4.0. The main
changes in this release are the addition of support for Purism's Librem
Key as well as for invocation of user-provided extensions.

Signed-off-by: Daniel Müller <deso@posteo.net>


app-crypt/nitrocli/nitrocli-0.4.0.ebuild
```diff
@@ -0,0 +1,124 @@
...
+		target/release/shell-complete bash > ${PN}.bash || die
+		newbashcomp ${PN}.bash ${PN}
+	fi
+
+	if use fish-completion; then
+		target/release/shell-complete fish > ${PN}.fish || die
+		insinto /usr/share/fish/vendor_conf.d/
+		insopts -m0755
+		newins ${PN}.fish ${PN}.fish
+	fi
```
juippis, line 120: We don't use USE flags to control installing these files, 
and especially since it looks like fish is not required to generate the completion file.

juippis (approved):

PR #20123, opened 2021-03-26
net-im/telegram-desktop, media-libs/tg_owt: Resolve missing dependencies

mid-kid:
"oops"

### DIFF
--- a/media-libs/tg_owt/tg_owt-0_pre20210309.ebuild
+++ b/media-libs/tg_owt/tg_owt-0_pre20210309.ebuild
@@ -40,6 +40,7 @@ DEPEND="
 	media-libs/libjpeg-turbo:=
 	media-libs/opus
 	media-video/ffmpeg:=
+	x11-libs/libXtst
 	!pulseaudio? ( media-sound/apulse[sdk] )
 	pulseaudio? ( media-sound/pulseaudio )
 "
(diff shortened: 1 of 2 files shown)


wellWINeo:
i manually installed glibmm and still have error that the cmake can't find it. Is it building successful on your system?

mid-kid:
@wellWINeo Install dev-cpp/glibmm:2, specificially SLOT 2.

gyakovlev:
I'll merge it with some fixes asap, but you should start using repoman or pkgcheck (recommended) to check your work before committing. just run `pkgcheck scan --commits` on the branch before pushing, it'll print some messages, especially that `RDEPEND` was modified without revbump.

we can't change `RDEPEND` without a revision bump. that's because users who already have telegram installed with improper `RDEPEND` will never get an updated ebuild until it's bumped, because version does not change.
so portage will technically allow removing libXtst and glibmm:2 from the system if nothing else requires it, breaking telegram.

yeah, it sucks that with -r1 users will have to rebuild it, but it's the preferred way.

mid-kid:
I see. That makes sense, though it isn't the most obvious thing. I'll be more careful about `RDEPEND` in the future. I don't mind using something to check my changes, so I'll just use whatever you recommend, but I sort of expected the qa-bot to do the same...

PR #20138, opened 2021-03-26
sys-devel/elftoolchain: Build and install all avalable tools

jsmolic:


sys-devel/elftoolchain/elftoolchain-0.7.1-r1.ebuild
```diff
@@ -40,16 +47,17 @@ _bmake() {
...
-	doheader common/elfdefinitions.h
-	doheader libelf/{gelf,libelf}.h
-
-	dolib.so libelf/libelf.so.1
-	dosym libelf.so.1 /usr/$(get_libdir)/libelf.so
-
-	dodoc README
+	_bmake \
+		DESTDIR="${D}" \
+		BINDIR="${EPREFIX}"/usr/bin/${CHOST}-elftoolchain \
```
thesamesam, line 56: Clang/LLVM do /usr/lib/llvm/11/bin/clang, but obviously they don't bother with tuples in the paths. I'm honestly not that bothered about what we do here as long as it doesn't collide with anything else or confuse people.

-elftoolchain is pretty unique although I feel like we're abusing that part of the tuple. @SoapGentoo?

Alessandro-Barbieri:
It doesn't build for me

jsmolic:
@Alessandro-Barbieri hmm, it works fine for me, not sure why you are hitting that

jsmolic:
@Alessandro-Barbieri Seems to work fine on ago's CI as well... What version of bmake do you have?

ionenwks:
@Alessandro-Barbieri I assume you're setting MAKEFLAGS given the usual, what is it set to now?

For example `MAKEFLAGS='-bad'` would cause this.

Alessandro-Barbieri:
Oh, I didn't file a bug because my chroot it's too experimental. I'm using `MAKEFLAGS="LIBTOOL=rlibtool"`

ionenwks:
I see, probably something else then given just that should be fine. bmake does read many env vars, so likely something you have set somewhere causing this, can't really reproduce here and it's fine.

Alessandro-Barbieri:
Oh, I've find out: I was using `MAKEOPTS="LIBTOOL=rlibtool ${MAKEOPTS}"` and that doesn't make sense. Removing `${MAKEOPTS}` will compile

PR #20141, opened 2021-03-27
dev-java/easymock: EAPI 7

vaukai:
Patch out missing dependency on "com.google.dexmaker.stock"


vaukai:
@fordfrog Do you see a chance to get the missing "com.google.dexmaker.stock" packaged?  If not we need to keep it patched out.

fordfrog:
@vaukai examples are not installed auto-magically, you have to copy that `src_install` from the current ebuild to install examples.

vaukai:
> @vaukai examples are not installed auto-magically, you have to copy that `src_install` from the current ebuild to install examples.

Not that simple.  Upstream moved to hosting their code on Github and the tarball there does not have the samples.jar or examples directoty.  Should I switch back to the Sourceforge-hosted tarball?

fordfrog:
@vaukai you can keep it this way, but then you should drop `examples` use flag as it is not used anymore.

vaukai:
> @vaukai you can keep it this way, but then you should drop `examples` use flag as it is not used anymore.

Done.

fordfrog:
@vaukai cool, thanks! :-)

PR #20143, opened 2021-03-27
www-apps/radicale: bump to 3.0.6 and added python3.9

scardracs:
Bumped to new version and added dev-python/defusedxml as RDEPEND.
Tested against python3.8/3.9 and gcc10. Not tested with musl.


scardracs:
@thesamesam @juippis looks like there is a problem with Manifest

PR #20144, opened 2021-03-27
media-sound/wildmidi: update to 0.4.4

winterheart:


media-sound/wildmidi/wildmidi-0.4.4.ebuild
```diff
@@ -0,0 +1,59 @@
...
+	player? (
+		alsa? ( media-libs/alsa-lib )
+		openal? ( media-libs/openal )
+	)
+"
+RDEPEND="${DEPEND}
+	media-sound/timidity-freepats
+"
+
+REQUIRED_USE="player? ( ^^ ( alsa oss openal ) )"
```
juippis, line 27: I'd like to transform this `REQUIRED_USE` into a priority-like contsruct within `src_configure`. Something like, if `+alsa` then choose alsa, if `-alsa +oss` choose oss and if `-alsa -oss +openal` use that.

juippis (approved):

PR #20147, opened 2021-03-27
media-video/qmplay2: bump v21.03.09

arthurzam:
- add `pipewire` use flag (add description to `metadata.xml`
- update the live version
- use `xdg.eclass` and remove custom calls to `xdg-utils.eclass`.

(It is easier to see my changes on the live version to compare).


media-video/qmplay2/qmplay2-21.03.09.ebuild
```diff
@@ -0,0 +1,139 @@
...
+
+LICENSE="LGPL-3"
+SLOT="0"
+
+IUSE="avdevice +audiofilters +alsa cdio cuvid dbus extensions
+	gme inputs libass modplug notifications opengl pipewire portaudio
+	pulseaudio sid shaders vaapi vdpau +videofilters visualizations vulkan xv"
+
+REQUIRED_USE="
+	audiofilters? ( || ( alsa portaudio pulseaudio ) )
```
juippis, line 28: Does pipewire need to be added here? (And same question for -9999)

arthurzam: You are right. `pipewire` is a possible audio output device, therefore should also be in this array.
Fixed.

juippis (approved):
Doesn't work with `+vulkan` for me, rest looks good.

PR #20151, opened 2021-03-27
x11-drivers/xf86-video-amdgpu: Adds postinst elog to fix #778494

pinkflames:
Due to EAPI=7 changes described in #661502 , the amdgpu DDX does not
work out of the box. This commit introduces an elog message
detailing the workaround that AMDGPU users on Gentoo need to apply
manually to get a working Xorg server.

Closes: https://bugs.gentoo.org/778494

Also updated the ebuild header with the current year, of course.

### DIFF
--- a/x11-drivers/xf86-video-amdgpu/xf86-video-amdgpu-19.1.0.ebuild
+++ b/x11-drivers/xf86-video-amdgpu/xf86-video-amdgpu-19.1.0.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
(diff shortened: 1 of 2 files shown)


mattst88:
> Due to EAPI=7 changes described in #661502 , the amdgpu DDX does not
> work out of the box.

The changes noted in that bug were reverted. Specifically, the problem was that xf86-video-amdgpu didn't inherit xorg-3, which does this:

````
      [[ ${PN} == xorg-server || ${PN} == xf86-video-* || ${PN} == xf86-input-* ]] \
          && append-ldflags -Wl,-z,lazy
````

But today it inherits xorg-3. So this cannot be the problem I don't believe.

pinkflames:
Thanks for the explaining that, that old bug report was a bit over my head. Should I keep this PR open to potentially update it?

mattst88:
Yeah, leaving it open until we figure out what's going on with the reporter's system seems fine to me.

Thanks!

mattst88:
Looks like `-fno-plt` is the problem.

ionenwks:
> Looks like `-fno-plt` is the problem.

Yeah, -fno-plt makes lazy bindings non-functional and we know how that goes with xorg.

PR #20154, opened 2021-03-27
gtksharp2-removal

a17r:


thesamesam (approved):
General approval as one of the only devs in Gentoo with an interest in .NET/Mono.

jsmolic (approved):
Thanks, fine for `app-misc/gnutu`

PR #20159, opened 2021-03-28
dev-db/sqldeveloper-20.4.1.407.0006 version bump

fedeliallalinea:


dev-db/sqldeveloper/sqldeveloper-20.4.1.407.0006.ebuild
```diff
@@ -0,0 +1,92 @@
...
+KEYWORDS="-* ~amd64"
+
+IUSE="mssql mysql postgres sybase"
+
+RDEPEND="mssql? ( dev-java/jtds:1.3 )
+	mysql? ( dev-java/jdbc-mysql:0 )
+	postgres? ( dev-java/jdbc-postgresql:0 )
+	sybase? ( dev-java/jtds:1.3 )
+	>=dev-java/openjdk-8:*[javafx]
+	>=virtual/jre-1.8:*"
```
juippis: We usually sort unconditional deps before conditional ones, 
https://devmanual.gentoo.org/general-concepts/dependencies/
(just FYI)

juippis (approved):

dev-db/sqldeveloper/sqldeveloper-20.4.1.407.0006.ebuild
fedeliallalinea: Next release I change order, thank you.

PR #20166, opened 2021-03-28
xorg-3.eclass: strip -fno-plt from *FLAGS

pinkflames:
As discussed in #778494, the GCC flag -fno-plt will break lazy
binding, which appears to still be necessary for Xorg. Stripping the
offending flag out is the next best solution for reliable user
experience on Gentoo.

Closes: https://bugs.gentoo.org/778494

I first tried to remove lazy binding but it was still producing broken Xorg drivers, so I then tried stripping out -fno-plt and that worked. I also checked that no other eclass (apart from the now unused xorg-2) is using z,lazy and may have needed the same fix.

Oh, and credit/thanks go to @AdelKS for discovering and reporting this to the Gentoo bug tracker and @mattst88 for correctly identifying the cause and describing how to fix it.

### DIFF
--- a/eclass/xorg-3.eclass
+++ b/eclass/xorg-3.eclass
@@ -334,9 +334,12 @@ xorg-3_flags_setup() {
 
 	# Win32 require special define
 	[[ ${CHOST} == *-winnt* ]] && append-cppflags -DWIN32 -D__STDC__
-	# hardened ldflags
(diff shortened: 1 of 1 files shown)


pinkflames:
I have reformated the patch to match the style of the one posted to the gentoo-dev ML, updated the associated comment and rebased on the new xorg-3.eclass in Portage.

I also want to point out that this PR contains a second commit that changes WANT_AUTO* variable placement that I have not received any kind of feedback on.

PR #20174, opened 2021-03-29
media-gfx/xpaint: dependency `sys-devel/automake:1.15`

ceamac:
`media-gfx/xpaint-2.10.2-r1` depends on an older version of automake

Closes: https://bugs.gentoo.org/778968

### DIFF
--- a/media-gfx/xpaint/xpaint-2.10.2-r1.ebuild
+++ b/media-gfx/xpaint/xpaint-2.10.2-r1.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-inherit desktop toolchain-funcs
+inherit autotools desktop toolchain-funcs
 
 DESCRIPTION="Image editor with tiff, jpeg and png support"
 HOMEPAGE="http://sf-xpaint.sourceforge.net/"
@@ -47,6 +47,11 @@ PATCHES=(
 	"${FILESDIR}"/${P}-libtool-clang.patch
 	"${FILESDIR}"/${P}-respect-ldflags.patch
 )
+
+src_prepare() {
+	default
+	eautoreconf
+}
 
 src_configure() {
 	econf \


media-gfx/xpaint/xpaint-2.10.2-r1.ebuild
```diff
@@ -36,7 +36,9 @@ RDEPEND="
 	)
 "
 DEPEND="${RDEPEND}"
+# sys-devel/automake:1.15 for bug #778968; not needed in xpaint 3+
 BDEPEND="
+	sys-devel/automake:1.15
```
Zlogene, line 41: it is a wrong way to depend on automake. Please consult https://devmanual.gentoo.org/eclass-reference/autotools.eclass/index.html

Zlogene (changes requested):

media-gfx/xpaint/xpaint-2.10.2-r1.ebuild
ceamac: I know that.  This is a fix for the old, stable xpaint.  In xpaint 3 I used eautoreconf, but I did not want to make big changes to a stable ebuild.  That's why I also added a comment.

If this is not acceptable I can add a proper fix, hoping I won't break anything.

ceamac: Please disregard the previous comment, I think I understand the problem now.  I'll come back with a new push later.

ceamac: I misunderstood last time, but now I hope it's right.  Please review.

thesamesam:
I'd prefer eautoreconf here -- maintainer mode is too unpredictable, if that's okay?

ceamac:
Ok, I created a diff between the current version with automake 1.15 and the version with eautoreconf (automake 1.16), I looked through it and it seems fine, I've seen nothing out of the ordinary.  The makefiles especially look rather similar, so I'll go on and use `eautoreconf`.

PR #20175, opened 2021-03-29
app-text/master-pdf-editor: bump to 5.7.40

jpataias:


juippis (approved):
Does this get updated daily? When I launch the program, it gives me a message saying "there's a new version available" and I see they have 5.7.53 release.

PR #20181, opened 2021-03-29
media-gfx/inkscape: Remove libX11

lanodan:
It is only used at link time when present, see `X11_FOUND` in
./CMakeScripts/DefineDependsandFlags.cmake

Closes: https://bugs.gentoo.org/768663

### DIFF
--- a/media-gfx/inkscape/files/inkscape-1.0.2-automagic-libX11.patch
+++ b/media-gfx/inkscape/files/inkscape-1.0.2-automagic-libX11.patch
@@ -1,0 +1,181 @@
+From 6d0ace0518f0da18c7e81be1edecd50d997230b1 Mon Sep 17 00:00:00 2001
+From: "Haelwenn (lanodan) Monnier" <contact@hacktivis.me>
+Date: Tue, 11 May 2021 10:43:27 +0200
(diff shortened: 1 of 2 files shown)


mattst88:
So it links with libX11 but doesn't actually use anything from it, so -Wl,--as-needed removes the linkage?

lanodan:
> So it links with libX11 but doesn't actually use anything from it, so -Wl,--as-needed removes the linkage?

Nah, it's an annoying issue.
Inkscape uses few X11 symbols when `<gdk/gdkconfig.h>` defines `GDK_WINDOWING_X11` (basically when `gtk+:3` had `USE=X`) and they don't have any configuration option.

Maybe something like this would be better?
```
COMMON_DEPEND="
  X? ( gtk+:3[X] libX11 )
  !X? ( gtk+:3[-X] )
  …
"
```

mattst88:
Sounds like it's an [automagic dependency](https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies)?

If so, we either have to unconditionally depend on libX11 or add a toggle to the build system to disable linking with libX11.

lanodan:
Sent a patch to inkscape, will apply it here: https://gitlab.com/inkscape/inkscape/-/merge_requests/3208

PR #20182, opened 2021-03-29
media-libs/waffle: Put libX11 behind existing USE=X

lanodan:


mattst88:
It should be more than that, right? With `USE=-X` I see

> Dependency x11-xcb skipped: feature glx disabled

so I think libxcb should be conditional as well. Maybe also xcb-proto? You'll have to check.

Also there's a 9999 ebuild. Please update it too.

mattst88:
Ping?

lanodan:
Done, as noted in the commit message I also removed `virtual/opengl` as mesa is already in the dependencies.

mattst88:
Thanks!

PR #20205, opened 2021-03-30
dev-util/bear: add 3.0.10

Jannik2099:
bear has been rewritten in C++ with version 3.0, hence we can drop python-single-r1.

bear no longer has a seperate "make check" target, instead we have to reconfigure & rebuild in src_test.
As with version 2.x tests require disabling various sandbox stuff, but they all pass.


dev-util/bear/bear-3.0.10.ebuild
```diff
@@ -0,0 +1,75 @@
...
+"
+
+DEPEND="${RDEPEND}
+test? (
+	>=dev-cpp/gtest-1.10
+)"
+
+BDEPEND="test? (
+	app-shells/bash
+	dev-python/lit
```
scardracs, line 32: If this test is still valid you can't drop PYTHON_COMPAT nor python-any-r1

scardracs: Also you don't really need to add bash sad requirement as it is a must have in Gentoo

dev-util/bear/bear-3.0.10.ebuild
```diff
@@ -0,0 +1,75 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~ppc64"
+IUSE="test"
+
+RDEPEND="
+>=net-libs/grpc-1.26
+>=dev-libs/libfmt-6.2
+>=dev-libs/spdlog-1.5
+>=dev-cpp/nlohmann_json-3.7
+>=dev-db/sqlite-3.14
```
scardracs, line 22: Please order alphabetically and use correct indentations

dev-util/bear/bear-3.0.10.ebuild
Jannik2099: OH, yeah that makes sense - ty

dev-util/bear/bear-3.0.10.ebuild
```diff
@@ -0,0 +1,81 @@
...
+DESCRIPTION="Build EAR generates a compilation database for clang tooling"
+HOMEPAGE="https://github.com/rizsotto/Bear"
+SRC_URI="https://github.com/rizsotto/Bear/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc64"
+IUSE="test"
+
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
```
thesamesam, line 19: Very close, we need to do pkg_setup later on too. see [here](https://dev.gentoo.org/~mgorny/python-guide/single.html#conditional-python-use)

thesamesam:
Re bash/sed: it's valid to add them as runtime dependencies because they're not guaranteed to be there, but they're guaranteed to exist for execution in the ebuild. It gets kind of complicated with cross and stuff but we don't run tests in those situations anyway. So, for tests, I guess it's right as-is now.

dev-util/bear/bear-3.0.10.ebuild
```diff
@@ -0,0 +1,81 @@
...
+	')
+)"
+
+RESTRICT="!test? ( test )"
+
+S="${WORKDIR}/${P^}"
+
+src_configure() {
+	local mycmakeargs=(
+		"-DENABLE_UNIT_TESTS=OFF"
```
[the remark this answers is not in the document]
Jannik2099: no, because then we'd build and execute tests during src_compile. bears cmake config has no way (that I found) of running tests seperately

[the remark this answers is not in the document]
scardracs: I think he meant $(usex test on off). This will check if actually test is required or not and add/remove it as arg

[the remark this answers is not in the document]
Jannik2099: Bear has tests on by default, which is why I disable them in configure and reenable them in test

dev-util/bear/bear-3.0.10.ebuild
```diff
@@ -0,0 +1,85 @@
...
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc64"
+IUSE="test"
+
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+RDEPEND="
+	${PYTHON_DEPS}
```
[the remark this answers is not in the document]
thesamesam: This is my fault: https://dev.gentoo.org/~mgorny/python-guide/any.html#conditional-python-use, python-any-r1, not single, I'm sorry!

dev-util/bear/bear-3.0.10.ebuild
```diff
@@ -0,0 +1,82 @@
...
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc64"
+IUSE="test"
+
+RDEPEND="
+	>=dev-cpp/nlohmann_json-3.7
+	>=dev-db/sqlite-3.14
+	>=dev-libs/libfmt-6.2
```
gyakovlev, line 22: needs a subslot dep: `>=dev-libs/libfmt-6.2:=`

dev-util/bear/bear-3.0.10.ebuild
```diff
@@ -0,0 +1,82 @@
...
+	>=dev-libs/spdlog-1.5
```
gyakovlev, line 23: ditto, add sublot

dev-util/bear/bear-3.0.10.ebuild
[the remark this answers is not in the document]
gyakovlev: I think if you change `TEST_BEFORE_INSTALL` to `0` it will be possible to skip tests, can you try and test it?

https://github.com/rizsotto/Bear/blob/d84687e2112e48263ebe18ed022a36c975c3ccf6/CMakeLists.txt#L63-L64

also it should be possible to run tests separately with `ctest --verbose || die`

gyakovlev (changes requested):

gyakovlev:
thanks for working on it.
I tried packaging earlier re-written versions but something was failing, looks like they've fixed failures.

Jannik2099:
Thanks for digging through the CMake files, that looks like it'll be a much prettier solution.

I'm away for the weekend and will fix on monday

dev-util/bear/bear-3.0.10.ebuild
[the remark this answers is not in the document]
Jannik2099: doesn't work, both test projects are still run during src_compile

gyakovlev:
thanks again, merged.

I've managed to fix tests to something normal in https://github.com/gentoo/gentoo/commit/d4e19111c464077a4675cb90ae46928778675431

also bumped to 3.0.11 while at it.

PR #20200, opened 2021-03-30
Miscellaneous eclass fixes

thesamesam:
Very much a work in progress.

PR #20207, opened 2021-03-30
flag-o-matic.eclass: Let's get rid of eutils, fix eclassdoc, make some funcs internal

a17r:


eclass/flag-o-matic.eclass
```diff
@@ -438,7 +470,23 @@ strip-flags() {
...
 }
 
+# @FUNCTION: test-flag-PROG
+# @USAGE: <compiler> <flag>
+# @INTERNAL
+# @DESCRIPTION:
+# Returns shell true if <flag> is supported by given <compiler>, else returns shell false.
 test-flag-PROG() {
+	[[ ${EAPI} == [0-7] ]] ||
+		die "Internal function ${FUNCNAME} is not available in >=EAPI-8."
```
thesamesam, line 480: Is anybody actually using these internals?

thesamesam (commented):
Oh [you!](https://github.com/gentoo/gentoo/pull/20200) You’ve done it better though, I hadn’t gone back to it yet.

eclass/flag-o-matic.eclass
a17r: Not in Gentoo ebuild repo, but better be safe than sorry.

thesamesam: I was about to write about how I was unsure about providing this safety net, and then I realised it was never prefixed with a '_' to begin with. As you were!

eclass/flag-o-matic.eclass
```diff
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 # @ECLASS: flag-o-matic.eclass
```
SoapGentoo, line 4: add supported EAPIs?

eclass/flag-o-matic.eclass
```diff
@@ -12,16 +12,37 @@
 if [[ -z ${_FLAG_O_MATIC_ECLASS} ]]; then
 _FLAG_O_MATIC_ECLASS=1
 
-inherit eutils toolchain-funcs multilib
+inherit toolchain-funcs
 
+case ${EAPI} in
```
SoapGentoo, line 17: missing the default specifier?

eclass/flag-o-matic.eclass
```diff
@@ -12,16 +12,37 @@
...
 	echo {ADA,C,CPP,CXX,CCAS,F,FC,LD}FLAGS
 }
 
+# @FUNCTION: setup-allowed-flags
+# @INTERNAL
+# @DESCRIPTION:
 # {C,CPP,CXX,CCAS,F,FC,LD}FLAGS that we allow in strip-flags
 # Note: shell globs and character lists are allowed
 setup-allowed-flags() {
+	[[ ${EAPI} == [0-7] ]] ||
```
SoapGentoo, line 35: if you really genuinely still want to support EAPI 0, don't you need a default specifier for all of these?

SoapGentoo: I would, just be prepared for bitching (if you're willing to take the heat)

SoapGentoo:
@a17r even once we're happy, please coordinate with @thesamesam to coalesce the cache hit

eclass/flag-o-matic.eclass
```diff
@@ -12,16 +13,38 @@
 if [[ -z ${_FLAG_O_MATIC_ECLASS} ]]; then
 _FLAG_O_MATIC_ECLASS=1
 
-inherit eutils toolchain-funcs multilib
+inherit toolchain-funcs
 
+case ${EAPI} in
```
ulm, line 18:
```suggestion
case ${EAPI:-0} in
```

eclass/flag-o-matic.eclass
```diff
@@ -438,7 +472,25 @@ strip-flags() {
...
+# @FUNCTION: test-flag-PROG
+# @USAGE: <compiler> <flag>
+# @INTERNAL
+# @DESCRIPTION:
+# Returns shell true if <flag> is supported by given <compiler>,
+# else returns shell false.
 test-flag-PROG() {
+	[[ ${EAPI} == [567] ]] ||
+		die "Internal function ${FUNCNAME} is not available in EAPI ${EAPI}."
+	_test-flag-PROG
```
ulm, line 484: Shouldn't this pass on its arguments?
```suggestion
	_test-flag-PROG "$@"
```

eclass/flag-o-matic.eclass
```diff
@@ -12,16 +13,38 @@
...
 
+# @FUNCTION: setup-allowed-flags
+# @INTERNAL
+# @DESCRIPTION:
 # {C,CPP,CXX,CCAS,F,FC,LD}FLAGS that we allow in strip-flags
 # Note: shell globs and character lists are allowed
 setup-allowed-flags() {
+	[[ ${EAPI} == [567] ]] ||
+		die "Internal function ${FUNCNAME} is not available in EAPI ${EAPI}."
+	_setup-allowed-flags
```
ulm, line 39: AFAICS this doesn't have arguments, but for consistency I'd still do: 
```suggestion
	_setup-allowed-flags "$@"
```

eclass/flag-o-matic.eclass
```diff
@@ -533,33 +585,51 @@ test-flag-PROG() {
...
+# @FUNCTION: test-flags-PROG
+# @USAGE: <compiler> <flag> [more flags...]
+# @INTERNAL
+# @DESCRIPTION:
+# Returns shell true if <flags> are supported by given <compiler>,
+# else returns shell false.
 test-flags-PROG() {
+	[[ ${EAPI} == [567] ]] ||
+		die "Internal function ${FUNCNAME} is not available in EAPI ${EAPI}."
+	_test-flags-PROG
```
ulm, line 623: Ditto.
```suggestion
	_test-flags-PROG "$@"
```

ulm (changes requested):

ulm (approved):

eclass/cron.eclass
```diff
@@ -17,18 +18,30 @@
...
-inherit eutils flag-o-matic
+case ${EAPI:-0} in
+	[0-5]) die "cron.eclass: EAPI ${EAPI} is too old." ;;
+	[67]) inherit eutils flag-o-matic ;;
+	*) die "EAPI ${EAPI} is not supported by cron.eclass." ;;
+esac
 
 EXPORT_FUNCTIONS pkg_postinst
 
 SLOT="0"
```
SoapGentoo, line 30: inherit guards?

eclass/cuda.eclass
```diff
@@ -26,9 +15,18 @@ esac
...
 
 inherit flag-o-matic toolchain-funcs
-[[ ${EAPI} == [56] ]] && inherit eapi7-ver
+
+case "${EAPI:-0}" in
+	[0-5]) die "Unsupported EAPI=${EAPI:-0} (too old) for ${ECLASS}" ;;
+	6) inherit eapi7-ver ;;
+	7) ;;
+	*) die "Unsupported EAPI=${EAPI} (unknown) for ${ECLASS}" ;;
+esac
```
SoapGentoo, line 29: don't we usually have the whitelist outside the inherit guards?

SoapGentoo (approved):

eclass/toolchain.eclass
```diff
@@ -1,14 +1,36 @@
...
+# adjacent packages, support for release and live ebuilds.
+#
+# This eclass unconditionally inherits toolchain-funcs.eclass and all its public
+# variables and helper functions may be considered as part of this eclass's API.
+#
+# This eclass's phase functions are not intended to be mixed and matched, so if
+# any phase functions are overridden the version here should also be called.
+
+case ${EAPI:-0} in
+	0|1|2|3|4*) die "Need to upgrade to at least EAPI=5" ;;
```
thesamesam, line 20: We never approved 4-*, 5-*, so Iw wouldn't worry too much about those, but up to you. I'd go for "I don't speak" in those cases as it's unclear if we could even support them properly in the past.

eclass/cron.eclass
```diff
@@ -4,6 +4,7 @@
 # @ECLASS: cron.eclass
```
thesamesam, line 4: This brought to mind the removal [bug](https://bugs.gentoo.org/693004), but I don't think it's too pressing

eclass/toolchain.eclass
```diff
@@ -1,14 +1,36 @@
...
+# @BLURB: Functions to build sys-devel/gcc
+# @DESCRIPTION:
+# Comprehensive helper and phase functions to build sys-devel/gcc and
+# adjacent packages, support for release and live ebuilds.
+#
+# This eclass unconditionally inherits toolchain-funcs.eclass and all its public
+# variables and helper functions may be considered as part of this eclass's API.
+#
+# This eclass's phase functions are not intended to be mixed and matched, so if
+# any phase functions are overridden the version here should also be called.
```
thesamesam, line 17:
```suggestion
# any phase functions are overridden, the toolchain.eclass version should also be called.
```

eclass/cron.eclass
a17r: hmm right. might as well just leave that one alone

eclass/toolchain.eclass
a17r: didn't want to change the existing logic, maybe we get a second opinion through dev ML review...

eclass/flag-o-matic.eclass
```diff
@@ -533,33 +585,51 @@ test-flag-PROG() {
 # @USAGE: <flag>
 # @DESCRIPTION:
 # Returns shell true if <flag> is supported by the C compiler, else returns shell false.
-test-flag-CC() { test-flag-PROG "CC" c "$@"; }
+test-flag-CC() { _test-flag-PROG "CC" c "$@"; }
```
ulm, line 588: This is very minor, but is there any reason for the inconsistent quoting? Either `CC` shouldn't be quoted (which I'd slightly prefer), or `c` should.

Same for the other functions below.

eclass/gnustep-base.eclass
```diff
@@ -63,22 +67,18 @@ gnustep-base_src_prepare() {
 		eend $?
 	fi
 
-	! has ${EAPI:-0} 0 1 2 3 4 5 && default
+	! has ${EAPI:-0} 5 && default
```
ulm, line 70:
```suggestion
	default
```

`default_src_prepare` in EAPI 5 is a no-op, so this could be simplified.

eclass/gnustep-base.eclass
```diff
@@ -171,36 +171,42 @@ egnustep_env() {
 	die "gnustep-make not installed!"
 }
 
+# @FUNCTION: egnustep_make
+# @DESCRIPTION:
 # Make utilizing GNUstep Makefiles
 egnustep_make() {
 	if [[ -f ./Makefile || -f ./makefile || -f ./GNUmakefile ]] ; then
-		emake ${*} "${GS_ENV[@]}" all || die "package make failed"
+		emake ${*} "${GS_ENV[@]}" all
```
ulm, line 179: Without having checked the details, I say that `${*}` for the function's parameters looks suspicious and should likely be replaced by `"$@"`.

eclass/gnustep-base.eclass
```diff
@@ -171,36 +171,42 @@ egnustep_env() {
...
 # Make-install utilizing GNUstep Makefiles
 egnustep_install() {
 	if [[ ! -d ${EPREFIX}/usr/share/GNUstep/Makefiles ]]; then
 		# avoid problems due to our "weird" prefix, make sure it exists
-		mkdir -p "${D}"${GNUSTEP_SYSTEM_TOOLS}
+		mkdir -p "${D}"${GNUSTEP_SYSTEM_TOOLS} || die
 	fi
 	if [[ -f ./[mM]akefile || -f ./GNUmakefile ]] ; then
-		emake ${*} "${GS_ENV[@]}" install || die "package install failed"
+		emake ${*} "${GS_ENV[@]}" install
```
ulm, line 194: Ditto.

eclass/gnustep-base.eclass
```diff
@@ -171,36 +171,42 @@ egnustep_env() {
...
 	die "no Makefile found"
 }
 
+# @FUNCTION: egnustep_doc
+# @DESCRIPTION:
 # Make and install docs using GNUstep Makefiles
 egnustep_doc() {
 	if [[ -d "${S}"/Documentation ]] ; then
 		# Check documentation presence
 		pushd "${S}"/Documentation || die
```
ulm, line 206: Suppress output of `pushd` (and of `popd` below)?

eclass/toolchain.eclass
ulm: `4*` will match EAPI 40 :smile: so please don't use it.

a17r: hah.

eclass/cron.eclass
```diff
@@ -17,18 +18,33 @@
...
+case ${EAPI:-0} in
+	[0-5]) die "cron.eclass: EAPI ${EAPI} is too old." ;;
+	[67]) inherit eutils flag-o-matic ;;
+	*) die "EAPI ${EAPI} is not supported by cron.eclass." ;;
+esac
+
+if [[ -z ${_CRON_ECLASS} ]]; then
+_CRON_ECLASS=1
 
 EXPORT_FUNCTIONS pkg_postinst
```
SoapGentoo, line 30: after the EAPI whitelist, but before the guard

eclass/cuda.eclass
```diff
@@ -25,10 +14,19 @@ esac
...
+
 if [[ -z ${_CUDA_ECLASS} ]]; then
+_CUDA_ECLASS=1
+
+EXPORT_FUNCTIONS src_prepare
```
SoapGentoo, line 27: after the EAPI whitelist, but before the guard

eclass/cuda.eclass
```diff
@@ -25,10 +14,19 @@ esac
 # @EXAMPLE:
 # inherit cuda
 
+case "${EAPI:-0}" in
```
SoapGentoo, line 17: remove quotes

eclass/gnustep-base.eclass
```diff
@@ -1,16 +1,27 @@
...
 # Inner gnustep eclass, should only be inherited directly by gnustep-base
 # packages
 
-inherit eutils flag-o-matic
+if [[ -z ${_GNUSTEP_BASE_ECLASS} ]]; then
+_GNUSTEP_BASE_ECLASS=1
+
+inherit flag-o-matic
+
+EXPORT_FUNCTIONS pkg_setup src_prepare src_configure src_compile src_install pkg_postinst
```
SoapGentoo, line 18: after the EAPI whitelist, but before the guard

eclass/gnustep-base.eclass
```diff
@@ -1,16 +1,27 @@
...
+
+case ${EAPI:-0} in
```
SoapGentoo, line 20: don't you usually put this before the guard?

SoapGentoo (changes requested):

eclass/cron.eclass
```diff
@@ -17,18 +18,33 @@
...
+# Iterates through a list of known cron packages to set up RDEPEND blockers.
+for _CRON_PN in bcron cronie dcron fcron; do
+	[[ ${_CRON_PN} == ${PN} ]] || RDEPEND+=" !sys-process/${_CRON_PN}"
 done
+unset _CRON_PN
+
+# vixie-cron last-rites: 2019-10-11, 844c30aa; drop eventually
+RDEPEND+="
+	>=sys-process/cronbase-0.3.2
+	!sys-process/vixie-cron
```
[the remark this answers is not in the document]
a17r: We usually keep these for 2 years after removal, alternatively we could already put it behind `[[ EAPI == [67] ]]`.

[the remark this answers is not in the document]
ulm: No, what I'm saying is that the code just works as is:
```
for _CRON_PN in vixie-cron bcron cronie dcron fcron; do
	[[ ${_CRON_PN} == ${PN} ]] || RDEPEND+=" !sys-process/${_CRON_PN}"
done
```
... will add sys-process/vixie-cron as a blocker to all packages, without the need to make it a special case below.

trofi (changes requested):
[going through my review backlog] For lack of a way to remove myself from reviewers marking as `Request changes`.

eclass/cron.eclass
```diff
@@ -17,18 +18,35 @@
 # chosen based on the most common setting among cron ebuilds.
 #
 
-inherit eutils flag-o-matic
+case ${EAPI:-0} in
+	[0-5]) die "cron.eclass: EAPI ${EAPI} is too old." ;;
+	[67]) ;;
+	*) die "EAPI ${EAPI} is not supported by cron.eclass." ;;
+esac
```
ulm, line 25: Looking at it again, this is no longer the recommended style. It should look like this:
```suggestion
case ${EAPI} in
	6|7) ;;
	*) die "${ECLASS}: EAPI ${EAPI:-0} not supported" ;;
esac
```
Same for the other eclasses.

ulm (changes requested):

eli-schwartz:
eutils was recently removed anyway so that much is moot.

PR #20213, opened 2021-04-01
*/*: miscellaneous missing inherits, EAPI 7

thesamesam:


app-accessibility/eflite/eflite-0.4.1-r3.ebuild
```diff
@@ -1,8 +1,9 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI="5"
-inherit autotools eutils user
+EAPI=5
```
scardracs, line 4: Could i ask why in some you have changed to 7 and in some (like that) no?

PR #20215, opened 2021-04-01
x11-wm/herbstluftwm: add 0.9.2

Flowdalic:
Closes: https://bugs.gentoo.org/779646


x11-wm/herbstluftwm/herbstluftwm-0.9.2.ebuild
```diff
@@ -0,0 +1,93 @@
...
+		popd >/dev/null || die
+	fi
+}
+
+src_install() {
+	cmake_src_install
+
+	if ! use examples; then
+		rm -r "${ED}"/usr/share/doc/${PF}/examples || die
+	fi
```
juippis, line 85: This kind of removing is really frowned upon, especially since
```
# du -csh /usr/share/doc/herbstluftwm-0.9.2/examples/
84K	/usr/share/doc/herbstluftwm-0.9.2/examples/
84K	total
```

juippis (approved):
You should also move the `.html` files, installed by `IUSE="+doc"` under `/usr/share/doc/${PF}/html`

x11-wm/herbstluftwm/herbstluftwm-0.9.2.ebuild
Flowdalic: Fair point, will drop the examples USE flag.

PR #20218, opened 2021-04-01
media-plugins/vdr-osdteletext: version bump to 1.1.1

madmartin:


media-plugins/vdr-osdteletext/vdr-osdteletext-1.1.1.ebuild
```diff
@@ -0,0 +1,46 @@
...
+DESCRIPTION="VDR Plugin: Osd-Teletext displays the teletext/videotext on the OSD"
+HOMEPAGE="https://projects.vdr-developer.org/projects/plg-osdteletext https://github.com/vdr-projects/vdr-plugin-osdteletext/"
+SRC_URI="https://github.com/vdr-projects/vdr-plugin-osdteletext/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-2+ public-domain" #teletext2.ttf, not copyrightable
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+DEPEND="media-video/vdr"
+RDEPEND="${DEPEND}"
```
juippis, line 19: You should depend on the acct-* packages here.

juippis (approved):

media-plugins/vdr-osdteletext/vdr-osdteletext-1.1.1.ebuild
madmartin: Why?
Or more clearly asked, do you propose an additional depend on the acct-* ? (which IMHO do not exist currently - the acct-*/vdr packages are in my private overlay currently...) 
or do you want to replace dependency from media-video/vdr to acct-user/vdr?

juippis: You have `fowners -R vdr:vdr "${vdr_user_home}/.local"` in your ebuild. I know they get inherited through *vdr* main package, but we normally don't depend on dependency-chain like that. What if *vdr* dropped the acct- dependencies, or someone uses an older versions, etc. They are needed in this ebuild, so this ebuild should depend on them.

madmartin: So next time, when acct-*/vdr is in the main tree, I'll add on every media-plugins/vdr-* and ADDITIONAL dependency on acct-user/vdr when it is referenced, right?

juippis: > So next time, when acct-_/vdr is in the main tree, 

Oh, right, **sigh**. I forgot it's not done yet...

but yeah when we get the acct-* packages, everything that uses them should depend on them. These ebuilds are so oldschool I assume everything's ported to GLEP-81 already :P

PR #20221, opened 2021-04-01
Fix failing `sys-devel/llvm-roc-4.1.0` build

rslawson:
CMake's use of `sys/resource.h` appears to attempt access to `/dev/nvidiactl`. Use `addpredict` to prevent this failing the build. I kinda forgot to file a bug for this, whoops. It was causing me some issues last night when I was trying to update all the ROCm packages, and applying this fix to the Gentoo repo locally before attempting to build fixed the issue.


candrews:
Thanks for the PR!

The bug is https://bugs.gentoo.org/789693

Fixed in f32ed0e9c59158047fb10f8e8c0d27d232f015b8

PR #20224, opened 2021-04-01
sys-libs/libselinux: add musl patch

xmhd:
pid_t function in libselinux requires patching for musl.

Update copyright date in header of -9999 ebuild.

@perfinion


thesamesam:
Fixed long ago.

PR #20227, opened 2021-04-02
x11-libs/libva: bump to v2.11.0

Karlson2k:
The shell grouping patch is not needed any more as it has been integrated upstream.

This time `.tar.gz` was used as sources tarball because `.tar.bz2` has timestamps in the future, which break `autotools` and `make` builds.


Karlson2k:
@thesamesam could you check this one and other linked, including https://github.com/gentoo/gentoo/pull/20235?

thesamesam:
Sure. Been on my list, sorry for delay. Could you possibly do them all in one PR next time around, so the CI is nice and green? :)

Thanks for your work!

Karlson2k:
> Sure. Been on my list, sorry for delay. Could you possibly do them all in one PR next time around, so the CI is nice and green? :)

My PRs were published in two-three hours after upstreams update, I rushed to bring the latest and the greatest. :)

OK. The idea was to simplify updating, but if combined PR will work better, I'll do it that way.

 
> I also suggest adding yourself to metadata.xml to be CCed on bugs etc

Sure, next time.

> Thanks for your work!

Thank you for checking and merging.

PR #20233, opened 2021-04-02
Bump lite 2.4.0

oz123:


dev-libs/libite/libite-2.4.0.ebuild
```diff
@@ -0,0 +1,23 @@
...
+HOMEPAGE="https://github.com/troglobit/libite"
+SRC_URI="https://github.com/troglobit/libite/releases/download/v${PV}/${P}.tar.xz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="static-libs"
+
+src_configure() {
+	econf --enable-static=$(usex static-libs)
```
juippis: Since nothing depends on this package with static libs, can we disable them?
https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0302

oz123: Sure, I'll just change:
```
```

juippis: No no, get rid of the USE flag altogether. Then do `econf --enable-static=false`

oz123: Thanks for the review. I have implemented the requested changes.

juippis (approved):
Thanks!

PR #20230, opened 2021-04-02
*/*: EAPI 5->7, indirect and superfluous inherits

thesamesam:

PR #20243, opened 2021-04-03
app-arch/makeself: Bump to 2.4.3

00-matt:
Includes changes from #20230.


00-matt:
There's also `app-arch/unmakeself` which I think can be last-rited and dropped?

thesamesam:
File a bug or similar re unmakeself?

PR #20249, opened 2021-04-03
Removing unused inherits - Updating EAPI - Updating copyright

scardracs:
All the packages with python have been tested personally against python 3.9.4


www-servers/nginx/nginx-1.19.9.ebuild
```diff
@@ -278,7 +278,7 @@ done
 # @TODO: jbergstroem: remove on next release series
 IUSE="${IUSE} nginx_modules_http_spdy"
 
-CDEPEND="
+COMMON_DEPEND="
```
thesamesam, line 281: I wouldn't bother with stuff like this. You are right because we should avoid using XDEPEND where X is a single-character - they're supposed to be reserved for future EAPIs.

But it's not worth messing with existing style and confusing maintainers, etc. If we're going to do it, we should do that in a separate PR and make it really clear in the commit message(s).

www-servers/nginx/Manifest
```diff
@@ -1,12 +1,5 @@
 DIST modsecurity-2.9.3.tar.gz 4307670 BLAKE2B 337ea15cc8805af7ab43aed8aecf4c72ccc586d0d7e9d9b91f036a61baa70d1ac8b4ad8045a2bb7a13515912a15fba7d3cdb9670ae6730de43b1e44ee90ded6d SHA512 4e1ea5dd8edadf8f630e4fe92a200d3a8e78963fce3128b5975a1e1ecd0e8bf9ceecd9905c95f8c508932ccd837f1d8ae8bb2ba423307718c3c6a4ae9b783ddd
 DIST nginx-1.18.0.tar.gz 1039530 BLAKE2B a8962a6af96acb043ff0c3dc4ad5192083773c449950aff53b01f6f7c46a2a540eb061a43432acccd753fa71067b1451d75f440ba5526575b78608be9d40a50b SHA512 8c21eeb62ab6e32e436932500f700bd2fb99fd2d29e43c08a5bfed4714c189c29c7141db551fcd5d2437303b7439f71758f7407dfd3e801e704e45e7daa78ddb
-DIST nginx-1.19.2.tar.gz 1048727 BLAKE2B db00b5945ed82fe90059269ba987931642981a2b8fa1cb24e0d842b97d89857993f99443c644d9217af29430b868ee8efd90dbc5daefd719d0f98c0766601554 SHA512 f6232d58a91db486cf36cc027ab84f04ad25aeafe586051043f0f3295dad5f475094f7cedad142ef0f56ff15a0971d700d6ec34b68697c462364650c49be8fcc
```
thesamesam, line 3: I wouldn't bother cleaning up old for maintained packages. Too controversial and it might be the maintainer want to only drop certain versions.

In this case, the maintainer is active, and you could just ask him if you need them cleaned up. But in any case, try not to mix tasks here - it's not clear to me why we're cleaning these up.

app-doc/elisp-manual/elisp-manual-26.3.ebuild
```diff
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
```
thesamesam, line 1: We only update copyright when the ebuilds changed. Did they change this year and someone forgot to update it?

app-doc/elisp-manual/Manifest
```diff
@@ -1,14 +1 @@
-DIST elisp-manual-18-1.03.tar.gz 1246385 BLAKE2B af23fd63bf56d5b76785e929aa05e628143d33baf03c30746ee18e8c278113c2874fbe2689c1b361a2d9efd6a92aa6c39a92317f2f36f282f313afa1fc4876bd SHA512 5ba37650482a34918924beafdf214e8beb9adb5e94976020fb8a231d6d5a2dfb4ceb0659172412b193415634a93929462aba4109ebbbef31feca96397a3b2c30
```
thesamesam, line 1: See other comments re nginx and cleaning up.

Specifically for emacs: these versions correspond to old(er) Emacs slots. It's useful to have an older elisp manual even if it's technically not needed(?) by the Emacs package. So, we shouldn't clean these up, even if the depgraph allows it.

app-accessibility/brltty/brltty-6.1.ebuild
```diff
@@ -56,15 +59,19 @@ DEPEND="
...
+	${DEPEND}
 	java? ( >=virtual/jre-1.4 )
 "
+
 BDEPEND="
-	virtual/pkgconfig
 	java? ( >=virtual/jdk-1.4 )
 	nls? ( virtual/libintl )
 	python? ( >=dev-python/cython-0.16[${PYTHON_USEDEP}] )
+	virtual/pkgconfig
```
thesamesam, line 74: Normally, we put all unconditional then conditional, so it was originally OK.

app-accessibility/brltty/brltty-6.1.ebuild
```diff
@@ -56,15 +59,19 @@ DEPEND="
 		x11-libs/libXfixes
 		x11-libs/libXt
 		x11-libs/libXtst
-	)"
-RDEPEND="${DEPEND}
+	)
```
thesamesam, line 62: I probably wouldn't mess with stuff like this for maintained packages unless it's _wrong_. It's style and we don't want to upset people.

app-accessibility/brltty/brltty-6.1.ebuild
```diff
@@ -46,6 +48,7 @@ DEPEND="
 		app-accessibility/flite
 		app-accessibility/speech-dispatcher
 	)
+	sys-apps/dbus
```
thesamesam, line 51: [We](https://devmanual.gentoo.org/general-concepts/dependencies/) do all unconditional then conditional, so it was right before

app-accessibility/brltty/brltty-6.1.ebuild
```diff
@@ -3,13 +3,12 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7,8} )
+PYTHON_COMPAT=( python3_{7..9} )
```
thesamesam, line 6: Did tests pass with 3.9?

www-servers/nginx/nginx-1.19.9.ebuild
scardracs: Good point. By now it had become so habitual that I hadn't thought about adding a note on it

www-servers/nginx/Manifest
scardracs: I have asked to whissi, but no answer since 04/04

app-accessibility/brltty/brltty-6.1.ebuild
scardracs: I have tested it and yes, it pass

app-doc/elisp-manual/elisp-manual-26.3.ebuild
scardracs: I have dropped the commit so now it is restored to original version

app-accessibility/brltty/brltty-6.1.ebuild
scardracs: Dropped commit

PR #20264, opened 2021-04-05
cmake-utils--

a17r:
In the year 2021 we don't use cmake-utils.


a17r:
@keriharris please test `dev-lang/mozart` and `dev-lang/mozart-stdlib` as it failed in my case and I didn't have time to look into it:
```
!!! ERROR: Couldn't find a VM dep
 * Unable to determine VM for building from dependencies
```

It seems to me that the following is used as a habitual copy-paste:
```
BUILD_DIR="${S}/build"
CMAKE_USE_DIR="${S}"
```

It was certainly not required in `dev-lang/swi-prolog` so I dropped it from there.

orlitzky:
the change to sci-mathematics/arb is fine, thanks

a17r:
@junghans commit 4af2eadb0a2508ab22bfba286961fd29adb9e2c4 gives no explanation why `dev-cpp/kokkos` needs to be pinned down to `3.0*`. Please consider making more verbose git summaries.

junghans:
> @junghans commit [4af2ead](https://github.com/gentoo/gentoo/commit/4af2eadb0a2508ab22bfba286961fd29adb9e2c4) gives no explanation why `dev-cpp/kokkos` needs to be pinned down to `3.0*` in `sci-physics/lammps-20200303`. Please consider making more verbose git summaries.

There is an api change between `kokkos:3.0` and `kokkos:3.1` and `lammps-20200303` only builds against the 3.0 api. New versions of lammps work with the 3.1 api.

a17r:
> There is an api change between `kokkos:3.0` and `kokkos:3.1` and `lammps-20200303` only builds against the 3.0 api. New versions of lammps work with the 3.1 api.

Thanks for commenting. ;) Fine to cleanup the `votca*` stuff?

atupone (approved):
cvc4 changes seems ok to me

junghans:
> Fine to cleanup the `votca*` stuff?

Looks good to me.

sarnex (approved):

dev-lang/mozart/mozart-2.0.1.ebuild
```diff
@@ -31,50 +32,47 @@ DEPEND="${RDEPEND}
...
+	cmake_src_prepare
 }
 
 src_configure() {
-	mycmakeargs=(
+	local mycmakeargs=(
 		-DMOZART_BOOST_USE_STATIC_LIBS=OFF
-		-DEMACS=$(if use emacs; then echo /usr/bin/emacs; fi)
-		)
+		-DEMACS=$(usex emacs "/usr/bin/emacs" "")
```
ulm, line 51: This is very suspicious, emacs USE flag without inheriting elisp-common. But of course it was like this before, so no blocker for this PR.

@keriharris

ulm (approved):
Otherwise LGTM.

I've double-checked that the package maintained by me (net-im/openmittsu) builds and installs.

dev-lang/mozart/mozart-2.0.1.ebuild
keriharris: This is being tracked as a separate issue in bug 194604 [1].

[1] https://bugs.gentoo.org/194604

dev-lang/mozart/mozart-2.0.1.ebuild
```diff
@@ -1,9 +1,10 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-inherit cmake-utils java-pkg-2 java-ant-2
+BUILD_DIR="${S}/build" # TODO: really?
```
keriharris, line 6: This line can be dropped. The default BUILD_DIR used by the cmake eclass is fine.

dev-lang/mozart-stdlib/mozart-stdlib-2.0.1.ebuild
```diff
@@ -1,11 +1,11 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-inherit cmake-utils
-
 MY_PN=mozart2-stdlib
+BUILD_DIR="${S}/build" # TODO: really?
```
keriharris, line 7: This line can be dropped. The default BUILD_DIR used by the cmake eclass is fine.

keriharris:
> @keriharris please test `dev-lang/mozart` and `dev-lang/mozart-stdlib` as it failed in my case and I didn't have time to look into it:
> 
> ```
> !!! ERROR: Couldn't find a VM dep
>  * Unable to determine VM for building from dependencies
> ```

I've tested both mozart and mozart-stdlib. They both install and run with the changes to use the cmake eclass.

I can't reproduce the Java "VM dep" error, but the output from "eselect java-vm list" and the NV_DEPEND output in the above error message should provide a clue.

> It seems to me that the following is used as a habitual copy-paste:
> 
> ```
> BUILD_DIR="${S}/build"
> CMAKE_USE_DIR="${S}"
> ```
> 
> It was certainly not required in `dev-lang/swi-prolog` so I dropped it from there.

Dropping both BUILD_DIR and CMAKE_USE_DIR is fine.

PR #20271, opened 2021-04-05
dev-lang/fpc: 3.2.2 bump

band-a-prend:
- Switch to EAPI-7.
- The `ide` sources is in "packages" subdirectory in 3.2.0 release.
It's now build unconditionally so USE="ide" is dropped.
- Remove `RDEPEND` `ide? (!dev-lang/fpc-ide)` as `dev-lang/fpc-ide` was dropped from portage in 2014.
- Drop `-j1` after emake calls. If appropriate issue will take place then it could be restored.
- In the src_install() for "ide" configs now excplicit filenames are used within "sed".
- For amd64 the built `libpas2jslib.so` conditionally moved to `/usr/lib64` directory.

Closes: https://bugs.gentoo.org/730720

@Amynka 
could you please review this pull request?


band-a-prend:
FPC 3.2.2 was released on 20 may 2021.

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+	mirror://sourceforge/freepascal/fpc-${PV}.source.tar.gz
+	amd64? ( mirror://sourceforge/freepascal/${P}.x86_64-linux.tar )
+	x86? ( mirror://sourceforge/freepascal/${P}.i386-linux.tar )
+	doc? ( mirror://sourceforge/freepascal/Documentation/${PV}/doc-html.tar.gz -> ${P}-doc-html.tar.gz )
+"
+
+SLOT="0"
+LICENSE="GPL-2 LGPL-2.1-with-linking-exception"
+KEYWORDS="~amd64 ~x86"
+IUSE="doc source"
```
SoapGentoo, line 20: Change to canonical order:
```
LICENSE
SLOT
KEYWORDS
IUSE
```

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+
+RESTRICT="strip" #269221
+
+S="${WORKDIR}/fpcbuild-${PV}/fpcsrc"
```
SoapGentoo, line 24: put after `SRC_URI`

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+		if "$(tc-getLD)" --version | grep -q "GNU gold"; then
+			eerror "fpc does not function correctly when built with the gold linker."
+			eerror "Please select the bfd linker with binutils-config."
+			die "GNU gold detected"
+		fi
+	fi
+}
+
+src_unpack() {
+	case "${ARCH}" in
```
SoapGentoo, line 38:
```suggestion
	case ${ARCH} in
```
not needed in bash

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+
+	unpack ${A}
+
+	tar -xf "${PN}-${PV_BIN}.${FPC_ARCH}-linux/binary.${FPC_ARCH}-linux.tar" || die "Unpacking binary.${FPC_ARCH}-linux.tar failed!"
+	tar -xzf "base.${FPC_ARCH}-linux.tar.gz" || die "Unpacking base.${FPC_ARCH}-linux.tar.gz failed!"
+}
+
+src_prepare() {
+	default
+	find "${WORKDIR}" -name Makefile -exec sed -i -e 's/ -Xs / /g' {} + || die
```
SoapGentoo, line 52: use the funky read while loop

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+
+	# let the pkg manager compress man files
+	sed -i '/find man.* gzip /d' "${WORKDIR}"/fpcbuild-"${PV}"/install/man/Makefile || die
+
+	# make the compiled binary check for fpc.cfg under the prefixed /etc/ path
+	hprefixify "${WORKDIR}"/fpcbuild-"${PV}"/fpcsrc/compiler/options.pas
+}
+
+set_pp() {
+	case "${ARCH}" in
```
SoapGentoo, line 62:
```suggestion
	case ${ARCH} in
```

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+		INSTALL_DOCDIR="${ED}/usr/share/doc/${PF}" \
+		INSTALL_MANDIR="${ED}/usr/share/man" \
+		INSTALL_SOURCEDIR="${ED}/usr/lib/fpc/${PV}/source"
+
+	emake "$@" compiler_install rtl_install packages_install utils_install
+
+	dosym ../lib/fpc/"${PV}"/ppc"${FPC_ARCH}" /usr/bin/ppc"${FPC_ARCH}"
+
+	cd "${S}"/../install/doc || die
+	emake "$@" installdoc
```
SoapGentoo, line 119: what about
```suggestion
	emake -C "${S}"/../install/doc "$@" installdoc
```
?

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+
+	cd "${S}"/../install/man || die
+	emake "$@" installman
```
SoapGentoo, line 122: same

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+
+	if use doc ; then
+		cd "${S}"/../../doc || die
+		dodoc -r *
+	fi
```
SoapGentoo, line 127: `use doc && dodoc "${S}"/../../doc/.`

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+
+	if use source ; then
+		cd "${S}" || die
+		shift
+		emake PP="${ED}/usr/bin/ppc${FPC_ARCH}" "$@" sourceinstall
+		find "${ED}/usr/lib/fpc/${PV}/source" -name '*.o' -exec rm {} \;
```
SoapGentoo, line 133:
```suggestion
		find "${ED}/usr/lib/fpc/${PV}/source" -name '*.o' -delete || die
```

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+		-i "${ED}/etc/fppkg.cfg" \
+		-i "${ED}/etc/fppkg/default" \
+		-i "${ED}/usr/lib/fpc/${PV}/ide/text/fp.cfg" \
+		|| die
+
+	rm -r "${ED}"/usr/lib/fpc/lexyacc || die
+
+	case "${ARCH}" in
+		amd64)
+			mkdir -p "${ED}/usr/$(get_libdir)/"
```
SoapGentoo, line 151:
```suggestion
			mkdir -p "${ED}/usr/$(get_libdir)/" || die
```

SoapGentoo (changes requested):

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Could you please clarify? There are numerous Makefiles in subdirectories and of it's subdirectories too.

ionenwks: Not sure I see the intend either, find {} + will pass every arguments to a single sed call and is clean/simple enough (it also returns false if the -exec command failed when using '+', wouldn't with '\;' -- so error checking should be okay).

If we're talking of loop, is this about doing 1 separate sed call for each 913 Makefiles? That doesn't sound too nice.

That aside, could drop the /g for efficiency, I don't think any Makefile has "-Xs -Xs"

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+			mv "${ED}"/usr/lib/*.so "${ED}/usr/$(get_libdir)/" || die ;;
+	esac
+}
+
+pkg_postinst() {
+	if [[ -z "${REPLACING_VERSIONS}" ]] ; then
+		einfo "To read the documentation in the fpc IDE, enable the doc USE flag"
+	fi
+}
```
ionenwks, line 160: I know of lot of these were already in the old ebuild, but einfo has very little visibility unless user stares at the build log (for a one time message, elog is typically preferable). Also shouldn't it not show if the doc USE is already enabled? But I question the need of this message altogether.

**If** you wish to clarify that USE=doc has runtime implications, then adding a local definition in metadata.xml may make more sense. Users may often look at use descriptions when something is amiss.

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+	case "${ARCH}" in
+		amd64)	FPC_ARCH="x86_64"    PV_BIN="${PV}" ;;
+		x86)	FPC_ARCH="i386"      PV_BIN="${PV}" ;;
+		*)	die "This ebuild doesn't support ${ARCH}." ;;
+	esac
+
+	unpack ${A}
+
+	tar -xf "${PN}-${PV_BIN}.${FPC_ARCH}-linux/binary.${FPC_ARCH}-linux.tar" || die "Unpacking binary.${FPC_ARCH}-linux.tar failed!"
+	tar -xzf "base.${FPC_ARCH}-linux.tar.gz" || die "Unpacking base.${FPC_ARCH}-linux.tar.gz failed!"
```
ionenwks, line 47: Can't these use `unpack ./file.tar.gz` instead of tar?

ionenwks (commented):
Gave it a quick look as well if want to improve the ebuild a bit more.

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+	emake PP="${pp}" compiler_cycle AS="$(tc-getAS)"
+
+	# Save new compiler from cleaning...
+	cp "${S}/compiler/ppc${FPC_ARCH}" "${S}/ppc${FPC_ARCH}.new" || die
+
+	# ...rebuild with current version...
+	emake PP="${S}/ppc${FPC_ARCH}.new" AS="$(tc-getAS)" compiler_cycle
+
+	# ..and clean up afterwards
+	rm "${S}/ppc${FPC_ARCH}.new" || die
```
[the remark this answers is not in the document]
band-a-prend: It seems that `emake` command requires full compiler path here. Otherwise compiler command couldn't be found.

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Done

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Done

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Done

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Drop `g` (global) and `|| die` if I undestand right, isn't it, @ionenwks 

Yes, there are only affected ` -Xs ` in default configuration and untouched conditional addition reassignments.

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Done

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Thank you, I fixed with offered way.

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Done

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Done

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Done

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Done

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+
+S="${WORKDIR}/fpcbuild-${PV}/fpcsrc"
+
+pkg_pretend() {
+	if [[ "${MERGE_TYPE}" != binary ]]; then
+		# Bug 475210
+		if "$(tc-getLD)" --version | grep -q "GNU gold"; then
+			eerror "fpc does not function correctly when built with the gold linker."
+			eerror "Please select the bfd linker with binutils-config."
+			die "GNU gold detected"
```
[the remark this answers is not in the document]
band-a-prend: Thank you! I replaced with suggested block.

dev-lang/fpc/fpc-3.2.2.ebuild
```diff
@@ -0,0 +1,160 @@
...
+	cd "${S}"/../install/man || die
+	emake "$@" installman
+
+	if use doc ; then
+		cd "${S}"/../../doc || die
+		dodoc -r *
+	fi
+
+	if use source ; then
+		cd "${S}" || die
```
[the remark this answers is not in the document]
band-a-prend: Drop `cd`, shift is still required here.

dev-lang/fpc/fpc-3.2.2.ebuild
ionenwks: No no, keep || die, I was just probing at what made have brought this up, e.g. `{} \;` is bad with return codes, but + isn't.

/g is to replace all instances on a single line, but if there's only one  there's no need for it. I just don't like seeing these when it's not actually needed.

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Add condition to check USE="-doc". That use flag installs common html documentation as well, so reading it from ide is feature in my mind rather than runtime option.

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Replaced by unpack.

dev-lang/fpc/fpc-3.2.2.ebuild
band-a-prend: Restore `|| die`

dev-lang/fpc/fpc-3.2.2.ebuild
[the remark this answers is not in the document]
ionenwks: Seem to have been a misunderstanding, I wasn't giving the replacement but rather instructions to do so and showing what this function did.

tc-ld-disable-gold is a function defined in [toolchain-funcs.eclass](https://devmanual.gentoo.org/eclass-reference/toolchain-funcs.eclass/index.html) which is already inherited by this ebuild.

In the event gold has known problems, you can invoke this function before configuring/compiling, like adding this early in src_compile:
```
tc-ld-disable-gold # bug 475210
```
After doing so, pkg_pretend() can be removed as it'll serve no purpose anymore.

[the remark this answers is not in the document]
band-a-prend: I checked that issue is still reproduced and FPC failed to build with `ld.gold` ( after `binutils[gold,default-gold]` rebuild ), so I place `tc-ld-disable-gold` in `src_configure` phase - it's usual place for it as I could see from other ebuilds.

PR #20277, opened 2021-04-06
dev-util/cutter: change radare2 to rizin in description

arkamar:
We should change the description as well.


ajakk:
Whoops, thanks!

PR #20285, opened 2021-04-06
media-sound/fluid-soundfont: Proxy maintenance request.

pinkflames:
This commit adds me to the proxy maintainers of the fluid-soundfont
package. As is customary, the ebuild has been overhauled:
- All SRC_URI were dead - this fixes that by using a new GH repository,
documenting the Fluid R3 soundfont and its open source history.
- Makes installation of the independent GM and GS soundfonts a user choice.
- Always installs the TiMidity config with the GM soundfont.
- Drops RDEPEND on app-eselect/eselect-timidity since TiMidity users
will have it anyway and soundfonts themselves do not need it.
- Adds BDEPEND on media-libs/flac to decompress the soundfonts (a
limitation imposed by GitHub's hard limit on max file size).

TODO:
  - [ ] Add readme-gentoo-r1.eclass to print elog only once


media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
```diff
@@ -1,29 +1,62 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-DESCRIPTION="The Fluid R3 soundfont"
-HOMEPAGE="http://musescore.org/en/handbook/soundfont"
```
vaukai, line 7: https for HOMEPAGE

fordfrog:
i'm glad you want to maintain it. i would also like to have sound@gentoo.org as the maintainer so that i (as a member of the sound project) can keep track of the package. there are many improvements and changes in your work so it will take some time to review it and test it :-)

pinkflames:
I have made the change locally but I'm still waiting for an ACK from the Proxy maintenance project before I push the commit.

thesamesam:
Go for it, the more help the better :)

media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
```diff
@@ -1,29 +1,62 @@
...
-		doins "${WORKDIR}"/timidity.cfg
+		doins contrib/timidity.cfg
+	fi
+}
+
+pkg_postinst() {
+	elog "We're not lawyers but, according to Debian, music produced from samples"
+	elog "under MIT licence may itself be subject to it, if the samples constitute"
+	elog "a substantial portion of the work. In particular this means reproducing"
+	elog "the copyright notice and the licence when distributing derived works."
```
juippis, line 55:
```suggestion
	elog "Music produced from samples under MIT licence may itself be subject to"
	elog "it, if the samples constitute a substantial portion of the work. In" 
	elog "particular this means reproducing the copyright notice and the licence"
	elog "when distributing derived works. This is not legal advice."
```

juippis: Besides I heard an opinion that this should be printed once, so you should utilize *readme-gentoo-r1.eclass*.

media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
```diff
@@ -1,29 +1,62 @@
...
 
 LICENSE="MIT"
 SLOT="0"
 KEYWORDS="amd64 ~ppc64 x86"
-IUSE="timidity"
+IUSE="+gm gs"
+
+BDEPEND="media-libs/flac"
+
+REQUIRED_USE="|| ( gm gs )"
```
juippis, line 23: No one likes `REQUIRED_USE`. Not users, not arch testers. Could you simply construct a way to choose gm if both are disabled, while printing a message to the user?

media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
```diff
@@ -1,29 +1,62 @@
...
+HOMEPAGE="https://github.com/pinkflames/fluid-soundfont"
 SRC_URI="
-	http://prereleases.musescore.org/soundfont/${PN}_${PV}.tar.gz
-	timidity? ( https://dev.gentoo.org/~hwoarang/distfiles/timidity.cfg.bz2 )"
+	https://github.com/pinkflames/${PN}/archive/${COMMIT}.tar.gz -> ${PN}-${COMMIT}.tar.gz
+"
 
 LICENSE="MIT"
 SLOT="0"
 KEYWORDS="amd64 ~ppc64 x86"
```
juippis, line 18: FYI, not really comfortable pushing an update like this to a stable ebuild. Please reset the KEYWORDS to ~ and revbump your work.

media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
```diff
@@ -1,29 +1,62 @@
...
+REQUIRED_USE="|| ( gm gs )"
+
+DOCS=( original/COPYING original/README )
+
+S="${WORKDIR}"/${PN}-${COMMIT}
+
+src_prepare() {
+	einfo "Using the flac tool to decompress the soundfont(s):"
+	use gm && flac --force-raw-format --endian=little --sign=signed -d original/FluidR3_GM.sf2.flacy -o FluidR3_GM.sf2
+	use gs && flac --force-raw-format --endian=little --sign=signed -d original/FluidR3_GS.sf2.flacy -o FluidR3_GS.sf2
```
juippis, line 32: Open properly, `if use; then`
and add `flac ... || die`

media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
```diff
@@ -1,29 +1,62 @@
...
 
-RDEPEND="timidity? ( app-eselect/eselect-timidity )"
+	default
```
juippis, line 34: `default` should be the first thing called in `src_prepare`.

media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
pinkflames: I understand. It's mostly a remnant from when I was trying to use GitHub's Large File Storage but had to go with the FLACy hack, because it would never have worked with the Gentoo mirror traffic alone, let alone any users accessing the files directly.

media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
pinkflames: Thank you for reminding me what I should have noticed myself by now. Will be fixed in the next commit.

media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
pinkflames: My understanding was that calling it afterwards was a valid choice but sure; fixed.

media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
pinkflames: I would have thought that installing a feature `foo` while `-foo` USE flag is set, would be a very, very bad and terrible thing to do but, if that really would be acceptable, that's not a difficult thing to write, I imagine. Alternatively GM could be installed unconditionally but it makes up the vast bulk of the package size, so it feels a bit bad to foist that on someone who maybe really needs just the SFX extension.

media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
pinkflames: On one hand, yes, this might be mundane enough to print it only once, so I have added it to the TODO for later but in general I do not believe that the issue is packages print too much elog messages - some people do not read them even when doing the first installation of one particular package and having been told to read and and follow instructions that the particular ebuild will print.

media-sound/fluid-soundfont/fluid-soundfont-3.1.ebuild
juippis: > I would have thought that installing a feature foo while -foo USE flag is set, would be a very, very bad and terrible thing to do but, if that really would be acceptable

It is arguable indeed. I'm doing my best to get rid of REQUIRED_USE as much as possible.

I hate to use existing ebuilds in the repository as an example, as there are some horrible ebuilds present, but check virtualbox, freeciv and qudos for example that does this already. There's many more...

> it feels a bit bad to foist that on someone who maybe really needs just the SFX extension.

Then enable `gs` if both are disabled, and do `IUSE="+gm gs"`?

Would this package be functional at all without either USE flag enabled, do we need the REQUIRED_USE?

pinkflames: -gm -gs would result in effectively NOOP as only README and LICENCE would be installed. The GS while small is also, IMO, not what people expect - it's literally effect sounds only. I'm sure there's someone out there that only wants cheesy effects like clapping hands, washing waves or dog barks but majority of people who use MIDI probably will want musical instrument samples. Besides we already set +gm - the required use would only kick in if someone had -gm USE set, which itself would be weird, since it's a local USE flag and should not be set globally.

fordfrog: and how about dropping the use flags and simply install both?

pinkflames: That would cause confusion - how many people would know if they need the GM or the GS file? I certainly didn't when I began my work on this PR.

PR #20286, opened 2021-04-07
*/*: add missing multilib inherit for get_libname

thesamesam:


ceamac:
Should also add `media-libs/netpbm-10.86.21` to the list
Not sure if I should do this in a different PR or just mention it here

thesamesam:
Already merged as part of some other work.

@ceamac new PR?

PR #20287, opened 2021-04-07
*/*: add missing virtual/pkgconfig (B)DEPEND

thesamesam:
TODO: do eclasses separately or just drop

### DIFF
--- a/sys-fs/xfsdump/xfsdump-3.1.9.ebuild
+++ b/sys-fs/xfsdump/xfsdump-3.1.9.ebuild
@@ -1,9 +1,9 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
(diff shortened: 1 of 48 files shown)


x11-misc/xbattbar/xbattbar-1.4.9.ebuild
```diff
@@ -18,6 +18,7 @@ DEPEND="
 "
 RDEPEND="
 	${DEPEND}
+	virtual/pkgconfig
```
SoapGentoo, line 21: wat?

sys-fs/xfsdump/xfsdump-3.1.9.ebuild
```diff
@@ -23,10 +23,12 @@ RDEPEND="
 	ncurses? ( sys-libs/ncurses:0= )
 "
 DEPEND="${RDEPEND}
+	ncurses? ( virtual/pkgconfig )
```
SoapGentoo, line 26: wrong dependency type

SoapGentoo (changes requested):

PR #20290, opened 2021-04-07
python eclasses: remove old EAPI support

mgorny:
Should require only cleaning up old Portage.

### DIFF
--- a/eclass/python-utils-r1.eclass
+++ b/eclass/python-utils-r1.eclass
@@ -7,7 +7,7 @@
 # @AUTHOR:
 # Author: Michał Górny <mgorny@gentoo.org>
 # Based on work of: Krzysztof Pawlik <nelchael@gentoo.org>
-# @SUPPORTED_EAPIS: 5 6 7
+# @SUPPORTED_EAPIS: 6 7
 # @BLURB: Utility functions for packages with Python parts.
 # @DESCRIPTION:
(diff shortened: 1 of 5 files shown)


eclass/python-utils-r1.eclass
```diff
@@ -1169,17 +1160,14 @@ python_fix_shebang() {
 		done < <(find -H "${path}" -type f -print0 || die)
 
 		if [[ ! ${any_fixed} ]]; then
-			local cmd=eerror
-			[[ ${EAPI} == 5 ]] && cmd=eqawarn
-
-			"${cmd}" "QA warning: ${FUNCNAME}, ${path#${D%/}} did not match any fixable files."
+			eerror "QA warning: ${FUNCNAME}, ${path#${D%/}} did not match any fixable files."
```
thesamesam, line 1163: Not eqawarn?

thesamesam (approved):
On mobile so I wasn’t able to super easily check for other possible cleanups but changes look good.

eclass/python-utils-r1.eclass
mgorny: Nope, it's a fatal warning now ;-).

mgorny: I've changed it to 'QA error'.

PR #20294, opened 2021-04-07
dev-python/spyder: bump some deps to py3.9

AndrewAmmerlaan:
Bumps some of my packages to py3.9.


juippis (commented):
I'm kind of 50-50 on merging this or not... because I see you've revbumped some packages with stable keywords. You don't have to do that, as python targets are controlled via profiles. You can add python-3.9 to a stable ebuild.

So yeah, just add python-3.9 to stable ebuilds, so we don't cause more work for arch teams.

AndrewAmmerlaan:
> I'm kind of 50-50 on merging this or not... because I see you've revbumped some packages with stable keywords. You don't have to do that, as python targets are controlled via profiles. You can add python-3.9 to a stable ebuild.
> 
> So yeah, just add python-3.9 to stable ebuilds, so we don't cause more work for arch teams.

I know, but the problem is that some dependencies have python3_9 only on versions with unstable keywords.
E.g. `dev-python/abydos` depends on `dev-python/nltk` but the only version of nltk with python3_9 has unstable keywords. So I can do two things: I either wait until the new version of `dev-python/nltk` becomes stable (which delays the transition to python3_9), or I revision bump `dev-python/abydos` to unstable keywords and then add python3_9 (which causes extra work for the arch teams). Neither solution is perfect, and I'm unsure what to do in this case. What would you do in this situation, delay or revision bump?

juippis:
Ah, indeed, I can see that happening. **sigh**.

PR #20300, opened 2021-04-08
net-im/ejabberd-21.04 bump

Flowdalic:
Besides dependency adjustments this
- Migrates dev-erlang/epam to GLEP-81
- Drops libressl in dev-erlang/fast_tls

GID allocation request for dev-erlang/epam in https://github.com/gentoo/api-gentoo-org/pull/385


dev-erlang/mqtree/mqtree-1.0.13.ebuild
```diff
@@ -0,0 +1,21 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
```
thesamesam, line 4: I mean, if we’re really going for this, how about EAPI 7?

dev-erlang/epam/epam-1.0.10.ebuild
```diff
@@ -0,0 +1,32 @@
...
+"
+RDEPEND="${DEPEND}"
+
+DOCS=( CHANGELOG.md README.md )
+
+src_install() {
+	rebar_src_install
+
+	local epam_path="$(get_erl_libs)/${P}/priv/bin/epam"
+	fowners root:"${PN}" "${epam_path}"
```
thesamesam, line 30:
```suggestion
	fowners root:${PN} "${epam_path}"
```

(It’s not going to contain a space)

dev-erlang/epam/epam-1.0.9.ebuild
```diff
@@ -1,9 +1,9 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=6
 
-inherit rebar user
+inherit rebar
```
thesamesam, line 6: I’d suggest revbumping for these to try get simpler bug reports (you know everybody’s on the same, new GLEP 81 version).

dev-erlang/p1_acme/p1_acme-1.0.12.ebuild
```diff
@@ -0,0 +1,31 @@
...
+	>=dev-erlang/jose-1.9.0
+	>=dev-erlang/base64url-1.0"
+RDEPEND="${DEPEND}"
+
+DOCS=( CHANGELOG.md README.md )
+
+src_prepare() {
+	rebar_src_prepare
+	# otherwise it wants to fetch base64url from git
+	sed -ri 's/\+\+ \[\{base64url.*//' rebar.config.script
```
thesamesam, line 30: What should we do on external commands that might fail?

net-im/ejabberd/ejabberd-21.04.ebuild
```diff
@@ -0,0 +1,292 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+SSL_CERT_MANDATORY=1
+
+inherit eutils pam rebar ssl-cert systemd tmpfiles
```
thesamesam, line 8: Is eutils used here? Need EAPI 6?

net-im/ejabberd/ejabberd-21.04.ebuild
```diff
@@ -0,0 +1,292 @@
...
+	sed -e "/EJABBERD_BIN_PATH=/{s:ejabberd:${P}:}" \
+		-i "${S}/ejabberdctl.template" \
+		|| die 'failed to set ejabberd path in ejabberdctl.template'
+}
+
+# Get epam-wrapper from 'files' directory and correct path to lib directory in
+# it. epam-wrapper is placed into work directory. It is assumed no epam-wrapper
+# file exists there already.
+customize_epam_wrapper() {
+	local epam_wrapper_src="$1"
```
thesamesam, line 98:
```suggestion
	local epam_wrapper_src="${1}”
```

net-im/ejabberd/ejabberd-21.04.ebuild
```diff
@@ -0,0 +1,292 @@
...
+		-e "s|\(LOGS_DIR:=\"\){{localstatedir}}[^\"]*|\1${JABBER_LOG}|" \
+		-e "s|\(SPOOL_DIR:=\"\){{localstatedir}}[^\"]*|\1${JABBER_SPOOL}|" \
+		-i "${S}/ejabberdctl.template" \
+		|| die 'failed to set paths ejabberdctl.template'
+}
+
+# Skip installing docs because it's only COPYING that's installed by Makefile.
+skip_docs() {
+	gawk -i inplace '
+/# Documentation/, /^[[:space:]]*#?[[:space:]]*$/ {
```
thesamesam, line 169: We should probably BDEPEND on gawk if GNU awk is genuinely needed for this.

net-im/ejabberd/ejabberd-21.04.ebuild
```diff
@@ -0,0 +1,292 @@
...
+
+	# Fix bug #591862. ERL_LIBS should point directly to ejabberd directory
+	# rather than its parent which is default. That way ejabberd directory
+	# takes precedence is module lookup.
+	local ejabberd_erl_libs="$(get_ejabberd_path):$(get_erl_libs)"
+	sed -e "s|\(ERL_LIBS=\){{libdir}}.*|\1${ejabberd_erl_libs}|" \
+		-i "${S}/ejabberdctl.template" \
+		|| die 'failed to set ERL_LIBS in ejabberdctl.template'
+
+	sed -e "s|\(AC_INIT(ejabberd, \)m4_esyscmd([^)]*)|\1[$PV]|" \
```
thesamesam, line 203: This is actually not needed right now given the below line is commented

net-im/ejabberd/ejabberd-21.04.ebuild
```diff
@@ -0,0 +1,292 @@
...
+		-i configure.ac || die "Failed to write correct version to configure"
+	# eautoreconf # required in case of download from github
+}
+
+src_configure() {
+	econf \
+		--docdir="${EPREFIX}/usr/share/doc/${PF}/html" \
+		--enable-user=jabber \
+		--disable-system-deps \
```
thesamesam, line 212: This sounds sad. Some context in a comment would be neat.

net-im/ejabberd/ejabberd-21.04.ebuild
```diff
@@ -0,0 +1,292 @@
...
+	default
+
+	keepdir /var/lib/lock/ejabberdctl
+	rm -rf "${ED%/}/var/log" || die
+	rm -rf "${ED%/}/var/spool" || die
+
+	if use pam; then
+		local epam_path="$(get_ejabberd_path)/priv/bin/epam"
+
+		pamd_mimic_system xmpp auth account || die "cannot create pam.d file"
```
thesamesam, line 246: I haven’t checked but does this die for us?

dev-erlang/mqtree/mqtree-1.0.13.ebuild
Flowdalic: Since ejabberd bumps in gentoo tend to be rather large. I tried to keep the changes as minimal as possible. But yes, an EAPI 7 migration would be a sensible thing to do at some point.

net-im/ejabberd/ejabberd-21.04.ebuild
Flowdalic: Good catch, `eutils` does not seem to be used. What do you mean with "Need EAPI 6"?

net-im/ejabberd/ejabberd-21.04.ebuild
Flowdalic: It does. Maybe this is just here to be extra defensive, as this behavior is not documented in pamd_mimic and pamd_mimic_system? Shall I keep the `die`, or remove it?

thesamesam: Almost all eclass functions die for us. They don’t _have_ to, but in reality, they do - things won’t pass review when they’re being added or updated on e.g gentoo-dev otherwise.

So, in theory, we should always check if eclass functions die for us, but in reality, there’s not _really_ a need.

Anyway, back to the point: because we generally expect eclass functions to die for us and such a change to not die would be a big deal and need review of all consumers and discussion on -dev, if it dies now, we can assume it always will. So, let’s drop the extra die.

dev-erlang/mqtree/mqtree-1.0.13.ebuild
thesamesam: I get the point, although my thinking was it’s necessarily invasive anyway and we may as well get the EAPI 7 in unless there’s a reason _not_ to given we’re trying to push for that right now.

but I also understand it’s not your package, so maybe it’s better to do this separately. It just seems like a shame to add a bunch more EAPI 6 in one go if we can help it.

net-im/ejabberd/ejabberd-21.04.ebuild
thesamesam: Always worth auditing eclasses! Especially eutils which is one of those horrible ones which creeps up a lot unnecessarily. multillib (no suffix) is similar as it was in the past needed for get_libdir.

Re EAPI, I was just asking if we couldn’t do 7.

github-k8n:
Any chance of getting this merged?
@Flowdalic, IIRC, reviewers are not notified on force-push, maybe adding a follow-up comment here after the push helps to keep visibility.

thesamesam:
Hanno let us know he was happy to go ahead with this. Sorry for the delay, university's been really busy. Thank you! (See also #20424)

PR #20304, opened 2021-04-08
sys-cluster/glusterfs: Version 9.1

jkroonza:


jkroonza:
Almost straight copy, just add liburing as a dependency.

Behaviour for 9.0 was an implicit automagic dependency, meaning if it was installed, glusterfs would use it, if not installed, it would not be used, now it's a hard dep unless --disable-linux_io-uring is specified.  Dependency looks fairly small, and it is preferred to use it rather than not (not sure why) from what I'm reading, so just hard depend on it.

thesamesam:
My only guess is older kernels or weird arches. Sounds good.

jkroonza:
> My only guess is older kernels or weird arches. Sounds good.

I've read up aboud io_uring, the principle looks insanely awesome, except for possible bugs I can't imagine anyone not wanting to use this.  You're right about kernel version though ... it's only available "since" 5.1 ... but quite frankly, if you're running glusterfs 9 you've probably got an up to date kernel.  If anyone files a bug about it I'll make the support optional.

PR #20305, opened 2021-04-08
dev-java/slf4j-{api,simple}: fix tests 

vaukai:


fordfrog:
@vaukai here's a patch to remove `impl` from the jar:
```
diff --git a/dev-java/slf4j-api/slf4j-api-1.7.30-r1.ebuild b/dev-java/slf4j-api/slf4j-api-1.7.30-r1.ebuild
index 10c48f27b302..a8dcc1d55bf0 100644
--- a/dev-java/slf4j-api/slf4j-api-1.7.30-r1.ebuild
+++ b/dev-java/slf4j-api/slf4j-api-1.7.30-r1.ebuild
@@ -20,6 +20,8 @@ LICENSE="MIT"
 SLOT="0"
 KEYWORDS="~amd64 ~arm64 ~ppc64 ~x86"
 
+BDEPEND="app-arch/zip"
+
 DEPEND="
        >=virtual/jdk-1.8:*
 "
@@ -37,8 +39,6 @@ JAVA_TEST_GENTOO_CLASSPATH="junit-4"
 JAVA_TEST_SRC_DIR="src/test/java"
 
 JAVA_TEST_EXCLUDES=(
-       # This code should have never made it into slf4j-api.jar
-       "org.slf4j.NoBindingTest"       
        # java.lang.InstantiationException
 #      "org.slf4j.helpers.MultithreadedInitializationTest"
 )
@@ -48,3 +48,10 @@ src_prepare() {
        java-pkg_clean
 #      rm -r "${JAVA_SRC_DIR}/org/slf4j/impl" || die
 }
+
+src_compile() {
+       java-pkg-simple_src_compile
+
+       # remove org/slf4j/impl/ from the zip file
+       zip -d ${PN}.jar org/slf4j/impl/\* || die "Failed to remove impl files"
+}
```
wrt that failing test, it still fails, but... after changing in the top level pom.xml java version from 1.5 to 1.8, running `mvn test` in `slf4j-api` dir produces no error... but at the same time i am not able to find the (in our case failing) method `multiThreadedInitialization` in the test output... can you verify that, just to be sure? if so, it would be safe to exclude this test.

vaukai:
> wrt that failing test, it still fails, but... after changing in the top level pom.xml java version from 1.5 to 1.8, running `mvn test` in `slf4j-api` dir produces no error... but at the same time i am not able to find the (in our case failing) method `multiThreadedInitialization` in the test output... can you verify that, just to be sure? if so, it would be safe to exclude this test.

Thanks for the patch.  
The java version, are you saying you changed [this value](https://github.com/qos-ch/slf4j/blob/0b97c416e42a184ff9728877b461c616187c58f7/pom.xml#L36) to `1.8`?  How could I do that in `slf4j-api` where I don't have the top-level pom?

dev-java/slf4j-api/slf4j-api-1.7.30-r1.ebuild
```diff
@@ -0,0 +1,57 @@
...
+src_prepare() {
+	default
+	java-pkg_clean
+#	rm -r "${JAVA_SRC_DIR}/org/slf4j/impl" || die
+}
+
+src_compile() {
+	java-pkg-simple_src_compile
+
+	# remove org/slf4j/impl/ from the zip file
```
fordfrog, line 55: eh, my bad, should be `jar` not `zip` :-) though technically it's a zip file, the extension is `.jar`...

fordfrog:
> The java version, are you saying you changed [this value](https://github.com/qos-ch/slf4j/blob/0b97c416e42a184ff9728877b461c616187c58f7/pom.xml#L36) to `1.8`?

yes

> How could I do that in `slf4j-api` where I don't have the top-level pom?

the top level pom is included from the `slf4j-api` pom so it takes the values from there. it itself does not define java version, for all the packages there it's inherited from the top level pom.

dev-java/slf4j-api/slf4j-api-1.7.30-r1.ebuild
vaukai: adjusted

fordfrog:
@vaukai i'm not sure about this solution, it might not even work. i'm thinking about a bit different solution:

1. lets leave `slf4j-api` as it was, without creating that test jar (it would be created only if tests are enabled i think, which would cause a mess, not even sure it would be installed... anyway, not even feel like i wanna test it)
2. wrt `slf4j-simple`, couldn't we just copy that `MultithreadedInitializationTest.class` (and possibly others if imported by this class) to the test source tree of `slf4j-simple`? wouldn't that solve the issues? none of the other slf4j packages needs it so this would be imo the simplest approach.

what do you think?

vaukai:
Well, after changing top-level `pom.xml`: [build.log.gz](https://github.com/gentoo/gentoo/files/6279043/build.log.gz)

>     2\. wrt `slf4j-simple`, couldn't we just copy that `MultithreadedInitializationTest.class` (and possibly others if imported by this class) to the test source tree of `slf4j-simple`?

Could this be done from inside the `slf4j-simple` ebuild?  
`MultithreadedInitializationTest.class` gets generated in `src_test` phase of the other package.
So how to copy?

fordfrog:
@vaukai i meant to modify `slf4j-simple` ebuild to copy file `slf4j-api/src/test/java/org/slf4j/helpers/MultithreadedInitializationTest.java` in `src_prepare` to `slf4j-simple/src/test/java/org/slf4j/helpers/MultithreadedInitializationTest.java`. not the difference in the first level directory. eclass will then compile even this file and if it does not import other classes from other packages, it might work.

vaukai:
>  not the difference in the first level directory. eclass will then compile even this file and if it does not import other classes from other packages, it might work.

With `MultithreadedInitializationTest.java` added I am now getting:
```
 * Compiling ...
warning: [options] bootstrap class path not set in conjunction with -source 8
src/test/java/org/slf4j/helpers/SimpleLoggerMultithreadedInitializationTest.java:37: error: cannot find symbol
public class SimpleLoggerMultithreadedInitializationTest extends MultithreadedInitializationTest {
                                                                 ^
  symbol: class MultithreadedInitializationTest
src/test/java/org/slf4j/helpers/SimpleLoggerMultithreadedInitializationTest.java:47: error: cannot find symbol
    final String loggerName = this.getClass().getName();
                                  ^
  symbol: method getClass()
src/test/java/org/slf4j/helpers/SimpleLoggerMultithreadedInitializationTest.java:52: error: cannot find symbol
        System.out.println("THREAD_COUNT=" + THREAD_COUNT);
                                             ^
  symbol:   variable THREAD_COUNT
  location: class SimpleLoggerMultithreadedInitializationTest
src/test/java/org/slf4j/helpers/SimpleLoggerMultithreadedInitializationTest.java:65: error: method does not override or implement a method from a supertype
    @Override
    ^
src/test/java/org/slf4j/helpers/SimpleLoggerMultithreadedInitializationTest.java:70: error: method does not override or implement a method from a supertype
    @Override
    ^
Note: src/test/java/org/slf4j/InvocationTest.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
5 errors
1 warning
```

fordfrog:
@vaukai i will have a look at it, proly tomorrow

vaukai:
> @vaukai i will have a look at it, proly tomorrow

Sure.  I'm getting tired of this stuff too.  
The errors are the same as without copying `MultithreadedInitializationTest.java`. The file has always been there, in the same directory.
I should learn how to read and to fix those `error: cannot find symbol`.

vaukai:
@fordfrog 
When all is ok, then after some days for testing it should be added to package list of https://bugs.gentoo.org/776796

fordfrog:
@vaukai so i had a look at the code but did not find whether the test is broken or there is a sychronization issue or there is an issue il slf4j itself:
```
1) multiThreadedInitialization(org.slf4j.helpers.SimpleLoggerMultithreadedInitializationTest)
java.lang.AssertionError: 1284 < 1262+16
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.slf4j.helpers.MultithreadedInitializationTest.multiThreadedInitialization(MultithreadedInitializationTest.java:46)
```
the assertion fails because it expects higher number than it gets. i would just disable this test with a note of the cause so it can be revisited.

i already committed `slf4j-api` with some adjustments, and the metadata aswell, so only `slf4j-simple` is left. you can disable this test and if you're happy with it, i will merge it.

vaukai:
> i already committed `slf4j-api` with some adjustments, and the metadata aswell, so only `slf4j-simple` is left. you can disable this test and if you're happy with it 

Done, happily

fordfrog:
@vaukai thanks! :-)

PR #20308, opened 2021-04-08
eclass/selinux-policy-2: Added MCS/MLS to policy types where unconfined is omitted

jpds:


jpds:
@perfinion This one should also be ready for review.

PR #20309, opened 2021-04-08
net-misc/gofish: Adopt package and drop user.eclass

00-matt:


acct-user/gopher/gopher-0.ebuild
```diff
@@ -0,0 +1,12 @@
+# Copyright 2020-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit acct-user
+
+DESCRIPTION="User for gofish"
+ACCT_USER_ID=295
```
ConiKost, line 9: NACK. Please reserve (fill a PR) a free UID at https://github.com/gentoo/api-gentoo-org/blob/master/files/uid-gid.txt. You can use the script https://github.com/gentoo/api-gentoo-org/blob/master/bin/used_free_uidgids.sh for help.

net-misc/gofish/gofish-1.2-r1.ebuild
```diff
@@ -0,0 +1,32 @@
...
+LICENSE="GPL-2+"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc ~x86"
+
+RDEPEND="
+	acct-group/gopher
+	acct-user/gopher
+"
+
+S=${WORKDIR}/${PN}
```
ConiKost, line 19: Please put `S` after `SRC_URI`. See `skel.ebuild` for example.

net-misc/gofish/gofish-1.2-r1.ebuild
```diff
@@ -0,0 +1,32 @@
...
+
+src_configure() {
+	econf \
```
ConiKost, line 22: Please try to avoid usage of backslashes.

Better like:
```
local myeconfargs=(
   --foo
   --baz
)

econf "${myeconfargs[@]}"
```

net-misc/gofish/gofish-1.2-r1.ebuild
```diff
@@ -0,0 +1,32 @@
...
+		--localstatedir=/var \
+		--disable-mmap-cache
+}
+
+src_install() {
+	default
+	dodoc Configure_GoFish
```
ConiKost, line 29: Maybe better define `DOCS=` globally?

net-misc/gofish/gofish-1.2.ebuild
```diff
@@ -9,7 +9,7 @@ DESCRIPTION="Gofish gopher server"
 HOMEPAGE="http://gofish.sourceforge.net"
 SRC_URI="mirror://sourceforge/gofish/${P}.tar.gz"
 
-LICENSE="GPL-2"
+LICENSE="GPL-2+"
 SLOT="0"
 KEYWORDS="amd64 ~ppc ~x86"
 IUSE=""
```
ConiKost, line 15: If empty, drop it.

net-misc/gofish/gofish-1.2.ebuild
```diff
@@ -9,7 +9,7 @@ DESCRIPTION="Gofish gopher server"
 HOMEPAGE="http://gofish.sourceforge.net"
```
ConiKost, line 9: Could you check, if `https` is supported?

ConiKost (changes requested):

acct-user/gopher/gopher-0.ebuild
00-matt: Thanks for the review. Sorry, I did run that script but I forgot to actually reserve what it suggested.

net-misc/gofish/gofish-1.2.ebuild
00-matt: It is not :(

net-misc/gofish/gofish-1.2-r1.ebuild
```diff
@@ -0,0 +1,35 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DESCRIPTION="Gofish gopher server"
+HOMEPAGE="http://gofish.sourceforge.net"
+SRC_URI="mirror://sourceforge/gofish/${P}.tar.gz"
+S=${WORKDIR}/${PN}
```
ConiKost, line 9: Please quote.

ConiKost:
I just have it a try, please update the init scripts.

```
 * Use of the opts variable is deprecated and will be
 * removed in the future.
 * Please use extra_commands, extra_started_commands or extra_stopped_commands.
 * Starting gofish  ...
 * start-stop-daemon: fopen `/var/run/gofish.pid': No such file or directory
 * Detaching to start `/usr/sbin/gofish' ...                                          
```

00-matt:
Sorry, forgot to add new init script.

ConiKost (approved):

PR #20316, opened 2021-04-09
dev-tex/biber: disable 64-bit only tests on 32-bit arches

j-licht:
Closes: https://bugs.gentoo.org/780879
Signed-off-by: Jonas Licht <jonas.licht@fem.tu-ilmenau.de>

### DIFF
--- a/dev-tex/biber/biber-2.16.ebuild
+++ b/dev-tex/biber/biber-2.16.ebuild
@@ -67,3 +67,10 @@ DEPEND="${RDEPEND}
 			dev-perl/Test-Differences )"
 
 mydoc="doc/biber.tex"
+
+src_prepare() {
+	#disable 64-bit only Tests on non 64-bit archs
+	use amd64 || use arm64 || eapply "${FILESDIR}/${P}-disable-64bit-only-tests.patch"
+
+	default
+}
--- a/dev-tex/biber/files/biber-2.16-disable-64bit-only-tests.patch
+++ b/dev-tex/biber/files/biber-2.16-disable-64bit-only-tests.patch
@@ -1,0 +1,45 @@
+Disable Tests which works only on 64bit architecture
+because they tests int overflows.
+From: Jonas Licht <jonas.licht@gmail.com>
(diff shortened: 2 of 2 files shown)


dev-tex/biber/biber-2.16.ebuild
```diff
@@ -67,3 +67,10 @@ DEPEND="${RDEPEND}
 			dev-perl/Test-Differences )"
 
 mydoc="doc/biber.tex"
+
+src_prepare() {
+	#disable 64-bit only Tests on non 64-bit archs
+	use amd64 || use arm64 || eapply "${FILESDIR}/${P}-disable-64bit-only-tests.patch"
```
thesamesam, line 73: I wonder if there's a cleaner way to do this (detect it), but hey, works for now.

dev-tex/biber/files/biber-2.16-disable-64bit-only-tests.patch
```diff
@@ -0,0 +1,45 @@
...
+ 
+@@ -880,14 +880,14 @@ $biber->prepare;
+ $out = $biber->get_output_obj;
+ 
+ # Test negative dates and eras
+-eq_or_diff($out->get_output_entry('era1', $main), $era1, 'Date meta information - 1');
+-eq_or_diff($out->get_output_entry('era2', $main), $era2, 'Date meta information - 2');
+-eq_or_diff($out->get_output_entry('era3', $main), $era3, 'Date meta information - 3');
+-eq_or_diff($out->get_output_entry('era4', $main), $era4, 'Date meta information - 4');
++#eq_or_diff($out->get_output_entry('era1', $main), $era1, 'Date meta information - 1');
```
thesamesam, line 22: We usually don't commit patches with comments, we just remove the lines. The patch shows us which lines have been removed so information is lost. I need to update [this](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#clean-patch-howto) accordingly

thesamesam:
Thanks! (And thanks for responding to feedback so quickly).

PR #20317, opened 2021-04-09
sci-libs/cantera: 2.5.1 fix USE="cti -python" test and installation

band-a-prend:
USE="cti -python" provides python_minimal Cantera installation but thisinterface had no appropriate patch to install
in proper site_packages directory in Gentoo. **Thus the installation in this case was broken.**
So there is no rev bump this time.

Also 'dev-python/ruamel-yaml' is required as RDEPEND in this case and as DEPEND for test phase.

### DIFF
--- a/sci-libs/cantera/files/cantera-2.5.1_env.patch
+++ b/sci-libs/cantera/files/cantera-2.5.1_env.patch
@@ -45,6 +45,50 @@ diff -Naur old/interfaces/cython/SConscript new/interfaces/cython/SConscript
      else:
          extra = '--user'
          localenv.AppendENVPath(
+diff -Naur old/interfaces/python_minimal/SConscript new/interfaces/python_minimal/SConscript
+--- old/interfaces/python_minimal/SConscript	2021-02-13 00:57:15.000000000 +0300
++++ new/interfaces/python_minimal/SConscript	2021-04-09 23:26:28.000000000 +0300
+@@ -8,7 +8,7 @@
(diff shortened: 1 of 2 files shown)


sci-libs/cantera/files/cantera-2.5.1_env.patch
```diff
@@ -45,6 +45,50 @@ diff -Naur old/interfaces/cython/SConscript new/interfaces/cython/SConscript
      else:
          extra = '--user'
          localenv.AppendENVPath(
+diff -Naur old/interfaces/python_minimal/SConscript new/interfaces/python_minimal/SConscript
```
thesamesam, line 48: Try[ clean up](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#clean-patch-howto) patches if you can by removing the diff lines, using a/ and b/, and removing dates.

thesamesam: (I'll fix this myself now to get it merged.)

band-a-prend: I sent non-Gentoo specific parts of patch to upstream too.

thesamesam:
Thank you!

sci-libs/cantera/files/cantera-2.5.1_env.patch
thesamesam: Excellent as ever. :)

PR #20318, opened 2021-04-09
sys-fs/zfs: adapt live template

thesamesam:


gyakovlev:
`awk` used at runtime (in some scripts,  dracut module and completions), and `DEPEND` gets included  into `RDEPEND`, so let's keep it in `DEPEND` as well or put it in `RDEPEND` in addition to `BDEPEND`

gyakovlev (changes requested):
drop/fix awk change and feel free to merge.

PR #20324, opened 2021-04-10
package.mask: Last rite app-arch/unmakeself

00-matt:
Bug: https://bugs.gentoo.org/781947


00-matt:
Rebased and amended date

PR #20328, opened 2021-04-10
dev-util/samurai: Install docs

orbea:
Signed-off-by: orbea <orbea@riseup.net>

### DIFF
--- a/dev-util/samurai/samurai-1.2.ebuild
+++ b/dev-util/samurai/samurai-1.2.ebuild
@@ -18,4 +18,5 @@ SLOT="0"
 
 src_install() {
 	emake DESTDIR="${D}" PREFIX="${EPREFIX}"/usr install
+	dodoc LICENSE README.md
 }
--- a/dev-util/samurai/samurai-9999.ebuild
+++ b/dev-util/samurai/samurai-9999.ebuild
@@ -18,4 +18,5 @@ SLOT="0"
 
 src_install() {
 	emake DESTDIR="${D}" PREFIX="${EPREFIX}"/usr install
+	dodoc LICENSE README.md
 }


dev-util/samurai/samurai-1.2.ebuild
```diff
@@ -18,4 +18,5 @@ SLOT="0"
 
 src_install() {
 	emake DESTDIR="${D}" PREFIX="${EPREFIX}"/usr install
+	dodoc LICENSE README.md
```
Zlogene, line 21: ypu should never install the license file

Zlogene (changes requested):

dev-util/samurai/samurai-1.2.ebuild
orbea: Is there some gentoo policy against this? Seems weird, having the license file for reference is useful especially when a reliable network may not be available.

orbea: On second thought not providing the `LICENSE` file would be a clear violation of all four licenses...

__ISC__
> Permission to use, copy, modify, and/or distribute this software for any purpose
with or without fee is hereby granted, provided that the above copyright notice
and this permission notice appear in all copies.

__Apache2__
> You must give any other recipients of the Work or
          Derivative Works a copy of this License; and

__MIT__ (x2)
> The above copyright notice and this permission notice shall be included in
all copies or substantial portions of the Software.

The `LICENSE` file **must** be included and all binary samurai packages could not be redistributed without it.

ionenwks: LICENSE files are already available (gentoo has them), for the most parts it's a duplicate.

orbea: Where? Its not part of the samurai package by default where it is required.

thesamesam: Okay, so the initial explanation was a bit blunt.

- Licenses are stored in /var/db/repos/gentoo/licenses/
- We don't install them _in_ the ebuild to avoid duplicates
- The LICENSE value in ebuilds must be accurate and for that to happen, we need to import the licences into the tree (in the licenses/) directory, so they're guaranteed to be on user systems too

juippis:
Sorry, the GCO requirement still applies. Please rebase with a proper sign-off line, or please stop wasting everyone's time with these.
https://github.com/gentoo/gentoo/pull/18300#issuecomment-729736581
https://github.com/gentoo/gentoo/pull/18307#issuecomment-729332847

PR #20334, opened 2021-04-11
net-misc/bird: Adding LTO useflag

alarig:
Closes: https://bugs.gentoo.org/781923

### DIFF
--- a/net-misc/bird/bird-2.0.8.ebuild
+++ b/net-misc/bird/bird-2.0.8.ebuild
@@ -12,7 +12,7 @@ LICENSE="GPL-2"
 
 SLOT="0"
 KEYWORDS="~amd64 ~arm64 ~x86 ~x64-macos"
-IUSE="+client debug libssh"
+IUSE="+client debug libssh +lto"
 
 RDEPEND="
 	client? ( sys-libs/ncurses:= )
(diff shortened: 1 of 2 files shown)


net-misc/bird/bird-2.0.8.ebuild
```diff
@@ -12,7 +12,7 @@ LICENSE="GPL-2"
 
 SLOT="0"
 KEYWORDS="~amd64 ~arm64 ~x86 ~x64-macos"
-IUSE="+client debug libssh"
+IUSE="+client debug libssh +lto"
```
SoapGentoo, line 15: why are you enabling this by default?

SoapGentoo (changes requested):

net-misc/bird/bird-2.0.8.ebuild
alarig: Because of https://bugs.gentoo.org/781923#c13

SoapGentoo: won't be merged in this state then, sorry

ionenwks: Personally felt the USE wasn't necessary considering this doesn't seem to need a special path/setup for lto to work, users can still add it to their normal FLAGS if they want it (the way want it without the package appending its own override flag at the end).

thesamesam: Right, TL:DR: let's not bother with the flag, and disregard upstream here

PR #20336, opened 2021-04-11
app-emulation/img: bump to 0.5.11

Z5483:
Signed-off-by: Khue Nguyen <Z5483Y@gmail.com>


app-emulation/img/img-0.5.11.ebuild
```diff
@@ -0,0 +1,34 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+EGO_PN="github.com/genuinetools/img"
+
+inherit golang-vcs-snapshot
```
juippis, line 8: Looking at upstream, you should be using the *go-module.eclass* with this package.
https://devmanual.gentoo.org/eclass-reference/go-module.eclass/index.html

Z5483: I will try to fix this as soon as possible. Thank you for the review :)

juippis (approved):
Well the ebuild became a lot uglier, but we are trying to phase away from using old *golang-\** eclasses. Thanks!

PR #20339, opened 2021-04-11
games-strategy/warzone2100: version bump to 4.0.1

0xC0ncord:
4.0.0 also dropped all Qt dependencies, which is reflected in this ebuild.

Closes: https://bugs.gentoo.org/777330
Signed-off-by: Kenton Groombridge <me@concord.sh>


thesamesam:
Can you do .1? I think they may have yanked the distfile

0xC0ncord:
> Can you do .1? I think they may have yanked the distfile

Done.

PR #20346, opened 2021-04-12
dev-util/cutter: improve configuration

arkamar:
- Introduce `graphviz` use flag
- Set `CUTTER_USE_ADDITIONAL_RIZIN_PATHS=OFF`; This should be disabled if rizin is installed as a proper system library according to an option description, otherwise it may search it in /usr/local, for example.

### DIFF
--- a/dev-util/cutter/cutter-2.0.1.ebuild
+++ b/dev-util/cutter/cutter-2.0.1.ebuild
@@ -16,6 +16,7 @@ SRC_URI="https://github.com/rizinorg/cutter/archive/v${PV}.tar.gz -> ${P}.tar.gz
 LICENSE="CC-BY-SA-3.0 GPL-3"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
+IUSE="graphviz"
 REQUIRED_USE="${PYTHON_REQUIRED_USE}"
 
 COMMON_DEPEND="${PYTHON_DEPS}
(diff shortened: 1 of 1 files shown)


dev-util/cutter/cutter-2.0.1.ebuild
```diff
@@ -17,14 +17,16 @@ LICENSE="CC-BY-SA-3.0 GPL-3"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
 REQUIRED_USE="${PYTHON_REQUIRED_USE}"
```
ajakk, line 19: IUSE before REQUIRED_USE

dev-util/cutter/cutter-2.0.1.ebuild
```diff
@@ -42,7 +44,8 @@ src_configure() {
 		-DCMAKE_CXX_COMPILER="$(tc-getCXX)"
 		-DCMAKE_C_COMPILER="$(tc-getCC)"
 		-DCMAKE_INSTALL_PREFIX="${EPREFIX}/usr"
-		-DCUTTER_ENABLE_GRAPHVIZ=OFF
+		-DCUTTER_USE_ADDITIONAL_RIZIN_PATHS=OFF
```
ajakk, line 47: Keep the options sorted alphabetically, please

dev-util/cutter/cutter-2.0.1.ebuild
```diff
@@ -42,7 +44,8 @@ src_configure() {
...
+		-DCUTTER_ENABLE_GRAPHVIZ=$(usex graphviz)
```
ajakk, line 48: Quote the subshell expression to follow suit with the other options that call a subshell

ajakk (changes requested):

dev-util/cutter/cutter-2.0.1.ebuild
arkamar: done

dev-util/cutter/cutter-2.0.1.ebuild
arkamar: ok

dev-util/cutter/cutter-2.0.1.ebuild
arkamar: fixed

ajakk:
Thanks!

PR #20354, opened 2021-04-12
sys-cluster/ipvsadm: bump to v1.31

rikatz:
Bump ipvsadm to v1.31 so it can add support to new features, as SCTP support


ultrabug:
Thank you @rikatz

PR #20358, opened 2021-04-13
Update elisp*.eclass, drop support for EAPI 4 and 5

ulm:


ulm:
Please review the eclass changes only.

The removals in app-emacs are only to pacify CI and will be handled individually (and require some stabilisations first).

eclass/elisp-common.eclass
```diff
@@ -165,7 +166,7 @@
 # to above calls of elisp-site-regen().
 
 case ${EAPI:-0} in
-	4|5|6) inherit eapi7-ver ;;
+	5|6) inherit eapi7-ver ;;
 	7) ;;
 	*) die "${ECLASS}: EAPI ${EAPI:-0} not supported" ;;
 esac
```
SoapGentoo, line 172: no inherit guard?

eclass/elisp-common.eclass
```diff
@@ -228,7 +231,7 @@ elisp-emacs-version() {
 		# Redirecting stdin and unsetting TERM and DISPLAY will cause
 		# most of them to exit with an error.
 		unset TERM DISPLAY
-		${EMACS} ${EMACSFLAGS} -l "${T}"/emacs-version.el </dev/null
+		${tmout} ${EMACS} ${EMACSFLAGS} -l "${T}"/emacs-version.el </dev/null
 	)
 	ret=$?
 	rm -f "${T}"/emacs-version.el
```
SoapGentoo, line 237: maybe `|| die`? I see there's handling for the error case, but purely for consistency

eclass/elisp-common.eclass
```diff
@@ -380,7 +362,7 @@ elisp-modules-install() {
 	shift
 	# Don't bother inheriting multilib.eclass for get_libdir(), but
 	# error out in old EAPIs that don't support it natively.
-	[[ ${EAPI} == [45] ]] \
+	[[ ${EAPI} == 5 ]] \
 		&& die "${ECLASS}: Dynamic modules not supported in EAPI ${EAPI}"
 	ebegin "Installing dynamic modules for GNU Emacs support"
 	( # subshell to avoid pollution of calling environment
```
SoapGentoo, line 368: (more a comment): isn't this a bit brittle, seeing that only EAPI>=7 handles errors in subshells correctly?

eclass/elisp.eclass
```diff
@@ -73,8 +73,7 @@ EXPORT_FUNCTIONS src_{unpack,prepare,configure,compile,install} \
 
 RDEPEND=">=app-editors/emacs-${NEED_EMACS}:*"
```
SoapGentoo, line 74: inherit guard here too?

eclass/elisp.eclass
```diff
@@ -118,17 +117,11 @@ elisp_src_prepare() {
 		else
 			die "Cannot find ${patch}"
```
SoapGentoo, line 118: `ELISP_PATCHES` looks like a whitespace-separated list? w/e, seems a bit out of line with modern practices

SoapGentoo (approved):
LGTM, if you could just add the inherit guards

eclass/elisp-common.eclass
ulm: Error handling is just below in the `if` block (checking `$?`). I'd rather not move any nontrivial code into `$( )`.

Also, `$( )` spawns a subshell, and `die` won't work there in EAPI 6.

eclass/elisp-common.eclass
ulm: Hm, unless I'm missing something, this test happens before the subshell is spawned.

eclass/elisp.eclass
ulm: You're right, and its docstring says "This variable is semi-deprecated, preferably use the PATCHES array instead if the EAPI supports it."

Maybe we can remove `ELISP_PATCHES` in EAPI 8.

eclass/elisp-common.eclass
ulm: These are "leaf" eclasses that aren't (passively) inherited by any other eclass. And in EAPI 7 they also don't (actively) inherit anything else. So no danger of cycles, and I think we can save these variables from the environment.

eclass/elisp.eclass
ulm: In fact, we can drop "if the EAPI supports it" from the comment if we're only supporting EAPI 6 and later.

PR #20359, opened 2021-04-13
dev-libs/libthreadar: bump to v1.3.2

ceamac:
Minor bugfix release


thesamesam:
Note that we should stabilise a newer version of dar soon. If we clean up the old version here, we're kind of delaying that. Should I drop the cleanup?

ceamac:
Sure, if it helps with new stable version let's keep two versions

thesamesam:
Thanks for the quick reply!

ceamac:
Thank you for the quick review :)

thesamesam:
https://github.com/gentoo/gentoo/commit/ec743ec9bd3fb29375403ba2392516e33d0842cf

PR #20362, opened 2021-04-13
dev-python/spyder: add python3.9

AndrewAmmerlaan:
Add python3.9 to the remaining deps of spyder and spyder itself.


juippis (approved):

AndrewAmmerlaan:
Thanks @juippis

PR #20365, opened 2021-04-13
www-apps/kibana-bin: restrict nodejs version

hydrapolic:
Closes: https://bugs.gentoo.org/782376


juippis (approved):
Wonder if a simple sed/patch to modify the restriction actually works... I hope upstream is aware and working on this.

hydrapolic:
That probably won't be that easy, just by checking the error message the user reported:

````
Error: The module '/opt/kibana/node_modules/re2/build/Release/re2.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 83. This version of Node.js requires
NODE_MODULE_VERSION 88. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
````

Thanks for the review/merge @juippis

PR #20366, opened 2021-04-13
dev-java/odfdom: Do not depend on dev-java/xml-commons-external

vaukai:

Compiling without `dev-java/xml-commons-external` produces errors: [build.log.gz](https://github.com/gentoo/gentoo/files/6304193/build.log.gz)

### DIFF
--- a/dev-java/odfdom/odfdom-0.8.6-r2.ebuild
+++ b/dev-java/odfdom/odfdom-0.8.6-r2.ebuild
@@ -14,8 +14,7 @@ LICENSE="Apache-2.0"
 SLOT="0"
 KEYWORDS="~amd64 ~x86-linux ~sparc-solaris ~x86-solaris"
 
-CDEPEND="dev-java/xerces:2
-	dev-java/xml-commons-external:1.4"
+CDEPEND="dev-java/xerces:2"
 
 BDEPEND="app-arch/unzip"
 RDEPEND=">=virtual/jre-1.5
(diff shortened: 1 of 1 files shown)


fordfrog:
@vaukai looks good, but it needs to be a revbump, because deps are written to `/usr/share/<package>/package.env` and this way the deps would not be updated which might cause issues later on runtime when `xml-commons-external` is removed but still present in `package.env`.

vaukai:
> @vaukai looks good,

Not at all. Sorry, I forgot to mention.
Going without `xml-commons-external` makes trouble:  [build.log.gz](https://github.com/gentoo/gentoo/files/6304316/build.log.gz)
But I'll move it to a revbump so we can proceed

fordfrog:
> > @vaukai looks good,
> 
> Not at all. Sorry, I forgot to mention.
> Going without `xml-commons-external` makes trouble: [build.log.gz](https://github.com/gentoo/gentoo/files/6304316/build.log.gz)
> But I'll move it to a revbump so we can proceed

i just did a visual check but did not do the testing because of the revbump. thanks :-)

fordfrog:
@vaukai i just checked what packages depend on this one, and found none. and the package does not provide launcher, so imo should be last-rited as obsolete? any objections? :-)

vaukai:
>  so imo should be last-rited as obsolete? any objections? :-)

Reading the [homepage](https://odftoolkit.org/odfdom/) it appears as something useful.  
But last-riting seems better tham carrying broken stuff

vaukai:
@fordfrog 
Another candidate to be last-rited, how do you think about `dev-java/saaj-1.3.19-r2`?

fordfrog:
@vaukai 

> > so imo should be last-rited as obsolete? any objections? :-)
> 
> Reading the [homepage](https://odftoolkit.org/odfdom/) it appears as something useful.
> But last-riting seems better than carrying broken stuff

yes, but it's a library for which we don't have any consumer. we might re-add it once something needs it.

> Another candidate to be last-rited, how do you think about `dev-java/saaj-1.3.19-r2`?

yes, again a library without a consumer :-)

i will last-rite them both tomorrow. if you find more, just push it my way :-)

fordfrog:
@vaukai both masked for removal

PR #20368, opened 2021-04-13
app-shells/bash-completion: add missing user-info eclass

mm1ke:

Hi,

The ebuild calls `egethome` but doesn't inherit the `user-info` eclass. This PR simply fixes this issue.

### DIFF
--- a/app-shells/bash-completion/bash-completion-9999.ebuild
+++ b/app-shells/bash-completion/bash-completion-9999.ebuild
@@ -1,10 +1,10 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
 PYTHON_COMPAT=( python3_{7..9} )
-inherit autotools git-r3 python-any-r1
+inherit autotools git-r3 python-any-r1 user-info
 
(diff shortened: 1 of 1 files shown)


thesamesam:
BTW, if you fancy helping me out (no obligation!):

1) pmaint regen (needed for improved accuracy)
2) pkgcheck scan -k UnusedInherits or pkgcheck scan -k IndirectInherits or pkgcheck scan -k MissingInherits

There are accuracy issues with Python and some other eclasses but the eutils stuff is accurate, just watch out in case we really need to replace it with epatch instead. Unfortunately, eutils->epatch introduces a lot of CI warnings so we'll have to tweak the threshold, but it's worth it because it is actually fixing issues (we can coordinate the needed tweak) - so in some cases, it's better where it's easy to port to EAPI 7 instead.

mgorny:
Thanks!

mm1ke:
@thesamesam 
Thanks for the hints. :)
In order to find unused/missing inherits i'm actually using following lists:
https://gentooqa.levelnine.at/results/gentoo/checks/ebuild_missing_eclasses/
https://gentooqa.levelnine.at/results/gentoo/checks/ebuild_unused_eclasses/
They are also not perfect but it's my main source for my PRs. 
...Some day i'll include pkgcheck results too, then i don't need to run my own scripts anymore ;)

PR #20373, opened 2021-04-14
media-libs/libpng{,-compat}: misc fixes (mostly CPU_FLAGS_ARM_NEON)

thesamesam:


stefson:
@thesamesam you mind to fix that note in the package.use.mask? It is rather incomprehensible as of now :-)

PR #20383, opened 2021-04-14
dev-java/bytelist: bump to 1.0.15

vaukai:


vaukai:
`jcodings` makes trouble:
```
Appending /var/db/repos/vaukai to PORTDIR_OVERLAY...
 * jcodings-1.0.55-sources.tar.gz BLAKE2B SHA512 size ;-) ...                                                                                       [ ok ]
 * Using: openjdk-11
>>> Unpacking source...
>>> Unpacking jcodings-1.0.55-sources.tar.gz to /var/tmp/portage/dev-java/jcodings-1.0.55/work
>>> Source unpacked in /var/tmp/portage/dev-java/jcodings-1.0.55/work
>>> Preparing source in /var/tmp/portage/dev-java/jcodings-1.0.55/work/jcodings-jcodings-1.0.55 ...
Searching for bundled jars:
None found.
Searching for bundled classes (no output if none found):
Search done.
>>> Source prepared.
>>> Configuring source in /var/tmp/portage/dev-java/jcodings-1.0.55/work/jcodings-jcodings-1.0.55 ...
>>> Source configured.
>>> Compiling source in /var/tmp/portage/dev-java/jcodings-1.0.55/work/jcodings-jcodings-1.0.55 ...
 * Verbose logging for "ejavac" function
 * Compiler executable: javac
 * Extra arguments: -source 1.8 -target 1.8
 * Complete command:
 * javac -source 1.8 -target 1.8 -d target/classes -encoding UTF-8 @sources.lst
 * Compiling ...
warning: [options] bootstrap class path not set in conjunction with -source 8
src/module-info.java:1: error: modules are not supported in -source 8
open module org.jruby.jcodings {
     ^
  (use -source 9 or higher to enable modules)
1 error
1 warning
```

dev-java/jcodings/jcodings-1.0.55.ebuild
```diff
@@ -0,0 +1,37 @@
...
+LICENSE="MIT"
+SLOT="1"
+KEYWORDS="~amd64 ~x86"
+
+DEPEND="
+	>=virtual/jdk-11:*
+"
+
+RDEPEND="
+	>=virtual/jre-11:*
```
fordfrog, line 28: @vaukai this is something new to introduce to you. this package does not compile with source 1.8 because it uses modules and those were first introduced in java 9. but still, the library can be successfully compiled for java 1.8 (that is in bytecode for java 1.8). in other words, the library needs java 11 for compilation but it can run using java 1.8, so the apps that use it are not restricted only to java 11 and newer. so `DEPEND` will stay as it is, but `RDEPEND` should be lowered to `>=virtual/jre-1.8:*`.

vaukai: Understand.  Will you please adjust it for `jcodings` and `joni`, so I need not touch them again, please

fordfrog:
@vaukai i just updated existing `dev-java/jcodings` for java 1.8, see 71ac53b92024b01dac8885042c6b0fe2a8ba1560.

fordfrog:
@vaukai `bytelist` commit merged at df5c82ffcda9d8d103ef8d0d7ae3c9628cde8156

fordfrog:
@vaukai merged `jvyamlb` in 033e436e1a3d433a4d8c5aed4047b6a9a1592316 + added a8f9cfe25a296bb27d40746b3486500fb26d5738 and b36e1fbb13b5178d3f0a8bcbad10493bf0dda0ef

i updated the dep slot as it compiles fine, so we could get rid of the old version. and i kept stable as this is not a bump but a simple change in bytecode version.

dev-java/jcodings/jcodings-1.0.55.ebuild
fordfrog: @vaukai anyway, it does not work in this case as `joni` wants to import module from `jcodings` and if compiled with 1.8 bytecode, it does not contain module information and hence compilation of `joni` fails. so java 11 is correct in this case.

fordfrog:
@vaukai thanks! on the masking commit i changed the sign to `>=`.

PR #20384, opened 2021-04-14
eutils.eclass: mark as @DEPRECATED officially (test)

thesamesam:
Note that this is already deprecated as per @DESCRIPTION. This
is just making it official and giving us the pkgcheck warnings
for new such ebuilds.


thesamesam:
1. A random sample of these (and from looking at the list), it looks like all of these are direct.
2. We should merge this with some cherry-picked epatch stuff from #20390 to avoid having to reset the violation counter twice.

I really think that if we can, we should get this in finally, because otherwise people are going to keep committing eutils ebuilds because it's "technically not deprecated" despite the description stating it.

PR #20389, opened 2021-04-15
sys-boot/grub: Remove carat separator from multiple initrd paths

Peter-Levine:
I co-maintain _sys-boot/os-prober_.  In its current state, grub-mkconfig/os-prober truncates all but the first path passed to `initrd` for arch-based distros.  This leads to a freeze when dual-booting Manjaro without using the fallback entry.  Patching os-prober alone can get it to emit multiple space-separated `initrd` paths to grub.  The way grub handles certain space-separated strings such as kernel parameters is to replace the spaces with `^`, and later to switch the carats back to spaces.  Unfortunately, it fails to do this with `initrd` paths, leading to entries such as `initrd /boot/intel-ucode.img^/boot/amd-ucode.img^/boot/initramfs-5.11-x86_64.img` (refer to https://bugzilla.suse.com/1101942, https://bugs.debian.org/820838, and https://savannah.gnu.org/bugs/index.php?47681)

This patch is a simple, one-line fix whose implementation  is used in [openSUSE](https://build.opensuse.org/package/view_file/Base:System/grub2/grub2-util-30_os-prober-multiple-initrd.patch?expand=1), [Arch](https://github.com/archlinux/svntogit-packages/blob/ccf0e81307b605ca0d9659ff73f1a1cd04bb8e64/trunk/0002-10_linux-detect-archlinux-initramfs.patch#L22), and [Manjaro](https://github.com/manjaro/packages-core/blob/d34a3310179b50bdf8364046ac49e1b834dfeec0/grub/grub-manjaro-modifications.patch#L142).

Signed-off-by: Peter Levine <plevine457@gmail.com>

### DIFF
--- a/sys-boot/grub/files/grub-2.06_rc1-multiple-initrd-paths.patch
+++ b/sys-boot/grub/files/grub-2.06_rc1-multiple-initrd-paths.patch
@@ -1,0 +1,22 @@
+When parsing output from os-prober, grub replaces spaces with "^" and, for some
+variables such as LPARAMS, replaces "^" again with " ".  It doesn't do this
+with initrd.  Since Manjaro passes multiple paths to initrd, grub emits something
+like "initrd /boot/intel-ucode.img^/boot/amd-ucode.img^/boot/initramfs-5.11".
+Dual-booting Manjaro thus fails. This patch enables spaces in the initrd line to avoid
+this issue.
+
+Bug: https://bugzilla.suse.com/1101942
+     https://bugs.debian.org/820838
+     https://savannah.gnu.org/bugs/index.php?47681
+
+--- a/util/grub.d/30_os-prober.in
++++ b/util/grub.d/30_os-prober.in
+@@ -207,7 +207,7 @@ EOF
+         LBOOT="`echo ${LINUX} | cut -d ':' -f 2`"
+         LLABEL="`echo ${LINUX} | cut -d ':' -f 3 | tr '^' ' '`"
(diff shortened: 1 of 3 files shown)


floppym:
It seems like this patch should be applied upstream instead. Please submit it there.

floppym:
Note that the grub bug tracker is usually ignored; patches should be submitted to the grub-devel mailing list.

https://lists.gnu.org/mailman/listinfo/grub-devel

Peter-Levine:
OK.  Found the [corresponding mailing-list post](https://lists.gnu.org/archive/html/bug-grub/2016-04/msg00014.html).  It appears they're going to wait for os-prober to accept their patch first, which should only leave the rest of eternity. 

Closing.

floppym:
That's actually the same bug report you linked earlier, just in mailing list form.

Again, a patch sent to the grub-devel mailing list (not bug-grub) has a much better chance of getting merged.

Peter-Levine:
I see.  Thanks.  I suppose it likely should be [fixed in os-prober](https://salsa.debian.org/installer-team/os-prober/-/merge_requests/4?commit_id=1745d1528cebe8e90796c6891af7724bd5d5ed09) first since **grub.d/30_os-prober** is dependent on its behaviour, but I'll eventually post on the grub mailing.

Peter-Levine:
Finally fixed upstream with https://git.savannah.gnu.org/cgit/grub.git/commit/?id=000b5cd04fd228f9741f5dca0491636bc0b89eb8.

PR #20404, opened 2021-04-16
mail-mta/exim: native srs support.

jkroonza:
Version 9.94 includes EXPERIMENTAL_SRS_NATIVE, which has already been
marked as supported for the next release.

Since users may already have the libsrs_alt configured, split the USE
flags into srs-alt and srs-native to distinguish between the two
implementations.

These options have changed for the 4.95 release to EXPERIMENTAL_SRS_ALT
and SUPPORT_SRS, leave comments to this effect in the ebuild so that
this isn't forgotten on the bump.

Closes: https://bugs.gentoo.org/783099

### DIFF
--- a/mail-mta/exim/exim-4.94-r3.ebuild
+++ b/mail-mta/exim/exim-4.94-r3.ebuild
@@ -5,7 +5,7 @@ EAPI="7"
 
 inherit db-use toolchain-funcs multilib pam systemd
 
-IUSE="arc +dane dcc +dkim dlfunc dmarc +dnsdb doc dovecot-sasl dsn elibc_glibc exiscan-acl gnutls idn ipv6 ldap libressl lmtp maildir mbx mysql nis pam perl pkcs11 postgres +prdr proxy radius redis sasl selinux spf sqlite srs +ssl syslog tcpd +tpda X"
+IUSE="arc +dane dcc +dkim dlfunc dmarc +dnsdb doc dovecot-sasl dsn elibc_glibc exiscan-acl gnutls idn ipv6 ldap libressl lmtp maildir mbx mysql nis pam perl pkcs11 postgres +prdr proxy radius redis sasl selinux spf sqlite srs-alt srs-native +ssl syslog tcpd +tpda X"
 REQUIRED_USE="
 	arc? ( dkim spf )
 	dane? ( ssl !gnutls )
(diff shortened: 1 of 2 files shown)


grobian:
I've pushed your work with a few slight changes to it.  Thanks!

jkroonza:
> I've pushed your work with a few slight changes to it. Thanks!

Thank you.  Few notes:

+       srs? (
+               exiscan-acl
+               ^^ ( srs-alt srs-native )
+       )

Not sure srs-native requires exiscan-acl.

+	srs? ( srs-alt? ( mail-filter/libsrs_alt ) )

I'd personally drop the wrapping srs? USE flag, unles you're going to add REQUIRED_USE srs-alt? ( srs) entries.

Similar concerns on the hunk @ lines ~440.

In short: whilst I'd keep the SRS USE flag for the moment, I'd only actually use it in REQUIRED_USE.  The two implementations are independent and can actually co-exist.  So you can use REQUIRED_USE srs ? ( || ( srs-alt srs-native ) ).

You're aiming to just drop back to USE=srs?  And that be the single implementation?  Switching between the two isn't exactly trivial, so I'd keep the split for as long as it's available from upstream.

Perfectly usable as it stands though.  So won't be filing bugs :).

grobian:
> > I've pushed your work with a few slight changes to it. Thanks!
> 
> Thank you. Few notes:
> 
>     * ```
>         srs? (
>       ```
> 
>     * ```
>                 exiscan-acl
>       ```
> 
>     * ```
>                 ^^ ( srs-alt srs-native )
>       ```
> 
>     * ```
>         )
>       ```
> 
> 
> Not sure srs-native requires exiscan-acl.

Me neither, kept it for safetly reasons.

>     * srs? ( srs-alt? ( mail-filter/libsrs_alt ) )
> 
> 
> I'd personally drop the wrapping srs? USE flag, unles you're going to add REQUIRED_USE srs-alt? ( srs) entries.

I did this on purpose, because +srs-alt.  This thing being "set" doesn't do anything without srs, as it should (and did before).

> Similar concerns on the hunk @ lines ~440.
> 
> In short: whilst I'd keep the SRS USE flag for the moment, I'd only actually use it in REQUIRED_USE. The two implementations are independent and can actually co-exist. So you can use REQUIRED_USE srs ? ( || ( srs-alt srs-native ) ).

Not sure I even want to go to this confusing scenario.  I just want to get rid of srs-alt.

> You're aiming to just drop back to USE=srs? And that be the single implementation? Switching between the two isn't exactly trivial, so I'd keep the split for as long as it's available from upstream.

yes, after I drop srs-alt, only srs remains.

> Perfectly usable as it stands though. So won't be filing bugs :).

Great, I've kept it deliberately "limited" here, so I can go back to a single impl.

PR #20406, opened 2021-04-16
media-gfx/jp2a: new upstream, update to 1.1.0, EAPI 7

myrvogna:
Changing the upstream source to [Talinx/jp2a](https://github.com/Talinx/jp2a), which has releases introducing features such as 256 colors and truecolor output, as per [cslarsen/jp2a#3](https://github.com/cslarsen/jp2a/issues/3).

Not sure what the policy would be in metadata.xml given the two versions (one stable, one not) have different upstream remote-ids – I have added the new one alongside the previous one for now.

Closes: https://bugs.gentoo.org/716744


myrvogna:
Added the related bug tag to the commit message. Given the linked Bugzilla bug is now over a year old, perhaps it should also be reassigned to someone else so the package may be updated?

PR #20410, opened 2021-04-16
net-misc/rsync: Fixes for glibc 2.32

m-gupta:
1. Add upstream glibc 2.32 fixes for rsync 3.2.3 with lchmod.
2. Always calls aclocal/eautoconf etc.

### DIFF
--- a/net-misc/rsync/rsync-3.2.3-r2.ebuild
+++ b/net-misc/rsync/rsync-3.2.3-r2.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-inherit prefix systemd
+inherit autotools prefix systemd
 
 DESCRIPTION="File transfer program to keep remote files into sync"
 HOMEPAGE="https://rsync.samba.org/"
@@ -54,13 +54,16 @@ python_check_deps() {
 	has_version "dev-python/commonmark[${PYTHON_USEDEP}]"
 }
 
+PATCHES=(
+	"${FILESDIR}/${PN}-3.2.3-glibc-lchmod.patch"
+	"${FILESDIR}/${PN}-3.2.3-glibc-lchmod-better-fix.patch"
+)
+
 src_prepare() {
 	default
-	if [[ "${PV}" == *9999 ]] ; then
-		eaclocal -I m4
-		eautoconf -o configure.sh
-		eautoheader && touch config.h.in
-	fi
+	eaclocal -I m4
+	eautoconf -o configure.sh
+	eautoheader && touch config.h.in
 }
 
 src_configure() {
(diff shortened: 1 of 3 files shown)


floppym:
> Disable simd detection for cross-compiles.

Why would we do that? `${ARCH}` should still be valid when cross-compiling.

Regardless, please don't mix this into the same commit; it's a distinct change and deserves its own commit/review.

m-gupta:
> Why would we do that? ${ARCH} should still be valid when cross-compiling.

We do x86 host -> x86 target cross compiles in Chrome OS. The target x86 device may not support the same features as the host. Or the target may not have same library versions etc.

> Regardless, please don't mix this into the same commit; it's a distinct change and deserves its own commit/review.

Will split.

m-gupta:
Removed the simd detection part from the commit.

net-misc/rsync/rsync-3.2.3-r2.ebuild
```diff
@@ -54,13 +54,16 @@ python_check_deps() {
...
 src_prepare() {
 	default
-	if [[ "${PV}" == *9999 ]] ; then
-		eaclocal -I m4
-		eautoconf -o configure.sh
-		eautoheader && touch config.h.in
-	fi
+	eaclocal -I m4
+	eautoconf -o configure.sh
+	eautoheader && touch config.h.in
```
floppym, line 66: We need to inherit autotools to call these functions. Currently, that only happens in the live ebuild.

net-misc/rsync/rsync-9999.ebuild
```diff
@@ -56,11 +56,9 @@ python_check_deps() {
 
 src_prepare() {
 	default
-	if [[ "${PV}" == *9999 ]] ; then
```
floppym, line 59: Please do not remove this condition in the live ebuild. When we copy this ebuild for the next version bump, we do not want to call the autotools functions unnecessarily.

floppym (changes requested):

net-misc/rsync/rsync-9999.ebuild
m-gupta: Sure, i have removed the changes from 9999 ebuild.

net-misc/rsync/rsync-3.2.3-r2.ebuild
m-gupta: done

floppym:
rsync-3.2.3-glibc-lchmod.patch is actually completely unnecessary since the other patch actually reverts it. I consolidated the patches and pushed it as a revbump.

Thanks for bringing the issue to my attention.

m-gupta:
thanks :)

floppym:
I pushed a change that should take care of cross-compiling as well.

https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=81e8e7d6db67b73553187c48893cbc36c546122e

PR #20411, opened 2021-04-16
dev-python/spyder: version bump 5.0.3

AndrewAmmerlaan:
add version 5.0.1, and remove 5.0.0


dev-python/spyder/spyder-5.0.1.ebuild
```diff
@@ -0,0 +1,202 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8,9} )
```
mgorny: Could you use the following from (everywhere)?

```suggestion
PYTHON_COMPAT=( python3_{7..9} )
```

Generally we can expect Python to work in complete ranges.

dev-python/spyder/spyder-5.0.1.ebuild
```diff
@@ -0,0 +1,202 @@
...
+# The warning that this is wrong is a false positive
+# Spyder has setuptools in install_requires
+DISTUTILS_USE_SETUPTOOLS=rdepend
```
mgorny: Is there still a warning, though? If yes, please report a bug. The check is supposed to handle `install_requires`.

dev-python/spyder/spyder-5.0.1.ebuild
```diff
@@ -0,0 +1,202 @@
...
+DOCS_PV="588cdf55cceffea7a3f31d98c0720bfe7df34b72"
+
+DESCRIPTION="The Scientific Python Development Environment"
+HOMEPAGE="
+	https://www.spyder-ide.org/
+	https://github.com/spyder-ide/spyder/
+	https://pypi.org/project/spyder/
+"
+SRC_URI="
+	https://github.com/spyder-ide/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz
```
mgorny: Wrap lines, please.

dev-python/spyder/spyder-5.0.1.ebuild
```diff
@@ -0,0 +1,202 @@
...
+
+# The test suite often hangs or does not work.
+# Technically spyder requires pyqt5<13, which
+# we do not have in ::gentoo any more. Likely
+# this is the reason many of the tests fail
+# or hang. RESTRICTing because IMO it is
+# not worth the several hours I spend every
+# single version bump checking which tests
+# do and do not work. Spyder itself works
+# fine with pyqt5>13.
```
mgorny: Please rewrap at 72 or 80 chars (choose one and follow it consistently).

dev-python/spyder/spyder-5.0.1.ebuild
```diff
@@ -0,0 +1,202 @@
...
+RESTRICT="test"
+
+# White space separated deps are expansion of python-language-server[all] dep
```
mgorny: I would recommend something like this:

```
RDEPEND="(general deps)"

# python-language-server[all] deps
RDEPEND+="(...)"
```

This is the next best thing to having comments inline.

dev-python/spyder/spyder-5.0.1.ebuild
```diff
@@ -0,0 +1,202 @@
...
+	<dev-python/pytest-6.0[${PYTHON_USEDEP}]
+	dev-python/pytest-lazy-fixture[${PYTHON_USEDEP}]
+	dev-python/pytest-mock[${PYTHON_USEDEP}]
+	dev-python/pytest-ordering[${PYTHON_USEDEP}]
+	dev-python/pytest-qt[${PYTHON_USEDEP}]
+	dev-python/pytest-xvfb[${PYTHON_USEDEP}]
+	dev-python/pyyaml[${PYTHON_USEDEP}]
+	dev-python/scipy[${PYTHON_USEDEP}]
+	dev-python/sympy[${PYTHON_USEDEP}]
+)"
```
mgorny:
```suggestion
BDEPEND="
￼	test? (
￼	￼	dev-python/cython[${PYTHON_USEDEP}]
￼	￼	dev-python/flaky[${PYTHON_USEDEP}]
￼	￼	dev-python/matplotlib[tk,${PYTHON_USEDEP}]
￼	￼	dev-python/pandas[${PYTHON_USEDEP}]
￼	￼	dev-python/pillow[${PYTHON_USEDEP}]
￼	￼	<dev-python/pytest-6.0[${PYTHON_USEDEP}]
￼	￼	dev-python/pytest-lazy-fixture[${PYTHON_USEDEP}]
￼	￼	dev-python/pytest-mock[${PYTHON_USEDEP}]
￼	￼	dev-python/pytest-ordering[${PYTHON_USEDEP}]
￼	￼	dev-python/pytest-qt[${PYTHON_USEDEP}]
￼	￼	dev-python/pytest-xvfb[${PYTHON_USEDEP}]
￼	￼	dev-python/pyyaml[${PYTHON_USEDEP}]
￼	￼	dev-python/scipy[${PYTHON_USEDEP}]
￼	￼	dev-python/sympy[${PYTHON_USEDEP}]
￼	￼)"
```

(in other ebuilds too)

dev-python/spyder/spyder-5.0.1.ebuild
```diff
@@ -0,0 +1,202 @@
...
+
+python_prepare_all() {
+	# move docs into workdir
+	mv ../spyder-docs-${DOCS_PV}* docs || die
+
+	# these dependencies are packaged separately:
+	#    dev-python/spyder-kernels,
+	#    dev-python/python-language-server,
+	#    dev-python/qdarkstyle
+	rm external-deps/* -r || die
```
mgorny:
```suggestion
rm -r external-deps/* || die
```

dev-python/spyder/spyder-5.0.1.ebuild
```diff
@@ -0,0 +1,202 @@
...
+
+	# skip online test
+	rm spyder/widgets/github/tests/test_github_backend.py || die
+
+	distutils-r1_python_prepare_all
+}
+
+# Calling pytest directly makes the tests freeze after completing even if successful
+# Exit code is nonzero even upon success, so can't add || die here
+# test results should be checked for success manually
```
mgorny: That's not acceptable. Grep the output then.

dev-python/spyder/spyder-5.0.1.ebuild
```diff
@@ -0,0 +1,202 @@
...
+python_test() {
+	${EPYTHON} runtests.py
```
mgorny:
```suggestion
	"${EPYTHON}" runtests.py
```

dev-python/mkdocs-git-revision-date-localized-plugin/mkdocs-git-revision-date-localized-plugin-0.9.ebuild
```diff
@@ -0,0 +1,48 @@
...
+distutils_enable_tests --install pytest
+
+python_prepare_all() {
+	# mkdocs-git-revision-date-localized-plugin's tests need git repo
+	if use test || use doc; then
+		git init || die
+		git config --global user.email "you@example.com" || die
+		git config --global user.name "Your Name" || die
+		git add . || die
+		git commit -m 'init' || die
```
mgorny, line 44: `-q` maybe?

dev-python/python-lsp-jsonrpc/python-lsp-jsonrpc-1.0.0.ebuild
```diff
@@ -0,0 +1,34 @@
+# Copyright 1999-2021 Gentoo Authors
```
mgorny, line 1: How do these forks work? Does spyder require them in the new version unconditionally, or prefers over the original packages, or...?

mgorny (changes requested):

dev-python/python-lsp-jsonrpc/python-lsp-jsonrpc-1.0.0.ebuild
AndrewAmmerlaan: These forks are the continued versions of python-language-server/python-jsonrpc-server the original versions are no longer actively developed. Spyder currently does not require them yet, but it will in the next release, and once that version is released we can remove the original package and add an move entry to `profiles/updates`

dev-python/spyder/spyder-5.0.1.ebuild
AndrewAmmerlaan: No this seems to have been fixed now, nice :)

dev-python/spyder/spyder-5.0.1.ebuild
AndrewAmmerlaan: I'd rather not mess with the test suite again, it takes two hours to complete the test suite, and it often fails randomly. It would be *a lot* of work, for zero gain since there is `RESTRICT="test"` anyway.

dev-python/spyder/spyder-5.0.1.ebuild
AndrewAmmerlaan: Thanks, this is much clearer

dev-python/python-lsp-jsonrpc/python-lsp-jsonrpc-1.0.0.ebuild
mgorny: No, you won't be able to add a move entry once a new package is added. But if they install under a different name (i.e. not collide with the old upstream), then a move entry is not the right thing to do anyway.

AndrewAmmerlaan: Why not? It's the same thing just distributed under a different name. We would want users who have `dev-python/python-language-server` installed to automatically migrate to `dev-python/python-lsp-server` right? A move entry would accomplish that, or am I mistaken?

mgorny: No, move entry would push two packages into the same CAT-PN-SLOT, resulting in invalid state.

AndrewAmmerlaan: Hmm, I see, so how about we add the new package, stabilize the new package, then mask the old package (with a message pointing to the new package), and then remove the old package later. That should work right?

mgorny: Yes, I think that's the best option.

dev-python/mkdocs-git-revision-date-localized-plugin/mkdocs-git-revision-date-localized-plugin-0.9.ebuild
```diff
@@ -0,0 +1,49 @@
...
+
+DOCS_BUILDER="mkdocs"
+DOCS_DEPEND="dev-python/mkdocs-material"
+
+inherit distutils-r1 docs
+
+DESCRIPTION="Display the localized date of the last git modification of a markdown file"
+HOMEPAGE="
+	https://github.com/timvink/mkdocs-git-revision-date-localized-plugin
+	https://pypi.org/project/mkdocs-git-revision-date-localized-plugin
```
mgorny, line 16: I'm pretty sure pkgcheck is going to flag this as redirect. Always add the trailing slash. And use `pkgcheck scan --commits --net`.

mgorny (commented):
Ok, I think that's the final remarks, provided that there are no test failures. For packages that have pypi remote and working tests, please add python@ as co-maint.

dev-python/mkdocs-git-revision-date-localized-plugin/mkdocs-git-revision-date-localized-plugin-0.9.ebuild
AndrewAmmerlaan: Fixed, Thanks

mgorny:
Bad news:

```
FAILED test/plugins/test_pylint_lint.py::test_lint_free_pylint - assert not [{'code': 'R1732', 'message': "[consider-using-with] Consider using...
FAILED test/plugins/test_pylint_lint.py::test_per_file_caching - assert not [{'code': 'R1732', 'message': "[consider-using-with] Consider using...
```
[dev-python:python-lsp-server-1.0.0:20210425-081440.log](https://github.com/gentoo/gentoo/files/6371518/dev-python.python-lsp-server-1.0.0.20210425-081440.log)

Might be because I've just bumped pylint.

AndrewAmmerlaan:
> Bad news:
> 
> ```
> FAILED test/plugins/test_pylint_lint.py::test_lint_free_pylint - assert not [{'code': 'R1732', 'message': "[consider-using-with] Consider using...
> FAILED test/plugins/test_pylint_lint.py::test_per_file_caching - assert not [{'code': 'R1732', 'message': "[consider-using-with] Consider using...
> ```
> 
> [dev-python:python-lsp-server-1.0.0:20210425-081440.log](https://github.com/gentoo/gentoo/files/6371518/dev-python.python-lsp-server-1.0.0.20210425-081440.log)
> 
> Might be because I've just bumped pylint.

That was indeed the problem, for now I have fixed it by adding `<dev-python/pylint-2.8.0`. I'll see if I can find a better solution soon.

AndrewAmmerlaan:
> Bad news:
> 
> ```
> FAILED test/plugins/test_pylint_lint.py::test_lint_free_pylint - assert not [{'code': 'R1732', 'message': "[consider-using-with] Consider using...
> FAILED test/plugins/test_pylint_lint.py::test_per_file_caching - assert not [{'code': 'R1732', 'message': "[consider-using-with] Consider using...
> ```
> 
> [dev-python:python-lsp-server-1.0.0:20210425-081440.log](https://github.com/gentoo/gentoo/files/6371518/dev-python.python-lsp-server-1.0.0.20210425-081440.log)
> 
> Might be because I've just bumped pylint.

@mgorny Added a patch that fixes this, and removed the `<dev-python/pyltin-2.8.0` dep. Got it from here: https://github.com/python-lsp/python-lsp-server/pull/20

AndrewAmmerlaan:
Bumped to 5.0.2, I'm pretty sure the CI check is only failing because of the pkgcheck upgrade.

PR #20412, opened 2021-04-16
dev-java/dom4j: upgrade jaxen:1.1 -> jaxen:1.2 

vaukai:
Trouble with
```
There was 1 failure:
1) warning(junit.framework.TestSuite$1)
junit.framework.AssertionFailedError: No tests found in org.apache.commons.jelly.core.BaseMemoryLeakTest
        at junit.framework.Assert.fail(Assert.java:57)
        at junit.framework.TestCase.fail(TestCase.java:223)
        at junit.framework.TestSuite$1.runTest(TestSuite.java:96)
        at junit.framework.TestCase.runBare(TestCase.java:142)
        at junit.framework.TestResult$1.protect(TestResult.java:122)
        at junit.framework.TestResult.runProtected(TestResult.java:142)
        at junit.framework.TestResult.run(TestResult.java:125)
        at junit.framework.TestCase.run(TestCase.java:130)
        at junit.framework.TestSuite.runTest(TestSuite.java:241)
        at junit.framework.TestSuite.run(TestSuite.java:236)
        at org.junit.internal.runners.JUnit38ClassRunner.run(JUnit38ClassRunner.java:90)
        at org.junit.runners.Suite.runChild(Suite.java:128)
        at org.junit.runners.Suite.runChild(Suite.java:27)
        at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
        at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
        at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
        at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
        at org.junit.runner.JUnitCore.runMain(JUnitCore.java:77)
        at org.junit.runner.JUnitCore.main(JUnitCore.java:36)

FAILURES!!!
Tests run: 121,  Failures: 1
```

### DIFF
--- a/dev-java/dom4j/dom4j-1.6.1-r9.ebuild
+++ b/dev-java/dom4j/dom4j-1.6.1-r9.ebuild
@@ -1,0 +1,84 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+JAVA_PKG_IUSE="doc source test"
+
+inherit java-pkg-2 java-ant-2
+
+DESCRIPTION="XML Java library"
+HOMEPAGE="https://dom4j.github.io/"
+SRC_URI="
+	mirror://sourceforge/dom4j/${P}.tar.gz
+	mirror://gentoo/${P}-java5.patch.bz2"
+
+LICENSE="dom4j"
+SLOT="1"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~sparc-solaris ~x64-solaris ~x86-solaris"
+IUSE=""
+
+CDEPEND="
+	dev-java/xpp2:0
+	dev-java/xpp3:0
+	dev-java/xsdlib:0
+	dev-java/relaxng-datatype:0"
+
+RDEPEND="
+	${CDEPEND}
+	>=virtual/jre-1.8:*"
+
+# restricted to jdk 1.8 because it needs javax.xml.bind
+DEPEND="
+	${CDEPEND}
+	test? (
+		dev-java/xalan:0
+		dev-java/ant-junit:0
+		dev-java/junitperf:0
+	)
+	virtual/jdk:1.8"
+
+PDEPEND="dev-java/jaxen:1.2"
+
+# Add missing methods to compile on Java 5 #137970
+PATCHES=( "${WORKDIR}/${P}-java5.patch" )
+
+JAVA_RM_FILES=(
+	src/test/org/dom4j/bean/BeansTest.java
+	src/test/org/dom4j/io/StaxTest.java
+	src/test/org/dom4j/rule/RuleTest.java
+	src/test/org/dom4j/ThreadingTest.java
+	src/test/org/dom4j/io/XPP3ReaderTest.java
+)
+
+src_prepare() {
+	default
+
+	# Circular deps with jaxen #212993
+	find -name '*.jar' -! -name jaxen-1.1-beta-6.jar -exec rm -v {} + || die
+
+	rm -v "${JAVA_RM_FILES[@]}" || die
+}
+
+JAVA_ANT_REWRITE_CLASSPATH="yes"
+JAVA_ANT_CLASSPATH_TAGS+=" javadoc"
+
+JAVA_ANT_ENCODING="ISO-8859-1"
+
+EANT_BUILD_TARGET="clean package"
+EANT_GENTOO_CLASSPATH="relaxng-datatype,xpp2,xpp3,xsdlib"
+EANT_GENTOO_CLASSPATH_EXTRA="lib/jaxen.jar"
+EANT_EXTRA_ARGS="-Dbuild.javadocs=build/doc/api"
+
+EANT_TEST_GENTOO_CLASSPATH="${EANT_GENTOO_CLASSPATH},junitperf,xalan"
+
+src_test() {
+	java-pkg-2_src_test
+}
+
+src_install() {
+	java-pkg_dojar "build/${PN}.jar"
+	java-pkg_register-dependency jaxen-1.2
+	use doc && java-pkg_dojavadoc build/doc/api
+	use source && java-pkg_dosrc src/java/*
+}


vaukai:
```
>>> Installing (23 of 23) dev-java/commons-jelly-1.0.1::testing
 * checking 2 files for package collisions
>>> Merging dev-java/commons-jelly-1.0.1 to /
 * Possibly unneeded dependencies found
 *      forehead
 *      xerces-2
 *      commons-collections-4
 * Class files not found via DEPEND in package.env
 *      org/apache/commons/collections/ArrayStack.class
 *      org/apache/commons/collections/iterators/EnumerationIterator.class
 *      org/apache/commons/collections/iterators/SingletonIterator.class
 *      org/apache/commons/collections/iterators/ArrayIterator.class
 *      org/apache/commons/collections/BeanMap.class
--- /usr/
--- /usr/share/
>>> /usr/share/commons-jelly-1/
>>> /usr/share/commons-jelly-1/lib/
>>> /usr/share/commons-jelly-1/lib/commons-jelly.jar
>>> /usr/share/commons-jelly-1/package.env
>>> dev-java/commons-jelly-1.0.1 merged.

 * Messages for package dev-java/dom4j-1.6.1-r8:

 * Class files not found via DEPEND in package.env
 *      org/jaxen/dom4j/DocumentNavigator.class
 *      org/jaxen/dom4j/Dom4jXPath.class

 * Messages for package dev-java/commons-jelly-1.0.1:

 * Possibly unneeded dependencies found
 *      forehead
 *      xerces-2
 *      commons-collections-4
 * Class files not found via DEPEND in package.env
 *      org/apache/commons/collections/ArrayStack.class
 *      org/apache/commons/collections/iterators/EnumerationIterator.class
 *      org/apache/commons/collections/iterators/SingletonIterator.class
 *      org/apache/commons/collections/iterators/ArrayIterator.class
 *      org/apache/commons/collections/BeanMap.class
>>> Auto-cleaning packages...

>>> No outdated packages were found on your system.
```

vaukai:
How could this be solved?
```
Could not load class: org.apache.commons.jelly.DOESNOTEXIST so taglib instantiation failed
```

dev-java/jaxen/jaxen-1.2.0-r1.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	# https://bugs.gentoo.org/739894#c9
+	rm -rv "${JAVA_SRC_DIR}"/org/jaxen/{jdom,xom} || die
+	use dom4j || rm -rv "${JAVA_SRC_DIR}"/org/jaxen/dom4j || die
+	use dom4j && JAVA_GENTOO_CLASSPATH="dom4j-1"
+}
+
+src_install() {
+	default
+	java-pkg-simple_src_install
+}
```
fordfrog: i guess this is not needed, i suppose this is the default behavior.

dev-java/jaxen/jaxen-1.2.0-r1.ebuild
```diff
@@ -0,0 +1,65 @@
...
+RDEPEND="
+	>=virtual/jre-1.8:*
+"
+
+S="${WORKDIR}"
+
+DOCS=( "${P}"/{LICENSE.txt,README.md} )
+
+JAVA_CLASSPATH_EXTRA="xerces-2"
+JAVA_SRC_DIR="${P}/src/java/main"
```
fordfrog: tests are in `IUSE` but settings for tests is missing here.

dev-java/jaxen/jaxen-1.2.0-r1.ebuild
vaukai: Ugly, I know. But without this, the `DOCS=()` array is ignored.

fordfrog: ah, tricky... the eclass would need some care it seems :-)

fordfrog:
> Trouble with
> 
> ```
> There was 1 failure:
> 1) warning(junit.framework.TestSuite$1)
> junit.framework.AssertionFailedError: No tests found in org.apache.commons.jelly.core.BaseMemoryLeakTest
>         at junit.framework.Assert.fail(Assert.java:57)
>         at junit.framework.TestCase.fail(TestCase.java:223)
>         at junit.framework.TestSuite$1.runTest(TestSuite.java:96)
>         at junit.framework.TestCase.runBare(TestCase.java:142)
>         at junit.framework.TestResult$1.protect(TestResult.java:122)
>         at junit.framework.TestResult.runProtected(TestResult.java:142)
>         at junit.framework.TestResult.run(TestResult.java:125)
>         at junit.framework.TestCase.run(TestCase.java:130)
>         at junit.framework.TestSuite.runTest(TestSuite.java:241)
>         at junit.framework.TestSuite.run(TestSuite.java:236)
>         at org.junit.internal.runners.JUnit38ClassRunner.run(JUnit38ClassRunner.java:90)
>         at org.junit.runners.Suite.runChild(Suite.java:128)
>         at org.junit.runners.Suite.runChild(Suite.java:27)
>         at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
>         at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
>         at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
>         at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
>         at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
>         at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
>         at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
>         at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
>         at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
>         at org.junit.runner.JUnitCore.runMain(JUnitCore.java:77)
>         at org.junit.runner.JUnitCore.main(JUnitCore.java:36)
> 
> FAILURES!!!
> Tests run: 121,  Failures: 1
> ```

the class can be ignored if it contains no tests, using the exclude variable.

dev-java/jaxen/jaxen-1.2.0-r1.ebuild
vaukai: @fordfrog Test are still `RESTRICT="test", just like in `jaxen-1.2.0.ebuild`.
But, how do you think about using the `test` flag instead of introducing the local `dom4j` flag?

So far I only unveiled the dom4j directory which is needed for commons-jelly.
But should (to be tested) also work for `jdom` and `xom`.

vaukai:
> the class can be ignored if it contains no tests, using the exclude variable.

Done.

dev-java/jaxen/jaxen-1.2.0-r1.ebuild
vaukai: > ah, tricky... the eclass would need some care it seems :-)

Good idea, good for the docs ...

dev-java/jaxen/jaxen-1.2.0-r1.ebuild
fordfrog: it came to my mind that we could provide all three use flags. and depending on enabled use flags appropriate tests might be run, if they are not mixed together. if they are mixed together, then all three flags would be needed for tests, as otherwise there is nothing to test. what do you think?

vaukai: My preference? Keep it simple and stupid, similar to what's proposed by java-ebuilder: [jaxen-1.2.0.ebuild.gz](https://github.com/gentoo/gentoo/files/6329290/jaxen-1.2.0.ebuild.gz)
I would like to replace the `dom4j` USE flag with the `test` flag and keep `jdom` and `xom` hidden at the moment.
Unhiding `jdom` and `xom` and enabling tests could be done later.  I know that's kind of misusing the `test` flag.

The other topic is the circular dependency. Presently `dom4j` pulls jaxen in the older slot.  Once you pushed `dom4j` to pulling `jaxen:1.2` ...
But the circular dependency is given upstream and should be solved there.

fordfrog: but `java-ebuilder` does not remove any code. and you cannot depend on `test` use flag from `commons-jelly`.

vaukai: > but `java-ebuilder` does not remove any code. and you cannot depend on `test` use flag from `commons-jelly`.

So it's not allowed to have `commons-jelly` with
```
dev-java/jaxen:1.2[test]
```
??
If so, could you please take it as is (Please check the flag description in `metadata.xml`) and let's keep `xom` and `jdom` for later.
The revbump is needed, right?

fordfrog: `dev-java/jaxen:1.2[test]` misuses `[test]` for something it's not intended for. i'm fine with `dev-java/jaxen:1.2[dom4j]`, i would not be even afraid to extend it with `[xom,jdom]`, as that might be needed in the future anyway. and portage has some useful hints when these cyclical deps occur. but we can leave it as it is now, for now. there is only one think i don't like about this ebuild and i did not notice it on our previous merge, and that is the fact that it has `[test]` use flag but if the restriction would be removed, the tests would not work as they are not defined. so i'd suggest to remove `[test]` completely as this is just a partial non-working attempt for implementation and the flag is useless without test setup in the ebuild. it still would be useful to keep there the note about why tests are not implemented for future reference.

vaukai: >  so i'd suggest to remove `[test]` completely

Could you please do that on merge?  Or is there any other important changes needed?  
I'm glad to have a solution for `commons-jelly`

fordfrog: ok, i'll take it as it is, will clean it up and do final testing.

vaukai: Thanks

fordfrog:
@vaukai both `jaxen` commits are merged. you can see my change in a871d2b01b7673788592d8eb1ab34101c187fa79

dev-java/commons-jelly/commons-jelly-1.0.1.ebuild
```diff
@@ -0,0 +1,76 @@
...
+	dev-java/commons-jexl:1.0
+	dev-java/commons-lang:2.1
+	dev-java/commons-logging:0
+	dev-java/dom4j:1
+	dev-java/forehead:0
+	dev-java/jaxen:1.2[dom4j]
+	dev-java/junit:0
+	dev-java/tomcat-jstl-spec:0
+	java-virtuals/servlet-api:3.0
+"
```
fordfrog, line 55: did you try that none of the latest slots of each dep cannot be used? following are those that should be tested: `commons-lang` (we really want to avoid anything older than `:3.6` as it does not compile with java source set to `1.8`), `servlet-api` (here you can in fact depend on `dev-java/tomcat-servlet-api`, just avoid slot `5.0`, there was a change in package names and it is not backward compatible).

jfyi, `commons-collections` has different package names for slot `:0` and `:4` so that one would not work for sure. similar for `commons-jexl` and slots `:1.0` and `:2`.

vaukai:
> @vaukai both `jaxen` commits are merged. you can see my change in [a871d2b](https://github.com/gentoo/gentoo/commit/a871d2b01b7673788592d8eb1ab34101c187fa79)

Thanks. Just rebased.

fordfrog:
@vaukai just searched for packages depending on `<commons-lang-3` and came again across `commons-jelly` and i did not find any packages depending on it. do we really need it for something or we can last-rite it? there's no point in keeping libraries in the tree that we do not have a use for.

vaukai:
> @vaukai just searched for packages depending on `<commons-lang-3` and came again across `commons-jelly` and i did not find any packages depending on it. do we really need it for something or we can last-rite it? there's no point in keeping libraries in the tree that we do not have a use for.

I would be glad to get rid of that one and I will just remove it from this PR. Pls go ahead and last-rite.
Then only the dom4j remains here.

dev-java/commons-jelly/commons-jelly-1.0.1.ebuild
vaukai: Keeping this information bookmarked

fordfrog:
@vaukai thanks! there is still a failing test though... but...

PR #20416, opened 2021-04-17
kde-plasma/plasma-workspace: Make qalculate USE flag enabled by default

zpuskas:
@a17r Please let me know if you need me to make any modifications in case the PR is not up to KDE project's standards.

### DIFF
--- a/kde-plasma/plasma-workspace/metadata.xml
+++ b/kde-plasma/plasma-workspace/metadata.xml
@@ -8,7 +8,7 @@
 	<use>
 		<flag name="appstream">Enable AppStream software metadata support</flag>
 		<flag name="geolocation">Enables dataengine providing location information</flag>
-		<flag name="qalculate">Enable Qalculate runner using <pkg>sci-libs/libqalculate</pkg></flag>
+		<flag name="qalculate">Use <pkg>sci-libs/libqalculate</pkg> instead of QJSEngine (<pkg>dev-qt/qtdeclarative</pkg>) for krunner calculator</flag>
 		<flag name="screencast">Enable screencast portal using <pkg>media-video/pipewire</pkg></flag>
 		<flag name="telemetry">Enable User Feedback control module for <pkg>kde-plasma/systemsettings</pkg></flag>
 	</use>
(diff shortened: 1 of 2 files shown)


kde-plasma/plasma-workspace/metadata.xml
```diff
@@ -8,7 +8,7 @@
 	<use>
 		<flag name="appstream">Enable AppStream software metadata support</flag>
 		<flag name="geolocation">Enables dataengine providing location information</flag>
-		<flag name="qalculate">Enable Qalculate runner using <pkg>sci-libs/libqalculate</pkg></flag>
+		<flag name="qalculate">Replace krunner calculator QJSEngine engine with Qalculate using <pkg>sci-libs/libqalculate</pkg></flag>
```
Arfrever, line 11: More standard phrasings would be `Use ... instead of ...`.
(See e.g. `gles2-only`, `graphicsmagick`, `libressl`, `llvm-libunwind` global USE flags).

So possible suggestions for this case:
```
<flag name="qalculate">Use <pkg>sci-libs/libqalculate</pkg> instead of QJSEngine (<pkg>dev-qt/qtdeclarative</pkg>) for krunner calculator</flag>
<flag name="qalculate">Use <pkg>sci-libs/libqalculate</pkg>-based engine instead of QJSEngine (<pkg>dev-qt/qtdeclarative</pkg>)-based engine for krunner calculator</flag>

zpuskas: Thank you for the suggestions, I've went with the more concise version.

a17r:
I tend to agree with upstream comment 4 - why provide an incorrect calculator at all. We could then use a more general "calculator" default-enabled use flag with clear description.

zpuskas:
> but instead I would recommend to move that PR, if you want to do it, over to gentoo/kde repository and change the live ebuilds. Then the change will come to tree with 5.21.5 version bump.

I've created https://github.com/gentoo/kde/pull/901. I'll look into submitting a patch upstream where I also gut the C++ QJSEngine based code.

zpuskas:
Abandoning this PR in favour of gentoo/kde PR.

PR #20418, opened 2021-04-17
media-libs/netpbm: new version 10.86.21

ceamac:
- New version from stable branch.
- Created a tarball hosted at github.
- Fixed most of the tests, still fails `ps-alt-roundtrip` if built with USE `postscript`.  Possible bug in `pbmtolps`, should contact upstream.  Related bug: https://bugs.gentoo.org/670362

About the remaining bugs:
- https://bugs.gentoo.org/618612 - no sure what to do about this, I could not find what version this is fixed in, but other distributions (fedora, redhat) ship 10.93, so maybe it's fixed there.
- https://bugs.gentoo.org/762298 - 10.93.00 is not from the *stable* branch, it's from *advanced*.
- https://bugs.gentoo.org/672762 - some of them fixed by upstream, some not in stable, cannot test for now
- https://bugs.gentoo.org/605436 - fixed in 10.76.00 and 10.86.21, not fixed in 10.70.00
- https://bugs.gentoo.org/672624 - works fine on amd64, seems to be on x86.  Cannot test for now
- https://bugs.gentoo.org/676200 - according to a comment was caused by `media-libs/urt`


media-libs/netpbm/files/make-tarball.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
```
vapier, line 1: rebase and drop this hunk as it's obsolete now

media-libs/netpbm/files/netpbm-10.86.21-build.patch
```diff
@@ -0,0 +1,66 @@
+- Remake netpbm-10.76.00-build.patch for 10.86.21; original description follows.
```
vapier, line 1: don't refer to old patches.  just create the new one and update the description as needed.

media-libs/netpbm/files/netpbm-10.86.21-format-security.patch
```diff
@@ -0,0 +1,18 @@
+Fix compiling with `CFLAGS` `-Werror=format-security`
```
vapier, line 1: have you sent this upstream ?

media-libs/netpbm/files/netpbm-10.86.21-misc-deps.patch
```diff
@@ -0,0 +1,70 @@
...
+ 
+ SUBDIRS = $(PRODUCT_SUBDIRS) $(SUPPORT_SUBDIRS)
+ 
+--- a/converter/other/Makefile
++++ b/converter/other/Makefile
+@@ -7,6 +7,7 @@
+ 
+ include $(BUILDDIR)/config.mk
+ 
++ifneq ($(XML2_LIBS),NONE)
```
vapier, line 25: pretty sure we don't need this anymore.  we should be able to force `TEST_PKGCONFIG_LIBXML2=...` in the ebuild instead.

media-libs/netpbm/files/netpbm-10.86.21-test.patch
```diff
@@ -0,0 +1,10 @@
+`print ("got color map\n");` removed from `editor/pnmquantall` so remove from test
```
vapier, line 1: have you sent this upstream ?

media-libs/netpbm/netpbm-10.86.21.ebuild
```diff
@@ -0,0 +1,208 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit toolchain-funcs
+
+DESCRIPTION="A set of utilities for converting to/from the netpbm (and related) formats"
+HOMEPAGE="http://netpbm.sourceforge.net/"
+SRC_URI="https://github.com/ceamac/netpbm-make-dist/releases/download/v10.86.21/${P}.tar.xz"
```
vapier, line 10: i don't think we want to do this.  we'll just maintain the archives in Gentoo mirrors.

vapier (changes requested):

media-libs/netpbm/netpbm-10.86.21.ebuild
thesamesam: Agreed it sucks, but infra disabled dumping tarballs into mirror://gentoo, so the only choice is this or devspace which are equivalent

media-libs/netpbm/files/make-tarball.sh
ceamac: Done

media-libs/netpbm/files/netpbm-10.86.21-build.patch
ceamac: Done

media-libs/netpbm/files/netpbm-10.86.21-format-security.patch
ceamac: Not yet, but I plan to, together with the postscript bug.  I'll add a comment when it's done.

media-libs/netpbm/files/netpbm-10.86.21-misc-deps.patch
ceamac: This part of the Makefile decides whether to use *pkg-config* or *xml2-config* to find `libxml2`.  Without the patch it would autodetect it regardless of the USE flag.  Same later for png and X.

media-libs/netpbm/files/netpbm-10.86.21-test.patch
ceamac: Same as above, I will.

media-libs/netpbm/netpbm-10.86.21.ebuild
ceamac: https://devmanual.gentoo.org/general-concepts/mirrors/#suitable-download-hosts

mirror://gentoo is deprecated and I don't have a dev space, so this was the closest I could do.  At least it's on github.

media-libs/netpbm/files/netpbm-10.86.21-test.patch
ceamac: It's been fixed by upstream in advanced *10.90.00*

media-libs/netpbm/files/netpbm-10.86.21-format-security.patch
ceamac: It's been fixed by upstream in advanced *10.91.03*

ceamac:
It seems upstream removed `pnmtolps` from the latest advanced build, so `ps-alt-roundtrip` doesn't fail anymore.  https://bugs.gentoo.org/670362 will stay open until that change reaches stable.

media-libs/netpbm/netpbm-10.86.21.ebuild
```diff
@@ -0,0 +1,208 @@
...
+}
+
+netpbm_ldshlib() {
+	case ${CHOST} in
+		*-darwin*) echo '$(LDFLAGS) -dynamiclib -install_name $(SONAME)';;
+		*)         echo '$(LDFLAGS) -shared -Wl,-soname,$(SONAME)';;
+	esac
+}
+netpbm_config() {
+	if use $1 ; then
```
juippis, line 60:
```suggestion
	if use ${1} ; then
```

media-libs/netpbm/netpbm-10.86.21.ebuild
```diff
@@ -0,0 +1,208 @@
...
+
+	doman userguide/*.[0-9]
+	dodoc README
+
+	cd doc || die
+	dodoc HISTORY Netpbm.programming USERDOC
+	docinto html
+	dodoc -r *.html
+
+	use doc && dodoc -r ../userguide/*.html
```
juippis, line 207: Since the generation of these files isn't being controlled by this USE flag, how about just installing them always and dropping `doc` IUSE altogether?

juippis (approved):
I wonder about the purpose of *make-tarball.sh* file.

Builds fine, waiting for other comments, please ping again if no one merges in couple of days. Didn't read patches or the ebuild that carefully...

media-libs/netpbm/netpbm-10.86.21.ebuild
ceamac: Most *html* files in `userguide` are already converted to *man* and installed as man pages.  The `doc` flag controls installing the original *html* files, as they are a bit redundant.

ceamac:
@juippis please look again

juippis (approved):

PR #20420, opened 2021-04-17
qtwebkit-removal

a17r:


AdelKS:
Unfortunately up until now, `net-misc/teamviewer` also still depends on this https://bugs.gentoo.org/750899

a17r:
> Unfortunately up until now, `net-misc/teamviewer` also still depends on this https://bugs.gentoo.org/750899

So do others, and they will need to be masked.

Pesa (approved):
Thanks!

steils:
But openshot-2.5.1 still requires qtwebkit. It will not in the next release which we are craving for...

a17r:
> But openshot-2.5.1 still requires qtwebkit. It will not in the next release which we are craving for...

I think we've kept the tracker running for long enough. All remaining revdeps should be masked together with qtwebkit, those without hope of porting should be last-rited already. Packages with ongoing upstream porting work (as far as I can see openshot is the *only* one) can remain masked provided they can link to an upstream porting inquiry so we get an idea of an ETA.

a17r:
> > But openshot-2.5.1 still requires qtwebkit. It will not in the next release which we are craving for...
> 
> Packages with ongoing upstream porting work (as far as I can see openshot is the _only_ one) can remain masked provided they can link to an upstream porting inquiry so we get an idea of an ETA.

@steils is there any such upstream issue/bug/tracker we can link to?

PR #20432, opened 2021-04-18
dev-python/py-gfm: bump to 1.0.2

scardracs:
Add support to python3.9 (tested with 3.9.4)

Closes: https://bugs.gentoo.org/680260
Closes: https://bugs.gentoo.org/748477


dev-python/py-gfm/py-gfm-1.0.2.ebuild
```diff
@@ -0,0 +1,20 @@
...
+HOMEPAGE="https://github.com/Zopieux/py-gfm"
+SRC_URI="https://github.com/Zopieux/py-gfm/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~x86"
+
+RDEPEND="dev-python/markdown[${PYTHON_USEDEP}]"
+DEPEND="${RDEPEND}
+	dev-python/setuptools[${PYTHON_USEDEP}]"
```
thesamesam, line 20: unnecessary

dev-python/py-gfm/py-gfm-1.0.2.ebuild
```diff
@@ -0,0 +1,20 @@
...
+DESCRIPTION="Github-Flavored Markdown for Python-Markdown"
+HOMEPAGE="https://github.com/Zopieux/py-gfm"
+SRC_URI="https://github.com/Zopieux/py-gfm/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~x86"
+
+RDEPEND="dev-python/markdown[${PYTHON_USEDEP}]"
+DEPEND="${RDEPEND}
```
thesamesam, line 19: likely unnecessary, also no tests?

scardracs: This should be the test: https://github.com/Zopieux/py-gfm/blob/master/tests/test_gfm.py and there are lots of "some-code" "foo" ecc... I don't think test actually works or do something useful

scardracs:
@thesamesam opened a new PR due to internal problem with branch... New PR here https://github.com/gentoo/gentoo/pull/20435

PR #20433, opened 2021-04-18
app-benchmarks/siege: bump to 4.0.9

scardracs:


thesamesam:
Ideally don't 'git mv' so we can easily revert commits. The exception is if the current version is critically broken, but even then...

scardracs:
> Ideally don't 'git mv' so we can easily revert commits. The exception is if the current version is critically broken, but even then...

Ok :) I'm adding back 4.0.7 too

app-benchmarks/siege/siege-4.0.9.ebuild
```diff
@@ -0,0 +1,52 @@
...
+	local myconf=(
+		$(use_with ssl ssl "${EPREFIX}/usr")
+	)
+	econf "${myconf[@]}"
+}
+
+src_install() {
+	emake DESTDIR="${ED}" install
+
+	dodoc AUTHORS ChangeLog INSTALL README* doc/siegerc doc/urls.txt
```
juippis, line 44: Ideally you'd define these as a global `DOCS` array, so you could just...

juippis, line 44: Call `default` here. This only seems to install *README.md*, so no globbing needed.

juippis (approved):

PR #20434, opened 2021-04-18
package.mask: mask >=dev-java/xerces-2.12.1 

vaukai:


profiles/package.mask
```diff
@@ -32,6 +32,10 @@
 
 #--- END OF EXAMPLES ---
 
+# Volkmar W. Pogatzki <gentoo@pogatzki.net> (2021-04-18)
+# dev-java/xerces-2.12.1 depends on masked >=virtual/{jdk,jre}-11
```
thesamesam, line 36: Usually a bug reference in the ebuild and package.mask is a good idea to let people know why it's definitely Java 11 it needs.

fordfrog:
thanks!

PR #20440, opened 2021-04-18
use ${EPYTHON}, not python3 in mm packages

trofi:


dev-cpp/atkmm/atkmm-2.28.1.ebuild
```diff
@@ -2,7 +2,9 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
-inherit gnome.org meson multilib-minimal
+
+PYTHON_COMPAT=(python3_{6..9})
```
mattst88, line 6: I'd prefer `PYTHON_COMPAT=( python3_{6..9} )` (e.g. with spaces) since that style is used 4463 times in the tree vs 88 times without spaces.

mattst88:
Thanks! Please commit.

dev-cpp/atkmm/atkmm-2.28.1.ebuild
mgorny: python3_6 has been removed in January.

trofi: Pushed with changed whitespace and dropped 3.6 as `PYTHON_COMPAT=( python3_{6..7} )`

PR #20446, opened 2021-04-18
mail-client/mutt-wizard: add sasl runtime neomutt dependency

epsilon-0:
Closes: https://bugs.gentoo.org/765763

---
No need for revision bump as package was unusable without this, so everyone had to enable it manually anyway.


juippis (approved):
I'll revbump though, there are other benefits to it, for example someone having `neomutt[sasl]` on by other reason, then removing it.

PR #20448, opened 2021-04-18
Add support to build rust, mesa and spidermonkey with llvm-12

J-Bu:
Update ebuilds and backport upstream patches to build successfully with llvm-12

### DIFF
--- a/dev-lang/rust/files/rustc-1.51.0-backport-pr81451-1.patch
+++ b/dev-lang/rust/files/rustc-1.51.0-backport-pr81451-1.patch
@@ -1,0 +1,497 @@
+From 55f345f32505c2095966a5dc46c4ae3290dbf7a1 Mon Sep 17 00:00:00 2001
+From: Nikita Popov <nikita.ppv@gmail.com>
+Date: Tue, 3 Nov 2020 22:47:16 +0100
+Subject: [PATCH 01/12] Support LLVM 12 in rustc
+
+---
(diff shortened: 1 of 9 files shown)


Whissi (changes requested):
NAK from mozilla project due to https://bugzilla.mozilla.org/show_bug.cgi?id=1658401 (LLVM 12 support will land with next version which should land this week). So please drop spidermonkey from your PR.

J-Bu:
@Whissi 
Strange, I did a successfull testbuild of spidermonkey-78.9.0 with llvm and clang 12 but I can remove it from the PR

stefson:
I'm certainly not a dev, but you might have better chances to get the dev-lang/rust changes merged if you don't git mv the -r2 to -r3, but rather copy the 1.51.0-r2 ebuild and add a silly high number to revision, such as -r99 or something along those lines, and you will have to drop keywords entirely. This will be enough to signal everyone to stay away from it, and everyone who wants to play with fire can keyword the ebuild locally. 

regaring mesa, I read through the upstream issue and it seems to cause further issues down the road with unreleased llvm-13, so maybe its best to wait and see this fixed properly?

telans:
> dev-lang/rust-1.51 is stable candidate (to address some security issues) and will not gain llvm:12 support.

Why not revbump rust-1.51 for one version in testing + llvm:12, and one version stable + llvm:11?

PR #20449, opened 2021-04-19
sci-libs/opencascade: fix Qt dependency

waebbl:
Changes a dependency on dev-qt/linguist-tool

Closes: https://bugs.gentoo.org/783810

### DIFF
--- a/sci-libs/opencascade/opencascade-7.4.0-r4.ebuild
+++ b/sci-libs/opencascade/opencascade-7.4.0-r4.ebuild
@@ -9,19 +9,21 @@ EAPI=7
 
 inherit check-reqs cmake flag-o-matic java-pkg-opt-2
 
+MY_PV="$(ver_rs 1- '_')"
+
 DESCRIPTION="Development platform for CAD/CAE, 3D surface/solid modeling and data exchange"
 HOMEPAGE="https://www.opencascade.com"
-MY_PV="$(ver_rs 1- '_')"
 SRC_URI="https://git.dev.opencascade.org/gitweb/?p=occt.git;a=snapshot;h=refs/tags/V${MY_PV};sf=tgz -> ${P}.tar.gz"
+S="${WORKDIR}/occt-V${MY_PV}"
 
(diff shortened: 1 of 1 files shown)


sci-libs/opencascade/opencascade-7.4.0-r4.ebuild
```diff
@@ -9,19 +9,21 @@ EAPI=7
...
 LICENSE="|| ( Open-CASCADE-LGPL-2.1-Exception-1.0 LGPL-2.1 )"
 SLOT="${PV}"
 KEYWORDS="~amd64 ~arm64 ~x86"
-# gl2ps
-IUSE="debug doc examples ffmpeg freeimage gles2 inspector java optimize qt5 tbb +vtk"
+IUSE="debug doc +examples ffmpeg freeimage gles2 inspector java optimize qt5 tbb +vtk"
 
 REQUIRED_USE="
 	inspector? ( qt5 )
+	qt5? ( examples )
```
a17r, line 26: That looks strange and begs for an explanation in the git summary.

a17r (changes requested):

sci-libs/opencascade/opencascade-7.4.0-r4.ebuild
waebbl: Qt is only needed with the examples. In my 7.5.1 bump (c.f. https://github.com/gentoo/gentoo/pull/20443) I completely removed the qt5 USE flags. The library code base itself has no dependency whatsoever on Qt. However, I didn't recognize that when making the 7.4.0 ebuild, but just now when working on the 7.5.1 ebuild.

Some of the examples, however, depend on qt5. I thought about changing it the way, I've done it with 7.5.1, but then decided against it, as this patch works, although the other way looks more clean and 7.5.1 is on the horizon.

Inspector is a binary tool, which depends on Qt5 as well. A deep inspection pro tool for models created with opencascade. It's more like an example as well, than a tool for the average user. In 7.5.1, I build it when examples is enabled and removed the USE flag as well.

If you're having a good idea on how to handle this, I'm happy to hear about them.

PR #20450, opened 2021-04-19
net-vpn/i2pd: version bump 2.37.0

khumarahn:
A regular version bump.

No changes, but took a while before I could test it.

Thank you!


juippis (commented):
Please still split adding a new version and removing old version to their separate commits. So this should have 2 commits. And preferably rebase against master, so we get a green CI check.

khumarahn:
ok. I was confused that the 2 old versions I used to keep here were suddenly removed.

juippis (approved):
Thanks!

Yeah i2pd might've been a victim of my huge proxy-maint spring cleaning, to get some sense into stabilization, libressl and new PYTHON_COMPAT situation.

PR #20454, opened 2021-04-19
media-video/pipewire: Wiring up of jack-sdk IUSE & multilib and various fixes.

pinkflames:
This draft PR is a WIP on implementing the jack-sdk IUSE for making media-video/pipewire a virtual/jack provider.

TODO:
  - [x] disable building of unused code for non-native ABIs (partially done - disabling more easily breaks tests)
  - [ ] either discard or fix ({README,INSTALL}.md NEWS) installation (man pages are built and installed, so this is a bit of a fluff anyway)
  - [ ] test that virtual/jack reverse dependencies build and run correctly with media-video/pipewire[jack-sdk]
    - [x] the small section of _less pro_-audio applications used by the PR's author build and seem to run fine
    - [ ] have a pro user report back on a DAW or similar use-case(?)
    - [x] and in the meantime mask jack-sdk IUSE
  - [ ] either fix or remove pw-jack loader when jack-sdk is enabled (since overriding the libdir breaks its built-in multilib support, and it's no longer needed once we replace system's JACK libraries)
  - [x] wait for the next PipeWire release to come out for this to be available ;)

The only important non-jack/multilib change is the removal of the docdir patch, [since it has been upstreamed](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1057).

### DIFF
--- a/media-video/pipewire/pipewire-9999.ebuild
+++ b/media-video/pipewire/pipewire-9999.ebuild
@@ -3,11 +3,11 @@
 
 EAPI=7
 
-inherit meson optfeature udev
+inherit meson optfeature udev multilib-minimal
 
 if [[ ${PV} == 9999 ]]; then
 	EGIT_REPO_URI="https://gitlab.freedesktop.org/${PN}/${PN}.git"
-#	EGIT_BRANCH="master"
+	EGIT_BRANCH="master"
 	inherit git-r3
 else
 	SRC_URI="https://gitlab.freedesktop.org/${PN}/${PN}/-/archive/${PV}/${P}.tar.gz"
@@ -19,7 +19,7 @@ HOMEPAGE="https://pipewire.org/"
 
 LICENSE="LGPL-2.1+"
 SLOT="0/0.3"
-IUSE="aac aptx bluetooth doc extra gstreamer jack-client ldac pipewire-alsa systemd test v4l"
+IUSE="aac aptx bluetooth doc extra gstreamer jack-client jack-sdk ldac pipewire-alsa systemd test v4l"
 
 # Once replacing system JACK libraries is possible, it's likely that
 # jack-client IUSE will need blocking to avoid users accidentally
@@ -30,6 +30,7 @@ IUSE="aac aptx bluetooth doc extra gstreamer jack-client ldac pipewire-alsa syst
 REQUIRED_USE="
 	aac? ( bluetooth )
 	aptx? ( bluetooth )
+	jack-sdk? ( !jack-client )
 	ldac? ( bluetooth )
 "
 
@@ -42,9 +43,10 @@ BDEPEND="
 "
 RDEPEND="
 	media-libs/alsa-lib
-	sys-apps/dbus
+	sys-apps/dbus[${MULTILIB_USEDEP}]
 	sys-libs/ncurses[unicode]
-	virtual/libudev
+	virtual/libintl[${MULTILIB_USEDEP}]
+	virtual/libudev[${MULTILIB_USEDEP}]
 	bluetooth? (
 		aac? ( media-libs/fdk-aac )
 		aptx? ( media-libs/libopenaptx )
@@ -61,14 +63,18 @@ RDEPEND="
 		media-libs/gst-plugins-base:1.0
 	)
 	jack-client? ( >=media-sound/jack2-1.9.10:2[dbus] )
+	jack-sdk? (
+		!media-sound/jack-audio-connection-kit
+		!media-sound/jack2
+	)
 	pipewire-alsa? (
-		>=media-libs/alsa-lib-1.1.7
+		>=media-libs/alsa-lib-1.1.7[${MULTILIB_USEDEP}]
 		|| (
 			media-plugins/alsa-plugins[-pulseaudio]
 			!media-plugins/alsa-plugins
 		)
 	)
-	!pipewire-alsa? ( media-plugins/alsa-plugins[pulseaudio] )
+	!pipewire-alsa? ( media-plugins/alsa-plugins[${MULTILIB_USEDEP},pulseaudio] )
 	systemd? ( sys-apps/systemd )
 	v4l? ( media-libs/libv4l )
 "
@@ -85,7 +91,8 @@ DEPEND="${RDEPEND}"
 # Ditto for DEPEND
 #	>=dev-util/vulkan-headers-1.1.69
 
-DOCS=( {README,INSTALL}.md NEWS )
+# It appears that meson.eclass with multilib-minimal.eclass breaks dodoc itself
+#DOCS=( {README,INSTALL}.md NEWS )
 
 RESTRICT="!test? ( test )"
 
@@ -95,7 +102,20 @@ PATCHES=(
 
 # limitsdfile related code taken from =sys-auth/realtime-base-0.1
 # with changes as necessary.
-limitsdfile=40-${PN}.conf
+limitsdfile=00-${PN}.conf
+
+# Copied with modifications from media-sound/pulseaudio-9999 from b.g.o (has sign-offs)
+meson_native_feature() {
+	echo "-D${2:-${1}}=$(multilib_native_usex ${1} enabled disabled)"
+}
+
+meson_native_enabled() {
+	if multilib_is_native_abi; then
+		echo "-D${1}=enabled"
+	else
+		echo "-D${1}=disabled"
+	fi
+}
 
 src_prepare() {
 	default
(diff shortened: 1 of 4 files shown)


pinkflames:
@Whissi - since you have been bumping PW lately, this PR is probably relevant to you. ;)

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -19,7 +19,8 @@ HOMEPAGE="https://pipewire.org/"
 
 LICENSE="LGPL-2.1+"
 SLOT="0/0.3"
-IUSE="aac aptx bluetooth doc extra gstreamer jack-client ldac pipewire-alsa systemd test v4l"
+IUSE="aac aptx bluetooth doc extra gstreamer jack-client jack-sdk ldac pipewire-alsa systemd test
+v4l"
```
Whissi, line 23: Keep a long line or break into multiple lines with
```bash
IUSE="aac ...."
IUSE+=" v4l"
```

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -42,9 +44,10 @@ BDEPEND="
 "
 RDEPEND="
 	media-libs/alsa-lib
-	sys-apps/dbus
+	sys-apps/dbus[${MULTILIB_USEDEP}]
 	sys-libs/ncurses[unicode]
```
Whissi, line 48: No multilib needed for ncurses?

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -42,9 +44,10 @@ BDEPEND="
...
-	virtual/libudev
+	virtual/libintl[${MULTILIB_USEDEP}]
+	virtual/libudev[${MULTILIB_USEDEP}]
 	bluetooth? (
 		aac? ( media-libs/fdk-aac )
```
Whissi, line 52: No multilib needed for fdk-aac?

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -85,7 +92,7 @@ DEPEND="${RDEPEND}"
 # Ditto for DEPEND
 #	>=dev-util/vulkan-headers-1.1.69
 
-DOCS=( {README,INSTALL}.md NEWS )
+#DOCS=( {README,INSTALL}.md NEWS )
```
Whissi, line 96: Probably leftover?

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -97,6 +104,29 @@ PATCHES=(
 # with changes as necessary.
 limitsdfile=40-${PN}.conf
 
+# Copied with modifications from media-sound/pulseaudio-9999 from b.g.o (has sign-offs)
+_meson_native_feature() {
+	echo "-D${2:-${1}}=$(multilib_native_usex ${1} enabled disabled)"
+}
+
+# Copied with modifications from sys-apps/systemd/systemd-9999
+meson_multilib() {
```
Whissi: Should be converted into `meson_use <USE flag> [option name]` style function.

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -107,90 +137,101 @@ src_prepare() {
...
 }
 
-src_install() {
+multilib_src_install() {
 	meson_src_install
 
-	insinto /etc/security/limits.d
-	doins ${limitsdfile}
+	if multilib_is_native_abi; then
+#		dodoc {README,INSTALL}.md NEWS
```
Whissi: Leftover and we have `multilib_src_install_all` for this.

virtual/jack/jack-1.ebuild
```diff
@@ -13,4 +13,5 @@ RDEPEND="
 	|| (
 		media-sound/jack2[${MULTILIB_USEDEP}]
 		media-sound/jack-audio-connection-kit[${MULTILIB_USEDEP}]
+		media-video/pipewire[${MULTILIB_USEDEP},jack-sdk(-)]
```
Whissi, line 16: Why `(-)`?

Whissi (changes requested):
Changes from "Adds meson.eclass style functions for multilib" should be squashed into first commit.

virtual/jack/jack-1.ebuild
pinkflames: This indicates that if the jack-sdk IUSE is absent from the ebuild (such as from versions before 0.3.26 that first made it available), then the package manager must treat it as if it was disabled when resolving dependencies. (+) would mean it's enabled if absent. Without any of them it may produce unsolvable reverse dependencies and is a CI failure. Thanks to @a17r for explaining this.

media-video/pipewire/pipewire-9999.ebuild
pinkflames: Alright, will do the former here.

media-video/pipewire/pipewire-9999.ebuild
pinkflames: Currently ncurses is only used for pw-top tool, which will be overwritten if built for non-native ABIs anyway. And if multilib for it was missing, it should merely silently disable building of pw-top from what I recall about the build system:
$ grep ncurses /var/db/pkg/media-video/pipewire-9999/NEEDED
/usr/bin/pw-top libpipewire-0.3.so.0,libncursesw.so.6,libtinfow.so.6,libc.so.6

media-video/pipewire/pipewire-9999.ebuild
pinkflames: Non-native ABIs have completely disabled Bluetooth, since we only need the jack libraries and, if pipewire-alsa is enabled, also the ALSA plugin, none of which need AAC or any other audio codec. I'm not entirely sure about future but I doubt libraries directly host SPA plugins, so unless things are re-architected, this will stay true even if applications start to directly use PipeWire rather than JACK/ALSA/PA APIs.

media-video/pipewire/pipewire-9999.ebuild
pinkflames: If you mean that installing {README,INSTALL}.md NEWS is a left-over from before good integration, I'm more than willing to agree and officially stop installing them. Just say so and they'll be gone. If you meant the commented out code - the dependencies are still valid, if we decided to turn back on the features they correspond to - they just make no sense to ship. In principle we could remove them completely.

media-video/pipewire/pipewire-9999.ebuild
pinkflames: I believe the current meson_native_feature and meson_native_enabled should do that. Please let me know if that's not it.

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -263,6 +262,8 @@ pkg_postinst() {
 		elog
 		ewarn "${EROOT}/etc/xdg/autostart/pipewire.desktop has been installed. Users of XDG"
 		ewarn "compliant desktops on OpenRC must not manually start pipewire anymore!"
+		ewarn "Users wishing to use PulseAudio must remove the file manually and add the"
+		ewarn "path above to the INSTALL_MASK variable in their make.conf"
```
Whissi, line 266: No, `INSTALL_MASK` is not a feature we will recommend in `elog` to use.

1. `elog` should only show the absolute necessary information for user to get a package into a usable state.
2. `elog` message should not repeat with every update, i.e. only show information when user interaction is required. If we would show same `elog` all the time user would ignore them and probably miss new important information once something has changed.

virtual/jack/jack-1.ebuild
Whissi: But this is wrong: If `USE=jack-sdk` for `media-video/pipewire` is not available, that pipewire ebuild does **not** satisfy `virtual/jack`.

media-video/pipewire/pipewire-9999.ebuild
Whissi: No, the other way. We need to install them.

virtual/jack/jack-1.ebuild
pinkflames: I fail to see how that would be wrong - virtual/jack only accepts media-video/pipewire[jack-sdk]. If its either disabled [-jack-sdk] or missing, then such configurations will not satisfy the dependency and instead pull in one of the existing JACK servers. In fact, got hit by just that when doing the 0.3.27 bump - no jack-sdk IUSE for that ebuild led to refusal by Portage to install such a package. As expected.

ionenwks: While that's true (this isn't `(+)`), I don't see the point of `(-)` here. Just `[...,jack-sdk]` is enough, portage already knows that a missing flag doesn't sastisfy a requested flag. This just makes it look more confusing.

afaik `(-)` only makes sense when disabling with `[-flag(-)]` given a missing flag wouldn't satisfy -flag otherwise.

pinkflames: That makes sense. Thanks, Ionen. I guess I misunderstood Andreas and why CI was failing.

jkroonza:
After some initial difficulty managed to get this operational, and at a minimum I'm no worse off than alsa + (jack | pulseaudio) or alsa + jack + pulseaudio.

So check from me.  Albeit, still unable to get sensible audio configs up and running, but that's a different issue.

IMHO, merge this and go.  And possibly back-port to the newest ~ version.

On the upside, this uses ~ 15% of the CPU compared to pulseaudio when using the pulseaudio interface and <10% when using the JACK interface.  With none of the noticable (using mplayer) audio latency, nor any of the stutter issues with pulseaudio being reproducable.

Still nowhere near a perfect solution for audio on Linux, but IMHO a small step in the right direction.

jkroonza:
Will get onto #pipewire and direct upstream regarding the issues I have with pipewire itself.

Whissi:
Landed via commit ba2b951973e0f5ab1e1647b8ac969d9e147cb6ae and commit e3978839cc6da41c74e9c65ced13fdf2b7c0f3db.

virtual/jack/jack-1.ebuild
Whissi: Without "(+)" or "(-)", we would get MissingUseDepDefault error for new virtual because <media-video/pipewire-0.3.27 has no USE=jack-sdk. Specifying "jack-sdk(-)" like you did is correct: Like written in devmanual (https://devmanual.gentoo.org/general-concepts/dependencies/#use-dependency-defaults), it sets a default in case the flag does not exist in the target package.

Sorry!

PR #20462, opened 2021-04-19
media-fonts/iosevka: version bump to 6.1.3

epsilon-0:

---

This time I'm adding a ttc for the default monospace variant, this ttc contains 54 ttf files.
Building is basically impossible, I haven't invested time to figure out how to work nodejs crap and stop it from downloading other cruft.


juippis (commented):
It looks like 6.0.1 is available too.

Can you please provide the file list for both versions, 1.14.1 and 6.0.1.

Although at this point, the version difference seems to be so huge, we maybe shouldn't focus on what's working with 1.xx at all.

epsilon-0:
Updated to 6.1.3.
The only font file installed by the ebuild is `iosevka.ttc` which bundles all weights together (for the default monospace variant). This is the recommended method by upstream.

juippis (approved):
Well if I upgrade from 1.14.1 to 6.1.3, the font I selected looks different. But there are a couple of version in-between, maybe it has changed...

PR #20467, opened 2021-04-20
Changed website because of 404 error on current homepage.

maciderio:


juippis (commented):
Hey, 

if you want to be credited, then please sign-off the commit with your legal name. It is a requirement for all commits to our repository,
https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin

Please let me know if this is not okay for you, and I'll fix the issue by myself ASAP.

maciderio:
I'm sorry, I hadn't noticed your request.

I would rather not provide my legal name at this stage. I would be very happy for you to use the change as seen fit.

Is a pull request the best way to propose such changes to the package repository ? If so, I'm wondering if requests can be filed here or would they better go to gitweb.gentoo.org ?

Thank you and all the best.

PR #20471, opened 2021-04-20
dev-libs/libxdg-basedir: Version bump 1.2.3.

tastytea:
Changes:
* Removed patch since it is in 1.2.3
* Added myself as proxied maintainer


juippis (approved):

dev-libs/libxdg-basedir/libxdg-basedir-1.2.3.ebuild
```diff
@@ -0,0 +1,47 @@
...
+	fi
+}
+
+src_install() {
+	emake DESTDIR="${D}" install
+
+	use doc && local HTML_DOCS=( doc/html/. )
+	einstalldocs
+
+	find "${ED}" -type f -name '*.la' -delete
```
juippis, line 46: Missing `|| die` here, but I'll add it, thanks!

PR #20473, opened 2021-04-20
net-analyzer/zabbix: Added reload function in the zabbix-server init script

zegeek:
As zabbix-server process supports configuration reload in runtime, it would be great to support with through the openrc commandline utility.

### DIFF
--- a/net-analyzer/zabbix/files/zabbix-server.init
+++ b/net-analyzer/zabbix/files/zabbix-server.init
@@ -17,3 +17,7 @@ start_pre() {
 	checkpath -q -d -m 0750 -o root:zabbix /etc/zabbix
 	checkpath -q -f -m 0640 -o root:zabbix /etc/zabbix/*.conf
 }
+
+reload() {
+    /usr/sbin/zabbix_server -c /etc/zabbix/zabbix_server.conf -R config_cache_reload
+}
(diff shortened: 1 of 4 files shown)


fordfrog:
@zegeek nice, thanks! though we will have to do revbumps on relevant ebuilds to get it emerged, this way it would be installed only on new installs and bumps.

zegeek:
@fordfrog True. To be honest, this is my first time working with gentoo package manager. Would you point me out to a good documentation on how to properly do ebuilds revision?

fordfrog:
@zegeek in this case, it's just copying the ebuilds of the latest versions of each main version and adding `-r1` at the end of the ebuild name. if you'd copy a stable ebuild, you'd have to drop it to unstable, the easiest way is to use `ekeywords ~all ...ebuild` (part of `app-portage/gentoolkit`). you can find more info [here](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/). it's also a good habit to run `repoman full` (`app-portage/repoman`) and `pkgcheck scan` (`dev-util/pkgcheck`) in the ebuild dir to check for any issues.

zegeek:
@fordfrog Thanks for the tips. I added the ebuild revision and now everyone can benefit from this "little" contribution :smiley:

fordfrog:
@zegeek thank you. there are still two things that should be resolved.
1. your commits are missing `signed-off-by] line which is required for all commits, see this comment: https://github.com/gentoo/gentoo/pull/20473#issuecomment-823397564
2. `zabbix-5.0.10`, though not lts, should also be revbumped to get your improvement

PR #20477, opened 2021-04-20
net-mail/automx2: Bump to version 2021.4

rseichter:
Upstream bugfix release.


rseichter:
@thesamesam Pinging you because mschiff is currently circumnavigating the globe (AFAIK) and because you have processed pull requests for automx2 in the past.

rseichter:
@thesamesam Polite bump. This PR has now been open since mid-April.

net-mail/automx2/automx2-2021.4.ebuild
```diff
@@ -0,0 +1,49 @@
...
+
+python_test() {
+	export AUTOMX2_CONF="tests/unittest.conf"
+	${EPYTHON} -m unittest discover tests/ || die "Tests failed with ${EPYTHON}"
+}
+
+python_install_all() {
+	if use doc; then
+		DOCS="*.adoc doc/*.adoc contrib/*sample.conf"
+		HTML_DOCS="doc/*.html doc/*.svg"
```
juippis, line 42: You should prefix these with `local` and transform to an array form, or use dodoc here respectively.

net-mail/automx2/automx2-2021.4.ebuild
```diff
@@ -0,0 +1,49 @@
...
+	fi
+	sed -e "s/@EPYTHON@/${EPYTHON}/" "${FILESDIR}/init" | newinitd - "${PN}"
```
juippis, line 44: This could also be reworked and made sure `sed` call doesn't fail with `|| die`. But... with distutils, I think having `EPYTHON="python"` fits more to the eclass spirit. The default interpreter is then selected via `eselect python`. Or you could use `DISTUTILS_SINGLE_IMPL`.

net-mail/automx2/automx2-2021.4.ebuild
```diff
@@ -0,0 +1,49 @@
...
+inherit distutils-r1
+
+DESCRIPTION="Email client autoconfiguration service"
+HOMEPAGE="https://automx.org/"
+SRC_URI="https://github.com/rseichter/automx2/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="doc"
```
juippis, line 17: Since it doesn't look like anything is required to generate those doc files, how much size do they take after being installed? Maybe this USE flag is redundant and they could be installed unconditionally?

juippis (approved):
Few nits / points to think about, but this works for me. Sorry it took a while after your ping.

PR #20478, opened 2021-04-20
x11-terms/kitty and kitty-terminfo bump to 0.20.1

end222:
x11-terms/kitty and kitty-terminfo bump to 0.20.1.


x11-terms/kitty/kitty-0.20.1.ebuild
```diff
@@ -0,0 +1,110 @@
...
+	x11-libs/libXi
+	x11-libs/libXinerama
+	x11-libs/libxkbcommon[X]
+	x11-libs/libXrandr
+	x11-terms/kitty-terminfo
+	wayland? (
+		dev-libs/wayland
+		>=dev-libs/wayland-protocols-1.17
+	)
+	$(python_gen_cond_dep 'dev-python/importlib_resources[${PYTHON_USEDEP}]' python3_6)
```
juippis, line 47: This can probably be removed with PYTHON_COMPAT updated to not have 3.6 anymore?

juippis (approved):
Tests fail, but kitty works. Thanks!

x11-terms/kitty-terminfo/kitty-terminfo-0.20.1.ebuild
```diff
@@ -0,0 +1,41 @@
+# Copyright 1999-2020 Gentoo Authors
```
juippis, line 1: Although you should use repoman to commit so it updates these automatically.

PR #20484, opened 2021-04-21
dev-java/aspectj: bump to 1.9.6

vaukai:
Closes: https://bugs.gentoo.org/601608
Bug: https://bugs.gentoo.org/784275


fordfrog:
@vaukai thanks! please also note that i switched the package to the latest `asm` slot, see b48e7407acca71b9391e04c358276b5fcee6fe3b. it's always good to check whether the package is compatible with newer slots so that we can later get easily rid of the old slots.

PR #20487, opened 2021-04-21
dev-libs/libuv: Bump to 1.41.0

jsmolic:
* Re-enable test suite
* Drop static-libs support
* Add live ebuild template


dev-libs/libuv/libuv-9999.ebuild
```diff
@@ -1,49 +1,67 @@
...
 
 multilib_src_configure() {
-	local myeconfargs=(
+	ECONF_SOURCE="${S}" \
+	econf \
+		--disable-static \
 		cc_cv_cflags__g=no
-		$(use_enable static-libs static)
-	)
-	ECONF_SOURCE="${S}" econf "${myeconfargs[@]}"
```
a17r, line 37: I much prefer `myeconfargs` array style, please keep existing style.

dev-libs/libuv/libuv-9999.ebuild
```diff
@@ -1,49 +1,67 @@
...
 SLOT="0/1"
-KEYWORDS=""
-IUSE="static-libs"
-RESTRICT="test"
 
 BDEPEND="
 	sys-devel/libtool
-	virtual/pkgconfig
-"
+	virtual/pkgconfig"
```
a17r, line 24: Unnecessary change, please keep existing style.

dev-libs/libuv/libuv-9999.ebuild
```diff
@@ -1,49 +1,67 @@
...
+			src/unix/darwin-proctitle.c || die
+		sed -i -e '/int uv_if_indextoname/i\
+		static size_t strnlen(char *x, size_t len) {\
+			char *r = memchr(x, '"'"'\\n'"'"', len);\
+			if (r == NULL)\
+				return len;\
+			else\
+				return r - x;\
+		}' \
+			src/unix/getaddrinfo.c || die
```
a17r, line 45: This is ugly to say the least. Please turn it into a patch.

a17r (changes requested):
lgtm except for the nitpicks

dev-libs/libuv/libuv-9999.ebuild
jsmolic: Sure, no problem

dev-libs/libuv/libuv-9999.ebuild
jsmolic: Done

dev-libs/libuv/libuv-9999.ebuild
jsmolic: Agreed, done

a17r (approved):
Thanks, I can merge this. For the future, I would recommend working on the live ebuild, then do a clean version bump; so the commit log would be something like:
- add release/live switch
- re-enable test suite
- drop static-libs
- 1.41.0 version bump

This makes an isolated look at the live ebuild git log a lot more productive than a bunch of 'sync live ebuild' summaries. If a 9999 version is present, we will always use it to create the next release version, not the other way round.

Patches, unless really trivial, should always contain git format-patch style header with a short description what they do.

Package builds, tests pass.

PR #18551, opened 2020-12-07
sys-apps/raspberrypi-sys-mods: new package

crabbedhaloablution:
This package sets up permissions on i2c, video, spi, gpio and input,
handles setting up symlinks /dev/serial{0,1} on serial devices, needed
by pi-bluetooth, blacklists unwanted drivers for the internal
ethernet chip, sets up "hotplug" on i2c and spi devices.
Included is also a systemd service to properly turn off backlight
for devices connected to the ribbon cable display connection.

Closes: https://bugs.gentoo.org/758917
Package-Manager: Portage-3.0.11, Repoman-3.0.2
Signed-off-by: Peter Alfredsen <crabbedhaloablution@icloud.com>

PR #20492, opened 2021-04-21
app-misc/ddcutil: GLEP81 and version bump

VTimofeenko:
This PR aims to provide the following fixes for app-misc/ddcutil:

* Fix GLEP 81 compliance
* Bump EAPI from 6 to 7
* Bump ddcutil version to 1.1.0

PR is marked as "work in progress" because at the time of writing there
is a conflicting PR #18551 in which acct-group/i2c may be introduced.

Depending on how that conflict will be resolved, the acct-group/i2c may
be added to this PR.

See bug https://bugs.gentoo.org/781557 for more information.

Bug: https://bugs.gentoo.org/781557


VTimofeenko:
Rebased onto master after acct-group/i2c was added in 8ee67694aefba13bdce3a5d20bd5a2acf2a5248c and removed the WIP flag.

app-misc/ddcutil/ddcutil-0.9.1.ebuild
```diff
@@ -27,9 +27,17 @@ RDEPEND="dev-libs/glib:2
 	X? (
 		x11-libs/libXrandr
 		x11-libs/libX11
+	)
+"
+RDEPEND="${COMMON_DEPEND}
+	usb-monitor? (
```
robbat2, line 33: Should `usb-monitor` be inside the `user-permissions` block of RDEPEND? Otherwise `USE='usb-monitor -user-permissions'` ends up adding the `video` group now when the old code did not trigger `enewgroup video` in that case.

Is `usb-monitor` even meaningful without `user-permissions`? If it's not meaningful, probably also add a `REQUIRED_USE` to block the case of `USE='usb-monitor -user-permissions'`

robbat2 (changes requested):

app-misc/ddcutil/ddcutil-0.9.1.ebuild
VTimofeenko: Thanks for the review!

The way I understand https://www.ddcutil.com/usb/, `usb-monitor -user-permissions` may make sense, since that would still allow managing such monitor by user that is in video group. I.e. if a monitor is managed through USB, there may not be a need to manage it through I2C. Sadly, none of the monitors I have work with USB, so I cannot test this.

I do agree that the ebuild in the form that I submitted would exhibit inconsistent behavior in regards to those useflags. 

If my understanding above is correct, I could flatten the nested conditionals and make the checks independent.

I would also like to clarify, should I add revisions to the updated versions of ddcutil ebuilds?

AndrewAmmerlaan:

PR #20494, opened 2021-04-22
dev-java/avalon-logkit: EAPI 7, min java 1.8

vaukai:
Not updating to the latest upstream version since this package should be last-rited if possible in the near future.


fordfrog:
@vaukai thanks! merged straight to stable, dropped old.

PR #20497, opened 2021-04-22
dev-java/bndlib: bump to 5.3.0

vaukai:
Bug: https://bugs.gentoo.org/676116

@fordfrog 
Here again your help is needed.  I would like to have `java-ebuilder` rewrite and move the ebuild to `java-pkg-simple`.
From version 2.1.0 I could identify `S` as `bnd-5.3.0.REL/biz.aQute.bndlib/`. 
But from those non less than 85 `pom.xml`files in the download it's hard to find the one to use. Can you help?

Also, there was trouble downloading the sources for dev-java/osgi-enterprise-api.  
Without the `pom.xml` file I could not check :-(  
Bug filed: https://bugs.gentoo.org/785043


fordfrog:
unfortunately this project uses `gradle` to build, so no `pom.xml`. and i'm not familiar with `gradle` so cannot help you off-head. anyway, what i would do is that

1.  i would try to build the project using `gradle` as upstream has documented,
2. then i would investigate what jar files were created and what do they contain,
3. then i would try to mimic that using `java-pkg-simple`
4. and finally i would check the resulting jar file against the one built by `gradle` using `japi-compliance-checker` and `pkgdiff`

vaukai:
> unfortunately this project uses `gradle` to build, so no `pom.xml`. and i'm not familiar with `gradle` so cannot help you off-head. anyway, what i would do is that
> 
>     1. i would try to build the project using `gradle` as upstream has documented,
> 
>     2. then i would investigate what jar files were created and what do they contain,
> 
>     3. then i would try to mimic that using `java-pkg-simple`
> 
>     4. and finally i would check the resulting jar file against the one built by `gradle` using `japi-compliance-checker` and `pkgdiff`

Sorry, guess I'd better leave this to you or someone else and concentrate on: 
```
git --no-pager grep -Elo "virtual/(jre|jdk)-1.[2-5]" | sort | grep '^dev-java' | wc -l
110
```
@fordfrog 
Could you mark https://bugs.gentoo.org/676116 blocking https://bugs.gentoo.org/697014, please?

fordfrog:
@vaukai sure, thank you for what you do :-)

PR #20501, opened 2021-04-22
media-sound/mumble: Update live ebuild

DarkDefender:
Mumble now depends on the cpp lib `poco` for their new plugin system.


DarkDefender:
Don't know if I can assign you, so I'll ping you instead @Polynomial-C

Polynomial-C:
Please use a more descriptive commit message. Something like
```
Add dev-libs/poco to RDEPEND
```

DarkDefender:
Done :)

PR #20502, opened 2021-04-22
net-im/telegram-desktop: drop old

mid-kid:
- net-im/telegram-desktop: drop old
- media-libs/libtgvoip: drop old
- media-libs/tg_owt: drop old


mid-kid:
@gyakovlev should I tag you specifically?

gyakovlev:
@mid-kid yeah, just @ me if you need me.
I have 25118 unread emails from all github activities and not tracking it actively, but mentioning me directly gets the message in my filtered inbox =)

gyakovlev (approved):

PR #20503, opened 2021-04-22
dev-haskell/{hakyll, hlint}: remove unused patch/file

mm1ke:


trofi:
Pulled as is. Thank you!

PR #20509, opened 2021-04-22
sci-mathematics/cgal: version bump to 5.2.1

epsilon-0:
move old ebuilds to header only mode


sci-mathematics/cgal/cgal-5.2-r1.ebuild
```diff
@@ -57,7 +55,7 @@ src_configure() {
 	local mycmakeargs=(
 		-DCGAL_INSTALL_LIB_DIR="$(get_libdir)"
 		-DCGAL_INSTALL_CMAKE_DIR="$(get_libdir)/cmake/CGAL"
-		-DCGAL_HEADER_ONLY=OFF
+		-DCGAL_HEADER_ONLY=ON
```
SoapGentoo, line 58: why? and why are you doing this in one commit?

SoapGentoo (changes requested):

sci-mathematics/cgal/cgal-5.2-r1.ebuild
epsilon-0: It's the default setting, we have no specific need for making it into a library. From https://doc.cgal.org/latest/Manual/installation.html
> Since CGAL version 5.0, CGAL is header-only be default, which means that there is no need to compile CGAL or its libraries before it can be used.
>
> This page is for advanced users that have a good reason to still use the old way. If this is not your case, head over back to the page Getting Started with CGAL. 

I'll split it into bump and header only changes.

SoapGentoo: Thanks for clearing that up. Here's the plan: Change it for 5.2.1 only, and in due course the old ebuilds will be washed out of the repo. I don't want to change stable ebuilds like this.

epsilon-0: done.

juippis (approved):
```
[ebuild   R    ] sci-mathematics/cgal-5.2.1:0/14::gentoo  USE="examples* gmp* mpfi qt5* -doc -ntl*" 0 KiB

Total: 1 package (1 reinstall), Size of downloads: 0 KiB

>>> Verifying ebuild manifests
>>> Emerging (1 of 1) sci-mathematics/cgal-5.2.1::gentoo
>>> Installing (1 of 1) sci-mathematics/cgal-5.2.1::gentoo
>>> Jobs: 1 of 1 complete                           Load avg: 1.92, 2.19, 1.07

 * Messages for package sci-mathematics/cgal-5.2.1:
 * Log file: /var/tmp/portage/vbslogs/build/sci-mathematics/cgal-5.2.1:20210505-060916.log

 * QA Notice: One or more compressed files were found in docompress-ed
 * directories. Please fix the ebuild not to install compressed files
 * (manpages, documentation) when automatic compression is used:
 * 
 *   /usr/share/doc/cgal-5.2.1/examples/Mesh_3/data/liver.inr.gz
```
```
# du -csh /usr/share/doc/cgal-5.2.1/examples/
21M	/usr/share/doc/cgal-5.2.1/examples/
21M	total
my-test-container-snap-20509 ~ # du -csh /usr/share/doc/cgal-5.2.1/demo/
11M	/usr/share/doc/cgal-5.2.1/demo/
11M	total
my-test-container-snap-20509 ~ # du -csh /usr/share/doc/cgal-5.2.1/html/
250M	/usr/share/doc/cgal-5.2.1/html/
250M	total
```
(Just to confirm I agree with having USE flags for these even though they don't pull external deps and don't apparently require any build-system steps.)

sci-mathematics/cgal/cgal-5.2-r1.ebuild
bircoph: This change breaks other packages in the tree which need this library:
https://bugs.gentoo.org/791811

epsilon-0: These packages should be patched to support the new CGAL.
Or we can restrict the CGAL to older version.
Will do this soon.

bircoph: > These packages should be patched to support the new CGAL.
> Or we can restrict the CGAL to older version.
> Will do this soon.

You should add a USE flag to control whether library is needed or not and let users to decide. This will also allow to keep current packages working without patching other than dependency update.

juippis: @epsilon-0 thanks for looking into it. Please either ping me in the PR with fixes to related packages, or and preferably, link the commits to this bug:
https://bugs.gentoo.org/791811

epsilon-0: Indeed, it seems like the simplest solution is to allow a `shared` USE flag. I've made a PR to do a version bump and add the USE flag - https://github.com/gentoo/gentoo/pull/21275
I am unsure what are all the packages that need the library, so I haven't touched them.

Overall, the better strategy would be to tell upstreams of these packages to work with the newer CGAL.

PR #20514, opened 2021-04-23
www-client/qutebrowser: add dbus pyqt5 dep

GuillaumeSeren:


steils:
Thanks for the PR. But why are you changing the revision, is it really necessary?
Also I think it's a good idea to keep the old version of qutebrowser in the tree for a while, if it doesn't have security issues or critical bugs.

ionenwks:
> But why are you changing the revision, is it really necessary?

It's a RDEPEND change that may not be properly propagated (dynamic-deps are unreliable, and qutebrowser is broken without it), and this isn't a package that takes time to "build" making revbumps cheap.

I don't see a reason not to revbump here.

GuillaumeSeren:
Hello,
thank you for the feedback.

@steils ,

> Thanks for the PR. But why are you changing the revision, is it really necessary?

I wasn't sure, but I checked the devmanual on this (https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html),
and it seemed to be a 'changes that warrant a new revision' (adding a missing runtime dependency to one of the existing blocks).

> Also I think it's a good idea to keep the old version of qutebrowser in the tree for a while, if it doesn't have security issues or critical bugs.

Yes I was going that way at first, but since there is no stable version yet,
we may just stick to the last release, I am open to change it if you want.

steils:
I'm not sure if adding the dependency changes already installed files of qutebrowser therefore affecting the runtime, since the Python module is loaded dynamically. Too lazy to check it. Not a big deal, let's keep the revision bump. I'll drop the patch that drops the old version, and merge.

GuillaumeSeren:
@steils ,

> I'm not sure if adding the dependency changes already installed files of qutebrowser therefore affecting the runtime, since the Python module is loaded dynamically (but I'm too lazy to check it).

It should be only a runtime deps as stated here (https://github.com/qutebrowser/qutebrowser/releases/tag/v2.2.0),

> Not a big deal, let's keep the revision bump. I'll drop the patch that drops the old version, and merge.

Yes cool thank you,
also we should open a stabilization bug for the 2.1.1 to get a stable version.

> Not a big deal, let's keep the revision bump. I'll drop the patch that drops the old version, and merge.

PR #20517, opened 2021-04-23
net-wireless/wireless-regdb: Bump to 20210421, Remove old 20190301, 20190603

jsmolic:


ZeroChaos-:
thanks for the poke, accepted.

Just for future reference, keep the keywords for this package, it's a binary shipped from upstream, we don't use the standard stabilization process unless actually changing more than the shipped binary.

PR #20518, opened 2021-04-23
media-fonts/nerd-fonts: icon patched versions of popular fonts

epsilon-0:


epsilon-0:
@juippis @jonasstein 
anyone interested in getting these fonts in gentoo?

media-fonts/nerd-fonts/nerd-fonts-2.1.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+	ProFont
+	ProggyClean
+	RobotoMono
+	ShareTechMono
+	SourceCodePro
+	SpaceMono
+	Terminus
+	Tinos
+	Ubuntu
+	UbuntuMono
```
juippis, line 63: Can we combine these under `ubuntu` IUSE? Just update the description in *metadata.xml* to include mono typeset too.

It's a bit more text in the ebuild, but easily doable with similar `IUSE+=` what you use already.

media-fonts/nerd-fonts/nerd-fonts-2.1.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+	InconsolataGo
+	InconsolataLGC
+	Iosevka
+	JetBrainsMono
+	Lekton
+	LiberationMono
+	Meslo
+	Monofur
+	Monoid
+	Mononoki
```
juippis, line 49: Are these related to each other?

media-fonts/nerd-fonts/nerd-fonts-2.1.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+	Gohu
+	Hack
+	Hasklig
+	HeavyData
+	Hermit
+	iA-Writer
+	IBMPlexMono
+	Inconsolata
+	InconsolataGo
+	InconsolataLGC
```
juippis, line 41: I'd combine these as well.

media-fonts/nerd-fonts/nerd-fonts-2.1.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+	BitstreamVeraSansMono
+	CascadiaCode
+	CodeNewRoman
+	Cousine
+	DaddyTimeMono
+	DejaVuSansMono
+	DroidSansMono
+	FantasqueSansMono
+	FiraCode
+	FiraMono
```
juippis, line 30: And these.

media-fonts/nerd-fonts/nerd-fonts-2.1.0.ebuild
```diff
@@ -0,0 +1,78 @@
+# Copyright 2019-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit font
+
+DESCRIPTION="Icon patched versions of popular fonts"
+HOMEPAGE="https://nerdfonts.com/"
```
juippis, line 9: Please add the Github page as well.

media-fonts/nerd-fonts/nerd-fonts-2.1.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+
+KEYWORDS="~amd64 ~arm64 ~x86"
```
juippis, line 11: I know these are fonts, but I can't commit with `~arm64` around as I'm not part of that arch team. Let's follow the procedure.

juippis (commented):
Disclaimer: I really dislike the amount of USE flags this has. Although we can argue we're not talking about small file policy in here, this is still an awful user experience. But acceptable.

Question: What would happen if I emerged this with `USE="-*"`, or even with default IUSE set without editing my *package.use*? Would you consider adding some core set always to be installed, or at least default some USEs on with `+`?

Question: Why aren't we installing the *10-nerd-font-symbols.conf* fontconfig file? I kind of understand due to these USEs it might be configured wrongly... but then again, is it a bit too much work to generate our own fontconfig file with the USEs we have enabled?

Nostalgia rambling: Fun fact, I was the first one to write an ebuild for nerd-fonts when a first release, 1.0.0 was made. https://github.com/juippis/jps-gentoo/commit/441f9d4dcd32e80b90ed4fbe4ac2d4b22348ffe9
(and a bit more modern ebuild, 1.1 https://github.com/juippis/jps-gentoo/commit/87aa636ab88471331b07aa271634766912512e98)
Looking at the only ebuild in https://gpo.zugaina.org/media-fonts/nerd-fonts still has the familiar `pkg_postinst` message...

I quickly dropped that from my overlay as it was pain to maintain, but bumped a couple of versions in Funtoo before leaving it to someone else's hands. Now looking at gpo.zugaina.org, looks like it's available only in one overlay anymore. It was painful to maintain, I'd like to know why a dozen other overlays also dropped it, as once it was rather popular. 

That brings me to my final question: Have you tried installing fontpatcher and getting that to work? It can be used to patch your own fonts with nerd glyphs. I'd say in Gentoo that's something useful, but if I remember correctly, it was truly painful trying to get to work, as it installed into weird directory with a weird file layout and had paths hardcoded. I'd still like you to take a look, it could be installed with USE `tools` or something.

epsilon-0:
> Disclaimer: I really dislike the amount of USE flags this has. Although we can argue we're not talking about small file policy in here, this is still an awful user experience. But acceptable.
> 
> Question: What would happen if I emerged this with `USE="-*"`, or even with default IUSE set without editing my _package.use_? Would you consider adding some core set always to be installed, or at least default some USEs on with `+`?
> 

Good point, I'll added an ||( ... )

> Question: Why aren't we installing the _10-nerd-font-symbols.conf_ fontconfig file? I kind of understand due to these USEs it might be configured wrongly... but then again, is it a bit too much work to generate our own fontconfig file with the USEs we have enabled?
> 

Totally missed this file!!!
I'll add it.

> Nostalgia rambling: Fun fact, I was the first one to write an ebuild for nerd-fonts when a first release, 1.0.0 was made. [juippis/jps-gentoo@441f9d4](https://github.com/juippis/jps-gentoo/commit/441f9d4dcd32e80b90ed4fbe4ac2d4b22348ffe9)
> (and a bit more modern ebuild, 1.1 [juippis/jps-gentoo@87aa636](https://github.com/juippis/jps-gentoo/commit/87aa636ab88471331b07aa271634766912512e98))
> Looking at the only ebuild in https://gpo.zugaina.org/media-fonts/nerd-fonts still has the familiar `pkg_postinst` message...
> 
> I quickly dropped that from my overlay as it was pain to maintain, but bumped a couple of versions in Funtoo before leaving it to someone else's hands. Now looking at gpo.zugaina.org, looks like it's available only in one overlay anymore. It was painful to maintain, I'd like to know why a dozen other overlays also dropped it, as once it was rather popular.
> 
> That brings me to my final question: Have you tried installing fontpatcher and getting that to work? It can be used to patch your own fonts with nerd glyphs. I'd say in Gentoo that's something useful, but if I remember correctly, it was truly painful trying to get to work, as it installed into weird directory with a weird file layout and had paths hardcoded. I'd still like you to take a look, it could be installed with USE `tools` or something.

I am still trying to get that to work :sob: .
Maybe it can be a separate ebuild? media-libs/nerdfont-patcher?
It needs python and what not, so I don't want to add it to this ebuild and complicate things further.

Also, I really don't want to merge the USE flags. These files are quite large, the total download was 2.1GB. So merging them makes it even larger. I'm not sure if it is a good idea.

epsilon-0:
On second thought, would it be better to have a separate ebuild for each font?
It will be easier to handle.

juippis:
> 
> Good point, I'll added an ||( ... )
> 

Nooo don't do that, AFAIK there's even a technical limit on pkgcore how many USEs you can have inside `|| ( )` (so it breaks the CI).

> 
> I am still trying to get that to work sob .
> Maybe it can be a separate ebuild? media-libs/nerdfont-patcher?
> It needs python and what not, so I don't want to add it to this ebuild and complicate things further.

Sure, if it's too complicated AND you don't use it, let's not add it here. I'd still some sort of comment to acknowledge such a tool exists, and is available upstream.

> 
> Also, I really don't want to merge the USE flags. These files are quite large, the total download was 2.1GB. So merging them makes it even larger. I'm not sure if it is a good idea.
> 

No no, it doesn't increase anything. Just do something like
```
for MY_FONT in ${MY_FONTS[@]}; do
	SRC_URI+=" ${MY_FONT,,}? ( https://github.com/ryanoasis/nerd-fonts/releases/download/v${PV}/${MY_FONT}.zip -> ${P}-${MY_FONT}.zip )"
done

SRC_URI+=" ubuntu? ( https://github.com/ryanoasis/nerd-fonts/releases/download/v${PV}/ubuntu.zip -> ${P}-ubuntu-${PV}.zip  https://github.com/ryanoasis/nerd-fonts/releases/download/v${PV}/ubuntumono.zip -> ${P}-ubuntumono-${PV}.zip"
``` 
(untested pseudo, you get the idea)
That shouldn't increase the download size at all.

epsilon-0:
@juippis there may be some hope still, I got the font-patcher to work, upstream accepted my patch - https://github.com/ryanoasis/nerd-fonts/pull/613
I'll get back to gentoo in a week and half or so.

PR #20531, opened 2021-04-25
dev-cpp/benchmark: Version bump to 1.5.3

vilhelmgray:
Closes: https://bugs.gentoo.org/785466


juippis (approved):

dev-cpp/benchmark/benchmark-1.5.3.ebuild
```diff
@@ -0,0 +1,28 @@
+# Copyright 1999-2020 Gentoo Authors
```
juippis, line 1: ...

PR #20533, opened 2021-04-25
metadata.xml: Update maintainer name

salahcoronya:
I've used and contributed to Gentoo for a long time (since 2004, at least). However my real name is not the one I've contributed under. While GLEP 76 only requires the signoff use my real name, it's confusing to contribute under one name and sign off under another, and sometimes I've gotten it wrong. So going forward I just want to use my real name everywhere in Gentoo to clear things up and avoid this error in the future.


mgorny:
Could you split this to one commit per package? The easiest way would probably be to make a fresh branch, `git cherry-pick --no-commit`, then loop over the packages.

PR #20534, opened 2021-04-25
games-roguelike/dwarf-fortress: bump to 0.47.05

ezzieyguywuf:
Updated to EAPI 7, otherwise just a simple bump.


ezzieyguywuf:
Here is a diff showing the changes from the previous version (since github diff doesn't include `--find-copies-harder`)

```diff
diff --git a/games-roguelike/dwarf-fortress/Manifest b/games-roguelike/dwarf-fortress/Manifest
index c42a144cfc7..f7f5a8d0f1f 100644
--- a/games-roguelike/dwarf-fortress/Manifest
+++ b/games-roguelike/dwarf-fortress/Manifest
@@ -1,2 +1,4 @@
 DIST df_44_12_linux.tar.bz2 12081683 BLAKE2B 2410e8da6af6a8ced45d3e95dc06e2d0f4e31dbdcd6b8d9945a761d2278532bb24c1095418aa22639eedded256b3d485fa128984e6282a491d4dc7bd9af8671b SHA512 03cd7f6800d44873b573cc5bffb12c7a246c3aba0ee8535001372df689a5179dc02592670fd87f2fd5de82421605f34152f6241307c8f7b9fb4d7b8704aed70c
 DIST df_44_12_linux32.tar.bz2 12686340 BLAKE2B bf1e1370930d91d2c9e72d3366bc79706092d1505853eb63c6160fea39822551d3276958ac6d2916b8f7825be0c054a037219210f0c2c18a731065a2dd1bb842 SHA512 5cd352acf310592c7e30442205b7fe590f421c5e3562959dcfd32a17d3d562347c2b32e3bc5835637c5550ce2c696c17b6dc910ee4cf766be07af306e83537db
+DIST df_47_05_linux.tar.bz2 12725420 BLAKE2B 80987f05de2db2796a53cc8767ee7aa5c78674dff725efc0dafe07dc4118e56c7814cbf92507fe9b76a385e182405c7d6759cb6bcc942385110df1dc1a3b3af4 SHA512 9ea46fa12a80266cd09363f1aea8cdd059a5ebeefb453c4a46ffbb1115486409003c3caca95a1b6010da4040d04b781182c59e57a6cc033cd7de4507299f7534
+DIST df_47_05_linux32.tar.bz2 13487455 BLAKE2B be604dd2fd480c5609da801013dfdae26964760c0a92c4b3d2d8369800eee7ca9c1b1c2313cc67357feee3c6d425bf7f9f67cab53d2e771ab59055bc4c3329d0 SHA512 e9b9cf6995c24b403347159a35cfe4b57603bdb777c78639beb2799d9bf31317edec54fb92b17dbb90df6001b905d43285a35e9446768d12a746492b2fd34fe7
diff --git a/games-roguelike/dwarf-fortress/dwarf-fortress-0.44.12.ebuild b/games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
similarity index 91%
copy from games-roguelike/dwarf-fortress/dwarf-fortress-0.44.12.ebuild
copy to games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
index dfcfe78cf3c..bcdf4ddf825 100644
--- a/games-roguelike/dwarf-fortress/dwarf-fortress-0.44.12.ebuild
+++ b/games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
@@ -1,15 +1,15 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
 
-inherit versionator toolchain-funcs
+inherit toolchain-funcs
 
-MY_PV=$(replace_all_version_separators _ "$(get_version_component_range 2-)")
+MY_PV=$(ver_rs 0- _ "$(ver_cut 2-)")
 MY_PN=df
 MY_P=${MY_PN}_${MY_PV}
 
-DESCRIPTION="A single-player fantasy game"
+DESCRIPTION="The deepest, most intricate simulation of a world that's ever been created."
 HOMEPAGE="https://www.bay12games.com/dwarves"
 SRC_URI="amd64? ( https://www.bay12games.com/dwarves/${MY_P}_linux.tar.bz2 )
 	x86? ( https://www.bay12games.com/dwarves/${MY_P}_linux32.tar.bz2 )"
@@ -26,6 +26,7 @@ RDEPEND="media-libs/glew:0
 	sys-libs/zlib
 	virtual/glu
 	x11-libs/gtk+:2"
+
 # Yup, libsndfile, openal and ncurses are only needed at compile-time; the code
 # dlopens them at runtime if requested.
 DEPEND="${RDEPEND}
```

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
```diff
@@ -0,0 +1,85 @@
...
+	elog "System-wide Dwarf Fortress has been installed to ${gamesdir}. This is"
+	elog "symlinked to ~/.dwarf-fortress when dwarf-fortress is run."
+	elog "For more information on what exactly is replaced, see /usr/bin/dwarf-fortress."
+	elog "Note: This means that the primary entry point is /usr/bin/dwarf-fortress."
+	elog "Do not run ${gamesdir}/libs/Dwarf_Fortress."
+	elog
+	elog "Optional runtime dependencies:"
+	elog "Install sys-libs/ncurses[unicode] for [PRINT_MODE:TEXT]"
+	elog "Install media-libs/openal and media-libs/libsndfile for audio output"
+	elog "Install media-libs/libsdl[opengl] for the OpenGL PRINT_MODE settings"
```
00-matt, line 84:
```suggestion
	optfeature "text PRINT_MODE" sys-libs/ncurses[unicode]
	optfeature "audio output" "media-libs/openal media-libs/libsndfile"
	optfeature "OpenGL PRINT_MODE" media-libs/libsdl[opengl]
```

(with `optfeature.eclass`)

ezzieyguywuf: I can add this, thanks for the review.

ezzieyguywuf: Done

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
```diff
@@ -0,0 +1,84 @@
...
+
+src_prepare() {
+	rm -f libs/*.so* || die
+	sed -i -e '1i#include <cmath>' g_src/ttf_manager.cpp || die
+	default
+}
+
+src_configure() {
+	tc-export CXX PKG_CONFIG
+	CXXFLAGS+=" -D$(use debug || echo N)DEBUG"
```
thesamesam, line 52: append-cxxflags from flag-o-matic instead **but** this looks like it's actually a define, so it's CPPFLAGS anyway

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
```diff
@@ -0,0 +1,84 @@
...
+}
+
+src_compile() {
+	emake -f "${FILESDIR}/Makefile.native"
+	sed -e "s:^gamesdir=.*:gamesdir=${gamesdir}:" "${FILESDIR}/dwarf-fortress" > dwarf-fortress || die
```
thesamesam, line 57: It seems a bit odd to do this in src_compile, but I suppose if it's needed...

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
ezzieyguywuf: @thesamesam I understanding the different between `CPPFLAGS` and `CXXFLAGS` (the latter being a superset of the former).

I don't, however, understand what you mean by "looks like it's actually a define"

Would you care to elaborate on this a bit?

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
ezzieyguywuf: moved it to `src_prepare`

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
ionenwks: Suggested to change `CXXFLAGS` to `append-cxxflags` but, given `-DNDEBUG` is "actually a define" (i.e. equivalent of `#define NDEBUG`), those would go in `CPPFLAGS` (C pre-processor, not cplusplus).

So, `append-cppflags` should be tried instead (if package respects it).

ionenwks: That aside, I'm questioning why USE=debug exists here if this is all it does.

ezzieyguywuf: @ionenwks thanks for your thoughts.

If you check the latest commit, [you'll notice](https://github.com/gentoo/gentoo/pull/20534/files#diff-960d65f2ff8959ba0548e979278eee2a8505e175ddf459502f467be84fbdc506R57) I've incorporated these changes.

I've tested, and `append-cppflags` works as expected. Also, I've grepped the code, and it is littered with `#if DEBUG`.

I opted to change the logic a bit - `-DNDEBUG` is a bit of a misnomer - on the one hand, it is benign, as there aren't any checks for it in the source code. On the other hand, I prefer to just selectively add `-DDEBUG`, and do nothing otherwise

00-matt: > I opted to change the logic a bit - -DNDEBUG is a bit of a misnomer - on the one hand, it is benign, as there aren't any checks for it in the source code.

There are a handful of assertions using the assert macro in the DF codebase, defining NDEBUG would make them not do anything.

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
```diff
@@ -0,0 +1,88 @@
...
+RDEPEND="media-libs/glew:0
+	media-libs/libsdl[joystick,video]
+	media-libs/sdl-image[png]
+	media-libs/sdl-ttf
+	sys-libs/zlib
+	virtual/glu
+	x11-libs/gtk+:2"
+
+# Yup, libsndfile, openal and ncurses are only needed at compile-time; the code
+# dlopens them at runtime if requested.
```
DarthGandalf, line 31: But it'll then fail at runtime if requested and the library is not there?

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
```diff
@@ -0,0 +1,88 @@
...
+src_prepare() {
+	# for eapply_user
+	default
+
+	# We don't want to use any of the bundled libraries
+	rm -f libs/*.so* || die
+	# This file seems to be missing this import, resulting in a compile error.
+	sed -i -e '1i#include <cmath>' g_src/ttf_manager.cpp || die
+	# This ensures that our wrapper script properly respects ${EPREFIX}
+	hprefixify "${FILESDIR}/dwarf-fortress"
```
DarthGandalf, line 53: This would modify file in /var/db/repos/gentoo/...
Perhaps copy it from there first

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
ezzieyguywuf: > > I opted to change the logic a bit - -DNDEBUG is a bit of a misnomer - on the one hand, it is benign, as there aren't any checks for it in the source code.
> 
> There are a handful of assertions using the assert macro in the DF codebase, defining NDEBUG would make them not do anything.

So you think i should add it back?

Can you point to a place where you think having `-DNDEBUG` would be helpful?

What about a compromise: `add-cppflags -DDEBUG=$(use debug)` (untested)

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
ezzieyguywuf: Good question! I've actually been meaning to test this - I think it just disables the option gracefully, but I'll double check.

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
ezzieyguywuf: It gets copied [later in `src_install`](https://github.com/gentoo/gentoo/pull/20534/files#diff-960d65f2ff8959ba0548e979278eee2a8505e175ddf459502f467be84fbdc506R70).

Would it be better/preferred to first copy, then hprefixify, then move to /bin?

DarthGandalf: Consider what happens if you install the package twice.
Best case, it will add the prefix twice, and therefore fail at runtime. Worst case, it will simply fail to build because checksum of the file doesn't match anymore.
Or should these case be reversed?

ezzieyguywuf: Ah, I understand your concern.

Isn't `${FILEDIR}` already a copied version from `/var/db/repos/gento...`? i.e. isn't it typically somewhere like `/tmp/portage/games-rogueline/${P}/files`?

that's been my experience at least, and my understanding of the PMS, but please correct me if I'm wrong.

DarthGandalf: `/tmp/portage/games-rogueline/${P}/files` is often a simple symlink to it

ezzieyguywuf: Dang never mind, I guess I was wrong about this. TIL!

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
ezzieyguywuf: I think this has been resolved. Please move review over to #20721

games-roguelike/dwarf-fortress/dwarf-fortress-0.47.05.ebuild
ezzieyguywuf: I have resolved this. Please move any further reviews to #20721

ezzieyguywuf:
Closing, as I've moved this to #20721

PR #20538, opened 2021-04-25
*/*: mostly system library EAPI ports, modernisation

thesamesam:


app-doc/abs-guide/abs-guide-10.ebuild
```diff
@@ -1,27 +1,26 @@
...
-RDEPEND=""
-
-S="${WORKDIR}/abs"
+IUSE="pdf"
 
 src_unpack() {
 	unpack abs-guide-final.tar.bz2
-	use pdf && cp "${DISTDIR}"/abs-guide.pdf "${S}"
+
+	if use pdf ; then
```
telans, line 21: Why the space here before `;`?

SoapGentoo (changes requested):

app-doc/abs-guide/abs-guide-10.ebuild
thesamesam: sam style ™️, but I think I’m in the minority here

PR #20540, opened 2021-04-25
games-emulation/mednafen-1.26.1 and mednaffe-0.9.1 bump

ionenwks:
1.27.0 is out but I'll leave it alone with its big -UNSTABLE and revisit this at 1.27.1

Edit: wrt trio license, hoped to skip adding given how simple/permissive it is but [upstream](https://daniel.haxx.se/projects/trio/) does call it the "trio license" and doesn't fully match anything (debian did call it ISC-Original instead, which is "close" minus 2nd paragraph.. albeit we only have the new ISC -- have no strong opinion on this myself but went with that seemed more proper)


thesamesam:
Do we think it's a free licence? @gentoo/licenses

PR #20548, opened 2021-04-26
app-arch/xdms: fix build issues with GCC 10.

gryf:
Patch origins from Debian.


gryf:
I'll do that really quick.

thesamesam:
Cheers!

PR #20553, opened 2021-04-27
dev-java/commons-beanutils: bump to 1.9.4 

vaukai:


fordfrog:
@vaukai thanks! i was not able to fix the remaining tests so merged it as it is for now, will try to fix that later.

PR #20559, opened 2021-04-27
app-metrics/prometheus: 2.26.0: Use builtinassets tag to readd web UI.

jpds:
Closes: https://bugs.gentoo.org/777882


jpds:
@Zlogene I've tested that this works on my system

Zlogene:
@jpds can you please apply the change to the all ebuilds without promu calls? i.e. not only the latest one

jpds:
@Zlogene Sure, pushed.

PR #20561, opened 2021-04-27
sys-power/uhubctl: export pkg-config

coconutruben:
2.4.0 uhubctl honors tc-export of pkg-config. Use it here.


ZeroChaos-:
ACK, accepted, thanks!

PR #20562, opened 2021-04-27
sys-libs/pam: allow optional builds against libxcrypt

10ne1:
Glibc supports being built without its libcrypt because it will
be removed in the future and libxcrypt is the main replacement
even though packages might use openssl or other alternatives.

Therefore it is necessary to build pam against libxcrypt when
it replaces GNU libcrypt via its "system" use flag, so allow it
as an option via a newly introduced "xcrypt" pam use flag.

### DIFF
--- a/sys-libs/pam/metadata.xml
+++ b/sys-libs/pam/metadata.xml
@@ -13,6 +13,10 @@
 		provided by <pkg>sys-libs/db</pkg>) installed in /usr/lib and
 		will thus not work for boot-critical services authentication.
 	</flag>
+	<flag name="xcrypt">
+		Build pam against <pkg>sys-libs/libxcrypt</pkg> instead of GNU
+		libcrypt provided by <pkg>sys-libs/glibc</pkg>.
+	</flag>
 		</use>
 <upstream>
 	<remote-id type="github">linux-pam/linux-pam</remote-id>
--- a/sys-libs/pam/pam-1.5.1.ebuild
+++ b/sys-libs/pam/pam-1.5.1.ebuild
@@ -16,7 +16,7 @@ SRC_URI="https://github.com/linux-pam/linux-pam/releases/download/v${PV}/${MY_P}
 LICENSE="|| ( BSD GPL-2 )"
 SLOT="0"
 KEYWORDS="~alpha amd64 arm arm64 hppa ~ia64 ~m68k ~mips ppc ppc64 ~riscv ~s390 sparc x86"
-IUSE="audit berkdb debug nis selinux"
+IUSE="audit berkdb debug nis selinux xcrypt"
 
 BDEPEND="
 	dev-libs/libxslt
@@ -33,6 +33,7 @@ DEPEND="
 	berkdb? ( >=sys-libs/db-4.8.30-r1:=[${MULTILIB_USEDEP}] )
 	selinux? ( >=sys-libs/libselinux-2.2.2-r4[${MULTILIB_USEDEP}] )
 	nis? ( net-libs/libnsl[${MULTILIB_USEDEP}]
+	xcrypt? ( >=sys-libs/glibc-2.30[-crypt] sys-libs/libxcrypt[system] )
 	>=net-libs/libtirpc-0.2.4-r2[${MULTILIB_USEDEP}] )"
 
 RDEPEND="${DEPEND}"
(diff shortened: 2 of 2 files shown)


sys-libs/pam/metadata.xml
```diff
@@ -12,6 +12,12 @@
 		flag will create a PAM module that links to the Berkeley DB (as
 		provided by <pkg>sys-libs/db</pkg>) installed in /usr/lib and
 		will thus not work for boot-critical services authentication.
+	</flag>
+	<flag name="xcrypt">
+		Build pam against <pkg>sys-libs/libxcrypt</pkg> instead of GNU
+		libcrypt provided by <pkg>sys-libs/glibc</pkg>. Please note that
+		libxcrypt is a binary-compatible superset of GNU libcrypt which
+		will enventualy be removed from <pkg>sys-libs/glibc</pkg>.
```
thesamesam, line 20:
```suggestion
		will eventually be removed from <pkg>sys-libs/glibc</pkg>.
```

10ne1: Thanks, fixed.

sys-libs/pam/metadata.xml
```diff
@@ -12,6 +12,12 @@
 		flag will create a PAM module that links to the Berkeley DB (as
 		provided by <pkg>sys-libs/db</pkg>) installed in /usr/lib and
 		will thus not work for boot-critical services authentication.
+	</flag>
+	<flag name="xcrypt">
```
Zlogene, line 16: well I am not sure it will eventually be merged by glibc, + this information is not really important

sys-libs/pam/pam-1.5.1.ebuild
```diff
@@ -51,11 +51,9 @@ multilib_src_configure() {
 	# Do not let user's BROWSER setting mess us up. #549684
 	unset BROWSER
 
-	# Disable automatic detection of libxcrypt; we _don't_ want the
-	# user to link libxcrypt in by default, since we won't track the
-	# dependency and allow to break PAM this way.
-
-	export ac_cv_header_xcrypt_h=no
+	if ! use xcrypt; then
```
Zlogene, line 54: this is simply not enough, because:

a.) you must also make sure glibc is built without the crypt use flag
b.) you must pull libxcrypt if this flag is enabled

Zlogene (changes requested):

sys-libs/pam/metadata.xml
10ne1: To verify I understand correctly: you are saying you are not sure glibc will ever remove libxcrypt from its codebase? That is in contradiction with current glibc dev plans...

I can simplify this message to just the first sentence just in case glibc changes it's mind :)

sys-libs/pam/pam-1.5.1.ebuild
10ne1: Very good observations, thank you, I will add these changes.

sys-libs/pam/metadata.xml
Zlogene: @10ne1  yes, that would be nice

sys-libs/pam/pam-1.5.1.ebuild
```diff
@@ -33,6 +33,7 @@ DEPEND="
 	berkdb? ( >=sys-libs/db-4.8.30-r1:=[${MULTILIB_USEDEP}] )
 	selinux? ( >=sys-libs/libselinux-2.2.2-r4[${MULTILIB_USEDEP}] )
 	nis? ( net-libs/libnsl[${MULTILIB_USEDEP}]
+	xcrypt? ( >=sys-libs/glibc-2.30[-crypt] sys-libs/libxcrypt[system] )
```
thesamesam, line 36: I think instead depend on >=virtual/libcrypt-2:= which does this, but doesn't break on non-glibc.

thesamesam (commented):
See also https://bugs.gentoo.org/799131. I think upstream have recently improved the automagic a _bit_, and @Zlogene wanted to take a new snapshot to get those fixes.

sys-libs/pam/metadata.xml
```diff
@@ -12,6 +12,10 @@
...
```
thesamesam, line 16: Could you please fix the CI failure (rebasing may do it, actually, given we've just unmasked for the general libxcrypt migration) and also add this to the package.use.force for libxcrypt we have in profiles (it's there for USE=system already, just add this on a new line)?

thesamesam (changes requested):

Zlogene:
> See also https://bugs.gentoo.org/799131. I think upstream have recently improved the automagic a _bit_, and @Zlogene wanted to take a new snapshot to get those fixes.
> 
> EDIT: ... that said, I now realise, given we're migrating to libxcrypt anyway, all this really does is improve the automagic situation (which is good). pam[-xcrypt] will just mean it looks to use compat (.so.1) or possibly in future some other provider(s), rather than looking for any new libxcrypt API.
> 
> So, I don't think there's a reason to block merging this.

well, I wanna less efforts done here, so I will indeed take a snapshot, I am curious to see how automagic flies there and if ```xcrypt? ( >=sys-libs/glibc-2.30[-crypt] sys-libs/libxcrypt[system] )```  would be enough to link properly

Zlogene:
https://github.com/gentoo/gentoo/commit/4be3f5aff505115f6d36ed916533333678dbc207 make this PR obsolete, but thanks for the submission!

PR #20567, opened 2021-04-28
dev-java/jansi: remove unused slot

vaukai:


fordfrog:
i changed the slot in 1.11 to `:0`, eapi 7, java 1.8. and i will file stabilization bug for the remaining arch. so keeping the newer version as it is reported to be 100% compatible with the old one, older will be removed after stabilization.

PR #20574, opened 2021-04-28
media-video/obs-studio: update python support and more

Chiitoo:
I'm not entirely sure if the masking of USE="python" is too heavy-handed or not, but at the very least it should make clear that there may be issues with it.

I've personally not witnessed any, but I do believe those saying that they are there.


profiles/base/package.use.mask
```diff
@@ -6,6 +6,11 @@
 # This file is only for generic masks. For arch-specific masks (i.e.
 # mask everywhere, unmask on arch/*) use arch/base.
 
+# Jimi Huotari <chiitoo@gentoo.org> (2021-04-28)
+# Known to cause issues with some scripts, while fixing others.
```
a17r, line 10: From your description it would seem odd to mask it. Notwithstanding the patch tipping the scales in one or the other way, at least it sounds like there have been existing issues that haven't lead to the flag be masked yet.

Maybe just add an `ewarn` in `pkg_postinst` and wait for the reports to be filed?

Chiitoo: Yep, done.

Chiitoo:
Regarding the Decklink plugin, I couldn't really come up with anything for it, though I did slightly change it.

Not sure if it's any better.

PR #20579, opened 2021-04-28
dev-util/nvidia-cuda-toolkit: bump to 11.3.0

myrvogna:


amadio:
Thank you! I see that even the profilers have the executable bit set now, very nice.
For the record, I updated the dependency on `virtual/jre` to fix the CI warnings above.

PR #20584, opened 2021-04-28
media-gfx/flameshot: version bump to 0.9.0

pavel-the-best:
- add system-singleapplication useflag (whether to use bundled
  singleapplication or system one)
- update dependencies (add spdlog)

Closes: https://bugs.gentoo.org/777414
Signed-off-by: Pavel Kalugin <pavel@pavelthebest.me>


media-gfx/flameshot/metadata.xml
```diff
@@ -9,4 +9,8 @@
 		<email>proxy-maint@gentoo.org</email>
 		<name>Proxy Maintainers</name>
 	</maintainer>
+	<use>
+		<flag name='system-singleapplication'>Use the system-wide
+			<pkg>dev-qt/qtsingleapplication</pkg> instead of bundled.</flag>
+	</use>
```
juippis, line 15: Since you're unbundling the dep already with a patch, why would we want to use the bundled one? I think this doesn't need to be a choice, just always prefer the system one.

media-gfx/flameshot/flameshot-0.9.0.ebuild
```diff
@@ -0,0 +1,69 @@
...
+
+inherit desktop cmake xdg-utils
+
+DESCRIPTION="Powerful yet simple to use screenshot software"
+HOMEPAGE="https://flameshot.org https://github.com/flameshot-org/flameshot"
+SRC_URI="https://github.com/flameshot-org/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="FreeArt GPL-3 Apache-2.0"
+SLOT="0"
+KEYWORDS="amd64 x86"
```
juippis, line 14: Please reset the KEYWORDS with `ekeyword ~all your.ebuild` whenever doing version bumps.

media-gfx/flameshot/flameshot-0.9.0.ebuild
```diff
@@ -0,0 +1,69 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit desktop cmake xdg-utils
```
juippis, line 6: You could just inherit *xdg.eclass*, and remove `pkg_post*` phases from this ebuild. Just to clean it up, but your choice.

media-gfx/flameshot/flameshot-0.9.0.ebuild
```diff
@@ -0,0 +1,69 @@
...
+
+DESCRIPTION="Powerful yet simple to use screenshot software"
+HOMEPAGE="https://flameshot.org https://github.com/flameshot-org/flameshot"
+SRC_URI="https://github.com/flameshot-org/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="FreeArt GPL-3 Apache-2.0"
```
juippis, line 12: Is there some reason to such sorting? If not, just sort alphabetically.

media-gfx/flameshot/metadata.xml
pavel-the-best: See the patch. It adds ifdefs and cmake flag whether to use system-provided (I think I will try to add this to mainstream).
The bundled singleapplication version is actually different from the system one, they should be compatible, but who knows. I will probably make this USEFLAG default selected.

media-gfx/flameshot/flameshot-0.9.0.ebuild
pavel-the-best: Will do

media-gfx/flameshot/metadata.xml
juippis: The patch is fine, you could just always use `-DUSE_EXTERNAL_SINGLEAPPLICATION=1` in `src_configure` while removing the bundled one nevertheless. I'd say let's always prefer the system one at least until it breaks, and even then, the choice is visible in the ebuild to quickly test whether bundled one still works.

juippis (approved):
Perfect, works fine too. Thanks!

PR #20585, opened 2021-04-28
media-gfx/openscad-9999: fix installation, media-gfx/openscad-2021.01: adjust variables

waebbl:
media-gfx/openscad-9999: fix installation
Fix the removal of .git{,ignore} for the MCAD library.
Upstream no longer installs those files, so we don't
need to remove them in src_install anymore.

Reported-by: Jan Psota <jasiupsota@gmail.com>
Closes: https://bugs.gentoo.org/785805

media-gfx/openscad-2021.01: adjust variables
Adjust the USE of EROOT in src_configure, due to newly
reported issues from repoman and pkgcheck.

This popped up during fixing of the 9999 ebuild.

### DIFF
--- a/media-gfx/openscad/openscad-2021.01.ebuild
+++ b/media-gfx/openscad/openscad-2021.01.ebuild
@@ -69,9 +69,9 @@ src_prepare() {
 
 src_configure() {
 	if has ccache ${FEATURES}; then
-		eqmake5 "PREFIX = ${EROOT}/usr" "CONFIG += ccache" "${PN}.pro"
+		eqmake5 "PREFIX = ${ESYSROOT}/usr" "CONFIG += ccache" "${PN}.pro"
 	else
-		eqmake5 "PREFIX = ${EROOT}/usr" "${PN}.pro"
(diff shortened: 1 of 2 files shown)


media-gfx/openscad/openscad-2021.01.ebuild
```diff
@@ -69,9 +69,9 @@ src_prepare() {
 
 src_configure() {
 	if has ccache ${FEATURES}; then
-		eqmake5 "PREFIX = ${EROOT}/usr" "CONFIG += ccache" "${PN}.pro"
+		eqmake5 "PREFIX = ${ESYSROOT}/usr" "CONFIG += ccache" "${PN}.pro"
```
thesamesam, line 72: ESYSROOT is definitely less wrong because it's valid and EROOT isn't, but if this is for the destination system, it's probably meant to be EPREFIX instead

ESYSROOT is where we can [find](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-720008.1) DEPEND-category dependencies installed during the build

PR #20586, opened 2021-04-28
sec-policy/selinux-wireguard: Added module for WireGuard.

jpds:


jpds:
@perfinion Can you please review this when you have a moment?

PR #20592, opened 2021-04-29
app-arch/tar: Added "parallel" USE flag to use parallel version of xz, bzip2 and gzip for tar

TobiasFaller:
Added the USE flag "parallel" to app-arch/tar which replaces the single-threaded compression programs used by tar with multi-threaded versions.
This configuration makes the compression of large archives on systems with multiple cores faster than when only using a single thread.

- `bzip2` -> `pbzip2`
- `gzip` -> `pigz`
- `xz` -> `pxz`

Tested on `amd64`. Feedback is appreciated.

Signed-off-by: Tobias Faller <TobiasFaller@gmx.net>


Whissi:
Sorry, I have to reject this PR:

- Violation of https://devmanual.gentoo.org/general-concepts/use-flags/index.html#when-not-to-use-use-flags
- pxz & Co. aren't 1:1 replacements

For example, try using to xz-compress initramfs with pxz and you will get a non-functional initramfs.

PR #20601, opened 2021-04-29
sys-libs/libxcrypt: add cross-* build support

10ne1:
This logic is required for cross LLVM toolchains built against
    glibc with its libcrypt disabled because compiler-rt requires a
    crypt.h header for its sanitizers, leading to a circular build
    dependency situation:
    
    libxcrypt -> compiler-rt -> libxcrypt
    
    To break this circular dep other cross-compilation build systems
    like Yocto have disabled compiler-rt sanitizers entirely [1] or
    a crypt.h header might be copied into the compiler-rt sources.
    
    A better solution than the above two is to directly be able to
    build libxcrypt before compiler-rt in the cross toolchain setup
    steps which is what this addition enables.
    
    [1] https://github.com/kraj/meta-clang/pull/426
    
### DIFF
--- a/sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild
+++ b/sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild
@@ -25,6 +25,18 @@ IUSE="+compat split-usr static-libs system test"
 REQUIRED_USE="split-usr? ( system )"
 RESTRICT="!test? ( test )"
 
+export CTARGET=${CTARGET:-${CHOST}}
+if [[ ${CTARGET} == ${CHOST} ]] ; then
+	if [[ ${CATEGORY/cross-} != ${CATEGORY} ]] ; then
+		export CTARGET=${CATEGORY/cross-}
+	fi
+fi
+
+is_cross() {
+	local enabled_abis=( $(multilib_get_enabled_abis) )
+	[[ "${#enabled_abis[@]}" -le 1 ]] && [[ ${CHOST} != ${CTARGET} ]]
+}
+
 DEPEND="system? (
 		elibc_glibc? (
 			sys-libs/glibc[-crypt(+)]
@@ -114,19 +126,46 @@ src_configure() {
 	multibuild_foreach_variant multilib-minimal_src_configure
 }
 
+get_xcprefix() {
+	if is_cross; then
+		echo "${EPREFIX}/usr/${CTARGET}"
+	else
+		echo "${EPREFIX}"
+	fi
+}
+
 get_xclibdir() {
-	printf -- "%s/%s/%s\n" \
+	printf -- "%s/%s/%s/%s\n" \
+		"$(get_xcprefix)" \
 		"$(usex split-usr '' '/usr')" \
 		"$(get_libdir)" \
 		"$(usex system '' 'xcrypt')"
 }
 
+get_xcincludedir() {
+	printf -- "%s/usr/include/%s\n" \
+		"$(get_xcprefix)" \
+		"$(usex system '' 'xcrypt')"
+}
+
+get_xcmandir() {
+	printf -- "%s/usr/share/man\n" \
+		"$(get_xcprefix)"
+}
+
+get_xcpkgconfigdir() {
+	printf -- "%s/usr/%s/pkgconfig\n" \
+		"$(get_xcprefix)" \
+		"$(get_libdir)"
+}
+
 multilib_src_configure() {
 	local -a myconf=(
 		--disable-werror
-		--libdir="${EPREFIX}"$(get_xclibdir)
-		--with-pkgconfigdir="${EPREFIX}/usr/$(get_libdir)/pkgconfig"
-		--includedir="${EPREFIX}/usr/include/$(usex system '' 'xcrypt')"
+		--libdir=$(get_xclibdir)
+		--with-pkgconfigdir=$(get_xcpkgconfigdir)
+		--includedir=$(get_xcincludedir)
+		--mandir="$(get_xcmandir)"
 	)
 
 	case "${MULTIBUILD_ID}" in
@@ -168,7 +207,7 @@ src_install() {
 		shopt -s failglob || die "failglob failed"
 
 		# Make sure our man pages do not collide with glibc or man-pages.
-		for manpage in "${ED}"/usr/share/man/man3/crypt{,_r}.?*; do
+		for manpage in "${D}$(get_xcmandir)"/man3/crypt{,_r}.?*; do
 			mv -n "${manpage}" "$(dirname "${manpage}")/xcrypt_$(basename "${manpage}")" \
 				|| die "mv failed"
 		done
@@ -177,6 +216,13 @@ src_install() {
 	# Remove useless stuff from installation
 	find "${ED}"/usr/share/doc/${PF} -type l -delete || die
 	find "${ED}" -name '*.la' -delete || die
+
+	# workaround broken upstream cross-* --docdir by installing files in proper locations
+	if is_cross; then
+		insinto "$(get_xcprefix)"/usr/share
+		doins -r "${ED}"/usr/share/doc
+		rm -r "${ED}"/usr/share/doc || die
+	fi
 }
 
 multilib_src_install() {
@@ -185,7 +231,7 @@ multilib_src_install() {
 	# Don't install the libcrypt.so symlink for the "compat" version
 	case "${MULTIBUILD_ID}" in
 		xcrypt_compat-*)
-			rm "${ED}"$(get_xclibdir)/libcrypt$(get_libname) \
+			rm "${D}"$(get_xclibdir)/libcrypt$(get_libname) \
 				|| die "failed to remove extra compat libraries"
 		;;
 		xcrypt_nocompat-*)
@@ -195,7 +241,7 @@ multilib_src_install() {
 						# .a files are installed to /$(get_libdir) by default
 						# Move static libraries to /usr prefix or portage will abort
 						shopt -s nullglob || die "failglob failed"
-						static_libs=( "${ED}"/$(get_xclibdir)/*.a )
+						static_libs=( "${D}"/$(get_xclibdir)/*.a )
 
 						if [[ -n ${static_libs[*]} ]]; then
 							dodir "/usr/$(get_xclibdir)"
@@ -209,13 +255,13 @@ multilib_src_install() {
 						# to allow linker to correctly find shared libraries.
 						shopt -s failglob || die "failglob failed"
 
-						for lib_file in "${ED}"$(get_xclibdir)/*$(get_libname); do
+						for lib_file in "${D}"$(get_xclibdir)/*$(get_libname); do
 							lib_file_basename="$(basename "${lib_file}")"
 							lib_file_target="$(basename "$(readlink -f "${lib_file}")")"
 							dosym "../../$(get_libdir)/${lib_file_target}" "/usr/$(get_xclibdir)/${lib_file_basename}"
 						done
 
-						rm "${ED}"$(get_xclibdir)/*$(get_libname) || die "Removing symlinks in incorrect location failed"
+						rm "${D}"$(get_xclibdir)/*$(get_libname) || die "Removing symlinks in incorrect location failed"
 					fi
 				)
 			fi


sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
```diff
@@ -10,10 +10,24 @@ SRC_URI="https://github.com/besser82/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
...
+		export CTARGET=${CATEGORY/cross-}
+	fi
+fi
+
+is_cross() { [[ ${CHOST} != ${CTARGET} ]] ; }
+
+if is_cross ; then
+	SLOT="${CTARGET}"
+else
+	SLOT="0/1"
```
thesamesam, line 28: The SLOT has to be constant within an ebuild or it’ll violate metadata invariance

Why do we actually need this? We don’t for any other packages?

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
```diff
@@ -52,19 +66,50 @@ src_configure() {
 	multibuild_foreach_variant multilib-minimal_src_configure
 }
 
+get_xcsysroot() {
+	if is_cross; then
+		echo "${EPREFIX}/usr/${CTARGET}"
```
thesamesam, line 71: ESYSROOT is probably more appropriate here, right? And then we can just drop this helper altogether?

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
```diff
@@ -52,19 +66,50 @@ src_configure() {
...
+}
+
+get_xcpkgconfigdir() {
+	printf -- "%s/%s/%s/pkgconfig\n" \
+		"$(get_xcsysroot)" \
+		"$(usex split-usr '' '/usr')" \
+		"$(get_libdir)"
+}
+
 multilib_src_configure() {
```
thesamesam, line 105: As I mentioned, I think we can get rid of most/all of these helpers above by using the right variables provided in EAPI 7

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
```diff
@@ -102,19 +147,25 @@ src_test() {
...
 			mv -n "${manpage}" "$(dirname "${manpage}")/xcrypt_$(basename "${manpage}")" \
 				|| die "mv failed"
 		done
 	) || die "failglob error"
+	fi
 
 	# remove useless stuff from installation
-	find "${D}"/usr/share/doc/${PF} -type l -delete || die
 	find "${D}" -name '*.la' -delete || die
+	if is_cross; then
```
thesamesam, line 164: ?

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
```diff
@@ -136,8 +187,8 @@ multilib_src_install() {
 						static_libs=( "${ED}"/$(get_xclibdir)/*.a )
 
 						if [[ -n ${static_libs[*]} ]]; then
-							dodir "/usr/$(get_xclibdir)"
-							mv "${static_libs[@]}" "${D}/usr/$(get_xclibdir)" \
+							dodir "$(get_xclibdir)"
+							mv "${static_libs[@]}" "${D}$(get_xclibdir)" \
```
thesamesam, line 191: Rolling back to D or ED should be fine with the suggested changes

thesamesam (changes requested):

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
thesamesam: (This [post](https://mgorny.pl/articles/the-ultimate-guide-to-eapi-7.html#improved-cross-compilation-support) is really helpful)

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
thesamesam: (See [PMS](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-600007.1) and the [devmanual](https://devmanual.gentoo.org/general-concepts/portage-cache/index.html))

10ne1: Thanks for bringing up the metadata invariance and pointing to the documentation, that is a very good point. I will drop the SLOT change as it's not necessary and I think the `is_cross` check can also be simplified to something like:
```
[[ ${CATEGORY} == cross-* ]]
```
I will do more testing then come back with an update.

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
10ne1: Will do.

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
thesamesam: No worries, tell me if you want more pointers or anything.

re detecting cross, let’s try use [toolchain.eclass](https://devmanual.gentoo.org/eclass-reference/toolchain-funcs.eclass/)’ helper tc-is-cross-compiler

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
10ne1: My problem here was that I could not find any way to remove the doc files from the standard `/usr/share/doc/libxcrypt` path which caused file conflicts between cross-* builds and plain sys-libs/libxcrypt, so I ended up removing those files from cross builds as you see above.

I tried a lot of methods including passing `--docdir` in the cross-* configure step, but that didn't work even though `./configure` documentation says it should work. For example `--mandir` works as documented and there are no conflicts there between cross an non-cross builds.

BTW, as a side-note, the `sys-apps/man-pages` package also provides the following files which conflict with `sys-apps/libxcrypt`, my solution there was to install the libxcrypt documentation in a subdirectory because --mandir works. Will post this in the next iteration unless I can come up with something better.

```
/usr/share/man/man3/crypt.3.bz2
/usr/share/man/man3/crypt_r.3
```

Any suggestion is welcome. Maybe we could remove the crypt documentation from the man-pages package if it is disabled in glibc?

I'll also try to see if I can somehow get `--docdir` to work, to avoid deleting the cross doc.

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
m-gupta: [[ ${CATEGORY} == cross-* ]] is correct but tc-is-cross-compiler to detect is_cross is not correct.
This is based on crossdev's [1] expectations. glibc/gcc/gdb ebuilds do the same thing [2].

crossdev will create an ebuild named cross-<abi>/libxcrypt e.g. cross-armv7a-linux-gnueabihf/libxcrypt. The expectation is building this package will install files to /usr/armv7a-linux-gnueabihf/. Use of CTARGET is used to install to correct location i.e. /usr/${CTARGET}.

[1] https://wiki.gentoo.org/wiki/Cross_build_environment
[2] https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-libs/glibc/glibc-2.33.ebuild#n60

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
m-gupta: crossdev will create an ebuild named cross-/libxcrypt e.g. cross-armv7a-linux-gnueabihf/libxcrypt. The expectation is building this package will install files to /usr/armv7a-linux-gnueabihf/. Use of CTARGET is used to install to correct location i.e. /usr/${CTARGET}.

So ESYSROOT is not right.

10ne1: I just verified the documentation as well as how the cross- code works in practice and what @m-gupta noted above is correct.

10ne1:
@thesamesam and @m-gupta I just pushed an improved version but I did not touch the helpers. Any specific pointers / suggestions what to improve there are be most welcome!

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
```diff
@@ -52,19 +61,50 @@ src_configure() {
 	multibuild_foreach_variant multilib-minimal_src_configure
 }
 
+get_xcsysroot() {
+	if is_cross; then
+		echo "${EPREFIX}/usr/${CTARGET}"
+	else
+		echo "/"
```
thesamesam, line 68: We're sure this shouldn't be ${ESYSROOT} in the non-cross case?

It seems wrong for it to be _just_ /.

thesamesam (commented):
Let's give this a little bit more time (we've just begun the libxcrypt migration) and we'll revisit it, possibly before we try stabilise the new libxcrypt + glibc ebuilds?

So, could you rebase this (possibly in ~2 weeks) and we'll go from there?

10ne1:
@thesamesam Sure, this can wait a bit longer, it's no rush.

thesamesam:
Would you mind rebasing when you get a chance (no rush) and then we can get this in?

10ne1:
> Would you mind rebasing when you get a chance (no rush) and then we can get this in?

Thanks for waiting and sorry for the delay, I'll try to wrap some other work ASAP and get back to this.

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
thesamesam: I'm still not actually convinced ${ESYSROOT}/usr/${CTARGET} is wrong. Just unusual.

The final destination for my system isn't necessarily where deps are. ESYSROOT is not for where our final location is.

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
thesamesam: Yeah, it should be, as otherwise we're [overriding what PMS does by default](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-12800012.3.8).

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
```diff
@@ -52,19 +61,50 @@ src_configure() {
...
+get_xcincludedir() {
+	printf -- "%s/%s/include/%s\n" \
+		"$(get_xcsysroot)" \
+		"$(usex split-usr '' '/usr')" \
+		"$(usex system '' 'xcrypt')"
+}
+
+get_xcmandir() {
+	printf -- "%s/%s/share/man\n" \
+		"$(get_xcsysroot)" \
```
thesamesam, line 89: sysroot is wrong for this; you want ${EPREFIX}/usr/${CTARGET} or ${EPREFIX} depending on cross or not, because this is for the final product - not where we find dependencies.

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
```diff
@@ -52,19 +61,50 @@ src_configure() {
...
+		"$(usex split-usr '' '/usr')"
+}
+
+get_xcpkgconfigdir() {
+	printf -- "%s/%s/%s/pkgconfig\n" \
+		"$(get_xcsysroot)" \
```
thesamesam, line 95: ditto with man

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
```diff
@@ -102,19 +142,27 @@ src_test() {
 src_install() {
 	multibuild_foreach_variant multilib-minimal_src_install
 
+	if ! use system; then
 	(
 		shopt -s failglob || die "failglob failed"
 
 		# Make sure our man pages do not collide with glibc or man-pages.
-		for manpage in "${ED}"/usr/share/man/man3/crypt{,_r}.?*; do
+		for manpage in "${ED}$(get_xcmandir)"/man3/crypt{,_r}.?*; do
```
thesamesam, line 150: This introduces double prefix (${ED} contains ${EPREFIX} and so does $(get_xcmandir) in the cross case).

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
thesamesam: FWIW: I think since this was started, the ebuild properly respects ${EPREFIX} consistently throughout which might make some of our discussions a bit easier.

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
thesamesam: All that said, binutils seems to either do it wrong too, so it's not a huge deal. I'll sort them all out together at some point, I guess.

10ne1:
Hi @thesamesam Thanks for waiting on this, I was finally able to get back to it. I pushed a cleaned up rebased revision.

I dropped the sysroot arg because it's indeed not needed as you suggested and used the prefix instead, however I could not get the `--prefix` to work (it had no effect), so I couldn't drop all the helpers as you suggested.

Another issue I had is I couldn't reproduce the double prefix you suggested with `${ED}`, in my case $EPREFIX was always empty.

sys-libs/libxcrypt/libxcrypt-4.4.19.ebuild
thesamesam: The issue here is, ED is defined to be `${EPREFIX}/${D}`. But `get_xcmandir` is guaranteed to return something prefixed with `${EPREFIX}`. 

As you've seen, it's no bother when `${EPREFIX}` is blank (most cases), but we can see it's wrong, so we should do something about it to help out Prefix users. The easiest fix here is to simply use `${D}`, not `${ED}` on this line.

(This is testable by setting up Gentoo Prefix but there's no need for you to do that)

sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild
```diff
@@ -114,19 +123,49 @@ src_configure() {
...
+	printf -- "%s/%s/%s/pkgconfig\n" \
+		"$(get_xcprefix)" \
+		"$(usex split-usr '' '/usr')" \
+		"$(get_libdir)"
+}
+
 multilib_src_configure() {
 	local -a myconf=(
 		--disable-werror
 		--libdir="${EPREFIX}"$(get_xclibdir)
```
thesamesam, line 165: Double prefix again -- just `$(get_xclibdir)` (`get_xclibdir` is guaranteed to return something prefixed with `${EPREFIX}`, so prepending it with `${EPREFIX}` is doubling it)

sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild
```diff
@@ -177,6 +216,12 @@ src_install() {
 	# Remove useless stuff from installation
 	find "${ED}"/usr/share/doc/${PF} -type l -delete || die
 	find "${ED}" -name '*.la' -delete || die
+
+	# workaround broken upstream cross-* --docdir by installing files in proper locations
+	if is_cross; then
+		install -d "${ED}$(get_xcprefix)"/usr/share
```
thesamesam, line 222: In line with other bits:
```suggestion
		install -d "${D}$(get_xcprefix)"/usr/share || die
```

... but please do `insinto` instead (and `doins`)

sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild
```diff
@@ -177,6 +216,12 @@ src_install() {
...
+		mv "${ED}"/usr/share/doc "${ED}$(get_xcprefix)"/usr/share
```
thesamesam, line 223: This won't matter I _think_ once using `insinto`/`doins`, but _second_ argument should have `${D}` not `${ED}` for earlier mentioned reasons.

Also, use `|| die` on external commands.

thesamesam (commented):
Thanks a lot! Do appreciate the effort you put into these. Tiny comments (easily fixed) and then we're good!

10ne1:
Thanks again @thesamesam for all your patience and clarifications. I'm not sure why I can't reply on the issue comments but I think I've addressed them all.

Indeed I am not using a Gentoo Prefix installed system, which is separate from the --prefix related attempts to drop the helper functions (I failed at that unfortunately) and was also confused by the sysroot vs prefix combinations, but I think it's all clear now in my head after your awesome explanations and guidance.

thesamesam:
Uh oh!

```
 *
 *   /include
 *   /share
 *   /usr/usr
 *
   /usr/lib64/libcrypt.a
   /lib64/libcrypt.so.2.0.0
   /lib64/libcrypt.so.1.1.0
   /usr/x86_64-pc-linux-gnu/lib/libcrypt.so.1.1.0
   /usr/usr/x86_64-pc-linux-gnu/lib/libcrypt.a
   /usr/x86_64-pc-linux-gnu/lib/libcrypt.so.2.0.0
installsources: rsyncing source files
```

So then I started to play because I didn't really see how we could end up with CHOST in /usr/*, and realised that `is_cross` misfires for `abi_x86_32` (multilib) builds on amd64, and it dies with a header mismatch. Not yet sure how to fix it.

Diff I was playing with:
```
diff --git a/sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild b/sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild
index 9036e42eb6cb..85a4598d65fe 100644
--- a/sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild
+++ b/sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild
@@ -140,22 +140,19 @@ get_xclibdir() {
 }

 get_xcincludedir() {
-       printf -- "%s/%s/include/%s\n" \
+       printf -- "%s/usr/include/%s\n" \
                "$(get_xcprefix)" \
-               "$(usex split-usr '' '/usr')" \
                "$(usex system '' 'xcrypt')"
 }

 get_xcmandir() {
-       printf -- "%s/%s/share/man\n" \
-               "$(get_xcprefix)" \
-               "$(usex split-usr '' '/usr')"
+       printf -- "%s/usr/share/man\n" \
+               "$(get_xcprefix)"
 }

 get_xcpkgconfigdir() {
-       printf -- "%s/%s/%s/pkgconfig\n" \
+       printf -- "%s/usr/%s/pkgconfig\n" \
                "$(get_xcprefix)" \
-               "$(usex split-usr '' '/usr')" \
                "$(get_libdir)"
 }

@@ -168,6 +165,12 @@ multilib_src_configure() {
                --mandir="$(get_xcmandir)"
        )

+        if is_cross ; then
+                echo "CHOST: ${CHOST}"
+                echo "CTARGET: ${CTARGET}"
+                die
+        fi
+
        case "${MULTIBUILD_ID}" in
                xcrypt_compat-*)
                        myconf+=(
@@ -231,7 +234,7 @@ multilib_src_install() {
        # Don't install the libcrypt.so symlink for the "compat" version
        case "${MULTIBUILD_ID}" in
                xcrypt_compat-*)
-                       rm "${ED}"$(get_xclibdir)/libcrypt$(get_libname) \
+                       rm "${D}"/$(get_xclibdir)/libcrypt$(get_libname) \
                                || die "failed to remove extra compat libraries"
                ;;
                xcrypt_nocompat-*)
@@ -241,11 +244,11 @@ multilib_src_install() {
                                                # .a files are installed to /$(get_libdir) by default
                                                # Move static libraries to /usr prefix or portage will abort
                                                shopt -s nullglob || die "failglob failed"
-                                               static_libs=( "${ED}"/$(get_xclibdir)/*.a )
+                                               static_libs=( "${D}"/$(get_xclibdir)/*.a )

                                                if [[ -n ${static_libs[*]} ]]; then
                                                        dodir "/usr/$(get_xclibdir)"
-                                                       mv "${static_libs[@]}" "${ED}/usr/$(get_xclibdir)" \
+                                                       mv "${static_libs[@]}" "${D}/usr/$(get_xclibdir)" \
                                                                || die "Moving static libs failed"
                                                fi
                                        fi
@@ -255,13 +258,13 @@ multilib_src_install() {
                                                # to allow linker to correctly find shared libraries.
                                                shopt -s failglob || die "failglob failed"

-                                               for lib_file in "${ED}"$(get_xclibdir)/*$(get_libname); do
+                                               for lib_file in "${D}"/usr/$(get_xclibdir)/*$(get_libname); do
                                                        lib_file_basename="$(basename "${lib_file}")"
                                                        lib_file_target="$(basename "$(readlink -f "${lib_file}")")"
-                                                       dosym "../../$(get_libdir)/${lib_file_target}" "/usr/$(get_xclibdir)/${lib_file_basename}"
+                                                       dosym "../../$(get_libdir)/${lib_file_target}" "$(get_xclibdir)/${lib_file_basename}"
                                                done

-                                               rm "${ED}"$(get_xclibdir)/*$(get_libname) || die "Removing symlinks in incorrect location failed"
+                                               rm "${D}"/usr/$(get_xclibdir)/*$(get_libname) || die "Removing symlinks in incorrect location failed"
                                        fi
                                )
                        fi
```

thesamesam:
Okay, this seems to work for me, does it work for you? I'll carry on testing now.

```
diff --git a/sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild b/sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild
index 9036e42eb6cb..3a86e64f4aa3 100644
--- a/sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild
+++ b/sys-libs/libxcrypt/libxcrypt-4.4.28.ebuild
@@ -32,7 +32,10 @@ if [[ ${CTARGET} == ${CHOST} ]] ; then
        fi
 fi

-is_cross() { [[ ${CHOST} != ${CTARGET} ]] ; }
+is_cross() {
+       local enabled_abis=( $(multilib_get_enabled_abis) )
+       [[ "${#enabled_abis[@]}" -le 1 ]] && [[ ${CHOST} != ${CTARGET} ]]
+}

 DEPEND="system? (
                elibc_glibc? (
@@ -140,22 +143,19 @@ get_xclibdir() {
 }

 get_xcincludedir() {
-       printf -- "%s/%s/include/%s\n" \
+       printf -- "%s/usr/include/%s\n" \
                "$(get_xcprefix)" \
-               "$(usex split-usr '' '/usr')" \
                "$(usex system '' 'xcrypt')"
 }

 get_xcmandir() {
-       printf -- "%s/%s/share/man\n" \
-               "$(get_xcprefix)" \
-               "$(usex split-usr '' '/usr')"
+       printf -- "%s/usr/share/man\n" \
+               "$(get_xcprefix)"
 }

 get_xcpkgconfigdir() {
-       printf -- "%s/%s/%s/pkgconfig\n" \
+       printf -- "%s/usr/%s/pkgconfig\n" \
                "$(get_xcprefix)" \
-               "$(usex split-usr '' '/usr')" \
                "$(get_libdir)"
 }

@@ -231,7 +231,7 @@ multilib_src_install() {
        # Don't install the libcrypt.so symlink for the "compat" version
        case "${MULTIBUILD_ID}" in
                xcrypt_compat-*)
-                       rm "${ED}"$(get_xclibdir)/libcrypt$(get_libname) \
+                       rm "${D}"$(get_xclibdir)/libcrypt$(get_libname) \
                                || die "failed to remove extra compat libraries"
                ;;
                xcrypt_nocompat-*)
@@ -241,7 +241,7 @@ multilib_src_install() {
                                                # .a files are installed to /$(get_libdir) by default
                                                # Move static libraries to /usr prefix or portage will abort
                                                shopt -s nullglob || die "failglob failed"
-                                               static_libs=( "${ED}"/$(get_xclibdir)/*.a )
+                                               static_libs=( "${D}"/$(get_xclibdir)/*.a )

                                                if [[ -n ${static_libs[*]} ]]; then
                                                        dodir "/usr/$(get_xclibdir)"
@@ -255,13 +255,13 @@ multilib_src_install() {
                                                # to allow linker to correctly find shared libraries.
                                                shopt -s failglob || die "failglob failed"

-                                               for lib_file in "${ED}"$(get_xclibdir)/*$(get_libname); do
+                                               for lib_file in "${D}"$(get_xclibdir)/*$(get_libname); do
                                                        lib_file_basename="$(basename "${lib_file}")"
                                                        lib_file_target="$(basename "$(readlink -f "${lib_file}")")"
                                                        dosym "../../$(get_libdir)/${lib_file_target}" "/usr/$(get_xclibdir)/${lib_file_basename}"
                                                done

-                                               rm "${ED}"$(get_xclibdir)/*$(get_libname) || die "Removing symlinks in incorrect location failed"
+                                               rm "${D}"$(get_xclibdir)/*$(get_libname) || die "Removing symlinks in incorrect location failed"
                                        fi
                                )
                        fi
```

10ne1:
> Okay, this seems to work for me, does it work for you?

Yes, it works for me as well. I applied the diff on top of my latest change from this pull request - it applied cleanly after I converted spaces to tabs and all the files are in the correct places for both the cross and non-cross build.

How do you wish to proceed? I can add the diff code to the PR if you think it's ready and I would also like to list you as co-author in the commit message :smile:

10ne1:
> Go ahead and merge my stuff in to yours!

Done. Thanks again for the awesome collaboration.

thesamesam:
> > Go ahead and merge my stuff in to yours!
> 
> Done. Thanks again for the awesome collaboration.

To many more, I hope! And thank you!

PR #20603, opened 2021-04-29
net-misc/openssh: systemd improvements

expeditioneer:
added patch for type notify
individual host key creation

Signed-off-by: Dennis Lamm <expeditioneer@gentoo.org>

### DIFF
--- a/net-misc/openssh/files/openssh-8.5_p1-systemd.patch
+++ b/net-misc/openssh/files/openssh-8.5_p1-systemd.patch
@@ -1,0 +1,74 @@
+Author: Jakub Jelen <jjelen@redhat.com>
+Date:   Mon Nov 21 15:04:06 2016 +0100
+
+    systemd stuff
+
+diff --git a/configure.ac b/configure.ac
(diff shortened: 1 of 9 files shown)


floppym:
I don't really see any real benefit from this.

expeditioneer:
@floppym

IMHO the service type notify is a good improvement over the default simple type.

The benefit of the creation of individual host files can be valued differently. I think the benefit is that you don't have unnecessary host keys on your system. Also I think this approach is 'cleaner' since a service is only doing one thing.

I personally use ansible for the config management on my nodes and i always get a 'change' when i delete my unwanted host keys. The separation keeps the system in a consistent state.

chutz:
@expeditioneer Would it be possible to upstream the patch to OpenSSH?

Generally there has to be a pretty good reason for yet another patch for OpenSSH, there already are quite a few patches that need updating for every version bump.

expeditioneer:
@chutz  I've created a pull-request upstream, see https://github.com/openssh/openssh-portable/pull/251

PR #20604, opened 2021-04-29
mail-mta/nullmailer: Set environment variables AR and RANLIB.

tastytea:
The build systems falls back to `ar` and `ranlib` otherwise.

Closes: https://bugs.gentoo.org/668850

### DIFF
--- a/mail-mta/nullmailer/nullmailer-2.2-r1.ebuild
+++ b/mail-mta/nullmailer/nullmailer-2.2-r1.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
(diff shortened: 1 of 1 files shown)


mail-mta/nullmailer/nullmailer-2.2-r1.ebuild
```diff
@@ -71,6 +71,12 @@ src_configure() {
 		$(use_enable ssl tls)
 }
 
+src_compile() {
+	export AR="$(tc-getAR)"
```
thesamesam, line 75: I've changed this to tc-export AR RANLIB. Thank you!

PR #20607, opened 2021-04-30
dev-vcs/git: Fix runtime dependencies for git-cvsimport

Nekun:
Closes: https://bugs.gentoo.org/787017

I consider it as a non-critical issue because it seems like RSH hosts is not alive in the real world at the moment (e.g., OpenBSD anon mirrors provides only SSH) and it's always necessary to do `export CVS_RSH=ssh`, so don't revbump and rely on Portage dynamic dependencies.


thesamesam:
@Polynomial-C

PR #20615, opened 2021-04-30
app-backup/duply: drop older version and fix typos

scardracs:


a17r (changes requested):
Pushed old version removal as a separate commit, as it should be, in commit f256f508b39e277be43e8e678ffe017aedd5ad9c.

The other changes do not strike me as typo fixes, much more unnecessary changes to a stable ebuild. I don't think there is anything to do here.

scardracs:
> Pushed old version removal as a separate commit, as it should be, in commit [f256f50](https://github.com/gentoo/gentoo/commit/f256f508b39e277be43e8e678ffe017aedd5ad9c).
> 
> The other changes do not strike me as typo fixes, much more unnecessary changes to a stable ebuild. I don't think there is anything to do here.

Ok, closing PR

PR #20616, opened 2021-04-30
net-im/telegram-desktop: bump to 2.7.4

mid-kid:
- media-libs/libtgvoip: Make pulseaudio optional again
- media-libs/tg_owt: bump to 0_pre20210422, unbundle libs
- media-libs/tg_owt: rebundle usrsctp
- net-im/telegram-desktop: bump to 2.7.4

I'm confident enough in the tg_owt patchset and this update is minor enough that people can just downgrade in the worst case scenario.

Also, @gyakovlev, I have no access to a ppc64 machine nor much of any idea how to emulate one. Can you build usrsctp and request keywording?  
I'm also unsure about the "causes forced inclusion of SSE2" sed line in tg_owt...


gyakovlev:
looking/testing now, will provide more feedback bit later.

I've just keyworded `usrsctp` with `~ppc64` so it should be ok now to unbundle, I'll drop the commit or you can remove it. thanks for leaving it separate.

next time just drop the keyword if you can't test.
the right way to do it is: drop keywords, open re-keywording bug.

as for sed line and sse, it's no longer needed. I'll comment more inline. in review.

gyakovlev (approved):
ok looks good.
I've altered `sse`2 workaround to be `ppc64` specific and added a comment explaining what happens and what I did to fix ;-)
it's a "bug" in upstream webrtc and tg will not accept the fix, I think.

arm/aarch64 seems to be handled there already.

haven't tested webkit flag/integration at all, so I assume it's tested.

gyakovlev:
https://github.com/gentoo/gentoo/commit/b69c36acab3ef74df82fe39d81b36feac34b23bc here's the sse2 fix

basically this code does not work

https://github.com/desktop-app/tg_owt/blob/18cb4cd9bb4c2f5f5f5e760ec808f74c302bc1bf/src/modules/desktop_capture/differ_block.cc#L29-L48

it seems that `have_sse2` is true on ppc64, which is pretty strange.
it's the only part it's used in the code according to grep(tm)

mid-kid:
@gyakovlev 

Thanks for documenting your hack, I wasn't sure of what it did. Patching a single file is significantly better than excluding a bunch of files which might become used later down the line.

I'll keep the keywording thing in mind in the future, I wasn't really sure what to do in this situation, but dropping the keyword seems sensible too.

I've tested webkit, it's a bit annoying to test because it's only used for payments, and you need to message `@ShopBot` to create an invoice that uses webkit. I might mention this for the stabilization process, but otherwise it seems to work on my machine.

PR #20617, opened 2021-04-30
mail-client/geary-40.0: Use release build profile

Leo3418:
Upstream documentation suggests that distributions must use the release build profile when providing a package for Geary:
https://gitlab.gnome.org/GNOME/geary/-/blob/gnome-40.0/BUILDING.md#build-profiles


mattst88:
Thanks!

PR #20620, opened 2021-04-30
app-text/mandoc: Add systemd service and timer

wylfen:
mandoc updates its databases with the makewhatis(8) tool which should be
invoked regularly to keep tools such as apropos(1) and whatis(1) fully
functional. This commit introduces a systemd service and timer to that
end.

Since makewhatis(8) needs to write to root-owned directories and there
is no infrastructure yet to allow an unprivileged approach, we make sure
to sufficiently lock down its invocation.

Whilst the cronjob shipped in this ebuild still relies on sourcing
/etc/profile (this is needed to make makewhatis(8) aware of MANPATH), we
can instead rely on /etc/environment.d/10-gentoo-env.conf which was
added for [1].

[1] https://bugs.gentoo.org/show_bug.cgi?id=704416

Signed-off-by: Wolfgang Müller <wolf@oriole.systems>


wylfen:
Sorry, deleted this by accident. If anyone wants to pick this up I can resubmit.

PR #20621, opened 2021-04-30
app-emulation/xen: bump to 4.14.2 / 4.15.0

hydrapolic:
Compile tested on ~amd64 (binutils-2.36.1, gcc-11).

Thanks <3


Whissi:
@hydrapolic Please notice commit 5871ab0ecdb76ffde3576f46bfcf02e7908d5f81. I had to change SRC_URI after user reported
```
!!! Fetched file: seabios-1.14.0.tar.gz VERIFY FAILED!
!!! Got:      628981
!!! Expected: 628985   
```
Is there a reason why we don't pull in `sys-firmware/seabios[binary]` at all?

hydrapolic:
Thanks @Whissi, IIRC SRC_URI was changed because the version used in Xen wasn't available on the official link. It was true for some previous version and during the bump this wasn't changed.

Actually sys-firmware/seabios is pulled in for USE=system-seabios.

hydrapolic:
@Whissi we need to take care of https://bugs.gentoo.org/789648

Whissi:
Thanks, fixed.

hydrapolic:
Thank you @Whissi

PR #20624, opened 2021-05-01
app-crypt/adcli: new package and optfeature for sssd

henning-schild:
ActiveDirectory setups can hand out a machine credential called the "kerberos keytab" to connected machines. This can be set to expire forcing machines to regularly renew it. While that renewal is supported by sssd, the feature can not be used in gentoo. That is because gentoo so far does not ship the application required for the renew process.

https://github.com/SSSD/sssd/blob/7313efba2cd668ce622c4bf54b94a725e7209617/src/providers/ad/ad_machine_pw_renewal.c#L30

This PR brings the application into gentoo. It in fact is useful on its own, if people have a kerberos keytab that needs renewal but do not use sssd. But it really seems to be required by sssd if expiring keytabs are used on the network the client is connected to.


henning-schild:
@alexxy

henning-schild:
Please let me know if that one needs a bug opened.

henning-schild:
@Whissi

henning-schild:
> > which version of make are you using. I would like to repro your doc issue and fix it upstream, but i can not repro
> 
> It seems to be looking for `app-text/docbook-xml-dtd:4.3`, in an ultra minimal system it /could/ be missing. Not particularly familiar with docbook stuff but guess need to depend on it.
> 
> (nevermind my deleted message, realized after it wasn't that)

Not sure which "deleted message" i might have missed. @juippis please tell me how you run into the issue so i can repro

If is is another "make" i would be surprised how that can happen but would push for the upstream change in the Makefile. Otherwise you might want to tell us whether or not having "docbook-xml-dtd" solves the problem.

Can not solve a problem i can not repo .. especially when wanting to ague things upstream.

PR #20631, opened 2021-05-01
EAPI bumps for virtual packages

dm0-:
@SoapGentoo Here are a few trivial EAPI bumps for the latest revisions of some virtual packages.  The first three are for [EAPI5Removal](https://bugs.gentoo.org/698100).  While [EAPI6Removal](https://bugs.gentoo.org/770247) probably isn't a concern yet, there were only a handful of other non-EAPI7 virtuals installed on my system, so I included those as well.  (I skipped `virtual/eject` due to #19828.)

### DIFF
--- a/virtual/opengl/opengl-7.0-r2.ebuild
+++ b/virtual/opengl/opengl-7.0-r2.ebuild
@@ -1,11 +1,12 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
 
 inherit multilib-build
 
 DESCRIPTION="Virtual for OpenGL implementation"
+
 SLOT="0"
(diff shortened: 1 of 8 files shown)


virtual/ttf-fonts/ttf-fonts-1-r1.ebuild
```diff
@@ -1,7 +1,7 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
 
 DESCRIPTION="Virtual for Serif/Sans/Monospace font packages"
```
SoapGentoo, line 6: empty line after

virtual/opengl/opengl-7.0-r2.ebuild
```diff
@@ -1,7 +1,7 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
 
 inherit multilib-build
```
SoapGentoo, line 7: same here

SoapGentoo (changes requested):

thesamesam:
Some are in #20538 but that's fine.

dm0-:
Oh, I just searched open PRs for "virtual" but it looks like that didn't match.  It seems only the EAPI 5 ones are duplicated there, so I could drop them if preferred.

PR #20637, opened 2021-05-01
www-client/qutebrowser: verbump to 2.2.1

BouncyMaster:
also remove v2.2.0-r1

Signed-off-by: Nicu Borta <nicuborta123@tutanota.com>


juippis (commented):
Please split the addition and removal of old ebuilds into separate commits, so this should have 2 commits instead of 1.

@GuillaumeSeren @steils any comments?

GuillaumeSeren:
Hey @BouncyMaster ,
Usually I think I think it is better to open a bug on the gentoo bugtracker,
to let other people / maintainer know that you are working on this like (https://bugs.gentoo.org/721544).

Also about cleanup an old version it should be the oldest one if not stable,
if stable one of the oldest unstable one.

steils:
Thanks for the PR, sorry for the delay.

> Please split the addition and removal of old ebuilds into separate commits, so this should have 2 commits instead of 1.
> 
> @GuillaumeSeren @steils any comments?

Thanks. Well, usually I'm okay with adding a new version and dropping the old one in a single commit, especially in simple cases.

PR #20638, opened 2021-05-01
sys-firmware/sof-firmware: Bump to 1.6.1

simonvanderveldt:
Closes: https://bugs.gentoo.org/777537


mattst88:
v1.7 was released a month ago. Add that instead?

simonvanderveldt:
> v1.7 was released a month ago. Add that instead?

Yeah, I'm aware. 1.7 has more changes, the script to install is no longer used and since it's not my ebuild I'd rather let a maintainer fix it. Apart from that 1.7 gave me warnings that it uses a kernel interface that's newer than my kernel (and I'm on 5.11.17) and the new hardware support I need (and I guess most others as well) was added in 1.6 so don't need 1.7.

mattst88:
Okay, that makes sense.

simonvanderveldt:
@mattst88 Can you merge this PR (assuming it's OK) or does another maintainer need to do that?

mattst88:
Done, thanks!

PR #20639, opened 2021-05-01
Support EAPI 7 in kernel-2.eclass

dm0-:
@SoapGentoo This is the kernel eclass update with additional QA stuff in separate commits.

https://bugs.gentoo.org/702280

### DIFF
--- a/eclass/kernel-2.eclass
+++ b/eclass/kernel-2.eclass
@@ -8,7 +8,7 @@
 # John Mylchreest <johnm@gentoo.org>
 # Mike Pagano <mpagano@gentoo.org>
 # <so many, many others, please add yourself>
-# @SUPPORTED_EAPIS: 2 3 4 5 6 7
+# @SUPPORTED_EAPIS: 6 7
 # @BLURB: Eclass for kernel packages
 # @DESCRIPTION:
 # This is the kernel.eclass rewrite for a clean base regarding the 2.6
@@ -199,10 +199,9 @@
 # that of course does not mean we're not willing to help.
 
 inherit estack toolchain-funcs
-[[ ${EAPI:-0} == [012345] ]] && inherit epatch
-[[ ${EAPI:-0} == [0123456] ]] && inherit eapi7-ver
+[[ ${EAPI:-0} == 6 ]] && inherit eapi7-ver
 case ${EAPI:-0} in
-	2|3|4|5|6|7)
+	6|7)
 		EXPORT_FUNCTIONS src_{unpack,prepare,compile,install,test} \
 			pkg_{setup,preinst,postinst,postrm} ;;
 	*) die "${ECLASS}: EAPI ${EAPI} not supported" ;;
@@ -597,7 +596,7 @@ kernel_is_2_6() {
 
 # Capture the sources type and set DEPENDs
 if [[ ${ETYPE} == sources ]]; then
-	[[ ${EAPI} == [0-6] ]] && DEPEND="!build? ( sys-apps/sed )" ||
+	[[ ${EAPI} == 6 ]] && DEPEND="!build? ( sys-apps/sed )" ||
 	BDEPEND="!build? ( sys-apps/sed )"
 	RDEPEND="!build? (
 		dev-lang/perl
@@ -633,7 +632,7 @@ if [[ ${ETYPE} == sources ]]; then
 			# tree has been dropped from the kernel.
 			kernel_is lt 4 14 && LICENSE+=" !deblob? ( linux-firmware )"
 
-			[[ ${EAPI} == [0-6] ]] && DEPEND+=" deblob? ( ${PYTHON_DEPS} )" ||
+			[[ ${EAPI} == 6 ]] && DEPEND+=" deblob? ( ${PYTHON_DEPS} )" ||
 			BDEPEND+=" deblob? ( ${PYTHON_DEPS} )"
 
 			if [[ -n KV_MINOR ]]; then
@@ -1559,12 +1558,7 @@ kernel-2_src_unpack() {
 
 kernel-2_src_prepare() {
 	debug-print "Applying any user patches"
-
-	# apply any user patches
-	case ${EAPI:-0} in
-		0|1|2|3|4|5) epatch_user ;;
-		*) eapply_user ;;
-	esac
+	eapply_user
 }
 
 # @FUNCTION: kernel-2_src_compile
--- a/eclass/kernel-2.eclass
+++ b/eclass/kernel-2.eclass
@@ -731,7 +731,7 @@ unpack_2_4() {
 	make -s mrproper ${xmakeopts} || die "make mrproper failed"
 	make -s symlinks ${xmakeopts} || die "make symlinks failed"
 	make -s include/linux/version.h ${xmakeopts} || die "make include/linux/version.h failed"
-	echo ">>> version.h compiled successfully."
+	elog ">>> version.h compiled successfully."
 }
 
 # @FUNCTION: unpack_2_6
@@ -839,7 +839,7 @@ compile_headers() {
 
 	if kernel_is 2 4; then
 		yes "" | make oldconfig ${xmakeopts}
-		echo ">>> make oldconfig complete"
+		elog ">>> make oldconfig complete"
 		make dep ${xmakeopts}
 	elif kernel_is 2 6; then
 		# 2.6.18 introduces headers_install which means we dont need any
@@ -961,7 +961,7 @@ install_sources() {
 
 	cd "${S}" || die
 	dodir /usr/src
-	echo ">>> Copying sources ..."
+	elog ">>> Copying sources ..."
 
 	file="$(find ${WORKDIR} -iname "docs" -type d)"
 	if [[ -n ${file} ]]; then
@@ -1030,11 +1030,9 @@ postinst_sources() {
 	# Don't forget to make directory for sysfs
 	[[ ! -d ${EROOT%/}/sys ]] && kernel_is 2 6 && { mkdir "${EROOT%/}"/sys || die ; }
 
-	echo
 	elog "If you are upgrading from a previous kernel, you may be interested"
 	elog "in the following document:"
 	elog "  - General upgrade guide: https://wiki.gentoo.org/wiki/Kernel/Upgrade"
-	echo
 
 	# if K_EXTRAEINFO is set then lets display it now
 	if [[ -n ${K_EXTRAEINFO} ]]; then
@@ -1062,12 +1060,10 @@ postinst_sources() {
 	#  And now the general message.
 	if [[ -n ${K_SECURITY_UNSUPPORTED} ]]; then
 		ewarn "This means that it is likely to be vulnerable to recent security issues."
-		echo
 		ewarn "Upstream kernel developers recommend always running the latest "
 		ewarn "release of any current long term supported Linux kernel version."
 		ewarn "To see a list of these versions, their most current release and "
 		ewarn "long term support status, please go to https://www.kernel.org ."
-		echo
 		ewarn "For specific information on why this kernel is unsupported, please read:"
 		ewarn "https://wiki.gentoo.org/wiki/Project:Kernel_Security"
 	fi
@@ -1079,7 +1075,6 @@ postinst_sources() {
 	if [[ "$(tc-arch)" = "sparc" ]]; then
 		if [[ $(gcc-major-version) -lt 4 && $(gcc-minor-version) -lt 4 ]]; then
 			if [[ ${KV_MAJOR} -ge 3 ]] || ver_test ${KV_MAJOR}.${KV_MINOR}.${KV_PATCH} -gt 2.6.24 ; then
-				echo
 				elog "NOTE: Since 2.6.25 the kernel Makefile has changed in a way that"
 				elog "you now need to do"
 				elog "  make CROSS_COMPILE=sparc64-unknown-linux-gnu-"
@@ -1087,7 +1082,6 @@ postinst_sources() {
 				elog "  make"
 				elog "to compile the kernel. For more information please browse to"
 				elog "https://bugs.gentoo.org/show_bug.cgi?id=214765"
-				echo
 			fi
 		fi
 	fi
@@ -1107,7 +1101,6 @@ setup_headers() {
 	done
 
 	if [[ ${H_ACCEPT_ARCH} != "yes" ]]; then
-		echo
 		eerror "This version of ${PN} does not support $(tc-arch)."
 		eerror "Please merge the appropriate sources, in most cases"
 		eerror "(but not all) this will be called $(tc-arch)-headers."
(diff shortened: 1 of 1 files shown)


eclass/kernel-2.eclass
```diff
@@ -600,10 +600,8 @@ kernel_is_2_6() {
 
 # Capture the sources type and set DEPENDs
 if [[ ${ETYPE} == sources ]]; then
-	DEPEND="!build? (
-		sys-apps/sed
-		>=sys-devel/binutils-2.11.90.0.31
-	)"
+	[[ ${EAPI} == [0-6] ]] && DEPEND="!build? ( sys-apps/sed )" ||
```
mjeveritt, line 600: We don't tend to use ranges for EAPI, as there is no guarantee (PMS) that it is, or will continue to be, numeric.

eclass/kernel-2.eclass
```diff
@@ -932,26 +931,26 @@ install_headers() {
 	# of this crap anymore :D
 	if kernel_is ge 2 6 18 ; then
 		env_setup_xmakeopts
-		emake headers_install INSTALL_HDR_PATH="${ED}"${ddir}/.. ${xmakeopts} || die
+		emake headers_install INSTALL_HDR_PATH="${ED%/}"${ddir}/.. ${xmakeopts} || die
```
mjeveritt, line 931: Is ${ddir} assumed to have a leading slash ?

mjeveritt (approved):
LGTM, thanks.

eclass/kernel-2.eclass
thesamesam: In this case, it’s okay because we die on unknown EAPIs above anyway, and this is only a compatibility hack for older ones

eclass/kernel-2.eclass
```diff
@@ -813,9 +809,9 @@ universal_unpack() {
 # handle EXTRAVERSION
 
 unpack_set_extraversion() {
-	cd "${S}"
+	pushd "${S}" >/dev/null || die
 	sed -i -e "s:^\(EXTRAVERSION =\).*:\1 ${EXTRAVERSION}:" Makefile
```
thesamesam, line 813: die

eclass/kernel-2.eclass
```diff
@@ -825,8 +821,8 @@ unpack_set_extraversion() {
 # Otherwise patches that modify the same area of Makefile will fail
 
 unpack_fix_install_path() {
-	cd "${S}"
-	sed	-i -e 's:#export\tINSTALL_PATH:export\tINSTALL_PATH:' Makefile
+	cd "${S}" || die
+	sed -i -e 's:#export\tINSTALL_PATH:export\tINSTALL_PATH:' Makefile
```
thesamesam, line 825: die

eclass/kernel-2.eclass
```diff
@@ -855,7 +851,7 @@ compile_headers() {
 
 		# autoconf.h isnt generated unless it already exists. plus, we have
 		# no guarantee that any headers are installed on the system...
-		[[ -f ${EROOT}usr/include/linux/autoconf.h ]] \
+		[[ -f ${EROOT%/}/usr/include/linux/autoconf.h ]] \
```
thesamesam, line 854: die

eclass/kernel-2.eclass
```diff
@@ -914,10 +910,10 @@ compile_headers_tweak_config() {
 # Fix permissions in tarball
 
 install_universal() {
-	cd "${WORKDIR}"
+	pushd "${WORKDIR}" >/dev/null || die
 	chown -R 0:0 * >& /dev/null
```
thesamesam, line 914: die, die

eclass/kernel-2.eclass
```diff
@@ -932,32 +928,32 @@ install_headers() {
...
+	cp -pPR "${S}"/include/asm/* "${ED%/}"${ddir}/asm || die
 
 	if kernel_is 2 6 ; then
 		dodir ${ddir}/asm-generic
-		cp -pPR "${S}"/include/asm-generic/* "${ED}"${ddir}/asm-generic || die
+		cp -pPR "${S}"/include/asm-generic/* "${ED%/}"${ddir}/asm-generic || die
 	fi
 
 	# clean up
 	find "${D}" -name '*.orig' -exec rm -f {} \;
```
thesamesam, line 954: die

eclass/kernel-2.eclass
```diff
@@ -968,7 +964,7 @@ install_headers() {
 install_sources() {
 	local file
 
-	cd "${S}"
+	cd "${S}" || die
 	dodir /usr/src
 	echo ">>> Copying sources ..."
```
thesamesam, line 969: Maybe change to a log?

eclass/kernel-2.eclass
```diff
@@ -1595,7 +1575,7 @@ kernel-2_src_prepare() {
 # conpile headers or run deblob script
 
 kernel-2_src_compile() {
-	cd "${S}"
+	cd "${S}" || die
 	[[ ${ETYPE} == headers ]] && compile_headers
 
 	if [[ $K_DEBLOB_AVAILABLE == 1 ]] && use deblob ; then
```
thesamesam, line 1581: {}

thesamesam (changes requested):
Not had a chance to look for other possible improvements. Thanks for looking at this!

eclass/kernel-2.eclass
thesamesam: I was wondering that too :)

eclass/kernel-2.eclass
```diff
@@ -1435,22 +1435,19 @@ unipatch() {
...
 		basefname=$(basename ${2})
 		basedname=$(dirname ${2})
 		unset ARCH
 
-		cd ${basedname}
-		echo -e "include ${basefname}\ne:\n\t@echo \$(${1})" | \
-			make ${BUILD_FIXES} -s -f - e 2>/dev/null
-		cd ${workingdir}
+		echo -e "include ${basefname}\ne:\n\t@echo \$(${1})" |
+		emake -C "${basedname}" ${BUILD_FIXES} -s -f - e 2>/dev/null
```
mjeveritt, line 1439: Perhaps keep indentation here, to indicate the continuation (I note dropping possibly redundant backslash)?

eclass/kernel-2.eclass
dm0-: Yes, `ddir` is assigned to the output of `kernel_header_destdir`, which always outputs a leading slash: https://github.com/gentoo/gentoo/blob/master/eclass/kernel-2.eclass#L692-L696

eclass/kernel-2.eclass
```diff
@@ -1197,7 +1186,6 @@ unipatch() {
 
 			if [ -n "${PIPE_CMD}" ]; then
```
thesamesam, line 1187: bash tests

eclass/kernel-2.eclass
```diff
@@ -1348,8 +1334,8 @@ unipatch() {
 			if [ -z "${PATCH_DEPTH}" ]; then PATCH_DEPTH=0; fi
```
thesamesam, line 1334: bash tests

mjeveritt (commented):
It can be useful to pad output to separate messages for clear visibility. This is somewhat of a cosmetic choice, and I'm not particularly for or against it personally.

eclass/kernel-2.eclass
mjeveritt: Hmmkay!

eclass/kernel-2.eclass
```diff
@@ -1579,13 +1560,12 @@ kernel-2_src_unpack() {
...
 
 kernel-2_src_prepare() {
-
 	debug-print "Applying any user patches"
 
 	# apply any user patches
 	case ${EAPI:-0} in
 		0|1|2|3|4|5) epatch_user ;;
-		6) eapply_user ;;
+		*) eapply_user ;;
```
thesamesam, line 1568: Let’s drop a bunch of ancient EAPIs given EROOT is only defined in 3+ anyway? Probably other uses too but can’t check yet

eclass/kernel-2.eclass
mjeveritt: `die` on the sed I assume....

eclass/kernel-2.eclass
```diff
@@ -1653,14 +1633,11 @@ kernel-2_pkg_postinst() {
 kernel-2_pkg_setup() {
 	if kernel_is 2 4; then
 		if [[ $(gcc-major-version) -ge 4 ]] ; then
```
thesamesam, line 1635: Inconsistent ;

eclass/kernel-2.eclass
mjeveritt: https://github.com/gentoo/gentoo/pull/20639#discussion_r624559875 ditto

eclass/kernel-2.eclass
mjeveritt: on the `find` damnit this is getting ridiculous :stuck_out_tongue_closed_eyes:

eclass/kernel-2.eclass
mjeveritt: oof good catch!

eclass/kernel-2.eclass
mjeveritt: `[[ ]]`

eclass/kernel-2.eclass
mjeveritt: ditto `[[ ]]`

dm0-:
I think I got everything.  There were a few more things that I thought were errors and fixed.

  - Like this not changing back to the previous directory: https://github.com/gentoo/gentoo/blob/master/eclass/kernel-2.eclass#L827-L830

  - Or these invalid array references: https://github.com/gentoo/gentoo/blob/master/eclass/kernel-2.eclass#L513

> It can be useful to pad output to separate messages for clear visibility. This is somewhat of a cosmetic choice, and I'm not particularly for or against it personally.

I'll let @SoapGentoo comment on his rejection of `echo`es.

eclass/kernel-2.eclass
```diff
@@ -1589,7 +1581,7 @@ kernel-2_src_prepare() {
 # conpile headers or run deblob script
 
 kernel-2_src_compile() {
-	cd "${S}"
+	cd "${S}" || die
 	[[ ${ETYPE} == headers ]] && compile_headers
 
 	if [[ $K_DEBLOB_AVAILABLE == 1 ]] && use deblob ; then
```
mjeveritt: `${K_DEBLOB_AVAILABLE}` I think was what was meant here ..

eclass/kernel-2.eclass
```diff
@@ -211,7 +211,7 @@ esac
 # Added by Daniel Ostrow <dostrow@gentoo.org>
 # This is an ugly hack to get around an issue with a 32-bit userland on ppc64.
 # I will remove it when I come up with something more reasonable.
-[[ ${PROFILE_ARCH} == "ppc64" ]] && CHOST="powerpc64-${CHOST#*-}"
+[[ ${PROFILE_ARCH} == ppc64 ]] && CHOST="powerpc64-${CHOST#*-}"
```
mjeveritt, line 214: I think ordinarily we tend to prefer `"${VAR}" == "str"` syntax ...

eclass/kernel-2.eclass
```diff
@@ -290,17 +290,17 @@ handle_genpatches() {
 
 		local use_cond_start="" use_cond_end=""
 
-		if [[ "${i}" == "experimental" && -z ${K_EXP_GENPATCHES_PULL} && -z ${K_EXP_GENPATCHES_NOUSE} ]] ; then
+		if [[ ${i} == experimental && -z ${K_EXP_GENPATCHES_PULL} && -z ${K_EXP_GENPATCHES_NOUSE} ]]; then
 			use_cond_start="experimental? ( "
```
mjeveritt, line 293: Again, for consistency's sake I think we'll stick to `""` quoting ..

eclass/kernel-2.eclass
dm0-: I did make that change.  It says this comment is on an outdated file.

eclass/kernel-2.eclass
```diff
@@ -415,7 +415,7 @@ detect_version() {
 	# We cannot trivally call kernel_is here, because it calls us to detect the
 	# version
 	#kernel_is ge 2 6 && RELEASE=${RELEASE/-pre/-git}
-	[ $(($KV_MAJOR * 1000 + ${KV_MINOR:-0})) -ge 2006 ] && RELEASE=${RELEASE/-pre/-git}
+	(( KV_MAJOR * 1000 + ${KV_MINOR:-0} >= 2006 )) && RELEASE=${RELEASE/-pre/-git}
 	RELEASETYPE=${RELEASE//[0-9]}
```
mjeveritt, line 418: Not sure about arithmetic expansions in bash, so deferring to another, but any reason this might fail?

eclass/kernel-2.eclass
```diff
@@ -907,7 +907,7 @@ compile_headers_tweak_config() {
 # Fix permissions in tarball
 
 install_universal() {
-	chown -R 0:0 "${WORKDIR}"/* >& /dev/null || die
+	chown -R 0:0 "${WORKDIR}"/* &>/dev/null || die
 	chmod -R a+r-w+X,u+w "${WORKDIR}"/* || die
```
mjeveritt, line 910: ooh, there's a classic typo! would probably preserve the space after `&>` for readability. </nitpick>

eclass/kernel-2.eclass
mjeveritt: just noticed that ! :)

mjeveritt:
@mpagano @Whissi @aliceinwire for final thoughts before ML review, etc

eclass/kernel-2.eclass
telans: Bash doesn't do any word splitting within `[[ ]]`. Quotes don't matter if they're there but it is redundant

eclass/kernel-2.eclass
a17r: In general we tell people to drop quotation inside ``[[ ]]``.

eclass/kernel-2.eclass
dm0-: Nearly all other instances of redirection in the file have no space, so I'd prefer to omit it for consistency rather than rewrite all the other lines, unless that is the coding standard.

dm0-:
> for final thoughts before ML review, etc

Note the EAPI 7 patch has not changed since [the last time I sent this to the mailing list](https://archives.gentoo.org/gentoo-dev/message/a032ddd8d54fb0d0797f855a69a4e612), which was never acknowledged.  This PR just has additional cleanup commits.

dm0-:
@SoapGentoo I had to revert that `emake` for reading variables that I mentioned earlier since it's apparently expected to fail sometimes and should not kill the build.

SoapGentoo:
@Whissi I'd like to have a hard deadline for this, and of not, we'll merge this

dm0-:
@mpagano I applied your patch from the bug here to run it through CI.  If you're a maintainer of this GitHub project, you should be able to modify the PR branch if needed.

PR #20642, opened 2021-05-01
net-misc/batctl: Version updated to 2021.0.

jpds:
Tested on `amd64`.


juippis (commented):
Please split adding a new version and removing an old one to separate commits. So this should have 2 commits instead of 1.

jpds:
@juippis Thanks for the review, done.

net-misc/batctl/batctl-2021.0.ebuild
```diff
@@ -0,0 +1,33 @@
+# Copyright 2020 Gentoo Authors
```
juippis: Should be updated. Repoman does this automatically for you if you use that to commit.

juippis (approved):

PR #20646, opened 2021-05-02
*/*: conditionally pass -DLINUX

thesamesam:


chewi:
While I'm very happy to not assume Linux everywhere, this seems like the wrong approach. The POSIX macro for detecting Linux is `__linux__` so if all `#ifdef` and `#if` lines have `LINUX` replaced with `__linux__` then it will just work without extra logic. Such patches could be sent upstream. It's not POSIX but `__unix__` is commonly defined and might be a better fit in some cases. [Here's a good list of macros for other operating systems](https://sourceforge.net/p/predef/wiki/OperatingSystems/). This does make me wonder what these projects will do when you don't define `LINUX`. Assume Windows?

PR #20653, opened 2021-05-02
net-misc/openssh: remove unused patch

mm1ke:


Zlogene:
@mm1ke you may want to run the global check, there are some patches left after libressl removal, but I left them up to maintainers to deal with

PR #20656, opened 2021-05-02
kde-apps/umbrello, kde-frameworks/kdelibs4support: remove unused patches

mm1ke:
Hi,

I guess the `kdelibs4support` patch was added accidentally since version 5.82.0 isn't released yet ;)


a17r (approved):
I'm going to pick the umbrello cleanup, but will leave kdelibs4support as new release is only a few days away.

PR #20662, opened 2021-05-03
vim-plugin.eclass: Replace has_version check with REPLACING_VERSIONS

mjsir911:
The intention is to show a message on first install, which ! has_version wasn't doing anymore.

Signed-off-by: Marco Sirabella <marco@sirabella.org>

### DIFF
--- a/eclass/vim-plugin.eclass
+++ b/eclass/vim-plugin.eclass
@@ -126,7 +126,7 @@ update_vim_afterscripts() {
 display_vim_plugin_help() {
 	local h
 
-	if ! has_version ${CATEGORY}/${PN} ; then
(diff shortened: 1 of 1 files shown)


mjsir911:
I see `app-vim/vim-latex` sort of duplicates this effort, but maybe works more properly?

https://github.com/gentoo/gentoo/blob/e2793693062c693368664e3fff03b21e4fc37f1b/app-vim/vim-latex/vim-latex-1.10.0-r2.ebuild#L56-L64

is `REPLACING_VERSIONS` what we want here instead of `has_version`?

laumann:
I think #24941 will be making the same fix as this PR.

mjsir911:
Actually was superseded by d2194b95 a while back lol but thanks @laumann for noticing!

PR #20664, opened 2021-05-03
games-action/*: EAPI7, eclass and other minor improvements

mm1ke:
Hi,

This udpates a few EAPI6 ebuilds towards EAPI7. Some notes:

* There were some ebuilds using `gnome2_icon_savelist`. I was reading the eclass about this and it was written that "This is only necessary for eclass implementations that call gnome2_icon_cache_update conditionally". I'm not really sure about this but since we don't call them conditionally we don't need this, do we?

* `games-action/brutal-legend` is has `~amd64 ~x86` keywords but doesn't have `-*` like other binary only games. Shouldn't we change that and add this too?

Please review


thesamesam:
1. Yeah, just use the xdg eclass (or xdg-utils if you prefer, but xdg is less work as it implements phases for us)
2. Yep, -* all the way.

mm1ke:
@thesamesam 
updated and rebased. I've updated the ebuilds to use the `xdg` eclass and also remove the phase functions completely.

ionenwks (commented):
Can't test most of these but looks good to me on paper, just 3 minor things I could see (one unrelated to your changes).

ionenwks:
Given trivial I went ahead and added the missing desktop, plus another commit wrt toolchain. Thanks!

mm1ke:
Thank you @ionenwks 
Sorry that i haven't updated my PR yet, i was afk the last few days ;)

PR #20672, opened 2021-05-03
x11-wm/herbstluftwm: add 0.9.3

Flowdalic:
New ebuild for 0.9.2 includes backported patches to
- save the HTML documentation in extra html/ directory
- add a missing include for GCC 11 compatibility

This also drops the 'examples' USE flag, as the examples are to small
to justify their own USE flag. Furthermore, app-text/asciidoc is only
needed when building from git, as the release tarballs of herbstluftwm
contain precompiled HTML files.

Closes: https://bugs.gentoo.org/787848


andrey-utkin (changes requested):
Please be consistent regarding revision bumps. Splitting the first commit into 2 or 3 parts should make that easier.

```
rename from x11-wm/herbstluftwm/herbstluftwm-0.9.2.ebuild
rename to x11-wm/herbstluftwm/herbstluftwm-0.9.2-r1.ebuild
```
So you've created the `r1`, but have removed the `r0`.
This is not what is suggested by the handbook and indeed the common sense in any of the cases: since it's a build failing bug, just make the change and don't increment the revision number, because nobody is supposed to want to go back.
If anybody can possibly want to go back to `r0`, then don't delete it.

Then, the rest of the changes in the first commit make things complicated as these are the things which usually warrant a revision increment so I guess that's why you did that.

* Address the build issue *only* without incrementing the revision number.
* Do the rest of the changes on that version's ebuild. The revision increment is appropriate for this. I would do that in two separate commits for clarity (but still with only one revision increment), but that's not critical to me and I leave this up to you.
* Apply that "sync 9999 with latest version ebuild". I personally prefer to apply things the other way around - whenever possible, I apply changes to 9999 first, because it has longer life and helps to dig up the history better. But I will accept what you have now in this case.

PR #20677, opened 2021-05-04
net-dns/avahi: use python-single-r1 instead of python-r1

thesamesam:
* Only builds one Python implementation => python-single-r1

* Fix REQUIRED_USE logic. We need to have it all together
  for python_gen_cond_dep, so we move the bookmarks? (...)
  dep within the python? ( ... ) one.

Closes: https://bugs.gentoo.org/710244
Closes: https://bugs.gentoo.org/788043

### DIFF
--- a/net-dns/avahi/avahi-0.8-r4.ebuild
+++ b/net-dns/avahi/avahi-0.8-r4.ebuild
@@ -5,7 +5,7 @@ EAPI="7"
 
 PYTHON_COMPAT=( python3_{7,8} )
 PYTHON_REQ_USE="gdbm"
-inherit autotools flag-o-matic multilib-minimal mono-env python-r1 systemd
+inherit autotools flag-o-matic multilib-minimal mono-env python-single-r1 systemd
 
 DESCRIPTION="System which facilitates service discovery on a local network"
 HOMEPAGE="http://avahi.org/"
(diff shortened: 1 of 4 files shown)


net-dns/avahi/avahi-0.8-r4.ebuild
```diff
@@ -166,18 +166,25 @@ multilib_src_compile() {
...
 
-	if multilib_is_native_abi && use doc; then
-		docinto html
-		dodoc -r doxygen/html/.
-		insinto /usr/share/devhelp/books/avahi
-		doins avahi.devhelp
+	if multilib_is_native_abi; then
+		if use python; then
+			python_optimize
+		fi
```
SoapGentoo, line 180: `use python && python_optimize`

net-dns/avahi/avahi-0.8-r4.ebuild
```diff
@@ -166,18 +166,25 @@ multilib_src_compile() {
...
 	# https://github.com/lathiat/avahi/issues/28
 	use howl-compat && dosym avahi-compat-howl.pc /usr/$(get_libdir)/pkgconfig/howl.pc
 	use mdnsresponder-compat && dosym avahi-compat-libdns_sd/dns_sd.h /usr/include/dns_sd.h
 
-	if multilib_is_native_abi && use doc; then
-		docinto html
-		dodoc -r doxygen/html/.
-		insinto /usr/share/devhelp/books/avahi
-		doins avahi.devhelp
+	if multilib_is_native_abi; then
```
SoapGentoo, line 177: consider moving this whole block into `_all`?

SoapGentoo (approved):

PR #20681, opened 2021-05-04
Enable python3_10 flags

mgorny:
CC @gentoo/python


thesamesam (approved):
That time already?

Thanks for doing split-arch masks, it’ll make life easier later on.

mgorny:
Yep. 3.10.0b1 was released today, and therefore 3.10 is now in feature freeze. Final is expected in September but we can start getting things rolling in ~arch already.

PR #20686, opened 2021-05-04
dev-ml/dose3: version bump for newer ocaml compat.

hlein:
Switched to the Dune build system by copy-paste from other ebuilds.
Caveats: I don't know the right way to make some components optional
with Dune build systems; therefore, everything that used to be
conditional on USE is now mandatory. Also, ocaml-base64 is only
KEYWORDS=~amd64, so as a result that is the only arch I could keep for
dose3-6 atm without repoman having a fit.

Closes: https://bugs.gentoo.org/769821


hlein:
Updated all opam packages that need a patch for dose3-6.x compat. But note, dose3-6.x depends on ocaml-base64 which has limited KEYWORDS, so all arches but ~amd64 are dropped in this round; see https://bugs.gentoo.org/788406

thesamesam (approved):
Excellent work, thank you so much for this!

~~Can you handle the rekeywording bug now? (All in one).~~ (I forgot we're handling it in [this](https://bugs.gentoo.org/788406) bug).

PR #20692, opened 2021-05-05
app-arch/snappy: update to 1.1.9

winterheart:
Taking maitainership, cleanup ebuild.


app-arch/snappy/snappy-1.1.9.ebuild
```diff
@@ -0,0 +1,46 @@
...
+HOMEPAGE="https://github.com/google/snappy"
+SRC_URI="https://github.com/google/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0/${PV%%.*}"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~mips ~ppc ~ppc64 ~s390 ~sparc ~x86 ~amd64-linux ~x86-linux ~x64-macos"
+IUSE="test"
+RESTRICT="!test? ( test )"
+
+BDEPEND="test? ( dev-cpp/gtest )"
```
Arfrever, line 19: `DEPEND` was correct here (as is in `snappy-1.1.8.ebuild`).
`dev-cpp/gtest` contains only headers and libraries, not executables.

winterheart: gtest is not runtime dependency, snappy does not require it for working, it's buildtime dependency.

Arfrever: Both `BDEPEND` and `DEPEND` are for build-time dependencies, needed in different locations. See e.g. [this](https://mgorny.pl/articles/the-ultimate-guide-to-eapi-7.html#improved-cross-compilation-support).
`RDEPEND` is for run-time dependencies.

`dev-cpp/gtest` should be in `DEPEND`.

app-arch/snappy/snappy-1.1.9.ebuild
```diff
@@ -0,0 +1,46 @@
...
+DOCS=( format_description.txt framing_format.txt NEWS README.md )
+
+PATCHES=(
+	"${FILESDIR}/${P}_gcc_inline.patch"
+	"${FILESDIR}/${P}_external_gtest.patch"
+)
+
+multilib_src_configure() {
+	local mycmakeargs=(
+		-DBUILD_SHARED_LIBS=ON
```
Arfrever, line 30: This is redundant, because the same value is already set in `cmake.eclass`.

app-arch/snappy/snappy-1.1.9.ebuild
```diff
@@ -0,0 +1,45 @@
...
+	"${FILESDIR}/${P}_external_gtest.patch"
+)
+
+multilib_src_configure() {
+	local mycmakeargs=(
+		-DSNAPPY_BUILD_TESTS=$(usex test)
+		-DSNAPPY_BUILD_BENCHMARKS=OFF
+		# we do not want to run benchmarks, and those are only used
+		# for benchmarks
+		-DHAVE_LIBZ=NO
```
juippis, line 34: zlib support?

app-arch/snappy/snappy-1.1.9.ebuild
```diff
@@ -0,0 +1,45 @@
...
+		-DHAVE_LIBLZO2=NO
+		-DHAVE_LIBLZ4=NO
```
juippis, line 36: lz4 compression support? Why not make these toggleable?

app-arch/snappy/snappy-1.1.9.ebuild
mgorny: Read the comment above.

juippis: Mmhm, somehow I connected it with the build option above the comment. I'll update the comment.

juippis (approved):

app-arch/snappy/snappy-1.1.9.ebuild
```diff
@@ -0,0 +1,45 @@
...
+
+CMAKE_ECLASS=cmake
+inherit cmake-multilib
+
+DESCRIPTION="A high-speed compression/decompression library by Google"
+HOMEPAGE="https://github.com/google/snappy"
+SRC_URI="https://github.com/google/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0/${PV%%.*}"
```
juippis, line 14: Not sure if this is necessary,
```
  SONAME               libsnappy.so.1
```

PR #20693, opened 2021-05-05
gui-apps/slurp: add version 1.3.2

tobiasjakobi:
Bug tracker entry for reference: https://bugs.gentoo.org/771966

This is the 1.2.0 ebuild with some minor modifications.
- I changed the keywords to ~x86 and ~amd64 as I can't test on arm64 system
- added missing libxkbcommon dependency (verified by inspecting the meson.build of the project)

Build and runtime check done. Compilation was done with LTO and without.


tobiasjakobi:
Just added the Bug tag and my Signed-off-by.

tobiasjakobi:
Added the distfile to the Manifest

yamabiiko:
Maybe we should add [please reassign] to the pull title to force it? The maintainer does not seem active to me

tobiasjakobi:
> Maybe we should add [please reassign] to the pull title to force it? The maintainer does not seem active to me

I already did this once, but apparantly that only triggered the rescan. So I should add the suffix again?

yamabiiko:
I don't know, but it seems it needs a new maintainer. If you feel confident and have time for it, you could add yourself as co-maintainer to the Manifest file according to  [wiki](https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide#Taking_over_an_existing_package). Otherwise I could try to do it too

tobiasjakobi:
@yamabiiko I'm not sure I'm really qualified for that. My knowledge about ebuild is very superficial. And time is another thing...

PR #20707, opened 2021-05-06
media-video/pipewire: Bump to 0.3.27 and slight improvements.

pinkflames:
This PR backports some of the changes from the jack-sdk/multilib PR while bumping to 0.3.27.

**Please note that group ID -1 for pipewire-desktop is a placeholder and a real one must be requested via Gentoo Data API system before committing!**


thesamesam:
One commit per package please. We can assign the UID/GID on merge (it’s somewhere in the proxy maint  guide)

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -165,7 +166,16 @@ src_compile() {
...
 		# Start of ${limitsdfile} from ${P}
 
-		1000:60000	-	memlock 256
+		# Required to memlock buffers for pipewire-pulse & JACK emulation
+		@pipewire-desktop	-	memlock	256
+
+		# May be lowered/removed in the future; PW runs at 20 via rtkit
+		#@pipewire-desktop	-	rtprio	19
+
+		# Should match Red Hat defaults for their realtime group
```
Whissi, line 175: No, we will not overwrite settings for realtime group in this package. `USE=jack-sdk` will probably pull in `sys-auth/realtime-base`.

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -95,7 +96,7 @@ PATCHES=(
 
 # limitsdfile related code taken from =sys-auth/realtime-base-0.1
 # with changes as necessary.
-limitsdfile=40-${PN}.conf
+limitsdfile=00-${PN}.conf
```
Whissi, line 99: No, we will stick to `40-` to stay in sync with other package. Also, you cannot go below `00-` so this wouldn't be an option at all.

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -41,6 +41,7 @@ BDEPEND="
 	)
 "
 RDEPEND="
+	acct-group/pipewire-desktop
```
Whissi, line 44: Like discussed in IRC, I'll probably prefer using `audio` group which is already in use by everyone following handbook. No final decision yet.

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -7,7 +7,7 @@ inherit meson optfeature udev
 
 if [[ ${PV} == 9999 ]]; then
 	EGIT_REPO_URI="https://gitlab.freedesktop.org/${PN}/${PN}.git"
-#	EGIT_BRANCH="master"
+	EGIT_BRANCH="master"
```
Whissi, line 10: Is `EGIT_BRANCH` needed at all?

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -229,10 +239,15 @@ pkg_postinst() {
 		elog "by commenting out the relevant command near the end of the file:"
 		elog "#\"/usr/bin/pipewire\" = { args = \"-c pipewire-pulse.conf\" }"
 		elog
-		elog "It is still necessary to manually enable PipeWire startup. Setup specific"
-		elog "instructions can be found at: https://wiki.gentoo.org/wiki/PipeWire"
+		ewarn "${EROOT}/etc/xdg/autostart/pipewire.desktop has been installed. Users of XDG"
+		ewarn "compliant desktops on OpenRC must not manually start pipewire anymore!"
+		ewarn "Users wishing to use PulseAudio must remove the file manually and add the"
```
Whissi, line 244: We do not advertise `INSTALL_MASK` feature. So please remove the last two `ewarn` messages.

media-video/pipewire/pipewire-9999.ebuild
```diff
@@ -201,15 +211,15 @@ pkg_postinst() {
 		elog
 	fi
 
-	elog "JACK emulation is incomplete and not all programs will work. PipeWire's"
```
Whissi, line 204: We shouldn't display all of this all the time when pipewire is getting merged to disk. Users tend to ignore it with the result to miss really important messages once we need to communicate them. See https://github.com/gentoo/gentoo/blob/13ec4ca6975058cf8cd810da440d1e42618b7986/sys-apps/lm-sensors/lm-sensors-3.6.0.ebuild#L172-L215 to get an idea what I mean.

Whissi (changes requested):

media-video/pipewire/pipewire-9999.ebuild
pinkflames: Because we install 00-foo.conf while the realtime-base installs 40-foo.conf, our @realtime settings will be replaced by that file's higher values - the intention is to provide slightly less insane settings for users who do not already have a pro-audio setups (frankly I don't like rtprio 70 either but as you discovered at least it's what Red Hat installs and it's not 99). With further testing I hope we can settle on no more than rtprio 49 but first we need JACK applications that attempt to schedule themselves at FIFO or RR priority levels in the first place.

media-video/pipewire/pipewire-9999.ebuild
pinkflames: But the entire point is that any other file will have higher priority and will override our settings. The higher the number, the higher the priority (but only within the matching rule type e.g. default vs default, group vs. group or UID vs UID). Would you prefer `10-` at least?

media-video/pipewire/pipewire-9999.ebuild
pinkflames: As I explained multiple times (and as you verified in your own tests), audio group should not be used at all for non-system daemons (which we do not support on Gentoo!). All it does is interfere with Access Control List maintained by either elogind or systemd-logind causing unpleasant experiences such as no access to audio devices when another logged in user has taken them, including that users audio being played back even though they have lost the active seat.

pinkflames:
> One commit per package please. We can assign the UID/GID on merge (it’s somewhere in the proxy maint guide)

Yeah, I remember about -1 but I wanted a functional file to test the behavior on my system. I'll fix that to -1 after a rebase for the acct-group/realtime.

media-video/pipewire/pipewire-9999.ebuild
pinkflames: Right, I missed that during backporting. Sorry.

media-video/pipewire/pipewire-9999.ebuild
pinkflames: Thank you for the example, if you really insist, I will implement that but I would like to point out that not printing messages will not make them be read more. ;)

media-video/pipewire/pipewire-9999.ebuild
pinkflames: You get a warning otherwise, and if someone was using the old ebuild, that one set the branch to some outdated one. I'm not sure whether the new ebuild would switch over to default master automatically or keep using the old branch value with the new `GIT_REPO_URI`, so better safe than sorry?

mattst88:
> After implementing README.gentoo via readme.gentoo-r1.eclass, I have decided that it's too horrible to look at or commit, and that I'll rather spend my effort elsewhere than on the ebuilds for media-video/pipewire.

Really?

Whissi:
Closed via commit 183aaa2dc9100c576b4dc7d158a3c91f362279c8.

@pinkflames I am sorry that you are no longer interested. Feel free to always ping us/continue querying me via IRC whenever you notice something which can be improved. Even if we disagree sometimes, your feedback is appreciated. Thank you for your past work on pipewire!

pinkflames:
Yeah, I will. Godspeed to PipeWire and Gentoo.

PR #20709, opened 2021-05-06
media-libs/mesa: Switch to main branch on live ebuild

FireBurn:
Mesa has now switched to using the "main" branch

### DIFF
--- a/media-libs/mesa/mesa-9999.ebuild
+++ b/media-libs/mesa/mesa-9999.ebuild
@@ -16,6 +16,7 @@ HOMEPAGE="https://www.mesa3d.org/ https://mesa.freedesktop.org/"
 
 if [[ ${PV} == 9999 ]]; then
 	EGIT_REPO_URI="https://gitlab.freedesktop.org/mesa/mesa.git"
+	EGIT_BRANCH="main"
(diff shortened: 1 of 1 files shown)


media-libs/mesa/mesa-9999.ebuild
```diff
@@ -16,6 +16,7 @@ HOMEPAGE="https://www.mesa3d.org/ https://mesa.freedesktop.org/"
 
 if [[ ${PV} == 9999 ]]; then
 	EGIT_REPO_URI="https://gitlab.freedesktop.org/mesa/mesa.git"
+	EGIT_BRANCH="main"
```
Arfrever, line 19: Why would this be needed?
`git-r3.eclass` does not hardcode branch name `master`.

ionenwks: Indeed, I don't define EGIT_BRANCH and use git-r3 on a few packages with "main" and there's no issues as-is.

FireBurn: Ah I thought it was hard coded, but I can see it uses HEAD

FireBurn:
Closing, as this isn't required

PR #20714, opened 2021-05-07
app-crypt/mkp224o: New package

00-matt:
Currently in ::guru.


app-crypt/mkp224o/mkp224o-1.5.0.ebuild
```diff
@@ -0,0 +1,54 @@
...
+EAPI=7
+
+DESCRIPTION="Vanity address generator for v3 Tor hidden service addresses"
+HOMEPAGE="https://github.com/cathugger/mkp224o"
+SRC_URI="https://github.com/cathugger/mkp224o/releases/download/v1.5.0/mkp224o-${PV}-src.tar.gz"
+
+LICENSE="CC0-1.0"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="amd64-51-30k amd64-64-24k besort binsearch +donna donna-sse2 pcre2 ref10 +statistics"
```
00-matt, line 13: With USE flags being used to select the ed25519 implementation, users can pick an implementation that doesn't support their arch.

Should I mask the USE flags in profiles to prevent this, or is there a better way?

PR #20724, opened 2021-05-07
dev-python/{django,google-api-python-client,nbformat}: remove unused patches

mm1ke:


mgorny (approved):
Thanks. I'll merge it once I'm done building new kernels (in ~45 mins).

PR #20727, opened 2021-05-08
Fix Rust

dm0-:
Here is a first attempt at solving some of the problems with Rust in Gentoo.  I haven't tested everything here yet.  The overall gist is to support arbitrary targets without caring where they came from, like `rust-std` with crossdev, the `RUST_CROSS_TARGETS` variable, or custom builds.  It solves two main issues.

  - Rust in Gentoo currently requires multilib ABI `USE` flags in dependencies, which is incompatible with EAPI 7 and later.  Rust must be declared in `BDEPEND` in EAPI 7+, but `BROOT` and `SYSROOT` can have different architectures or use a `no-multilib` profile, so it will end up with impossible dependencies.

This removes the individual multilib requirements and adds a single `multilib` flag to `dev-lang/rust` instead.  It is enabled or disabled in profiles depending on whether they support multilib or not, just like GCC and glibc.  This makes Rust work the same as other compilers like GCC and Go, where they can be used for any supported target without breaking Gentoo dependencies across roots.

  - Gentoo-supported Rust targets are very limited, which prevents the distro from building consistently (or at all) for many systems.  Even with the supported targets or those installed separately, there is no easy way to change the target used by ebuilds.

This standardizes on the `RUST_TARGET` variable, which is set in each profile with a corresponding Rust builtin target.  It is handled correctly by multilib by supporting different targets per ABI, and cargo ebuilds will now successfully cross-compile.  (`RUST_HOST` is also supported and defaults to `RUST_TARGET`, but it should be customized when cross-compiling.)  Users can override the value in `make.conf` or the environment to change targets as needed.  These targets accept arbitrary values, which supports using completely custom targets.

Some issues:
  - Maybe different variables should be used like `RBUILD`/`RHOST`/`RTARGET` to correspond to the `C` versions so it's less confusing.  `RUST_TARGET` corresponds to `CHOST`, and `RUST_HOST` corresponds to `CBUILD`.  These were picked because they're what Mozilla uses.
  - Mozilla projects need to be patched to use these environment variables, which is not included in this PR.  (I.e. Firefox, spidermonkey, etc. won't use the given targets out of the box.)  The patches can be stolen from Debian.
  - I haven't really done anything with `rust-bin`.

### DIFF
--- a/profiles/default/linux/x86/17.0/musl/make.defaults
+++ b/profiles/default/linux/x86/17.0/musl/make.defaults
@@ -1,5 +1,7 @@
-# Copyright 1999-2018 Gentoo Foundation
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 CHOST="i686-gentoo-linux-musl"
+RUSTHOST="i686-unknown-linux-musl"
 CHOST_x86="${CHOST}"
+RUSTHOST_x86="${RUSTHOST}"
--- a/dev-lang/rust-bin/rust-bin-1.59.0.ebuild
+++ b/dev-lang/rust-bin/rust-bin-1.59.0.ebuild
@@ -63,7 +63,7 @@ src_unpack() {
 
 	default_src_unpack
 
-	mv "${WORKDIR}/${MY_P}-$(rust_abi)" "${S}" || die
+	mv "${WORKDIR}/${MY_P}-${RUSTHOST}" "${S}" || die "Unsupported target binary: ${RUSTHOST}"
 }
 
 patchelf_for_bin() {
@@ -184,7 +184,7 @@ multilib_src_install() {
 
 	else
 		local rust_target
-		rust_target="$(rust_abi $(get_abi_CHOST ${v##*.}))"
+		rust_target="$(get_abi_RUSTHOST ${v##*.})"
 		dodir "/opt/${P}/lib/rustlib"
 		cp -vr "${WORKDIR}/rust-${PV}-${rust_target}/rust-std-${rust_target}/lib/rustlib/${rust_target}"\
 			"${ED}/opt/${P}/lib/rustlib" || die
--- a/dev-lang/rust-bin/rust-bin-1.60.0.ebuild
+++ b/dev-lang/rust-bin/rust-bin-1.60.0.ebuild
@@ -68,7 +68,7 @@ src_unpack() {
 
 	default_src_unpack
 
-	mv "${WORKDIR}/${MY_P}-$(rust_abi)" "${S}" || die
+	mv "${WORKDIR}/${MY_P}-${RUSTHOST}" "${S}" || die "Unsupported target binary: ${RUSTHOST}"
 }
 
 patchelf_for_bin() {
@@ -189,7 +189,7 @@ multilib_src_install() {
 
 	else
 		local rust_target
-		rust_target="$(rust_abi $(get_abi_CHOST ${v##*.}))"
+		rust_target="$(get_abi_RUSTHOST ${v##*.})"
 		dodir "/opt/${P}/lib/rustlib"
 		cp -vr "${WORKDIR}/rust-${PV}-${rust_target}/rust-std-${rust_target}/lib/rustlib/${rust_target}"\
 			"${ED}/opt/${P}/lib/rustlib" || die
--- a/dev-lang/rust-bin/rust-bin-1.61.0.ebuild
+++ b/dev-lang/rust-bin/rust-bin-1.61.0.ebuild
@@ -72,7 +72,7 @@ src_unpack() {
 
 	default_src_unpack
 
-	mv "${WORKDIR}/${MY_P}-$(rust_abi)" "${S}" || die
+	mv "${WORKDIR}/${MY_P}-${RUSTHOST}" "${S}" || die "Unsupported target binary: ${RUSTHOST}"
 }
 
 patchelf_for_bin() {
@@ -193,7 +193,7 @@ multilib_src_install() {
 
 	else
 		local rust_target
-		rust_target="$(rust_abi $(get_abi_CHOST ${v##*.}))"
+		rust_target="$(get_abi_RUSTHOST ${v##*.})"
 		dodir "/opt/${P}/lib/rustlib"
 		cp -vr "${WORKDIR}/rust-${PV}-${rust_target}/rust-std-${rust_target}/lib/rustlib/${rust_target}"\
 			"${ED}/opt/${P}/lib/rustlib" || die
--- a/dev-lang/rust-bin/rust-bin-1.62.0.ebuild
+++ b/dev-lang/rust-bin/rust-bin-1.62.0.ebuild
@@ -70,7 +70,7 @@ src_unpack() {
 
 	default_src_unpack
 
-	mv "${WORKDIR}/${MY_P}-$(rust_abi)" "${S}" || die
+	mv "${WORKDIR}/${MY_P}-${RUSTHOST}" "${S}" || die "Unsupported target binary: ${RUSTHOST}"
 }
 
 patchelf_for_bin() {
@@ -191,7 +191,7 @@ multilib_src_install() {
 
 	else
 		local rust_target
-		rust_target="$(rust_abi $(get_abi_CHOST ${v##*.}))"
+		rust_target="$(get_abi_RUSTHOST ${v##*.})"
 		dodir "/opt/${P}/lib/rustlib"
 		cp -vr "${WORKDIR}/rust-${PV}-${rust_target}/rust-std-${rust_target}/lib/rustlib/${rust_target}"\
 			"${ED}/opt/${P}/lib/rustlib" || die
--- a/dev-lang/rust/rust-1.59.0.ebuild
+++ b/dev-lang/rust/rust-1.59.0.ebuild
@@ -248,7 +248,7 @@ pkg_setup() {
 src_prepare() {
 	if ! use system-bootstrap; then
 		local rust_stage0_root="${WORKDIR}"/rust-stage0
-		local rust_stage0="rust-${RUST_STAGE0_VERSION}-$(rust_abi)"
+		local rust_stage0="rust-${RUST_STAGE0_VERSION}-${RUSTHOST}"
 
 		"${WORKDIR}/${rust_stage0}"/install.sh --disable-ldconfig \
 			--without=rust-docs --destdir="${rust_stage0_root}" --prefix=/ || die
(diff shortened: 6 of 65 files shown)


Arfrever:
`multilib` USE flag in `sys-devel/gcc` and `sys-libs/glibc` is something deprecated, which still exists only because nobody had time to work on replacing it with `abi_*` USE flags.

dm0-:
> `multilib` USE flag in `sys-devel/gcc` and `sys-libs/glibc` is something deprecated, which still exists only because nobody had time to work on replacing it with `abi_*` USE flags.

The only way to move forward with multilib ABIs in compilers is to ban depending on them (or officially declare that Gentoo no longer supports cross-compilation).  It might not be an issue with glibc since that shouldn't be in `BDEPEND`, and maybe it won't matter for gcc since things don't tend to depend on it directly, but that is not the case for Rust.

dm0-:
> I like the idea of `RUST_HOST/RHOST` in profiles instead of that rigid list in rust_abi(), that will let users willingly override it to arbitrary values as long as they have capable compiler.
> 
> https://bugs.gentoo.org/747760 also relevant here and maybe it should be mixture of both?

I am indifferent to new ways to build more targets.  This PR is just about selecting targets, and I think these two concepts should remain separate (i.e. selecting a target cannot have the package depend on that target being built--at least considering portage's current capabilities--so it should work like GCC and `CHOST`).  As long as that is true, defining another list of targets to build should be fine.

My concern with that method would be that it is unmanageable to maintain enough target `USE` flags to reflect the settings that Gentoo is capable of building, since Rust targets include combinations of any number of the CPU architecture, operating system, CPU flags, bitness, ABI, and probably other weird stuff.  So the `USE` target list would be restricted to only supporting a few common targets, forcing some users to ignore it and continue using existing options to build for their odd platforms.  For example, a case like the `arm` keyword would need at least the following 13 builtin Linux targets available to be compatible with Gentoo profiles, and these lists can be different for each Rust version (e.g. `i386` and `i486` Linux builtin targets were just added a release or two ago).
  - armv4 × glibc
  - armv5 × (glibc + musl)
  - armv6 × (glibc + musl) × (softfloat + hardfloat)
  - armv7 × (glibc + musl) × (softfloat + hardfloat × (nosimd + `CPU_FLAGS_ARM="neon thumb"`))

dm0-:
> > ```
> > * Maybe different variables should be used like `RBUILD`/`RHOST`/`RTARGET` to correspond to the `C` versions so it's less confusing.  `RUST_TARGET` corresponds to `CHOST`, and `RUST_HOST` corresponds to `CBUILD`.  These were picked because they're what Mozilla uses.
> > ```
> 
> Ugh. That's very ugly. I would very much prefer if we use towards our users and in our config files the same semantics as with gcc (but I dont know how complex that will make the internal code).
> 
> This would mean
> RUST_TARGET=RHOST
> RUST_HOST=RBUILD
> (Are you sure this is correct?)

Yes, it would probably mean having an eclass export `RUST_TARGET` for compatibility.  Upstream projects like librsvg can use it from the environment like that, and other distros (Debian, Void) are patching projects to do the same, e.g. [spidermonkey](https://salsa.debian.org/gnome-team/mozjs/-/blob/debian/78/master/debian/patches/Bug-1545437-Add-options-to-specify-Rust-target-name.patch).

eclass/cargo.eclass
```diff
@@ -348,7 +351,7 @@ cargo_src_compile() {
 
 	tc-export AR CC CXX PKG_CONFIG
 
-	set -- cargo build $(usex debug "" --release) ${ECARGO_ARGS[@]} "$@"
+	set -- cargo build $(usex debug "" --release) --target "${RUST_TARGET}" ${ECARGO_ARGS[@]} "$@"
```
gyakovlev, line 354: please keep it as part of toml file, populated in `cargo_gen_config`
command line argument has highest priority and cargo bails out (usually) if same arg/param specified twice.
so passing it here prevents possibly passing it in ebuild without modifying the eclass.

refer to my patch https://gist.github.com/gyakovlev/f9bdcb59ceb3741f411ad23ebe0feed4


also note, specifying target for cargo changes output directory

`target/release` becomes `target/${RUST_TARGET}/release`

that's why I create stable symlink as target/gentoo

also it should handle USE=debug (that's not in my patch)

so stable symlink should be something like that

`target/${RUST_TARGET}/$(usex debug debug release)` `->` `target/gentoo`

and ebuilds just install files from `target/gentoo`

at least those ebuilds are affected, maybe more.
```
rg target/release --files-with-matches
gui-libs/greetd/greetd-0.6.1.ebuild
gui-libs/greetd/greetd-0.7.0.ebuild
net-libs/quiche/quiche-9999.ebuild
net-libs/quiche/quiche-0.8.1.ebuild
net-libs/quiche/quiche-0.7.0.ebuild
net-libs/quiche/quiche-0.8.0.ebuild
net-libs/quiche/quiche-0.5.1.ebuild
net-libs/quiche/quiche-0.6.0.ebuild
sys-apps/bat/bat-0.17.1-r1.ebuild
sys-apps/ripgrep/ripgrep-12.1.1.ebuild
dev-java/oracle-javamail/oracle-javamail-1.5.6-r1.ebuild
app-crypt/nitrocli/nitrocli-0.3.4.ebuild
app-crypt/nitrocli/nitrocli-0.4.0.ebuild
app-crypt/nitrocli/nitrocli-0.3.5.ebuild
```

gyakovlev (changes requested):
adding some initial `cargo.eclass` comments

eclass/cargo.eclass
gyakovlev: lost some parts of my comment...
I see you've updated `quiche`, but it still may require updating for USE=debug

dm0-:
@gyakovlev Maybe it will be easier to drop `cargo.eclass` from this PR.  It could implement cross-compilation separately from selecting targets.

@akhuettel I changed the variables to `RUSTBUILD` and `RUSTHOST` to look like `RUSTFLAGS`.  I could use `RHOST` or whatever if that is preferable.

gyakovlev:
thanks for updating it.
`RUSTHOST` looks nicer than already used `RUST_TARGET` and is better named imo.

I'll give it a test, still bit busy, and maybe will start merging profile changes.

I prefer functional approach to emit variables, usually, so need to think if it would be better to make `rust_abi()` `echo ${RUSTHOST}` value instead of direct assignment in ebuilds, as it's convenient place to error-out or transform values if needed. also less ebuilds to edit =)

Some magic variable that's assumed to be present is simple, but may backfire.
opinions on that?

dm0-:
> I prefer functional approach to emit variables, usually, so need to think if it would be better to make `rust_abi()` `echo ${RUSTHOST}` value instead of direct assignment in ebuilds, as it's convenient place to error-out or transform values if needed. also less ebuilds to edit =)

I don't feel too strongly about it, but I prefer a bare variable especially due to multilib supporting for different targets for different ABIs.  I think it's easier to follow when it works the same way as other multilib variables without other abstractions, but I don't encounter `rust-toolchain.eclass` usage often, so maybe it won't be an issue.  (I should probably also make `RUSTFLAGS` a multilib variable for ABI-specific settings.)

> Some magic variable that's assumed to be present is simple, but may backfire.
> opinions on that?

My thought was that `RUSTHOST` would be used exactly like `CHOST` and assumed to be everywhere, except undefined by default only for systems that have no builtin Rust target, but those systems would continue to have Rust packages masked in profiles.  I think it should be considered an error for `RUSTHOST` to not be present when building for a supported system, since it can specify CPU settings that may be required to produce executable binaries.

gyakovlev:
hope to find time this weekend, have not forgotten.

dm0-:
I pushed a commit to make `RUSTFLAGS` configurable per ABI (tested with `RUSTFLAGS="-C target-cpu=native" RUSTFLAGS_x86="-C target-cpu=geode"`) and rebased to fix merge conflicts.

dev-lang/rust/rust-1.51.0-r2.ebuild
```diff
@@ -41,7 +41,7 @@ LLVM_TARGET_USEDEPS=${ALL_LLVM_TARGETS[@]/%/(-)?}
 
 LICENSE="|| ( MIT Apache-2.0 ) BSD-1 BSD-2 BSD-4 UoI-NCSA"
 
-IUSE="clippy cpu_flags_x86_sse2 debug doc miri nightly parallel-compiler rls rustfmt system-bootstrap system-llvm test wasm ${ALL_LLVM_TARGETS[*]}"
+IUSE="clippy cpu_flags_x86_sse2 debug doc miri multilib nightly parallel-compiler rls rustfmt system-bootstrap system-llvm test wasm ${ALL_LLVM_TARGETS[*]}"
```
mgorny, line 44: Is there a very good reason to support non-multilib builds in the first place?

dm0-: I don't think so, aside from wasting time building unusable code in `no-multilib` profiles.  I just tried to make the interface consistent with glibc, since the behavior is to build more std libs for additional ABIs.

mgorny: Well, I don't think that there's anything stopping you from controlling multilib via profile without extra flags. In fact, `get_all_abis` should be doing already, I think.

dm0-:
@SoapGentoo The branch is rebased, and I applied your librsvg changes on the EAPI 7 commit from this branch.  I didn't get a chance to do thorough testing with the new versions, but some basic builds look okay.

profiles/default/linux/x86/17.0/musl/make.defaults
```diff
@@ -1,5 +1,7 @@
-# Copyright 1999-2018 Gentoo Foundation
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 CHOST="i686-gentoo-linux-musl"
+RUSTHOST="i686-unknown-linux-musl"
```
SoapGentoo, line 5: instead of introducing `RUSTHOST` as a variable into all profiles, couldn't we just rewrite `CHOST` on the fly? In practice it seems to just be about turning all tuples into Debian/Ubuntu ones?

SoapGentoo (changes requested):

stefson:
I will give the pullrequest a testride on armv7, please have some patience for the results its super slow :D

profiles/default/linux/x86/17.0/musl/make.defaults
SoapGentoo: sure, but we can remap those host tuples there? having a single mathematical map that translates CHOST/CTARGET to the correct rust values is far preferable to me than RUSTHOST spread all over the profiles, so let's go that way.

dm0-: I don't think I can continue developing this branch.  Someone else can take the PR from here if it is useful.  You can run `grep -FarB1 RUSTHOST profiles` to see which values don't match (all arm and riscv profiles and some of the mips ABIs).

mattst88:
librsvg-2.55.1 is now using EAPI=8 since I had to switch to using cargo.eclass.

Would be happy to see someone work on this PR again.

mattst88:
Sorry, I have no idea.

stefson:
@mattst88 you can patch the values into the cargo.eclass file, works perfectly: 

file /usr/armv7a-unknown-linux-gnueabihf/usr/lib/librsvg-2.so.2.48.0
```
/usr/armv7a-unknown-linux-gnueabihf/usr/lib/librsvg-2.so.2.48.0: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked, stripped
```

arthurzam:
I'm guessing this is obsolete already.

PR #20728, opened 2021-05-08
www-apps/netbox: Bump to 2.10.5 with py support upgrade

alarig:


thesamesam:
netbox was treecleaned a while ago. If stilil interested, we can look at bringing it back though (but PR would need rebasing).

PR #20730, opened 2021-05-08
games-fps/quakeforge: Fix config error

weshmek:
Closes: https://bugs.gentoo.org/649088
Closes: https://bugs.gentoo.org/681192
Closes: https://bugs.gentoo.org/713168
Signed-off-by: Wesley Chalmers <wwchalme@uwaterloo.ca>

### DIFF
--- a/games-fps/quakeforge/files/quakeforge-0.7.2-p20210508-fix-makefile-includes.patch
+++ b/games-fps/quakeforge/files/quakeforge-0.7.2-p20210508-fix-makefile-includes.patch
@@ -1,0 +1,11 @@
+--- a/tools/qfcc/test/Makefile.am
++++ b/tools/qfcc/test/Makefile.am
+@@ -144,7 +144,7 @@ voidfor.dat: $(voidfor_obj) $(QFCC_DEP)
+ voidfor.run: Makefile build-run
+ 	$(srcdir)/build-run $@
+ 
+-include ./$(DEPDIR)/*.Qo
++-include ./$(DEPDIR)/*.Po
+ 
+ EXTRA_DIST= test-bi.h build-run
+ CLEANFILES= *.dat *.sym *.qfo *.run


weshmek:
I was having an issue building quakeforge on my machine, and I found that this patch resolved it.
So far as I can tell, this is an error in the 0.7.2 and earlier branches of the repo, and does not exist in the master branch

ionenwks:
The fix itself is fine but I honestly don't know what to do with this version of quakeforge, it's segfaulting left and right and hardly usable for me either way. Is it working right for you?

[Upstream](https://github.com/quakeforge/quakeforge) snapshot I tried works fine, but it seems rough around the edges with build failures and flags combinations (vulkan especially.. even if I disable it).

chewi:
It segfaults for me too. This is pretty ancient though so I think we should be looking at the snapshot, even if it is a bit rough. We could stick to a build configuration that works. Also upstream is still active so we could report bugs.

I think DarkPlaces is the better port. That's also still active upstream but there haven't been any releases and our snapshot is years behind. I used to play it so maybe I'll give it a poke.

ionenwks:
I used to play it as well (albeit moreso the original than quakeforge), already mostly-cleaned up the ebuild locally but fell apart when I tried to test run it and pondering what to do next.

Since you're getting segfaults too, I'll have a closer look at using a snapshot.

Haven't tried darkplaces yet, but it's where I'd have gone if I wanted to play again.

weshmek:
I looked into the segfaults and found that disabling code optimizations fixes it....
...and then I join DM servers and I'm unable to fire my weapon (although that appears to be a server-side issue, near as I can tell).
Of all the Quake engines I've tried, I like this one the least.

ionenwks:
Quakeforge can be "*okay*" after you set it up, but it's very not user friendly with horrible defaults. I do recommend to use something else.

That aside, snapshot from a year ago (before vulkan branch got merged) seems best to me, optimizations are no issue too -- it also doesn't need this Makefile fix (so won't need this PR, but thanks still).

I'll still update it given some its tools can be useful, but probably shouldn't be playing using it.

PR #20735, opened 2021-05-09
dev-libs/apr: fix underlinking for atomics on ppc, sparc

thesamesam:
Fixes errors in the produced library like:
> undefined reference to `__sync_val_compare_and_swap_8'

Forcing linkage against libatomic didn't work. May be worth
investigation in future but need to get ppc and sparc working
again.

Closes: https://bugs.gentoo.org/740464


thesamesam:
@Polynomial-C

PR #20738, opened 2021-05-09
sys-devel/binutils{,-hppa64}, sys-libs/binutils-libs: remove unused patches

mm1ke:


trofi:
Pushed as is. Thank you!

PR #20739, opened 2021-05-09
games-engines/frobtads: Bump version to cmake-based 2.0, remove old

realnc:
FrobTADS 2.0 uses cmake. Ebuild has been adapted accordingly. The `tads2compiler` and `tads3compiler` USE flags are enabled by default now, to reflect upstream defaults. Note that the primary use of this package is to actually get the TADS compilers, not the command-line interpreter. So having those disabled by default was a bit silly to begin with.

1.2.4 is removed, as it fails to build nowadays with recent GCC versions.


realnc:
Ping.

thesamesam:
Thanks!

PR #20740, opened 2021-05-09
dev-python/*: Port to python3.10

jsmolic:


dev-python/sniffio/sniffio-1.2.0.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7..9} )
+PYTHON_COMPAT=( python3_{7..10} )
 inherit distutils-r1
 
 DESCRIPTION="Sniff out which async library your code is running under"
```
mgorny, line 9: Convert to `epytest`.

mgorny (changes requested):

dev-python/sniffio/sniffio-1.2.0.ebuild
jsmolic: done

dev-python/fastimport/fastimport-0.9.12.ebuild
```diff
@@ -13,7 +13,6 @@ SRC_URI="mirror://pypi/${PN::1}/${PN}/${P}.tar.gz"
 
 LICENSE="GPL-2+"
 SLOT="0"
-KEYWORDS=""
```
mgorny, line 16: Did you actually test this version?

jsmolic: Yeah, tests passed on my end successfully

mgorny:
Merging all but fastimport.

dev-python/fastimport/fastimport-0.9.12.ebuild
mgorny: But you do realize that this version has no keywords?

jsmolic: Yup, I was just about to add 3.10 to 0.9.11 as well

PR #20749, opened 2021-05-10
net-wireless/wavemon-0.9.3: fix ncurses(w) libs

mehw:
Fixes a segmentation fault due to linking with libncursesw and libtinfo, rather than libtinfow.

See also https://github.com/uoaerg/wavemon/issues/88 about the segmentation fault error.

If libncursesw is available uses ncursesw_LIBS, otherwise uses ncurses_LIBS if libncurses remains the only option.

Closes: https://bugs.gentoo.org/785562
Signed-off-by: Matteo Bianco <mehw.is.me@inventati.org>


alarig:
The CI doesn’t speak about wavemon, so this PR looks good to me

mehw:
> The CI doesn’t speak about wavemon, so this PR looks good to me

@alarig thanks for checking.

juippis (approved):
I'm going to revbump the 0.9.3 since you mentioned this is a runtime issue rather than a build issue. Thanks!

PR #20751, opened 2021-05-10
app-editors/juffed: bump snapshot version

arthurzam:
- added the github repo to `metadata.xml`
- moved `S=` closer to `SRC_URI=` (as stated in `skel.ebuild`).
- Take latest (2020-01-03) snapshot of code. This snapshot incorporates most patches that were used with old version.
- One patch, which fixed `qscintilla` library name, was removed to be set from configure step (using `-DQSCINTILLA_NAMES`). Works better that way.
- Removed versioned dependency of `qscintilla`, as we don't have such old version in tree.
- Removed the `[qt5(+)]` USE dependency as all those libraries don't have such USE flag.
- added `dev-qt/linguist-tools:5` dependency, as the bug report states (it is used during build).

After bump, removed old version (unstable only) and unused patches.


arthurzam:
bump

app-editors/juffed/juffed-0.10_p20200103.ebuild
```diff
@@ -0,0 +1,53 @@
...
+
+EAPI=7
+
+COMMIT=74ab7236a30be17351edc5b83eb3579affd96913
+inherit cmake xdg
+
+DESCRIPTION="QScintilla-based tabbed text editor with syntax highlighting"
+HOMEPAGE="http://juffed.com/en/"
+SRC_URI="https://github.com/Mezomish/${PN}/archive/${COMMIT}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/Mezomish-${PN}-74ab723"
```
mgorny, line 12: How about replacing the inline short commit id with cutting it out of `${COMMIT}`?

app-editors/juffed/juffed-0.10_p20200103.ebuild
```diff
@@ -0,0 +1,53 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+COMMIT=74ab7236a30be17351edc5b83eb3579affd96913
+inherit cmake xdg
+
+DESCRIPTION="QScintilla-based tabbed text editor with syntax highlighting"
+HOMEPAGE="http://juffed.com/en/"
+SRC_URI="https://github.com/Mezomish/${PN}/archive/${COMMIT}.tar.gz -> ${P}.tar.gz"
```
mgorny, line 11: Please inline the PN here.

app-editors/juffed/juffed-0.10_p20200103.ebuild
```diff
@@ -0,0 +1,53 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+COMMIT=74ab7236a30be17351edc5b83eb3579affd96913
+inherit cmake xdg
+
+DESCRIPTION="QScintilla-based tabbed text editor with syntax highlighting"
+HOMEPAGE="http://juffed.com/en/"
```
mgorny, line 10: Probably worth including the github repo URL for convenience.

app-editors/juffed/juffed-0.10_p20200103.ebuild
```diff
@@ -0,0 +1,53 @@
...
+SRC_URI="https://github.com/Mezomish/${PN}/archive/${COMMIT}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/Mezomish-${PN}-74ab723"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc ~ppc64 ~x86"
+IUSE="debug"
```
mgorny: Where is it being used?

mgorny (changes requested):

mgorny:
GitHub doesn't seem to like you:

```
!!! Fetched file: juffed-0.10_p20200103.tar.gz VERIFY FAILED!
!!! Got:      1628731
!!! Expected: 1628670
```

arthurzam:
> GitHub doesn't seem to like you:
> 
> ```
> !!! Fetched file: juffed-0.10_p20200103.tar.gz VERIFY FAILED!
> !!! Reason: Filesize does not match recorded size
> !!! Got:      1628731
> !!! Expected: 1628670
> ```

I refetched the archive, and hope it would be the same at your side. Else, we have a problem.
Thanks for noting it.

mgorny:
```
 * ERROR: app-editors/juffed-0.10_p20200103::gentoo failed (prepare phase):
 *   The source directory '/tmp/portage/app-editors/juffed-0.10_p20200103/work/Mezomish-juffed-74ab723' doesn't exist
```

Told you that commit id stripping looks suspicious to me. Didn't you have local snapshot or sth?

PR #20762, opened 2021-05-10
media-libs/opencv: version bump to 4.5.2

rossbridger:
Relevant outstanding bugs:

Bug: https://bugs.gentoo.org/787998
Bug: https://bugs.gentoo.org/788712
Bug: https://bugs.gentoo.org/783102


a17r (changes requested):
This reintroduces dependency on `virtual/lapacke` which was dropped in de5c0d7b596a13d9b4fdc14a0e9160f0da797092
Also you are changing virtual/jre and virtual/jdk dependency (has implications on how it is built) and slot op (looks to be a mistake), maybe @vaukai or @fordfrog can tell you more about it. See also https://archives.gentoo.org/gentoo-dev/message/71ec6a505f3245496d343c506ad21717

media-libs/opencv/opencv-4.5.2.ebuild
```diff
@@ -0,0 +1,569 @@
...
+		media-libs/libdc1394:=[${MULTILIB_USEDEP}]
+		sys-libs/libraw1394[${MULTILIB_USEDEP}]
+	)
+	java? ( virtual/jre:= )
+	jpeg? ( virtual/jpeg:0[${MULTILIB_USEDEP}] )
+	jpeg2k? ( media-libs/openjpeg:2=[${MULTILIB_USEDEP}] )
+	lapack? (
+		virtual/cblas
+		virtual/lapack
+		virtual/lapacke
```
epsilon-0, line 120: Please try building without having `virtual/lapacke` installed and if it works remove this dependency.

vaukai:
> maybe @vaukai or @fordfrog can tell you more about it. See also https://archives.gentoo.org/gentoo-dev/message/71ec6a505f3245496d343c506ad21717

nothing to add, `1.8:*` should be tried in first place.
```
>=virtual/jdk-1.8:*
>=virtual/jre-1.8:*
```

steils:
I'm not sure any of these three bugs are relevant to the version upgrade. Also I'm not able to reproduce them with 4.5.1 either (but I don't have a no-mutilib system or a system with nvidia, so I could really try only the bug 787998).
I suggest not to mention them in the commit message. Perhaps close the bugs as obsolete. If some of them are still actual for 4.5.2, they can be reopened.

PR #20772, opened 2021-05-11
dev-lang/rust-{bin}: request add ARCH=riscv support

dlan17:
this series will enable ARCH=riscv support for dev-lang/rust,
current it only enable abi=riscv64/lp64d support,

I've tested both rust and rust-bin on sifive unleashed machine.

it's possible to support riscv32, but haven't tested due to less interests and man power..


gyakovlev:
otherwise LGTM, with `use.force` changes.

but let's wait for infra to fix CI so we can see the report.
I suspect it needs more masks.

gyakovlev:
also seems `virtual/rust` needs a keyword

gyakovlev (approved):

gyakovlev:
it has been pushed, closing.

PR #20781, opened 2021-05-12
www-servers/pound: various changes (read first comment)

scardracs:
Bump to version 3.0

Unfortunately this release drop support for alpha, hppa, ppc and sparc because a required dep (dev-libs/nanomsg) is not supported for these architectures.

Bug: https://bugs.gentoo.org/714084
Closes: https://bugs.gentoo.org/657942
Closes: https://bugs.gentoo.org/527278
Closes: https://bugs.gentoo.org/657946
Closes: https://bugs.gentoo.org/674064


www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,57 @@
...
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~mips ~x86"
+
+DEPEND="
+	dev-libs/libpcre
+	dev-libs/libyaml
+	dev-libs/nanomsg
+	dev-libs/openssl
+	net-libs/mbedtls
```
juippis, line 22: Since these are needed in `DEPEND` they're most likely linked while compiling. You should define a subslot binder to these libs where such is available.

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,57 @@
...
+"
+
+RDEPEND="${DEPEND}"
+
+S="${WORKDIR}/${MY_P}"
+
+src_compile() {
+	rm -r GPL.txt || die
```
juippis, line 30: This operation does not belong in `src_compile`.

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	cmake_src_compile
+}
+
+src_install() {
+	dodir /usr/sbin
+	cp "${BUILD_DIR}"/pound "${ED}"/usr/sbin/ || die
```
juippis, line 36: `dosbin` / `newsbin`?
https://devmanual.gentoo.org/function-reference/install-functions/

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	doman "${S}"/man/pound.8
+	dodoc README.md
+
+	dodir /etc/init.d
+	newinitd "${FILESDIR}"/pound.init-1.9 pound
+
+	insinto /etc
+	newins "${FILESDIR}"/pound-2.2.cfg pound.cfg
+}
```
juippis, line 45: What would `cmake_src_install` do here? I'd imagine you could replace many of these steps with that single call.

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,57 @@
...
+
+pkg_postinst() {
+	elog "No demo-/sample-configfile is included in the distribution; read the man-page for more info."
+	elog "A sample (localhost:8888 -> localhost:80) for gentoo is given in \"/etc/pound.cfg\"."
+	ewarn
+	ewarn "You will have to upgrade you configuration file, if you are"
+	ewarn "upgrading from a version <= 2.0."
```
juippis, line 52: This seems useless, as <2.0 hasn't existed in the tree during git-era at all.

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	ewarn
+	ewarn "The 'WebDAV' config statement is no longer supported!"
+	ewarn "Please adjust your configuration, if necessary."
```
juippis, line 55: and I wonder if this is a similar case.

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,57 @@
...
+EAPI=7
+
+inherit cmake
+
+MY_P=${P/p/P}
+DESCRIPTION="A http/https reverse-proxy and load-balancer"
+HOMEPAGE="https://www.apsis.ch/pound.html"
+SRC_URI="http://www.apsis.ch/pound/${MY_P}.tgz"
+
+LICENSE="GPL-3"
```
juippis, line 13: Are you sure it's GPL-3? Please verify by reading the included license file.

juippis, line 13: Then their homepage says:

> the uthash library. A copy is included in the Pound distribution, but you may want to download the most up-to-date version. The copyright notice is here
> the hpack library. A copy is included in the Pound distribution, with adjustments for Linux compatability. The copyright is held by Reyk Floeter (BSD license).

So you'll have to include the licenses for these bundled libs too, **if** they're being used in here (please verify).

www-servers/pound/files/pound.init-1.9
```diff
@@ -1,5 +1,5 @@
 #!/sbin/openrc-run
-# Copyright 1999-2004 Gentoo Foundation
+# Copyright 1999-2021 Gentoo Authors
```
juippis, line 2: I don't know why you're making this change, as nothing else is touched in this file.

www-servers/pound/pound-2.7f-r1.ebuild
```diff
@@ -13,9 +13,11 @@ LICENSE="GPL-3"
 SLOT="0"
 KEYWORDS="~alpha amd64 ~hppa ~mips ~ppc ~sparc x86"
 
-DEPEND="dev-libs/libpcre
+DEPEND="
+	dev-libs/libpcre
```
juippis, line 17: And this seems rather useless too?

juippis (commented):
Hey, I know you're working on a security bump, but let's do it properly so no one has to fix it after.

www-servers/pound/pound-3.0.ebuild
juippis: Also I want to point out that pound-2.7f installs two binaries into /usr/sbin, you're only installing one.

scardracs: On 3.0 there is only one binary file

www-servers/pound/pound-3.0.ebuild
scardracs: Fixed

www-servers/pound/pound-3.0.ebuild
scardracs: Fixed

www-servers/pound/pound-2.7f-r1.ebuild
scardracs: Why?

www-servers/pound/pound-3.0.ebuild
scardracs: The GPL.txt says GPL-3

scardracs: Yes they are used

www-servers/pound/pound-3.0.ebuild
scardracs: It does not work because of -> ninja: error: unknown target 'install'

www-servers/pound/files/pound.init-1.9
scardracs: I haven't changed it, repoman did it

www-servers/pound/pound-3.0.ebuild
scardracs: Fixed

www-servers/pound/pound-3.0.ebuild
scardracs: Now it should be fixed

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,51 @@
...
+
+EAPI=7
+
+inherit cmake
+
+MY_P="${P/p/P}"
+
+DESCRIPTION="A http/https reverse-proxy and load-balancer"
+HOMEPAGE="https://www.apsis.ch/pound.html"
+SRC_URI="http://www.apsis.ch/pound/${MY_P}.tgz"
```
juippis, line 12: `https://`?

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,51 @@
...
+LICENSE="BSD GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~mips ~x86"
+
+DEPEND="
+	dev-libs/libpcre:=
+	dev-libs/libyaml:=
+	dev-libs/nanomsg:=
+	dev-libs/openssl:=
+	net-libs/mbedtls:=
```
juippis, line 23: Well, not **every** lib here has a subslot defined... but it's fine, at least it's better now.

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,51 @@
...
+"
+
+RDEPEND="${DEPEND}"
+
+S="${WORKDIR}/${MY_P}"
+
+DOCS=( README.md )
```
juippis, line 30: Does this get installed? You seem to be missing `einstalldocs` call.

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,51 @@
...
+
+src_compile() {
+	cmake_src_compile
+}
+
+src_install() {
+	rm -r GPL.txt || die
```
juippis, line 37: `-r`?

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,51 @@
...
+
+RDEPEND="${DEPEND}"
+
+S="${WORKDIR}/${MY_P}"
+
+DOCS=( README.md )
+
+src_compile() {
+	cmake_src_compile
+}
```
juippis, line 34: Do you need to define this phase with *cmake.eclass* inherited? I have a feeling not.

www-servers/pound/pound-3.0.ebuild
```diff
@@ -0,0 +1,51 @@
...
+	dodir /etc/init.d
+	newinitd "${FILESDIR}"/pound.init-1.9 pound
+
+	insinto /etc
+	newins "${FILESDIR}"/pound-2.2.cfg pound.cfg
+}
+
+pkg_postinst() {
+	elog "No demo-/sample-configfile is included in the distribution; read the man-page for more info."
+	elog "A sample (localhost:8888 -> localhost:80) for gentoo is given in \"/etc/pound.cfg\"."
```
juippis, line 50: Wrap earlier, 
https://devmanual.gentoo.org/ebuild-writing/messages/index.html

www-servers/pound/pound-3.0.ebuild
juippis: I wonder if it works with `CMAKE_MAKEFILE_GENERATOR="emake"`

juippis: (note that I haven't actually checked the upstream build system at all, which would tell this instantly)

www-servers/pound/pound-3.0.ebuild
scardracs: Fixed

www-servers/pound/pound-3.0.ebuild
scardracs: The CMakeLists.txt does not offer any install way, only compilation

www-servers/pound/pound-3.0.ebuild
scardracs: Done

www-servers/pound/pound-3.0.ebuild
scardracs: If I don't define it it fails saying `/var/tmp/portage/www-servers/pound-3.0/temp/environment: line 492: cmake_src_compile: command not found`

www-servers/pound/pound-3.0.ebuild
scardracs: Done, it was a refuse

www-servers/pound/pound-3.0.ebuild
scardracs: Done

www-servers/pound/pound-3.0.ebuild
scardracs: I leave it as is due to that

www-servers/pound/pound-3.0.ebuild
juippis: That doesn't sound right, but I'll test it then.

juippis (approved):

PR #20783, opened 2021-05-12
QPL and QPL-1.0 are duplicates

ulm:


ulm:
> _Status_: x **broken**

No it isn't.

Zlogene:
ack from @gentoo/tex

PR #20786, opened 2021-05-12
Add CPE tag for app-emulation/runc.

vaibhav-rustagi:
Based on the CPE tags from official CPE tag database
official-cpe-dictionary_v2.3.xml downloaded from
https://nvd.nist.gov/products/cpe, it uses
linuxfoundation as the vendor for runc vulnerabilities.

### DIFF
--- a/app-emulation/runc/metadata.xml
+++ b/app-emulation/runc/metadata.xml
@@ -26,5 +26,6 @@
 	</use>
 	<upstream>
 		<remote-id type="github">opencontainers/runc</remote-id>
+		<remote-id type="cpe">cpe:/a:linuxfoundation:runc</remote-id>
 	</upstream>
 </pkgmetadata>


vaibhav-rustagi:
I don't think adding the CPE tag in app-emulation/runc resulted in any of the failure specified in the report above. All those failures should be present before this commit too.

@gyakovlev: Do you have any suggestion for this?

gyakovlev:
I'll check, but we had outage yesterday, maybe related.

gyakovlev:
ah you are missing `Signed-off-by:` line in commit message. Please add it.

vaibhav-rustagi:
@gyakovlev: Thanks for checking that. I updated the commit with the "Signed-off-by:" tag.

gyakovlev:
thanks, pushed.
I think I took care of issues in next commit, it's unrelated to your change.

PR #20791, opened 2021-05-13
Fix LICENSE on clockspeed and checkpassword

lanodan:


juippis (approved):

net-mail/checkpassword/checkpassword-0.90-r4.ebuild
```diff
@@ -9,7 +9,8 @@ DESCRIPTION="A uniform password checking interface for root applications"
 HOMEPAGE="https://cr.yp.to/checkpwd.html"
 SRC_URI="https://cr.yp.to/checkpwd/${P}.tar.gz"
 
-LICENSE="all-rights-reserved"
+# http://cr.yp.to/distributors.html
+LICENSE="public-domain"
```
eli-schwartz, line 13: If you're going to remove all-rights-reserved dedications you should also check for outdated RESTRICTs, see https://bugs.gentoo.org/444424#c16

PR #20794, opened 2021-05-13
sys-apps/exa: bump to v0.10.1

dAgbeve:


dAgbeve:
@a17r  Could you look at this also?, Thanks

dAgbeve:
@thesamesam

dAgbeve:
Version 0.10.1 of `sys-apps/exa` uses `app-text/pandoc` to build `man pages`  and since `app-text/pandoc` does not support `ppc64` keyword, it had to be removed to meet `QA` requirement. The tests passed in this release.

@thesamesam Kindly have a look at this, any recommendation and improvement is welcome. Thanks

sys-apps/exa/exa-0.10.1.ebuild
```diff
@@ -0,0 +1,130 @@
...
+tinyvec_macros-0.1.0
+"
+
+inherit bash-completion-r1 cargo
+
+DESCRIPTION="A modern replacement for 'ls' written in Rust"
+HOMEPAGE="https://the.exa.website/"
+SRC_URI="$(cargo_crate_uris ${CRATES})"
+
+LICENSE="Apache-2.0 MIT Unlicense"
```
juippis, line 82: Are these all licensed used by this package? Please check with *cargo-license* package, each update.

sys-apps/exa/exa-0.10.1.ebuild
```diff
@@ -0,0 +1,130 @@
...
+DEPEND="git? ( dev-libs/libgit2:= )"
+RDEPEND="${DEPEND}"
+BDEPEND+="man? ( >=app-text/pandoc-2.11.3 )"
+
+QA_FLAGS_IGNORED="/usr/bin/exa"
+
+src_prepare() {
+	default
+	if use man
+	then
```
juippis, line 96:
```suggestion
	if use man; then
```

sys-apps/exa/exa-0.10.1.ebuild
```diff
@@ -0,0 +1,130 @@
...
+		mkdir -p contrib/man && \
+		pandoc --standalone -f markdown -t man man/exa_colors.5.md \
+		-o contrib/man/exa_colors.1 && \
+		pandoc --standalone -f markdown -t man man/exa.1.md -o contrib/man/exa.1 && \
+		rm -rf man || die "failed to creat directory"
```
juippis, line 101: If you chain these commands like this, it's impossible to know which part specifically died. So please just separate them,,
```suggestion
		mkdir -p contrib/man || die
		pandoc --standalone -f markdown -t man man/exa_colors.5.md -o contrib/man/exa_colors.1 || die
		pandoc --standalone -f markdown -t man man/exa.1.md -o contrib/man/exa.1 || die
		rm -r man || die
```
We don't use `rm -f` if the outcome can be predicted.

juippis (commented):
Also there's a weird commit for gcr?

sys-apps/exa/exa-0.10.1.ebuild
dAgbeve: Yes, gcr is lingering in this commit and should be ignored. A further check on the licensing shows that it uses MIT, so the needed update has been made.

sys-apps/exa/exa-0.10.1.ebuild
```diff
@@ -0,0 +1,130 @@
...
+tinyvec_macros-0.1.0
+"
+
+inherit bash-completion-r1 cargo
+
+DESCRIPTION="A modern replacement for 'ls' written in Rust"
+HOMEPAGE="https://the.exa.website/"
+SRC_URI="$(cargo_crate_uris ${CRATES})"
+
+LICENSE="MIT"
```
juippis, line 82: Noo you misunderstood, we need to include licenses for **all** bundled crates as well in here. And that's where we use *cargo-license* package to check all licenses.
https://packages.gentoo.org/packages/dev-util/cargo-license

dAgbeve: Right, just did and the earlier licenses were all there is, so updated it. Thanks for the `education`.

juippis (approved):

juippis:
11df9df46ba64e423ae7a005abaf98b762b70e31
e68ffa34eb4846cfcaec58874cd374cd83842681
1aa094ce87be782275704e21490dfc45cfc252c7

PR #20795, opened 2021-05-13
dev-python/*: bump some packages to 3.10

vaartis:
The tests pass for these.


mgorny:
Have you tested it? I'm pretty sure I've tried pytest-asyncio and it had test failures.

vaartis:
I was sure I did, but you are correct, terribly sorry, it does indeed have test failures, I have removed it from the PR. I have also removed the old gunicorn version that was just removed from the tree. I've rerun all the remaining packages' tests again just now and they have passed on my computer.

mgorny:
Thanks. I've merged most of 'em, however gunicorn fails tests for me:
```
>   from eventlet.wsgi import ALREADY_HANDLED as EVENTLET_ALREADY_HANDLED 
E   ImportError: cannot import name 'ALREADY_HANDLED' from 'eventlet.wsgi' (/usr/lib/python3.7/site-packages/eventlet/wsgi.py)
```

I'm going to guess it's incompatible with my eventlet version. It doesn't strictly block py3.10 but blocks me from testing it ;-). I'll try to take a look at it later but if you could try to reproduce it and find a solution (probably putting `<` version restriction on eventlet), that'd help a lot.

vaartis:
Unfortunately it seems that eventlet depends on dnspython which depends on ecds which depends on gmpy, the latter of which seems to really dislike python 3.10 and may need a version bump to the latest beta version (because the current version in tree is from 2016). I'll try looking into it and will get back to you with the results.

vaartis:
I have updated the gmpy ebuild with a new version, which still had to be slightly patched (explained in the ebuild). Had to comment out some weird doctests that didn't make sense and failed, otherwise the tests now pass for both python 3.8 and 3.10. I'll try to get the rest of gunicorn dependencies now.

vaartis:
@mgorny Unfortunately it seems that eventlet itself doesn't work on 3.10 and requires major work: eventlet/eventlet#687. I did however successfully bump ecdsa and dnspython while working on eventlet and tests pass for them. Had to patch dnspython 1.16.0 for 3.10 a bit, so I bumped the revision number. Seems like gunicorn will have to wait, perhaps for its own PR that would update both eventlet and gevent.

vaartis:
Ah, seems like despite the tests passing some dependencies aren't resolved. Looking into it.

vaartis:
So, dnspython needed cryptography. And the current version of cryptography doesn't work with 3.10. However, the problem in tests has been fixed 2 days ago (https://github.com/pyca/cryptography/commit/1b922ed1dee0cd7e165a868639ce6d0869c8b2f5). I made a snapshot package for this commit, and it turned out that cryptography now _requires_ rust and will not build without it. They use a package called `setuptools_rust`, which I have also added and made it RDEPEND on virtual/rust. 

I imagine after the next release of cryptography, this snapshot package can be easily converted back to using pypi (as of now it downloads a tarball from github).

All the tests pass for cryptography using rust :) (at least for me) and dnspython has all dependencies resolved (at least pkgcheck says so)

vaartis:
I think i've mostly figured out the keywording issues. Unfortunately I had to drop some keywords from cryptography and only leave those supported by rust.

mgorny:
> So, dnspython needed cryptography. And the current version of cryptography doesn't work with 3.10. However, the problem in tests has been fixed 2 days ago ([pyca/cryptography@1b922ed](https://github.com/pyca/cryptography/commit/1b922ed1dee0cd7e165a868639ce6d0869c8b2f5)). I made a snapshot package for this commit, and it turned out that cryptography now _requires_ rust and will not build without it. They use a package called `setuptools_rust`, which I have also added and made it RDEPEND on virtual/rust.

Did they actually start using Rust? Last I checked it was just boilerplate. In either case, depending on Rust for Python 3.10 is not an option as it effectively kills half of supported arches. Could you try backporting the fix instead? It seems relatively simple, so even if it doesn't apply as-is rebasing it sould be easy.

vaartis:
> Did they actually start using Rust? Last I checked it was just boilerplate. In either case, depending on Rust for Python 3.10 is not an option as it effectively kills half of supported arches. Could you try backporting the fix instead? It seems relatively simple, so even if it doesn't apply as-is rebasing it sould be easy.

Yes, they do use rust. I'll try backporting it, though simply applying the diff didn't work last time I tried.

vaartis:
@mgorny you were correct, i guess I should have looked into porting it more thoroughly. I have modifed their commit a bit and it applied and the tests now pass. The same tests also pass for 3.8 even with the patch. I have bumped the ebuild to r1. 

Now, concerning the new snapshot version: i imagine we will have to use it sooner or later. Should I leave it in somehow, perhaps mask it or something? I imagine that would save some time later when someone will inevitably have to write a ebuild for the new version.

dev-python/gmpy/files/gmpy-2.1.0_beta5-test-input.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/test/runtests.py b/test/runtests.py
+index 5e5842d..7d64e52 100644
+--- a/test/runtests.py
++++ b/test/runtests.py
+@@ -81,7 +81,7 @@ if sys.version.startswith('3.1.'):
+     print("with Python 3.1. The doctest module in Python 3.2 and later does not")
+     print("have this issue.")
+     print()
+-    input("Press ENTER to continue.. ")
++    # input("Press ENTER to continue.. ")
```
mgorny, line 10: Don't add comments in patch, that's just inflating it unnecessarily.

dev-python/gmpy/files/gmpy-2.1.0_beta5-pyhash-nan.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/src/gmpy2_hash.c b/src/gmpy2_hash.c
```
mgorny, line 1: Is this an upstream patch? Please add a link or a comment where it comes from.

dev-python/dnspython/dnspython-1.16.0-r2.ebuild
```diff
@@ -23,6 +23,11 @@ RDEPEND="dev-python/pycryptodome[${PYTHON_USEDEP}]
 	!dev-python/dnspython:py2
 	!dev-python/dnspython:py3"
 
+PATCHES=(
+	# Mutable mapping was moved to a different module in 3.10 and removed from the old one
+	"${FILESDIR}"/abc-mutable-mapping.patch
```
mgorny, line 28: Start the patch name with `${P}-`, I'd go for simple `${P}-py310.patch`.

dev-python/cryptography/cryptography-3.4.7-r1.ebuild
```diff
@@ -41,6 +41,10 @@ DEPEND="
 	"
 RDEPEND+=${DEPEND}
 
+PATCHES=(
+	"${FILESDIR}/well-defined-reprs.patch"
```
mgorny, line 45: Same here.

mgorny (changes requested):
Let's avoid adding cryptography snapshots prematurely. I'd suggest moving setuptools_rust and the cryptography snapshot (so you don't lose your work) to a separate draft PR, and we'll look into it when a proper release is made. Well, unless I get so angry I decide to spit in upstream's face and fork cryptography.

dev-python/gmpy/files/gmpy-2.1.0_beta5-pyhash-nan.patch
vaartis: There is a comment in the ebuild explaining the patch. The upstream doesn't work with python 3.10 because the constant was removed in 3.10. Should probably submit the patch upstream, but that'll only be in the next release.

dev-python/gmpy/files/gmpy-2.1.0_beta5-test-input.patch
vaartis: Removed the comment

dev-python/dnspython/dnspython-1.16.0-r2.ebuild
vaartis: Done

dev-python/cryptography/cryptography-3.4.7-r1.ebuild
vaartis: Done

dev-python/gmpy/files/gmpy-2.1.0_beta5-pyhash-nan.patch
mgorny: The point is that we generally prefer patches blessed upstream, or at least submitted somewhere official where we can figure out what's happening.

vaartis: I'll try submitting it upstream and see what they say.

vaartis: It also seems that the patch that I made that disables the failings tests have been resolved upstream already, at least so said in https://github.com/aleaxit/gmpy/issues/296 but it's not in a release yet.

mgorny: Find the upstream commit then.

vaartis: It was this commit: https://github.com/aleaxit/gmpy/commit/db7ce2ef46fab84e7b9c32b05725e9b02e8cf206. The patch did't apply cleanly so I had to manually copy-past from the commit, because too much changed in a ~year since the last release, for it to apply, plus some functions were renamed.

mgorny: Cool, I see you've backported more changes and made things work ;-).

dev-python/semantic_version/semantic_version-2.8.5.ebuild
```diff
@@ -13,7 +13,7 @@ SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
 
 LICENSE="BSD"
 SLOT="0"
-KEYWORDS="amd64 arm arm64 ~ppc ~ppc64 ~sparc x86"
+KEYWORDS="amd64 arm arm64 ~ppc ~ppc64 ~riscv ~sparc x86"
```
mgorny, line 16: Just to confirm, did you test it on some RISCV hardware or emulator?

mgorny:
I'm going to skip semantic_version here since it is failing tests for me:

```
ERROR tests/test_django.py::FieldMigrationTests::test_nondefault_spec_field - RuntimeError: setup_test_environment() was already called and can...
ERROR tests/test_django.py::FieldMigrationTests::test_spec_field - RuntimeError: setup_test_environment() was already called and can't be calle...
ERROR tests/test_django.py::FieldMigrationTests::test_version_field - RuntimeError: setup_test_environment() was already called and can't be ca...
[...]
```

I'm guessing django version incompatibility? Could you try with a few django versions, figure out an appropriate `<` test-dep and submit it separately?

mgorny:
dnspython-2 tests fail too:
```
FAILED tests/test_tokenizer.py::TokenizerTestCase::testStdinFilename - AssertionError: '<file>' != '<stdin>'
```

dev-python/semantic_version/semantic_version-2.8.5.ebuild
vaartis: Initially added the keyword for the rust update if I remember correctly, but it doesn't matter now. It's a very simple python package that _should_ just work everywhere. That said, I'll refrain from touching keywords from now on and will remove this when submitting the fixed version.

mgorny: I've seen so many weird platform-specific Python failures I won't say 'should' ever again ;-).

vaartis:
> dnspython-2 tests fail too:
> 
> ```
> FAILED tests/test_tokenizer.py::TokenizerTestCase::testStdinFilename - AssertionError: '<file>' != '<stdin>'
> ```

That's strange. it works fine for me..
> I'm guessing django version incompatibility? Could you try with a few django versions, figure out an appropriate < test-dep and submit it separately?

I'll try looking into it, but it seems like django hasn't been ported to 3.10 yet and it has some complicated dependencies like numpy.

vaartis:
FYI: https://github.com/aleaxit/gmpy/pull/297 (the 3.10 fix for gmpy) got merged upstream

PR #20796, opened 2021-05-13
mail-client/neomutt: Do not call `cc`

nicolasbock:
Closes: https://bugs.gentoo.org/724716
Signed-off-by: Nicolas Bock <nicolasbock@gentoo.org>


thesamesam (approved):
Looks good, thank you! I'll add the change to the 9999* ebuild too

PR #20797, opened 2021-05-13
profiles: info_vars add more variables, sorting

Alessandro-Barbieri:

Bug: https://bugs.gentoo.org/722456
Closes: https://bugs.gentoo.org/780477


Alessandro-Barbieri:
Ping

profiles/info_vars
```diff
@@ -7,35 +7,69 @@
...
 CONFIG_PROTECT
 CONFIG_PROTECT_MASK
+CONFIG_SHELL
+CPP
 CPPFLAGS
 CTARGET
 CXXFLAGS
 DISTDIR
-FEATURES
+DWP
```
thesamesam, line 30: What is this?

profiles/info_vars
```diff
@@ -7,35 +7,69 @@
...
 ACCEPT_LICENSE
+ADDR2LINE
+AR
+ARFLAGS
+AS
+ASFLAGS
 CC
+CCLD
 CXX
+CXXFILT
```
thesamesam, line 18: I'm not convinced this needs to be in every single emerge --info call.

profiles/info_vars
```diff
@@ -7,35 +7,69 @@
 ##
 ACCEPT_KEYWORDS
 ACCEPT_LICENSE
+ADDR2LINE
+AR
+ARFLAGS
+AS
+ASFLAGS
 CC
+CCLD
```
thesamesam, line 16: People are really typically setting this? ditto several others

profiles/info_vars
thesamesam: .. oh, these are only shown if set. I guess it's okay.

profiles/info_vars
thesamesam: I'll drop this as I have no idea what it is.

PR #20803, opened 2021-05-14
media-sound/mumble: Update live ebuild

DarkDefender:
Mumble added a new sub-module that provides some cmake files to find python.
Their cmake files now expects this submodule to be present as cmake will error out otherwise as it can't find files that belongs to this module.

@Polynomial-C


DarkDefender:
Hmm, I don't see any new breakages caused by the new commit.
Perhaps the CI report is wrong? :thinking:

DarkDefender:
@Polynomial-C poke.
I'm sure you are busy with other stuff, but I just want to make sure that you are still alive.

PR #20806, opened 2021-05-14
Gluster bumps

jkroonza:


jkroonza:
Changes from ionen incorporated, along with tmpfiles.d fixes (please refer updated changelog).

sys-cluster/glusterfs/glusterfs-9.2.ebuild
```diff
@@ -0,0 +1,195 @@
...
+	elog "You can now treat glusterfsd2 like any other service"
+	echo
+	ewarn "You need to use a ntp client to keep the clocks synchronized across all"
+	ewarn "of your servers. Setup a NTP synchronizing service before attempting to"
+	ewarn "run GlusterFS."
+	echo
+	elog "If you are upgrading from a previous version of ${PN}, please read:"
+	elog "  http://docs.gluster.org/en/latest/Upgrade-Guide/upgrade_to_$(ver_cut '1-2')/"
+
+	tmpfiles_process
```
juippis, line 188: I'm getting 
```
 * ERROR: sys-cluster/glusterfs-9.2::gentoo failed (postinst phase):
 *   tmpfiles_process: Must specify at least one filename
 * 
 * Call stack:
 *     ebuild.sh, line  125:  Called pkg_postinst
 *   environment, line 2230:  Called tmpfiles_process
 *   environment, line 3437:  Called die
 *       [[ ${#} -gt 0 ]] || die "${FUNCNAME}: Must specify at least one filename";
```
In every run, with 8.5 and 9.2. I can't see how this would be your fault, but... it doesn't look like this call is doing much in here. at least for me. @ionenwks care to test too?

ionenwks: Before it was already installing tmpfiles.d files but without processing them. This `tmpfiles_process` is new and doesn't have automagic to find new .conf files, so they need to be specified.

If a single gluster.conf the only thing it can install, then this should be simple.

On a side-note, the init script seems to still be using /var/run/gluster while .conf is /run/gluster (it's normally a symlink but who knows with users' systems).

jkroonza: I'm a little uncertain of this particular line:

```
101         --with-tmpfilesdir="${EPREFIX}"/usr/lib/tmpfiles.d \
```

(The normal question of when to specify EPREFIX in paths and when not to ... --localstatedir (next line) was with, so I followed that.  If someone can please confirm.

I've fixed the other stuff I believe.

ionenwks: > ```
> 101         --with-tmpfilesdir="${EPREFIX}"/usr/lib/tmpfiles.d \
> ```
Seeing how it's a top level path (/usr), it'd mean it's not adding what was passed to --prefix= and so you need it.

And sure enough there's no issue and I'm getting:
`home/ionen/prefix/usr/lib/tmpfiles.d/gluster.conf`

However, ***if*** you want to worry about prefix, there's also this:
```
 * QA Notice: the following files are outside of the prefix:
 * /sbin
 * /sbin/mount.glusterfs
 * ERROR: sys-cluster/glusterfs-9.2::gentoo failed:
 *   Aborting due to QA concerns: there are files installed outside the prefix
```

sys-cluster/glusterfs/glusterfs-9.2.ebuild
```diff
@@ -0,0 +1,195 @@
...
+	elog "You can now treat glusterfsd2 like any other service"
+	echo
+	ewarn "You need to use a ntp client to keep the clocks synchronized across all"
+	ewarn "of your servers. Setup a NTP synchronizing service before attempting to"
+	ewarn "run GlusterFS."
+	echo
+	elog "If you are upgrading from a previous version of ${PN}, please read:"
+	elog "  http://docs.gluster.org/en/latest/Upgrade-Guide/upgrade_to_$(ver_cut '1-2')/"
+
+	tmpfiles_process /usr/lib/tmpfiles.d/gluster.conf
```
ionenwks, line 188:
```suggestion
	tmpfiles_process gluster.conf
```
No need for full path.

sys-cluster/glusterfs/glusterfs-9.2.ebuild
jkroonza: That /sbin is an interesting one.  This comes from the package itself I guess, if you don't mind let's let that one slip for the moment.  Not really looking forward to digging into that particular code, but we can report an issue upstream perhaps?

juippis (approved):
Works for me, thank you both!

PR #20810, opened 2021-05-14
dev-python/cryptography: the snapshot of the new version with rust

vaartis:
Separated from #20795. The new version of cryptography now requires rust. Putting this up here in case we need it in the future.


gyakovlev:
there's practically no rust extention in 3.4.7
https://github.com/pyca/cryptography/blob/3.4.7/src/rust/src/lib.rs

vaartis:
> there's practically no rust extention in 3.4.7
> https://github.com/pyca/cryptography/blob/3.4.7/src/rust/src/lib.rs

This isn't 3.4.7, but a snapshot of the latest version from github. 3.4.7, which we already have, has the ability to disable rust, which we do. After that release, they removed the ability to disable rust and started converting C code to rust. Initially this was made when I worked on porting to python 3.10, but it turned out backporting a patch was enough for that. If we ever chose to support the newer versions which, if the developers of the library do not decide otherwise or someone doesn't fork the library, will require rust, this snapshot ebuild could be the starting point for this new version.

gyakovlev:
> > there's practically no rust extention in 3.4.7
> > [https://github.com/pyca/cryptography/blob/3.4.7/src/rust/src/lib.rs](https://github.com/pyca/cryptography/blob/3.4.7/src/rust/src/lib.rs?rgh-link-date=2021-05-15T07%3A48%3A24Z)
> 
> This isn't 3.4.7, but a snapshot of the latest version from github. 3.4.7, which we already have, has the ability to disable rust, which we do. After that release, they removed the ability to disable rust and started converting C code to rust. Initially this was made when I worked on porting to python 3.10, but it turned out backporting a patch was enough for that. If we ever chose to support the newer versions which, if the developers of the library do not decide otherwise or someone doesn't fork the library, will require rust, this snapshot ebuild could be the starting point for this new version.

Sorry I don't see snapshot, just a revbump with patch.
did you forget to push something or still working on it? I that case I might be early to review =)

vaartis:
>  did you forget to push something or still working on it? I that case I might be early to review =) 
You are correct.. seems like i pushed the wrong commit. I'm going to push the correct one soon. I was suggested to make a draft PR for this since it's not known whether we will need this, but it may be useful in the future.

sbraz:
There's a typo in the PR title: "cryptorgaphy".

dev-python/cryptography/cryptography-3.4.7_p12052021.ebuild
```diff
@@ -0,0 +1,115 @@
...
+		dev-python/pytest-subtests[${PYTHON_USEDEP}]
+		dev-python/pytest-xdist[${PYTHON_USEDEP}]
+	)"
+
+distutils_enable_tests pytest
+
+DEPEND="
+	>=dev-libs/openssl-1.0.2o-r6:0=
+	"
+RDEPEND+=${DEPEND}
```
gyakovlev, line 89: quote it.

dev-python/cryptography/cryptography-3.4.7_p12052021.ebuild
```diff
@@ -0,0 +1,115 @@
...
+		if [[ ${CHOST##*-darwin} -ge 16 ]] ; then
+			darwinok=1
+		fi
+		sed -i -e 's/__builtin_available(macOS 10\.12, \*)/'"${darwinok}"'/' \
+			src/_cffi_src/openssl/src/osrandom_engine.c || die
+	fi
+}
+
+python_test() {
+	local -x PYTHONPATH=${PYTHONPATH}:${S}/vectors
```
gyakovlev, line 113: quote

dev-python/setuptools_rust/metadata.xml
```diff
@@ -0,0 +1,12 @@
...
+<pkgmetadata>
+  <maintainer type="project">
+    <email>python@gentoo.org</email>
+    <name>Python</name>
+  </maintainer>
+  <upstream>
+    <remote-id type="pypi">setuptools_rust</remote-id>
+    <remote-id type="github">PyO3/setuptools-rust</remote-id>
+  </upstream>
+</pkgmetadata>
```
gyakovlev, line 12: normalize xml with tabs
`XMLLINT_INDENT=$'\t' xmllint --format metadata.xml  > newname.xml`

dev-python/setuptools_rust/metadata.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+  <maintainer type="project">
+    <email>python@gentoo.org</email>
+    <name>Python</name>
+  </maintainer>
```
gyakovlev, line 7: add `rust@gentoo.org` as second maintainer.

dev-python/cryptography/cryptography-3.4.7_p12052021.ebuild
```diff
@@ -0,0 +1,115 @@
...
+	' 'python*')
+	"
+BDEPEND="
+	dev-python/setuptools_rust[${PYTHON_USEDEP}]
+	test? (
+		>=dev-python/hypothesis-1.11.4[${PYTHON_USEDEP}]
+		dev-python/iso8601[${PYTHON_USEDEP}]
+		dev-python/pretend[${PYTHON_USEDEP}]
+		dev-python/pyasn1-modules[${PYTHON_USEDEP}]
+		dev-python/pytz[${PYTHON_USEDEP}]
```
gyakovlev, line 79: sort

gyakovlev (changes requested):
looks good, left some comments.

gyakovlev:
also I think it's ok to add `<stabilize-allarches/>` to `setuptools_rust` as it's python only code.

vaartis:
I think I got it all. Also rebased on master, where the older cryptography version was removed.

dev-python/setuptools_rust/setuptools_rust-0.12.1.ebuild
```diff
@@ -0,0 +1,35 @@
...
+MY_PN="${PN/_/-}"
+MY_P="${MY_PN}-${PV}"
+
+DESCRIPTION="a plugin for setuptools to build Rust Python extensions"
+HOMEPAGE="https://github.com/PyO3/setuptools-rust"
+SRC_URI="mirror://pypi/${PN::1}/${MY_PN}/${MY_P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64 ~ppc ~ppc64 ~riscv ~x86"
```
gyakovlev, line 21: ditto with keywords here. only add `~amd64`, the bot check will become green.

vaartis: Right, of course. Done.

vaartis:
Unfortunately there's still a semantic_version bump to 3.10 in another PR that is needed for this one, so the warnings won't go away until it's merged.

gyakovlev:
ok let's wait then until rest is processed, thanks for working on it btw. setuptoos_rust was my on my to-do list, now it's not ;-)

vaartis:
Rebased on master since semantic_version got merged. This should make the QA bot happy.

gyakovlev:
found time to test.

you can add `~ppc64` keywords, tested by me, tests pass.
`2722 passed, 3252 skipped, 86631 warnings in 241.48s (0:04:01) `

`RUSTFLAGS` seem to work fine,  good.

qa warning 1 for `setuptools_rust`

```
 * 
 * QA Notice: this package seems to contain tests but they are not enabled.
 * Please either run tests (via distutils_enable_tests or declaring
 * python_test yourself), or add RESTRICT="test" along with an explanatory
 * comment if tests cannot be run.
 * 
 ```
 

qa warning 2 for `cryptography`

```
 * /usr/lib/python3.9/site-packages/cryptography/hazmat/bindings/_rust.abi3.so


 * /usr/lib/python3.9/site-packages/cryptography/hazmat/bindings/_rust.abi3.so
 ```
probably 
`QA_FLAGS_IGNORED="usr/lib.*/py.*/site-packages/cryptography/hazmat/bindings/_rust.abi3.so"`
should shut it up, it's acceptable for rust.
 

also `src_unpack` spams with 
```
>>> Unpacking winapi-0.3.9.crate to /var/tmp/portage/dev-python/cryptography-3.4.7_p12052021/work                                                                                                                                                                                     
unpack winapi-0.3.9.crate: file format not recognized. Ignoring.
....
```

you should not call `default` in `src_unpack` to avoid it, `cargo_src_unpack` alone is enough and will handle tarballs just fine.
unknown formats in `SRC_URI` passed to unpack  may become errors in future `EAPI`.

otherwise it's good =)

vaartis:
OK, so I _sort of_ got `setuptools_rust` "tests" to work: they use the examples dir with several rust things for testing. Unfortunately two of them I was not able to get to work. html_py_ever works if I run the same commands as portage does by hand, but when running with emerge it can't import the module it's looking for, and rust_with_cffi uses _a different version of pyo3_ for some reason and i think cargo.eclass can't handle that (maybe because it's not even specified in cargo.toml, but instead is marked as a transitive dependency and `cargo ebuild` doesn't even show it).

thesamesam:
(Try using distutils_install_for_testing with its various options?)

gyakovlev:
I don't think examples are real tests.
just add 
`RESTRICT=test #no tests`
is sucks ofc, but we tried.

gyakovlev:
but I did not look at examples, maybe worth keeping. I'll give it another go.

vaartis:
> but I did not look at examples, maybe worth keeping. I'll give it another go.

The examples have test in them, and are considered "tests" by upstream.

vaartis:
@gyakovlev, i was able to get html-py-ever to work, plus worked out the annoying thing of it trying to install the built examples into the system. unforunately the src_test phase is a bit complex, becase the tests are super inconsistent, but it works. still no luck with `rust_with_cffi` unfortunately, i don't think we can do anything about that. but `html-py-ever` seems to be the biggest real test example.

gyakovlev:
thanks.
I might have an idea how to fix different cargo crates.
we could define a var, which lists deps required by `rust_with_cffi`
```
TEST_CRATES="
...
"
```

and extend `SRC_URI` with

```
...
    test?  ( $(cargo_crate_uris ${TEST_CRATES}) )
```

but again, haven't looked at it in detail, it's just an assumption based on fact the example needs different crate version.

gyakovlev:
I got lost here, seems it's what you've done already.

vaartis:
> I got lost here, seems it's what you've done already.

Yes, but if I list both versions there I think only one gets downloaded by cargo.eclass..

gyakovlev:
I see where problem is.
version `0.12.1` of `setuptools_rust` requires git checkout of pyo3, hence it's not handled by eclass/cargo-ebuild.

current main branch switched versioned pyo3, so it should work in the future.

https://github.com/PyO3/setuptools-rust/commit/21fc0105dc7ba9b3bfdce1530a80190c01dff560#diff-a3f14f562689def4c1df269f3e37bafcf6036a58798d825b8a58438cb479600e

you can list multiple versions of pyo3 in `CRATES`, that will be fine btw.

also, cargo-ebuild was a bit out of maintenance. it may fail to catch all crates.
I recently taken over maintenance of it and will fix it soon.

what I use is right now is
```
CARGO_HOME=/tmp/crates cargo fetch
ls -1 /tmp/crates/registry/src/github.com-*/
```

it will give you a bit more reliable list of crates.

to really catch all deps of all examples - just repeat `fetch` step for all examples containing cargo.toml and `ls -1` shoud contain all crates that can be possibly used. 
as bonus point - you can copy downloaded crates from `/tmp/crates/registry/cache/github.com-*/` to your distdir.
and ofc clean-up /tmp/crates after that and before run too.

gyakovlev:
```
python3.9 -m pytest -vv -ra -l
/usr/bin/python3.9: No module named pytest
```

looks like setuptools_rust needs pytest test dep.

gyakovlev:
after adding pytest it passes.
```
tests/test_namespace_package.py::test_rust PASSED                                                                                   [ 50%]
tests/test_namespace_package.py::test_cffi PASSED                                                                                   [100%]
```
we should look at merging setuptools_rust tbh and leave crypto in Draft for now.

vaartis:
Yes, it does. Sorry, forgot to add it.

vaartis:
>current main branch switched versioned pyo3, so it should work in the future.
> you can list multiple versions of pyo3 in CRATES, that will be fine btw.

Ah, i see why it couldn't fetch it then. my bad.

vaartis:
I guess we still can't use the crate from git easily, so that's for the next release.

vaartis:
The html-py-ever tests actually require beautifulsoup and it hasn't been ported to 3.10 yet, it has some problems with compatibility unfortunately. Didn't notice initially because I have it installed in 3.10 locally.

gyakovlev:
merged `setuptools_rust`, and committed just a bit of fixes in https://github.com/gentoo/gentoo/commit/6f304b0a6d911510c647e880df2c5efc3c13323f
don't want to nitpick it anymore. you've done a lot already.

vaartis:
Thank you. Should I watch when the next release comes out and update this?

gyakovlev:
> Thank you. Should I watch when the next release comes out and update the package then?

as you wish, actual help is always appreciated =)

gyakovlev:
something to put into commit message:

`Bug: https://bugs.gentoo.org/769482`

thesamesam:
Thanks a lot @vaartis! Merged and then updated to 36.0.0.

PR #20815, opened 2021-05-15
dev-php/pecl-parallel: Fix SRC_URI.

jkroonza:
Something changed, not sure what, but he defaults from the eclass now
works.  So use that.

Closes: https://bugs.gentoo.org/789393


dev-php/pecl-parallel/pecl-parallel-1.1.4.ebuild
```diff
@@ -11,8 +11,6 @@ USE_PHP="php7-3 php7-4"
 
 inherit php-ext-pecl-r3
 
-SRC_URI="https://pecl.php.net/get/pecl-${PHP_EXT_NAME}-${PV/_/}.tgz"
```
juippis, line 14: Well we did this because "parallel" is too ambiguous, so I'd like to remain pecl-parallel in the name. For example, 
there are 3 different programs named "parallel" in Gentoo tree under different categories, and many packages having "\*parallel\*" in their name.

Didn't this get mirrored? Does the mirrored tarball still work? If you need to update the upstream tarball, add `-1` before `.tgz` in `SRC_URI`.

jkroonza: Oh yes!  Now I remember!

Apparently no mirroring no.  And I used the -> operator here, so no, SRC_URI was committed incorrectly.  So I had this:

```SRC_URI="${SRC_URI} -> ${P}.tgz"```

Which you then updated to:

```SRC_URI="https://pecl.php.net/get/pecl-${PHP_EXT_NAME}-${PV/_/}.tgz```

Which was OK for me since I already had the file downloaded, so I didn't pick up on this.

May I recommend simply reverting commit c5f4efa8338aed1ff81259e882f0920c62434f43 ?

juippis: Hmm sure, a bit embarrassing on my behalf... although I don't still get how the manifest hash changes with that.

Let's try then.

jkroonza: It's not the hash, I originally said whatever the eclass determines the SRC_URI to be, rename it, your well intended more readable broke that by way of providing the wrong download location.  No change required to the Manifest since the hash is fine, the reason this patch changes the Manifest is because of the different filename.

I hope that babble makes sense.

PR #20817, opened 2021-05-15
gtk2removal

a17r:


x11-misc/macopix/files/macopix-3.4.0-openssl-1.1.0.patch
```diff
@@ -0,0 +1,10 @@
+--- a/src/http-client.c	2019-12-30 22:05:51.000000000 +0100
++++ b/src/http-client.c	2021-05-15 14:41:27.703133114 +0200
```
SoapGentoo, line 2: remove metadata noise plox

games-strategy/freeciv/freeciv-2.6.4.ebuild
```diff
@@ -0,0 +1,182 @@
...
+			./tools/freeciv-manual || die
+			docinto html
+			dodoc classic*.html
+		fi
+		if use sdl ; then
+			make_desktop_entry freeciv-sdl "Freeciv (SDL)" freeciv-client
+		else
+			rm -f "${ED}"/usr/share/man/man6/freeciv-sdl*
+		fi
+		rm -f "${ED}"/usr/share/man/man6/freeciv-xaw*
```
arthurzam, line 176: missing `|| die`

arthurzam, line 176: missing `|| die`

NP-Hardass (changes requested):
NAK from me.  I'm OK with the core change to remove the gtk3 USE flag, EAPI=7 and the change regarding the .desktop files, but there is no reason to change the ebuild style throughout.  Gentoo policy, unless it has changed, is to not change another developer's style unnecessarily when working on their ebuilds. 
Leave COMMON_DEPEND, src_unpack as is, and don't monkey with PV comparisons unnecessarily, and I'd be OK with the other changes

a17r:
> Leave COMMON_DEPEND

@NP-Hardass That makes no sense. Please look at it again, and if you don't know what `BDEPEND` is, read EAPI-7 docs.

NP-Hardass:
> > Leave COMMON_DEPEND
> 
> @NP-Hardass That makes no sense. Please look at it again, and if you don't know what `BDEPEND` is, read EAPI-7 docs.

COMMON_DEPEND is being used in DEPEND and RDEPEND, not BDEPEND.  Makes absolute sense.  Only thing BDEPEND does is split out build time runtime deps... COMMON_DEPEND, as I'm using it, is for dependencies shared between DEPEND and RDEPEND, so that I don't have to deal with figuring out if a dep is misshared between DEPEND and RDEPEND when I come across a DEPEND=${RDEPEND} or vice versa.

Makes complete sense.  Please look at it again.

a17r:
So you want me to keep a completely useless `COMMON_DEPEND`. Fine, but that's completely unusual, I just wanted that on the record.

NP-Hardass:
> So you want me to keep a completely useless `COMMON_DEPEND`. Fine, but that's completely unusual, I just wanted that on the record.

I try, when possible, to use a similar style in all my ebuilds.  Just makes it easier to deal with.  So, if it means that occasionally a COMMON_DEPEND is superfluous, I'm OK with that as long as it makes the dependencies easier to think about across the board.

Make sense?

a17r:
Keeping a useless COMMON_DEPEND is on the same level as MY_P etc. - it is just adding clutter when it serves no purpose (as it is 90% of the time).

But I do as you please.

PR #20818, opened 2021-05-15

arthurzam:
Not a big list of packages, but all of those were tested and had passed.


arthurzam:
Added `css-parser`:
- needed patch from upstream (those were multiple commits, that were merged using `git diff`)
- adding `chardet` dependency removes warning during tests
- fully passes tests after patch (doesn't without it)

mgorny:
Hmm, network-sandbox seems to no longer be a problem for css-parser, so I'll unrestrict tests entirely.

PR #20824, opened 2021-05-15
net-firewall/firewalld: Updates PLOCALE for 0.9.x

pinkflames:
Both firewalld 0.9.1 and 0.9.3 ebuilds are missing `ast` and `fa` from the PLOCALES variable, which is triggering a warning (probably due to FEATURES=stricter). This commit seeks to remedy that and hopefully is not violating some policy on mandatory revbumps or rekeywording.

### DIFF
--- a/net-firewall/firewalld/firewalld-0.9.1-r1.ebuild
+++ b/net-firewall/firewalld/firewalld-0.9.1-r1.ebuild
@@ -12,7 +12,7 @@ SRC_URI="https://github.com/${PN}/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
 
 LICENSE="GPL-2+"
 SLOT="0"
-KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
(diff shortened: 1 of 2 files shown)


thesamesam:
Does it change installed files? If so, revbump, and probably git mv for it if it’s an obvious omission or typo.

Could you show the error?

pinkflames:
> Does it change installed files? If so, revbump, and probably git mv for it if it’s an obvious omission or typo.
Yeah, updating the ebuild does install two new locale files. I was mostly concerned with the warning and didn't think of users not getting the new files, if there's no revision change. Thanks for pointing that out, Sam.

> Could you show the error?
 * There are changes in locales! This ebuild should be updated to:
 * PLOCALES="ar as ast bg bn_IN ca cs da de el en_GB en_US es et eu fa fi fr gl gu hi hu ia id it ja ka kn ko lt ml mr nl or pa pl pt pt_BR ru sk sq sr sr@latin sv ta te tr uk zh_CN zh_TW"

juippis (approved):

PR #20828, opened 2021-05-16
dev-python/*: bump semantic_version and dnspython + dependencies

vaartis:
A continuation of #20795. 

@mgorny, I was able to bump django and all dependencies to 3.10 (with some patches), and now the semantic_version tests for django pass for me. Still not sure about the dnspython test fails, since they work fine for me.

### DIFF
--- a/dev-python/selenium/selenium-3.141.0-r1.ebuild
+++ b/dev-python/selenium/selenium-3.141.0-r1.ebuild
@@ -1,9 +1,9 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7..9} )
+PYTHON_COMPAT=( python3_{7..10} )
 
 inherit distutils-r1
 
--- a/dev-python/urllib3/urllib3-1.26.4.ebuild
+++ b/dev-python/urllib3/urllib3-1.26.4.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7..9} pypy3 )
+PYTHON_COMPAT=( python3_{7..10} pypy3 )
 PYTHON_REQ_USE="ssl(+)"
 
 inherit distutils-r1
@@ -40,6 +40,10 @@ BDEPEND="
 	)
 "
 
+PATCHES=(
+	"${FILESDIR}/${P}-test-ssltransport.patch"
+)
+
 python_prepare_all() {
 	# tests failing if 'localhost.' cannot be resolved
 	sed -e 's:test_dotted_fqdn:_&:' \
@@ -58,6 +62,34 @@ python_test() {
 	local deselect=(
 		# TODO?
 		test/with_dummyserver/test_socketlevel.py::TestSocketClosing::test_timeout_errors_cause_retries
+	)
+	[[ ${EPYTHON} == python3.10 ]] && deselect+=(
+		# Fail because they rely on warnings and there are new deprecation warnings in 3.10
+		test/contrib/test_pyopenssl.py
+		test/with_dummyserver/test_https.py::TestHTTPS::test_verified
+		test/with_dummyserver/test_https.py::TestHTTPS::test_verified_with_context
+		test/with_dummyserver/test_https.py::TestHTTPS::test_context_combines_with_ca_certs
+		test/with_dummyserver/test_https.py::TestHTTPS::test_ca_dir_verified
+		test/with_dummyserver/test_https.py::TestHTTPS::test_ssl_correct_system_time
+		test/with_dummyserver/test_https.py::TestHTTPS::test_ssl_wrong_system_time
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_2::test_verified
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_2::test_verified_with_context
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_2::test_context_combines_with_ca_certs
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_2::test_ca_dir_verified
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_2::test_ssl_correct_system_time
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_2::test_ssl_wrong_system_time
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_2::test_default_tls_version_deprecations
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_2::test_no_tls_version_deprecation_with_ssl_version
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_2::test_no_tls_version_deprecation_with_ssl_context
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_3::test_verified
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_3::test_verified_with_context
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_3::test_context_combines_with_ca_certs
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_3::test_ca_dir_verified
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_3::test_ssl_correct_system_time
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_3::test_ssl_wrong_system_time
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_3::test_default_tls_version_deprecations
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_3::test_no_tls_version_deprecation_with_ssl_version
+		test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_3::test_no_tls_version_deprecation_with_ssl_context
 	)
 
 	epytest ${deselect[@]/#/--deselect }
--- a/dev-python/urllib3/files/urllib3-1.26.4-test-ssltransport.patch
+++ b/dev-python/urllib3/files/urllib3-1.26.4-test-ssltransport.patch
@@ -1,0 +1,31 @@
+diff --git a/test/test_ssltransport.py b/test/test_ssltransport.py
+index 72b06b006..98682bd43 100644
+--- a/test/test_ssltransport.py
++++ b/test/test_ssltransport.py
+@@ -246,6 +246,7 @@ def proxy_handler(listener):
+                 )
+                 self._read_write_loop(client_sock, upstream_sock)
+                 upstream_sock.close()
++                client_sock.close()
+ 
+         self._start_server(proxy_handler)
+ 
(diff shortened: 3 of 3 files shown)


mgorny:
Every day a new challenge, right? ;-)

vaartis:
> Every day a new challenge, right? ;-)

So it seems, so it seems. For constantly, the problem is that the latest twisted version doesn't have an `alpha` keyword, but other versions do. For twisted, there are some dependency problems, I'll try setting all use flags for everything and checking what emerge says. Same thing for tornado.

vaartis:
I've solved the problems that existed for twisted itself. For tornado and constantly, we need an ~alpha keyword on the latest twisted version, could this somehow be requested? The old twisted versions seem to be pretty broken and their tests also probably depended on an older version of pyopenssl which were removed from the tree recently. I could just disable these tests and things would probably work out, but having the ~alpha keyword on the newer version would solve this too, and the old version could be left with python 3.9.

vaartis:
Seems like pyopenssl also requires ~riscv from cryptography, the new version of which does not have that keyword, but the old does,

mgorny:
[Rekeywording of twisted](https://bugs.gentoo.org/773451) is already requested. We just have to wait.

My suggestion would be to generally revbump and drop keywords if really necessary, or just defer adding Python 3.10 for now.

mgorny:
Also, if you managed to split the changes into smaller PRs that would be helpful. Ideally do the packages that don't cause problems first.

vaartis:
> [Rekeywording of twisted](https://bugs.gentoo.org/773451) is already requested. We just have to wait.
> 
> My suggestion would be to generally revbump and drop keywords if really necessary, or just defer adding Python 3.10 for now.

I'll subscribe to the rekeywording thread and wait a bit, then. In the meanwhile, i'll probably move twisted and everything that needs later in the commit history, so everything before it could be merged.

As for cryptography, I was able to get a riscv qemu user-emulator working and am currently testing cryptography there. [The RISC-V project page](https://wiki.gentoo.org/wiki/Project:RISC-V) says that is enough for keywording as of now.

vaartis:
Checked cryptography and pytest-subtests (since cryptography needed it for tests) in qemu-user riscv chroot, both worked fine and tests passed, so I added ~riscv for them.

vaartis:
Separated everything that needs twisted into #20835. The rest don't need it, or at least pkgcheck doesn't find any problems with that.

dev-python/pytest-asyncio/pytest-asyncio-0.15.1.ebuild
```diff
@@ -23,4 +23,11 @@ BDEPEND="
 		>=dev-python/hypothesis-3.64[${PYTHON_USEDEP}]
 	)"
 
+python_prepare_all() {
+	# Ignore deprecation warnings
+	sed -i -E 's/^filterwarnings = error/&\n    ignore::DeprecationWarning/' setup.cfg || die
```
mgorny, line 28: I'd rather not do that, unless upstream actually blesses it. Then, it would be more readable as a patch.

mgorny (changes requested):
Also, please sort commits dependency-first.

dev-python/pytest-asyncio/pytest-asyncio-0.15.1.ebuild
mgorny: Hmm, I see upstream's working hard on a major change, so I suppose this is fine after all.

mgorny:
Ok, I'm completely lost as to which PR is current and what I'm supposed to do with it all ;-). Could you split the packages that are not blocked by anything out, and make a small PR with them specifically?

dev-python/pytest-asyncio/pytest-asyncio-0.15.1.ebuild
vaartis: Should I still make it a patch instead?

vaartis:
> Ok, I'm completely lost as to which PR is current and what I'm supposed to do with it all ;-). Could you split the packages that are not blocked by anything out, and make a small PR with them specifically?

This is the PR with those not blocked, the blocked ones are in the one mentioned a few comments above. I'll sort the things dependency-first in a bit.
The one being blocked is based on this one, so it has everything this one has + blocked deps, I'll rebase when this is merged, so it'll only have the ones blocked by twisted.

dev-python/pytest-asyncio/pytest-asyncio-0.15.1.ebuild
mgorny: Yes, please.

vaartis:
Sorted dependency-first to the best of my ability and converted the sed config change to a patch.

vaartis:
Will do the same thing in a moment for jinja.

vaartis:
Alright, that's done. Going to rebase the twisted PR on this one again.

mgorny:
Please skip jinja for now. I've allowed pytest-asyncio because upstream confirmed that it's a problem with their tests and not the package itself. In case of jinja, the buggy calls occur inside the main code.

vaartis:
Moved it to the twisted PR for now, and removed from here.

dev-python/urllib3/urllib3-1.26.4.ebuild
```diff
@@ -58,6 +58,35 @@ python_test() {
 	local deselect=(
 		# TODO?
 		test/with_dummyserver/test_socketlevel.py::TestSocketClosing::test_timeout_errors_cause_retries
+		# Failed: Timeout >60.0s
```
mgorny, line 61: I'm sorry but this is a big *no* from me. There is clearly some breakage involved here, and I would really like to have upstream at least look into it first.

dev-python/selenium/selenium-3.141.0-r1.ebuild
```diff
@@ -1,9 +1,9 @@
-# Copyright 1999-2020 Gentoo Authors
```
mgorny, line 1: Had to skip it because of urllib3.

dev-python/semantic_version/semantic_version-2.8.5.ebuild
```diff
@@ -3,7 +3,7 @@
```
mgorny, line 3: Skipped it because it still fails for me.

dev-python/hpack/hpack-3.0.0.ebuild
```diff
@@ -2,7 +2,7 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=6
```
mgorny, line 4: I've backported the changes from 4.0.0 too to make this a more modern ebuild.

dev-python/hpack/hpack-3.0.0.ebuild
```diff
@@ -34,6 +34,11 @@ python_prepare_all() {
 	# Remove a test that is not part of the mainstream tests
 	# Also, it's data directory is not included in the release
 	rm test/test_hpack_integration.py || die
+
+	# depends on old enum representation
+	sed -e 's:test_get_by_index_out_of_range:_&:' \
```
mgorny, line 39: We don't do seds on package using pytest anymore, we use `--deselect` instead. Done that.

dev-python/hyper-h2/hyper-h2-3.2.0.ebuild
```diff
@@ -29,6 +29,21 @@ DEPEND="${RDEPEND}
 	)
 "
 
+PATCHES=(
+	# From https://github.com/python-hyper/h2/pull/1248
+	# Disables some failing healthchecks
+	"${FILESDIR}/${P}-failed-healthcheck.patch"
```
mgorny, line 35: I've also had to add this patch to 4.0.0.

dev-python/hyper-h2/hyper-h2-3.2.0.ebuild
```diff
@@ -29,6 +29,21 @@ DEPEND="${RDEPEND}
...
+)
+
+python_prepare_all() {
+	# These tests rely on enum representation which changed in python 3.10
+	sed -E 's/def (.+_repr)/def _\1/' \
```
mgorny, line 40: Same and same (`--deselect`, updates from newer version).

dev-python/hyper-h2/files/hyper-h2-3.2.0-failed-healthcheck.patch
```diff
@@ -0,0 +1,74 @@
...
+         """
+         If the connection flow control window is empty, but the stream flow
+@@ -916,5 +920,6 @@ def test_connection_only_empty(self, frame_factory, increment):
+         assert c.data_to_send() == expected_data
+ 
+     @given(integers(min_value=1025, max_value=DEFAULT_FLOW_WINDOW))
++    @settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
+     def test_mixing_update_forms(self, frame_factory, increment):
+         """
+         If the user mixes ackowledging data with manually incrementing windows,
```
mgorny, line 74: Missing LF at EOF.

dev-python/cryptography/cryptography-3.3.2.ebuild
```diff
@@ -3,7 +3,7 @@
```
mgorny, line 3: Now I'm confused. This ebuild shouldn't be needed by anything. In fact, I think I can remove the old version already.

mgorny (commented):
I've merged what I could, please rebase and we'll see where we can go from there.

Note that you're going to see some merge conflicts due to my changes to the ebuilds.

dev-python/cryptography/cryptography-3.3.2.ebuild
vaartis: Sorry, did not actually look into reverse-dependencies. If we don't need it anymore, that's good.

dev-python/urllib3/urllib3-1.26.4.ebuild
vaartis: Seems like they already know about it: https://github.com/urllib3/urllib3/pull/2209. So we'll just have to wait until this is fixed, I guess..

vaartis:
Moved semantic_version to the twisted PR, because (and i completely forgout about this) it depends on django for tests and that one's in the other PR. For now, just waiting on urllib3 to deal with that test failure, I've subscribed to the PR they have.

dev-python/urllib3/urllib3-1.26.4.ebuild
vaartis: Alright, I applied the patch mentioned there and the tests mentioned are now passing.

dev-python/urllib3/urllib3-1.26.4.ebuild
```diff
@@ -58,6 +62,32 @@ python_test() {
 	local deselect=(
 		# TODO?
 		test/with_dummyserver/test_socketlevel.py::TestSocketClosing::test_timeout_errors_cause_retries
+		# Fail because they rely on warnings and there are new deprecation warnings in 3.10
```
mgorny, line 65:
```suggestion
	)
	[[ ${EPYTHON} == python3.10 ]] && deselect+=(
		# Fail because they rely on warnings and there are new deprecation warnings in 3.10
```

mgorny (changes requested):
Thanks for doing this! I've planned on doing it myself but you saved me the work ;-).

dev-python/urllib3/urllib3-1.26.4.ebuild
vaartis: Amended

mgorny:
Thanks. I'm going to do a few final touches and merge it!

mgorny:
(this will take some more time as I'm building kernels right now)

PR #20830, opened 2021-05-16
media-libs/openimageio-2.2.13.0: required for current stable blender

samuelbernardo:
This PR fixes the following issue
Bug: https://bugs.gentoo.org/788982


AndrewAmmerlaan:
I see version 2.2.13.1 is already in the tree now, is that sufficient for blender? Or does it have to be exactly 2.2.13.0?

DarkDefender:
It doesn't have to be exactly 2.2.13.0. So I think we can close this.

thesamesam:
Thanks!

PR #20831, opened 2021-05-16
media-libs/openimageio-2.2.13.1: required for blender

samuelbernardo:
Update previous added version of openimageio to 2.2.13.1. This also works with current stable blender version.

Also fixes bug
Bug: https://bugs.gentoo.org/788982


thesamesam:
Use repoman/pkgdev! :)

PR #20835, opened 2021-05-16
dev-python/*: bump twisted and some packages that depend on it to python 3.10

vaartis:
Separated from #20828, and will be rebased after it's merged. Waiting on the twisted keyword for ~alpha (https://bugs.gentoo.org/773451)

### DIFF
--- a/www-servers/tornado/tornado-6.1.ebuild
+++ b/www-servers/tornado/tornado-6.1.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7..9} )
+PYTHON_COMPAT=( python3_{7..10} )
 PYTHON_REQ_USE="threads(+)"
 
 inherit distutils-r1
@@ -31,6 +31,14 @@ BDEPEND="
 distutils_enable_sphinx docs \
 	dev-python/sphinx_rtd_theme \
 	dev-python/sphinxcontrib-asyncio
+
+python_prepare_all() {
+	# Disable deprecation-warnings-as-errors because tornado has a lot of stuff deprecated in 3.10
+	sed 's/warnings.filterwarnings("error", category=DeprecationWarning, module=r"tornado\\..*")//' \
+		-i tornado/test/runtests.py || die
+
+	distutils-r1_python_prepare_all
+}
 
 python_test() {
 	local -x ASYNC_TEST_TIMEOUT=60
--- a/dev-python/constantly/constantly-15.1.0-r1.ebuild
+++ b/dev-python/constantly/constantly-15.1.0-r1.ebuild
@@ -2,7 +2,7 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
-PYTHON_COMPAT=( python3_{7..9} )
+PYTHON_COMPAT=( python3_{7..10} )
 DISTUTILS_USE_SETUPTOOLS=bdepend
 
 inherit distutils-r1
@@ -22,6 +22,4 @@ DEPEND="${RDEPEND}
 	test? ( dev-python/twisted[${PYTHON_USEDEP}] )
 "
 
-python_test() {
-	esetup.py test
-}
+distutils_enable_tests setup.py
--- a/dev-python/django/django-3.2.3-r1.ebuild
+++ b/dev-python/django/django-3.2.3-r1.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7..9} )
+PYTHON_COMPAT=( python3_{7..10} )
 PYTHON_REQ_USE='sqlite?,threads(+)'
 
 inherit bash-completion-r1 distutils-r1 optfeature verify-sig
@@ -50,6 +50,8 @@ BDEPEND="
 
 PATCHES=(
 	"${FILESDIR}"/${PN}-3.1-bashcomp.patch
+	# From https://github.com/django/django/pull/14228
+	"${FILESDIR}"/${P}-py310-repr.patch
 )
 
 distutils_enable_sphinx docs --no-autodoc
@@ -65,6 +67,13 @@ src_unpack() {
 	fi
 
 	default
+}
+
+python_prepare_all() {
+	# Fails because of warnings
+	sed -i 's/test_dumpdata_proxy_with_concrete/_&/' tests/fixtures/tests.py
+
+	distutils-r1_python_prepare_all
 }
 
 python_test() {
--- a/dev-python/django/files/django-3.2.3-py310-repr.patch
+++ b/dev-python/django/files/django-3.2.3-py310-repr.patch
@@ -1,0 +1,92 @@
+diff --git a/django/db/models/constraints.py b/django/db/models/constraints.py
+index b073df17636a..6dfc42942f79 100644
+--- a/django/db/models/constraints.py
++++ b/django/db/models/constraints.py
+@@ -4,6 +4,7 @@
+ from django.db.models.indexes import IndexExpression
+ from django.db.models.query_utils import Q
+ from django.db.models.sql.query import Query
++from django.utils.version import PY310
+ 
+ __all__ = ['CheckConstraint', 'Deferrable', 'UniqueConstraint']
+ 
+@@ -85,6 +86,11 @@ class Deferrable(Enum):
+     DEFERRED = 'deferred'
+     IMMEDIATE = 'immediate'
+ 
++    # A similar format is used in Python 3.10+.
++    if not PY310:
++        def __repr__(self):
++            return '%s.%s' % (self.__class__.__qualname__, self._name_)
++
+ 
+ class UniqueConstraint(BaseConstraint):
+     def __init__(
+@@ -218,7 +224,7 @@ def __repr__(self):
+             '' if not self.expressions else ' expressions=%s' % repr(self.expressions),
+             ' name=%s' % repr(self.name),
+             '' if self.condition is None else ' condition=%s' % self.condition,
+-            '' if self.deferrable is None else ' deferrable=%s' % self.deferrable,
++            '' if self.deferrable is None else ' deferrable=%r' % self.deferrable,
+             '' if not self.include else ' include=%s' % repr(self.include),
+             '' if not self.opclasses else ' opclasses=%s' % repr(self.opclasses),
+         )
+diff --git a/django/db/models/enums.py b/django/db/models/enums.py
+index 7082a397c237..dd9088597d4d 100644
+--- a/django/db/models/enums.py
++++ b/django/db/models/enums.py
+@@ -2,6 +2,7 @@
+ from types import DynamicClassAttribute
+ 
+ from django.utils.functional import Promise
++from django.utils.version import PY310
+ 
+ __all__ = ['Choices', 'IntegerChoices', 'TextChoices']
+ 
+@@ -74,6 +75,11 @@ def __str__(self):
+         """
+         return str(self.value)
+ 
++    # A similar format is used in Python 3.10+.
++    if not PY310:
++        def __repr__(self):
++            return '%s.%s' % (self.__class__.__qualname__, self._name_)
++
+ 
+ class IntegerChoices(int, Choices):
+     """Class for creating enumerated integer choices."""
+diff --git a/tests/model_enums/tests.py b/tests/model_enums/tests.py
+index 78f8b146be92..cda835010d7e 100644
+--- a/tests/model_enums/tests.py
++++ b/tests/model_enums/tests.py
+@@ -48,7 +48,7 @@ def test_integerchoices(self):
+         self.assertEqual(Suit.values, [1, 2, 3, 4])
+         self.assertEqual(Suit.names, ['DIAMOND', 'SPADE', 'HEART', 'CLUB'])
+ 
+-        self.assertEqual(repr(Suit.DIAMOND), '<Suit.DIAMOND: 1>')
++        self.assertEqual(repr(Suit.DIAMOND), 'Suit.DIAMOND')
+         self.assertEqual(Suit.DIAMOND.label, 'Diamond')
+         self.assertEqual(Suit.DIAMOND.value, 1)
+         self.assertEqual(Suit['DIAMOND'], Suit.DIAMOND)
+@@ -89,7 +89,7 @@ def test_textchoices(self):
+         self.assertEqual(YearInSchool.values, ['FR', 'SO', 'JR', 'SR', 'GR'])
+         self.assertEqual(YearInSchool.names, ['FRESHMAN', 'SOPHOMORE', 'JUNIOR', 'SENIOR', 'GRADUATE'])
+ 
+-        self.assertEqual(repr(YearInSchool.FRESHMAN), "<YearInSchool.FRESHMAN: 'FR'>")
++        self.assertEqual(repr(YearInSchool.FRESHMAN), 'YearInSchool.FRESHMAN')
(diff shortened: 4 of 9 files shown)


mgorny:
Did you manage to test PySocks? This bleeping thing technically requires ancient Tornado that can't work with py3.10 but you might be able to get them mostly working by collecting random patches from the Internet. I've basically started pinging people to stop using it as it's abandoned.

mgorny:
(I'm entirely happy if you managed to test it locally/in venv and not bother with making the tests work inside the ebuild)

vaartis:
> Did you manage to test PySocks? This bleeping thing technically requires ancient Tornado that can't work with py3.10 but you might be able to get them mostly working by collecting random patches from the Internet. I've basically started pinging people to stop using it as it's abandoned.

Tested it locally with requests on both 3.8 and 3.10, both seem to work fine.

vaartis:
Was able to get the tests to actually run, they work with 3.8 fine, in 3.10 tornado needs a single patch to run successfully and then the tests pass also.

mgorny:
> Was able to get the tests to actually run locally, they work with 3.8 fine, in 3.10 tornado needs a single patch to run successfully and then the tests pass also.

How much effort did you need to put into it? I'm wondering if there's a point in trying to make them ebuild-friendly after all.

vaartis:
> > Was able to get the tests to actually run locally, they work with 3.8 fine, in 3.10 tornado needs a single patch to run successfully and then the tests pass also.
> 
> How much effort did you need to put into it? I'm wondering if there's a point in trying to make them ebuild-friendly after all.

Changed the version in requirements_dev.txt for test server to 0.0.29 and after tox fetched the dependencies changed a single line in tornado from collections.MutableMapping to collections.abc.MutableMapping. That was enought for PySocks tests to pass.

mgorny:
BTW I'm surprised that Twisted works with py3.10. Back when I was doing py3.9 (and it was after py3.9 final), I had to fix a lot of issues upstream.

vaartis:
> BTW I'm surprised that Twisted works with py3.10. Back when I was doing py3.9 (and it was after py3.9 final), I had to fix a lot of issues upstream.

Surprisingly enough, yes, it seems to work fine, except those tests that depend on warning count. Perhaps thanks to your work on 3.9, but also may be because it's a relatively recent release (just two months ago).

mgorny:
Good news! Jinja has just been released in Python 3.10-compatible version, and I've pushed it to ::gentoo. Please remove jinja from the PR when rebasing ;-).

vaartis:
> Good news! Jinja has just been released in Python 3.10-compatible version, and I've pushed it to ::gentoo. Please remove jinja from the PR when rebasing ;-).

Great. Rebased and removed it, that's one problem less.

mgorny:
FYI, I've reverted asgiref because the warnings from the code break revdeps. Normally I'd say it's revdeps' fault for relying on warning-free dependencies but fighting 'fatal warnings' all over the place sounds like fighting windmills, so it's probably better to fix the issue in asgiref.

mgorny:
That said, I've forced keywords on twisted, so you may want to try rebasing and seeing where we can go from there.

vaartis:
Will look into it later today, probably going to also look into asigref stuff and patch whatever problems they have in their code and suggesting the patch upstream, depending on how much work it is.

vaartis:
https://github.com/django/asgiref/pull/262

Gathered a patch for asgiref to fix all their deprecation warnings, let's see what upstream says about them.

I saw in the bugzilla issue for alpha keywording that you have chose to instead mask 3.10 on alpha. Should I revbump all the ebuilds affected removing the keyword?

mgorny:
Don't remove keywords. The mask makes them irrelevant.

vaartis:
Now that upstream has approved the asgiref patches, I've re-added the ebuild here.

mgorny:
Great, thanks! Lemme check if/how much I can merge now.

vaartis:
There's still some keyword problem with semantic_version (nonsolvable deps), i think it's because of pypy3, I'll have to test that one and add it to the packages required.

mgorny:
I've merged the subset of packages that didn't depend on twisted. Could you rebase now?

dnspython-2 still fails for me.

vaartis:
Rebased. Can revbump + drop pypy3 from semantic_version? I don't know how it got there, but it needs a lot of depdendencies that haven't had pypy3 added yet, like all the packages from the last PR.

mgorny:
Sure, if no revdeps are broken. You can try removing it from all existing versions and using `check-revdep` to check if anything got broken (from mgorny-dev-tools).

vaartis:
> Sure, if no revdeps are broken. You can try removing it from all existing versions and using `check-revdep` to check if anything got broken (from mgorny-dev-tools).

According to the tool, there's dev-python/flask-login. If I remove it from both, then there are no more problems reported by it, neither with dev-python/semantic_version, nor with dev-python/flask-login.

vaartis:
Dropped pypy3 from both for now. Will look into the dnspython problem again tomorrow, probably. @mgorny, could you perhaps describe how you're testing it and it what environtment (maybe emerge --info?)

dev-python/semantic_version/semantic_version-2.8.5-r1.ebuild
```diff
@@ -3,7 +3,7 @@
```
mgorny, line 3: Good news! I've figured out that my failures were due to pytest-django. Will commit a fix.

www-servers/tornado/tornado-6.1.ebuild
```diff
@@ -32,6 +32,14 @@ distutils_enable_sphinx docs \
 	dev-python/sphinx_rtd_theme \
 	dev-python/sphinxcontrib-asyncio
 
+python_prepare_all() {
+	# Disable deprecation-warnings-as-errors because tornado has a lot of stuff deprecated in 3.10
+	sed 's/warnings.filterwarnings("error", category=DeprecationWarning, module=r"tornado\\..*")//' \
```
mgorny, line 37: This isn't going to fly. The warnings come from tornado itself, not the tests. You're shifting the problem to another package that will detect the warnings and fail.

dev-python/twisted/twisted-21.2.0.ebuild
```diff
@@ -4,7 +4,7 @@
 EAPI=7
 
 DISTUTILS_USE_SETUPTOOLS=rdepend
-PYTHON_COMPAT=( python3_{7..9} )
+PYTHON_COMPAT=( python3_{7..10} )
```
mgorny, line 7: Does twisted pass for you?

I get:
```
  NewFetchTests
    testFetchAll ...                                                    [ERROR]
    testFetchAllUID ...                                                 [ERROR]
                                                [ERROR]
    testFetchBody ...                                                   [ERROR]
                                                  [ERROR]
    testFetchBodyParts ...                                              [ERROR]
                                             [ERROR]
    testFetchBodyStructureUID ...                                       [ERROR]
                                      [ERROR]
    testFetchBodyUID ...                                                [ERROR]
                                               [ERROR]
    testFetchEnvelope ... 
```
and at this point it hangs.

dev-python/urllib3/urllib3-1.26.4.ebuild
```diff
@@ -64,6 +64,35 @@ python_test() {
 	local deselect=(
 		# TODO?
 		test/with_dummyserver/test_socketlevel.py::TestSocketClosing::test_timeout_errors_cause_retries
+		# Failed: Timeout >60.0s
```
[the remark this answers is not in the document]
vaartis: Correct.

dev-python/semantic_version/semantic_version-2.8.5-r1.ebuild
vaartis: Great

www-servers/tornado/tornado-6.1.ebuild
vaartis: I'll look into fixing it in the package itself and submitting it upstream

dev-python/twisted/twisted-21.2.0.ebuild
vaartis: Strange. Yes, these do pass for me.

www-servers/tornado/tornado-6.1.ebuild
vaartis: Seems like this one is too complicated for me, unfortunately, there are _a lot_ of deprecated things used.

vaartis: Seems like this one's too hard for me, there's too much deprecated in there that I don't understand. I'll try at least opening an issue upstream about this.

dev-python/twisted/twisted-21.2.0.ebuild
mgorny: Yes.

www-servers/tornado/tornado-6.1.ebuild
mgorny: Could you make a PR with at least these that you know how to fix? I think I've seen TLS-related stuff somewhere (urllib3? requests?), so you may want to try looking for a solution there.

dev-python/semantic_version/semantic_version-2.8.5-r1.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7..9} )
+PYTHON_COMPAT=( python3_{7..10} )
```
mgorny, line 6: You've lost the django dep here. But then, this makes it possible to merge it as-is for the time being, and add django when it's 3.10-ready. I'll file a bug not to forget about it.

mgorny:
I'm sorry, I was wrong. I've just retested and the twisted errors/hang is happening on 3.7 too. I'm going to investigate.

mgorny:
```
===============================================================================                                                                   [ERROR]                                                                                                                                           
  File "/tmp/portage/dev-python/twisted-21.2.0/work/twisted-twisted-21.2.0-python3_7/test/lib/twisted/internet/test/test_gireactor.py", line 100, 
in test_gtkApplicationActivate                                                                                                                    
    reactor = gtk3reactor.Gtk3Reactor()                                                                                                           
  File "/tmp/portage/dev-python/twisted-21.2.0/work/twisted-twisted-21.2.0-python3_7/test/lib/twisted/internet/gtk3reactor.py", line 35, in __init
__                                                                                                                                                
    gireactor.GIReactor.__init__(self, useGtk=True)                                                                                               
  File "/tmp/portage/dev-python/twisted-21.2.0/work/twisted-twisted-21.2.0-python3_7/test/lib/twisted/internet/gireactor.py", line 73, in __init__
    _glibbase.GlibReactorBase.__init__(self, GLib, _gtk, useGtk=useGtk)                                                                             File "/tmp/portage/dev-python/twisted-21.2.0/work/twisted-twisted-21.2.0-python3_7/test/lib/twisted/internet/_glibbase.py", line 115, in __init_
_                                                                                                                                                     self._pending = self._gtk.events_pending                                                                                                        File "/usr/lib/python3.7/site-packages/gi/overrides/__init__.py", line 32, in __getattr__                                                       
    return getattr(self._introspection_module, name)                                                                                              
  File "/usr/lib/python3.7/site-packages/gi/module.py", line 124, in __getattr__                         
    self.__name__, name))                                                                                                                         
builtins.AttributeError: 'gi.repository.Gtk' object has no attribute 'events_pending'                                

twisted.internet.test.test_gireactor.GApplicationRegistrationTests.test_gtkApplicationActivate
```

What GTK+ version do you have?

mgorny:
Ok, I'm 90% convinced that it's GTK+4 getting in the way. I've unmerged it and I'm trying again. If all tests pass this time, I'm going to merge it.

mgorny:
No such luck:

```
  RunnersTests                                                                                                                           [35/1864]
    test_basicTrialIntegration ...                                         [OK]
    test_expectedResults ... /tmp/portage/dev-python/twisted-21.2.0/temp/environment: line 2913:   127 Segmentation fault      (core dumped) "${EPYTHON}" -m twisted.trial twisted
```

www-servers/tornado/tornado-6.1.ebuild
vaartis: When i try to fix things unfortunately it breaks even more :/ 
Unfortunately this one might be fit more for a real python developer

vaartis:
> Ok, I'm 90% convinced that it's GTK+4 getting in the way. I've unmerged it and I'm trying again. If all tests pass this time, I'm going to merge it.

That's curious. I don't have GTK installed at all.

dev-python/semantic_version/semantic_version-2.8.5-r1.ebuild
vaartis: Could you link the bug here?

mgorny: Sure. https://bugs.gentoo.org/791367

mgorny:
I've filed https://twistedmatrix.com/trac/ticket/10200 for GTK+4, and I'll submit a patch to force GTK+3 shortly. If I don't forget, I'm going to test git twisted against py3.10. Or you can try (uninstall gtk:4 first):

```
git clone https://github.com/twisted/twisted
cd twisted
python3.10 setup.py build
PYTHONPATH=build/lib python3.10 -m twisted.trial twisted
```

vaartis:
> I've filed https://twistedmatrix.com/trac/ticket/10200 for GTK+4, and I'll submit a patch to force GTK+3 shortly. If I don't forget, I'm going to test git twisted against py3.10. Or you can try (uninstall gtk:4 first):

Thank you. I have tested the git twisted version and the test suite seems to pass fine, without GTK 4 that is. That said, I only just found out that there are two GTK packages in different categories. Turns out I did have GTK3 installed, so the tests do indeed work fine with it.

vaartis:
I've applied your patch to force gtk 3 to the newest version of twisted, and also tried to apply it to an older one, however it seems that older versions' tests no longer pass at all, because dev-python/pyopenssl is too new and we only have 20.0.1 in tree.

mgorny:
So you can't reproduce the segfault?

vaartis:
> So you can't reproduce the segfault?

Don't think so. It doesn't segfault for me.

www-servers/tornado/tornado-6.1.ebuild
vaartis: Got a response on the issue, they said it's probably going to be a major pain for them because of the get_current_loop deprecation.

mgorny:
So I've given it more thought and I've decided to merge it and stop blocking other packages on tornado. Worst case, other packages will hit the warnings and we'll either decide to wait or to ignore them there as well, and so on.

That is, 'it' without dnspython since that still fails for me and doesn't seem to be needed for the rest.

vaartis:
> So I've given it more thought and I've decided to merge it and stop blocking other packages on tornado. Worst case, other packages will hit the warnings and we'll either decide to wait or to ignore them there as well, and so on.
> 
> That is, 'it' without dnspython since that still fails for me and doesn't seem to be needed for the rest.

I see. Would be good to figure out what the problem with dnspython is..

mgorny:
I think it's related to output capture. I can get the test to pass by passing `-s` to pytest.

Could you try different versions of pytest, to see if this is some kind of regression?

vaartis:
> I think it's related to output capture. I can get the test to pass by passing `-s` to pytest.
> 
> Could you try different versions of pytest, to see if this is some kind of regression?

The only version of pytest that works with 3.10 is the latest one, others are broken on 3.10. It works on 8, 9 and 10 for me. The older versions of pytest work with 8 and 9 for me too.

PR #20854, opened 2021-05-17
net-voip/captagent: New Package.  Version 6.3.1.

jkroonza:
autoconf patch from upstream (in response to issue report).
gcc10 patches pending on PR.


jkroonza:
Note:  the code base itself requires some work.  For example, --disable-redis has been fixed to not bother linking against redis, but the database_redis.s file still gets built and installed, but will mostly become no-op in nature.  I'm not going to fix this here but will attempt to push PRs upstream towards the project.

net-voip/captagent/captagent-6.3.1.ebuild
```diff
@@ -0,0 +1,48 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools
+
+DESCRIPTION="CaptAgent:  100% Open-Source Packet Capture Agent for HEP"
```
juippis, line 8:
```suggestion
DESCRIPTION="CaptAgent: 100% Open-Source Packet Capture Agent for HEP"
```

net-voip/captagent/captagent-6.3.1.ebuild
```diff
@@ -0,0 +1,48 @@
...
+HOMEPAGE="https://sipcapture.org/"
```
juippis, line 9:
```suggestion
HOMEPAGE="https://sipcapture.org/ https://github.com/sipcapture/captagent"
```

net-voip/captagent/captagent-6.3.1.ebuild
```diff
@@ -0,0 +1,48 @@
...
+SRC_URI="https://github.com/sipcapture/captagent/archive/refs/tags/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3"
```
juippis, line 12: Read the source files under upstreams src/ dir, GPL-3 doesn't seem exactly correct ;)

net-voip/captagent/captagent-6.3.1.ebuild
```diff
@@ -0,0 +1,48 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="ipv6 mysql pcre redis ssl uv"
```
juippis, line 15: I can't see `uv` having a function in this ebuild.

net-voip/captagent/captagent-6.3.1.ebuild
```diff
@@ -0,0 +1,48 @@
...
+PATCHES=(
+	# https://github.com/sipcapture/captagent/pull/239 (should be accepted).
+	"${FILESDIR}/${P}-gcc10.patch"
+	# Already upstreamed for next version.
+	"${FILESDIR}/${P}-configure.patch"
+)
+
+DEPEND="dev-libs/json-c
+	net-libs/libpcap
+
```
juippis, line 26: Could we drop this empty line from here.

net-voip/captagent/captagent-6.3.1.ebuild
jkroonza: Updated locally so long.

net-voip/captagent/captagent-6.3.1.ebuild
jkroonza: Ha!  Old typing rule habits die hard.  Fixed locally so long.

net-voip/captagent/captagent-6.3.1.ebuild
jkroonza: In DEPEND ... but that should be associated with a ./configure option which there is not ... rechecking ./configure, it's a hard dep, moving that accordingly and dropping the IUSE.

net-voip/captagent/captagent-6.3.1.ebuild
jkroonza: Done.  I just prefer having a blank line between the unconditional and conditional depends, but that's personal preference, if the policy is to not have it, so be it.

net-voip/captagent/captagent-6.3.1.ebuild
jkroonza: Updated to GPL-3+

The Bison generated file doesn't specify a version, which means any version applies (as I understand).

There is also one file supplied by Aladdin, but it basically says for any purpose 'as-is' so it's not the standard a Aladdin license and I'm having trouble finding an associated one in ::gentoo/licenses/

jkroonza: Updated to GPL-3+ which is more in line with most of the source files.  Two notable exceptions:

capplan.tab.c (generated from .y) which doesn't mention a version.
md5.c which basically says public domain as I interpret it.

I'm happy that we're allowed to redistribute it as part of a GPL-3+ project.  Please confirm.

juippis (approved):

PR #20855, opened 2021-05-17
*/*: Remove last-rited pkgs

jsmolic:


fordfrog:
@jsmolic and @thesamesam, thank you both for the cleanup :-)

PR #20856, opened 2021-05-17
sci-libs/opencacade: patch cmake config so that find_package works properly

ezzieyguywuf:
The upstream cmake configuration include a cmake macro that fails - the failure
seems pretty specify to gentoo and the fact that we have an intermediate
installation directory.

Rather than fixing the macro, I just removed the mechanism - the original intent
is to allow a side-by-side installation of a release and debug version of
opencascade. I *did* try to fix the macro, but honestly its a bit of a mess and
already implements some pretty obscure cmake hackery, so I thought it best to
just leave it alone.

Another option is to fix the files in `src_install` if we want to go that route
, but then we'd need to add logic to append a suffix if it's a debug build.

Signed-off-by: Wolfgang E. Sanyer

### DIFF
--- a/sci-libs/opencascade/opencascade-7.5.1-r1.ebuild
+++ b/sci-libs/opencascade/opencascade-7.5.1-r1.ebuild
@@ -80,6 +80,13 @@ src_prepare() {
 
 	sed -e 's/\/lib\$/\/'$(get_libdir)'\$/' \
 		-i adm/templates/OpenCASCADEConfig.cmake.in || die
+
+	# There is an OCCT_UPDATE_TARGET_FILE cmake macro that fails due to some
+	# assumptions it makes about installation paths. Rather than fixing it, just
+	# get rid of the mechanism altogether - its purpose is to allow a
+	# side-by-side installation of release and debug libraries.
+	sed -e 's|\\${OCCT_INSTALL_BIN_LETTER}||' \
+		-i "adm/cmake/occt_toolkit.cmake"
 }
 
 src_configure() {


waebbl:
Could someone please fix the assignment for the PR? Thanks!

ezzieyguywuf:
> Could someone please fix the assignment for the PR? Thanks!

What should the proper assignment be? I asked the bot to reassign but I'm sure it'll just do the same thing 

@thesamesam ping

sci-libs/opencascade/opencascade-7.5.1-r1.ebuild
```diff
@@ -80,6 +80,12 @@ src_prepare() {
 
 	sed -e 's/\/lib\$/\/'$(get_libdir)'\$/' \
 		-i adm/templates/OpenCASCADEConfig.cmake.in || die
+
+	# There is an OCCT_UPDATE_TARGET_FILE cmake macro that fails due to some
+	# assumptions it makes about installation paths. Rather than fixing it, just
+	# get rid of the mechanism altogether - its purpose is to allow a
+	# side-by-side installation of release and debug libraries.
+	sed -e 's|\${OCCT_INSTALL_BIN_LETTER}||' -i "${BUILD_DIR}/"
```
juippis, line 88: Hmm...
```
sed: couldn't edit /var/tmp/portage/sci-libs/opencascade-7.5.1-r1/work/opencascade-7.5.1_build/: not a regular file
```

ezzieyguywuf: yikes, I get the same. wonder why that's happening now but not earlier when I submitted the PR....

either way, I'll fix it.

sci-libs/opencascade/opencascade-7.5.1-r1.ebuild
```diff
@@ -80,6 +80,13 @@ src_prepare() {
 
 	sed -e 's/\/lib\$/\/'$(get_libdir)'\$/' \
 		-i adm/templates/OpenCASCADEConfig.cmake.in || die
+
+	# There is an OCCT_UPDATE_TARGET_FILE cmake macro that fails due to some
+	# assumptions it makes about installation paths. Rather than fixing it, just
+	# get rid of the mechanism altogether - its purpose is to allow a
+	# side-by-side installation of release and debug libraries.
+	sed -e 's|\\${OCCT_INSTALL_BIN_LETTER}||' \
+		-i "adm/cmake/occt_toolkit.cmake"
```
juippis, line 89: Gonna add the missing `|| die`, but works now, thanks!

juippis (approved):

sci-libs/opencascade/opencascade-7.5.1-r1.ebuild
waebbl, line 89: Don't we need to cover `adm/cmake/occt_macros.cmake` as well with this statement? A `find` leads me to these two files with `OCCT_INSTALL_BIN_LETTER` in it:
```
waebbl $ find . -name *.cmake -exec grep -H OCCT_INSTALL_BIN_LETTER {} \;
./adm/cmake/occt_toolkit.cmake:           DESTINATION "${INSTALL_DIR_BIN}\${OCCT_INSTALL_BIN_LETTER}")
./adm/cmake/occt_toolkit.cmake:    install (FILES  ${CMAKE_BINARY_DIR}/${OS_WITH_BIT}/${COMPILER}/bin\${OCCT_INSTALL_BIN_LETTER}/${PROJECT_NAME}.pdb
./adm/cmake/occt_toolkit.cmake:             DESTINATION "${INSTALL_DIR_BIN}\${OCCT_INSTALL_BIN_LETTER}")
./adm/cmake/occt_toolkit.cmake:           RUNTIME DESTINATION "${INSTALL_DIR_BIN}\${OCCT_INSTALL_BIN_LETTER}"
./adm/cmake/occt_toolkit.cmake:           ARCHIVE DESTINATION "${INSTALL_DIR_LIB}\${OCCT_INSTALL_BIN_LETTER}"
./adm/cmake/occt_toolkit.cmake:           LIBRARY DESTINATION "${INSTALL_DIR_LIB}\${OCCT_INSTALL_BIN_LETTER}")
./adm/cmake/occt_toolkit.cmake:      set (LINK_NAME    "${INSTALL_DIR}/${INSTALL_DIR_LIB}\${OCCT_INSTALL_BIN_LETTER}/lib${PROJECT_NAME}${CMAKE_SHARED_LIBRARY_SUFFIX_DEFAULT}")
./adm/cmake/occt_toolkit.cmake:      set (LIBRARY_NAME "${INSTALL_DIR}/${INSTALL_DIR_LIB}\${OCCT_INSTALL_BIN_LETTER}/lib${PROJECT_NAME}${CMAKE_SHARED_LIBRARY_SUFFIX}")
./adm/cmake/occt_macros.cmake:# Adds OCCT_INSTALL_BIN_LETTER variable ("" for Release, "d" for Debug and 
./adm/cmake/occt_macros.cmake:      string (REGEX REPLACE \"[\\\\]?[\\\$]{OCCT_INSTALL_BIN_LETTER}\" \"\${OCCT_INSTALL_BIN_LETTER}\" line \"\${line}\")
./adm/cmake/occt_macros.cmake:    set (OCCT_INSTALL_BIN_LETTER \"\")
./adm/cmake/occt_macros.cmake:    set (OCCT_INSTALL_BIN_LETTER \"i\")
./adm/cmake/occt_macros.cmake:    set (OCCT_INSTALL_BIN_LETTER \"d\")
```

waebbl:
@juippis looks like my review and your push happened at the same time O.o

juippis:
@waebbl I thought you acked this? Did you have something important?

waebbl:
I started a review, but unfortunately didn't submit it...

sci-libs/opencascade/opencascade-7.5.1-r1.ebuild
waebbl: Did a check of the code. Looks like it's ok to only strip it from `occt_toolkit.cmake`. The `occt_macros.cmake` only contains the macro definition and another macro `OCCT_INSERT_CODE_FOR_TARGET` which uses the `OCCT_INSTALL_BIN_LETTER` macro, which is only used with `if(MSVC)` guards and therefore not relevant for Gentoo users.

juippis: Thanks, it'd be easy to just add a 2nd line in `sed -i` :)

waebbl:
> @waebbl I thought you acked this? Did you have something important?

I acked the earlier version, which didn't work, and the process to remove this code, but couldn't review the latest version earlier.
Please ignore the review, it's ok without changing the `occt_macros.cmake` file.

waebbl:
Thanks @ezzieyguywuf for the patch

PR #20861, opened 2021-05-18
www-client/chromium: add ppc64le patchset from void linux

gyakovlev:
Hi =)

I've been maintaining ppc64le ebuild in https://github.com/gyakovlev/chromium-ppc64le for quite some time.

And it kinda works.
I no longer maintain patchset itself, but Void people do.

What do you think about landing it in ::gentoo instead?
we already have working ppc64le qtwebengine and it also been fine.

Updating patchset will be easy, just update commit sha in `PPC64LE_COMMIT` and boom, done.
patchset rarely breaks within 1 major version.
even if it breaks and stops applying - it does not affect amd64/arm64 users.
I will update `PPC64LE_COMMIT` if required as soon as it stops applying on ppc64le.

Just to make it clear, I do not expect you to update patchset, that's on me.

not touching beta/testing ebuilds as there's no patchset.
no stable keywords required.

I tried to make it as hassle-free as possible for you.


www-client/chromium/chromium-90.0.4430.212.ebuild
```diff
@@ -15,13 +15,20 @@ DESCRIPTION="Open-source version of Google Chrome web browser"
...
 PATCHSET="7"
 PATCHSET_NAME="chromium-$(ver_cut 1)-patchset-${PATCHSET}"
+PPC64LE_COMMIT="3f575325dcc3bdfc419824518bac6d4c38241859"
+PPC64LE_URI="https://raw.githubusercontent.com/void-linux/void-packages/${PPC64LE_COMMIT}/srcpkgs/chromium/patches"
 SRC_URI="https://commondatastorage.googleapis.com/chromium-browser-official/${P}.tar.xz
 	https://files.pythonhosted.org/packages/ed/7b/bbf89ca71e722b7f9464ebffe4b5ee20a9e5c9a555a56e2d3914bb9119a6/setuptools-44.1.0.zip
-	https://github.com/stha09/chromium-patches/releases/download/${PATCHSET_NAME}/${PATCHSET_NAME}.tar.xz"
+	https://github.com/stha09/chromium-patches/releases/download/${PATCHSET_NAME}/${PATCHSET_NAME}.tar.xz
+	ppc64? (
+		${PPC64LE_URI}/xxx-ppc64le-libvpx.patch -> ${PN}-ppc64le-${PPC64LE_COMMIT}-libvpx.patch
```
floppym, line 24: The default `src_unpack` function will emit a warning about an unrecognized archive. I have heard rumblings that this may become an error in some future EAPI.

We should probably define our own `src_unpack` that unpacks the tarballs and skips the patch files.

stha09: Why not bundle the patches in a tarball and host it on d.g.o? If the changes to patches are minimal we could also try to adjust them for dev and beta releases (beta changes are minimal usually).

gyakovlev: I can do tarball, just did it the lazy way =)

patches are not minimal and usually need basic rebases between major versions, like adjusting line numbers, handling file moves and adjusting context lines.

but if it applies it usually works.

gyakovlev: pushed tarred patches

gyakovlev:
image

stha09 (approved):
LGTM, feel free to push yourself :)

PR #20862, opened 2021-05-18
dev-java/trident: min java 1.8, swt:4.10

vaukai:

### DIFF
--- a/dev-java/trident/trident-1.3-r1.ebuild
+++ b/dev-java/trident/trident-1.3-r1.ebuild
@@ -1,0 +1,50 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+JAVA_PKG_IUSE="source"
+
+inherit java-pkg-2 java-ant-2
+
+DESCRIPTION="Java animation library"
+HOMEPAGE="https://kenai.com/projects/trident/pages/Home"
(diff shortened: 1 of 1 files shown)


dev-java/trident/trident-1.3-r1.ebuild
```diff
@@ -0,0 +1,50 @@
...
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE=""
+
+RESTRICT="test"
+
+CDEPEND="dev-java/swt:4.10"
+
+DEPEND=">=virtual/jdk-1.8:*
+	app-arch/unzip
```
fordfrog, line 24: proly should go to `BDEPEND` instead.

vaukai: of course, changed

fordfrog:
@vaukai thanks! it did not compile for me though so i had to fix it in f1f7e7e2b8ada6ae59377d55d7416df77c80168f. re-keyword req should be filed because of dropped `x86`. after rekeyworded it can be stabilized if no issues till then.

vaukai:
> @vaukai thanks! it did not compile for me though so i had to fix it in [f1f7e7e](https://github.com/gentoo/gentoo/commit/f1f7e7e2b8ada6ae59377d55d7416df77c80168f). re-keyword req should be filed because of dropped `x86`. after rekeyworded it can be stabilized if no issues till then.

Sorry to hear you had to rework. No idea why it compiled here :-( 
kw request added to https://bugs.gentoo.org/790761

PR #20863, opened 2021-05-18
x11-terms/rxvt-unicode: remove query for graphics mode.

gryf:
There is a security flaw in using ANSI escape sequence for querying
graphics mode in rxvt-unicode-9.22 which can lead to remote code
execution[1]. Although Marc has turning it off almost two years ago[2],
there was no urxvt release since then. This patch is fixing it on for
urxvt version 9.22.

Versions 9.26 and 9999 are unaffected.

[1] https://seclists.org/oss-sec/2021/q2/145
[2] https://github.com/exg/rxvt-unicode/commit/be631862

Closes: https://bugs.gentoo.org/790782


thesamesam:
Please file a bug on Bugzilla in the Gentoo security component too.

gryf:
ack.

mkszuba:
Thank you for the report! That said, please keep the following in mind for future reference: revision bumps must revert all keywords to ~arch, and security tickets should only be closed by members of the Gentoo Security team. Moreover, seeing as rxvt-unicode already has a way of marking escape sequences as insecure I would rather backport the upstream fix marking ESC G Q as such than disable it altogether with a third-party patch.

PR #20866, opened 2021-05-18
Remove CMake workaround from boost, and fix prefix build of spidermonkey

amadio:
@gentoo/mozilla and @SoapGentoo Please ACK. Thanks.

### DIFF
--- a/dev-lang/spidermonkey/spidermonkey-78.10.1.ebuild
+++ b/dev-lang/spidermonkey/spidermonkey-78.10.1.ebuild
@@ -13,7 +13,7 @@ PYTHON_COMPAT=( python3_{7..9} )
 
 WANT_AUTOCONF="2.1"
 
-inherit autotools check-reqs flag-o-matic llvm multiprocessing python-any-r1 toolchain-funcs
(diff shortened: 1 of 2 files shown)


dev-lang/spidermonkey/spidermonkey-78.10.1.ebuild
```diff
@@ -230,6 +230,9 @@ src_prepare() {
 		python/mozbuild/mozbuild/configure/check_debug_ranges.py \
 		|| die "sed failed to set toolchain prefix"
 
+	# use prefix shell in wrapper linker scripts, bug #789660
+	hprefixify "${WORKDIR}"/firefox-${MY_PV}/build/cargo-{,host-}linker
```
Whissi, line 234: Could you please use path relative to `$S`?

Whissi (changes requested):

dev-lang/spidermonkey/spidermonkey-78.10.1.ebuild
amadio: Sure, that would be like `${S}/../../build/...`, is that ok?

PR #20870, opened 2021-05-18

arthurzam:
- Some more packages that passed bump to 3.10
- removed redundant versions (based on CI output)
- Some of them could get the `doc` support, so I added them.

A little question about stable editing:
I saw in `gentoo-dev` we can bump the python version to 3.10 as it isn't stable at all now, so it is fine to not bump revision.
But what about adding `IUSE=doc` support to stable ebuild?
If the answer is "not okay without bump", please skip the "enable doc" commits.


mgorny:
Thanks!

PR #20871, opened 2021-05-18
dev-python/QtPy: allow use of pyside2, but better this time

AndrewAmmerlaan:
DO NOT MERGE for now

Implementation of proposed solution 1 in linked bug

Closes: https://bugs.gentoo.org/790764

### DIFF
--- a/dev-python/QtPy/QtPy-1.9.0-r4.ebuild
+++ b/dev-python/QtPy/QtPy-1.9.0-r4.ebuild
@@ -1,0 +1,136 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7..8} )
+
+inherit distutils-r1 virtualx
+
+DESCRIPTION="Abstraction layer on top of PyQt5 and PySide2 and additional custom QWidgets"
+HOMEPAGE="https://github.com/spyder-ide/qtpy"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64 ~x86"
+
+_IUSE_QT_MODULES="
+	declarative designer gui help location multimedia network opengl
+	positioning printsupport sensors serialport sql svg testlib webchannel
+	webengine websockets widgets x11extras xml xmlpatterns
+"
+IUSE="+pyqt5 pyside2 ${_IUSE_QT_MODULES}"
+
+REQUIRED_USE="|| ( pyqt5 pyside2 )"
+
+# These flags are currently *not* common to both the PySide2 and PyQt5 ebuild
+# Disable them for now, please check periodically if this is still up to date.
+# 	bluetooth? ( pyqt5 )
+# 	dbus? ( pyqt5 )
+# 	networkauth? ( pyqt5 )
+# 	webkit? ( pyqt5 )
+#
+# 	3d? ( pyside2 )
+# 	charts? ( pyside2 )
+# 	concurrent? ( pyside2 )
+# 	datavis? ( pyside2 )
+# 	scxml? ( pyside2 )
+# 	script? ( pyside2 )
+# 	scripttools? ( pyside2 )
+# 	speech? ( pyside2 )
+
+# WARNING: the obvious solution of using || for PyQt5/pyside2 is not going
+# to work.  The package only checks whether PyQt5/pyside2 is installed, it does
+# not verify whether they have the necessary modules (i.e. satisfy the USE dep).
+#
+# Webengine is a special case, because PyQt5 provides this in a separate package
+# while PySide2 ships it in the same package.
+#
+# declarative/qml/quick is a special case, because PyQt5 bundles the bindings
+# for qml and quick in one flag: declarative PySide2 does not.
+#
+# The PyQt5 ebuild currently enables xml support unconditionally, the flag is
+# added anyway with a (+) to make it future proof if the ebuild were to change
+# this behaviour in the future.
+#
+# The PySide2 ebuild currently enables opengl and serialport support
+# unconditionally, the flag is added anyway with a (+) to make it future proof
+# if the ebuild were to change this behaviour in the future.
+RDEPEND="
+	pyqt5? (
+		dev-python/PyQt5[${PYTHON_USEDEP}]
+		dev-python/PyQt5[declarative?,designer?,gui?,help?,location?]
+		dev-python/PyQt5[multimedia?,network?,opengl?,positioning?]
+		dev-python/PyQt5[printsupport?,sensors?,serialport?,sql?,svg?]
+		dev-python/PyQt5[testlib?,webchannel?,websockets?,widgets?]
+		dev-python/PyQt5[x11extras?,xml(+)?,xmlpatterns?]
+		webengine? ( dev-python/PyQtWebEngine[${PYTHON_USEDEP}] )
+	)
+	pyside2? (
+		dev-python/pyside2[${PYTHON_USEDEP}]
+		dev-python/pyside2[designer?,gui?,help?,location?,multimedia?]
+		dev-python/pyside2[network?,opengl(+)?,positioning?,printsupport?]
+		dev-python/pyside2[sensors?,serialport(+)?,sql?,svg?,testlib?]
+		dev-python/pyside2[webchannel?,webengine?,websockets?,widgets?]
+		dev-python/pyside2[x11extras?,xml?,xmlpatterns?]
+		declarative? ( dev-python/pyside2[qml,quick] )
+	)
+"
+
+# The QtPy testsuite skips tests for bindings that are not installed, so here we
+# ensure that everything is available and all tests are run. Note that not
+# all flags are available in PyQt5/PySide2, so some tests are still skipped.
+BDEPEND="
+	test? (
+		dev-python/mock[${PYTHON_USEDEP}]
+		dev-python/PyQt5[${PYTHON_USEDEP}]
+		dev-python/PyQt5[bluetooth,dbus,declarative,designer,gui,help,location]
+		dev-python/PyQt5[multimedia,network,networkauth,opengl,positioning]
+		dev-python/PyQt5[printsupport,sensors,serialport,sql,svg,testlib]
+		dev-python/PyQt5[webchannel,webkit,websockets,widgets,x11extras,xml(+)]
+		dev-python/PyQt5[xmlpatterns]
+		dev-python/PyQtWebEngine[${PYTHON_USEDEP}]
+		dev-python/pyside2[${PYTHON_USEDEP}]
+		dev-python/pyside2[3d,charts,concurrent,datavis,designer,gui,help]
+		dev-python/pyside2[location,multimedia,network,opengl(+),positioning]
+		dev-python/pyside2[printsupport,qml,quick,script,scripttools,scxml]
+		dev-python/pyside2[sensors,serialport(+),speech,sql,svg,testlib]
+		dev-python/pyside2[webchannel,webengine,websockets,widgets,x11extras]
+		dev-python/pyside2[xml,xmlpatterns]
+)"
+
+distutils_enable_tests pytest
+
+src_prepare() {
+	default
+	if ! use pyqt5; then
+		sed -i -e "s/from PyQt5.QtCore import/raise ImportError #/" qtpy/__init__.py || die
+	fi
+	if ! use pyside2; then
+		sed -i -e "s/from PySide2 import/raise ImportError #/" qtpy/__init__.py || die
+	fi
+
+	# Disable outdated PyQt4 and PySide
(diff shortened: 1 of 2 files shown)


dev-python/QtPy/QtPy-1.9.0-r4.ebuild
```diff
@@ -0,0 +1,117 @@
...
+# We have to use at least one of PyQt5 or PySide, but if we want to use bindings
+# that are only available in one of the implementations we have to use that
+# implementation.
+REQUIRED_USE="
+	|| ( pyqt5 pyside2 )
+	pyqt5? (
+		qml? ( declarative )
+		quick? ( declarative )
+	)
+	pyside2? ( declarative? ( qml quick ) )
```
mgorny, line 34: Do we really need these 3 flags separately? At a first glance, it would be easier on the user if you put all three under, say, `USE=declarative`.

dev-python/QtPy/QtPy-1.9.0-r4.ebuild
```diff
@@ -0,0 +1,117 @@
...
+
+	bluetooth? ( pyqt5 )
+	dbus? ( pyqt5 )
+	networkauth? ( pyqt5 )
+	ssl? ( pyqt5 )
```
mgorny, line 39: This looks weird. PySide2 doesn't support SSL?

dev-python/QtPy/QtPy-1.9.0-r4.ebuild
```diff
@@ -0,0 +1,117 @@
...
+		dev-python/PyQt5[${PYTHON_USEDEP},bluetooth,dbus,declarative,designer,gui,help,location,multimedia,network,networkauth,opengl,positioning,printsupport,sensors,serialport,sql,ssl,svg,testlib,webchannel,webkit,websockets,widgets,x11extras,xml(+),xmlpatterns]
+		dev-python/PyQtWebEngine[${PYTHON_USEDEP}]
+		dev-python/pyside2[${PYTHON_USEDEP},3d,charts,concurrent,datavis,designer,gui,help,location,multimedia,network,opengl(+),positioning,printsupport,qml,quick,script,scripttools,scxml,sensors,serialport(+),speech,sql,svg,testlib,webchannel,webengine,websockets,widgets,x11extras,xml,xmlpatterns]
+)"
+
+distutils_enable_tests pytest
+
+src_prepare() {
+	default
+	use pyqt5 || sed -i -e "s/from PyQt5.QtCore import/raise ImportError #/" qtpy/__init__.py || die
```
mgorny, line 95: Please don't chain `||`, that is unnecessarily hard to read.

dev-python/QtPy/QtPy-1.9.0-r4.ebuild
```diff
@@ -0,0 +1,117 @@
...
+	sed -i -e "s/from PyQt4.QtCore import/raise ImportError #/" qtpy/__init__.py || die
+	sed -i -e "s/from PySide import/raise ImportError #/" qtpy/__init__.py || die
+}
+
+python_test() {
+	export QT_API="pyqt5"
+	virtx pytest -vv
+	export QT_API="pyside2"
+	virtx pytest -vv
+	unset QT_API
```
mgorny, line 109:
```suggestion
	local -x QT_API
	for QT_API in pyqt5 pyside2; do
		virtx epytest
	done
```

Also, it may be possible to move `virtx` into `src_test` to wrap the whole test phase in a single virtx.

dev-python/QtPy/QtPy-1.9.0-r4.ebuild
```diff
@@ -0,0 +1,117 @@
...
+}
+
+pkg_postinst() {
+	if use pyqt5 && use pyside2; then
+		ewarn "You have enabled both PyQt5 and PySide2, note that QtPy will
```
mgorny, line 114: Use two ewarns.

mgorny (changes requested):

dev-python/QtPy/QtPy-1.9.0-r4.ebuild
AndrewAmmerlaan: Good point

dev-python/QtPy/QtPy-1.9.0-r4.ebuild
AndrewAmmerlaan: Fair point, this is indeed wrong

dev-python/QtPy/QtPy-1.9.0-r4.ebuild
```diff
@@ -0,0 +1,119 @@
...
+# added anyway with a (+) to make it future proof if the ebuild were to change
+# this behaviour in the future.
+#
+# The PySide2 ebuild currently enables opengl and serialport support
+# unconditionally, the flag is added anyway with a (+) to make it future proof
+# if the ebuild were to change this behaviour in the future.
+#
+RDEPEND="
+	pyqt5? (
+		dev-python/PyQt5[${PYTHON_USEDEP},bluetooth?,dbus?,designer?,declarative?,gui?,help?,location?,multimedia?,network?,networkauth?,opengl?,positioning?,printsupport?,sensors?,serialport?,sql?,svg?,testlib?,webchannel?,webkit?,websockets?,widgets?,x11extras?,xml(+)?,xmlpatterns?]
```
mgorny, line 66: To make CI happy, you need to split this, i.e.:

```
dev-python/PyQt5[someflags...]
dev-python/PyQt5[moreflags...]
```

dev-python/QtPy/QtPy-1.9.0-r4.ebuild
```diff
@@ -0,0 +1,141 @@
...
+
+DESCRIPTION="Abstraction layer on top of PyQt5 and PySide2 and additional custom QWidgets"
+HOMEPAGE="https://github.com/spyder-ide/qtpy"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64 ~x86"
+IUSE="
+	pyqt5 pyside2 3d bluetooth dbus charts concurrent datavis declarative
```
mgorny, line 18: Maybe separate `pyqt5 pyside2` from the rest of the flags, and choose one of them as the default (I guess pyqt5 since that's what upstream defaults to).

mgorny (commented):
Hmm, this is still wrong ;-/.

If you depend on `QtPy[gui]` it will force either `PyQt5[gui]` or `pyside2[gui]` but the final package will still have no way of knowing which one was installed. This would have worked if you used `^^` for the switch but not `||`.

To be honest, `^^` is my best bet right now.

AndrewAmmerlaan:
> Hmm, this is still wrong ;-/.
> 
> If you depend on `QtPy[gui]` it will force either `PyQt5[gui]` or `pyside2[gui]` but the final package will still have no way of knowing which one was installed. This would have worked if you used `^^` for the switch but not `||`.
> 
> To be honest, `^^` is my best bet right now.

I had this thought too yesterday, but I concluded then that `||` should not be a problem. Because in principle there are three options if a package depends on QtPy[gui]:
1) I have USE="pyqt5", in which case pyside2 is disabled and PyQt5[gui] is pulled in, end result is that PyQt5 is used.
2) I have USE="pyside2", in which case PyQt5 is disabled and pyside2[gui] is pulled in, end result is that pyside2 is used.
3) I have USE="pyside2 pyqt5", in which case nothing is disabled and both PyQt5[gui] *and* pyside2[gui] are pulled in, end result is that PyQt5 is used because it is the default, unless QT_API overrides this (the user is notified about this). 

In all three situations the package will work, and we should not run into ModuleNotFound errors even in situation three because both PyQt5[gui] *and* pyside2[gui] are pulled in (which is precisely what was not happening in -r3 because it had the || dep). Correct me if I'm wrong, but from this I concluded that the added constraint of `^^` is not necessary.

mgorny:
Right, you are correct. However, there's still another problem: if you USE-dep on a flag that's only supported by pyside2 but PyQt5 is installed too. The packages could technically force pyside2 use but I don't think we can really expect people to figure that out.

That said, are you sure that there are this many unique features? Or is that a problem of missing USE flags on either ebuild?

AndrewAmmerlaan:
> That said, are you sure that there are this many unique features? Or is that a problem of missing USE flags on either ebuild?

AFAIK, all/most of these features are in fact not unique to one implementation, it is merely a matter of the corresponding ebuild missing a flag to enable building this feature. (Last I checked the pyside2/pyqt5 ebuilds have some comments about these missing features). If you check the output of the test phase, you'll notice that some tests are skipped even though we have enabled every single flag on pyqt5/pyside2. The reason is that even though we have enabled everything we are still missing some features that this implementation in principle supports, but that the corresponding ebuild unconditionally disables.

mgorny:
I'm sorry to bother you with this but could you try looking into enabling missing features in PyQt5, to at least reduce the divergence?

My personal preference would be to decrease the number of flags in the long run (if only to avoid having to ask devs to check 40 possible flags for what their packages need) but I can live with that many.

AndrewAmmerlaan:
> I'm sorry to bother you with this but could you try looking into enabling missing features in PyQt5, to at least reduce the divergence?

Sure, I'll add this to the to-do list, this will take some time though. I actually briefly looked into this before when I was working on -r3, and the main issues seems to be that some deps (e.g. qt3d) do not have all the keywords that e.g. pyqt5 has. So this will probably involve quite some keywording requests and/or some specific use masks for a couple of arches.

Anyway, would it be okay to push this as is for now, and just relax the REQUIRED_USE constraints as more flags are added to pyqt5/pyside2?

mgorny:
No, it's not ok, as this still involves the possibility of breaking revdeps.

As a temporary solution, I suppose you could remove all the flags that aren't common to both packages. I'm not 100% sure but I think this will be ok then. If you do that, I'll take one final look.

dev-python/QtPy/QtPy-1.9.0-r4.ebuild
```diff
@@ -0,0 +1,136 @@
...
+IUSE="+pyqt5 pyside2 ${_IUSE_QT_MODULES}"
+
+REQUIRED_USE="|| ( pyqt5 pyside2 )"
+
+# These flags are currently *not* common to both the PySide2 and PyQt5 ebuild
+# Disable them for now, please check periodically if this is still up to date.
+# 	bluetooth? ( pyqt5 )
+# 	dbus? ( pyqt5 )
+# 	networkauth? ( pyqt5 )
+# 	webkit? ( pyqt5 )
```
mgorny, line 32: For the record, I think we don't need to be concerned about webkit as it's deprecated.

AndrewAmmerlaan:
> No, it's not ok, as this still involves the possibility of breaking revdeps.
> 
> As a temporary solution, I suppose you could remove all the flags that aren't common to both packages. I'm not 100% sure but I think this will be ok then. If you do that, I'll take one final look.

Removed those flags, I think it should be fine like this. Because now the QtPy ebuild doesn't advertise bindings that are not supported by both implementations, so ebuilds can't depend on these bindings and therefore unexpected failures should not happen. If we ever add a package that should depend on e.g. QtPy[3d] then we should just fix e.g. the PyQt5 ebuild to support this flag. (I will look into enabling support for some of those missing flags in the PyQt5/pyside2 ebuilds later)

dev-python/QtPy/QtPy-1.9.0-r4.ebuild
```diff
@@ -0,0 +1,136 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7..8} )
+
+inherit distutils-r1 virtualx
+
+DESCRIPTION="Abstraction layer on top of PyQt5 and PySide2 and additional custom QWidgets"
+HOMEPAGE="https://github.com/spyder-ide/qtpy"
```
mgorny, line 11:
```suggestion
HOMEPAGE="https://github.com/spyder-ide/qtpy/"
```

dev-python/QtPy/QtPy-1.9.0-r4.ebuild
```diff
@@ -0,0 +1,136 @@
...
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64 ~x86"
+
+_IUSE_QT_MODULES="
```
mgorny, line 18: unset it after use

mgorny (approved):
Few nits inline. If you've confirmed that the tests pass now, ship it!

AndrewAmmerlaan:
>  If you've confirmed that the tests pass now, ship it!

Awesome, I think it also makes sense to last-rites `app-eselect/eselect-QtPy` since it does not really do anything any more and will only confuse people, not to mention that it will break things if the eselect setting does not match the USE setting.

mgorny:
Yes, please. I suppose you beat the record of shortest time to your first last rites as a dev ;-).

juippis:
Thank you **both** for working on this! Really refreshing to see compromises and collaboration lead into satisfying outcome.

PR #20872, opened 2021-05-18
fix python_targets_python3_10 USE flag description[please reassign]

arthurzam:
I think it is one of the most useless pull request I have posted, but when I noticed it it bugged me.
@mgorny , could you check that it is truly a typo?


mgorny:
It's not useless at all. It's a very good catch!

PR #20877, opened 2021-05-19
dev-util/rebar-bin: Bump to 3.16.1

00-matt:


juippis (commented):
Can we remove the 3.15.1 in a **separate** commit to get rid of the vulnerable version ASAP? 

Since this is a -bin version, doubt it will introduce breakages.

juippis (approved):

PR #20878, opened 2021-05-19

arthurzam:
Continuing scanning the CI output for `PythonCompatUpdate`, and calling `ebuild ... test`
Some of them even add support for `python3_9`

### DIFF
--- a/dev-python/pysvg/pysvg-0.2.2_p3.ebuild
+++ b/dev-python/pysvg/pysvg-0.2.2_p3.ebuild
@@ -1,9 +1,9 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7,8} )
+PYTHON_COMPAT=( python3_{7..10} )
 
 inherit distutils-r1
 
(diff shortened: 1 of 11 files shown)


dev-python/python-editor/python-editor-1.0.4-r1.ebuild
```diff
@@ -1,9 +1,9 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-PYTHON_COMPAT=( pypy3 python3_{7..9} )
+PYTHON_COMPAT=( pypy3 python3_{7..10} )
```
mgorny, line 6: How did you test it?

dev-python/pysvg/pysvg-0.2.2_p3.ebuild
```diff
@@ -1,9 +1,9 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7,8} )
+PYTHON_COMPAT=( python3_{7..10} )
```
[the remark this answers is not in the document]
arthurzam: merged and tested the sample from upstream at location:
https://raw.githubusercontent.com/alorence/pysvg-py3/master/sample/tutorial.py

dev-python/python-editor/python-editor-1.0.4-r1.ebuild
arthurzam: Viewed the source code [here](https://github.com/fmoo/python-editor/blob/master/editor.py) (very simple therfore no problemt with 3.10)
and tried to run the test [here](https://github.com/fmoo/python-editor/blob/master/test.py).

mgorny:
Ok, thanks.

PR #20881, opened 2021-05-19
ARCH=riscv keywords request

dlan17:
do not merge!


dlan17:
pushed to main tree, so close

PR #20884, opened 2021-05-19
dev-java/sun-jai-bin: min java 1.8

vaukai:


fordfrog:
@vaukai did you check bytecode version of the binary jar? if it's lower than 1.8, it won't run with java 11 and will need `:1.8`.

fordfrog:
@vaukai thanks!

PR #20886, opened 2021-05-19
*/*: more python3_10 support

dAgbeve:


mgorny:
I'm sorry but a few screenfulls of commits described as `*/*` aren't helpful. Start small.

thesamesam:
1. Could you fix the commit messages please?
2. Have you revbumpd everything or something? It’s not needed for Python compatibility changes unless we have to patch.

PR #20887, opened 2021-05-19
x11-terms/kitty and kitty-terminfo: bump to 0.20.3

end222:
Update kitty and kitty-terminfo to 0.20.3 and update Python compatibility to include Python 3.9.


thesamesam:
Thanks!

PR #20888, opened 2021-05-19
*/* drop deprecated ltprune eclass

ConiKost:
Time to get ride of ltprune eclass.

### DIFF
--- a/sci-libs/hdf/hdf-4.2.13-r1.ebuild
+++ b/sci-libs/hdf/hdf-4.2.13-r1.ebuild
@@ -5,7 +5,7 @@ EAPI=6
 
 FORTRAN_NEEDED=fortran
 
-inherit fortran-2 toolchain-funcs autotools flag-o-matic ltprune
+inherit fortran-2 toolchain-funcs autotools flag-o-matic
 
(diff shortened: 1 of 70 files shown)


sci-libs/hdf/hdf-4.2.13-r1.ebuild
```diff
@@ -61,7 +61,7 @@ src_configure() {
 
 src_install() {
 	default
-	use static-libs || prune_libtool_files --all
+	use static-libs || find "${ED}" -name '*.la' -delete || die
 	dodoc release_notes/{RELEASE,HISTORY,bugs_fixed,misc_docs}.txt
```
thesamesam, line 65: Convert these into an if for readability?

ConiKost: I can do that. I just took it that way, as several ebuilds start using this way, to make it consistent. But I don't mind to change it to an full if.

ConiKost: Done.

SoapGentoo (approved):

PR #20891, opened 2021-05-20
net-im/skypeforlinux: Support All Upstream Stable Release Versions

digitall:
This reverts commit 76ddb59e4466c89cdc99be3fd01fd943cfd20ded.

This ensures that all versions present in upstream stable i.e.
https://repo.skype.com/rpm/stable/ are present in Portage.

The older versions are still available there as they are still
supported and the newer stable releases may have problems on some
systems so best to track the upstream.

Bug: https://bugs.gentoo.org/777606

Signed-off-by: David G Turner <digitall@scummvm.org>


SoapGentoo:
Obsolete, both `8.67.0.96` and `8.68.0.100` are gone upstream.

digitall:
@SoapGentoo : You are correct and thus since there are no mirrors for the source RPMs, then this pull is obselete. However, this does not fix https://bugs.gentoo.org/777606 which is still open. I have just checked and both of the new ebuilds for the latest versions in portage fail to work correctly with the same bug rendering them unusable :|

PR #20893, opened 2021-05-20
x11-libs/libxkbcommon: Add support for wayland

lanodan:

### DIFF
--- a/x11-libs/libxkbcommon/libxkbcommon-1.3.0.ebuild
+++ b/x11-libs/libxkbcommon/libxkbcommon-1.3.0.ebuild
@@ -18,7 +18,7 @@ inherit meson multilib-minimal ${GIT_ECLASS} python-any-r1 virtualx
 DESCRIPTION="keymap handling library for toolkits and window systems"
 HOMEPAGE="https://xkbcommon.org/ https://github.com/xkbcommon/libxkbcommon/"
 LICENSE="MIT"
-IUSE="X doc static-libs test"
+IUSE="doc static-libs test wayland X"
 RESTRICT="!test? ( test )"
 SLOT="0"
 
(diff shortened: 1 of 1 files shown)


x11-libs/libxkbcommon/libxkbcommon-1.3.0.ebuild
```diff
@@ -18,17 +18,22 @@ inherit meson multilib-minimal ${GIT_ECLASS} python-any-r1 virtualx
...
 RESTRICT="!test? ( test )"
 SLOT="0"
 
 BDEPEND="
 	sys-devel/bison
 	doc? ( app-doc/doxygen )
 	test? ( ${PYTHON_DEPS} )
+	wayland? (
+		dev-util/wayland-scanner
+		>=dev-libs/wayland-protocols-1.12
```
mattst88, line 31: I think `wayland-protocols` should be in `DEPEND`.

x11-libs/libxkbcommon/libxkbcommon-1.3.0.ebuild
```diff
@@ -45,7 +50,7 @@ multilib_src_configure() {
 	local emesonargs=(
 		-Ddefault_library="$(usex static-libs both shared)"
 		-Dxkb-config-root="${EPREFIX}/usr/share/X11/xkb"
-		-Denable-wayland=false # Demo applications
+		$(meson_use wayland enable-wayland)
```
mattst88, line 53: This controls the building and installation of
````
./usr/libexec/xkbcommon/xkbcli-interactive-wayland
./usr/share/man/man1/xkbcli-interactive-wayland.1.bz2
````

I see that we're already installing

````
./usr/libexec/xkbcommon/xkbcli-interactive-x11
./usr/share/man/man1/xkbcli-interactive-x11.1.bz2
````

with USE=X. (It also controls some other things)

So... seems fine. Is this tool useful for you? I'm just unsure of the rationale.

mattst88 (changes requested):

x11-libs/libxkbcommon/libxkbcommon-1.3.0.ebuild
lanodan: Yeah, the USE=X requires to have `xorg-server` installed, which isn't required for wayland usage in gentoo and I'm not aware of another wayland tool similar to this one or x11-apps/xev (`WAYLAND_DEBUG=1` is nice but it doesn't decodes events in something a human could directly use).

mattst88:
Thanks!

PR #20894, opened 2021-05-20
dev-python/*: python3_10 support

dAgbeve:


mgorny:
Twisted is already a part of #20835.

PR #20900, opened 2021-05-20
www-servers/lighttpd: Version bump (v1.4.59)

herbmillerjr:
- Eliminated `remove_non_essential()` and tied USE flags to condition against `minimal` USE flag to create a minimal installation instead
  - Installs basic docs even with `minimal` USE flag set 
  - First line of modules that used to be deleted with `rm` are controlled by PCRE USE flag 
  - Remaining modules are controlled by their respective USE flags 
- Replaced `use <module> &&` syntax with `if use <module>; then; fi` syntax 
- spawn-fcgi no longer gets built by this package, so removed line that deletes it after compiling
- fixed a bug where lighttpd was looking in the wrong place for nspr headers
- added Zstandard output compression support


www-servers/lighttpd/lighttpd-1.4.59.ebuild
```diff
@@ -0,0 +1,250 @@
...
+		!brotli
+		!bzip2
+		!doc
+		!lua
+		!maxminddb
+		!pcre
+		!rrdtool
+		!webdav
+		!zlib
+	)
```
juippis, line 31: Yea no this is unacceptable, if you really want to introduce the minimal flag, you should do it purely in `src_configure` and `pkg_postinst`.

herbmillerjr: I don't see why we need it. I was just afraid to rip it out. Seems to be all controlled by the USE flags now, as I was told in the previous pull request. I've removed it now.

juippis (approved):

mattst88:
@herbmillerjr: please respond to https://bugs.gentoo.org/855146

PR #20903, opened 2021-05-21
sys-boot/syslinux: Add DEPEND on sys-apps/util-linux

m-gupta:
Some syslinux files needs headers from util-linux
e.g. extlinux/xfs_sb.h includes <uuid/uuid.h>.

### DIFF
--- a/sys-boot/syslinux/syslinux-6.04_pre1-r2.ebuild
+++ b/sys-boot/syslinux/syslinux-6.04_pre1-r2.ebuild
@@ -28,6 +28,7 @@ RDEPEND="sys-fs/mtools
 DEPEND="${RDEPEND}
 	dev-lang/nasm
 	>=sys-boot/gnu-efi-3.0u
+	sys-apps/util-linux
 	virtual/os-headers"
 
 S=${WORKDIR}/${P/_/-}


m-gupta:
friendly ping for review..

sys-boot/syslinux/syslinux-6.04_pre1-r2.ebuild
```diff
@@ -28,6 +28,7 @@ RDEPEND="sys-fs/mtools
 DEPEND="${RDEPEND}
 	dev-lang/nasm
 	>=sys-boot/gnu-efi-3.0u
+	sys-apps/util-linux:=
```
mattst88, line 31: No need for the subslot operator (`:=`) since `util-linux` doesn't have sublots. Also, if the only dependency is due to headers, a subslot operator wouldn't be warranted.

mattst88 (changes requested):

sys-boot/syslinux/syslinux-6.04_pre1-r2.ebuild
m-gupta: done

thesamesam: We need libuuid too, so I've added to RDEPEND.

m-gupta:
thanks!

PR #20906, opened 2021-05-21
app-emacs/jde: min java 1.8

vaukai:
Bug: https://bugs.gentoo.org/787326


ulm:
Have you tested this at runtime?

vaukai:
> Have you tested this at runtime?

Not using emacs here and no idea how to test at runtime.  Guess that's also the reason why this package is homed in `app/emacs`?
Could you help with such runtime test?

PR #20909, opened 2021-05-21

arthurzam:
- Most packages was simple bump and test with `ebuild ... test`
- Please check that my change for `dev-python/python-gnupg` is correct and OK. Wasn't sure here so please extra check here.
- for `dev-python/pyasn1` and `dev-python/pyasn1-modules` a small redundant USE flag remove.
- for `dev-python/testpath` I checked multiple times about the `pathlib2` dependency, and after checking the code I found it isn't needed (it is a backup if `import pathlib` fails, which shouldn't happen now). Tested on all runtime versions supported and passed tests on all of them, even `pypy3`. So I removed it. Could you please also check it also passes tests for you (just to be sure)?

### DIFF
--- a/dev-python/python-gnupg/python-gnupg-0.4.7.ebuild
+++ b/dev-python/python-gnupg/python-gnupg-0.4.7.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( pypy3 python3_{7,8,9} )
+PYTHON_COMPAT=( pypy3 python3_{7..10} )
 
 inherit distutils-r1
 
@@ -21,9 +21,10 @@ SLOT="0"
 RDEPEND="app-crypt/gnupg"
 DEPEND="${RDEPEND}"
 
+distutils_enable_tests unittest
+
 python_test() {
 	# NO_EXTERNAL_TESTS must be enabled,
 	# to disable all tests, which need internet access.
-	NO_EXTERNAL_TESTS=1 "${EPYTHON}" -m unittest discover -v ||
-		die "Tests failed with ${EPYTHON}"
+	NO_EXTERNAL_TESTS=1 eunittest
 }
--- a/dev-python/python-jose/python-jose-3.2.0-r1.ebuild
+++ b/dev-python/python-jose/python-jose-3.2.0-r1.ebuild
@@ -25,12 +25,14 @@ RDEPEND="
 "
 
 distutils_enable_tests pytest
+distutils_enable_sphinx docs
 
 python_prepare_all() {
 	sed -e '/pytest-runner/d' \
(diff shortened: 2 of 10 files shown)


dev-python/python-jose/python-jose-3.2.0-r1.ebuild
```diff
@@ -1,9 +1,9 @@
-# Copyright 1999-2020 Gentoo Authors
```
mgorny, line 1: If you don't mind some extra work, please switch to `epytest --deselect` here.

dev-python/python-gnupg/python-gnupg-0.4.7.ebuild
```diff
@@ -21,9 +21,10 @@ SLOT="0"
...
 DEPEND="${RDEPEND}"
 
+distutils_enable_tests unittest
+
 python_test() {
 	# NO_EXTERNAL_TESTS must be enabled,
 	# to disable all tests, which need internet access.
-	NO_EXTERNAL_TESTS=1 "${EPYTHON}" -m unittest discover -v ||
-		die "Tests failed with ${EPYTHON}"
+	NO_EXTERNAL_TESTS=1 distutils-r1_python_test
```
mgorny, line 29: `eunittest`

dev-python/testpath/testpath-0.5.0.ebuild
```diff
@@ -16,13 +16,5 @@ LICENSE="MIT"
 SLOT="0"
 KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~ppc ~ppc64 ~s390 ~sparc ~x86 ~x64-macos"
 
-BDEPEND="
-	>=dev-python/pyproject2setuppy-15
```
mgorny, line 20: That's not redundant. It doesn't work with older versions.

mgorny (changes requested):

dev-python/testpath/testpath-0.5.0.ebuild
arthurzam: Fixed.
Sorry for missing it.

dev-python/python-gnupg/python-gnupg-0.4.7.ebuild
arthurzam: Done.

dev-python/python-jose/python-jose-3.2.0-r1.ebuild
arthurzam: Have been done gladly. I collect all your comments to make every PR better than the previous one (and to fix various python ebuilds found in tree).

I did it using a full array (similar to how it was done at `dev-python/virtualenv`) so it will be better ready to future needs.

PR #20911, opened 2021-05-21
app-editors/vim: add flag ipv6

vitaly-zdanevich:
Why not --disable-ipv6 see
https://vi.stackexchange.com/questions/31335/building-cannot-disable-ipv6-unrecognized-option
looks like this is the only correct way.

Closes: https://bugs.gentoo.org/790446

Signed-off-by: Vitaly Zdanevich <zdanevich.vitaly@ya.ru>


monsieurp:
Thank you for your contribution! I have merged the PR. :)

PR #20912, opened 2021-05-21
dev-libs/wt: new package

captaincrutches:
This adds Wt as a new package, at the current version 4.5.0.

Wt is a C++ framework for developing web applications.  Its source is [here](https://github.com/emweb/wt) and installation instructions are [here](https://www.webtoolkit.eu/wt/doc/reference/html/InstallationUnix.html).

The build itself is a fairly straightforward Cmake build, with just a few tweaks needed:
- Patch a [build failure](https://redmine.webtoolkit.eu/issues/8467) in gcc-11
- Blank out [WtInstall.cmake](https://github.com/emweb/wt/blob/master/WtInstall.cmake) since all it does is create `/var/run/wt` and make it owned by the `apache` user, which may not exist.  According to a QA notice, we don't want to install into `/var/run` anyway, and the FCGI connector does [attempt to create that directory](https://github.com/emweb/wt/blob/master/src/fcgi/Server.C#L354) at runtime.
- Install libraries to the correct libdir (i.e. `lib64`) instead of just `lib`

Besides that, we just turn on/off a few different cmake switches based on use flags and call it a day.

The reasoning behind the `REQUIRED_USE` is:
- If you're building `Wt::Dbo`, you should pick at least one backend
- Opengl is used for server-side image rendering bug [also requires GraphicsMagick to be installed](https://github.com/emweb/wt/blob/master/CMakeLists.txt#L528).

Tested on my ~amd64 machine with various combinations of flags and it seems good.  Everything's installed that I expect, and a local project that consumes Wt via `find_package(wt)` builds successfully against it.


captaincrutches:
Yeah, sounds fine to me - while this would be nice to have in gentoo, my own motivating cases to add it (and thus maintain it!) have since disappeared.  If someone else wants to pick up the slack, do feel free.

PR #20914, opened 2021-05-21
app-arch/upx: fix CVE-2020-24119

winterheart:
Bug: https://bugs.gentoo.org/790281


a17r (changes requested):
- Should `dev-lang/perl` be in `BDEPEND` instead?
- `!app-arch/upx-bin` blocker should be in `RDEPEND` only, easily achieved after perl moves to `BDEPEND`

winterheart:
Done.

PR #20917, opened 2021-05-21
dev-games/aseprite: fix #680060

winterheart:
Shared libloadpng comes with shared allegro[png]. USE=bundled-libs will
disable allegro dependency, which leads to building error.

Closes: https://bugs.gentoo.org/680060

### DIFF
--- a/dev-games/aseprite/aseprite-1.1.9.ebuild
+++ b/dev-games/aseprite/aseprite-1.1.9.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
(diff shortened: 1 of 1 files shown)


a17r (changes requested):
What about "Fix unused CMake var" instead of just #bugno in git summary.

winterheart:
Because "unused CMake var" is not cause of error.

ionenwks:
Set something appropriate then, the point was that the summary should shortly describe what it's fixing while the bugref be left in the message's body.

winterheart:
I got super-descriptive summary in commit.

a17r:
That's very good, but doesn't let you get away with just a number for the summary. `git log` does not look that way either, we're humans communicating in words not numbers.

winterheart:
Done.

a17r (commented):
Thanks for fixing, this is ready for merge now. Unfortunately building with USE=bundled-libs got us a new error: https://bugs.gentoo.org/791682

PR #20921, opened 2021-05-22
media-libs/libsdl2: Add live ebuild

lanodan:

### DIFF
--- a/media-libs/libsdl2/libsdl2-9999.ebuild
+++ b/media-libs/libsdl2/libsdl2-9999.ebuild
@@ -1,0 +1,212 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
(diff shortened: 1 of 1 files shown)


a17r:
We already have too many 9999 ebuilds in tree that no one cares about. This can go in only if
- @gentoo/games proj decides to make it part of their workflow
- you want to co-maintain this package with @gentoo/games provided they are ok with it

lanodan:
> We already have too many 9999 ebuilds in tree that no one cares about. This can go in only if
> 
> * @gentoo/games proj decide to make it part of their workflow
> * you want to co-maintain this package with @gentoo/games provided they are ok with it [and respond to your PRs]

Yeah, makes total sense, I would be fine with co-maintaining it but could also just stay in my overlay for my own needs.

chewi:
I don't think we need this, to be honest. Its releases aren't too infrequent and the changes usually aren't that exciting.

PR #20923, opened 2021-05-22
app-backup/dar: v2.7.1

ceamac:
Bump dar to 2.7.1 (minor release, mostly bugfixes)
2.7.0 is only a few days old in trunk, so it can be removed.


a17r (changes requested):
`REQUIRED_USE` should be moved below `IUSE`, otherwise lgtm

ceamac:
Thank you!

a17r:
Thank you!

PR #20924, opened 2021-05-22
net-libs/libnsl: provided static-libs USE flag

dannftk:


trofi:
Can you add a note to description why do you need static libraries for it?

a17r:
Dropped in commit 8a757dd2fd37b50a5223f56ec4880b15f064adc3, why reintroduce?

floppym:
As others have said, please provide some justification for this.

dannftk:
@a17r, why has it been dropped by @floppym? I made it be capable of being conditionally configured by using static-libs as a USE flag. If you do not need 'static-libs' for the package then do not provided USE flag, isn't this a wise consideration?
Sometimes I need static linkage with this library, therefore I need this USE flag enabled

ionenwks:
> why has it been dropped

See policy guide:
https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0302

Some exceptions can be *considered*, but otherwise slowly removing support where possible is something any developer would do.

floppym:
> Sometimes I need static linkage with this library, therefore I need this USE flag enabled

What are you building that needs to be linked statically with libnsl? If there's a good reason for it, we can certainly add the USE flag.

dannftk:
@floppym, I am building [openBcm](https://github.com/Broadcom-Network-Switching-Software/OpenBCM) project, and if I do not use 'shared'-libraries mode while compiling it requires, when linking, everything to be static including libnsl. That is why I have been interested in a reason of not providing `static-libs` USE flag for the package libnsl whereas other libraries have this flag as possible
I do not know if there are cases where this flag is possible to be harmful and I do not know if building `openBcm`, that is not a part of the portage, is a good reason to re-introduce this flag. But, as you know, you might disable this flag by providing `-static-libs` in particular for the libnsl, mightn't you?

a17r:
Is there any reason to not build openBcm in shared-lib mode?

PR #20926, opened 2021-05-22
By default /usr/lib64/pppd/2.4.7/ path used. This will use right path…

tochlab:
… from portage.

Bug: https://bugs.gentoo.org/722718

### DIFF
--- a/net-vpn/networkmanager-l2tp/networkmanager-l2tp-1.8.2.ebuild
+++ b/net-vpn/networkmanager-l2tp/networkmanager-l2tp-1.8.2.ebuild
@@ -46,8 +46,12 @@ src_prepare() {
 }
 
 src_configure() {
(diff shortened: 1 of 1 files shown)


jkroonza (commented):
This looks good to me.  Don't use networkmanager, but this did catch my eye.

net-misc/dahdi-tools has the following in ./configure so perhaps an upstream patch may be appropriate:

```
19101 level_file="$ppp_path/include/pppd/patchlevel.h"
19102 PPP_VERSION=
19103 if test "$ppp_path" != '' && test -r "$level_file"; then
19104     PPPD_VERSION=`awk -F '"' '/VERSION/ { print $$2; }' $level_file`
19105 fi
```

ppp_path gets set based on --with-ppp (and defaults to /usr)

This then gets replaced into appropriate Makefile.am files.  Just a suggestion.

PR #20927, opened 2021-05-22
[DO NOT MERGE] OpenStack removal Ci

mgorny:
This thing seems to be a maintenance nightmare. At this point, starting over will probably be less effort than trying to fix all the ebuilds. The maintainers don't seem to care at all. In a few days, it all will again be behind default Python. Let's see how much we can remove right now.


chutz:
Please keep `dev-python/bandit` (I can take maintainership if current maintainer is not keeping it up to date).

Also these are optional dependencies for `app-admin/salt`, they are currently required for tests to pass (though I could theoretically delete the tests that require them):
```
dev-python/python-neutronclient
dev-python/python-novaclient
```

mgorny:
Ok, bandit should be doable but please enable tests in oslotest as a sign of good faith.

mgorny:
Now includes kombu and py-amqp.

chutz:
> Ok, bandit should be doable but please enable tests in oslotest as a sign of good faith.

Done, `dev-python/oslotest-4.4.1` has working tests.

prometheanfire:
why is the python group removing a package not owned by the python group?

mgorny:
You can call it QA removing a huge PITA package set after you've been called multiple times to start improving them, and you ignored every single of these requests. Right now they're blocking Python 3.9.

chutz:
> Also these are optional dependencies for `app-admin/salt`, they are currently required for tests to pass (though I could theoretically delete the tests that require them):
> 
> ```
> dev-python/python-neutronclient
> dev-python/python-novaclient
> ```
It turns out I was mistaken about this, it appears that `app-admin/salt` just needs the USE flags deleted to no longer depend on these.

prometheanfire:
think we should close this now?

PR #20930, opened 2021-05-22
media-libs/{ilmbase,openexr}, dev-libs/imath, dev-python/pyilmbase: implement proper slotting

waebbl:
dev-libs/imath: bump to 3.0.2 and drop 3.0.1
dev-libs/openexr:3: bump to 3.0.3 and drop 3.0.1

Improves slotting to enable parallel installation of media-libs/{ilmbase,openexr}:0 and dev-libs/imath, media-libs/openexr:3, as well as dev-libs/imath[python] and dev-python/pyilmbase:0.

**Things to discuss**

- revdeps in this PR or in a seperate (big) PR or each one in it's own PR? Latter will probably need masking to avoid breaking of revdep building
- is this a good way to handle slotting on these packages? What about using an eselect module for this? Advantages, disadvantages of both ways

Please review. Feedback on above points welcome.


atupone (approved):
I am OK with this (for media-gfx/exact-image)

waebbl:
> I am OK with this (for media-gfx/exact-image)

@atupone Thanks for your feedback, but note this is not yet finished. Currently only the packages which lead to CI issues are covered. There are more revdeps on which I'm still working.

media-gfx/gmic/gmic-2.9.7-r1.ebuild
```diff
@@ -0,0 +1,175 @@
...
+DEPEND="${COMMON_DEPEND}
+	gimp? ( dev-qt/linguist-tools )
+	krita? ( dev-qt/linguist-tools )
+	qt5? ( dev-qt/linguist-tools )
+"
+BDEPEND="virtual/pkgconfig"
+
+PATCHES=(
+	"${FILESDIR}"/${PN}-2.9.2_ipa-sra.patch
+	"${FILESDIR}"/${P}-find-slotted-openexr.patch
```
mkszuba, line 72: Please use ${PN}-2.9.7 instead of ${P} here, I hate having to update _PATCHES_ on version bumps.

media-gfx/darktable/darktable-3.4.1-r1.ebuild
```diff
@@ -0,0 +1,168 @@
...
+	openexr? ( <media-libs/openexr-3.0.0:0= )
+	webp? ( media-libs/libwebp:0= )"
+RDEPEND="${DEPEND}
+	kwallet? ( >=kde-frameworks/kwallet-5.34.0-r1 )"
+
+PATCHES=(
+	"${FILESDIR}"/${PN}-find-opencl-header.patch
+	"${FILESDIR}"/${PN}-3.0.2_cmake-march-autodetection.patch
+	"${FILESDIR}"/${PN}-3.4.0_jsonschema-automagic.patch
+	"${FILESDIR}"/${P}-find-slotted-openexr.patch
```
mkszuba, line 82: Please use ${PN}-x.y.z instead of ${P} here, I hate having to update PATCHES on version bumps.

mkszuba (changes requested):
To be honest, I am not a fan. My comments regarding media-gfx/darktable and media-gfx/gmic aside, it feels to me that at least in case of CMake it would make more sense to handle slot management in the modules installed by media-gfx/openexr rather than patching ebuilds all over the place to look for OpenEXR-2 rather than OpenEXR, and so on.

I also wonder why you haven't moved openexr-2 from SLOT=0 to SLOT=2.

waebbl:
> To be honest, I am not a fan. My comments regarding media-gfx/darktable and media-gfx/gmic aside, it feels to me that at least in case of CMake it would make more sense to handle slot management in the modules installed by media-gfx/openexr rather than patching ebuilds all over the place to look for OpenEXR-2 rather than OpenEXR, and so on.
> 

I agree. I'm all but happy to have to patch all revdeps. Do you have any suggestion in how this could be done by the modules? I was thinking of this way too, but didn't find a way. The packages uses new style config mode modules and not old style `Find<PACKAGE>.cmake` mode. The code to generate these files is only a few lines of code[1] and the basic config file is almost entirely created by internal cmake magic from the skeleton[2], the `*Target*.cmake` files are created completely by internal cmake magic.

If you have a good idea in how I can add slot handling into the modules, I'd be more than happy to implement this. All my attempts to get to a proper solution upfront were left unanswered. Several attempts to ask for advice in `gentoo-dev-help` and `gentoo-proxy-maint` were not answered. I was told by several people to not post a thread in the `gentoo-dev` mailing list, when I asked if this would help in getting this solved. For almost a year now, I'm trying to get attention for the issues involved with the release of openexr/imath v3, with very little feedback.

> I also wonder why you haven't moved openexr-2 from SLOT=0 to SLOT=2.

I didn't want to add an extra slotmove. The package is the same as before the slot 3 versions were added. Additionally, I expect to have different slots for only a limited time and the v2.x versions to go away relatively quick. AFAICS upstream will support versions <3 only for a limited time.

[1] https://github.com/AcademySoftwareFoundation/openexr/blob/master/cmake/CMakeLists.txt#L225-L246
[2] https://github.com/AcademySoftwareFoundation/openexr/blob/master/cmake/OpenEXRConfig.cmake.in

mkszuba:
Let me be more straightforward this time: as it stands now this change is WAY too intrusive and I do NOT want it in any of the packages I maintain.

waebbl:
> Let me be more straightforward this time: as it stands now this change is WAY too intrusive and I do NOT want it in any of the packages I maintain.

Let me rephrase again. Do you have any *constructive* idea on how this could be done?

mkszuba:
As a matter of fact, I do. How about just leaving SLOT=0 ebuilds OpenEXR and friends exactly as they are and only modifying v3? That way you'll only have to modify ebuilds requiring OpenEXR-3 (and a cursory glance at your PR suggests these are currently in minority), and if a new ebuild requiring OpenEXR-3 makes to the tree without having been adapted for slotted versions building it will result in a simple-enough-to-understand "I'm sorry, your version is too old" error.

Everything else aside, if you really expects these slots not to persist for too long (which may or may not be wishful thinking, will have to see how it goes with revdeps of this particular library...) it would make sense to keep the amount of slotting-related overhead to an absolute minimum.

PS. Sorry for sounding harsh yesterday, GitHub did something odd in my browser and even having been told to uncollapse the comment list it still didn't show your response to my review - which led me to believe there hadn't been one.

waebbl:
Thanks for your feedback @mkszuba. At a first glance, your idea sound promising. It has, however some glitches from my POV:
- If I add a new slot and change only the v3 installation locations, I still have to change almost each of the packages, which are now in this PR. There are only a few, which already have a `:0=` slot operator specified. Most only have a `:=` slot operator, some don't even have this, so I get a lot of *Missing slot dependency* pkgcheck errors in the CI. I still like the idea, because even if I need to touch almost the same amount of packages, the impact will be smaller, if I only need to add a slot operator and this pulls in v2 and builds without further change.
- If I keep going with only one slot, I too, have to change most of the packages to restrict those, which don't yet support v3 to some <v3
- I also would like to focus on v3, because this is the future. Of course, we need to support v2, as long as we have packages which are not yet compatible with v3. For this, I'd like to go for *supporting* v3 and not for *requiring* v3. IMO the latter will probably take more time than the first.

> PS. Sorry for sounding harsh yesterday, GitHub did something odd in my browser and even having been told to uncollapse the comment list it still didn't show your response to my review - which led me to believe there hadn't been one.

All is good. I probably sounded too harsh as well a few days ago. I got close to my frustration tolerance with this issue.

mkszuba:
>     * If I add a new slot and change only the v3 installation locations, I still have to change almost each of the packages, which are now in this PR. There are only a few, which already have a `:0=` slot operator specified. Most only have a `:=` slot operator, some don't even have this, so I get a lot of _Missing slot dependency_ pkgcheck errors in the CI.

True that, then again this will be one-time, single-line changes with no patches to maintain.

>     * If I keep going with only one slot, I too, have to change most of the packages to restrict those, which don't yet support v3 to some <v3

I don't think I get your point here. Just to make it clear, I have suggested neither keeping one slot only nor having to eventually merge SLOT=3 into SLOT=0 - what I've had in mind has been something along the lines of "SLOT 0 for v2, for every other major version X use SLOT=X, keep slot numbers stable for as long as respective major versions are kept in the tree".

>     * I also would like to focus on v3, because this is the future. Of course, we need to support v2, as long as we have packages which are not yet compatible with v3. For this, I'd like to go for _supporting_ v3 and not for _requiring_ v3. IMO the latter will probably take more time than the first.

While it would certainly be nice if we could focus on v3 in Gentoo, I wouldn't hold my breath waiting for OpenEXR revdeps to be migrated to this version. Even though this transition will in all likelihood go more smoothly than that from Lua 5.1 to newer API versions (if only because there isn't much room for it going any _less_ smoothly), I am pretty sure it will at the very least take quite some time. Which is why I advocate the "work on v3 but in a way which keeps v2 maintenance overhead at absolute minimum" approach.
It would be rather helpful here were ilmbase/openexr upstream to embrace the idea of side-by-side presence of different major versions. Just look how (relatively) simple it is for Gentoo to maintain Gtk+ slots, for example! Have you by any chance tried upstreaming the idea yet?

waebbl:
> I don't think I get your point here. Just to make it clear, I have suggested neither keeping one slot only nor having to eventually merge SLOT=3 into SLOT=0 - what I've had in mind has been something along the lines of "SLOT 0 for v2, for every other major version X use SLOT=X, keep slot numbers stable for as long as respective major versions are kept in the tree".
> 

I actually did get you here. My point was only to show an alternative, someone has talked of in one of the bugs related to this and which I was rethinking after your reply. The way I'd like this to go, is in par with your suggestion, i.e. using different slots for major versions.

> It would be rather helpful here were ilmbase/openexr upstream to embrace the idea of side-by-side presence of different major versions. Just look how (relatively) simple it is for Gentoo to maintain Gtk+ slots, for example! Have you by any chance tried upstreaming the idea yet?

I've thought about upstreaming this several times, but haven't yet started a discussion on this topic. This suggests itself by the presence of the `<PKG>_OUTPUT_SUBDIR` cmake variable for all packages, thus already having a basic support for side-by-side installations of different versions. On a recent thread on their forum I learned they will have to support even v2.4 for some more time, because upstream wants (or needs) to support the VFX reference platform, which suggests that version in its current version. Might be a good time to actually start a topic about this.

waebbl:
@mkszuba May I ask what you're doing here?

waebbl:
Thanks for your review @mkszuba. although I haven't seen this as finished, due to the issue still present in povray.

mkszuba:
No worries, that issue was simply because the old version needed to be pointed to the right slot as well _i.e._ something trivial for me to address at merge time.

kde-frameworks/kimageformats/kimageformats-5.84.0-r1.ebuild
```diff
@@ -21,8 +21,16 @@ DEPEND="
...
+		|| (
+			(
+			media-libs/ilmbase:0
+			<media-libs/openexr-3.0.0:0
+			)
+			(
+			dev-libs/imath:3
+			media-libs/openexr:3
+			)
+		)
```
waebbl, line 33: @mkszuba: @a17r pointed me to an issue, we've missed with this ebuild. After an upgrade of openexr, no rebuild will be triggered. I'm gonna open a new PR for this.

waebbl:
Another issue happens with your darktable-3.6.0 ebuild @mkszuba. You restrict to `<openexr-3`, but the build system prefers v3, as suggested by OpenEXR upstream in their [PortingGuide](https://github.com/AcademySoftwareFoundation/Imath/blob/master/docs/PortingGuide2-3.md).
So, if both, openexr-2 and openexr-3 are installed, the build system will pick up v3 and fails to link `bin/darktable`, although the correct libraries are passed:

```
FAILED: bin/darktable 
: && /usr/bin/x86_64-pc-linux-gnu-gcc -O2 -pipe -march=bdver2 -frecord-gcc-switches -fstack-check -fstack-protector-strong -msse3 -Wall -Wformat -Wformat-security -Wshadow -Wtype-limits -Wvla -Wold-style-declaration -Wno-unknown-pragmas -Wno-error=varargs -Wno-format-truncation -Wno-error=address-of-packed-member -std=c99 -fopenmp -Wl,-O1 -Wl,--as-needed -Wl,--defsym=__gentoo_check_ldflags__=0 bin/CMakeFiles/darktable.dir/main.c.o -o bin/darktable  -Wl,-rpath,/var/tmp/portage/media-gfx/darktable-3.6.0/work/darktable-3.6.0_build/bin:/usr/lib64/OpenEXR-3:/usr/lib64/Imath-3:  bin/libdarktable.so  -lpthread  -lm  /usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/libgomp.so  /usr/lib64/libpthread.so  /usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/libgomp.so  /usr/lib64/libpthread.so  /usr/lib64/OpenEXR-3/libOpenEXR-3_0.so.29.0.0  /usr/lib64/OpenEXR-3/libIlmThread-3_0.so.29.0.0  /usr/lib64/OpenEXR-3/libIex-3_0.so.29.0.0  /usr/lib64/Imath-3/libImath-3_0.so.28.0.2  /usr/lib64/libz.so  /usr/lib64/libglib-2.0.so  -lgtk-3  -lgdk-3  -lpangocairo-1.0  -lpango-1.0  -lharfbuzz  -latk-1.0  -lcairo-gobject  -lcairo  -lgdk_pixbuf-2.0  -lgio-2.0  -lgobject-2.0  -lglib-2.0  /usr/lib64/libxml2.so  /usr/lib64/libsoup-2.4.so  /usr/lib64/libgio-2.0.so  /usr/lib64/libgobject-2.0.so  /usr/lib64/liblensfun.so  /usr/lib64/libsqlite3.so  /usr/lib64/libglib-2.0.so  -lgtk-3  -lgdk-3  -lpangocairo-1.0  -lpango-1.0  -lharfbuzz  -latk-1.0  -lcairo-gobject  -lcairo  -lgdk_pixbuf-2.0  -lgio-2.0  -lgobject-2.0  -lglib-2.0  /usr/lib64/libxml2.so  /usr/lib64/libsoup-2.4.so  /usr/lib64/libgio-2.0.so  /usr/lib64/libgobject-2.0.so  /usr/lib64/liblensfun.so  /usr/lib64/libsqlite3.so  /usr/lib64/libgthread-2.0.so  /usr/lib64/libgmodule-2.0.so  /usr/lib64/libpangocairo-1.0.so  /usr/lib64/librsvg-2.so  /usr/lib64/libcurl.so  /usr/lib64/libpng.so  -lz  /usr/lib64/libjpeg.so  /usr/lib64/libtiff.so  /usr/lib64/liblcms2.so  /usr/lib64/libjson-glib-1.0.so  /usr/lib64/libexiv2.so  /usr/lib64/libicui18n.so  /usr/lib64/libicuuc.so  /usr/lib64/libicudata.so  /usr/lib64/libpugixml.so && :
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::Header::dataWindow() const'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::Chromaticities::Chromaticities(Imath_2_5::Vec2<float> const&, Imath_2_5::Vec2<float> const&, Imath_2_5::Vec2<float> const&, Imath_2_5::Vec2<float> const&)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::InputFile::header() const'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::globalThreadCount()'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::hasChromaticities(Imf_2_5::Header const&)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::chromaticities(Imf_2_5::Header const&)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::InputFile::setFrameBuffer(Imf_2_5::FrameBuffer const&)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::TiledInputFile::numXTiles(int) const'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::Header::channels() const'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::ChannelList::end() const'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::isOpenExrFile(char const*, bool&)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::TiledInputFile::TiledInputFile(char const*, int)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::TiledInputFile::header() const'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::setGlobalThreadCount(int)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::hasWhiteLuminance(Imf_2_5::Header const&)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::Header::displayWindow() const'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `typeinfo for Imf_2_5::Attribute'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::InputFile::InputFile(char const*, int)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::FrameBuffer::insert(char const*, Imf_2_5::Slice const&)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::RGBtoXYZ(Imf_2_5::Chromaticities const&, float)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::TiledInputFile::readTiles(int, int, int, int, int)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::InputFile::readPixels(int, int)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::Slice::Slice(Imf_2_5::PixelType, char*, unsigned long, unsigned long, int, int, double, bool, bool)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::TiledInputFile::numYTiles(int) const'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::TiledInputFile::setFrameBuffer(Imf_2_5::FrameBuffer const&)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::whiteLuminance(Imf_2_5::Header const&)'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: bin/libdarktable.so: undefined reference to `Imf_2_5::ChannelList::begin() const'
```

waebbl:
> It would be rather helpful here were ilmbase/openexr upstream to embrace the idea of side-by-side presence of different major versions. Just look how (relatively) simple it is for Gentoo to maintain Gtk+ slots, for example! Have you by any chance tried upstreaming the idea yet?

@mkszuba https://lists.aswf.io/g/openexr-dev/message/4946

PR #20933, opened 2021-05-22
cmake-utils--

a17r:
In the year 2021 we don't use cmake-utils anymore.


ZeroChaos-:
ACK, description could use some work, this is just a cleanup of old ebuilds, but yeah, ACK, thanks

a17r:
> ACK, description could use some work, this is just a cleanup of old ebuilds, but yeah, ACK, thanks

How would you improve it?

ZeroChaos-:
"cleanup old ebuilds".  No changes were made, saying things like
"cmake-utils shouldn't be used in 2021" makes me think that you were
actually modifying something, rather than just dropping old ebuilds.

a17r:
But that's just on github.

PR #20937, opened 2021-05-22
eclass/common-lisp-3: Properly escape recursive install files

mjsir911:
This is a bit of a roundabout way to do it, but since we want to pass the results of `find` to a bash function, this is what we need to do.

All of these functions are a bit convoluted, I believe something similar could be hacked together that looks like `find -type f -name '*.lisp' -exec doins` (or xargs) without the recursive calling or escaping mess

Signed-off-by: Marco Sirabella <marco@sirabella.org>

### DIFF
--- a/eclass/common-lisp-3.eclass
+++ b/eclass/common-lisp-3.eclass
@@ -120,7 +120,9 @@ common-lisp-install-sources() {
 		if [[ -f ${path} ]] ; then
 			common-lisp-install-one-source ${fpredicate} "${path}" "$(dirname "${path}")"
 		elif [[ -d ${path} ]] ; then
-			common-lisp-install-sources -t ${ftype} $(find "${path}" -type f)
+			local files
+			readarray -d '' files < <(find "${path}" -type f -print0 || die "cannot traverse ${path}" )
+			common-lisp-install-sources -t ${ftype} "${files[@]}" || die
 		else
 			die "${path} is neither a regular file nor a directory"
 		fi


mjsir911:
cc @ulm @thesamesam

eclass/common-lisp-3.eclass
```diff
@@ -120,7 +120,8 @@ common-lisp-install-sources() {
 		if [[ -f ${path} ]] ; then
 			common-lisp-install-one-source ${fpredicate} "${path}" "$(dirname "${path}")"
 		elif [[ -d ${path} ]] ; then
-			common-lisp-install-sources -t ${ftype} $(find "${path}" -type f)
+			readarray -d '' FILES < <(find "${path}" -type f -print0)
```
ulm, line 123: `FILES` is a local variable AFAICS? Then is should be declared as one, and the convention is to use lowercase for its name.

`|| die` is missing.

mjsir911: Where would the `die` go? `readarray` won't fail (I don't think), would it go inside the `<(...)`? Previous subshelling of find didn't `die`, already checks that path is dir but not readable, so could fail there.

Next line with `common-lisp-install-sources` also `die`s properly in recursion, so shouldn't be needed there.

ulm:
>>>>> On Fri, 17 Sep 2021, Marco Sirabella wrote:

> @mjsir911 commented on this pull request.
>> @@ -120,7 +120,8 @@ common-lisp-install-sources() {
>  		if [[ -f ${path} ]] ; then
>  			common-lisp-install-one-source ${fpredicate} "${path}" "$(dirname "${path}")"
>  		elif [[ -d ${path} ]] ; then
> -			common-lisp-install-sources -t ${ftype} $(find "${path}" -type f)
> +			readarray -d '' FILES < <(find "${path}" -type f -print0)

> Where would the `die` go? `readarray` won't fail (I don't think),
> would it go inside the `<(...)`?

Yes, inside the parentheses.

> Previous subshelling of find didn't `die`, already checks that path is
> dir but not readable, so could fail there.

> Next line with `common-lisp-install-sources` also `die`s properly in
> recursion, so shouldn't be needed there.

It's preferable to die where the error occurs. It makes debugging
easier.

Ulrich

mjsir911:
@ulm ping on this one ;-)

ulm:
This hasn't been posted for review to gentoo-dev yet, I suppose?

ulm:
https://archives.gentoo.org/gentoo-dev/message/811364e2b426c21fab2f26e33f1b9705

ulm:
Merged, thank you.

PR #20938, opened 2021-05-22
9999cleanup

a17r:
@aballier


Alessandro-Barbieri:
What do you think of https://bugs.gentoo.org/713802?

a17r:
> What do you think of https://bugs.gentoo.org/713802?

This is my go-to reply if <random person> wants to add a live ebuild: https://github.com/gentoo/gentoo/pull/20921#issuecomment-846521842
I don't think we need to escalate it to a policy though.

chutz (approved):

monsieurp (approved):

PR #20942, opened 2021-05-23
Net irc/inspricd 3.10.0

clinew:
Needs a bit of work on the config file, but I'm trying to get this out there sooner rather than later.


thesamesam:
Thank you!

PR #20936, opened 2021-05-22
dev-python/apsw: add 3.35.4_p1, bump to python 3.10

arthurzam:
This package needed some more work, so I put it into a different PR.

- Bumped to newest version (which fixes the tests that failed in the bug report)
- Fixed the direct call to `gcc` in `setup.py` - done some testing (using custom setting of CC env to false in the steps) - I think it works.
- Bumped to `python3.10` (my original goal)

My changes are:

```diff
--- apsw-3.32.2_p1.ebuild	2021-03-24 14:22:29.343546908 +0200
+++ apsw-3.35.4_p1.ebuild	2021-05-22 21:17:26.690629301 +0300
@@ -1,10 +1,11 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
-PYTHON_COMPAT=( python3_{7..9} )
 
-inherit distutils-r1 flag-o-matic
+PYTHON_COMPAT=( python3_{7..10} )
+
+inherit distutils-r1 toolchain-funcs
 
 MY_PV=${PV/_p/-r}
 MY_P=${PN}-${MY_PV}
@@ -12,27 +13,31 @@
 DESCRIPTION="APSW - Another Python SQLite Wrapper"
 HOMEPAGE="https://github.com/rogerbinns/apsw/"
 SRC_URI="https://github.com/rogerbinns/apsw/releases/download/${MY_PV}/${MY_P}.zip -> ${P}.zip"
+S="${WORKDIR}/${MY_P}"
 
 LICENSE="ZLIB"
 SLOT="0"
-KEYWORDS="amd64 ~arm ~ppc64 x86"
+KEYWORDS="~amd64 ~arm ~ppc64 ~x86"
 IUSE="doc"
 
 RDEPEND=">=dev-db/sqlite-${PV%_p*}"
-DEPEND="${RDEPEND}
-	app-arch/unzip"
-
-S=${WORKDIR}/${MY_P}
+DEPEND="${RDEPEND}"
+BDEPEND="app-arch/unzip"
 
 PATCHES=( "${FILESDIR}/${PN}-3.6.20.1-fix_tests.patch" )
 
+python_prepare_all() {
+	sed -e "s/gcc/$(tc-getCC)/" -i setup.py || die
+	distutils-r1_python_prepare_all
+}
+
 python_compile() {
 	distutils-r1_python_compile --enable=load_extension
 }
 
[...]

PR #20943, opened 2021-05-23

arthurzam:
- Those were simple bump and test with `ebuild ... test`
- for `dev-python/colorclass`, imported patch from a pull request upstream
- for `dev-python/tabulate` I changed the way we turn off the tests from a patch to deselect - I think it is cleaner

Small question:
Should we add `"` around `${WORKDIR}`?
As I think this variable might have spaces (user set), so we must add quotes, but I understood it isn't QA policy so isn't enforced.


dev-python/et_xmlfile/et_xmlfile-1.1.0.ebuild
```diff
@@ -3,18 +3,18 @@
...
 MY_PV=${PV%.0}
 MY_P=${PN}-${MY_PV}
 
 DESCRIPTION="An implementation of lxml.xmlfile for the standard library"
-HOMEPAGE="https://pypi.org/project/et-xmlfile/"
+HOMEPAGE="https://pypi.org/project/et-xmlfile/ https://foss.heptapod.net/openpyxl/et_xmlfile"
 SRC_URI="
 	https://foss.heptapod.net/openpyxl/et_xmlfile/-/archive/${MY_PV}/${MY_P}.tar.gz"
-S=${WORKDIR}/${MY_P}
+S="${WORKDIR}/${MY_P}"
```
mgorny, line 17: Normally we don't really quote `S` since it's not supposed to contain spaces. I'll revert this line.

dev-python/et_xmlfile/et_xmlfile-1.1.0.ebuild
```diff
@@ -3,18 +3,18 @@
...
+PYTHON_COMPAT=( python3_{7..10} )
 
 inherit distutils-r1
 
 MY_PV=${PV%.0}
 MY_P=${PN}-${MY_PV}
 
 DESCRIPTION="An implementation of lxml.xmlfile for the standard library"
-HOMEPAGE="https://pypi.org/project/et-xmlfile/"
+HOMEPAGE="https://pypi.org/project/et-xmlfile/ https://foss.heptapod.net/openpyxl/et_xmlfile"
```
mgorny, line 14: Please wrap at 72 or 80 chars.

mgorny (approved):
Thanks for doing this. I've noted two things and I'll fix them while merging.

I'll let you in on a secret. If you set

    DISTUTILS_STRICT_ENTRY_POINTS=1

in make.conf, the QA checks will tell you when to update DISTUTILS_USE_SETUPTOOLS for py3.8+.

Also, if you're interested, I'd use some help porting stuff to py3.9 before we switch the default next week. However, you can skip openstack and robotos since both are bad quality and will probably be removed.

dev-python/et_xmlfile/et_xmlfile-1.1.0.ebuild
arthurzam: Noted, thank you.

dev-python/et_xmlfile/et_xmlfile-1.1.0.ebuild
arthurzam: Noted, thank you.

arthurzam:
Ok, I will move to porting 3.9 - I will gladly help.

Also, could you please look at https://github.com/gentoo/gentoo/pull/20936, so we can close previous PR for py 3.10?

mgorny:
I'm currently looking through packages inheriting distutils-r1, so preferably choose another subset ;-).

PR #20948, opened 2021-05-23
sys-fs/ntfs2btrfs: Initial import

maharmstone:
This is a new ebuild for my NTFS to Btrfs filesystem conversion tool, which is already in Arch and Fedora. See also https://bugs.gentoo.org/715532.


maharmstone:
Updated to fix CI failures.

PR #20956, opened 2021-05-23
*/*: bump python version

arthurzam:
@mgorny First batch of bumps. Where I can, I also add `py3.10` support on the way.

1. `dev-python/pdfrw` passes fully thanks to tests
2. `media-gfx/img2pdf` builds fully, and tested various conversations (works properly). The test suite still doesn't work (and shows the same errors as with previous python version - so I can conclude no change in tests state).
3. `app-arch/unrpa` builds and tested some rpa file found online - I guess it should work (no problematic code found in source)
4. `sys-apps/razercfg` builds but unable to test (no razer hardware). But I viewed the python code and haven't found something that shouldn't work.
5. `dev-util/apitrace` builds and works. Fixed the build so it really depends on passed python version.


sys-apps/razercfg/razercfg-0.42.ebuild
```diff
@@ -77,8 +77,8 @@ src_install() {
 
 pkg_postinst() {
 	if use udev ; then
-		udevadm control --reload-rules
-		udevadm trigger --subsystem-match=usb
+		udevadm control --reload-rules || die
```
mgorny, line 80: Not sure if dying in middle of postinst is a good idea.

arthurzam: I thought every call must "die" somehow - if not needed will remove.

mgorny: postinst is special. It's run *after* the package is installed, so it can't prevent it from being installed. Here's better to either ignore errors if they don't affect further code, or use things like `&&` to connect things that are interdependent. In either case, it's not for you to worry right now ;-).

PR #20964, opened 2021-05-24
*/*: bump python version

arthurzam:
@mgorny Second batch of bumps. Where I can, I also add `py3.10` support on the way.

1. `dev-util/gcovr` I have enabled partial tests. It still fails for a big part (because upstream are based on very old gcc version), but I guess a little tests is better that none. Tested by merging and running.
2. `sys-kernel/bliss-initramfs` did a small cleanup. Reviewed the code and shouldn't have any problem with newer python versions.
3. `media-libs/libiptcdata` tested the python bindings using the [examples](https://github.com/ianw/libiptcdata/tree/master/python/examples).
4. `app-editor/{,g}vim` merged and tested
5. `app-vim/*` reviewed the python code, should work without problems. There are a little more packages in this category needing python bump, but the batch was already quite big.

### DIFF
--- a/app-editors/gvim/gvim-8.2.0814-r100.ebuild
+++ b/app-editors/gvim/gvim-8.2.0814-r100.ebuild
@@ -4,7 +4,7 @@
 EAPI=7
 VIM_VERSION="8.2"
 LUA_COMPAT=( lua5-1 luajit )
-PYTHON_COMPAT=( python3_{7,8} )
+PYTHON_COMPAT=( python3_{7..10} )
 PYTHON_REQ_USE="threads(+)"
 USE_RUBY="ruby24 ruby25 ruby26 ruby27"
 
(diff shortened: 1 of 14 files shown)


chutz:
If the `src_test` for `dev-util/gcovr` does not pass as is, please set `RESTRICT=test`. Disabling tests is better than having known failing tests in the tree. I am fine also disabling all the tests that fail, and leaving the working ones enabled.

arthurzam:
> If the `src_test` for `dev-util/gcovr` does not pass as is, please set `RESTRICT=test`. Disabling tests is better than having known failing tests in the tree. I am fine also disabling all the tests that fail, and leaving the working ones enabled.

Thanks for the feedback, returned the RESTRICT. Sorry for changing it.

mgorny:
I'm sorry, I've forgotten that I had the old version ready to push. Could you rebase / redo the gcovr change?

arthurzam:
@mgorny No problem :-)
Here is the change (a simple change)
https://github.com/arthurzam/gentoo/commit/bf52beabd1df39d470ffb84abb394ca11a1d44c9

PR #20965, opened 2021-05-24
app-text/zathura-cb: add jpeg USE flag

N-R-K:
solves this: https://git.pwmt.org/pwmt/zathura/-/issues/213


juippis (commented):
Sounds like this either needs to use `optfeature` instead, or you need to make it a configurable option during `src_configure` and `src_compile`.
https://devmanual.gentoo.org/eclass-reference/optfeature.eclass/index.html

N-R-K:
Sorry for leaving this hanging, but after reading about GCO I'm not interested in this anymore, as I'm not comfortable giving away my legal name due to privacy reasons. Closing.

PR #20976, opened 2021-05-25
app-backup/dar: drop 2.6.13

ceamac:
Drop old versions.
app-backup/dar-2.6.14 is stable, also dev-libs/libthreadar-1.3.2


thesamesam:
Fixed second commit message to use Bug: reference, thanks!

ceamac:
I tried to avoid having 2 Bug: references to the same bug in the same PR, to avoid spamming the bug database.  Good to know there is no problem with this.  Thanks!

PR #20981, opened 2021-05-25
dev-libs/libxml2: Fix regression introduced by lxml patch

zippy2:
Turns out that the lxml patch introduced a regression. So far
I've identified as libvirt being the only one affected, but
that's because it's the only package I checked.

Bug: https://bugs.gentoo.org/790737


zippy2:
@thesamesam I've noticed you switched to a compressed patchset. Feel free to just include the patch there and close this one without merging.

thesamesam:
@zippy2 Huge thanks for picking this up. I've been checking regularly because of all the disruption but hadn't caught this yet. I'll add it to the patchset now but I'll --author.

thesamesam:
Done in 5dd398f1e5a3acda39f669bd9d94a6f9d715ac59. Thank you again and please let me know if there’s anything else!

PR #20986, opened 2021-05-25
Add and use meson-multilib.eclass

mattst88:
There are a lot of private (i.e. to individual ebuilds and/or packages) functions implementing the necessary `meson_use`/`meson_feature`-style functions for multilib.

This PR factors those out and shows how we can use them in existing packages.

Consider this an RFC. I'm open to all feedback and I don't feel strongly about anything I've written here.

A couple of unorganized thoughts:

- I feel like the function names are too long. Should we remove `multilib_` from the names?

- Trying to make the function names match `meson_use`/`meson_feature` exposes the fact that the `meson_use` name is weird. We want functions for the different meson options:
    - `feature` (enabled/disabled)
    - `bool` (true/false)
    - maybe something for yes/no (a common `combo` configuration).

Within those, we want multilib functions that operate with and without a USE flag. One natural way to name those functions differently is to put `_use` in the name of the ones that take a USE flag, e.g., could be something like `meson_multilib_native_use_enable` vs `meson_multilib_native_enable`. But if we already are using `_use` in a name like `meson_multilib_native_use` then adding another `_use` is kind of nonsensical. It's for that reason that I instead have `meson_multilib_native_use` and `meson_multilib_native_true`. Perhaps we should consider deprecating `meson_use` and replacing it with a `meson_bool`? :shrug:

### DIFF
--- a/app-crypt/libsecret/libsecret-0.20.4.ebuild
+++ b/app-crypt/libsecret/libsecret-0.20.4.ebuild
@@ -5,7 +5,7 @@ EAPI=7
 PYTHON_COMPAT=( python3_{7..9} )
 VALA_USE_DEPEND=vapigen
 
-inherit gnome2 meson multilib-minimal python-any-r1 vala virtualx
+inherit gnome2 meson-multilib python-any-r1 vala virtualx
 
 DESCRIPTION="GObject library for accessing the freedesktop.org Secret Service API"
 HOMEPAGE="https://wiki.gnome.org/Projects/Libsecret"
@@ -71,32 +71,20 @@ src_prepare() {
 		-i libsecret/secret-enum-types.h.template || die
 }
 
-meson_multilib_native_use() {
-	multilib_native_usex "$1" "-D${2-$1}=true" "-D${2-$1}=false"
-}
-
 multilib_src_configure() {
 	local emesonargs=(
 		$(meson_use crypt gcrypt)
 
 		# Don't build docs multiple times
-		-Dmanpage=$(multilib_is_native_abi && echo true || echo false)
-		$(meson_multilib_native_use gtk-doc gtk_doc)
+		$(meson_native_true manpage)
+		$(meson_native_use_bool gtk-doc gtk_doc)
 
-		$(meson_multilib_native_use introspection)
-		$(meson_multilib_native_use vala vapi)
+		$(meson_native_use_bool introspection)
+		$(meson_native_use_bool vala vapi)
 	)
 	meson_src_configure
 }
 
-multilib_src_compile() {
-	meson_src_compile
-}
-
 multilib_src_test() {
 	virtx meson_src_test
-}
-
-multilib_src_install() {
-	meson_src_install
 }
--- a/eclass/meson-multilib.eclass
+++ b/eclass/meson-multilib.eclass
@@ -1,0 +1,131 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# @ECLASS: meson-multilib.eclass
+# @MAINTAINER:
+# Author: Matt Turner <mattst88@gentoo.org>
+# @AUTHOR:
+# Author: Michał Górny <mgorny@gentoo.org>
+# Author: Matt Turner <mattst88@gentoo.org>
+# @SUPPORTED_EAPIS: 7
+# @BLURB: meson wrapper for multilib builds
+# @DESCRIPTION:
+# The meson-multilib.eclass provides a glue between meson.eclass(5)
+# and multilib-minimal.eclass(5), aiming to provide a convenient way
+# to build packages using meson for multiple ABIs.
+#
+# Inheriting this eclass sets IUSE and exports default multilib_src_*()
+# sub-phases that call meson phase functions for each ABI enabled.
+# The multilib_src_*() functions can be defined in ebuild just like
+# in multilib-minimal, yet they ought to call appropriate meson
+# phase rather than 'default'.
+
+if [[ -z ${_MESON_MULTILIB_ECLASS} ]] ; then
+_MESON_MULTILIB_ECLASS=1
+
+case ${EAPI:-0} in
+	7) ;;
+	*) die "EAPI=${EAPI} is not supported" ;;
+esac
+
+inherit meson multilib-minimal
+
+EXPORT_FUNCTIONS src_configure src_compile src_test src_install
+
+# @FUNCTION: meson_native_use_bool
+# @USAGE: <USE flag> [option name]
+# @DESCRIPTION:
+# Given a USE flag and a meson project option, output a string like:
+#
+#   -Doption=true
+#   -Doption=false
+#
+# if building for the native ABI (multilib_is_native_abi is true). Otherwise,
+# output -Doption=false. If the project option is unspecified, it defaults
+# to the USE flag.
+meson_native_use_bool() {
+	multilib_native_usex "${1}" "-D${2-${1}}=true" "-D${2-${1}}=false"
+}
+
+# @FUNCTION: meson_native_use_feature
+# @USAGE: <USE flag> [option name]
+# @DESCRIPTION:
+# Given a USE flag and a meson project option, output a string like:
+#
+#   -Doption=enabled
+#   -Doption=disabled
+#
+# if building for the native ABI (multilib_is_native_abi is true). Otherwise,
+# output -Doption=disabled. If the project option is unspecified, it defaults
+# to the USE flag.
+meson_native_use_feature() {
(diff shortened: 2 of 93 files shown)


floppym:
> I feel like the function names are too long. Should we remove multilib_ from the names?

Yeah, I think "native" already conveys the multilib-nature of the functions.

> Trying to make the function names match meson_use/meson_feature exposes the fact that the meson_use name is weird.

I like the idea of including the output type in the function name.

mattst88:
Removed `_multilib`  from the names and changed `meson_native_use` -> `meson_native_bool` to better indicate the return type.

Also added a patch to `meson.eclass` to cd into `${S}` around `einstalldocs` like cmake.eclass does.

Still need eclass documentation for `meson-multilib.eclass`.

mattst88:
meson-multilib.eclass documentation completed. I've sent the two eclass patches at the beginning of the series to gentoo-dev@ for review.

eclass/meson-multilib.eclass
```diff
@@ -0,0 +1,126 @@
...
+
+case ${EAPI:-0} in
+	7) ;;
+	*) die "EAPI=${EAPI} is not supported" ;;
+esac
+
+inherit meson multilib-minimal
+
+EXPORT_FUNCTIONS src_configure src_compile src_test src_install
+
```
SoapGentoo, line 31: missing inherit guard

media-libs/harfbuzz/harfbuzz-9999.ebuild
```diff
@@ -100,14 +92,6 @@ multilib_src_configure() {
...
 
-multilib_src_compile() {
-	meson_src_compile
-}
-
-multilib_src_install() {
-	meson_src_install
-}
-
 multilib_src_install_all() {
```
SoapGentoo, line 95: is this one necessary?

media-libs/mesa/mesa-20.3.5.ebuild
```diff
@@ -526,14 +526,6 @@ multilib_src_configure() {
...
 
-multilib_src_compile() {
-	meson_src_compile
-}
-
-multilib_src_install() {
-	meson_src_install
-}
-
 multilib_src_install_all() {
```
SoapGentoo, line 529: necessary?

media-libs/mesa/mesa-21.0.3.ebuild
```diff
@@ -524,14 +524,6 @@ multilib_src_configure() {
...
 
-multilib_src_compile() {
-	meson_src_compile
-}
-
-multilib_src_install() {
-	meson_src_install
-}
-
 multilib_src_install_all() {
```
SoapGentoo, line 527: necessary?

media-libs/mesa/mesa-21.1.1.ebuild
```diff
@@ -525,14 +525,6 @@ multilib_src_configure() {
...
 
-multilib_src_compile() {
-	meson_src_compile
-}
-
-multilib_src_install() {
-	meson_src_install
-}
-
 multilib_src_install_all() {
```
SoapGentoo, line 528: necessary?

media-libs/mesa/mesa-9999.ebuild
```diff
@@ -524,14 +524,6 @@ multilib_src_configure() {
...
 
-multilib_src_compile() {
-	meson_src_compile
-}
-
-multilib_src_install() {
-	meson_src_install
-}
-
 multilib_src_install_all() {
```
SoapGentoo, line 527: necessary?

x11-libs/libxkbcommon/libxkbcommon-1.3.0.ebuild
```diff
@@ -55,15 +55,3 @@ multilib_src_configure() {
 	)
 	meson_src_configure
 }
-
-multilib_src_compile() {
-	meson_src_compile
-}
-
-multilib_src_test() {
-	virtx meson_src_test
```
SoapGentoo, line 64: sure?

SoapGentoo (changes requested):

eclass/meson-multilib.eclass
mattst88: Is it needed? `cmake-multilib.eclass` doesn't have it, and both `meson.eclass` and `multilib-minimal.eclass` have it?

Doesn't hurt to have it I guess.

media-libs/harfbuzz/harfbuzz-9999.ebuild
mattst88: Nope. Thanks!

media-libs/mesa/mesa-20.3.5.ebuild
mattst88: Nope!

media-libs/mesa/mesa-21.0.3.ebuild
mattst88: Nope!

media-libs/mesa/mesa-21.1.1.ebuild
mattst88: Nope!

media-libs/mesa/mesa-9999.ebuild
mattst88: Nope!

x11-libs/libxkbcommon/libxkbcommon-1.3.0.ebuild
mattst88: Oops. I didn't catch this because the only one of the tests needs an Xserver and it nicely skips if it doesn't have one.

mattst88:
Found and replaced a few more `-Doption=$(usex ... true false)` and removed some more unnecessary einstalldocs.

Planning to push this tomorrow if there are no further comments.

eclass/meson-multilib.eclass
```diff
@@ -0,0 +1,131 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# @ECLASS: meson-multilib.eclass
+# @MAINTAINER:
+# gx86-multilib team <multilib@gentoo.org>
```
mgorny, line 6: ENOSUCHTEAM.

eclass/meson-multilib.eclass
```diff
@@ -0,0 +1,131 @@
...
+# @DESCRIPTION:
+# Given a USE flag and a meson project option, output a string like:
+#
+#   -Doption=true
+#   -Doption=false
+#
+# if building for the native ABI (multilib_is_native_abi is true). Otherwise,
+# output -Doption=false. If the project option is unspecified, it defaults
+# to the USE flag.
+meson_native_bool() {
```
mgorny, line 46: Personally, I find the naming to be confusing but I get your point. Still, I'd prefer if stuff using USE flags had `use` in its name.

app-crypt/libsecret/libsecret-0.20.4.ebuild
```diff
@@ -71,16 +71,12 @@ src_prepare() {
...
-}
-
 multilib_src_configure() {
 	local emesonargs=(
 		$(meson_use crypt gcrypt)
 
 		# Don't build docs multiple times
-		-Dmanpage=$(multilib_is_native_abi && echo true || echo false)
+		$(meson_native_true manpage)
 		$(meson_multilib_native_use gtk-doc gtk_doc)
```
mgorny, line 80: But you've renamed the function, right?

mattst88: Damn. Not sure how that slipped through. Fixed.

eclass/meson-multilib.eclass
mattst88: You're so sneaky!

eclass/meson-multilib.eclass
mattst88: What should the names be then? How about:

`meson.eclass`:
* `meson_use` → `meson_use_bool`
* `meson_feature` → `meson_use_feature`

`meson-multilib.eclass`:
* `meson_native_bool` → `meson_native_use_bool`
* `meson_native_feature` → `meson_native_use_feature`
* `meson_native_true`
* `meson_native_enabled`

Seems to fit with `use_enable` and `use_with`.

mgorny: I would go a step further and use either `bool` or `true` consistently, e.g. `meson_native_use_bool` and `meson_native_bool` etc.

mattst88: Discussed on IRC and mgorny said

> <mgorny> mattst88: at a second look, i'd probably use _true and _enabled

sam and floppym acked.

thesamesam (approved):

PR #20988, opened 2021-05-25
dev-java/antlr4-runtime: new package

vaukai:
Bug: https://bugs.gentoo.org/742554

pkgdiff to jar file from https://mvnrepository.com/artifact/org.antlr/antlr4-runtime/4.9.2 seems ok


fordfrog:
@vaukai why you closed this?

vaukai:
> @vaukai why you closed this?

It was about slot 4 but first we need slot 3.5. Also [tests](https://github.com/antlr/antlr4/tree/master/runtime-testsuite) were missing.

fordfrog:
ah, ok, thanks :-) i think we need all the three packages (as described in the bug) so that we do not have two packages with the same libraries.

PR #20999, opened 2021-05-26
dev-java/jzlib: remove obsolete version

vaukai:
Closes: https://bugs.gentoo.org/786519


fordfrog:
@vaukai thanks! i checked `dev-java/jsch` whether it builds against the newer version and it does so instead i updated the newer version, moved it to slot 0 and dropped the old one.

PR #21010, opened 2021-05-27
media-libs/netpbm: v10.86.22

ceamac:
Minor release
Backported a fix from netpbm-10.94.3 so it passes all tests


thesamesam:
Thanks!

PR #21020, opened 2021-05-27
remove some openstack stuff

prometheanfire:
there are likely some oslo packages that can be cleaned up as well

Signed-off-by: Matthew Thode <prometheanfire@gentoo.org>


mgorny:
Thanks for doing this. Let's pmask them and then look for more that can go.

Do you want treecleaners to do it or will you do it?

prometheanfire:
I made a commit already, but the message was too snarky for soap.  I can make a reminder to myself to remove in a month what I just p.masked

PR #21024, opened 2021-05-28
dev-python/*sphinx*: bump to python 3.9

arthurzam:
@mgorny Fifth batch. Here is a little more specialized python bump. I tried here to bump all sphinx related ebuilds to python 3.9. All were tested using various examples I found, while for most of them `dev-python/aiohttp[doc]` was really good test.

1. For `dev-python/sphinxcontrib-doxylink` I moved to `github` archives for using the test suite (and now it has tests and even passes them!)
2. For `dev-python/sphinxcontrib-newsfeed` I fixed QA notice given by `pkgcheck` about `insinto` for newer version
3. For `dev-python/sphinxcontrib-asyncio` I managed to port and fix for `sphinx>=4.0`. Now it successfully builds self docs and `aiohttp` so I think my change is working properly. Sadly, upstream is dead so no reason to push upstream the patch.

As always, I put every change info and explanation into commit description.

**Note:**
I think we should mask and remove `dev-python/sphinxcontrib-httpdomain`. This package doesn't import and has non trivial error since `sphinx>=4.0`, as a result fails current tests, and only revdep is `dev-python/sphinxcontrib-httpexample` which is masked (by you [here](http://bugs.gentoo.org/743889) with same problem). I think we can even reuse the same description for the mask :)


dev-python/sphinxcontrib-asyncio/sphinxcontrib-asyncio-0.2.0-r2.ebuild
```diff
@@ -13,17 +13,18 @@ SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
...
 RDEPEND="dev-python/sphinx[${PYTHON_USEDEP}]
 	dev-python/namespace-sphinxcontrib[${PYTHON_USEDEP}]"
 
-python_compile_all() {
-	use doc && emake -C docs html
+distutils_enable_sphinx docs
+
+src_prepare() {
+	# fix for sphinx >= 4.0
+	sed -i sphinxcontrib/asyncio.py
```
mgorny, line 24: You've missed a `\` here. I'll fix it while merging.

mgorny (approved):
Thanks. Awesome work, as always. I have kernel bumps batched, so it will take a while before I push this.

If you have a minute, PM me on IRC and let's discuss making you a dev.

PR #21026, opened 2021-05-28
dev-libs/weston: Add ability to use seatd

lanodan:


chewi:
You didn't run this through repoman or similar. It needs a description for the new USE flag.

chewi:
Turns out the option is actually called `launcher-libseat` so I fixed that for you.

PR #21029, opened 2021-05-28
x11-wm/muffin: Bump to 5.0.0

rndevfx:
Signed-off-by: Richard Nespithal <richard@nespithal.com>


SparkyBluefang:
Cinnamon 5.0 is in the process of being released, but not all packages have been released yet (cinnamon, nemo, cinnamon-settings-daemon). I'm working on a comprehensive update to all Cinnamon packages (as I have the previous 2 releases), so it probably doesn't make sense to do it piecemeal package-by-package.

rndevfx:
Totally agree but muffin in this case "only" has a few fixes which work fine with cinnamon-4.8.x as well. And there is an interesting patch available https://github.com/linuxmint/muffin/issues/544#issuecomment-841654789

Patch itself: https://gist.github.com/NikitaKarnauhov/3e8d19db835d14e21520ebcc18c28ab3

rndevfx:
Done in https://github.com/gentoo/gentoo/pull/21047

PR #21030, opened 2021-05-28
app-shells/bash: 5.x: set SELinux context for /bin/sh.

jpds:
Portage performs package builds under the portage_tmp_t label, and as
/bin/sh is not a part of the package file set, it's installed with that
temporary label and isn't assigned the correct label by the Portage
SELinux integration (causing system failures later). Adding the -Z option
installs the symlink with the correct label.

### DIFF
--- a/app-shells/bash/bash-5.1_p8.ebuild
+++ b/app-shells/bash/bash-5.1_p8.ebuild
@@ -262,7 +262,7 @@ pkg_preinst() {
 		local target=$(readlink "${EROOT}"/bin/sh)
 		local tmp="${T}"/sh
 		ln -sf "${target}" "${tmp}"
-		mv -f "${tmp}" "${EROOT}"/bin/sh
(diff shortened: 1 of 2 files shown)


app-shells/bash/bash-5.1_p8.ebuild
```diff
@@ -262,7 +262,7 @@ pkg_preinst() {
 		local target=$(readlink "${EROOT}"/bin/sh)
 		local tmp="${T}"/sh
 		ln -sf "${target}" "${tmp}"
-		mv -f "${tmp}" "${EROOT}"/bin/sh
+		mv -fZ "${tmp}" "${EROOT}"/bin/sh
```
thesamesam, line 265: Can you check for userland_gnu and only do -Z in that case?

floppym:
Related bug: https://bugs.gentoo.org/778311

floppym:
I removed the /bin/sh logic from pkg_preinst.

PR #21033, opened 2021-05-29
app-misc/recoll: Version bump to 1.31.0

ppurka:
Signed-off-by: Punarbasu Purkayastha ppurka@gmail.com

* No corresponding Gentoo bug report is present.
* No changes from previous (latest) ebuild.


app-misc/recoll/recoll-1.31.0.ebuild
```diff
@@ -0,0 +1,131 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{6..9} )
```
juippis, line 6: Could you update this? The current active python supports are 3.9 and 3.10. Nothing bad having 3.7-3.8 in there either, but 3.6 is a bit...

app-misc/recoll/recoll-1.31.0.ebuild
```diff
@@ -0,0 +1,131 @@
...
+
+DESCRIPTION="Personal full text search package"
+HOMEPAGE="https://www.lesbonscomptes.com/recoll/"
+SRC_URI="https://www.lesbonscomptes.com/recoll/${P}.tar.gz"
+
+LICENSE="GPL-2+"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+IUSE="camelcase chm doc +inotify qt5 session +spell webengine"
```
juippis, line 18: I don't think we need `doc` here. Can you check how much worth diskspace the html docs are when installed? If only a few KBs, just install them unconditionally.

app-misc/recoll/recoll-1.31.0.ebuild
```diff
@@ -0,0 +1,131 @@
...
+	optfeature "7zip archives support"          dev-python/pylzma
+	optfeature "iCalendar files support"        dev-python/icalendar
+	optfeature "Postscript files support"       app-text/pstotext
+	optfeature "RTF files support"              app-text/unrtf
+	optfeature "TeX files support"              dev-text/detex
+	optfeature "DVI files support"              virtual/tex-base
+	optfeature "DJVU files support"             app-text/djvu
+	optfeature "tags in audio files support"    media-libs/mutagen
+	optfeature "tags in image files support"    media-libs/exiftool
+	optfeature "Midi karaoke files support"     dev-python/chardet
```
juippis, line 130: Also let's fix this, don't attempt to align.
https://projects.gentoo.org/qa/policy-guide/ebuild-format.html#pg0101

app-misc/recoll/recoll-1.31.0.ebuild
ionenwks: 3.7 is the same as 3.6 at this point, it's not a valid target, i.e. can't select it, it's just dead code -- removing 7 whenever possible will save mgorny some work later when cleaning these up in a `*/*` commit

So check if 10 works, if it does then `3_{8..10}`, if it doesn't, then `3_{8..9}`

ppurka:
Updated the ebuild as instructed. I have not added Python 3.10 since it is still in beta.

ionenwks:
> Updated the ebuild as instructed. I have not added Python 3.10 since it is still in beta.

***If*** it works there's no reason not to add it now, running behind will just be grief for users later on.

ppurka:
Thanks! Added Python 3.10. It works. The setuptools needs to be installed - recoll uses PYTHON_SINGLE_TARGET.

juippis (approved):

thesamesam:
Merged in https://gitweb.gentoo.org/repo/gentoo.git/commit/app-misc/recoll?id=55945ef3fbc09cd2e71d1ac117f49e462936b2be.

PR #21036, opened 2021-05-29
media-gfx/waifu2x-ncnn-vulkan: add 20210521 and ncnn-20210525

ionenwks:
Added myself as co-maint, hopefully get to update this to at g.o soon'ish then I'll mostly take care of it directly.

Gave ncnn a bit more work, seems some people do use these tools. Still a few things that could be done (i.e. python, and hack in static lib build for tests), but I'll leave alone *for now*.

For segfault issue, mostly hoping it won't come back given it'd need a SLOT="0/${PV}" on glslang or maybe vulkan-loader. I don't recall this being an issue before, maybe just a fluke.


ionenwks:
Added a workaround for https://bugs.gentoo.org/730468

Don't particularly like it but better than having it fail for everyone with avx without avx2

media-gfx/waifu2x-ncnn-vulkan/waifu2x-ncnn-vulkan-20210521.ebuild
```diff
@@ -0,0 +1,53 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake
+
+NCNN_VERSION="20210525" # pin near bundled copy until more stable
```
slashbeast, line 8: You use it only once and I rather not have global variables defined here. Can you please just specify it in DEPEND?

PR #21039, opened 2021-05-29
sci-geosciences/grass: fix build failure with GCC 11

hamarituc:
Closes: https://bugs.gentoo.org/792801


thesamesam:
Thanks! No need for a [revbump](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html) because it's a build-time fix (no need to make people rebuild if they already built it successfully), so I'll undo that if that's alright.

PR #21043, opened 2021-05-29
app-admin/logstash-bin: drop vulnerable

ferki:
This PR is an attempt to fix [bug 790287](https://bugs.gentoo.org/790287) by dropping a vulnerable app-admin/logstash-bin version.

Please review and merge, or let me know how to improve it further!


thesamesam:
Looks good, thanks, although avoid using a Closes: tag for security bugs (just use Bug:). Thanks for doing this!

ferki:
> Looks good, thanks, although avoid using a Closes: tag for security bugs (just use Bug:). 

@thesamesam: thanks for the hint, will follow this in the future!

Could you point me a good document about this convention, please? I've checked GLEP 66 for the correct tag, and that didn't mention such rule for security bugs. I'd like to contribute a doc fix if there's a good place for this bit of information :)

> Thanks for doing this!

Thanks for merging! :pray:

ionenwks:
> > Looks good, thanks, although avoid using a Closes: tag for security bugs (just use Bug:).
>
> Could you point me a good document about this convention, please? I've checked GLEP 66 for the correct tag, and that didn't mention such rule for security bugs. I'd like to contribute a doc fix if there's a good place for this bit of information :)

It's indicated in the bug itself, on security bugs you'll see (under the comment box) standing-out text saying "please don't mark this bug as resolved" (aka close it) and that the security team will take care of it.

ferki:
> It's indicated in the bug itself, on security bugs you'll see (under the comment box) standing-out text saying "please don't mark this bug as resolved" (aka close it) and that the security team will take care of it.

@ionenwks: Thanks for the clarification! That note seems to be added via CSS to the bug's page, and apparently it's only visible in graphical browsers, but not in the plaintext versions (bugzilla emails, pybugz output, or terminal browsers). I'll keep that in mind, and will use the `Bug:` tag in security bugs.

PR #21045, opened 2021-05-29
app-editors/vscodium: new package

AdelKS:
Hello @juippis,

Here is my first draft on the `vscodium` EBUILD. I have been using it for a few weeks and it is working properly, I learned a few things on it recently, for example it does not have by default the same extension app store as `vscode` and that's why I couldn't find a few extensions when I looked for them. More information here: https://github.com/VSCodium/vscodium/blob/master/DOCS.md

The EBUILD is the same as the `vscode` package except for the name, the licenses, and that we can use `S="${WORKDIR}`.

Thanks!


app-editors/vscodium/vscodium-1.56.2.ebuild
```diff
@@ -0,0 +1,94 @@
...
+	/opt/vscodium/swiftshader/libGLESv2.so
+	/opt/vscodium/resources/app/extensions/*
+	/opt/vscodium/resources/app/node_modules.asar.unpacked/*
+"
+
+S="${WORKDIR}"
+
+src_install() {
+	# Cleanup
+	rm -r "${S}/resources/app/LICENSE.txt" || die
```
juippis, line 74: No need for `-r` here.

app-editors/vscodium/vscodium-1.56.2.ebuild
```diff
@@ -0,0 +1,94 @@
...
+	media-libs/libpng:0/16
+	x11-libs/cairo
+	x11-libs/gtk+:3
+	x11-libs/libnotify
+	x11-libs/libxkbcommon
+	x11-libs/libxkbfile
+	x11-libs/libXScrnSaver
+	x11-libs/libXtst
+	x11-libs/pango
+"
```
juippis, line 54:
```
 * 
 * 	/opt/vscodium/codium: libcups.so.2
 * 
```
```
$ vscodium 
/opt/vscodium/bin/../codium: error while loading shared libraries: libcups.so.2: cannot open shared object file: No such file or directory
```
Works fine after installing cups.

AdelKS: Does that mean `vscode` needs it, is there a simple way to get these missing soname dependencies ? Or do I need a clean chroot ?

app-editors/vscodium/vscodium-1.56.2.ebuild
AdelKS: Ah right, thanks!

app-editors/vscodium/vscodium-1.56.2.ebuild
juippis: I'm not sure why I didn't hit this error with *vscode*. Could you please provide the list for: 
*/var/db/pkg/app-editors/vscode-version/NEEDED* and */var/db/pkg/app-editors/vscode-version/NEEDED.ELF.2*

Yes, a clean stage3 chroot is usually the most accurate way of testing your deps. You'll need `FEATURES="qa-unresolved-soname-deps` to catch them.

Then you can compare the above file I gave you with your RDEPEND list after it's installed, strace / scanefl binaries and libraries...
```#!/bin/bash

qlist -e ${1} \
	| xargs --no-run-if-empty scanelf -L -n -q -F '%n #F' \
	| tr , '\n' \
	| xargs --no-run-if-empty readlink -f \
	| uniq \
	| xargs --no-run-if-empty qfile -CS \
	| sort -u
```

AdelKS: > Could you please provide the list for:
/var/db/pkg/app-editors/vscode-version/NEEDED and /var/db/pkg/app-editors/vscode-version/NEEDED.ELF.2

You mean the ones I have on my machine right now ? I attached them.

[vscode-NEEDED.zip](https://github.com/gentoo/gentoo/files/6566293/vscode-NEEDED.zip)

> Yes, a clean stage3 chroot is usually the most accurate way of testing your deps. You'll need FEATURES="qa-unresolved-soname-deps to catch them.

Alright, I will set that up for the next times. I liked the BTRFS snapshot approach when I was reading of the various ways to do it.

app-editors/vscodium/vscodium-1.56.2.ebuild
```diff
@@ -0,0 +1,95 @@
...
+S="${WORKDIR}"
+
+src_install() {
+	# Cleanup
+	rm "${S}/resources/app/LICENSE.txt" || die
+
+	# Install
+	pax-mark m codium
+	insinto "/opt/${PN}"
+	doins -r *
```
juippis, line 80: I think `doins -r ./.` is the preferred way to globbing, or `doins -r .` or something like that.

juippis (approved):
Thanks!

AdelKS:
Thank you too!

I just had a look at vscode's [build from source instructions](https://github.com/microsoft/vscode/wiki/How-to-Contribute) and it seems that all the needed dependencies are actually in Gentoo! I will see how easy it is to compile it, we may see a `vscode-oss` coming in hot xD

PR #21046, opened 2021-05-30
dev-ruby/rdoc: fix shebang for gentoo prefix

littlewu2508:

### DIFF
--- a/dev-ruby/rdoc/rdoc-6.1.2.ebuild
+++ b/dev-ruby/rdoc/rdoc-6.1.2.ebuild
@@ -11,7 +11,7 @@ RUBY_FAKEGEM_EXTRADOC="History.rdoc README.rdoc RI.rdoc TODO.rdoc"
 RUBY_FAKEGEM_BINWRAP=""
 RUBY_FAKEGEM_BINDIR="exe"
 
-inherit ruby-fakegem eutils
+inherit prefix ruby-fakegem eutils
 
 DESCRIPTION="An extended version of the RDoc library from Ruby 1.8"
 HOMEPAGE="https://github.com/ruby/rdoc/"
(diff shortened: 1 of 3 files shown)


dev-ruby/rdoc/rdoc-6.1.2.ebuild
```diff
@@ -82,6 +82,7 @@ all_ruby_install() {
 				ruby_fakegem_binwrapper $bin /usr/bin/${bin}${version}
 				sed -i -e "1s/env ruby/ruby${version}/" \
 					"${ED}/usr/bin/${bin}${version}" || die
+				if use prefix; then hprefixify ${ED}/usr/bin/${bin}${version}; fi
```
graaff, line 85: The argument to hprefixify needs to be quoted here, since ${ED} may contain spaces.

graaff:
The rdoc ebuilds did already change a bit so this commit did no longer apply as-is. I have added your change to the 6.3.2 ebuild in 8c3c4b660050062f63f964abd75e83f39ea92136, but using the more compact "use prefix &&" syntax.

Thanks for the fix!

heroxbd:
This patch back fires with `ruby-fakegem.eclass`.  The line 
```bash
rubycmd="$(ruby_implementation_command ${implementation})"
```
finds a ruby implementation inside a Prefix, not needed to be prefixified again.

thesamesam:
> This patch back fires with `ruby-fakegem.eclass`. The line
> 
> ```shell
> rubycmd="$(ruby_implementation_command ${implementation})"
> ```
> 
> finds a ruby implementation inside a Prefix, not needed to be prefixified again.

Actually, thinking about it, probably better to just open a new PR. Opening the old one is easier to miss (and confusing as there's no patch to be applied). It's more of a bug report.

heroxbd:
> Actually, thinking about it, probably better to just open a new PR. Opening the old one is easier to miss (and confusing as there's no patch to be applied). It's more of a bug report.

Sure, I have come up with a revised patch.  Will open another PR.

PR #21051, opened 2021-05-30
dev-python/logfury: Allow Python3.9

WPettersson:
Allows building of logfury for python 3.9

Tests and the package itself both run fine on my PC

### DIFF
--- a/dev-python/logfury/logfury-0.1.2-r1.ebuild
+++ b/dev-python/logfury/logfury-0.1.2-r1.ebuild
@@ -1,9 +1,9 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
 
-PYTHON_COMPAT=( python3_7 python3_8 )
+PYTHON_COMPAT=( python3_{7..9} )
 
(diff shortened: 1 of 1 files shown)


dev-python/logfury/logfury-0.1.2-r1.ebuild
```diff
@@ -1,9 +1,9 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=6
 
-PYTHON_COMPAT=( python3_7 python3_8 )
+PYTHON_COMPAT=( python3_{7,8,9} )
```
mgorny, line 6:
```suggestion
PYTHON_COMPAT=( python3_{7..9} )
```

mgorny (changes requested):
Please also bump to EAPI 7 while you're touching it, DEPEND → BDEPEND.

mgorny (approved):
Thanks! Tested and added to my merge queue™.

WPettersson:
Thanks for the very quick response, I was at least going to see what the QA bot says first before commenting :)

mgorny:
Just run `pkgcheck scan --commits --net` and you shouldn't have a bot to worry about.

PR #21053, opened 2021-05-30
dev-python/*: some more bumps to 3.10

vaartis:
- Small packages with other dependencies already bumped to 3.10
- Removed the python2 compat patch from send2trash
- Replaced setup.py test run with with whatever the package actually uses since setup.py test run is deprectaed


dev-python/send2trash/send2trash-1.5.0-r1.ebuild
```diff
@@ -21,8 +21,4 @@ SLOT="0"
 LICENSE="BSD"
 KEYWORDS="amd64 ~arm64 x86"
 
-distutils_enable_tests setup.py
-
-PATCHES=(
-	"${FILESDIR}/${P}-fix-broken-tests-on-py2.patch"
```
mgorny, line 27: Good thinking!

mgorny (approved):
Thanks. Lemme just finish testing it and I'll merge.

mgorny:
Actually, are you sure that lz4 tests worked for you? I seem to get with py3.10:

```
FAILED tests/frame/test_frame_5.py::test_frame_open_decompress_mem_usage[data0] - MemoryError
```

I thought it was just a fluke the first time but I can reproduce it every time.

vaartis:
> Actually, are you sure that lz4 tests worked for you? I seem to get with py3.10:
> 
> ```
> FAILED tests/frame/test_frame_5.py::test_frame_open_decompress_mem_usage[data0] - MemoryError
> ```
> 
> I thought it was just a fluke the first time but I can reproduce it every time.

It's a complicated one, The mem_usage tests require an insane amount of ram. Whether they work or not depends on how much free ram you have, at least that's how it was for me.

mgorny:
Sure but they pass on py<3.10, so it seems that something is off. Also no symptoms of running OOM.

vaartis:
This was from when i was tesing it previously, and that's on 3.8
`Out of memory: Killed process 25168 (python3.8) total-vm:4526916kB, anon-rss:4392644kB, file-rss:0kB, shmem-rss:0kB, UID:250 pgtables:8708kB oom_score_adj:0`

When re-running it with portage i got a error on 3.10 now too, strangely, even though I'm pretty sure it worked. The error is coming from C code [here](https://github.com/python-lz4/python-lz4/blob/v3.1.3/lz4/frame/_frame.c#L1263), i'll spend some time debugging this and will come back to you with the results.

vaartis:
OK, so problem is that somewhere there is an integer returned which is interpreted as an unsigned integer, and then python tries to allocate (unsigned integer)-1 bytes of memory, which is obviously not possible.

mgorny:
Could you please file a bug upstream in the meantime? Maybe they'll know what to do ;-). Maybe start by testing from git clone.

Thanks for doing this. Good to have someone knowing their way around this around.

vaartis:
> Could you please file a bug upstream in the meantime? Maybe they'll know what to do ;-). Maybe start by testing from git clone.
> 
> Thanks for doing this. Good to have someone knowing their way around this around.

I've been doing this in a git clone already, and it happens in the latest version too. I've figured out what actually messes things up: it's `PyArg_ParseTupleAndKeywords`, considreing that this is a function that is "part of the stable ABI", this _may_ be a python bug..

mgorny:
I honestly doubt it's a Python bug — more likely some change that just uncovered a hidden bug.

Apparently it's not supposed to touch the variable if the respective Python parameter is not passed. Are you saying that it's overwriting it nevertheless, or failing to convert valid Python value correctly?

mgorny:
I think your diagnosis is wrong: the test passes `max_length = 9223372036854775807` and it gets set correctly on the C side: `max_length=9223372036854775807` (my printf).

mgorny:
Ok, I think I know what's wrong. Let's move to upstream bug.

vaartis:
I see, thank you for looking into it. I'm unfortunately not very knowledgeble in these python nuances.. Hopefully upstream responds with something at least, they haven't had any commits in a few months.

mgorny:
Thank you! I just probably have more experience, that's all. You're doing a really great job.

mgorny:
That said, if you feel that doing lz4 is important right now, I think we might get away with simply trimming the value passed to `read()` somewhere. After all, I think it's perfectly legal for a partial `read()` to happen there.

PR #21061, opened 2021-05-31
Bumps for media-libs/kcolorpicker and media-libs/kimageannotator

dflogeras:


media-libs/kcolorpicker/kcolorpicker-0.1.6.ebuild
```diff
@@ -0,0 +1,41 @@
...
+	dev-qt/qtwidgets:5[png]
+"
+DEPEND="${RDEPEND}
+	test? ( dev-qt/qttest:5 )
+"
+
+S="${WORKDIR}/${MY_P}"
+
+src_configure() {
+	local mycmakeargs+=(
```
a17r, line 33: No need for that `+=`, these are your local mycmakeargs and `cmake.eclass` will make sure of adding them/not replacing its own defaults. :)

a17r (changes requested):

juippis (approved):

media-libs/kcolorpicker/kcolorpicker-0.1.6.ebuild
```diff
@@ -30,7 +30,7 @@ DEPEND="${RDEPEND}
 S="${WORKDIR}/${MY_P}"
 
 src_configure() {
-	local mycmakeargs+=(
+	local mycmakeargs=(
 		-DBUILD_TESTS=$(usex test)
 	)
```
juippis: This fix is in a wrong commit though :(

dflogeras: Shoot, I would have happily fixed that before the merge.  Sorry, I'll do better.  Thanks for merging!

PR #21062, opened 2021-05-31
app-misc/klavaro: drop unneeded dependency to gtk-builder-convert.

gryf:
Bug: https://bugs.gentoo.org/793218


juippis (approved):
Thanks!

PR #21070, opened 2021-06-01
Enable tests for app-shells/autojump

mattst88:
To do this, I needed to change `dev-vcs/pre-commit` from using `python-single-r1` to `python-r1`.

Closes: https://bugs.gentoo.org/530742

### DIFF
--- a/app-shells/autojump/autojump-22.5.3-r1.ebuild
+++ b/app-shells/autojump/autojump-22.5.3-r1.ebuild
@@ -1,0 +1,84 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{8..9} )
+DISTUTILS_SINGLE_IMPL=1
+DISTUTILS_USE_SETUPTOOLS=no
+
+inherit distutils-r1 bash-completion-r1 vcs-snapshot prefix
+
+DESCRIPTION="change directory command that learns"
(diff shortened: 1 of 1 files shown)


profiles/arch/x86/package.use.mask
```diff
@@ -1,6 +1,10 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Matt Turner <mattst88@gentoo.org> (2021-05-31)
```
mgorny, line 4: Just add `~x86`, nobody cares anyway.

mgorny (commented):
Well, I suppose you could make autojump single-impl but I suppose either way is fine.

Now, I wouldn't do all that profile masking because that's going to be PITA in the future. Just revbump it and request rekeywording.

profiles/arch/x86/package.use.mask
mattst88: Fair point wrt keywording and revbump vs masking. Will do

mattst88:
> Well, I suppose you could make autojump single-impl but I suppose either way is fine.

I actually wanted to do this, but I don't know how to handle this bit with python-single-r1:

````
	dobin bin/"${PN}"
	python_replicate_script "${ED}"/usr/bin/"${PN}"
````

since autojump doesn't have a standard build system. Any suggestions?

**Comment by mgorny** [comment]
```
python_doscript ...
```

(in python-r1 you'd add `python_foreach_impl ...`)

PR #21071, opened 2021-06-01
dev-java/bndlib: bump to 5.3.0 

vaukai:


fordfrog:
@vaukai that's a lot of work :-) i still did not have enough time to test it all, but what i noticed is that some ebuilds miss dependency on `zip` for unpacking, and at least one has it in `DEPEND` instead of `BDEPEND`. also, wanted to ask, as i did not have time to test it yet, do the osgi packages contain osgi information in `MANIFEST.MF`? though we so far did not bother to provide the osgi info there, at least for these packages we should, and in future it might be needed we mind it for all relevant packages. we already have [this eclass](https://github.com/gentoo/gentoo/blob/master/eclass/java-osgi.eclass) that provides some osgi support. and there are already some packages using the eclass, just grep the tree for `java-osgi`. if this is a new topic for you, i will give you a brief info if you want.

vaukai:
> @vaukai that's a lot of work :-) i still did not have enough time to test it all, but what i noticed is that some ebuilds miss dependency on `zip` for unpacking, and at least one has it in `DEPEND` instead of `BDEPEND`. 

@fordfrog Thanks for this first review.  I will add / adjust all the missing / misplaced `app-arch/unzip`.  Please be aware of https://bugs.gentoo.org/793863 where I guess that also anything from [JAVA_CLASSPATH_EXTRA](https://github.com/gentoo/gentoo/blob/3f1cd0c161eeec3a5b0173da31f8949fa5eb38b5/eclass/java-pkg-simple.eclass#L66-L70) should go to BDEPEND..

> also, wanted to ask, as i did not have time to test it yet, do the osgi packages contain osgi information in `MANIFEST.MF`? though we so far did not bother to provide the osgi info there, at least for these packages we should, and in future it might be needed we mind it for all relevant packages. 

To answer this question I need to check the details while I'm at those ebuilds.

>we already have [this eclass](https://github.com/gentoo/gentoo/blob/master/eclass/java-osgi.eclass) that provides some osgi support. and there are already some packages using the eclass, just grep the tree for `java-osgi`. if this is a new topic for you, i will give you a brief info if you want.

On solving some compile errors I came across that `eclass` but didn't see any need to care for it.  Should I file a bug for `java-ebuilder` to have this `eclass` automatically added to `inherit` in case it's needed??

Another point where I need help is the [is not abstract and does not override ...](https://github.com/gentoo/gentoo/pull/21071/commits/4138fc4d87fb4e50dd13a5f8b87d0bf3beea5d22#diff-b8478db75c2ca9d454346c62c539f7e1e20ffdf244f1f88dd824d3b133904195R91-R93) which prevents `dev-java/bndlib` from using the `dev-java/osgi-service-log` ebuild.

fordfrog:
the issue with `unzip` not in `BDEPEND` should be fixed now.

wrt `JAVA_CLASSPATH_EXTRA`, when considering java and osgi, iirc the build copies the implemented interfaces from osgi packages to the jar itself and so it's not a build tool but rather a library so it really goes to `DEPEND` in this case. that probably explains the issue with implementing the logging interface. to show you how it looks for one of my libraries, here's the content:
```
Archive:  /home/fordfrog/.m2/repository/com/fordfrog/com.fordfrog.osgi.datasources/2.0.3/com.fordfrog.osgi.datasources-2.0.3.jar
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  META-INF/
     711  Defl:N      406  43% 02-03-2021 14:23 282b4bff  META-INF/MANIFEST.MF
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  META-INF/maven/
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  META-INF/maven/com.fordfrog/
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  META-INF/maven/com.fordfrog/com.fordfrog.osgi.datasources/
      76  Defl:N       64  16% 02-03-2021 14:23 3d858fe0  META-INF/maven/com.fordfrog/com.fordfrog.osgi.datasources/pom.properties
    2801  Defl:N      764  73% 02-03-2021 14:22 f8352634  META-INF/maven/com.fordfrog/com.fordfrog.osgi.datasources/pom.xml
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  OSGI-INF/
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  OSGI-INF/metatype/
    1444  Defl:N      596  59% 09-10-2015 09:21 d4128c2f  OSGI-INF/metatype/metatype.xml
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  com/
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  com/fordfrog/
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  com/fordfrog/osgi/
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  com/fordfrog/osgi/datasources/
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  com/fordfrog/osgi/datasources/impl/
    2309  Defl:N     1025  56% 02-03-2021 14:23 1205279b  com/fordfrog/osgi/datasources/impl/Activator.class
    7301  Defl:N     3167  57% 02-03-2021 14:23 0c34e82e  com/fordfrog/osgi/datasources/impl/DataSourceInfo.class
    4681  Defl:N     1899  59% 02-03-2021 14:23 5f5a0b29  com/fordfrog/osgi/datasources/impl/DataSourcesFactory.class
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  org/
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  org/osgi/
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  org/osgi/service/
       0  Defl:N        2   0% 02-03-2021 14:23 00000000  org/osgi/service/jdbc/
    1624  Defl:N      765  53% 02-01-2018 21:06 9a75115f  org/osgi/service/jdbc/DataSourceFactory.class
     231  Defl:N      181  22% 02-01-2018 21:06 35b6102a  org/osgi/service/jdbc/package-info.class
--------          -------  ---                            -------
   21178             8897  58%                            24 files
```
notice the `org.osgi.service.jdbc` package there.

wrt inheriting `java-osgi`, it really depends on whether the package provides osgi support or not, which imo is not a simple way to detect from `pom.xml`, but proly would be doable.

vaukai:
First step done, `BDEPEND="app-arch/unzip"`for all

vaukai:
Second step done, `dev-java/osgi-service-log: try with java-osgi.eclass` doesn't help.  
Error still occurs :-(

fordfrog:
i think adding the java interface from `osgi-service-log` is the correct approach as i showed on my library which contains `org/osgi/service/jdbc/DataSourceFactory.class` though it is an interface from osgi and not coded by myself. this is how osgi works. i still wonder why the other osgi packages are needed when there is no need to copy anything from them to bnd.

vaukai:
> i think adding the java interface from `osgi-service-log` is the correct approach [...]

Yes. And it definitely needs `osgi-service-log-1.3.0`and fails with later versions.  

`osgi-service-log-1.3.0.ebuild ` now inherits `java-osgi.eclass` and has osgi informations in `MANIFEST.MF`.  This however doesn't help adding its java interface to bndlib.  Am I confusing different things?
```
src_install() {
   java-osgi_dojar "${S}"/${PN}.jar org.osgi.service.log org.osgi.service.log "org.osgi.service.log"
}
```
Before I start reworking all the other osgi packages could you please check the changed `osgi-service-log`

fordfrog:
here is a brief simplified info on osgi. osgi is a java runtime framework. it loads jar files, reads their `MANIFEST.MF` files including what they export and import. so for this to work, we need to provide that information in `MANIFEST.MF`. and this is what `java-osgi` does. so this is the runtime part.
when building, i suppose usually `bndlib` is used to collect and provide all the needed information, classes etc. but in our case we use the lowest approach, that is calling `javac` etc and hence we need to copy the implemented interfaces ourselves. so we should take the implemented interface from `orgi-service-log` and copy it to the source tree of `bndlib`. the best approach would probably be to depend on `dev-java/osgi-service-log[source]` and then unpack the needed java file (the implemented interface) from the source jar into the source tree of `bndlib`. and then we can compile.
these are just the basics. maybe i'll find out more that we need to do when i test the ebuilds. but this should be enough to provide the correct information about the exports in `MAINFEST.MF` for runtime and to compile `bndlib`.

vaukai:
> [...] and then unpack the needed java file (the implemented interface) from the source jar into the source tree of `bndlib`. and then we can compile.

Thanks for clarification.  That's exactly my approach with having `osgi-service-log` in `SRC_URI`.  
Then there seems no need at all to have it packaged or even installed.

vaukai:
And it's a possible last-rite candidate (once ant-ivy-2.5.0 hits the tree).

PR #21079, opened 2021-06-01
[Bug 793092] x11-libs/xapps-2.0.6 fails to build on prefix

SparkyBluefang:
https://bugs.gentoo.org/793092

`$PYTHON_GI_OVERRIDESDIR` contains the fully qualified python path, so no need to prefix it with `$ED`.

Also update xapp[s] maintainers after request from Olivier Laurantin:
```
When cinnamon was about to be removed from official gentoo tree, I was asked to be proxy maintainer of
dev-python/xapp and x11-libs/xapps because they were dependencies of another package I am maintaining
(net-wireless/blueberry). Now that you are here to maintain cinnamon packages, I think it would be better to
remove me from the proxy maintainers of those 2 packages. Is it ok for you? And can you deal with this bug?
```


thesamesam:
Thank you!

PR #21081, opened 2021-06-02
sys-libs/libtermcap-compat: install .so symlink

zavorka:
Linker fails to find '-ltermcap' unless there's a .so symlink to it.


zavorka:
Sorry, I failed to notice the comment
```
# We only want this for binary-only packages, so we will only be installing
# the lib used at runtime; no headers and no files to link against
```

PR #21085, opened 2021-06-02
app-emulation/virt-manager: Update live ebuild

zippy2:
There was a patch merged recently into the upstream repo
(v3.2.0-82-g2024068b) that introduced new dependency on
dev-python/tqdm. And while I'm at it - let's declare branch which
should be checked out (master).


zippy2:
CC @trofi

trofi:
Pulled as is. Thank you!

PR #21092, opened 2021-06-02
libvirt: 7.4.0 update

jpds:
Tested on `amd64`.


kobboi:
Ebuild is failing -> https://bugs.gentoo.org/795144

PR #21093, opened 2021-06-02
dev-java/jdbc-mysql: bump to 8.0.25

vaukai:


fordfrog:
@vaukai thanks! i can see only `dev-db/sqldeveloper` depends on this package. still, i would give it at least a week if somethings pops up.

vaukai:
> @vaukai thanks! i can see only `dev-db/sqldeveloper` depends on this package. still, i would give it at least a week if somethings pops up.

Except [removal (last-rite) of gnu-classpath](https://bugs.gentoo.org/786093) I can't see anything urgent here. So we might give it the regular 30 days.

PR #21097, opened 2021-06-02
*/*: fix BadDefaultUseFlags

thesamesam:


idl0r:
The HAProxy one, -zlib, is intended, because the preferred one is the internal SLZ instead and zlib is default in the linux profile.
Also there's no repoman warning about being a bad USE-flag.

ionenwks:
> The HAProxy one, -zlib, is intended, because the preferred one is the internal SLZ instead and zlib is default in the linux profile.

That's nice on paper, but it doesn't do anything. If try to emerge it'll still nag about about disabling zlib, having IUSE=-zlib doesn't auto-disable it. If want it easier on users, rather than REQUIRED_USE it may work better to default to slz if both are set (with at most a warning to inform, a bit like man-db for berkdb+gdbm).

mkszuba:
LGTM as far as my ebuilds are concerned.

ulm:
> -foo in IUSE is invalid

These changes look good, but can you update commit messages please? The quoted statement is not correct.

> Dropping '-' from these flags has no impact on the actual default behaviour.

That's not entirely accurate either, depending on the `USE_ORDER` setting it may or may not have impact.

PR #21099, opened 2021-06-03
www-apps/gitea: version bump

nemunaire:
Hi!

Please find here an updated gitea ebuild!


onkelbeh (commented):
Can confirm it works on production since 16 hours, medium load.
Forced this update because 1.13.7 crashed from time to time.
Will come back if this one should crash, too. Compiled with go-1.16.5.

PR #21110, opened 2021-06-03
sci-libs/vtk: add py39 support and improve all-modules USE flag

waebbl:
- add missing flags to REQUIRED_USE
- add new mysql USE flag - needed by all-modules, adds new dependency on
	dev-db/mysql-connector-c. Dev-db/mariadb-connector-c isn't supported
	by this version. Fix an issue with using of deprecated my_bool when
	using mysql-connector-c.
- remove xdmf2 USE flag, block sci-libs/xdmf2 with all-modules:
	sci-libs/xdmf2 doesn't ship any cmake configuration files, which leads
	to configuration errors when trying to build against external xdmf2
- block sci-libs/xdmf2: when built against internal xdmf2, there are header
	file collisions with the header files installed by sci-libs/xdmf2
- add dependency on sci-libs/kissfft, bug #754684) and drop bundled
- fix an issue with xdmf3 library being installed in both library locations,
	lib and lib64
- fix QA issues with RPATH
- byte compile python files
- disable searching for zfp (no package in tree)
- addresses some QA issues with defined but unused cmake variables: fixing all
	of them would need a thorough testing of each single USE flag and checking
	where these belong to, in this over-complex build system (bug #660582)

Closes: https://bugs.gentoo.org/793839
Closes: https://bugs.gentoo.org/754684
Closes: https://bugs.gentoo.org/646846
Bug: https://bugs.gentoo.org/660582

### DIFF
--- a/sci-libs/oce/oce-0.18.3-r2.ebuild
+++ b/sci-libs/oce/oce-0.18.3-r2.ebuild
@@ -29,7 +29,7 @@ DEPEND="
 	freeimage? ( media-libs/freeimage )
 	gl2ps? ( x11-libs/gl2ps )
 	tbb? ( dev-cpp/tbb )
-	vtk? ( =sci-libs/vtk-8*[boost,imaging,qt5,python,rendering,views,xdmf2] )"
+	vtk? ( =sci-libs/vtk-8*[boost,imaging,qt5,python,rendering,views] )"
 RDEPEND="${DEPEND}"
 
 CHECKREQS_MEMORY="256M"
--- a/sci-libs/vtk/files/vtk-8.2.0-0001-fix-library-installation-dir-for-xdmf3.patch
+++ b/sci-libs/vtk/files/vtk-8.2.0-0001-fix-library-installation-dir-for-xdmf3.patch
@@ -1,0 +1,57 @@
+From 3aad9ec6004b7e3a82817f2f1e2840fc5f07c050 Mon Sep 17 00:00:00 2001
+From: Bernd Waibel <waebbl-gentoo@posteo.net>
+Date: Thu, 3 Jun 2021 13:13:38 +0200
+Subject: [PATCH] fix library installation dir for xdmf3
+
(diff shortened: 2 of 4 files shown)


Alessandro-Barbieri:
The idea was to package ZFP first

waebbl:
Go on, if you wish. This was just a by-product of adding python-3.9 support. The support can easily be brought in later.

Alessandro-Barbieri:
I'll make an ebuild after dinner

waebbl:
Fine, I like that! :)
It can take a considerable amount of time until a new package finds its way into the tree, while having support for python-3.9 and fixing those all-modules issues need some relatively quick update.

waebbl:
Please note, I'm not really happy about
- remove xdmf2 USE flag, block sci-libs/xdmf2 with all-modules:
sci-libs/xdmf2 doesn't ship any cmake configuration files, which leads
to configuration errors when trying to build against external xdmf2
- block sci-libs/xdmf2: when built against internal xdmf2, there are header
file collisions with the header files installed by sci-libs/xdmf2

I couldn't think of a better way, though. Let me know if there is any.

@thesamesam ready for review

waebbl:
@zpuskas Could you please review and/or comment on the oce changes? Thanks!

zpuskas:
> @zpuskas Could you please review and/or comment on the oce changes? Thanks!

I've compiled OCE against recent VTK-8 without xdmf2 flag and it seems to build fine.

waebbl:
> I've compiled OCE against recent VTK-8 without xdmf2 flag and it seems to build fine. I'm fine with this change.

Thanks for the review :)

waebbl (commented):
Need to undo the changes in vtk-8.2.0-r1.ebuild. Started by doing the changes in `-r1` first, but then decided to revbump the ebuild. Seems I have not cleanly undone the changes in `-r1`.

NVM, got confused. The `-r1` changes commit the changes needed for support of py39.

PR #21111, opened 2021-06-03
net-misc/gallery-dl: remove python 3.7 and bump to 1.18.1

xgqt:
net-misc/gallery-dl: remove python 3.7 compat

net-misc/gallery-dl: drop old 1.17.3

net-misc/gallery-dl: bump to 1.17.5


xgqt:
out: 1.17.3 & 1.17.4
in: 1.17.5, 1.18.0 & 1.18.1

net-misc/gallery-dl/gallery-dl-1.17.5.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=(python3_{7..9})
+PYTHON_COMPAT=(python3_{8,9})
```
ionenwks, line 6: This isn't worth its own commit, 7 does nothing (can't be selected anymore), and eventually will be a `*/*` commit to remove all dead compat in the tree at once.

Wouldn't say if you were doing other changes at same time in which case you can silently drop it (not even worth a summary mention).  On that note, can 10 compat be added?

net-misc/gallery-dl/gallery-dl-1.18.1.ebuild
```diff
@@ -0,0 +1,42 @@
...
+if [[ ${PV} == *9999 ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/mikf/${PN}.git"
+else
+	SRC_URI="https://github.com/mikf/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64"
+fi
+
+# tests require network access
+RESTRICT="test"
```
ionenwks, line 24: Move this right below SLOT to respect skel.ebuid ordering, and given comment is short may want to inline it too, `"test" # comment`.

Also, not something I've messed with myself but, if tests are expected to work with network, then:
https://mgorny.pl/articles/the-ultimate-guide-to-eapi-8.html#properties-test-network-to-ease-reenabling-tests-requiring-internet

Can't bump this ebuild to eapi-8 yet but this new feature is retroactive.

net-misc/gallery-dl/gallery-dl-1.18.1.ebuild
```diff
@@ -0,0 +1,42 @@
...
+SLOT="0"
+
+RDEPEND=">=dev-python/requests-2.11.0[${PYTHON_USEDEP}]"
+
+distutils_enable_tests setup.py
+
+src_compile() {
+	emake data/completion/gallery-dl
+	emake data/completion/_gallery-dl
+	emake man
```
ionenwks, line 35: Do these need to be executed in order? (i.e. can -jX break this) If not:
```suggestion
	emake data/completion/{,_}gallery-dl man
```

net-misc/gallery-dl/gallery-dl-1.18.1.ebuild
```diff
@@ -0,0 +1,42 @@
...
+PYTHON_COMPAT=(python3_{8,9})
+PYTHON_REQ_USE="sqlite,ssl,xml"
+DISTUTILS_USE_SETUPTOOLS=rdepend
+
+inherit distutils-r1 optfeature
+
+DESCRIPTION="Download image galleries and collections from several image hosting sites"
+HOMEPAGE="https://github.com/mikf/gallery-dl"
+
+if [[ ${PV} == *9999 ]]; then
```
ionenwks, line 15: Trivial but I prefer to avoid `*` if not going to make use of it, this is always exactly 9999

ionenwks (commented):
Why 3 versions at once, 1.75.5, 1.18.0 and 1.18.1, are they all needed? This is also 3 new version that doesn't leave any of the previous for temporary safekeeping.

I'd normally imagine add 1.18.1, drop 1.17.3, and keep 1.17.4 for now unless it's unusable.

net-misc/gallery-dl/gallery-dl-1.17.5.ebuild
xgqt: >  On that note, can 10 compat be added?

havent tested (im not on 3.10 yet)

xgqt:
update

remaining versions are 1.17.5, 1.18.1 & 9999

net-misc/gallery-dl/gallery-dl-1.17.5.ebuild
ionenwks: Well, it's not urgent yet. I do suggest to add 3_10 alongside 3_9 to PYTHON_TARGETS if going to be working on some python ebuilds, would be nice to check if tests & runtime are fine for next version.

ionenwks:
> remaining versions are 1.17.5, 1.18.1 & 9999

If going to add 2 versions, please do 1 commit per version (dropping multiple at once is fine on the other hand, so can leave the other one alone). My previous comment was mostly about not having the separate commit for trivial changes on a version bump.

Also, manifest changes should be included rather than separate commit, i.e. manifest addition for 1.17.5 in the 1.17.5 commit. If commit using repoman or pkgdev, they'd add these for you and don't really have to think about it.

Other than that, looks good.

Guess will pass on test_network for now, tried and these tests have other problems.

xgqt:
> If going to add 2 versions, please do 1 commit per version (dropping multiple at once is fine on the other hand, so can leave the other one alone). My previous comment was mostly about not having the separate commit for trivial changes on a version bump.
> 
> Also, manifest changes should be included rather than separate commit, i.e. manifest addition for 1.17.5 in the 1.17.5 commit. If commit using repoman or pkgdev, they'd add these for you and don't really have to think about it.

done

ionenwks:
Thanks, merged as-is.

xgqt:
Thanks for cooperation & especially for short response times.

PR #21112, opened 2021-06-03
dev-dotnet/dotnet-sdk: Add

mid-kid:
This pull request adds a source-built dev-dotnet/dotnet-sdk to the tree, along with its unique dependency, dev-dotnet/dotnet-reference-packages.

The package is built using Fedora's sources for https://github.com/dotnet/source-build, since they're not provided by Microsoft themselves. This SDK is self-hosting, so it's bootstrapped using Microsoft's own build of it.

Bootstrapping is achieved through the use of USE=!system-bootstrap, in a similar fashion to dev-lang/rust. Unlike Rust, however, there isn't an equivalent to the minimal stage0 tarball, so bootstrapping from an equivalent dev-dotnet/dotnet-sdk-bin would be just as appropriate. However, doing that has two issues:
- The Private.SourceBuilt.Artifacts tarball that's required to build the source-built version offline, isn't shipped with the binary SDK, which would require a temporary dev-dotnet/dotnet-source-artifacts-bin
- The /usr/bin/dotnet symlink would conflict. This can be solved through an eselect module, but eselect doesn't make much sense for this package.

Ultimately, I'm not sure how to resolve this situation.

The package consists of three components:
- An SDK /usr/lib/dotnet/sdk/5.0.203
- A runtime /usr/lib/dotnet/shared/Microsoft.NETCore.App/5.0.6
- An ASP.NET runtime /usr/lib/dotnet/shared/Microsoft.AspNetCore.App/5.0.6
- A CLI /usr/lib/dotnet/dotnet

The CLI is capable of picking different versions of the SDK and runtimes through various flags and configuration files, and is capable of installing multiple of these side-by-side, into these directories. Previous art in terms of -bin packages for dotnet as seen in the ::dotnet repository separates these four components, and puts the different SDK and runtime versions in different SLOTs, while making sure only one CLI is installed.

That's why I hesitate to make an eselect helper for this, since the CLI practically voids the necessity of having eselect in the first place. However, the entire repository is built as one whole, and while there's different flags to build different parts of it, figuring it all out would be a significant undertaking. The dotnet developers expect distribution developers to split the packages up post-build.

Clang is required to build this package, as a GCC-built runtime crashes mid-build, the workaround has been borrowed from www-client/firefox, with some workarounds for LTO. I hesitate to add a USE=lto flag, since while I successfully built the package using LTO, I don't think it's supported upstream, and I don't want to encourage it.

A patch has been added to add additional architectures to the "Runtime IDs" file of dotnet. Upstream only includes gentoo-x64, but this adds the rest. It's not required, as it would otherwise fall back to linux-<arch>, but upstream seems adamant about advertising your exact distribution when downloading nuget packages. I haven't added the bootstrap SDKs nor the KEYWORDs for these architectures yet, as I lack the means to test them.


dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
```diff
@@ -0,0 +1,97 @@
...
+	if ! use system-bootstrap; then
+		DOTNET_ROOT="${WORKDIR}/dotnet"
+		return
+	fi
+
+	for x in "${DOTNET_ROOT}" "${BROOT}/usr/lib/dotnet"; do
+		if [ -d "${x}" ]; then
+			DOTNET_ROOT="${x}"
+		fi
+	done
```
ezzieyguywuf, line 60: Is the idea here to find whether one of `${DOTNET_ROOT}` or `${BROOT}/usr/lib/dotnet` exist? Is an `for` loop really needed for that, or would perhaps a `case` statement make more sense?

More importantly, what happens when both exist? As far as I can tell, in that case the last item (`${BROOT}/usr/lib/dotnet`) is selected for `DOTNET_ROOT` - is that the intent?

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+	export EXTRA_CFLAGS="${CFLAGS}"
+	export EXTRA_CXXFLAGS="${CXXFLAGS}"
+	export EXTRA_LDFLAGS="${LDFLAGS}"
+
+	./build.sh "${mybuildargs[@]}" || die
+}
+
+src_install() {
+	dodir /usr/lib/dotnet
+	tar xf artifacts/*/Release/dotnet-sdk-*.tar.gz -C "${ED}/usr/lib/dotnet"
```
thesamesam, line 219: Always use || die on external commands:
```suggestion
	tar xf artifacts/*/Release/dotnet-sdk-*.tar.gz -C "${ED}"/usr/lib/dotnet || die
```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+	dodir /usr/lib/dotnet/source-artifacts
```
thesamesam, line 220:
```suggestion

	dodir /usr/lib/dotnet/source-artifacts
```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+	tar xf artifacts/*/Release/Private.SourceBuilt.Artifacts.*.tar.gz -C "${ED}/usr/lib/dotnet/source-artifacts"
```
thesamesam, line 221:
```suggestion
	tar xf artifacts/*/Release/Private.SourceBuilt.Artifacts.*.tar.gz -C "${ED}"/usr/lib/dotnet/source-artifacts || die
	
```

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
```diff
@@ -0,0 +1,97 @@
...
+DESCRIPTION=".NET Source-build Reference Packages"
+HOMEPAGE="https://github.com/dotnet/source-build-reference-packages/"
+
+BOOT_PV="5.0.202"
+BOOT_ARTIFACTS_PV="0.1.0-${BOOT_PV}-1088763-20210414.1"  # See eng/Versions.props
+BOOT_SRC_URI="https://dotnetcli.azureedge.net/dotnet/Sdk/${BOOT_PV}/dotnet-sdk-${BOOT_PV}-linux-x64.tar.gz
+	https://dotnetcli.azureedge.net/source-built-artifacts/assets/Private.SourceBuilt.Artifacts.${BOOT_ARTIFACTS_PV}.tar.gz
+"
+
+MY_PN=source-build-reference-packages
```
thesamesam, line 21:
```suggestion
MY_PN="source-build-reference-packages"
```

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
```diff
@@ -0,0 +1,97 @@
...
+KEYWORDS="~amd64"
+IUSE="system-bootstrap"
+BDEPEND="
+	system-bootstrap? ( >=dev-dotnet/dotnet-sdk-${BOOT_PV} )
+"
+
+CHECKREQS_DISK_BUILD="14G"
+CHECKREQS_DISK_USR="400M"
+
+dotnet_unpack() {
```
thesamesam, line 39: Ideally, we'd make this a 'local' (not really) function within src_unpack, although you mentioned that may make it a pain for copying b/t ebuilds. We can consider an eclass (or hijacking one of the existing ones with a revision, maybe) later.

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
```diff
@@ -0,0 +1,97 @@
...
+		unpack "dotnet-sdk-${BOOT_PV}-linux-x64.tar.gz"
+		mkdir source-artifacts || die
+		cd source-artifacts || die
+		unpack "Private.SourceBuilt.Artifacts.${BOOT_ARTIFACTS_PV}.tar.gz"
+	fi
+}
+
+dotnet_find() {
+	if ! use system-bootstrap; then
+		DOTNET_ROOT="${WORKDIR}/dotnet"
```
thesamesam, line 52: Nit:
```suggestion
		DOTNET_ROOT="${WORKDIR}"/dotnet
```

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
```diff
@@ -0,0 +1,97 @@
...
+	fi
+}
+
+src_unpack() {
+	unpack "${P}.tar.gz"
+	dotnet_unpack
+}
+
+src_prepare() {
+	eapply "${FILESDIR}/dotnet-reference-packages-fix-bashisms.patch"
```
thesamesam, line 72: Nit:
```suggestion
	eapply "${FILESDIR}"/dotnet-reference-packages-fix-bashisms.patch
```

Could we use PATCHES instead?

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
```diff
@@ -0,0 +1,97 @@
...
+	sed -i -e '/setupSourceBuiltArtifacts.proj/c:' build.sh
+	default
+}
+
+src_configure() {
+	# --with-packages only likes being pointed at a tarball, so we symlink it
+	#  manually instead.
+	dotnet_find
```
thesamesam, line 80:
```suggestion
	dotnet_find
	
```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit check-reqs toolchain-funcs flag-o-matic multiprocessing
+
+DESCRIPTION=".NET SDK (includes runtime + aspnet)"
```
thesamesam, line 8: Need to check what the canonical style is:
```suggestion
DESCRIPTION=".NET SDK (includes runtime + ASP.NET)"
```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+RDEPEND="
+	app-crypt/mit-krb5
+	dev-libs/icu
+	dev-util/lttng-ust
+	system-libunwind? ( sys-libs/libunwind )
+"
+DEPEND="${RDEPEND}"
+BDEPEND="
+	dev-util/cmake
+	dev-vcs/git
```
thesamesam, line 49: What for?

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+IUSE="+system-libunwind system-bootstrap"
+RDEPEND="
+	app-crypt/mit-krb5
+	dev-libs/icu
+	dev-util/lttng-ust
+	system-libunwind? ( sys-libs/libunwind )
+"
+DEPEND="${RDEPEND}"
+BDEPEND="
+	dev-util/cmake
```
thesamesam, line 48: A shame but I guess it'll do given the build script is pretty large IIRC.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+	ninja|*)
+		BDEPEND+=" dev-util/ninja"
+		CMAKE_GENERATOR_FLAG="-ninja"
+		;;
+esac
+
+# If the package can't be built using gcc currently, set this to "true"
+FORCE_CLANG="${FORCE_CLANG:-true}"
+
+if ${FORCE_CLANG}; then
```
thesamesam, line 73: I know Chromium does this but I don't really want environment variables controlling dependencies, it [breaks](https://devmanual.gentoo.org/general-concepts/portage-cache/index.html) metadata invariance.

Need a USE flag for now.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+	BDEPEND+=" sys-devel/clang"
+fi
+
+dotnet_unpack() {
+	if ! use system-bootstrap; then
+		mkdir "${WORKDIR}/dotnet" || die
+		cd "${WORKDIR}/dotnet" || die
+		unpack "dotnet-sdk-${BOOT_PV}-linux-x64.tar.gz"
+		mkdir source-artifacts || die
```
thesamesam, line 82:
```suggestion

		mkdir source-artifacts || die
```

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
```diff
@@ -0,0 +1,97 @@
...
+src_unpack() {
+	unpack "${P}.tar.gz"
+	dotnet_unpack
+}
+
+src_prepare() {
+	eapply "${FILESDIR}/dotnet-reference-packages-fix-bashisms.patch"
+	sed -i -e '/setupSourceBuiltArtifacts.proj/c:' build.sh
+	default
+}
```
ezzieyguywuf, line 75: While not compulsory, I tend to prefer [using the `PATCHES` variable](https://devmanual.gentoo.org/ebuild-writing/functions/src_prepare/epatch/index.html#basic-eapply), since this is built-in to `EAPI 7`. It'd look something like this:

```suggestion
PATCHES=( "${FILESDIR}/${PN}-fix-bashisms.patch" )

src_prepare() {
    default
    sed -i -e '/setupSourceBuiltArtifacts.proj/c:' build.sh
}
```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+
+	default
+}
+
+src_configure() {
+	dotnet_find
+
+	# SDK is modified in-place...
+	if use system-bootstrap; then
+		mkdir -p dotnet
```
thesamesam, line 142:
```suggestion
		mkdir -p dotnet || die
```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+		cp -a "${DOTNET_ROOT}"/{dotnet,sdk} dotnet || die
+	else
+		ln -s "${DOTNET_ROOT}" dotnet || die
+	fi
+}
+
+src_compile() {
+	dotnet_find
+	dotnet_find_reference
+	rm -rf packages/archive  # Required for --with-packages to work
```
thesamesam, line 153:
```suggestion

    # Required for --with-packages to work
	rm -rf packages/archive || die

```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+	if ${FORCE_CLANG} && ! tc-is-clang; then
+		einfo "Enforcing the use of Clang due to build errors..."
+		AR=llvm-ar
+		CC=${CHOST}-clang
+		CXX=${CHOST}-clang++
+		NM=llvm-nm
+		OBJCOPY=llvm-objcopy
+		OBJDUMP=llvm-objdump
+		RANLIB=llvm-ranlib
+		if is-flagq '-flto*' && ! has_version 'sys-devel/clang[default-lld]' && ! is-ldflagq '-fuse-ld=lld'; then
```
thesamesam, line 188:
```suggestion

		if is-flagq '-flto*' && ! has_version 'sys-devel/clang[default-lld]' && ! is-ldflagq '-fuse-ld=lld'; then
```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+			has_version sys-devel/lld || die "Please install sys-devel/lld (or sys-devel/clang[default-lld]) to build with LTO"
+			einfo "Enforcing the use of LLD for Clang LTO..."
+			append-ldflags '-fuse-ld=lld'
+		fi
+		strip-unsupported-flags
```
thesamesam, line 193:
```suggestion

		strip-unsupported-flags
```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+	fi
+
+	# For the CLR_* variables, the full path is necessary, see:
+	# - src/runtime.*/eng/native/init-compiler.sh
+	# - src/runtime.*/eng/native/configuretools.cmake
+	# NM, OBJCOPY and OBJDUMP don't seem to be used currently (ver 5.0.203).
+	for var in AR CC CXX NM OBJCOPY OBJDUMP RANLIB; do
+		local prog="$(tc-get${var})"
+		local path="$(type -p "${prog}")"
+		test -n "${path}" || die "Can't find ${prog} (\$${var}) in \$PATH!"
```
thesamesam, line 203:
```suggestion
		test -n "${path}" || die "Can't find ${prog} (\$${var}) in \${PATH}!"
```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+}
+
+src_unpack() {
+	unpack "${MY_P}.tar.gz"
+	dotnet_unpack
+}
+
+src_prepare() {
+	add_patch() {
+		cp "$2" "patches/$1/9000-$(basename "$2")" || die
```
thesamesam, line 124:
```suggestion
		cp "${2}" "patches/${1}/9000-$(basename "${2}")" || die
```

dev-dotnet/dotnet-reference-packages/files/dotnet-reference-packages-fix-bashisms.patch
```diff
@@ -0,0 +1,13 @@
+Index: source-build-reference-packages-bcdc58a2ae9c2c06d769acfb8375d27adf5b2332/src/targetPacks/assemble.targets
+===================================================================
+--- source-build-reference-packages-bcdc58a2ae9c2c06d769acfb8375d27adf5b2332.orig/src/targetPacks/assemble.targets
++++ source-build-reference-packages-bcdc58a2ae9c2c06d769acfb8375d27adf5b2332/src/targetPacks/assemble.targets
```
ezzieyguywuf, line 4: Consider cleaning this up, per [these guidelines](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#clean-patch-howto). It might look like:

```suggestion
# Add a helpful comment here (this is optional though)
--- a/src/targetPacks/assemble.targets
+++ b/src/targetPacks/assemble.targets
```

dev-dotnet/dotnet-sdk/files/dotnet-sdk-5.0.203-runtime-locate-ranlib.patch
```diff
@@ -0,0 +1,21 @@
+Pick up the CLR_RANLIB variable properly, even when using Clang as a compiler
```
thesamesam, line 1: Any of these upstreamed? If so, link

thesamesam (changes requested):
Initial comments -- all surface level, not had a chance to think properly about the layout or anything yet (or test), but it seems reasonable.

I haven't repeated comments for the same functions across different ebuilds but of course they still apply.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+		CMAKE_GENERATOR_FLAG="-ninja"
+		;;
+esac
+
+# If the package can't be built using gcc currently, set this to "true"
+FORCE_CLANG="${FORCE_CLANG:-true}"
+
+if ${FORCE_CLANG}; then
+	BDEPEND+=" sys-devel/clang"
+fi
```
ezzieyguywuf, line 75: I don't understand the need for this - per your comments on the PR, it doesn't work with gcc. So why not just do away with `FORCE_CLANG` variable altogether?

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,224 @@
...
+	fi
+}
+
+src_unpack() {
+	unpack "${MY_P}.tar.gz"
+	dotnet_unpack
+}
+
+src_prepare() {
+	add_patch() {
```
ezzieyguywuf, line 123: consider adding a short comment here, explaining how the arguments work.

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
```diff
@@ -0,0 +1,97 @@
...
+	#  manually instead.
+	dotnet_find
+	mkdir artifacts || die
+	ln -s "${DOTNET_ROOT}/source-artifacts" artifacts/source-built || die
+}
+
+src_compile() {
+	dotnet_find
+
+	# Avoid dependency on dev-libs/icu
```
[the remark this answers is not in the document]
mid-kid: It's an optional runtime dependency of the prebuilt SDK, dotnet-sdk itself depends on this, so it doesn't really matter.

dev-dotnet/dotnet-reference-packages/files/dotnet-reference-packages-fix-bashisms.patch
```diff
@@ -0,0 +1,13 @@
+Index: source-build-reference-packages-bcdc58a2ae9c2c06d769acfb8375d27adf5b2332/src/targetPacks/assemble.targets
```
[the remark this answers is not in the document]
mid-kid: This is `dev-util/quilt` generates patches for me, and I do it the same way for `net-im/telegram-desktop`. Does it matter? Is there any tool to format patches according to these requirements? I really don't want to go in and hand-edit every patch I generate.

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
mid-kid: For `dotnet_find`, yeah, I agree there should be an eclass for it, but I'm not convinced dotnet_unpack won't change with future versions of these two bootstrap ebuilds.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: The build procedure uses `git apply` to apply all the patches. Don't ask me, ask upstream.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: Huh?

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: USE flags cannot be forced unless I edit files under `profile/`. Would it be OK to just set this to true, and not depend on the environment?

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: Can `mkdir -p` even fail?

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: I've been fully going by copying the style and use patterns of other ebuilds/eclasses. It might be useful to make `toolchain-funcs.eclass` comply with this as well, as I was reading that at the time.

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
mid-kid: I intended to break the loop early if one of them was found, preferring "${DOTNET_ROOT}" (set by /etc/env.d/) over a hardcoded path. Thanks for pointing this out. I should also make source-artifacts/ exists in the selected SDK.

dev-dotnet/dotnet-sdk/files/dotnet-sdk-5.0.203-runtime-locate-ranlib.patch
mid-kid: No, not sure if upstream would like them, will consider.

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
mid-kid: Yeah, I'd been fiddling too much with patching this and different sed lines that I forgot to clean this up back to a PATCHES array when I was done. Thanks.

dev-dotnet/dotnet-reference-packages/files/dotnet-reference-packages-fix-bashisms.patch
mid-kid: Same reply as here: https://github.com/gentoo/gentoo/pull/21112#discussion_r645053761
I'd rather not go into each patch and edit it manually if I don't have to (except for the header), though I might be able to configure quilt to get rid of that index line.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: I want to make it simple to disable the clang enforcement. The presence of a lot of `-gcc` switches in upstream's build scripts indicates that they at least *intend* it to work on GCC as well, so this situation might change in the future, be it through downstream or upstream patches and maintainance.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: Sure.

dev-dotnet/dotnet-reference-packages/files/dotnet-reference-packages-fix-bashisms.patch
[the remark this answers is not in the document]
mid-kid: Disregard. Apparently this can be fixed by setting `QUILT_REFRESH_ARGS="--no-index --no-timestamp -p ab"` in ~/.quiltrc

dev-dotnet/dotnet-reference-packages/files/dotnet-reference-packages-fix-bashisms.patch
mid-kid: Disregard. Apparently this can be fixed by setting `QUILT_REFRESH_ARGS="--no-index --no-timestamp -p ab"` in ~/.quiltrc

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,238 @@
...
+LICENSE="MIT Apache-2.0 BSD"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="+system-libunwind system-bootstrap"
+
+RDEPEND="
+	app-crypt/mit-krb5
+	dev-libs/icu
+	dev-util/lttng-ust
+	system-libunwind? ( sys-libs/libunwind )
```
telans, line 45:
```suggestion
IUSE="llvm-libunwind +system-libunwind system-bootstrap"

RDEPEND="
	app-crypt/mit-krb5
	dev-libs/icu
	dev-util/lttng-ust
	system-libunwind? (
		llvm-libunwind? ( sys-libs/llvm-libunwind )
		!llvm-libunwind? ( sys-libs/libunwind )
	)
```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,238 @@
...
+		cp "${2}" "patches/${1}/9000-$(basename "${2}")" || die
+	}
+
+	eapply "${FILESDIR}"/dotnet-sdk-5.0.203-additional-arguments.patch
+	add_patch sdk "${DISTDIR}"/dotnet-sdk-telemetry-optout.patch
+	add_patch runtime "${FILESDIR}"/dotnet-sdk-5.0.203-runtime-add-gentoo-ids.patch
+	add_patch runtime "${FILESDIR}"/dotnet-sdk-5.0.203-runtime-additional-arguments.patch
+	add_patch runtime "${FILESDIR}"/dotnet-sdk-5.0.203-runtime-as-needed.patch
+	add_patch runtime "${FILESDIR}"/dotnet-sdk-5.0.203-runtime-locate-ranlib.patch
+
```
telans, line 140:
```suggestion
	add_patch runtime "${FILESDIR}"/dotnet-sdk-5.0.203-runtime-locate-ranlib.patch
	add_patch runtime "${FILESDIR}"/dotnet-sdk-5.0.203-runtime-lld-build.patch
```

Please add the following to fix building the runtime with LLD:

```
--- a/eng/native/configurecompiler.cmake
+++ b/eng/native/configurecompiler.cmake
@@ -595,7 +595,7 @@
         ERROR_QUIET
         OUTPUT_VARIABLE ldVersionOutput)
 
-    if("${ldVersionOutput}" MATCHES "GNU ld" OR "${ldVersionOutput}" MATCHES "GNU gold")
+    if("${ldVersionOutput}" MATCHES "GNU ld" OR "${ldVersionOutput}" MATCHES "GNU gold" OR "${ldVersionOutput}" MATCHES "GNU linkers")
         set(LD_GNU 1)
     elseif("${ldVersionOutput}" MATCHES "Solaris Link")
         set(LD_SOLARIS 1)

```

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
telans: metadata.xml probably needs to be updated too

telans:
I'm having troubles building and I have no idea how to even begin looking at what's wrong:

```
/var/tmp/portage/dev-dotnet/dotnet-sdk-5.0.203/work/dotnet/sdk/5.0.202/MSBuild.dll /nologo -logger:/var/tmp/portage/dev-dotnet/dotnet-sdk-5.0.203/work/dotnet-v5.0.203-SDK/Tools/source-built/Microsoft.DotNet.Arcade.Sdk/tools/./net5.0/Microsoft.DotNet.Arcade.Sdk.dll -maxcpucount /m -verbosity:m /v:normal /bl:/var/tmp/portage/dev-dotnet/dotnet-sdk-5.0.203/work/dotnet-v5.0.203-SDK/src/msbuild.5e4b48a27efce55a613664b58d353ab4c8d1f6c1/artifacts/log/Release/Build.binlog /clp:Summary /nr:false /p:TreatWarningsAsErrors=false /p:ContinuousIntegrationBuild=true /p:Configuration=Release /p:RepoRoot=/var/tmp/portage/dev-dotnet/dotnet-sdk-5.0.203/work/dotnet-v5.0.203-SDK/src/msbuild.5e4b48a27efce55a613664b58d353ab4c8d1f6c1 /p:Restore=true /p:Build=true /p:Rebuild=false /p:Test=false /p:Pack=true /p:IntegrationTest=false /p:PerformanceTest=false /p:Sign=false /p:Publish=true /p:DotNetBuildFromSource=true /p:DotNetCoreSdkDir=/var/tmp/portage/dev-dotnet/dotnet-sdk-5.0.203/work/dotnet/ /p:DisableNerdbankVersioning=true /p:DotNetPackageVersionPropsPath=/var/tmp/portage/dev-dotnet/dotnet-sdk-5.0.203/work/dotnet-v5.0.203-SDK/artifacts/obj/x64/Release/PackageVersions.props /p:DotNetRestoreSourcePropsPath=/var/tmp/portage/dev-dotnet/dotnet-sdk-5.0.203/work/dotnet-v5.0.203-SDK/artifacts/obj/x64/Release/RestoreSources.props /p:DotNetBuildOffline=true /p:Projects=/var/tmp/portage/dev-dotnet/dotnet-sdk-5.0.203/work/dotnet-v5.0.203-SDK/src/msbuild.5e4b48a27efce55a613664b58d353ab4c8d1f6c1/MSBuild.SourceBuild.slnf /var/tmp/portage/dev-dotnet/dotnet-sdk-5.0.203/work/dotnet-v5.0.203-SDK/Tools/source-built/Microsoft.DotNet.Arcade.Sdk/tools/Build.proj
         MSBUILD : error MSB4016: The build stopped unexpectedly because the "ReusableLogger" logger failed unexpectedly during initialization. The underlying compression routine could not be loaded correctly.
         ##vso[task.logissue type=error]Build failed (exit code '1').
```

"The build stopped unexpectedly because the "ReusableLogger" logger failed unexpectedly during initialization. The underlying compression routine could not be loaded correctly."

Any ideas?

mid-kid:
@telans That sounds like an error similar to when building with GCC. I haven't tried building with clang 12 yet, but I suspect it might only build with clang 11.

telans:
> That sounds like an error similar to when building with GCC. I haven't tried building with clang 12 yet, but I suspect it might only build with clang 11. Alternatively, try to disable LTO...

I should've mentioned I did try without LTO but still with the same issue. I just tried Clang/LLVM 11 and it also failed at the same point

mid-kid:
Interesting, because I can't reproduce it. Maybe it's because you're using llvm-libunwind? I'll mess around in a prefix later, but building this package takes ages...

telans:
> Interesting, because I can't reproduce it. Maybe it's because you're using llvm-libunwind? I'll mess around in a prefix later, but building this package takes ages...

Solved it (loooong build times took a while). default-libcxx is the culprit.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,238 @@
...
+"
+DEPEND="${RDEPEND}"
+BDEPEND="
+	dev-util/cmake
+	dev-vcs/git
+	dev-dotnet/dotnet-reference-packages
+	system-bootstrap? ( >=dev-dotnet/dotnet-sdk-${BOOT_PV} )
+"
+
+CHECKREQS_DISK_BUILD="20G"
```
telans, line 55:
```suggestion
CHECKREQS_DISK_BUILD="24G"
```

My build failed with a 21G tmpfs. I saw it used just shy of 23G when installing.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,238 @@
...
+		OBJDUMP=llvm-objdump
+		RANLIB=llvm-ranlib
+
+		if is-flagq '-flto*' && ! has_version 'sys-devel/clang[default-lld]' && ! is-ldflagq '-fuse-ld=lld'; then
+			has_version sys-devel/lld || die "Please install sys-devel/lld (or sys-devel/clang[default-lld]) to build with LTO"
+			einfo "Enforcing the use of LLD for Clang LTO..."
+			append-ldflags '-fuse-ld=lld'
+		fi
+
+		strip-unsupported-flags
```
telans, line 205:
```suggestion
		fi

		has_version 'sys-devel/clang[default-libcxx]' && append-cxxflags -stdlib=libstdc++
		strip-unsupported-flags
```

Build fails with -stdlib=libcxx

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
```diff
@@ -0,0 +1,105 @@
...
+#  considered not to be *source code* per se, none of the files here contain
+#  any actual code. Only symbols/references for a variety of libs.
+
+DESCRIPTION=".NET Source-build Reference Packages"
+HOMEPAGE="https://github.com/dotnet/source-build-reference-packages/"
+
+BOOT_PV="5.0.202"
+BOOT_ARTIFACTS_PV="0.1.0-${BOOT_PV}-1088763-20210414.1"  # See eng/Versions.props
+BOOT_SRC_URI="https://dotnetcli.azureedge.net/dotnet/Sdk/${BOOT_PV}/dotnet-sdk-${BOOT_PV}-linux-x64.tar.gz
+	https://dotnetcli.azureedge.net/source-built-artifacts/assets/Private.SourceBuilt.Artifacts.${BOOT_ARTIFACTS_PV}.tar.gz
```
telans, line 18: Should this be renamed to something a little more related to the package?

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,238 @@
...
+
+inherit check-reqs toolchain-funcs flag-o-matic multiprocessing
+
+DESCRIPTION=".NET SDK (includes .NET Runtime + ASP.NET)"
+HOMEPAGE="https://dotnet.microsoft.com/"
+
+BOOT_PV="5.0.202"
+BOOT_ARTIFACTS_PV="0.1.0-${BOOT_PV}-1088763-20210414.1"  # See eng/Versions.props
+BOOT_SRC_URI="https://dotnetcli.azureedge.net/dotnet/Sdk/${BOOT_PV}/dotnet-sdk-${BOOT_PV}-linux-x64.tar.gz
+	https://dotnetcli.azureedge.net/source-built-artifacts/assets/Private.SourceBuilt.Artifacts.${BOOT_ARTIFACTS_PV}.tar.gz
```
telans, line 14: Likewise renamed

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: Wow, and I thought I was being generous. I might have forgotten to count the size of the bootstrap sdk in case of USE=-system-bootstrap...

dev-dotnet/dotnet-reference-packages/dotnet-reference-packages-1.0.0_p20210210.ebuild
mid-kid: I don't know. Any suggestions? Dotnet.Private.SourceBuilt.Artifacts?

telans: I would just append `dotnet-` so it's a tad easier to know what it's about

mid-kid:
Thanks for testing it and solving the issue! I take it USE=llvm-libunwind works fine?

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
telans: That's likely the reason. Though it does create a whole lot of archives near the end of the build - there's probably some switch somewhere to turn that off.

telans:
> I take it USE=llvm-libunwind and Clang 12 works fine?

Yep, I don't know of any run-time implications though but I highly doubt there are any

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,238 @@
...
+		RANLIB=llvm-ranlib
+
+		if is-flagq '-flto*' && ! has_version 'sys-devel/clang[default-lld]' && ! is-ldflagq '-fuse-ld=lld'; then
+			has_version sys-devel/lld || die "Please install sys-devel/lld (or sys-devel/clang[default-lld]) to build with LTO"
+			einfo "Enforcing the use of LLD for Clang LTO..."
+			append-ldflags '-fuse-ld=lld'
+		fi
+
+		strip-unsupported-flags
+	fi
```
telans, line 206:
```suggestion
	if ${DOTNET_FORCE_CLANG}; then
		if ! tc-is-clang; then
			einfo "Enforcing the use of Clang due to build errors..."
			AR=llvm-ar
			CC=${CHOST}-clang
			CXX=${CHOST}-clang++
			NM=llvm-nm
			OBJCOPY=llvm-objcopy
			OBJDUMP=llvm-objdump
			RANLIB=llvm-ranlib
			strip-unsupported-flags
		fi

		if is-flagq '-flto*' && ! has_version 'sys-devel/clang[default-lld]' && ! is-ldflagq '-fuse-ld=lld'; then
			has_version sys-devel/lld || die "Please install sys-devel/lld (or sys-devel/clang[default-lld]) to build with LTO"
			einfo "Enforcing the use of LLD for Clang LTO..."
			append-ldflags '-fuse-ld=lld'
		fi

		# Build failures with libcxx
		if has_version 'sys-devel/clang[default-libcxx]' || is-flagq '-stdlib=libc++'; then
			append-cxxflags '-stdlib=libstdc++'
			append-ldflags '-stdlib=libstdc++'
		fi
	fi
```

The LTO & libcxx workarounds should be applied regardless of `tc-is-clang`, if DOTNET_FORCE_CLANG is true. I assume DOTNET_FORCE_CLANG will eventually be replaced by a `clang` use flag? Perhaps it could be replaced now and just forced in profile?

mid-kid:
Just saw the commits surrounding dotnet, I'll go over the comments tomorrow. Sorry for taking so long.

zmedico:
We've got this WIP app-shells/pwsh ebuild that will need to make use of sdk:

https://github.com/gentoo/gentoo/pull/21451

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
```diff
@@ -0,0 +1,238 @@
...
+#  the proprietary "Darc CLI" tool, and downloading a bunch of nuget packages.
+
+# Fedora package script:
+# https://src.fedoraproject.org/rpms/dotnet5.0/blob/rawhide/f/build-dotnet-tarball
+# Or just run ./build-source-tarball.sh from the root of the source-build repo.
+# (the --base-tarball option might be worth using now it's added)
+
+# https://github.com/dotnet/runtime/blob/release/5.0/src/installer/pkg/packaging/deb/dotnet-runtime-deps-debian_config.json#L27
+LICENSE="MIT Apache-2.0 BSD"
+SLOT="0"
```
zmedico, line 37: We definitely need to be able to install multiple slots, like Java.

zmedico (changes requested):

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: Should I make the install location `/usr/lib/dotnet/dotnet-sdk-5.0`? Or just `/usr/lib/dotnet-sdk-5.0`? Do I need to support the `dotnet-symlink` USE flag?

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: It's a global USE flag so I'd reckon not.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: I see this is only required for clang built with `default-lld`, annoyingly enough. Added.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: Unfortunately I can't do much about that. The build system isn't very well documented and finding flags for anything at all seems near impossible at times.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: The `clang` and `lto` USE flags seem very inconsistent in the tree right now, and since we only want to control whether we *force* clang or not (probably on a version-by-version basis), I think this is the saner approach. Unlike firefox (where using clang allows LTO to work across the rust/c++ boundary, and LTO is handled by the build system instead of `*FLAGS`), there's no inherent benefit to using clang when GCC also builds correctly, so I doubt users would want to toggle it. I'm open to changing it, but if you phrase it like a question, my answer is I wouldn't.

I was mostly assuming that if the user has set up clang as their default compiler *and* enabled LTO, they would likely be using LLD already, as not doing that breaks many other builds or just doesn't LTO at all. This check is purely for GCC users running full-system LTO setups that might wish to install LLD to also LTO this package. That said, `-flto=n` will get stripped by `strip-unsupported-flags`, so this might not be extremely effective unless the user is using `-flto` raw.
That said, the libstdc++ check is different, I think I'll leave that outside of the DOTNET_FORCE_CLANG variable, instead only apply it if `tc-is-clang`, as this should be applied regardless of whether the ebuild forces clang or not.

mid-kid:
@zmedico Updated the ebuild, please lmk what you think.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
mid-kid: I used `/usr/lib/dotnet-sdk-5.0`, added support for `dotnet-symlink` and made the env.d file contain EPREFIX.

dev-dotnet/dotnet-sdk/dotnet-sdk-5.0.203.ebuild
ezzieyguywuf: I suppose if the intent in upstream is to work with GCC it's cool. Personally, I prefer to add these types of complexities/indirections when they're actually needed, rather than planning for some eventual future need.

However if it is easier to do this now rather then later, then it definitely makes sense.

telans:
I'd still love to see this make it into ::gentoo

mid-kid:
Dotnet 6.0 released, providing a new build system, and this, combined with the realization that this is yet another language that will be a massive pain to wrangle in terms of getting apps written in it packaged, and how little interest there seems to be, killed my mood in attempting to maintain this package. (not to mention it takes an hour and a half to build, it's insane.)

For anyone trying to wrangle this turd, here's some resources that might be useful:
- Fedora's dotnet6.0 package sources: https://src.fedoraproject.org/rpms/dotnet6.0/tree/rawhide
- The new build system repository, I hope it's not as big an XML mess as the previous one (it probably is): https://github.com/dotnet/installer
- Please take a look at how the ebuild in this PR is written, particularly the way the bootstrap binaries are optional and fixed to a specific version (the oldest known to bootstrap fine), the way the source tarball was being generated without included bootstrap binaries, the various patches to make sure MAKEOPTS and other CMake options are passed properly through the build system (as well as the other patches), they way I ensured LTO is supported and the proper compilers are picked (see src_compile), and everything else. I spent weeks to make sure everything worked right.

It's worth noting that the old build system used a proprietary DARC Cli to parse metadata, fetch the sources with git and generate some files and the source tarball. Once the source tarball was generated this didn't contain any prebuilts anymore and was fully bootstrappable from scratch, but be aware that not everything Microsoft makes is free software. This situation might have improved with the new build system.

Good luck to whoever attempts this in the future, they'll need it.

mausys:
Thanks you very much for the ebuild. I updated and uploaded it to my repository:
https://github.com/mausys/portage/blob/master/dev-dotnet/dotnet-sdk/dotnet-sdk-6.0.102.ebuild 
but without all the cflags/cmake/compiler options and only for amd64.

mid-kid:
Unfortunately an ebuild like that likely won't work with GCC as the default compiler, unless they somehow fixed support for it in 6.x

thesamesam:
Would you mind rebasing and then I'll get this in?

xgqt:
`--with-sdk` does not work with reference-packages >= v6.0.102

xgqt:
@mid-kid

`installer` source build is broken for me, it passes `./build.sh /p:ArcadeBuildTarball=true /p:TarballDir="$(pwd)/dotNET_SDK"` and `./prep.sh --bootstrap` but fails `./build.sh`.

Have you tried with 6.0 or 7.0? Any success?

xgqt:
> @mid-kid
> 
> `installer` source build is broken for me, it passes `./build.sh /p:ArcadeBuildTarball=true /p:TarballDir="$(pwd)/dotNET_SDK"` and `./prep.sh --bootstrap` but fails `./build.sh`.
> 
> Have you tried with 6.0 or 7.0? Any success?

@zmedico @telans @mid-kid @thesamesam @mausys  Initial 7.0.105 version here: https://gitlab.gentoo.org/dotnet/gentoo-dotnet-2023/-/blob/d2d55eafde0df1e29d82fc143b7e20f7776a6188/dev-dotnet/dotnet-sdk/dotnet-sdk-7.0.105.ebuild

xgqt:
Very experimental version of 7.0.105 added to the tree: https://packages.gentoo.org/packages/dev-dotnet/dotnet-sdk

As discussed previously >=7.0 bootstrap works differently and since this PR also introduced EOL SDK version - 5.0 I will close this now.

Thanks a lot @mid-kid for help!

PR #21113, opened 2021-06-03
dev-python/pyjsparser: remove py3.7 & dev-python/js2py: bump to 0.71

xgqt:
dev-python/pyjsparser: remove py3.7 compatibility

dev-python/js2py: remove py3.7 compatibility

dev-python/js2py: bump to 0.71


juippis (approved):
Don't think we'd need to drop python-3.7 like this, but since it doesn't hurt either...
thanks!

xgqt:
Thanks for cooperation!

PR #21115, opened 2021-06-03
dev-libs/zfp: initial import

Alessandro-Barbieri:


Alessandro-Barbieri:
@waebbl I can hand over the ebuild to you if it can speedup things

waebbl:
I doubt it will speed up anything. I don't have merge permission either. Just wait until a dev picks up on it. It can take considerable time, until this happens, so just be patient. There's no need whatsoever to pressure on it. In fact, experience tells, that pressuring too much, often leads to the opposite effect.

juippis:
@waebbl what do you feel about this, is this still necessary? Wasn't vtk fixed so it won't care about zfp anymore even if it is installed?

In other words, is there **any** worth having this in Gentoo? If you agree there is, then let's push it.

waebbl:
@juippis the zfp library support in vtk is currently disabled. Yet it is available and could probably be used. I never had a chance to test whether it will build against an external zfp, but in theory it should, as it's location in the ThirdParty subdir tells. There are only a very few libraries which need special actions to work, like the libharu library.

Vtk (the ebuild) was fixed to not care about it, yes. But this was due to the fact, that we didn't had the library in the tree. From my understanding from old vtk-{6,7,8} ebuild comments is, that the goal was to get as much of the third party libraries being external. From this POV i'd say it would be worth having it (provided vtk builds with the external library).

waebbl:
vtk-9 still supports zfp, see https://gitlab.kitware.com/vtk/vtk/-/tree/master/ThirdParty/zfp

PR #21118, opened 2021-06-04
net-im/telegram-desktop: Fix USE=-dbus

mid-kid:
@gyakovlev


gyakovlev:
man I'm so glad you've taken over maintenance, doing a great job!
thanks, merged.

PR #21126, opened 2021-06-05
 sys-block/targetcli-fb: update to python3

DocMAX:


scardracs:
Please update PR name to  sys-block/targetcli-fb: update to python3.9

sys-block/targetcli-fb/targetcli-fb-2.1.53.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7,8} )
+PYTHON_COMPAT=( python3_{7,8,9} )
```
scardracs, line 6: Change with PYTHON_COMPAT=( python3_{8,9} ) as python3.7 is not used anymore. Also try 10 if it works

scardracs:
You have also to check for qa errors

juippis:
Hey, 

you're also missing python-3.9 from the required deps - targetcli-fb wouldn't work with python-3.9 unless they supported python-3.9 too. 
Then this package, and the dep that's missing python-3.9, could also use a version bump. 

I've pinged the maintainer about it, and since you haven't fixed the QA issue I assume you don't wish to work on this anymore. If you do, happy to merge your work - but please do the version bumps and add python-3.9 to necessary deps while at it.

PR #21127, opened 2021-06-05
dev-games/openscenegraph-openmw: fix build with gcc 11

DarthGandalf:
Closes: https://bugs.gentoo.org/788301
Signed-off-by: Alexey Sokolov <sokolov@google.com>


ionenwks:
Thanks, builds fine -- I added the bugref to the patch header too.

PR #21130, opened 2021-06-05
sci-libs/ceres-solver: add python-3.9 support

waebbl:
Bugday 2021-06-05

### DIFF
--- a/sci-libs/ceres-solver/ceres-solver-1.14.0.ebuild
+++ b/sci-libs/ceres-solver/ceres-solver-1.14.0.ebuild
@@ -1,10 +1,10 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
 CMAKE_ECLASS=cmake
-PYTHON_COMPAT=( python3_{7,8} )
+PYTHON_COMPAT=( python3_{7,8,9} )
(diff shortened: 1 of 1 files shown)


AndrewAmmerlaan:
Thanks, I also changed this in a follow-up commit:
```
	doc? (
		$(python_gen_any_dep '
			dev-python/sphinx[${PYTHON_USEDEP}]
			dev-python/sphinx_rtd_theme[${PYTHON_USEDEP}]
		')
```
Since we need to ensure that we have both sphinx and sphinx_rtd_theme with the same PYTHON_TARGET

waebbl:
> Thanks, I also changed this in a follow-up commit:
> 
> ```
> 	doc? (
> 		$(python_gen_any_dep '
> 			dev-python/sphinx[${PYTHON_USEDEP}]
> 			dev-python/sphinx_rtd_theme[${PYTHON_USEDEP}]
> 		')
> ```
> 
> Since we need to ensure that we have both sphinx and sphinx_rtd_theme with the same PYTHON_TARGET

Is it working this way? `python-any-r1.eclass` states, that they should be checked in a `python_check_deps()` function declared in the ebuild, and that PYTHON_USEDEP is set locally for python_check_deps() calls.

AndrewAmmerlaan:
> Is it working this way? `python-any-r1.eclass` states, that they should be checked in a `python_check_deps()` function declared in the ebuild, and that PYTHON_USEDEP is set locally for python_check_deps() calls.

It worked when i tried it earlier, but yes `python_check_deps()` should probably be defined as well.

PR #21131, opened 2021-06-05
sci-libs/plplot: add support for python-3.9

waebbl:
Bugday 2021-06-05
Byte-compile python package to avoid QA warning


AndrewAmmerlaan:
Thanks, there's a test failure in `examples_python.py`, but it appears to be unrelated to this since it also fails in older python versions.

PR #21135, opened 2021-06-05
media-libs/gegl[openexr]: fix build against >=dev-libs/glib-2.67.3, =media-libs/openexr-3.0.1

band-a-prend:
This pull request fix two issues of built media-libs/gegl[openexr].

Closes: https://bugs.gentoo.org/793998  
Closes: https://bugs.gentoo.org/788400


thesamesam:
Note that you don't need to revbump for broken builds

band-a-prend:
They become broken with USE=openexr after upgrade of glib/openexr to certain versions. Sould I anyway just to update current ebuilds?

PR #21136, opened 2021-06-05
games-engines/openmw: 0.46 requires OSG 3.4

DarthGandalf:
This is in preparation for future 0.47. The discussion with upstream
revealed that 0.46 works best with the forked OSG 3.4, 3.6 is not
recommended for it. But 0.47 needs 3.6, also with a preference to use
the forked OSG.


thesamesam:
This needs a revbump to enforce the dep.

DarthGandalf:
Done

PR #21138, opened 2021-06-05
sys-libs/musl: fix HOMEPAGE

bjacquin:
As per http://www.musl-libc.org/, "musl has moved to a new domain:
musl.libc.org"

Bug: https://bugs.gentoo.org/794484
Signed-off-by: Bertrand Jacquin <bertrand@jacquin.bzh>


trofi:
Pushed as is. Thank you!

PR #21145, opened 2021-06-06
app-misc/OpenRGB

DarthGandalf:


app-misc/OpenRGB/OpenRGB-0.6.ebuild
```diff
@@ -12,44 +12,46 @@ else
...
+	eqmake5 INCLUDEPATH+="${ESYSROOT}/usr/include/nlohmann"
 }
 
 src_install() {
 	emake INSTALL_ROOT="${ED}" install
 
 	dodoc README.md OpenRGB.patch
-
-	if use udev; then
-		udev_dorules 60-openrgb.rules
```
chutz, line 53: Why are we not installing the udev rules anymore? AFAICS the rules file is still in the repository, and the upstream instructions seem to suggest they are still needed.

DarthGandalf: It's installed by "make install" already

PR #21147, opened 2021-06-07
net-misc/youtube-dl: Inherit git-r3 for live ebuild

zippy2:
The interit of git-r3 eclass was mistakenly removed in
a27a5dfb92de4aed3922446d5521aaf0cafc475e. Without this,
the ebuild fails to fetch and unpack the source.


zippy2:
CC @mgorny

mkszuba:
Good catch, thanks. Just make sure you preserve the alphabetical order of eclass names in the future.

PR #21156, opened 2021-06-08
Zfs rebuild triggers

gyakovlev:


gyakovlev:
it seems I've figured out a proper way to trigger portage to suggest enabling `dist-kernel` if only 1 package of 2 has it set, using `PDEPEND` with use conditional in full form.

also adding subslots to zfs-kmod, and subslot to zfs.
needed to trigger rebuilds if only 1 package got a revbump, to keep things in sync nicely.
sometimes files in /usr/src/zfs* get updated/patched, and we want userspace to pick it up.

it's actually not fragile at all and 0.8 works fine with 2.0 pools. but consistency is the key to robustness.

sys-fs/zfs-kmod/zfs-kmod-9999.ebuild
```diff
@@ -43,10 +43,25 @@ if [[ ${PV} != "9999" ]] ; then
...
 DOCS=( AUTHORS COPYRIGHT META README.md )
 
+pkg_pretend() {
+	use rootfs || return 0
+	use dist-kernel || return 0
+	if has_version virtual/dist-kernel; then
+		ewarn "You have virtual/dist-kernel installed, but"
+		ewarn "USE=\"dist-kernel\" is not enabled for ${CATEGORY}/${PN}"
+		ewarn "It's recommended to globally enable dist-kernel USE flag"
+		ewarn "to auto-trigger initrd rebuils with kernel updates"
```
thesamesam, line 61:
```suggestion
		ewarn "to auto-trigger initrd rebuilds with kernel updates"
```

sys-fs/zfs-kmod/zfs-kmod-9999.ebuild
```diff
@@ -43,10 +43,25 @@ if [[ ${PV} != "9999" ]] ; then
...
+PDEPEND="dist-kernel? ( ~sys-fs/zfs-${PV}[dist-kernel] )"
+
 RESTRICT="debug? ( strip ) test"
 
 DOCS=( AUTHORS COPYRIGHT META README.md )
 
+pkg_pretend() {
+	use rootfs || return 0
+	use dist-kernel || return 0
+	if has_version virtual/dist-kernel; then
```
thesamesam, line 57:
```suggestion

	if has_version virtual/dist-kernel; then
```

thesamesam (approved):
Looks good!

gyakovlev:
merged in https://github.com/gentoo/gentoo/commit/80c9b77b4b7dd9355a28b7067f5d84687775beb6 and around that.

PR #21162, opened 2021-06-08
Bump20210606

thebitpit:
Use newer mythtv fixes/31 branch to fix bug 794838
Add the Python 3.9 target


thebitpit:
Throw this away and try again. I can not figure out what is not signed.

PR #21171, opened 2021-06-09
eclass/kernel-2: fix USE=symlink

tillschaefer:
the symlink to /usr/src/linux is longer removed bofore linking
(see 7446cb850c5e0412e4d88b3aa5d207751241fb02). the latter ln -sf
does not replace existing symlinks, since destination is a
directory.

Closes: https://bugs.gentoo.org/795057
Signed-off-by: Till Schäfer <till2.schaefer@uni-dortmund.de>


mjeveritt (approved):
Correct! Good catch!

mpagano:
Thanks for your work

PR #21173, opened 2021-06-09
dev-java/jna: no longer fail with jdk-11

vaukai:

### DIFF
--- a/dev-java/jna/jna-4.2.2-r2.ebuild
+++ b/dev-java/jna/jna-4.2.2-r2.ebuild
@@ -1,0 +1,87 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+JAVA_PKG_IUSE="doc source"
+MAVEN_ID="net.java.dev.jna:jna:4.2.2"
+
+inherit java-pkg-2 java-ant-2 toolchain-funcs
+
+DESCRIPTION="Java Native Access (JNA)"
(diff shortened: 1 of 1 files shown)


dev-java/jna/jna-4.2.2-r2.ebuild
```diff
@@ -0,0 +1,89 @@
...
+# fine. Someone bisect it please. :)
+RESTRICT="test"
+
+CDEPEND="dev-libs/libffi"
+
+RDEPEND="${CDEPEND}
+	virtual/jre:1.8"
+
+DEPEND="${CDEPEND}
+	virtual/jdk:1.8
```
fordfrog, line 34: the message says it does not fail with jdk-11 but the ebuild is restricted to `jdk:1.8`, i'm confused :-) could you please explain that?

vaukai: > the message says it does not fail with jdk-11 but the ebuild is restricted to `jdk:1.8`, i'm confused :-) could you please explain that?

It was unlimited `>=virtual/jdk-1.7` before, causing it to fail with jdk-11.  And setting `-1.8:*` didn't work.
Will fix that soon ...

fordfrog: ah, so it does not fail with jdk-11 because it does not compile with jdk-11 anymore, got that :-)

fordfrog:
@vaukai thanks! also dropped old.

PR #21175, opened 2021-06-09
dev-lang/zig: version bump to 0.8.0

strikersh:


juippis:
@epsilon-0 any comments?

epsilon-0:
> @epsilon-0 any comments? 

Ah, I hadn't seen this one. The other PR also enabled tests so I believe we can use that one.

PR #21186, opened 2021-06-10
net-mail/queue-repair: Allowed for python 3.9 compatability

Ralphred1:
Bug: https://bugs.gentoo.org/793212
Signed-off-by: Andrew Foster <gentoothings@liquid.me.uk>

### DIFF
--- a/net-mail/queue-repair/queue-repair-0.9.0-r2.ebuild
+++ b/net-mail/queue-repair/queue-repair-0.9.0-r2.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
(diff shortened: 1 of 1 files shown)


DerDakon:
When you are at it anyway, could you check if this works with 3.10, too? I don't expect much problems, but just to be sure.

Ralphred1:
> When you are at it anyway, could you check if this works with 3.10, too?
The 3.10 porting section is smaller, I'll put it on my todo list but it'll be a couple of days as I add the 3.10 changes to my "will it work with 3.9" scripts.

PR #21191, opened 2021-06-10
net-libs/libmicrohttpd: fixed testing

Karlson2k:
No revbump as only test was fixed.

Closes: https://bugs.gentoo.org/795105


thesamesam:
Huge thanks for working on this so quickly!

PR #21139, opened 2021-06-06
dev-java/fontbox: bump to 2.0.24 | dev-java/pdfbox: bump to 2.0.24

vaukai:
These version bumps are precondition for https://bugs.gentoo.org/794241.

All `bouncycastle` bumps are moved to slot 0 and dependencies changed from slot to version like
```
dev-java/bcprov:${SLOT}
```
to
```
~dev-java/bcprov-${PV}:0
```

PR #21197, opened 2021-06-11
dev-java/ant-ivy: bump to 2.5.0 | BC_SLOT="1.50" removal

vaukai:
This `dev-java/ant-ivy` version bump is precondition for [`BC_SLOT="1.50"` removal](https://bugs.gentoo.org/717950).
I didn't want to overload the [pdfbox PR](https://github.com/gentoo/gentoo/pull/21139), so those 5 bouncy castle  commits are the same as posted there.


fordfrog:
i at least merged those `MAVEN_ID` commits to make the list a bit shorter. i will probably have process this piece by piece.

vaukai:
> [...] i will probably have process this piece by piece.

guess those dev-java/bc... bumps can only be done all four together and would solve the first part of https://github.com/gentoo/gentoo/pull/21139

fordfrog:
@vaukai `dev-java/httpcore` merged, thanks!

fordfrog:
@vaukai `dev-java/commons-httpclient` merged, thanks!

fordfrog:
i looked at the sources of `dev-java/jna-platform` and it seems it provides some tests. are these for this package? or were there any issues with adding support for testing in this package?

vaukai:
> i looked at the sources of `dev-java/jna-platform` and it seems it provides some tests. are these for this package? or were there any issues with adding support for testing in this package?

Did you mean those in `/var/tmp/portage/dev-java/jna-platform-4.2.2/work/jna-4.2.2/contrib/platform/test`?  Guess I didn't even try while being focused on `ant-ivy`. 
Also how do you think about integrating `jna-platform` into `dev-java/jna`, adding a `platform` USE flag.? It's anyway coming from the same source tarball ...

fordfrog:
well, the stuff from maven repo is usually stripped down afaics. wrt `jna-platform` and `jna`, i would keep them separate. they would even have different `MAVEN_ID` so for the purpose of `java-ebuilder` this way it's better. i suppose the latest version of jna is not compatible with `jsch-agent-proxy`, is it? i mean does it compile against it or not? also, `jsch-agent-proxy` provides its jars under different artifacts, so again, for `java-ebuilder` it would be better for us to split the package. what do you think?

vaukai:
> i suppose the latest version of jna is not compatible with `jsch-agent-proxy`, is it? 

Did you mean [5.8.0](https://github.com/java-native-access/jna/releases/tag/5.8.0)? Haven't tried since it's not in the tree. Is there any problem with [4.2.2-r3](https://packages.gentoo.org/packages/dev-java/jna)?

> i mean does it compile against it or not? also, `jsch-agent-proxy` provides its jars under different artifacts, so again, for `java-ebuilder` it would be better for us to split the package. what do you think?

Here I kind of dislike the idea of maintaining 7 widely identical `ebuilds` that all have the same and only one single consumer. 
However, a proper handling of `javadocs` and `source` still needs to be added.

fordfrog:
yes, i mean the latest version. sooner or later we'll have to bump it anyway.

wrt `jsch-agent-proxy`, i already thought about your argument before, but the project seems to be dead so it would be done once and for all it seems. though currently it really has just a single consumer, and i hope they rather get rid of it :-)

vaukai:
> i looked at the sources of `dev-java/jna-platform` and it seems it provides some tests. are these for this package? or were there any issues with adding support for testing in this package?

Now trying with tests enabled ...
```
 * Compiling ...
contrib/platform/test/com/sun/jna/platform/StructureFieldOrderTest.java:21: error: cannot find symbol
import com.sun.jna.StructureFieldOrderInspector;
                  ^
  symbol:   class StructureFieldOrderInspector
  location: package com.sun.jna
contrib/platform/test/com/sun/jna/platform/StructureFieldOrderTest.java:59: error: cannot find symbol
        StructureFieldOrderInspector.batchCheckStructureGetFieldOrder(FileUtils.class, ignoreConstructorError);
        ^
  symbol:   variable StructureFieldOrderInspector
  location: class StructureFieldOrderTest
Note: contrib/platform/test/com/sun/jna/platform/win32/Kernel32Test.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
2 errors
```
Needs more work ...

fordfrog:
there seems to be a problem with `dev-java/jsch-agent-proxy` wrt generating documentation, as for each compilation run the documentation is generated over the previous documentation so the documentation is far from complete.

vaukai:
> there seems to be a problem with `dev-java/jsch-agent-proxy` wrt generating documentation, as for each compilation run the documentation is generated over the previous documentation so the documentation is far from complete.

It's a regression. Doesn't even compile here ...

fordfrog:
there are two options how to fix it. you can either split the package to separate packages, one for each jar, or you can ignore the generated javadoc in the existing ebuild, delete the javadoc at the end of compilation, move all sources to a single dir, and then generate the javadoc from there using `ejavadoc`.

vaukai:
> there are two options how to fix it. you can either split the package to separate packages, one for each jar, or you can ignore the generated javadoc in the existing ebuild, delete the javadoc at the end of compilation, move all sources to a single dir, and then generate the javadoc from there using `ejavadoc`.

Will try after solving some other trouble ...

vaukai:
Several tests excluded ...

vaukai:
> there are two options how to fix it. you can either split the package to separate packages, one for each jar, or you can ignore the generated javadoc in the existing ebuild, delete the javadoc at the end of compilation, move all sources to a single dir, and then generate the javadoc from there using `ejavadoc`.

Solved this with another `java-pkg-simple_src_compile`:
```
	JAVA_SRC_DIR=(
		"${PN}-core"
		"${PN}-jsch"
		"${PN}-pageant"
		"${PN}-sshagent"
		"${PN}-usocket-jna"
		"${PN}-usocket-nc"
		"${PN}-connector-factory"
	)
	JAVA_JAR_FILENAME="ignoreme.jar"
	java-pkg-simple_src_compile
```
and installation like
```
	if use doc; then
		java-pkg_dojavadoc target/api
	fi
	if use source; then
		java-pkg_dosrc "${PN}-core" "${PN}-jsch" "${PN}-pageant" "${PN}-sshagent" "${PN}-usocket-jna" "${PN}-usocket-nc" "${PN}-connector-factory"
	fi
```

fordfrog:
this looks like an elegant solution :-)

fordfrog:
@vaukai `jsch-agent-proxy` merged, thanks!

vaukai:
And there are still [98 test failures](https://gist.github.com/vaukai/ae7d738bd83f648372abc7d25f047cb2).

fordfrog:
wrt the failing tests, i guess we could start with those missing files

vaukai:
For the sake of [several CVEs](https://bugs.gentoo.org/showdependencytree.cgi?id=794598&hide_resolved=1): READY for review.

PR #21200, opened 2021-06-11
app-misc/anki: patch Python 3.9 incompatibility

Austin-Ray:
This commit modifies the existing anki-2.1.15 ebuild with a patch to
replace a deprecated, undocumented method available in Python 3.8[0] but
removed in Python 3.9[1]. Without the patch, anki throws a runtime error
(see bug for more details) as anki 2.1.15 pre-dates Python 3.9.

[0] https://github.com/python/cpython/blob/4844abdd700120120fc76c29d911bcb547700baf/Lib/html/parser.py#L466
[1] https://bugs.python.org/issue37328

Closes: https://bugs.gentoo.org/795309

Signed-off-by: Austin Ray <austin@austinray.io>

### DIFF
--- a/app-misc/anki/anki-2.1.15-r1.ebuild
+++ b/app-misc/anki/anki-2.1.15-r1.ebuild
@@ -44,6 +44,7 @@ BDEPEND="test? (
 PATCHES=(
 	"${FILESDIR}"/${PN}-2.1.0_beta25-web-folder.patch
 	"${FILESDIR}"/${PN}-2.1.15-mpv-args.patch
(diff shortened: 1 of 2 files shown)


thesamesam:
Done, thank you!

(Help needed on bumping Anki to something newer, btw.)

Austin-Ray:
> (Help needed on bumping Anki to something newer, btw.)

I'm willing to help with that. Do you have any resources for working with Gentoo packages (my changes have all been minor)? Where's a good place to ask for help if I get stuck?

PR #21209, opened 2021-06-12
dev-cpp/pystring: Don't use hardcoded g++, bug #795168

DarkDefender:
Convert the project into cmake so we get compiler switching for free.


thesamesam:
And fixes that pesky libtool-related issue too!

DarkDefender:
Ah, should I add that to the commit message as well then?
(The libtool bug id)

eli-schwartz:
By the way, similar to https://github.com/gentoo/gentoo/pull/39761#issuecomment-2712186093

This switched to a custom hand-brewed build system that upstream never accepted, which set the library soname to "0.0" -- libtool used "0".

Now in 2024 it is using no soname version at all. That's three different subslots we *should* have used.

PR #21215, opened 2021-06-12
app-editors/vim: add flag ipv6 for 9999

vitaly-zdanevich:


thesamesam:
See https://bugs.gentoo.org/790446#c5.

PR #21217, opened 2021-06-13
net-analyzer/neti: min java 1.8

vaukai:
Bug: https://bugs.gentoo.org/787449

### DIFF
--- a/net-analyzer/neti/neti-2.0-r3.ebuild
+++ b/net-analyzer/neti/neti-2.0-r3.ebuild
@@ -1,0 +1,55 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DESCRIPTION="NETI@Home research project from GATech"
+HOMEPAGE="http://www.neti.gatech.edu"
+SRC_URI="mirror://sourceforge/project/neti/neti/${PV}/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~ppc ~x86"
+IUSE="zlib java"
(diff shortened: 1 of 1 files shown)


net-analyzer/neti/neti-2.0-r3.ebuild
```diff
@@ -0,0 +1,55 @@
...
+HOMEPAGE="http://www.neti.gatech.edu"
+SRC_URI="mirror://sourceforge/project/neti/neti/${PV}/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~ppc ~x86"
+IUSE="zlib java"
+
+DEPEND="
+	java? ( || ( >=virtual/jdk-1.8:* >=virtual/jre-1.8:* ) )
```
fordfrog, line 16: this looks weird to me. if it compiles any java classes, it should depend on jdk. wrt jre, why would it need just runtime for compilation?

vaukai: > this looks weird to me. if it compiles any java classes, it should depend on jdk. wrt jre, why would it need just runtime for compilation?

But you don't really expect me to answer that question ... [It's always been that way](https://gitweb.gentoo.org/repo/gentoo.git/commit/net-analyzer/neti?id=56bd759df1d0c750a065b8c845e93d5dfa6b549d)
[Latest release was in 2005](https://sourceforge.net/projects/neti/files/neti/) and package maintainers seem not to care.

fordfrog: we'll see whether maintainers will react. if not, i'll check it when i get to it.

PR #21225, opened 2021-06-13
dev-java/miglayout: remove doc flag (broken)

vaukai:
Closes: https://bugs.gentoo.org/786732

No fix available, 
vbump might be possible later ( https://github.com/mikaelgrev/miglayout )


fordfrog:
@vaukai thanks!

PR #21226, opened 2021-06-13
dev-db/hsqldb: min java 1.8

vaukai:
fails with 1.8:*
Bug: https://bugs.gentoo.org/784380

No version bump here.  I am working on a bump to 2.6.0 which doesn't support java8. For java 8 we can keep the present version.
Also tried building [2.5.2](https://sourceforge.net/p/hsqldb/news/2021/04/maven-releases-of-version-260/) for `java 8` but with no success.

Help needed for GLEP81

### DIFF
--- a/acct-group/hsqldb/hsqldb-0.ebuild
+++ b/acct-group/hsqldb/hsqldb-0.ebuild
@@ -1,0 +1,9 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit acct-group
+
+DESCRIPTION="Group for hsqldb"
+ACCT_GROUP_ID=217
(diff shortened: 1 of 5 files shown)


fordfrog:
creating users and groups (glep 81) is described [here](https://devmanual.gentoo.org/ebuild-writing/users-and-groups/). if you need more assistance, just let me know.

vaukai:
> creating users and groups (glep 81) is described [here](https://devmanual.gentoo.org/ebuild-writing/users-and-groups/). if you need more assistance, just let me know.

Let's start with https://github.com/gentoo/api-gentoo-org/pull/403

fordfrog:
@vaukai thanks! i also switched `hsqldb` to the latest compatible `tomcat-servlet-api`.

PR #20710, opened 2021-05-07
app-editors/vscode: New package

AdelKS:
Hello,

I am here to submit, then maintain, the vscode EBUILD. 

Two things are left to do:
- Add the Microsoft VSCode license to the list of licenses: https://code.visualstudio.com/License Should I open a bug report to request it ?
- I am still hesitating on whether add a "-bin" suffix  or not. Input welcome!

I picked up this EBUILD from the GURU overlay and my knowledge is still quite limited, I am willing to learn ! Any feedback is welcome.

Adel

PR #21227, opened 2021-06-13
app-editors/vscode app-editors/vscodium: version bump to 1.57.0

AdelKS:
Hello @juippis 

A little version bump, they both install and lunch just fine. Is there a quick way to check if a new binary EBUILD brings in new files with respect to an older version ?

There's a warning message with the Jupyter extension but that's something upstream (same issue happens in Arch's versions). If anything wrong needs to be fixed, I will open another PR.


dakyskye:
Been waiting for this PR for days, it's getting little bit annoying VSCode won't stop suggesting me to update, hahah.

AdelKS:
I opened this PR the day after I got the notification ! I will do my best to be on time :D

juippis (approved):
> A little version bump, they both install and lunch just fine. Is there a quick way to check if a new binary EBUILD brings in new files with respect to an older version ?

You might have some luck with *pkgdiff* from https://github.com/mgorny/mgorny-dev-scripts - but I doubt it'll work with binary packages...

And to the wait-time: That was one thing I told you to consider when bringing this over from GURU. Realistically, you can't ever expect us to merge these within 3 days of opening PRs. 

(But these packages are a good candidate for my upcoming autogen-project)

AdelKS:
> You might have some luck with pkgdiff from https://github.com/mgorny/mgorny-dev-scripts - but I doubt it'll work with binary packages...

Thanks! Will give it a try

> And to the wait-time: That was one thing I told you to consider when bringing this over from GURU. Realistically, you can't ever expect us to merge these within 3 days of opening PRs.

It's not me, it's our dear Gentoo users who can't wait to be on the bleeding edge (Just like me xD) and take all the new bugs in the face haha. I have the update before everyone already :sunglasses: But don't worry, I will soon become a Gentoo dev/maintainer then I will be able to push these small updates myself (I suppose ?)

Thanks! See ya for the next update, probably next month. Except if there's some packaging bugs to fix.

Adel

dakyskye:
Wait, does GURU receive package updates earlier?

AdelKS:
The time it takes to approve a change from a maintainer is very short, usually the same day : the QA check is not a blocker, only a superficial check for malicious intent is done by a trusted contributor before approval, and the ratio N(trusted contributors) / N(packages) in GURU is most certainly bigger. So yes, Updates in `::gentoo` are currently slower than `::guru`. I have the vast impression that the number of maintainers in Gentoo is relatively tiny for the amount of work needed.

dashezup:
@AdelKS `vscodium` 1.57.0 does not start when I run `vscodium` from a terminal emulator. (I need to launch it from CLI because I'm a WM user and I can't start application with `.desktop` file)

> $ vscodium --verbose
> [3397:0701/102055.441246:FATAL:setuid_sandbox_host.cc(158)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /opt/vscodium/chrome-sandbox is owned by root and has mode 4755.

Workaround (any):
- `/optc/vscodium/chrome-sandbox` comes with permission, change it to 4755 and it will work
- `vscodium --no-sandbox`

btw seems it works in wayland as well, with `--enable-features=UseOzonePlatform --ozone-platform=wayland` and one of the workarounds.

Related issues:
- https://github.com/microsoft/vscode/issues/81056#issuecomment-532753469
- https://github.com/microsoft/vscode/issues/126027 (especially this one)
- https://github.com/VSCodium/vscodium/issues/604#issuecomment-789477784

TL; DR sandbox support is a wip feature in vscode, they dropped `--no-sandbox`, vscode does not start for lots of users and they added it back.

The `.desktop` files in this repo does not include `--no-sandbox` argument, maybe you should add it for now.

I wonder did you face this issue as well, why it works for you if not. I wonder how to use the sandbox feature either, I'll just chmod 4755 chrome-sandbox for now.

AdelKS:
Hello @dashezup,

Thank you for your feedback! The original EBUILDs I started from were using actually `--no-sandbox`, while I was submitting the first EBUILD for `::gentoo`, [I got asked about it](https://github.com/gentoo/gentoo/pull/20710#discussion_r628705789) and found [this merged PR](https://github.com/microsoft/vscode/pull/81096) that seems to enable `--no-sandbox` anyway even if it's not specified. So I decided to drop it, so if in the future, they start supporting sandboxing again it would work as-is or some issues will arise. It looks like it's the latter option is happening after they dropped the embedded `--no-sandbox`

> `/optc/vscodium/chrome-sandbox` comes with permission, change it to 4755 and it will work

I have zero knowledge on sandboxing, is this equivalent to using `--no-sandbox` or is it the correct approach to sandboxing ? Because if it's the correct approach I will prefer it. I will read further on the matter.

> I wonder did you face this issue as well, why it works for you if not. I wonder how to use the sandbox feature either, I'll just chmod 4755 chrome-sandbox for now.

I did not face any issue and it just works:tm:, I wonder what makes it different for other people.

dashezup:
> I have zero knowledge on sandboxing, is this equivalent to using --no-sandbox or is it the correct approach to sandboxing?

I just read this, seems setuid should be avoided.

https://chromium.googlesource.com/chromium/src/+/HEAD/docs/linux/sandboxing.md#user-namespaces-sandbox
> The namespace sandbox aims to replace the setuid sandbox. It has the advantage of not requiring a setuid binary. It's based on (unprivileged) user namespaces in the Linux kernel. It generally requires a kernel >= 3.10, although it may work with 3.8 if certain patches are backported.

 > Give a try to 1.57.1?

I just tired, it 1.57.1 works here without setuid. It surprised me as well, I've double-checked, it works on 1.57.1 but not on 1.57.0.

It must be vscode/vscodium fix it on 1.57.1, so maybe just forget about the `--no-sandbox` thing.

> PS: thanks for the wayland tip, I will make a new .desktop entry with those extra arguments and test it out. Some people are running pure wayland without any xwayland and they would benefit from this if it will actually works!

Maybe not add to the package, one more icon may make it confusing. I read that from Arch Wiki, maybe others would also read it there if they want to use it in wayland-only mode.

AdelKS:
> it works on 1.57.1 
Great! Yea that's probably one of the fixes that come with 1.57.1, since that version came out quite fast.

> Maybe not add to the package, one more icon may make it confusing. I read that from Arch Wiki, maybe others would also read it there if they want to use it in wayland-only mode.

Well, Firefox uses the same approach. But running any .desktop file actually runs a bash script that decides which arguments to use based on if it's Wayland or not.

PR #21237, opened 2021-06-14
net-firewall/nfacct: add init script

zuckschwerdt:
Save and restore nfacct accounting objects.

Signed-off-by: Christian W. Zuckschwerdt <christian@zuckschwerdt.org>

### DIFF
--- a/net-firewall/nfacct/files/nfacct.confd
+++ b/net-firewall/nfacct/files/nfacct.confd
@@ -1,0 +1,7 @@
+# /etc/conf.d/nfacct
+
+# Location for nfacct initscript to save and restore the counters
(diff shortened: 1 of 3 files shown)


zuckschwerdt:
This adds an initscript to save and restore nfacct accounting objects.
All accounting objects need to be present before iptables can be started/restored if the nfacct module is used in some iptables rules.
E.g. `nfacct add ipv4` otherwise `iptables -I INPUT -m nfacct --nfacct-name ipv4` fails.

The confd and initd is largely copied from iptables and ipset.

zuckschwerdt:
Sorry to bump this, anything else needed? Is there a process or someone to ping?

thesamesam:
I absolutely forgot about it, thank you for pinging!

PR #21239, opened 2021-06-14
fcaps.eclass: EAPI 8

dm0-:
@floppym This updates fcaps.eclass to EAPI 8 so it can finally correctly define its dependency with `IDEPEND`.  In particular, this will fix a dependency loop with PAM.

The list of supported EAPIs was being added to eclasses a while ago, but this one was skipped, so I added it here.  I can drop it if it's preferable for this eclass to be EAPI-agnostic.  Since there were only two ebuilds using EAPI 5, this updates those and declares eclass compatibility with EAPIs 6-8 (and takes advantage of bash 4 syntax that comes with that).

This should probably wait to be merged after the stalled stablereq [bug #756946](https://bugs.gentoo.org/756946) so all arches have a stable EAPI 7 arping version.  There is also [bug #632225](https://bugs.gentoo.org/632225) to replace intel-performance-counter-monitor, but it can be trivially updated until then.

### DIFF
--- a/eclass/fcaps.eclass
+++ b/eclass/fcaps.eclass
@@ -1,9 +1,10 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 # @ECLASS: fcaps.eclass
 # @MAINTAINER:
 # base-system@gentoo.org
+# @SUPPORTED_EAPIS: 6 7 8
 # @BLURB: function to set POSIX file-based capabilities
 # @DESCRIPTION:
 # This eclass provides a function to set file-based capabilities on binaries.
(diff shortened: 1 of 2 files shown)


eclass/fcaps.eclass
```diff
@@ -33,15 +34,12 @@ _FCAPS_ECLASS=1
...
+# Since it is needed in pkg_postinst() it must be in IDEPEND
 case "${EAPI:-0}" in
-	[0-6])
-		RDEPEND="filecaps? ( sys-libs/libcap )"
-	;;
-	*)
-		BDEPEND="filecaps? ( sys-libs/libcap )"
-		RDEPEND="${BDEPEND}"
-	;;
+	7) BDEPEND="filecaps? ( sys-libs/libcap )" ;&
```
floppym, line 39: I have never seen `;&` before. That's neat!

floppym (approved):
Looks good to me.

eclass/fcaps.eclass
mgorny: But this means the eclass will throw a syntax error on bash-3.2, instead of giving unsupported EAPI die.

I think the best way would be to split the `case` into a more traditional ebuild thingie, i.e.:

```
case ${EAPI} in
    [6-8]) ;;
    *) die "EAPI ${EAPI:-0} is unsupported";;
esac

case ${EAPI} in
    7) ...;&
    6) ...;;
    *) ...;;
esac
```

SoapGentoo: also, put the whitelist before the inherit guard, and the `IDEPEND` assignment inside the inherit guard. Finally, maybe consider

```sh
case ${EAPI} in
    7) ...;& # fallthrough
    6) ...;;
    *) ...;;
esac
```

so people don't get triggered?

PR #21241, opened 2021-06-14
dev-libs/tvision: bump to 2.2.3

waebbl:
dev-libs/tvision: drop 2.1.0_pre2
Cleanup old

Closes: https://bugs.gentoo.org/760576

dev-libs/tivison: bump to 2.2.3
- Dependency on gpm, X and allegro is automagic, so I made it mandatory
  for now. See set-soft/tvision#1. Drop gpm
  USE flag for this.
- Add support for media-libs/allegro
- Add USE flag to install examples for testing
- Drop no-op debug USE flag

Closes: https://bugs.gentoo.org/730022


dev-libs/tvision/tvision-2.2.3.ebuild
```diff
@@ -0,0 +1,81 @@
...
+		--without-static
+		--x-include="${EPREFIX}/usr/include/X11"
+	)
+
+	use nls || myconf+=( --no-intl )
+
+	# Note: Do not use econf here, this isn't an autoconf configure script,
+	# but a perl based script which simply calls config.pl
+	einfo "Running ./configure ${myconf[@]}"
+	./configure ${myconf[@]} || die
```
thesamesam, line 66: You may want to add ${EXTRA_ECONF} here to be nice but it's not required

dev-libs/tvision/tvision-2.2.3.ebuild
```diff
@@ -0,0 +1,81 @@
...
+PATCHES=(
+	"${FILESDIR}"/${PN}-2.2.1.4-flags.patch
+	"${FILESDIR}"/${PN}-2.2.1.4-ldconfig.patch
+	"${FILESDIR}"/${P}-0001-use-proper-AR.patch
+)
+
+src_configure() {
+	tc-export CC CXX AR
+
+	local myconf=()
```
thesamesam, line 52:
```suggestion
	local myconf=(
		--fhs
		...
	)
```

dev-libs/tvision/tvision-2.2.3.ebuild
waebbl: Ah yes, I can do this the next time, an update is needed. Although there are not many options left which the user might set independantly.

PR #21242, opened 2021-06-14
media-gfx/freecad: add python-3.9 support

waebbl:
Also restrict dependency for opencascade

Closes: https://bugs.gentoo.org/789780

I will open separate PR's for building against opencascade-7.5.2 (once it is merged), which I already tested and get it working, and building against vtk-9, which has still to be addressed issues.


fordfrog:
i suppose `freecad-9999` should also receive `py3_9` support, or not?

waebbl:
Uh, yes absolutely. Darn, I forgot to add this

PR #21252, opened 2021-06-15
sci-libs/vtk: update dependency on dev-python/sip, remove old

waebbl:
drop 8.2.0{-r{1,2}}

Use slotted dev-python/sip package
Remove obsolete commented cmake variable


thesamesam:
I'd squash the second 'update dependency' commit

waebbl:
Should I drop the `-r1` and `-r2` too then, and only push an `-r3`

thesamesam:
I usually wimp out and do r1 -> r3 and r2 -> r4, but I don't see a point in keeping old versions around here actually, so go for it

waebbl:
I was thinking about dropping r1 anyway, and if a revbump is needed it makes even more sense :)

PR #21253, opened 2021-06-15
criterion/boxfort - bump and add py3.10

hydrapolic:
Tests pass on ~amd64, used it to test syslog-ng, it passed fine.


hydrapolic:
Thanks @thesamesam

PR #21262, opened 2021-06-16
Dictd usr share

juippis:


juippis:
Merged with CI fix, and reverted --libexecdir change.

PR #21272, opened 2021-06-16
games-emulation/hatari: remove unused patches

mm1ke:


ionenwks:
Thanks, I knew I may have forgot some of these in the last cleanup spree, went over but looks like I forgot hatari.

PR #21274, opened 2021-06-16
sys-cluster/lmod: version bump to 8.5.3

epsilon-0:


sys-cluster/lmod/lmod-8.5.3.ebuild
```diff
@@ -0,0 +1,130 @@
...
+RDEPEND="${LUA_DEPS}
+	dev-lang/tcl
+	dev-lang/tk
+	$(lua_gen_cond_dep '
+		>=dev-lua/luafilesystem-1.8.0[${LUA_USEDEP}]
+		dev-lua/luajson[${LUA_USEDEP}]
+		dev-lua/luaposix[${LUA_USEDEP}]
+		dev-lua/lua-term[${LUA_USEDEP}]
+	')
+	virtual/pkgconfig
```
AndrewAmmerlaan, line 36: pkgconfig as an RDEPEND? Are you sure that is correct?

amadio: The ebuild is identical to the previous version, so I've merged this.
I will leave it for the maintainer (@zlogene) to review the pkgconfig dependency, as it was there before.

PR #21275, opened 2021-06-16
sci-mathematics/cgal: version bump to 5.2.2

epsilon-0:
also add shared USE flag


juippis (commented):
Should the shared lib be enabled unconditionally?

epsilon-0:
> Should the shared lib be enabled unconditionally?

Upstream says that this is not enabled by default and maybe deprecated later. So I don't think we should.

juippis (approved):

PR #21278, opened 2021-06-17
dev-qt/qt-creator: Bump to 4.15.1

hansfn:
The dev-libs/libclangformat-ide package is no longer needed. Synced
qt-creator-9999.ebuild and qt-creator-4.15.1.ebuild so the latest
changes from both files are used.

Closes: https://bugs.gentoo.org/750527


AdelKS:
Hello! thanks for this PR.

I gave a try to this version, clang code model works great! The issue I am having is that debugging doesn't seem to be working, I get a "command aborted" popup, using GDB directly works, can you test on your end on a simple hello world C++ and put a break point before the program ends ? Maybe it's only an issue with my machine. For an unknown reason, I cannot downgrade back to the original qt-creator version either x)

hansfn:
I tested on a small, real project and debugging worked. Could it be that your gdb isn't updated? I'm using sys-devel/gdb-10.2 (which no longer has the client use flag). The ebuild doesn't really change anything related to debugging so if it didn't work, it would have been an upstream bug.

dev-qt/qt-creator/qt-creator-9999.ebuild
```diff
@@ -115,15 +114,8 @@ done
 unset x
 
 PATCHES=(
-	"${FILESDIR}"/${PN}-4.12.0-dylib-fix.patch
-	"${FILESDIR}"/${PN}-4.12.0-libclangformat-ide.patch
 )
 
-llvm_check_deps() {
-	has_version -d "sys-devel/clang:${LLVM_SLOT}" && \
```
Pesa, line 123: why did you delete the check on sys-devel/clang?

hansfn: I don't really mind keeping the sys-devel/clang check, but I didn't see any purpose of the check when we removed dev-libs/libclangformat. In other words, I failed to see why Qt Creator needs this check when it only covers clang. I thought llvm and clang would be kept in sync without Qt Creator forcing it, but I might very well be wrong. Sorry.

hansfn: Just let me know if you want the PR updated with the `llvm_check_deps()` function preserved (without the libs/libclangformat check).

PR #21282, opened 2021-06-17
media-fonts/joypixels: add licensing postinst

epsilon-0:
Closes: https://bugs.gentoo.org/796461

### DIFF
--- a/media-fonts/joypixels/joypixels-6.5.0.ebuild
+++ b/media-fonts/joypixels/joypixels-6.5.0.ebuild
@@ -4,7 +4,7 @@
 EAPI=7
 
 FONT_SUFFIX="ttf"
-inherit font
+inherit font readme.gentoo-r1
 
 DESCRIPTION="A font for better emoji and unicode support"
 HOMEPAGE="https://www.joypixels.com/"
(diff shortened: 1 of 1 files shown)


media-fonts/joypixels/joypixels-6.5.0.ebuild
```diff
@@ -20,7 +20,16 @@ RESTRICT="bindist mirror"
...
+
 src_prepare() {
 	default
 	cp "${DISTDIR}"/${P}.ttf "${S}"/${P}.ttf || die
 }
+
+src_install() {
+	font_src_install
+	readme.gentoo_create_doc
+}
```
juippis, line 35:
```suggestion
}

pkg_postinst() {
	readme.gentoo_print_elog
}
```
(I thought it'd be on by default, but it's not)
I'll add it, thanks!

juippis (approved):

media-fonts/joypixels/joypixels-6.5.0.ebuild
epsilon-0: thanks!

PR #21285, opened 2021-06-17
dev-lang/clojure: add 1.10.3

Flowdalic:
This is a "new kind" of clojure ebuilds, which does not longer depend
on dev-java/spec-alpha and dev-java/core-specs-alpha, but instead is
able to bootstrap itself. Major thanks to Göktürk Yüksek for coming up
with a way to bootstrap clojure (see
add-compile-spec-ant-build-target.patch).

Closes: https://bugs.gentoo.org/661442


dev-lang/clojure/clojure-1.10.3.ebuild
```diff
@@ -0,0 +1,108 @@
...
+)
+
+RESTRICT="!test? ( test )"
+
+COMMON_DEPEND="
+	dev-java/ant-core:0
+"
+
+RDEPEND="
+	${COMMON_DEPEND}
```
gktrk, line 52: Why do we need ant for runtime?

dev-lang/clojure/files/add-compile-spec-ant-build-target.patch
```diff
@@ -0,0 +1,54 @@
...
++++ b/build.xml
+@@ -85,6 +85,24 @@
+     </java>
+   </target>
+ 
++  <target name="compile-spec"
++          description="Compile Clojure spec sources."
++          depends="compile-java">
++    <java classname="clojure.lang.Compile"
++          classpath="${maven.compile.classpath}:${build}:${cljsrc}"
```
gktrk, line 26: Do you wanna leave `maven.compile.classpath` in classpath? You were afraid that it might cause some issues.

gktrk (changes requested):

Flowdalic:
@gktrk Thanks for the review. :) Pushed new revision

dev-lang/clojure/clojure-1.10.3.ebuild
```diff
@@ -0,0 +1,97 @@
...
+	https://github.com/${PN}/${PN}/archive/${P}.tar.gz
+	https://github.com/clojure/spec.alpha/archive/spec.alpha-${SPEC_ALPHA_VER}.tar.gz
+	https://github.com/clojure/core.specs.alpha/archive/core.specs.alpha-${CORE_SPECS_ALPHA_VER}.tar.gz
+	test? (
+		https://github.com/clojure/tools.namespace/archive/refs/tags/tools.namespace-${TOOLS_NAMESPACE_VER}.tar.gz
+		https://github.com/clojure/java.classpath/archive/refs/tags/java.classpath-${JAVA_CLASSPATH_VER}.tar.gz
+		https://github.com/clojure/tools.reader/archive/refs/tags/tools.reader-${TOOLS_READER_VER}.tar.gz
+		https://github.com/clojure/test.generative/archive/refs/tags/test.generative-${TEST_GENERATIVE_VER}.tar.gz
+		https://github.com/clojure/data.generators/archive/refs/tags/data.generators-${DATA_GENERATORS_VER}.tar.gz
+		https://github.com/clojure/test.check/archive/refs/tags/test.check-${TEST_CHECK_VER}.tar.gz
```
telans, line 32: You might want to prepend `clojure-` or similar to the distfile names so it's clear what packages these files refer to

gktrk (approved):

dev-lang/clojure/clojure-1.10.3.ebuild
gktrk: We should consider it if we have distfiles collisions in the future

PR #21292, opened 2021-06-17
kill 0ad properly

holgersson32644:
- games-strategy/0ad: Fix stopping the binary with USE="-nvtt"
- games-strategy/0ad: Remove the broken version


thesamesam:
Thanks!

PR #21295, opened 2021-06-17
dev-python/fonttools: disable py3.7, move fs to RDEPEND

Alessandro-Barbieri:


Alessandro-Barbieri:
Can I get some love now since builds are failing?

PR #21297, opened 2021-06-18
gui-wm/hikari: version bump to 2.3.0

epsilon-0:


ionenwks (commented):
Looks like 2.3.1 is out too if you'd rather get that one in, then again I assume you still don't have something to properly test these right now :(

I'll do some checks from my end later.. using nouveau which kinda worked last time I tried it fwiw

ionenwks:
done in commit 31f1c5a908248358ae1955f2b706828180198ae9 since seems to be away

PR #21301, opened 2021-06-18
games-engines/odamex: Version bump to 0.9.3

vilhelmgray:
Closes: https://bugs.gentoo.org/796575


ionenwks:
Seems to be missing a dep on `net-misc/curl` (albeit that was missing in 0.9.2 as well, be a good time to add it).

What's up with jsoncpp? I don't see it linking with it, is it using bundled copy?

Also, is there a reason to keep 0.9.0? Feel free to drop it at same time if not.
```
$ scanelf -n /usr/bin/odamex
 TYPE   NEEDED FILE 
ET_DYN libX11.so.6,libSDL2-2.0.so.0,libSDL2_mixer-2.0.so.0,libcurl.so.4,libportmidi.so,libpng16.so.16,libstdc++.so.6,libm.so.6,libgcc_s.so.1,libc.so.6 /usr/bin/odamex

$ cat /var/db/pkg/games-engines/odamex-0.9.3/REQUIRES
x86_64: libSDL2-2.0.so.0 libSDL2_mixer-2.0.so.0 libX11.so.6 libc.so.6 libcurl.so.4 libgcc_s.so.1 libm.so.6 libminiupnpc.so.17 libpng16.so.16 libportmidi.so libpthread.so.0 libstdc++.so.6 libwx_baseu-3.0-gtk3.so.0 libwx_baseu_net-3.0-gtk3.so.0 libwx_gtk3u_adv-3.0-gtk3.so.0 libwx_gtk3u_core-3.0-gtk3.so.0 libwx_gtk3u_xrc-3.0-gtk3.so.0
```

ionenwks:
> What's up with jsoncpp? I don't see it linking with it, is it using bundled copy?

Oh right, this was the package that needed a old jsoncpp wasn't it, but if so why is it listed in RDEPEND?

vilhelmgray:
> Seems to be missing a dep on `net-misc/curl` (albeit that was missing in 0.9.2 as well, be a good time to add it).

`net-misc/curl` is a core package isn't it? I omitted it because I assumed every portage user would already have it installed. If my assumption is incorrect, I'll add it to the dep list.

> What's up with jsoncpp? I don't see it linking with it, is it using bundled copy?

I think at one point in the past we had it linking to the system-provided jsoncpp, but Odamex never updated so they are expecting an old version of jsoncpp that is incompatible with the versions Gentoo provides. I think we just forgot to remove this from the dep list so I'll update this ebuild to take care of that.

> Also, is there a reason to keep 0.9.0? Feel free to drop it at same time if not.

Ack. :slightly_smiling_face:

ionenwks:
> `net-misc/curl` is a core package isn't it? I omitted it because I assumed every portage user would already have it installed. If my assumption is incorrect, I'll add it to the dep list.

wget is in base set, but curl isn't, it's entirely possible for it to be missing on a minimal server without build deps or the like.

> I think at one point in the past we had it linking to the system-provided jsoncpp, but Odamex never updated so they are expecting an old version of jsoncpp that is incompatible with the versions Gentoo provides. I think we just forgot to remove this from the dep list so I'll update this ebuild to take care of that.

:+1: rdep being there just had me wondering

vilhelmgray:
> > `net-misc/curl` is a core package isn't it? I omitted it because I assumed every portage user would already have it installed. If my assumption is incorrect, I'll add it to the dep list.
> 
> wget is in base set, but curl isn't, it's entirely possible for it to be missing on a minimal server without build deps or the like.

Ack.

ionenwks:
By dropping at same time, I meant in the same PR but in a separate commit.

ionenwks:
Thanks. I just dropped the `:=` given net-misc/curl doesn't define a subslot (not wrong per-se, but unnecessarily adding can make portage dep checks slower).

I'll likely handle merging these in the future, hopefully in a timely fashion :)

vilhelmgray:
Thanks @ionenwks feel free to add yourself as a maintainer to the `metadata.xml` as well.

ionenwks:
> Thanks @ionenwks feel free to add yourself as a maintainer to the `metadata.xml` as well.

I am indirectly given I'm in games project.

PR #21308, opened 2021-06-18
media-sound/scream: 3.7

DarthGandalf:


media-sound/scream/scream-3.7.ebuild
```diff
@@ -0,0 +1,44 @@
...
+
+if [[ ${PV} == *9999* ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/duncanthrax/scream.git"
+else
+	SRC_URI="https://github.com/duncanthrax/scream/archive/${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64"
+fi
+
+S="${WORKDIR}/${P}/Receivers/unix"
```
SoapGentoo, line 19: doesn't this need to go into the "else" block above?

SoapGentoo (changes requested):

media-sound/scream/scream-3.7.ebuild
DarthGandalf: No, why?
This is true for both cases.

SoapGentoo (approved):

PR #21313, opened 2021-06-18
media-libs/libaom: version bump to v3.1.1

novomesk:
New, faster AV1 encoder/decoder to be used by media-libs/libavif for example.


novomesk:
@Whissi Thomas, you comited previous 3.0.0 version.
Could you review this one please?

Whissi:
Looks good, wondering how stable the SRC_URI will be but I guess we will find out. New release got support for butteraugli but we neither have butteraugli nor libjxl packaged in Gentoo yet so we can ignore it.

novomesk:
Thanks,
I have libjxl in ::guru overlay but only live ebuild.

PR #21316, opened 2021-06-18
dev-lang/ispc: bump to 1.16.0

Fat-Zer:
Closes: https://bugs.gentoo.org/796632
Signed-off-by: Alexander Golubev <fatzer2@gmail.com>

-------------------------

It seems virtually no changes compare to the live ebuild required (besides adding keywords).


dev-lang/ispc/ispc-1.16.0.ebuild
```diff
@@ -0,0 +1,84 @@
...
+DESCRIPTION="Intel SPMD Program Compiler"
+HOMEPAGE="https://ispc.github.io/"
+
+if [[ ${PV} = *9999 ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/ispc/ispc.git"
+	KEYWORDS=""
+else
+	SRC_URI="https://github.com/${PN}/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64 ~x86"
```
ionenwks, line 21: `KEYWORDS=` shouldn't appear twice in an ebuild (notably so ekeyword tool is not confused), normally you shouldn't have the empty `=""` one and define keyboards only in the else block. This is how the live ebuild should be as well so these simple copies work.

dev-lang/ispc/ispc-1.16.0.ebuild
```diff
@@ -0,0 +1,84 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8,9} )
+
+inherit cmake toolchain-funcs python-any-r1 llvm
```
ionenwks, line 8: toolchain-funcs appears unused

dev-lang/ispc/ispc-1.16.0.ebuild
```diff
@@ -0,0 +1,84 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8,9} )
```
ionenwks, line 6: remove 7 and check if 10 works

ionenwks (commented):
Yeah, this live ebuild been cared for recently, so this bump should be fine and tests are passing for me (not that I know this package, just make sure it's working as expected for you).

ebuild style aside, just a few things that should be done (can also add an additional commit to sync live ebuild, or I could do it after)..

dev-lang/ispc/ispc-1.16.0.ebuild
Fat-Zer: dropped for both: release and live

dev-lang/ispc/ispc-1.16.0.ebuild
Fat-Zer: dropped for both: release and live

dev-lang/ispc/ispc-1.16.0.ebuild
Fat-Zer: fixed, and moved it in live ebuild as well

PR #21318, opened 2021-06-19
dev-java/guava: version bump to 30.1.1

anaveragehuman:
Going back to SLOT 0 as [upstream claims][1] "APIs without `@Beta` will remain binary-compatible for the indefinite future."
It compiles, but I didn't check reverse dependencies.

Bug: https://bugs.gentoo.org/760111
Closes: https://bugs.gentoo.org/809974

[1]: https://github.com/google/guava#important-warnings


vaukai:
Is it supposed to solve https://bugs.gentoo.org/760111? (if so it should be mentioned in the commit message)

thesamesam:
Big thanks for this!

PR #21319, opened 2021-06-19
(CVE-2020-10683) dev-java/dom4j: bump to 2.1.3

vaukai:
dev-java/dom4j: bump to 2.1.3 (CVE-2020-10683) 
Bug: https://bugs.gentoo.org/719318
rewritten with java-pkg-simple.eclass
introducing "jaxen" USE flag

To break the cyclic dependency on `jaxen` I added a prebuilt `jaxen-1.2.0.jar` to `JAVA_GENTOO_CLASSPATH_EXTRA`
However the problem with `ProxyXmlStartTag.java` remains:
```
 * Compiling ...
warning: [options] bootstrap class path not set in conjunction with -source 8
src/main/java/org/dom4j/xpp/ProxyXmlStartTag.java:31: error: ProxyXmlStartTag is not abstract and does not override abstract method removeAttributeByRawName(String) in XmlStartTag
public class ProxyXmlStartTag implements XmlStartTag {
       ^
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error
1 warning
```
Maybe caused by `dev-java/xpp2` or `dev-java/xpp3`?

PR #21326, opened 2021-06-19
dev-java/jakarta-activation-api: allow java 1.8:*

vaukai:


dev-java/jakarta-activation-api/jakarta-activation-api-2.0.1-r1.ebuild
```diff
@@ -19,15 +19,23 @@ LICENSE="EPL-1.0"
...
 
 S="${WORKDIR}/jaf-${PV}/activation"
 
 JAVA_SRC_DIR="src/main/java"
 JAVA_RESOURCE_DIRS="src/main/resources"
+
+src_prepare() {
+	default
+	# src/main/java/module-info.java:11: error: modules are not supported in -source 8
+	rm "${JAVA_SRC_DIR}"/module-info.java || die
```
fordfrog, line 35: this is a bit problematic as removing `module-info.java` might cause the jar won't be able to be used when running using java 11.

vaukai: > this is a bit problematic as removing `module-info.java` might cause the jar won't be able to be used when running using java 11.

How to test, how to verify?
How else to solve that problem?

fordfrog: we need to solve this on an eclass level i suppose. there are imo two solutions:
1. exclude module-info.java from compilation if jdk 1.8 is used and compile it separately using jdk > 1.8 - this isn't probably doable atm, but that's the way we should go
2. provide compiled version of module-info.java - this is an ugly hack

vaukai: As I understand`jdk:1.8` will be used if we restrict to `dev-java/jdk:1.8`. Could this be a temporary workaround?

fordfrog: i'd rather find a solution that will work forever than creating workarounds that we will cause other issues.

fordfrog: here are my thoughts:

if we use `>=virtual/jdk-11` in `DEPEND` and `>=virtual/jdk-1.8` in `RDEPEND`, the package does not compile (using jdk 11) because the target is 1.8 and `module-info.java` is not supported in that target - that means we have to split (in `java-pkg-simple.eclass`) compilation of the sources and `module-info.java` in every case where target is < 9, no matter what jdk version is used for compilation - but at least when using jdk > 1.8, we can use the current jdk and do not have to switch to a newer jdk - but this would work only in case `[gentoo-vm]` is unmasked for jdk 11.

if we would like to not break the packages by removing `module-info.java` (i do not want to break them), we need jdk > 1.8 for compilation of at least `module-info.java`.

so, anything else except `DEPEND=">=virtual/jdk-11"` and `RDEPEND=">=virtual/jdk-1.8"` and updating `java-pkg-simple.eclass` i consider a workaround that might cause issues and that we will have to remove and solve anyway. from this point of view, the straight solution would be to update `java-pkg-simple.eclass` accordingly and unmask `[gentoo-vm]`. but the tracker still has a long list of unresolved issues.

maybe someone can come with a more elegant solution.

vaukai: Referenced in https://bugs.gentoo.org/719318

fordfrog: i created patch for `java-pkg-simple.eclass` to handle correctly `module-info.java`:
```
diff --git a/eclass/java-pkg-simple.eclass b/eclass/java-pkg-simple.eclass
index ca03fea78cf4..c8c73c8962d9 100644
--- a/eclass/java-pkg-simple.eclass
+++ b/eclass/java-pkg-simple.eclass
@@ -317,7 +317,7 @@ java-pkg-simple_prepend_resources() {
 # If USE FLAG 'binary' exists and is set, it will just copy
 # ${JAVA_BINJAR_FILENAME} to ${S} and skip the rest of src_compile.
 java-pkg-simple_src_compile() {
-       local sources=sources.lst classes=target/classes apidoc=target/api
+       local sources=sources.lst classes=target/classes apidoc=target/api moduleinfo
 
        # auto generate classpath
        java-pkg_gen-cp JAVA_GENTOO_CLASSPATH
@@ -335,7 +335,14 @@ java-pkg-simple_src_compile() {
        fi
 
        # gather sources
-       find "${JAVA_SRC_DIR[@]}" -name \*.java > ${sources}
+       # if target < 9, we need to compile module-info.java separately
+       # as this feature is not supported before Java 9
+       if [[ java-pkg_get-target -lt 9 ]]; then
+               find "${JAVA_SRC_DIR[@]}" -name \*.java ! -name module-info.java > ${sources}
+               moduleinfo=$(find "${JAVA_SRC_DIR[@]}" -name module-info.java)
+       else
+               find "${JAVA_SRC_DIR[@]}" -name \*.java > ${sources}
+       fi
 
        # create the target directory
        mkdir -p ${classes} || die "Could not create target directory"
@@ -349,6 +356,22 @@ java-pkg-simple_src_compile() {
                ${classpath:+-classpath ${classpath}} ${JAVAC_ARGS}\
                @${sources}
 
+       # handle module-info.java separately as it needs at least JDK 9
+       if [[ -n ${moduleinfo} ]]; then
+               if java-pkg_is-vm-version-ge "9" ; then
+                       local tmp_source=${JAVA_PKG_WANT_SOURCE} tmp_target=${JAVA_PKG_WANT_TARGET}
+
+                       JAVA_PKG_WANT_SOURCE="9"
+                       JAVA_PKG_WANT_TARGET="9"
+                       ejavac -d ${classes} -encoding ${JAVA_ENCODING} ${JAVAC_ARGS} "${moduleinfo}"
+
+                       JAVA_PKG_WANT_SOURCE=${tmp_source}
+                       JAVA_PKG_WANT_TARGET=${tmp_target}
+               else
+                       eerror "Need at least JDK 9 to compile module-info.java"
+               fi
+       fi
+
        # javadoc
        if has doc ${JAVA_PKG_IUSE} && use doc; then
                mkdir -p ${apidoc}
```
this way the package compiles but issues the error `Need at least JDK 9 to compile module-info.java` if jdk 1.8 is used. it should exit after the error, which is not part of the patch atm. the problem with this is that without exiting if users are already using java 11, packages are compiled with java 1.8 and some feature depends on java modules then things will get broken. also, after jdk 11 would be unmasked, for safety everything should be re-emerged, but first all packages using `module-info.java` would have to be updated to use jdk 11, so that `module-info.java` files are correctly compiled and included in the jar files. if we create a tracker for this and bump every such a package, only the affected packages would be re-emerged. but it does not solve the possible issues for those already using java 11.

vaukai: > if we create a tracker for this and bump every such a package, only the affected packages would be re-emerged.

Such tracker should be very helpful. At first it could collect those bugs waiting for module support. First  candidates:
- [dev-java/jakarta-activation-api](https://bugs.gentoo.org/show_bug.cgi?id=792666) could be moved there
- [dev-java/jaxb-api](https://github.com/gentoo/gentoo/pull/21319/commits/1c5ad2c2ec84a7a8942bb09440353b9228fe55ae)
- [dev-db/hsqldb-2.6.0](https://bugs.gentoo.org/show_bug.cgi?id=383445)

fordfrog: it's probably the only way to go atm. so if you agree, the steps should be probably the following:
1. do more tests for the `java-pkg-simple.eclass` patch. i already did some testing yesterday but if you could do some testing too, it would be better as i might have omitted some cases.
2. merge the eclass patch
3. create the tracker, lower the java requirements of the packages that provide `module-info.java` to 1.8, commit the changes and create related bugs
what do you think? it might cause some issues for those that have `[gentoo-vm]` enabled on java 11 but don't have it set as the default vm, but i have no better solution atm.

vaukai: Of course. I'm seriously planning to test the `java-pkg-simple.eclass` patch. I need it in my [jaxb-api](https://github.com/vaukai/gentoo/tree/jaxb-api) branch and also in my `testing` custom repository.
But I'd like to see the tracker right now so we can start feeding it with the packages we are aware.
And, `lower the java requirements` would not be possible for upcoming [dev-db/hsqldb-2.6.0](https://sourceforge.net/p/hsqldb/news/2021/03/-hypersql-version-260-released/). It wants Java >= 11

fordfrog: so [here](https://bugs.gentoo.org/796875) is the tracker. i also included the description of the issue and what needs to be done once jdk 11 is unmasked so that we have all needed info at a single place.

vaukai:
@fordfrog 
Just only for testing I had to include your patch in this git branch. Please do not merge it.

I've tested `jakarta-activation-api` both with `>=virtual/jdk-1.8:*` and `>=virtual/jdk-11:*` and both with `openjdk:8` and `opendjk:11` as system vm.

What other useful tests could I try?

fordfrog:
i think emerging few other packages, both with and without `module-info.java`, would be useful to be sure the patch won't break anything.

fordfrog:
[java-pkg-simple.eclass.patch](https://github.com/gentoo/gentoo/files/6682653/java-pkg-simple.eclass.patch.txt)
here's the updated patch. not much tested though :-)

vaukai:
> [java-pkg-simple.eclass.patch](https://github.com/gentoo/gentoo/files/6682653/java-pkg-simple.eclass.patch.txt)
> here's the updated patch. not much tested though :-)

With this patch applied to [dev-java/jaxb-api](https://github.com/gentoo/gentoo/pull/21319#issue-673838260)
```
FAILURES!!!
Tests run: 232,  Failures: 130
```
This high number of failures makes me think there is something in a wrong directory.  Before switching it was like:
```
INFO: [scenario-12]  TEST PASSED
..
Time: 1.471

OK (232 tests)

>>> Completed testing dev-java/jaxb-api-2.3.3
```

fordfrog:
i can't see anything wrong in the patch. do you have some output of the failures? just to have a quick look, can't investigate it more atm.

vaukai:
> i can't see anything wrong in the patch. do you have some output of the failures? just to have a quick look, can't investigate it more atm.

See https://bugs.gentoo.org/796995

dev-java/jakarta-activation-api/jakarta-activation-api-2.0.1-r1.ebuild
```diff
@@ -0,0 +1,35 @@
...
+DESCRIPTION="Jakarta Activation API jar"
+HOMEPAGE="https://eclipse-ee4j.github.io/jaf/"
+SRC_URI="https://github.com/eclipse-ee4j/jaf/archive/refs/tags/${PV}.tar.gz -> jakarta-activation-${PV}.tar.gz"
+
+LICENSE="EPL-1.0"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
+
+DEPEND="virtual/jdk:1.8"
+RDEPEND="virtual/jre:1.8"
```
fordfrog, line 23: these two lines will cause `module-info.java` is never compiled, even for those that have jdk 11 enabled.

vaukai: misunderstanding, sorry.
Re-done.

vaukai:
Just realized the ebuild is missing the [activationapi](https://github.com/eclipse-ee4j/jaf/tree/58f91fb5e1e128f13b5e44053839cbf07ebaa5a6/activationapi) part. 
But there is a [recent change](https://github.com/eclipse-ee4j/jaf/commit/93fb38633ca0da77c6c92e9cd2e52fd4478186d2#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L95-R95) replacing that part.

fordfrog:
@vaukai thanks! the api is really confusing, even the upstream repo is different from the tarball content... anyway, the final jar produced by your ebuild contained also stuff that should not be there. for details please see commit 45bbfeed8b0f88ab12ee33d90e386dfca1e9357c. it's always better to grab upstream's jar and compare it to ours, when unsure. some of the info i used is present in `activationapi/pom.xml`.

vaukai:
>  for details please see commit [45bbfee](https://github.com/gentoo/gentoo/commit/45bbfeed8b0f88ab12ee33d90e386dfca1e9357c).

If I had used a recent [snapshot ... ](https://github.com/eclipse-ee4j/jaf/commit/93fb38633ca0da77c6c92e9cd2e52fd4478186d2)

PR #21327, opened 2021-06-19
Add keyword ~riscv to ebuild

HomeHongZhang:


scardracs:
If you want to add an arches to some ebuild you must have to file a bug in bugs.gentoo.org with the list of packages you want to add. That's not the way to do it properly and no one will merge that pr

PR #21331, opened 2021-06-19
dev-ml/camlbz2: add ocamlopt USE flag

pinicarus:
Signed-off-by: Pierre-Nicolas Clauss <pinicarus@protonmail.com>

Without this USE flag, emerge-ing the package fails with this error:

masked by: invalid: RDEPEND: USE flag 'ocamlopt' referenced in conditional 'ocamlopt?' in atom 'dev-lang/ocaml:=[ocamlopt?]' is not in IUSE

### DIFF
--- a/dev-ml/camlbz2/camlbz2-0.7.0.ebuild
+++ b/dev-ml/camlbz2/camlbz2-0.7.0.ebuild
@@ -13,7 +13,7 @@ SRC_URI="https://gitlab.com/irill/camlbz2/-/archive/${PV}/${P}.tar.gz"
 LICENSE="LGPL-2.1"
 SLOT="0/${PV}"
 KEYWORDS="amd64 arm arm64 ppc ppc64 x86"
-IUSE="doc"
+IUSE="doc +ocamlopt"
 
 DEPEND="
 	app-arch/bzip2


pinicarus:
Hi,

Kind reminder that as such, this ebuild is not installable.
I'd be nice to see this PR accepted.

ionenwks:
Randomly spotted this, it'd be good to find a proxy else these will just be lost. One reason `proxied="yes"` tags were added was for these at g.o addresses without commit access.

I don't really know much about ML stuff though, maybe @thesamesam could see what to do with this.

thesamesam:
> Hi,
> 
> Kind reminder that as such, this ebuild is not installable.
> I'd be nice to see this PR accepted.

Are you sure this is right? We fixed the [eclass](679f4f3a7554cdf45dbee9186f274c60966597ed).

pinicarus:
> > Hi,
> > Kind reminder that as such, this ebuild is not installable.
> > I'd be nice to see this PR accepted.
> 
> Are you sure this is right? We fixed the [eclass](679f4f3a7554cdf45dbee9186f274c60966597ed).

You're right, I tried again with the latest portage tree and it works fine : this PR is moot

PR #21335, opened 2021-06-19
sys-kernel/gentoo-kernel: remove unused patch(es)

mm1ke:


gyakovlev:
thanks!

PR #14722, opened 2020-02-20
dev-qt/qtwebengine: 5.14.1 version is enabled to support clang

dannftk:
5.14.1 version is enabled to support clang

Signed-off-by: Denis Pronin <dannftk@yandex.ru>

PR #21337, opened 2021-06-19
dev-qt/qtwebengine: fix build with sys-devel/clang[default-libcxx]

telans:
`dev-qt/qtwebengine` passes `-static-libstdc++` to the compiler when building `src/3rdparty/gn`. When using `libc++` as the defualt stdlib the build fails with `error: unable to find library -lc++`.

This issue is not able to be worked around with CXXFLAGS as all flags are stripped for the majority of the qtwebengine build. It's also why there is no check such as `is-flagq '-stdlib=libc++` in the ebuild. This issue only appears to exist when `sys-devel/clang` has been built with `[default-libcxx]` as the build system expects libstdc++ as the default stdlib.

With this change I am able to successfully build qtwebengine with `sys-devel/clang[default-libcxx]`

Closes: https://github.com/gentoo/gentoo/pull/14722

### DIFF
--- a/dev-qt/qtwebengine/files/qtwebengine-5.15.2_p20210521-clang-libc++.patch
+++ b/dev-qt/qtwebengine/files/qtwebengine-5.15.2_p20210521-clang-libc++.patch
@@ -1,0 +1,10 @@
+--- a/src/buildtools/gn.pro
++++ b/src/buildtools/gn.pro
+@@ -25,6 +25,7 @@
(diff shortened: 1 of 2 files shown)


gyakovlev:
@telans I only maintain ppc64 patchset here, so not my decision.
try poking @Pesa and maybe @a17r can chime in indirectly too.

thesamesam:
Thanks all. I concur.

PR #21343, opened 2021-06-20
python-r1 suite EAPI 8 support

mgorny:
CC @gentoo/python

### DIFF
--- a/eclass/multibuild.eclass
+++ b/eclass/multibuild.eclass
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 # @ECLASS: multibuild.eclass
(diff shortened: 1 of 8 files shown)


eclass/multibuild.eclass
```diff
@@ -1,23 +1,23 @@
...
 # multiple 'variants' of a package (e.g. multilib, Python
 # implementations).
 
 case "${EAPI:-0}" in
-	0|1|2|3)
+	[0-3])
 		die "Unsupported EAPI=${EAPI:-0} (too old) for ${ECLASS}"
 		;;
-	4|5|6|7)
+	[4-8])
```
SoapGentoo, line 20: why not tighten this further?

SoapGentoo (changes requested):

eclass/multibuild.eclass
mgorny: I'm waiting for the last EAPI=5 consumers to go, then I will be able to remove a lot of old code.

SoapGentoo: but you're still declaring EAPI 4 valid?

mgorny: Yes because this commit is 'add new EAPI', 'remove old EAPIs' will come later.

PR #21350, opened 2021-06-21
media-gfx/freecad: compatibility update

waebbl:
update package to work with sci-libs/opencascade-7.5.2 and sci-libs/vtk-9

profiles: mask pcl USE flag on non-live freecad
sci-libs/pcl-1.11.1 doesn't support >=sci-libs/vtk-9
Bug: https://bugs.gentoo.org/796368

drop 0.19.2-r1

media-gfx/freecad: improve display of optional packages
See https://forum.freecadweb.org/viewtopic.php?f=42&t=37806 for a list of
supported optional packages

media-gfx/freecad: sync changes from 0.19.2 to live ebuild

- support sci-libs/opencascade-7.5.2
- support sci-libs/vtk-9
- add more optfeature's

Note that the live ebuild currently depends on the live ebuild of pcl (sci-libs/pcl-9999),
which works with vtk-9. So this version is not use masked.


thesamesam:
Thanks!

waebbl:
Thanks for merging. I hope they work as they are. I've tested 0.19.2 against v7.5.1 and v7.5.2 of opencascade and against v8.2.0 and v9.0.1 of vtk, but live ebuild has only been tested against latest versions.

PR #21352, opened 2021-06-21
dev-python/agate: Fix the wrong description & homepage

vaartis:
Signed-off-by: Ekaterina Vaartis <vaartis@kotobank.ch>

### DIFF
--- a/dev-python/agate/agate-1.6.1.ebuild
+++ b/dev-python/agate/agate-1.6.1.ebuild
@@ -7,8 +7,8 @@ DISTUTILS_USE_SETUPTOOLS=bdepend
 
 inherit distutils-r1
 
-DESCRIPTION="Python charting for 80% of humans."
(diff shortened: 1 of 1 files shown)


vaartis:
The bot is incorrect, it's _not_ a new package.

mgorny:
> The bot is incorrect, it's _not_ a new package. @zmedico

I'm going to take a look why it gets confused. Maybe GH API changed or sth.

vaartis:
> 
> 
> > The bot is incorrect, it's _not_ a new package. @zmedico
> 
> I'm going to take a look why it gets confused. Maybe GH API changed or sth.

May be because the package is just about an hour old, if that means anything.

mgorny:
> May be because the package is just about an hour old, if that means anything.

Yep, it's on yet on the host that does the assignment. Maybe I should make it fetch `metadata.xml` via gitweb…

zmedico:
Thanks!

PR #21354, opened 2021-06-21
sys-process/lsof: backport arm sigbus fix

10ne1:
sys-process/lsof: backport arm sigbus crash fix
    
    This backports an upstream fix for a crash which happens on
    armv7a + glibc 2.33 due to a buffer misalignment.
    
    Upstream issue: https://github.com/lsof-org/lsof/issues/160
    Upstream commit: 21cb1dad1243f4c0a427d893babab12e48b60f0e
    
    Bug: https://bugs.gentoo.org/797358
    

thesamesam:
Thank you! Give me a week or so and I'll figure out what we need to do for the other PRs, by the way. Ping me if I haven't by then (it involves getting the right people together).

PR #21358, opened 2021-06-21
www-servers/hinsightd new package

tiotags1:
Hello, I would like to add my server to ::gentoo, feedback requested


www-servers/hinsightd/files/initd-v1.sh
```diff
@@ -0,0 +1,50 @@
...
+RUN_DIR=/var/run/$NAME
+LOG_DIR=/var/log/$NAME
+TMP_DIR=/var/tmp/$NAME
+CWD_DIR=/var/www/localhost
+PID_FILE=$RUN_DIR/$NAME.pid
+CFG_FILE=/etc/hinsightd/main.lua
+LOG_FILE=$LOG_DIR/hindsight.log
+
+RUN_FILE=/usr/sbin/hinsightd
+RUN_USER="hinsightd"
```
monsieurp, line 13: Please declare those variables in a conf.d file instead of here.

www-servers/hinsightd/files/initd-v1.sh
```diff
@@ -0,0 +1,50 @@
...
+
+reload() {
+  if ! service_started "${NAME}" ; then
+    eerror " * ERROR ${NAME} isn't running"
+    return 1
+  fi
+
+  checkconfig || return 1
+
+  echo " * Reloading ${NAME} ..."
```
monsieurp, line 44: `einfo` is prefered instead of `echo`.

monsieurp (changes requested):

www-servers/hinsightd/files/initd-v1.sh
tiotags1: you mean how nginx does it ?
command_user=${HINSIGHTD_USER:-hinsightd}

www-servers/hinsightd/files/initd-v1.sh
tiotags1: I was looking for that green star, didn't think it was that simple to generate

PR #21359, opened 2021-06-21
app-shells/bash: remove /bin/sh logic from pkg_preinst

floppym:
This was added back in 2008 to handle some migration in file ownership.
I expect all users have upgraded many times since then.

Closes: https://bugs.gentoo.org/778311
Signed-off-by: Mike Gilbert <floppym@gentoo.org>


floppym:
Replaces #21030.

Polynomial-C (approved):

PR #21360, opened 2021-06-21
sys-libs/lwp: EAPI7, ltprune--

mm1ke:
This is the last ebuild using `ltprune`.

### DIFF
--- a/sys-libs/lwp/lwp-2.8.ebuild
+++ b/sys-libs/lwp/lwp-2.8.ebuild
@@ -1,8 +1,7 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
-inherit ltprune
+EAPI=7
 
 DESCRIPTION="Light-weight process library (used by Coda)"
 HOMEPAGE="http://www.coda.cs.cmu.edu/"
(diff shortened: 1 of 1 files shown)


sys-libs/lwp/lwp-2.8.ebuild
```diff
@@ -1,8 +1,7 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
-inherit ltprune
+EAPI=7
```
ionenwks, line 4: Hrm, I'd kinda want this to be EAPI-8 given it'd allow to remove `--disable-static` and drop src_configure -- but, given this is straight-to-stable and still kinda early, I'll pass on it. Still wanted to mention it though.

sys-libs/lwp/lwp-2.8.ebuild
```diff
@@ -23,5 +21,5 @@ src_configure() {
 
 src_install() {
 	default
-	prune_libtool_files
+	find "${D}" -name '*.la' -delete || die
```
ionenwks, line 24: `${D}` is technically fine for this, but `${ED}` is more precise and what the [policy guide example](https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0303) uses, so I'll change this on merge.

ionenwks (approved):
Thanks, good to see the last one get rid of this

Well, technically there's still more consumers: eutils, gnome2, and gstreamer eclasses still inherit it -- but it's all in old-eapi deprecated state, and gstreamer will be replaced by gstreamer-meson.

sys-libs/lwp/lwp-2.8.ebuild
mm1ke: Thanks, I'll keep an eye on this for the future :)

PR #21371, opened 2021-06-22
networkmanager-fortisslvpn: bumped to version 1.3.90

axelgenus:
New official release of the NetworkManager plugin for the unofficial FortiVPN client.

Signed-off-by: Alessandro Calorì <alessandro.calori@pm.me>


net-vpn/networkmanager-fortisslvpn/networkmanager-fortisslvpn-1.3.90.ebuild
```diff
@@ -0,0 +1,43 @@
...
+		>=app-crypt/libsecret-0.18
+		>=net-libs/libnma-1.2.0
+		>=x11-libs/gtk+-3.4:3
+	)
+"
+DEPEND="${RDEPEND}
+	dev-util/gdbus-codegen
+	>=sys-devel/gettext-0.19
+	virtual/pkgconfig
+"
```
juippis, line 34: These all look to be `BDEPEND` instead.
```suggestion
DEPEND="${RDEPEND}"
BDEPEND="dev-util/gdbus-codegen
	>=sys-devel/gettext-0.19
	virtual/pkgconfig"

**Review by juippis — COMMENTED** [review]
```
 * QA Notice: This package seems to install metainfo files into the
 * following location(s):
 * 
 *   /usr/share/appdata
 * 
 * This location is deprecated, it should not be used anymore by new software.
 * Appdata/Metainfo files should be installed into /usr/share/metainfo directory.
 * For more details, please see the freedesktop Upstream Metadata guidelines at
 * https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html
 * 
 * Gentoo Tracker bug: https://bugs.gentoo.org/709450
 * 
 * 
 *   /var/lib
 *   /var/lib/NetworkManager-fortisslvpn
 * 
```
These should be fixed.

juippis (approved):
Few minor fixes still needed, but I'll add a separate commit. Thanks!

PR #21372, opened 2021-06-22
dev-libs/tvision: avoid pre-stripping

waebbl:
Closes: https://bugs.gentoo.org/797454

### DIFF
--- a/dev-libs/tvision/tvision-2.2.3.ebuild
+++ b/dev-libs/tvision/tvision-2.2.3.ebuild
@@ -56,6 +56,8 @@ src_configure() {
 	myconf+=(
 		--fhs
 		--prefix="${EPREFIX}/usr"
+		# avoid stripping
+		--with-debug
 		--with-pthread
 		--without-static
 		--x-include="${EPREFIX}/usr/include/X11"


dev-libs/tvision/tvision-2.2.3.ebuild
```diff
@@ -56,6 +56,8 @@ src_configure() {
 	myconf+=(
 		--fhs
 		--prefix="${EPREFIX}/usr"
+		# avoid stripping
+		--with-debug
```
thesamesam, line 60: Checked this doesn't do anything irritating like append -g* etc?

waebbl: These two lines get added to the Makefile:
```
$stripDebug=($OSf eq 'Darwin') ? '-S' : '--strip-debug';
$rep.="\tstrip $stripDebug \$(libdir)/$ver\n" unless $conf{'debugInfo'} eq 'yes';
```
From my POV it only calls `strip --strip-debug` with this option enabled. I also didn't notice any unusal compiler flags added during my test. I had enabled in older ebuilds with the USE=debug flag, but noticed this actually didn't add any debugging information, so removed it. But I didn't recognized that it's enabling stripping.
The `$conf{'debugInfo'}` is set to yes with the `--with-debug` option.

waebbl: For some context: https://github.com/set-soft/tvision/blob/master/tvision/config.pl#L1281

juippis (approved):
WFM.

PR #21373, opened 2021-06-22
media-libs/{ilmbase,openexr}, dev-python/pyilmbase: bump to 2.5.7

waebbl:
Closes: https://bugs.gentoo.org/656680
Bug: https://bugs.gentoo.org/776808
Bug: https://bugs.gentoo.org/787452


media-libs/openexr/files/openexr-2.5.7-0001-disable-testRgba-on-sparc.patch
```diff
@@ -0,0 +1,31 @@
...
+diff --git a/IlmImfTest/main.cpp b/IlmImfTest/main.cpp
+index 3245fc4..3ba3e60 100644
+--- a/IlmImfTest/main.cpp
++++ b/IlmImfTest/main.cpp
+@@ -206,7 +206,7 @@ main (int argc, char *argv[])
+     TEST (testXdr, "core");
+     TEST (testHuf, "core");
+     TEST (testWav, "core");
+-    TEST (testRgba, "basic");
++//    TEST (testRgba, "basic");
```
thesamesam, line 25: In future, just drop the line -- we lose no information because the diff still shows what was there in the past.

I'm a bit uneasy about disabling it because it could be a genuine issue in the code (i.e. the test is _exposing_ a problem), but I suppose it's only one. Let's keep an eye on the bug upstream.

waebbl: The upstream bug doesn't had any progress lately. I keep an eye on it. It probably is an issue dedicated only to sparc machines. In the upstream bug, someone said on other big endian architectures, the issue didn't show up.

dev-python/pyilmbase/files/pyilmbase-2.5.7-0003-fix-pkgconfig-file.patch
```diff
@@ -0,0 +1,30 @@
...
++++ b/config/CMakeLists.txt
+@@ -15,9 +15,9 @@ if(PYILMBASE_INSTALL_PKG_CONFIG)
+   # use a helper function to avoid variable pollution, but pretty simple
+   function(pyilmbase_pkg_config_help pcinfile)
+     set(prefix ${CMAKE_INSTALL_PREFIX})
+-    set(exec_prefix ${CMAKE_INSTALL_BINDIR})
+-    set(libdir ${CMAKE_INSTALL_LIBDIR})
+-    set(includedir ${CMAKE_INSTALL_INCLUDEDIR})
++    set(exec_prefix "\${prefix}")
++    set(libdir "\${exec_prefix}/${CMAKE_INSTALL_LIBDIR}")
```
DerDakon, line 23: In the future please replace those by ${CMAKE_INSTALL_FULL_LIBDIR} for the case the patch eventually goes upstream. This will then also works for cases where the user explicitely sets e.g. an include dir that is not inside $prefix.

waebbl: @DerDakon The patch is already upstream, but for some reason it only has found it's way into the new Imath library, but has not yet been backported to the older 2.5 branch of OpenEXR.

DerDakon: Filed an issue for that: https://github.com/AcademySoftwareFoundation/Imath/issues/171

waebbl: Thanks @DerDakon

PR #21375, opened 2021-06-22
dev-python/js2py: patch description_file

xgqt:
Closes: https://bugs.gentoo.org/797520
Signed-off-by: Maciej Barć <xgqt@riseup.net>


mgorny:
This is really an upstream problem and patching ebuilds for it is just noise.

PR #21376, opened 2021-06-22
media-video/vlc: fix building with newer live

ajakk:


media-video/vlc/files/vlc-3.0.14-fix-live-address-api.patch
```diff
@@ -0,0 +1,50 @@
+From 94845266b705dc9de7921408531b9d7704f4458f Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: Minor but I'd add the bug link at the top (the Gentoo one at least, we could do the VLC one too for convenience)

PR #21380, opened 2021-06-22
net-misc/netkit-rsh: security fix, add myself as p-m

hlein:
This updates one of our existing patches with Debian's fix for
CVE-2019-7282 and CVE-2019-7283.

Bug: https://bugs.gentoo.org/717794

### DIFF
--- a/net-misc/netkit-rsh/files/patches/000_all_sectty.patch
+++ b/net-misc/netkit-rsh/files/patches/000_all_sectty.patch
@@ -1,0 +1,32 @@
+--- a/rexecd/rexecd.c
++++ b/rexecd/rexecd.c
+@@ -300,6 +300,7 @@
+        PAM_password = pass;
+        pam_error = pam_start("rexec", PAM_username, &PAM_conversation,&pamh);
+        PAM_BAIL;
++       (void) pam_set_item (pamh, PAM_TTY, "rexec");   /* we don't have a tty yet! */
+        pam_error = pam_authenticate(pamh, 0);
+        PAM_BAIL;
+        pam_error = pam_acct_mgmt(pamh, 0);
+--- a/rlogind/auth.c
++++ b/rlogind/auth.c
+@@ -123,7 +123,7 @@
+     pam_set_item(pamh, PAM_USER, localuser);
+     pam_set_item(pamh, PAM_RUSER, remoteuser);
+     pam_set_item(pamh, PAM_RHOST, host);
+-    pam_set_item(pamh, PAM_TTY, "tty");   /* ? */
++    pam_set_item(pamh, PAM_TTY, "rlogin");   /* we don't have a tty yet! */
+ 	
+     network_confirm();
+     retval = attempt_auth();
+--- a/rshd/rshd.c
++++ b/rshd/rshd.c
+@@ -243,7 +243,7 @@
+     }
+     pam_set_item (pamh, PAM_RUSER, remuser);
+     pam_set_item (pamh, PAM_RHOST, hostname);
+-    pam_set_item (pamh, PAM_TTY, "tty");
++    pam_set_item (pamh, PAM_TTY, "rsh");  /* we don't use a tty, so punt */
+     
+     retcode = pam_authenticate(pamh, 0);
+     if (retcode == PAM_SUCCESS) {
--- a/net-misc/netkit-rsh/netkit-rsh-0.17-r12.ebuild
+++ b/net-misc/netkit-rsh/netkit-rsh-0.17-r12.ebuild
@@ -1,0 +1,78 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit pam toolchain-funcs fcaps
+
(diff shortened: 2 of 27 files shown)


net-misc/netkit-rsh/netkit-rsh-0.17-r12.ebuild
```diff
@@ -0,0 +1,74 @@
...
+SRC_URI="ftp://ftp.uk.linux.org/pub/linux/Networking/netkit/${P}.tar.gz
+	mirror://gentoo/rexec-1.5.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~s390 ~sparc ~x86 ~amd64-linux ~x86-linux"
+IUSE="pam"
+
+RDEPEND="
+	sys-libs/ncurses:0
```
thesamesam, line 19: := if it links against it

net-misc/netkit-rsh/netkit-rsh-0.17-r12.ebuild
```diff
@@ -0,0 +1,74 @@
...
+DEPEND="${RDEPEND}"
+BDEPEND="app-arch/xz-utils"
+
+FILECAPS=( cap_net_bind_service usr/bin/r{cp,login,sh} )
+
+src_prepare() {
+	rm -r rexec || die
+	mv ../rexec rexec || die
+
+	eapply "${FILESDIR}"/patches
```
thesamesam, line 30: Could you convert this to use PATCHES=() in global scope and drop this + eapply_user? (This is just a nit, really).

net-misc/netkit-rsh/netkit-rsh-0.17-r12.ebuild
```diff
@@ -0,0 +1,74 @@
...
+		sed -i -e "s|./__conftest|: ./__conftest|" configure || die
+	fi
+}
+
+src_configure() {
+	tc-export CC
+	${CONFIG_SHELL:-/bin/sh} ./configure $(usex pam '' '--without-pam') || die
+
+	sed -i \
+		-e "s:-pipe -O2:${CFLAGS}:" \
```
thesamesam, line 44: Let's use another [delimiter](https://bugs.gentoo.org/685160) to close [this](https://bugs.gentoo.org/710960) bug.

net-misc/netkit-rsh/files/patches/000_all_sectty.patch
```diff
@@ -0,0 +1,32 @@
+--- netkit-rsh-0.17-pre20000412/rexecd/rexecd.c.sectty	Sat Oct  2 17:50:52 1999
```
thesamesam, line 1: This is just to be nice IF you can: can you drop the dates and the directory names at the beginning? Change to a/ and b/?


See here: https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#clean-patch-howto. Alright if not though, it's a nit.

thesamesam, line 1: There are a lot of patches. They already come to 39K on my machine (with the in-tree copy).

The limiter for individual patches is 20K but FILESDIR often grows beyond that: https://qa-reports.gentoo.org/output/filesdir-usage.txt.

Could you pop these in a tarball on e.g. your own git repo or similar? I can do this if you prefer (I suppose it makes review easier).

net-misc/netkit-rsh/netkit-rsh-0.17-r12.ebuild
thesamesam: You can do: PATCHES=(
	"${FILESDIR}"/patches/
}

thesamesam (changes requested):
Thank you for working on this!

net-misc/netkit-rsh/netkit-rsh-0.17-r12.ebuild
hlein: Good call! I'll address this and your other feedback shortly. Sheesh, you volunteer to maintain something, and suddenly are expected to do WORK? =)

net-misc/netkit-rsh/files/patches/000_all_sectty.patch
hlein: Well, there's a catch: after this security fix gets accepted and the project gets off the last-rites list, I would really like to update our patchset to/against current Debian, which has other feature-improvements like adding a -p port flag. It's going to be *way* easier to do that and to make clear what has changed, if I'm dropping/modifying individual patches, vs a new opaque tarball that has no clear history from the last one.

If you don't think that'd be a problem, then sure, I can make this a new PATCHVER instead, and ++ again when doing the next set of changes.

thesamesam: Got it. Let's stick with the status quo and remember to do that later.

thesamesam: Oh, and you can add a commit to drop this from package.mask by the way!

hlein:
@thesamesam I think this is ready for another look, thanks!

PR #21381, opened 2021-06-23
sci-mathematics/singular: Fix QA issues in 4.2.0_p3

kiwifb:
Bug: https://bugs.gentoo.org/797655


kiwifb:
This fix the fact that singular segfault and remove noise due to some configuration options.
It also update options passed to configure slightly, removing one that is obsolete and adding a new one.

kiwifb:
@orlitzky pinging to make sure you are aware of my issues. I am not fussy, if we find a different solution for the optimization flags.
I also have 4.2.1 in the wings on the development branch of the sage-on-gentoo overlay but I would rather solve this first.

PR #21382, opened 2021-06-23
dev-util/bear: add libfmt subslot dep

doctaweeks:
Bug: https://bugs.gentoo.org/797715


thesamesam:
I know I told you to do the exact opposite but would you mind revbumping please?

I’m sorry for the inconsistency, I’ve realised since that it’s wrong to rely on Portage’s preserved-libs and that this generally leaves the system in an inconsistent state. So if it helps, you were right!

gyakovlev:
thanks, did a revbump myself.
I swear I checked all deps for it when adding grpc/protobuf subslot operators, maybe got lost in my git dir.

PR #21388, opened 2021-06-23
add ~riscv keyword to dev-python/pillow; fix bug: #797334

HomeHongZhang:
add ~riscv keyword to dev-python/pillow, as well as some dependent packages

fixing bug: https://bugs.gentoo.org/797334


profiles/arch/riscv/package.use.mask
```diff
@@ -3,7 +3,7 @@
 
 # Marek Szuba <marecki@gentoo.org> (2021-06-22)
 # Dependencies not keyworded here yet
-dev-python/pillow imagequant test tk
+dev-python/pillow tk
```
dlan17, line 7: should have separate commit for unmask USEs for pillow, comment it well
please remember it doesn't hurt to have more commits while using git ..

PR #21390, opened 2021-06-23
l10n.eclass: Rename, support EAPI 8, split off strip-linguas

ulm:


ulm:
> _Status_: x **broken**
> 
> New issues caused by PR:"

"Deprecated eclass" as intended :smile:

eclass/plocale.eclass
```diff
@@ -0,0 +1,134 @@
...
+# @SUPPORTED_EAPIS: 6 7 8
+# @BLURB: convenience functions to handle localizations
+# @DESCRIPTION:
+# The plocale (localization) eclass offers a number of functions to more
+# conveniently handle localizations (translations) offered by packages.
+# These are meant to prevent code duplication for such boring tasks as
+# determining the cross-section between the user's set LINGUAS and what
+# is offered by the package.
+
+case ${EAPI:-0} in
```
SoapGentoo, line 18:
```suggestion
case ${EAPI} in
```

eclass/plocale.eclass
```diff
@@ -0,0 +1,134 @@
...
+	[[ $# -eq 3 ]] || die "Exactly 3 arguments are needed!"
+	ebegin "Looking in ${1} for new locales"
+	pushd "${1}" >/dev/null || die "Cannot access ${1}"
+	local current= x=
+	for x in ${2}*${3}; do
+		x=${x#"${2}"}
+		x=${x%"${3}"}
+		current+="${x} "
+	done
+	popd >/dev/null
```
SoapGentoo, line 91:
```suggestion
	popd >/dev/null || die
```

eclass/strip-linguas.eclass
```diff
@@ -0,0 +1,67 @@
...
+
+# @ECLASS: strip-linguas.eclass
+# @MAINTAINER:
+# Ulrich Müller <ulm@gentoo.org>
+# @AUTHOR:
+# Mike Frysinger <vapier@gentoo.org>
+# @SUPPORTED_EAPIS: 5 6 7 8
+# @BLURB: convenience function for LINGUAS support
+
+case ${EAPI:-0} in
```
SoapGentoo, line 12:
```suggestion
case ${EAPI} in
```

eclass/strip-linguas.eclass
```diff
@@ -0,0 +1,67 @@
...
+			done
+			ls=${newls}
+		done
+	else
+		ls="$@"
+	fi
+
+	nols=""
+	newls=""
+	for f in ${LINGUAS}; do
```
SoapGentoo, line 55: is `f` still local here? :thinking:

SoapGentoo (changes requested):

eclass/l10n.eclass
```diff
@@ -20,6 +20,8 @@ case ${EAPI:-0} in
 	*) die "${ECLASS}: EAPI ${EAPI:-0} not supported" ;;
 esac
 
+inherit strip-linguas
```
thesamesam, line 23: You want to go the full migration route, not just force people to migrate on EAPI bump?

It feels better to do what we did with eutils and limit the inheritance to certain EAPIs. So, here, we'd inherit only on <= EAPI 7.

(I know eutils is slightly different because it's gone in EAPI 8, but same principle).

ulm: But that's what we do? In EAPI 8, l10n.eclass will be replaced by plocale.eclass which doesn't inherit strip-linguas.

thesamesam: My fault for doing this just before bed. Absolutely, yes.

thesamesam (approved):

PR #21394, opened 2021-06-23
rpm.eclass: Support EAPI 8, some cleanup

ulm:


eclass/rpm.eclass
```diff
@@ -19,8 +22,8 @@ _RPM_ECLASS=1
 
 inherit estack
 
-case "${EAPI:-0}" in
-	[0-6]) DEPEND=">=app-arch/rpm2targz-9.0.0.3g" ;;
+case ${EAPI} in
+	5|6) DEPEND=">=app-arch/rpm2targz-9.0.0.3g" ;;
 	*) BDEPEND=">=app-arch/rpm2targz-9.0.0.3g" ;;
```
thesamesam, line 27: The >= bound is long-obsolete, we could drop it if we fancy.

eclass/rpm.eclass
```diff
@@ -34,15 +37,16 @@ rpm_unpack() {
 	for a in "$@" ; do
 		echo ">>> Unpacking ${a} to ${PWD}"
 		if [[ ${a} == ./* ]] ; then
-			: nothing to do -- path is local
-		elif [[ ${a} == ${DISTDIR}/* ]] ; then
+			: # nothing to do -- path is local
+		elif [[ ${a} == "${DISTDIR}"/* ]] ; then
 			ewarn 'QA: do not use ${DISTDIR} with rpm_unpack -- it is added for you'
```
thesamesam, line 42: eqawarn? (we're already inheriting eutils too for older EAPIs)

eclass/rpm.eclass
```diff
@@ -90,11 +94,15 @@ rpm_src_unpack() {
...
 # @FUNCTION: rpm_spec_epatch
 # @USAGE: [spec]
+# @DEPRECATED: none
 # @DESCRIPTION:
 # Read the specified spec (defaults to ${PN}.spec) and attempt to apply
 # all the patches listed in it.  If the spec does funky things like moving
 # files around, well this won't handle that.
 rpm_spec_epatch() {
+	# no epatch in EAPI 7 and later
+	[[ ${EAPI} == [56] ]] || die "${FUNCNAME} is banned in EAPI ${EAPI}"
```
thesamesam, line 104: As a minor hint for users:
```suggestion
	[[ ${EAPI} == [56] ]] || die "${FUNCNAME} is banned in EAPI ${EAPI}, use eapply"
```

ulm: No ebuild is using that function, and nobody noticed that it was broken in EAPI 7.
If someone will need it again, it may be revived with eapply. Until then, I'd rather not add any incomplete instructions how to replace it.

PR #21395, opened 2021-06-23
net-libs/NativeThread: min java 1.8

vaukai:
Bug: https://bugs.gentoo.org/787629
restricted to :1.8 as it doesn't work with java 11


vaukai:
@fordfrog  :-)

PR #21401, opened 2021-06-24
app-emulation/virt-manager: Update live ebuild

zippy2:
There was a patch merged recently into upstream repo
(v3.2.0-110-geb6b7939) that removes shebangs from library paths.
Thus we no longed need to fix them during install phase.


zippy2:
cc @trofi

trofi:
Pulled as is. Thank you!

PR #21403, opened 2021-06-24
dev-lisp/c2ffi: Bump version to 12.0.0

hjudt:


juippis (approved):
Ideally that eutils change should've been rebased to 1st commit, but oh well, it's removed in the end. Thanks!

PR #21409, opened 2021-06-24
meson.eclass: EAPI 8 support

dm0-:
@floppym Here is an update to allow EAPI 8 with meson packages (excluding the `meson-multilib.eclass`, since that can be done with all of its multilib eclass dependencies).

This updates `meson.eclass` to conform to conventions that other eclasses seem to follow.  E.g. conditional inherits are first (presumably for function precedence), and defining the inherit guard at the end.  It also removes the split inherit guard.  The only reason I saw for it is [bug #619178 comment #4](https://bugs.gentoo.org/619178#c4) with no further details.  Are there actually any examples of this being a problem?  A couple Python eclasses seem to be the only other instances of this.

### DIFF
--- a/eclass/meson.eclass
+++ b/eclass/meson.eclass
@@ -5,7 +5,7 @@
 # @MAINTAINER:
 # William Hubbs <williamh@gentoo.org>
 # Mike Gilbert <floppym@gentoo.org>
-# @SUPPORTED_EAPIS: 6 7
+# @SUPPORTED_EAPIS: 6 7 8
 # @BLURB: common ebuild functions for meson-based packages
 # @DESCRIPTION:
 # This eclass contains the default phase functions for packages which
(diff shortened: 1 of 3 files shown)


eclass/meson.eclass
```diff
@@ -35,31 +35,24 @@
 # @CODE
 
 case ${EAPI:-0} in
```
ulm, line 37:
```suggestion
case ${EAPI} in
```

ulm (approved):

eclass/meson.eclass
dm0-: Updated.  (This is another case where it seems most other eclasses are doing that.)

floppym:
I would prefer to keep EXPORT_FUNCTIONS outside the include guard. It doesn't make the eclass that much more complex, and the resulting behavior is nicer.

dm0-:
I've added a quick update to the inherited `eapi7-ver.eclass` so that it doesn't say EAPI 8 is unknown.

thesamesam (approved):

PR #21411, opened 2021-06-24
distutils-r1.eclass: Fix paths for dosym -r

ulm:
In EAPI 6, variable D has a trailing slash, so "${path#${D}}" will end up with one slash too many removed. Fix it by using "${path#${D%/}}" instead.


benkohler:
Tested and this does fix my failure in EAPI=6 catalyst ebuilds

benkohler (approved):

mgorny:
Sure, please push if it fixes the issue for you.

thesamesam:
Please commit with the Closes tags for the relevant bugs (there’s a few)

ulm:
Looks like this doesn't work with prefix.

ulm:
Closing per discussion in `#gentoo-python`. This will require more refactoring.

PR #21412, opened 2021-06-24
app-text/cmark: update to 0.30.0

winterheart:


juippis (approved):
Thanks, tests seem to pass with python-3.10 so I'll add that too.

PR #21418, opened 2021-06-25
media-sound/a2jmidid: return python support

nick87720z:
Fixes bug https://bugs.gentoo.org/798411
Also adds USE=python to make this dependency optional (just by not installing a2j_control if off)


nick87720z:
unbeatable

nick87720z:
What's point in `bug #NNNNNN` in summary if it has no effect? (from glep 66)

fordfrog:
@nick87720z your commit still misses `Signed-off-by`.

fordfrog:
@nick87720z thanks!

PR #21419, opened 2021-06-25
net-irc/konversation: Update Server default in konversationrc

ulm:


a17r (changes requested):
What about `20.12.3`?

ulm:
> What about `20.12.3`?

Stable revbump? Sure.

a17r (approved):

ulm:
Merged.

PR #21420, opened 2021-06-25
dev-python/logfury: Fix dash in setuptools variable

WPettersson:
Just changing description-file to description_file

Fixes #798057


mgorny:
Please don't patch this in Gentoo and just submit a fix upstream.

WPettersson:
Sorry, my bad.

PR #21424, opened 2021-06-25
app-shells/mksh: Switch to git mirror

lanodan:
It seems like there is an issue with cvs on the authentication,
ECVS_AUTH=ext seems specific to mksh and I don't think there is a strong
reason to keep using CVS, at the moment only 4 ebuilds are using it,
including mksh.

Draft: I asked mirabilos on the status of the github mirror.


lanodan:
Closing in preference to #21428 as the github mirror should be avoided.

PR #21425, opened 2021-06-25
net-irc/ergo: Fixed OpenRC init script + reload function

jpds:


thesamesam:
Thanks! Revbumped too.

PR #21432, opened 2021-06-25
Vulkan fixes from chromiumos-overlay

mattst88:
Some cross compiling fixes found in the chromiumos-overlay.

cc: @sarnex, @dm0-, @thesamesam, @SoapGentoo

### DIFF
--- a/dev-util/vulkan-tools/vulkan-tools-9999.ebuild
+++ b/dev-util/vulkan-tools/vulkan-tools-9999.ebuild
@@ -65,11 +65,11 @@ multilib_src_configure() {
 		-DBUILD_WSI_WAYLAND_SUPPORT=$(usex wayland)
 		-DBUILD_WSI_XCB_SUPPORT=$(usex X)
 		-DBUILD_WSI_XLIB_SUPPORT=$(usex X)
-		-DVULKAN_HEADERS_INSTALL_DIR="${EPREFIX}/usr"
+		-DVULKAN_HEADERS_INSTALL_DIR="${ESYSROOT}/usr"
 	)
 
 	use cube && mycmakeargs+=(
-		-DGLSLANG_INSTALL_DIR="${EPREFIX}/usr"
+		-DGLSLANG_INSTALL_DIR="${ESYSROOT}/usr"
 		-DCUBE_WSI_SELECTION=$(usex X XCB WAYLAND)
 	)
 
--- a/dev-util/vulkan-tools/vulkan-tools-9999.ebuild
+++ b/dev-util/vulkan-tools/vulkan-tools-9999.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
@@ -26,7 +26,7 @@ SLOT="0"
 IUSE="cube wayland +X"
 
 # Cube demo only supports one window system at a time
-REQUIRED_USE="!cube? ( || ( X wayland ) ) cube? ( ^^ ( X wayland ) )"
+REQUIRED_USE="cube? ( ^^ ( X wayland ) )"
 
 BDEPEND="${PYTHON_DEPS}
 	>=dev-util/cmake-3.10.2
--- a/media-libs/vulkan-layers/vulkan-layers-9999.ebuild
+++ b/media-libs/vulkan-layers/vulkan-layers-9999.ebuild
@@ -26,10 +26,11 @@ SLOT="0"
 IUSE="wayland X"
 
 BDEPEND=">=dev-util/cmake-3.10.2"
-DEPEND="${PYTHON_DEPS}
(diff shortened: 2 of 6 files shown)


media-libs/vulkan-loader/vulkan-loader-1.2.179.ebuild
```diff
@@ -36,6 +35,10 @@ DEPEND="${PYTHON_DEPS}
 "
 PDEPEND="layers? ( media-libs/vulkan-layers:=[${MULTILIB_USEDEP}] )"
 
+PATCHES=(
+	"${FILESDIR}"/CHROMIUM-Fix-cross-compilation.patch
```
thesamesam, line 39: Missing?

media-libs/vulkan-layers/vulkan-layers-9999.ebuild
```diff
@@ -46,9 +47,9 @@ multilib_src_configure() {
 		-DBUILD_WSI_XCB_SUPPORT=$(usex X)
 		-DBUILD_WSI_XLIB_SUPPORT=$(usex X)
 		-DBUILD_TESTS=OFF
-		-DGLSLANG_INSTALL_DIR="${EPREFIX}/usr"
-		-DCMAKE_INSTALL_INCLUDEDIR="${EPREFIX}/usr/include/vulkan/"
-		-DSPIRV_HEADERS_INSTALL_DIR="${EPREFIX}/usr/include/spirv"
+		-DGLSLANG_INSTALL_DIR="${ESYSROOT}/usr"
```
thesamesam, line 50: This doesn’t seem right for any of these cases if the variables are genuinely used for installation — we never write to ESYSROOT, we just find headers there.

May need to check if the build system is abusing these variables to find stuff OR if these changes were just added with other changes and there was confusion about which had the needed effect.

Given this package is only going to install one thing, maybe the other variables are for finding? “spirv is installed here”, not “install spirv here”?

thesamesam (approved):

thesamesam (changes requested):
(Didn’t mean to approve, on mobile)

media-libs/vulkan-loader/vulkan-loader-1.2.179.ebuild
mattst88: Oops. Didn't mean to leave that hunk in.

media-libs/vulkan-layers/vulkan-layers-9999.ebuild
mattst88: I think these variables are for informing cmake where to find various dependencies, not for specifying locations to install files from this project. I'll double check.

mattst88: Yes, I believe this is correct:

* `GLSLANG_INSTALL_DIR`: https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/CMakeLists.txt#L272
* `SPIRV_HEADERS_INSTALL_DIR`: https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/CMakeLists.txt#L268

I'm unsure about `CMAKE_INSTALL_INCLUDEDIR` though.

thesamesam: Okay, fantastic, I've checked and I agree with those two. CMAKE_INSTALL_INCLUDEDIR seems like it should genuinely be ```${EPREFIX}``` though as it's only used for installing headers to the system:
```
install(FILES ${LAYER_UTIL_FILES} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/vulkan)
```

dev-util/vulkan-tools/vulkan-tools-9999.ebuild
```diff
@@ -26,7 +26,7 @@ SLOT="0"
 IUSE="cube wayland +X"
 
 # Cube demo only supports one window system at a time
-REQUIRED_USE="!cube? ( || ( X wayland ) ) cube? ( ^^ ( X wayland ) )"
+REQUIRED_USE="cube? ( ^^ ( X wayland ) )"
```
thesamesam, line 30: I think VULKAN_HEADERS_INSTALL_DIR and GLSLANG_INSTALL_DIR need to be ESYSROOT here as per other discussions (and consistency with the others)?

thesamesam (changes requested):

dev-util/vulkan-tools/vulkan-tools-9999.ebuild
mattst88: Indeed, fixed. I also noticed that `vulkan-loader` was previously fixed in commit fce28df7e8304ff4b93b1c194a1f8cb76876b571.

media-libs/vulkan-layers/vulkan-layers-9999.ebuild
SoapGentoo: I've rechecked all these variables, and they're fine now

SoapGentoo (approved):

mattst88:
Merged. Thanks for the reviews!

PR #21434, opened 2021-06-26
Cinnamon Updates (06.25)

SparkyBluefang:
Latest Cinnamon 5.0.x updates.


gnome-extra/cinnamon/cinnamon-5.0.3.ebuild
```diff
@@ -128,16 +128,15 @@ src_configure() {
...
 	meson_src_configure
 }
 
 src_install() {
-	python_domodule files/usr/lib/python3/dist-packages/cinnamon
-	rm -rf files/usr/lib/python3 || die
-
 	meson_src_install
 
+	python_optimize "${ED}$(python_get_sitedir)"
```
SparkyBluefang, line 139: Not sure this is semantically correct, but it works correctly on my non-prefix system.

SparkyBluefang: `${ED}$(python_get_sitedir): concatenates two paths containing EPREFIX on line 139`

Presuming `python_get_sitedir` contains `EPREFIX` because it's running in `src_install`(?), should this be `${D}$(python_get_sitedir)`? Or just plain old `$(python_get_sitedir)`?

juippis (commented):
I know this seems silly, but please don't do `git mv` version bumps like this. Add a new file for each version bump. For example, this is unmergeable because (I guess?) xapps-2.2.1 got stabilized meanwhile.

SparkyBluefang:
> I know this seems silly, but please don't do `git mv` version bumps like this. Add a new file for each version bump. For example, this is unmergeable because (I guess?) xapps-2.2.1 got stabilized meanwhile.

Are you specifically talking about separating the adds and removes into separate commits? Or are you asking me to not remove the old version as part of an unstable version bump?

juippis (approved):
Yeah, so 1 commit adding a new file, 1 commit removing old version(s). I'll push this now but that's the way we prefer it.

nemo-extensions seem to have 5.0.1 release tagged upstream, that patch is a bit wierd...

SparkyBluefang:
> Yeah, so 1 commit adding a new file, 1 commit removing old version(s). I'll push this now but that's the way we prefer it.

Good to know. I'll be sure to do it that way in the future.

> nemo-extensions seem to have 5.0.1 release tagged upstream, that patch is a bit wierd...

Yeah, I was trying to do a bit of forward thinking. With all of the extensions grouped in a monolithic repo/tarball, I wanted to avoid downloading multiple full copies if I end up packaging other extensions. So I took a leaf from kernel packaging and use the repo-wide tag as a base, and patches for per-extension point releases. I'm not exactly happy with it either, but it seems like a decent compromise all things considered.

I suppose another potential strategy would be to always use the latest tag on the repo, and then update all of the extension packages at the same time (via -r# versions) to use the same tarball. But that also seems like a lot of work too.

PR #21436, opened 2021-06-26
dev-python/{grpcio, logbook, mysqlclient, pyu2f, keyring, unittest2}: add ~riscv keyword

caoye275:
Closes: https://bugs.gentoo.org/798783

Also add ~riscv keyword to some related packages


caoye275:
Closes: https://bugs.gentoo.org/798783

PR #21440, opened 2021-06-26
EAPI 8 eclass updates

dm0-:
@mgorny Here are the last few eclass updates.  The `meson-multilib.eclass` change has a dependency on #21409.

Aside from just adding EAPI 8 to the list of supported EAPIs, most of these changes fall into:
  - Dropping support for EAPIs that are not supported in inherited eclasses.
  - Moving the EAPI check before the inherit guard so using newer bash functionality won't result in syntax errors for old versions.
  - Setting the inherit guard variable before inherits so it prevents loops.
  - Trivial formatting unification where different styles are mixed in the same file.

@mattst88 @floppym @SoapGentoo

### DIFF
--- a/eclass/systemd.eclass
+++ b/eclass/systemd.eclass
@@ -4,7 +4,7 @@
 # @ECLASS: systemd.eclass
 # @MAINTAINER:
 # systemd@gentoo.org
-# @SUPPORTED_EAPIS: 0 1 2 3 4 5 6 7
+# @SUPPORTED_EAPIS: 5 6 7 8
 # @BLURB: helper functions to install systemd units
 # @DESCRIPTION:
 # This eclass provides a set of functions to install unit files for
@@ -24,14 +24,14 @@
 # }
 # @CODE
 
-inherit toolchain-funcs
-
-case ${EAPI:-0} in
-	0|1|2|3|4|5|6|7) ;;
-	*) die "${ECLASS}.eclass API in EAPI ${EAPI} not yet established."
+case ${EAPI} in
+	5|6|7|8) ;;
+	*) die "${ECLASS}: EAPI ${EAPI:-0} not supported" ;;
 esac
 
-if [[ ${EAPI:-0} == [0123456] ]]; then
+inherit toolchain-funcs
+
+if [[ ${EAPI} == [56] ]]; then
 	DEPEND="virtual/pkgconfig"
 else
 	BDEPEND="virtual/pkgconfig"
@@ -72,7 +72,6 @@ _systemd_get_systemunitdir() {
 # ${D}).  This function always succeeds, even if systemd is not
 # installed.
 systemd_get_systemunitdir() {
-	has "${EAPI:-0}" 0 1 2 && ! use prefix && EPREFIX=
 	debug-print-function ${FUNCNAME} "${@}"
 
 	echo "${EPREFIX}$(_systemd_get_systemunitdir)"
@@ -82,7 +81,7 @@ systemd_get_systemunitdir() {
 # @DESCRIPTION:
 # Deprecated alias for systemd_get_systemunitdir.
 systemd_get_unitdir() {
-	[[ ${EAPI} == [012345] ]] || die "${FUNCNAME} is banned in EAPI 6, use systemd_get_systemunitdir instead"
+	[[ ${EAPI} == 5 ]] || die "${FUNCNAME} is banned in EAPI 6, use systemd_get_systemunitdir instead"
 
 	systemd_get_systemunitdir
 }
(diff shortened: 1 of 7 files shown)


eclass/xdg-utils.eclass
```diff
@@ -86,21 +86,17 @@ xdg_icon_cache_update() {
...
+	for dir in "${EROOT%/}"/usr/share/icons/*; do
+		if [[ -f ${dir}/index.theme ]]; then
+			if ! gtk-update-icon-cache -qf "${dir}"; then
 				debug-print "Updating cache failed on ${dir}"
 				# Add to the list of failures
 				fails+=( "${dir}" )
 				retval=2
 			fi
-		elif [[ $(ls "${dir}") = "icon-theme.cache" ]]; then
+		elif [[ $(ls "${dir}") = icon-theme.cache ]]; then
```
floppym, line 99:
```suggestion
		elif [[ -e ${dir}/icon-theme.cache ]]; then
```

eclass/systemd.eclass
```diff
@@ -242,19 +234,19 @@ systemd_install_serviced() {
 # @DESCRIPTION:
 # Deprecated in favor of tmpfiles.eclass.
```
floppym, line 235: We should ban the systemd tmpfiles functions in EAPI 8.

floppym (changes requested):
Left comments for systemd and xdg-utils.

meson-multilib and udev look ok.

eclass/xdg-utils.eclass
dm0-: Is that the intent of that condition?  I thought it was testing if `icon-theme.cache` is the *only* file in the directory.

eclass/systemd.eclass
dm0-: Should they be banned, or just removed?  I asked about [bug #740638](https://bugs.gentoo.org/740638) in the gentoo-devel e-mail: is there a removal date?

eclass/xdg-utils.eclass
floppym: Ah, I see. Never mind then.

eclass/systemd.eclass
floppym: Yeah, let's remove them in that case.

dm0-: Rebased with that removal in its own commit.

dm0-:
I'll drop the deprecated tmpfiles removal here after #21462 is merged.

eclass/autotools.eclass
```diff
@@ -4,7 +4,7 @@
 # @ECLASS: autotools.eclass
```
ulm, line 4: I was told by its maintainers that there are some other changes for autotools.eclass pending, which should be done together with the EAPI 8 update. So I'd suggest that you drop it from the PR for now.

eclass/xdg-utils.eclass
```diff
@@ -7,7 +7,7 @@
 # freedesktop-bugs@gentoo.org
 # @AUTHOR:
 # Original author: Gilles Dartiguelongue <eva@gentoo.org>
-# @SUPPORTED_EAPIS: 0 1 2 3 4 5 6 7
+# @SUPPORTED_EAPIS: 0 1 2 3 4 5 6 7 8
```
ulm, line 10: Drop EAPIs 0 to 4? Keeping them doesn't make much sense, especially since `xdg.eclass` no longer supports 0 to 3 either.
```suggestion
# @SUPPORTED_EAPIS: 5 6 7 8
```

eclass/xdg-utils.eclass
```diff
@@ -17,9 +17,9 @@
 #  * XDG .desktop files cache management
 #  * XDG mime information database management
 
-case "${EAPI:-0}" in
-	0|1|2|3|4|5|6|7) ;;
-	*) die "EAPI=${EAPI} is not supported" ;;
+case ${EAPI:-0} in
+	0|1|2|3|4|5|6|7|8) ;;
```
ulm, line 21:
```suggestion
case ${EAPI} in
	5|6|7|8) ;;
```

eclass/xdg-utils.eclass
```diff
@@ -86,21 +86,17 @@ xdg_icon_cache_update() {
...
-			rv=$?
-			if [[ ! $rv -eq 0 ]] ; then
+	local dir retval=0
+	local fails=()
+	for dir in "${EROOT%/}"/usr/share/icons/*; do
+		if [[ -f ${dir}/index.theme ]]; then
+			if ! gtk-update-icon-cache -qf "${dir}"; then
 				debug-print "Updating cache failed on ${dir}"
 				# Add to the list of failures
 				fails+=( "${dir}" )
```
ulm, line 96: For consistency, either do
```suggestion
				fails+=("${dir}")
```
here, or keep the blank in `local fails=( )` above.

ulm (changes requested):

ulm (approved):
LGTM

eclass/autotools.eclass
ulm: I know. Time frame is about one month (or so I was told).

dm0-:
I dropped the deprecation cleanup commit since #21462 was merged and the conflict was resolved.

mgorny (approved):
LGTM.

PR #21443, opened 2021-06-27
sys-apps/miller-5.10.0: build html docs when doc flag is enabled

mehw:
Hello,

You may consider this as a proof of concept to build the html docs when the `doc` use flag is enabled.  The rebuild of man pages is kept disabled, though.

New files used to build the html docs are `sort-within-records.json`, which is required but missing in miller-5.10.0, and the shell script `creach`, which is a replacement to [John Kerl`s creach perl script](https://github.com/johnkerl/scripts/blob/master/fundam/creach).

[https://johnkerl.org/miller/doc/index.html](https://johnkerl.org/miller/doc/index.html) redirects to the home page [https://miller.readthedocs.io](https://miller.readthedocs.io), hence `HOMEPAGE` is changed accrodingly.


sys-apps/miller/miller-5.10.0.ebuild
```diff
@@ -34,13 +52,21 @@ src_prepare() {
...
+}
+
 src_test() {
 	emake -C c/reg_test
 	emake -C c/unit_test
 }
 
 src_install() {
-	local HTML_DOCS=( $(usev doc) )
+	local HTML_DOCS=( $(usex doc docs/_build/html/.) )
```
juippis, line 69: You might want `use doc && local HTML_DOCS=( docs/_build/html/ )`

```
  File "/usr/lib/portage/python3.10/doins.py", line 594, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/usr/lib/portage/python3.10/doins.py", line 582, in main
    if _doins(
  File "/usr/lib/portage/python3.10/doins.py", line 434, in _doins
    return install_runner.install_file(source, os.path.dirname(dest))
  File "/usr/lib/portage/python3.10/doins.py", line 370, in install_file
    return self._ins_runner.run(source, dest_dir)
  File "/usr/lib/portage/python3.10/doins.py", line 179, in run
    sstat = os.stat(source)
FileNotFoundError: [Errno 2] No such file or directory: b'no'
 * ERROR: sys-apps/miller-5.10.0::gentoo failed (install phase):
 *   dodoc failed
```

mehw: Thanks ;)

Since you modified the file, there's no need for me to force a push, right?

juippis: No that's just a suggestion based on a feeling, you should try to figure out why it fails and if that solution helps at all :P I didn't test it.

PR #21448, opened 2021-06-27
Command -v stdout fix

hlein:
command -v writes to stdout, not stderr. Fixed the instance causing emails from cron (app-antivirus/clamav/files/clamav-milter.logrotate) and another instance whose output was coincidentally already being eaten (mail-filter/spamassassin/files/update-spamassassin-rules-r1.cron). Did not -rN bump the latter since it's harmless as-installed, only bad as an incorrect example in the tree to be copied by others.


orlitzky:
Thumbs up for clamav from me, thanks!

LordVan:
looks good to me. Thanks

Philippe23:
Looks good to me from the SpamAssassin side as well.

PR #21450, opened 2021-06-27
net-im/telegram-desktop: Bump to 2.8.1

mid-kid:
@gyakovlev

Some fundamental changes in this one:

media-libs/rnnoise is introduced as a dependency (~ppc64 dropped from net-im/telegram-desktop, please re-add as appropriate)

media-libs/tg_owt has a huge update adding support for desktop capturing, which requires alsa and pulseaudio again. This has spurred a few patches to soften the dependencies on pipewire, pulseaudio and x11, but upstream support for configurations like these should be considered spotty at best. I'm not exacly sure why upstream [allows disabling both audio backends but not just pulse](https://github.com/desktop-app/tg_owt/issues/63), and why disabling pipewire is currently broken despite the option being available (requiring a patch), so I'm hoping for the best but prepared for the worst. Upstream webrtc seems to support these configurations, at least.

Along with the huge tg_owt update is ABI breakage, requiring media-libs/libtgvoip to be rebuilt. I'm not sure if I did this correctly, but I did the following:
* Add a subslot to media-libs/tg_owt
* Revbump media-libs/libtgvoip to depend on the subslot (might be possible to straight-to-stable, but doesn't matter much)
* Made net-im/telegram-desktop depend on this new revision or higher


mid-kid:
Oh, and it should be noted that I've not gotten around to testing desktop capturing at all yet, as I don't want to bother my friends and making a 2nd telegram account for testing being a major pain in the ass.

media-libs/tg_owt/tg_owt-0_pre20210626.ebuild
```diff
@@ -0,0 +1,104 @@
...
+
+TG_OWT_COMMIT="91d836dc84a16584c6ac52b36c04c0de504d9c34"
+LIBYUV_COMMIT="ad890067f661dc747a975bc55ba3767fe30d4452"
+SRC_URI="https://github.com/desktop-app/tg_owt/archive/${TG_OWT_COMMIT}.tar.gz -> ${P}.tar.gz
+	https://archive.org/download/libyuv-${LIBYUV_COMMIT}.tar/libyuv-${LIBYUV_COMMIT}.tar.gz"
+S="${WORKDIR}/${PN}-${TG_OWT_COMMIT}"
+# Fetch libyuv archive from: https://chromium.googlesource.com/libyuv/libyuv/+archive/${LIBYUV_COMMIT}.tar.gz
+
+LICENSE="BSD"
+SLOT="0/20210626"
```
gyakovlev, line 19: `SLOT="0/${PV##*pre}"`
is likely better.

I can fix on merge

media-libs/tg_owt/tg_owt-0_pre20210626.ebuild
```diff
@@ -0,0 +1,104 @@
...
+"
+RDEPEND="${DEPEND}"
+BDEPEND="virtual/pkgconfig"
+
+PATCHES=(
+	"${FILESDIR}/tg_owt-0_pre20210604-allow-disabling-pipewire.patch"
+	"${FILESDIR}/tg_owt-0_pre20210604-allow-disabling-X11.patch"
+	"${FILESDIR}/tg_owt-0_pre20210604-allow-disabling-pulseaudio.patch"
+	"${FILESDIR}/tg_owt-0_pre20210604-expose-set_allow_pipewire.patch"
+)
```
gyakovlev, line 65: all those patch names are wrong.

should be `20210626` and not `20210604`.
can fix on merge.

gyakovlev:
fixed outstanding issues and merged, thanks again.
also restored ppc64 keyword, so no bugs from you needed.

media-libs/tg_owt/tg_owt-0_pre20210626.ebuild
mid-kid: whoops, forgot fixing that after renaming them... Thanks

mid-kid:
Thanks!

PR #21374, opened 2021-06-22
dev-python/graph-tool: bump to 2.39

gerion0:
Patch is backported from
https://git.skewed.de/count0/graph-tool/-/commit/ee5cd8e8da8c35b56694d03eb9d63fb30b61702e

Closes: https://bugs.gentoo.org/795603

The entire ebuild is _not_ well tested. I have just compiled it and it can be imported within Python without a crash. Currently, when linking against the C++ headers, I got compilation errors, will evaluate this further.

PR #21454, opened 2021-06-28
dev-python/graph-tool: bump to 2.40

gerion0:
Closes: https://bugs.gentoo.org/795603
Signed-off-by: Gerion Entrup <gerion.entrup@flump.de>

Compilation and general working tested. This version is compatible with boost-1.76 (the current version in the tree is not).

graph-tool needs _lots_ of RAM during compilation, several GB per compilation unit. Some versions ago graph-tool was compiled with "-j1" (see the old comment in the ebuild). However, this changed with the ebuild rewrote to the new Python eclass (I guess). I guess, it would be meaningful to compile with -j1 again.

A better method would some calculation how much RAM is free and a simple heuristic (`-j $(free_ram/7GB)` for example), how much units can be compiled in parallel. I can measure the RAM usage together with the parallelisation, if you think, such an heuristic is meaningful.


gerion0:
See #21374

dev-python/graph-tool/graph-tool-2.40.ebuild
```diff
@@ -0,0 +1,84 @@
...
+
+LICENSE="GPL-3"
+SLOT="0"
+IUSE="+cairo openmp"
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+RDEPEND="
+	${PYTHON_DEPS}
+	>=dev-libs/boost-1.70:=[context,python,${PYTHON_USEDEP}]
+	dev-libs/expat:=
```
thesamesam, line 29:
```suggestion
	dev-libs/expat
```

We link against expat but it doesn't expose a subslot, so let's drop the operator as we don't know what it might be used for in future.

dev-python/graph-tool/graph-tool-2.40.ebuild
```diff
@@ -0,0 +1,84 @@
...
+	dev-python/numpy[${PYTHON_USEDEP}]
+	dev-python/scipy[${PYTHON_USEDEP}]
+	sci-mathematics/cgal:=
+	cairo? (
+		dev-cpp/cairomm:0
+		dev-python/pycairo[${PYTHON_USEDEP}]
+	)
+	dev-python/matplotlib[${PYTHON_USEDEP}]"
```
thesamesam, line 37: Nit: move this above the conditional (put conditional deps last).

dev-python/graph-tool/graph-tool-2.40.ebuild
```diff
@@ -0,0 +1,84 @@
...
+}
+
+src_prepare() {
+	default
+	[[ ${PV} == *9999 ]] && eautoreconf
+	python_copy_sources
+}
+
+src_configure() {
+	configure() {
```
thesamesam, line 61:
```suggestion
	myconfigure() {
```
or similar?

dev-python/graph-tool/graph-tool-2.40.ebuild
```diff
@@ -0,0 +1,84 @@
...
+	python_foreach_impl run_in_build_dir configure
+}
+
+src_compile() {
+	# most machines don't have enough ram for parallel builds
+	python_foreach_impl run_in_build_dir default
+}
+
+src_install() {
+	python_install() {
```
thesamesam, line 77:
```suggestion
	my_python_install() {
```

Too similar to a genuine phase/function name from some Python eclasses.

dev-python/graph-tool/graph-tool-2.40.ebuild
```diff
@@ -0,0 +1,84 @@
...
+			--disable-static \
+			$(use_enable openmp) \
+			$(use_enable cairo) \
+			--with-boost-python="boost_${EPYTHON/./}"
+	}
+	python_foreach_impl run_in_build_dir configure
+}
+
+src_compile() {
+	# most machines don't have enough ram for parallel builds
```
thesamesam, line 72: This comment seems obsolete now we dropped -j1?

thesamesam (changes requested):

dev-python/graph-tool/graph-tool-2.40.ebuild
gerion0: I would prefer to leave the comment as is and force -j1 again :). Is this possible?

gerion0:
Thanks for the review. I have splitted the "-j1" in an extra commit. As written, I would prefer some kind of metric here to enable partial compilation, but every metric will result in smart guessing. Not sure, if you want that.

PR #21455, opened 2021-06-28
sys-apps/bleachbit: Bump to 4.4.0

tietzc:
Build log with tests enabled: https://dpaste.com/FN8VAM7YG


ionenwks (approved):
Thank you for keeping tests in a good shape despite the restriction.

Passed for me with python3.10 too.

PR #21459, opened 2021-06-28
dev-lua/luaposix: add virtual/libcrypt:= to DEPEND

winterheart:
Fix Portage QA warning.
Closes: https://bugs.gentoo.org/799050


thesamesam:
Thanks! Needs a revbump though: https://wiki.gentoo.org/wiki/Project:Toolchain/libcrypt_implementation#Developer_information.

(It's changing (R)DEPEND and it needs to be rebuilt in future. It's very likely that libcrypt is on the system already, so we can't assume it wouldn't build.)

thesamesam:
Thanks!

PR #21462, opened 2021-06-28
systemd.eclass: drop deprecated tmpfiles functions

dm0-:
@floppym So oddly, the old versions of these packages used `tmpfiles.eclass` correctly and then downgraded to the deprecated functions in newer versions.  I omitted `TMPFILES_OPTIONAL=1` on them because the old versions didn't use it, but I did add it to the eclass since all other glibc versions use it due to [bug #774855](https://bugs.gentoo.org/774855).

### DIFF
--- a/eclass/systemd.eclass
+++ b/eclass/systemd.eclass
@@ -237,48 +237,6 @@ systemd_install_serviced() {
 	)
 }
 
-# @FUNCTION: systemd_dotmpfilesd
-# @USAGE: <tmpfilesd>...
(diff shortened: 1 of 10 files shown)


floppym (approved):

fordfrog:
i'm fine with the changes to `net-analyzer/zabbix`. thanks!

zmedico (approved):
Looks great!

akhuettel:
fine with toolchain

PR #21464, opened 2021-06-28
games-rpg/{draci-historie,drascula,dreamweb,gwiz,soltys}: EAPI8, ebuild improvements

mm1ke:
Hi,

This PR updates some `games-rpg` ebuilds. Some notes:
* some ebuilds have some big `if .. then` blocks to control the installation of the language. While I don't have any strong opinion about that i think it would look nicer and more compact to have it like in the ebuild `drascula` to install them.
* all ebuilds were updated to `EAPI=7` but i see that `EAPI=8` is getting more and more used so i think updating them straight to `EAPI=8` would be a good idea too? Any thoughts about this?

@ionenwks @SoapGentoo


ionenwks:
> some ebuilds have some big `if .. then` blocks to control the installation of the language. While I don't have any strong opinion about that i think it would look nicer and more compact to have it like in the ebuild `drascula` to install them.

Yes, this wall of copy/paste is looking pretty worthless. But rather than drascula's, [l10n.eclass](https://devmanual.gentoo.org/eclass-reference/l10n.eclass/index.html) notably has a `l10n_for_each_locale_do` which could call a custom function (not that it's an eclass I'm very familiar with, never needed it so far).

It's okay if not, but if you're interested in trying to improve these it'd be welcome. Have a tendency to semi-rewrite most of those old game ebuilds when I touch them myself, some of them were barely improved since 2003 and their original maintainer is gone.

> all ebuilds were updated to `EAPI=7` but i see that `EAPI=8` is getting more and more used so i think updating them straight to `EAPI=8` would be a good idea too? Any thoughts about this?

Early adoption may not always be the best thing to do, but we're talking about old ~arch-only game ebuilds here that may not be touched again for a long time, I don't see a reason not to start using 8 wherever the eclasses allow it (using l10n.eclass would notably mean no eapi8 yet, but that's fine).

I notably started doing this in my [last small batch of commits](https://github.com/gentoo/gentoo/compare/8c2bdf0a74b2ad651b7b362c1d4a4480e07b311e...e2d347fd03422da521dd933def9982f63b5b889f) anyhow, not that many devs jumped into this yet.

games-rpg/draci-historie/draci-historie-2012-r1.ebuild
```diff
@@ -56,24 +56,28 @@ src_prepare() {
...
 	for lingua in $(find * -type d); do
 		doins -r ${lingua}
 	done
-	if use l10n_en || ( ! use l10n_cs && ! use l10n_de && ! use l10n_en && ! use l10n_pl ) ; then
-		make_wrapper draci-historie-en "scummvm -f -p \"/usr/share/${PN}/en\" draci" .
-		make_desktop_entry ${PN}-en "Dračí Historie (English)" /usr/share/pixmaps/draci-historie.ico
+
+	if use l10n_en || ( ! use l10n_cs && ! use l10n_de && ! use l10n_en && ! use l10n_pl ); then
+		make_wrapper draci-historie-en "scummvm -f -p \"${EPREFIX}/usr/share/${PN}/en\" draci" .
+		make_desktop_entry ${PN}-en "Dračí Historie (English)" "${EPREFIX}/usr/share/pixmaps/draci-historie.ico"
```
ionenwks, line 67: Not a full review but this bothers me. Nice to have thought of prefix but this whole thing was wrong to start with, not only should desktop entries not have full path to icons, but icons other than png, xpm, and svg are all invalid as far as XDG is concerned and won't be displayed in most cases.

Optimally the idea would be to convert them and add them to SRC_URI but I've personally been lazy with these and just use the generic "applications-games" icon (i.e. `make_desktop_entry name Name applications-games`) for now and remove bad icons. Going through all of these and eventually giving them proper icons in my devspace is a TODO I'm keeping for later (for old games anyway).

ionenwks: Actually, if any icons are bad / missing I'll fix these on merge now -- don't want to get too backlogged on this so started adding them to devspace now.

mm1ke:
> > some ebuilds have some big `if .. then` blocks to control the installation of the language. While I don't have any strong opinion about that i think it would look nicer and more compact to have it like in the ebuild `drascula` to install them.
> 
> Yes, this wall of copy/paste is looking pretty worthless. But rather than drascula's, [l10n.eclass](https://devmanual.gentoo.org/eclass-reference/l10n.eclass/index.html) notably has a `l10n_for_each_locale_do` which could call a custom function (it seems soon to be replaced by not-yet-in-tree plocale.eclass, but easy conversion so don't worry about that).
> **Edit:** on side-note, single-use custom functions should typically be defined inside the phase function, still with some namespace like packagename/abbreviation_someaction()
> 
> It's okay if not, but if you're interested in trying to improve these it'd be welcome. Have a tendency to semi-rewrite most of those old game ebuilds when I touch them myself, some of them were barely improved since 2003 and their original maintainer is gone -- sometimes even broken by drive-by / games.eclass migration too.
> **Edit:** when doing some more major refactoring, I'd suggest comparing image/filelists before and after / with and without all USEs to see if everything's normal

I've now looked into `l10n.eclass` and tried to use the `l10n_for_each_locale_do` function. However i've decided to choose another solution (more on that later). 
As far as i could see `l10n.eclass` works only with the `LINGUAS` variable, which means the user would need to set wanted languages in `make.conf` (don't know if it's possible per package). This would also make the usage of `l10n_*` USE-Flags obsolete (since they are controlled by the `L10N` variable) and i would need to initially download all versions of the game and decide later which version i should install (based on the `LINGUAS` setting - which i think is also not easily done too).

Mainly I didn't like that because for two reason's:
* for users it's not really visible that there are even different langues available (since there are no `IUSE` settings) for this game.
* unless there is as way to set `LINGUAS` variable per package, users are forced to install the language version of the game based on the global setting.

However, my other solution works similar like the function `l10n_for_each_locale_do`, but with some major advantages:
* it works with `IUSE` settings
* doesn't require `l10n.eclass`, thus `EAPI=8`
* is almost as simple as with `l10n_for_each_locale_do`

In order to make it even more simple i've also added `REQUIRED_USE` for the languages. One language needs to be decided anyway and now users get more aware of this. One thing i'm not so sure about is if i should set a default for the language via `+l10n_en` (like before, were `en` was used too if nothing was set). Any opinions about that?
Since a default would be set anyway as long as a languages are set in the `L10N` variable i hesitated.

One more thing to note: I've added a `for .. do` loop to control the installed versions with passing `IUSE` there. This only works because the `draci-historie` ebuild has _only_ `l10n_*` use flags. I hope this is fine, otherwise i would do something like this to highlight the difference:
```
LANGS_IUSE="l10n_cs l10n_de l10n_en l10n_pl"
IUSE="${LANGS_IUSE}"
.
.
for loc in ${LANGS_IUSE}; do
...
```

I've also checked the installed files with the original ebuild and afaics it's the same except for the `ico` file which i'm now already remove in `src_prepare`

I think this is a better solutions as with `l10n` but i'm open for other solutions, which is why i've updated only `drace-historie` for now.

games-rpg/draci-historie/draci-historie-2012-r1.ebuild
mm1ke: Thanks for doing that. As you suggested i've now simply use the ` applications-games` placeholder (only for `draci-historie` now, but going to update the others too) :)

ionenwks:
Haven't reviewed the changes yet, but just to reply.

> As far as i could see `l10n.eclass` works only with the `LINGUAS` variable [...]

Sorry for making you waste time on checking this, as I mentioned I didn't have a close look at it myself yet.

Thought we did have an eclass for generic l10n_ handling, but looking closer every ebuild is doing their own thing -- so it makes sense too here then. And, indeed I didn't mean to use LINGUAS at all here, so there's no need to convince me.

> In order to make it even more simple i've also added `REQUIRED_USE` for the languages. One language needs to be decided anyway and now users get more aware of this. One thing i'm not so sure about is if i should set a default for the language via `+l10n_en` (like before, were `en` was used too if nothing was set). Any opinions about that?

Using REQUIRED_USE sounds good to me, but yes, l10n_en should be default given not every users set L10N and then they won't be able to emerge the package without setting it (annoyance for them). There's no profile default for L10N (all disabled) unlike there is for PYTHON_TARGETS and the like.

ionenwks:
> Using REQUIRED_USE sounds good to me, but yes, l10n_en should be default given not every users set L10N and then they won't be able to emerge the package without setting it (annoyance for them). There's no profile default for L10N (all disabled) unlike there is for PYTHON_TARGETS and the like.

Although using REQUIRED_USE does have some disadvantages, making it default is nice but if users set say.. L10N="nl" it'll disable the default and now you have no language if _nl isn't provided as well, forcing them to add "en".

mm1ke:
> Haven't reviewed the changes yet, but just to reply.
> 
> > As far as i could see `l10n.eclass` works only with the `LINGUAS` variable [...]
> 
> Sorry for making you waste time on checking this, as I mentioned I didn't have a close look at it myself yet.
> 
> Thought we did have an eclass for generic l10n_ handling, but looking closer every ebuild is doing their own thing -- so it makes sense too here then. And, indeed I didn't mean to use LINGUAS at all here, so there's no need to convince me.
> 
No Problem. :)
I don't mind trying out different solutions and in this case i even got to know the `l10n.eclass` a bit. ;)

mm1ke:
> > Using REQUIRED_USE sounds good to me, but yes, l10n_en should be default given not every users set L10N and then they won't be able to emerge the package without setting it (annoyance for them). There's no profile default for L10N (all disabled) unlike there is for PYTHON_TARGETS and the like.
> 
> Although using REQUIRED_USE does have some disadvantages, making it default is nice but if users set say.. L10N="nl" it'll disable the default and now you have no language if _nl isn't provided as well, forcing them to add "en".

Hmm, you're right. It's a pity that the default setting with `+l10n_en` doesn't work in this case :/
I'm sure i can work around this (i'll remove `REQUIRED_USE` and defaulting to `en` unconditionally again)

I'll update my PR tomorrow, it's already late, but thanks again for the hints.

games-rpg/draci-historie/draci-historie-2012-r1.ebuild
```diff
@@ -1,79 +1,60 @@
...
+			mv *.{dfw,fon,mid,sam} ${loc}/ || die
+		fi
+	done
 }
 
 src_prepare() {
 	default
-	rm -f *.{bat,exe,ins} readme.* || die
+	rm -f *.{bat,exe,ins,ico} readme.* || die
 }
```
ionenwks, line 42: Hm, I was going to mention removing the -f, but I don't see the use for this src_prepare. Only the ${loc}/ directories are installed, and cleaning those seem pointless.

games-rpg/draci-historie/draci-historie-2012-r1.ebuild
```diff
@@ -1,79 +1,60 @@
...
 
 src_install() {
-	newicon bert.ico draci-historie.ico
 	insinto /usr/share/${PN}
+	local lingua
 	for lingua in $(find * -type d); do
 		doins -r ${lingua}
 	done
```
ionenwks, line 49: Since we have our own loop, just remove this block and add a `doins -r ${loc}` in the new one.

games-rpg/draci-historie/draci-historie-2012-r1.ebuild
```diff
@@ -1,79 +1,60 @@
...
+
+	local loc
+	for loc in cs de en pl; do
+		if use l10n_${loc}; then
+			make_wrapper draci-historie-${loc} "scummvm -f -p \"${EPREFIX}/usr/share/${PN}/${loc}\" draci" .
+			make_desktop_entry ${PN}-${loc} "Dračí Historie (${loc})" applications-games
+		fi
+	done
+
 	einstalldocs
```
ionenwks, line 59: This einstalldocs isn't installing anything, you can drop it.

The readme.txt in the .zip doesn't seem to really apply to using this with our packaging and is kinda messy (src_prepare was deleting it), I don't think it's worth trying to preserve it.

ionenwks (commented):
Only had a quick look at draci-historie to give potential ideas, I'll check the rest after the next set of updates. Thanks for doing this :+1:

games-rpg/draci-historie/draci-historie-2012-r1.ebuild
ionenwks: You can add the icon now if you want (similar to the old .ico):
`https://dev.gentoo.org/~ionen/distfiles/${PN}.png` (draci-historie.png)
`doicon "${DISTDIR}"/${PN}.png`
And update Manifest.

For make_desktop_entry, remove the applications-games placeholder it'll use ${PN}.png by default.

In this batch, was there any other missing/wrong? I see xpm and png on others, and that sounds okay (assuming they are square, I've had a case where the icon was 32x33 and it wouldn't display :eyes:).

mm1ke:
@ionenwks 
Some more updates. I've included all your suggestions and used your `MY_L10N` usage (thx btw for the example).
Additional notes:
* for `dreamweb` i had to adapt the languages names (en -> us, en_GB -> uk) so that they work with the filenames. Other solutions would probably to be renaming the `.zip` files.
* I've also fixed `drascula`. The `dat` file was too old (used newer `dat` file) and music were missing (scummvm[vorbis]). However since i'm using different files now i guess a revision bump would be required.
* The same would apply for `draci-historie` i guess? Since we are installing a new icon file i think we should make a revision bump?

mm1ke:
One more note. `games-rpg/eschalon-book-1-demo` segfaults when trying to start. I've tried a few things to see if i could fix it but without luck. Since i don't have a 32bit environment i don't know if its a multilib issue either. My plan now is to remove it from the PR. Maybe it's simply too old and doesn't work with newer packages (treeclean candidate?)

games-rpg/draci-historie/draci-historie-2012-r1.ebuild
mm1ke: I've checked all icons and the others seems to be fine except for `games-rpg/eschalon-book-1-demo` (which doesn't seems to have a icon). I guess using `applications-games` would be the right way to fix it (or getting a icon for this game) but since it doesn't even start (see comment below) i haven't touched the ebuild for now.

ionenwks:
> I've also fixed `drascula`. The `dat` file was too old (used newer `dat` file) and music were missing (scummvm[vorbis]). However since i'm using different files now i guess a revision bump would be required.

yep, adding [vorbis] also warrant a revbump on its own, dynamic deps are unreliable and a revbump will ensure the requirement is known on users' systems

> The same would apply for `draci-historie` i guess? Since we are installing a new icon file i think we should make a revision bump?

More of a grey area here, I don't typically consider a .desktop icon relevant enough for a revbump, opinions may differ. 

> One more note. `games-rpg/eschalon-book-1-demo` segfaults when trying to start. I've tried a few things to see if i could fix it but without luck. Since i don't have a 32bit environment i don't know if its a multilib issue either. My plan now is to remove it from the PR. Maybe it's simply too old and doesn't work with newer packages (treeclean candidate?)

https://www.gog.com/forum/eschalon_series/linux_crash_on_startup/post5

Someone attempted a fix but it needs updating and is very fragile, and alternative is no good. Sounds like it'd be easy for upstream to fix but doesn't look like they've really done anything in years.

I'll make a note of it and throw it in a last-rite batch later. Don't waste more time with it.

mm1ke:
> > I've also fixed `drascula`. The `dat` file was too old (used newer `dat` file) and music were missing (scummvm[vorbis]). However since i'm using different files now i guess a revision bump would be required.
> 
> yep, adding [vorbis] also warrant a revbump on its own, dynamic deps are unreliable and a revbump will ensure the requirement is known on users' systems

Done, I've split the commit to reflect this (revbump + removal). Keeping the old is probably not required, especially since it has no stable keywords.
> 
> > The same would apply for `draci-historie` i guess? Since we are installing a new icon file i think we should make a revision bump?
> 
> More of a grey area here, I don't typically consider a .desktop icon relevant enough for a revbump, opinions may differ.

Ok, i'll keep it then :)
> 
> > One more note. `games-rpg/eschalon-book-1-demo` segfaults when trying to start. I've tried a few things to see if i could fix it but without luck. Since i don't have a 32bit environment i don't know if its a multilib issue either. My plan now is to remove it from the PR. Maybe it's simply too old and doesn't work with newer packages (treeclean candidate?)
> 
> https://www.gog.com/forum/eschalon_series/linux_crash_on_startup/post5
> 
> Someone attempted a fix but it needs updating and is very fragile, and alternative is no good. Sounds like it'd be easy for upstream to fix but doesn't look like they've really done anything in years. If someone really want to play this I'd suggest they use Wine at this point.
> 
> I'll make a note of it and throw it in a last-rite batch later. Don't waste more time with it.

Agreed and removed from the PR

games-rpg/dreamweb/dreamweb-1.1-r1.ebuild
```diff
@@ -16,50 +17,30 @@ SRC_URI="doc? ( mirror://sourceforge/scummvm/${PN}-manuals-en-highres.zip )
...
-	fi
-	if use doc ; then
+	MY_L10N=( $(usev l10n_de) $(usev l10n_es) $(usev l10n_fr) $(usev l10n_it) )
+	use l10n_en && MY_L10N+=( l10n_us )
+	use l10n_en-GB && MY_L10N+=( l10n_uk )
+	[[ ${MY_L10N} ]] || MY_L10N=( l10n_us )
+
+	local lang
+	for lang in "${MY_L10N[@]//l10n_/}"; do
+		mkdir -p "${S}"/${lang} || die
```
ionenwks, line 38: Drop the `-p` everywhere (including other ebuilds), it would be strange if "${S}" didn't already exist and this was somehow needed.

games-rpg/soltys/soltys-1.0-r2.ebuild
```diff
@@ -1,64 +1,52 @@
...
-	if use l10n_pl ; then
-		make_wrapper soltys-pl "scummvm -f -p \"/usr/share/${PN}/pl\" soltys" .
-		make_desktop_entry ${PN}-pl "Soltys (Polski)" soltys
-	fi
+	newicon "${DISTDIR}"/cat-${PN}.png ${PN}.png
+
+	local lang
+	for lang in "${MY_L10N[@]//l10n_/}"; do
+		make_wrapper ${PN}-${lang} "scummvm -f -p \"${EPREFIX}/usr/share/${PN}/${lang}\" ${PN}" .
+		make_desktop_entry ${PN}-${lang} "Soltys (${lang})" ${PN}
```
ionenwks, line 50: The third argument on either of those isn't needed for any of these ebuilds.
- `.` makes the wrapper change directory.. to the current dir, unspecifying does nothing which is current dir
- `${PN}` is the default for icon if unspecified

ionenwks (commented):
More full review this time, new ebuilds sure look a lot nicer.

Drascula's commit message could use a small note explaining why it's being revbumped (vorbis/data).

games-rpg/dreamweb/dreamweb-1.1-r1.ebuild
```diff
@@ -68,42 +49,19 @@ src_unpack() {
 
 src_prepare() {
 	default
-	rm -rf */license.txt */*.EXE || die
+	rm -f */license.txt */*.EXE || die
 }
```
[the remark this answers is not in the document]
mm1ke: AFAICS this is required, unless it's fine to install two EXE files and this `licesne.txt` file.

mm1ke:
@ionenwks 
Thanks again for the detailed review. 
I've addressed all you comments and i hope i've fixed all. I've removed a few more case were the third argument shouldn't be needed for `make_wrapper` and `make_desktop_entry`.
Also thanks for the hints (usev) and suggestions. I've also compiled tested all ebuilds with all languages enabled.

games-rpg/dreamweb/dreamweb-1.1-r1.ebuild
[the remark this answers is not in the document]
ionenwks: Ah yes, this one is different, good catch. I didn't notice these were going inside `*/directories`

ionenwks:
Merged as-is, thanks :)

Much better than the old ebuilds.

PR #21465, opened 2021-06-28
dev-libs/libphonenumber: add phone supports

scardracs:
evolution-data-server has phone's ability thanks to libphonenumber and it's required for mobile's phones/tablets like Librem 5, PinePhone and PineTab. Adding it.


gnome-extra/evolution-data-server/evolution-data-server-3.40.2-r1.ebuild
```diff
@@ -0,0 +1,153 @@
...
+	use berkdb && append-cppflags "-I$(db_includedir)"
+
+	local google_enable
+	if use oauth || use gnome-online-accounts; then
+		google_enable="ON"
+	else
+		google_enable="OFF"
+	fi
+
+	# phonenumber does not exist in tree
```
Arfrever, line 107: ^^

gnome-extra/evolution-data-server/evolution-data-server-3.40.2-r1.ebuild
```diff
@@ -0,0 +1,153 @@
...
+		-DSYSCONF_INSTALL_DIR="${EPREFIX}"/etc
+		-DENABLE_GTK_DOC=$(usex gtk-doc)
+		-DWITH_PRIVATE_DOCS=$(usex gtk-doc)
+		-DENABLE_SCHEMAS_COMPILE=OFF
+		-DENABLE_INTROSPECTION=$(usex introspection)
+		-DWITH_KRB5=$(usex kerberos)
+		-DWITH_KRB5_INCLUDES=$(usex kerberos "${EPREFIX}"/usr "")
+		-DWITH_KRB5_LIBS=$(usex kerberos "${EPREFIX}"/usr/$(get_libdir) "")
+		-DWITH_OPENLDAP=$(usex ldap)
+		-DWITH_PHONENUMBER=OFF
```
Arfrever, line 118: ^^

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
```diff
@@ -0,0 +1,32 @@
+# Copyright 1999-2021 Gentoo Authors
```
Arfrever, line 1: `# Copyright 2021 Gentoo Authors`

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
```diff
@@ -0,0 +1,32 @@
...
+HOMEPAGE="https://github.com/google/libphonenumber"
+SRC_URI="https://github.com/google/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+KEYWORDS="~amd64 ~arm64"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+
+RDEPEND="
+	dev-libs/boost
```
Arfrever, line 20: `dev-libs/boost:=`

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
```diff
@@ -0,0 +1,32 @@
...
+	dev-libs/icu
```
Arfrever, line 21: `dev-libs/icu:=`

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
```diff
@@ -0,0 +1,32 @@
...
+	dev-libs/protobuf
```
Arfrever, line 22: `dev-libs/protobuf:=`

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
```diff
@@ -0,0 +1,32 @@
...
+"
+
+BDEPEND="
+	dev-cpp/gtest
```
Arfrever, line 26: `dev-cpp/gtest` is C++ library for testing, so when it is used it should be in `DEPEND` and usually in `test? ( ... )` conditional.
But with `RESTRICT="test"`, maybe disable any usage of `dev-cpp/gtest` and leave `dev-cpp/gtest` only in comment.
You also need `DEPEND` anyway:
```
DEPEND="${RDEPEND}"
# test? ( dev-cpp/gtest )

File: gnome-extra/evolution-data-server/metadata.xml [hunk 8]
```diff
@@ -12,5 +12,6 @@
 			If gnome-online-accounts is enabled and used, this is not necessary,
 			but both can be supported at the same time with different setup at runtime</flag>
 		<flag name="weather">Enable optional weather calendar support</flag>
+		<flag name="phonenumber">Enable phone supports</flag>
```
ajakk, line 15: The rest of the <flag> tags are sorted by the name attribute, might as well keep the sorting

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
scardracs: I have always used 1999-2021 as both skel.ebuild and header.txt reports it

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
scardracs: Does not build without great, that's why it is on BDEPEND (as it is required only during build time)

profiles/arch/alpha/package.use.mask
```diff
@@ -1,6 +1,10 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Marco Scardovi <marco@scardovi.com> (2021-06-28)
```
ajakk, line 4: Instead of masking, it's better to drop arches where you've not tested and then file a rekeywording bug after the affected package is in tree. This eases arch testing because tooling is already built around working with an unkeyworded package, and some automated tooling may not know to unmask.

ajakk (changes requested):

profiles/arch/alpha/package.use.mask
scardracs: You mean drop/rekeyword evolution-data-server?

gnome-extra/evolution-data-server/evolution-data-server-3.40.2-r1.ebuild
scardracs: Oh, I've completely forgot to update it

ajakk:
On Mon, Jun 28, 2021 at 03:30:31PM -0700, Marco Scardovi wrote:
> @scardracs commented on this pull request.
>
>
>
> > @@ -1,6 +1,10 @@
>  # Copyright 1999-2021 Gentoo Authors
>  # Distributed under the terms of the GNU General Public License v2
>
> +# Marco Scardovi ***@***.***> (2021-06-28)
>
> You mean drop/rekeyword evolution-data-server?

Drop the keywords in the revbump, and remove the masks since the keyword
dropping achieves the same desired effect as masking.

Once in tree, the keywords can be added again via the normal
keywording process.

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
```diff
@@ -0,0 +1,32 @@
...
+	virtual/jdk
```
Arfrever, line 27: `DEPEND=">=virtual/jdk-1.8:*"`
If Java files are installed, then also: `RDEPEND=">=virtual/jre-1.8:*"`

Even in `EAPI="7"` or `EAPI="8"`, java-config still parses `DEPEND` instead of `BDEPEND`:
https://gitweb.gentoo.org/proj/java-config.git/tree/src/java_config_2/Package.py?id=82a5a6343e7b6172af6aaa1b3ea0a61a43afd206#n68

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
thesamesam: We should make the gtest dependency optional. DEPEND is also build-time only but it's about where the dependency is satisfied.

gnome-extra/evolution-data-server/evolution-data-server-3.40.2-r1.ebuild
```diff
@@ -0,0 +1,152 @@
...
+		${gdata_depend}
+	)
+	gnome-online-accounts? (
+		>=net-libs/gnome-online-accounts-3.8:=
+		${gdata_depend} )
+	introspection? ( >=dev-libs/gobject-introspection-0.9.12:= )
+	kerberos? ( virtual/krb5:= )
+	ldap? ( >=net-nds/openldap-2:= )
+	weather? ( >=dev-libs/libgweather-3.10:2= )
+	phonenumber? ( >=dev-libs/libphonenumber-8.12.26 )
```
scardracs, line 57: Personal note: reorder alphabetically

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
Arfrever: Use real initial year when it is known.

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
scardracs: Done

profiles/arch/alpha/package.use.mask
scardracs: Done

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
scardracs: Yeah, java files have to be installed but actually I'm not able to handle them, yet

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
```diff
@@ -17,16 +17,21 @@ LICENSE="Apache-2.0"
...
 "
-
-BDEPEND="
-	dev-cpp/gtest
-	virtual/jdk
+DEPEND="
+	${RDEPEND}
+	>=virtual/jdk-1.8:*
 "
+BDEPEND="dev-cpp/gtest"
```
Arfrever, line 29: If cmake files have unconditional references to gtest, then you can add patch to remove them.
But even without it, valid place for `dev-cpp/gtest` would be `DEPEND`, not `BDEPEND`.

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
scardracs: Java is not required at the end

dev-libs/libphonenumber/libphonenumber-8.12.26.ebuild
scardracs: CMakeFiles are messed up with gtest and It's a bit difficult to patch them. I'm trying to take it out

scardracs: @Arfrever anyway why DEPEND instead of BDEPEND? It's required only during build time

stealthgun:
libphonenumber is now in Guru Dev: https://github.com/gentoo/guru/tree/dev/dev-libs/libphonenumber

Wich is needed for this merge to work.

stealthgun:
The PR is needed because Chatty is dependend on it see: https://github.com/stealthgun/gjdwebserver-overlay/blob/master/net-im/chatty/chatty-0.3.3.ebuild
This is needed for SMS to work on a phone

PR #21467, opened 2021-06-28
media-sound/pulseaudio: Cleanup non-native ABI from deps only used with USE=daemon

i-garrison:
USE=daemon can utilize systemd or elogind, and also tcp-wrappers, drop non-native ABI requirement from these.

Signed-off-by: Igor V. Kovalenko <igor.v.kovalenko@gmail.com>


media-sound/pulseaudio/pulseaudio-14.99.2_pre1.ebuild
```diff
@@ -213,9 +213,9 @@ multilib_src_configure() {
 		#$(meson_use cpu_flags_arm_neon neon-opt)
 		$(meson_native_use_feature tcpd tcpwrap)
 		$(meson_feature dbus)
-		$(meson_feature elogind)
+		$(meson_native_use_feature elogind)
 		$(meson_feature X x11)
-		$(meson_feature systemd)
+		$(meson_native_use_feature systemd)
```
Arfrever, line 218: You should also change `systemd? ( sys-apps/systemd:=[${MULTILIB_USEDEP}] )` to `systemd? ( sys-apps/systemd:= )`.

i-garrison: thanks, fixed this and dropped multilib requirement from `sys-apps/tcp-wrappers` as well

PR #21468, opened 2021-06-29
Add media-libs/liblrdf to gst-plugins-ladspa-1.18.4 dependencies

lanodan:
This also means making media-libs/liblrdf and media-libs/raptor multilib as gstreamer is.


fordfrog:
@lanodan both `media-libs/raptor` and `media-libs/liblrdf` merged, thanks!

PR #21471, opened 2021-06-29
dev-python/{ipython,pylint}, net-libs/grpc, dev-cpp/abseil-cpp: add ~riscv keyword

caoye275:
Also add ~riscv keyword to related packages

### DIFF
--- a/profiles/arch/riscv/package.use.mask
+++ b/profiles/arch/riscv/package.use.mask
@@ -6,7 +6,7 @@
 sci-libs/fftw mpi
 media-sound/alsa-utils libsamplerate
 dev-python/ipython nbconvert notebook
-dev-python/matplotlib cairo excel tk wxwidgets 
+dev-python/matplotlib cairo excel wxwidgets 
 app-text/texlive-core xetex
 
 # Marek Szuba <marecki@gentoo.org> (2021-07-01)
--- a/profiles/arch/riscv/package.use.mask
+++ b/profiles/arch/riscv/package.use.mask
@@ -4,6 +4,7 @@
 # Ye Cao <ye.c@rioslab.org> (2021-07-02)
 # Dependencies not keyworded, not tested
 sci-libs/fftw mpi
+media-sound/alsa-utils libsamplerate
 
 # Marek Szuba <marecki@gentoo.org> (2021-07-01)
 # Dependencies not keyworded here yet:
--- a/profiles/arch/riscv/package.use.mask
+++ b/profiles/arch/riscv/package.use.mask
@@ -161,7 +161,6 @@ net-irc/irssi socks5
 net-libs/gnutls doc test-full
 net-libs/ldns doc python
 net-libs/libpcap rdma
-net-misc/curl curl_ssl_nss nss
 net-misc/ntp openntpd snmp vim-syntax zeroconf
 net-nds/openldap cxx kerberos sasl
 sys-apps/kmod doc
--- a/profiles/arch/riscv/package.use.mask
+++ b/profiles/arch/riscv/package.use.mask
@@ -7,7 +7,7 @@ sci-libs/fftw mpi
 media-sound/alsa-utils libsamplerate
 dev-python/ipython nbconvert notebook
 dev-python/matplotlib cairo excel tk wxwidgets 
-app-text/texlive-core tk xetex
+app-text/texlive-core xetex
 
 # Marek Szuba <marecki@gentoo.org> (2021-07-01)
 # Dependencies not keyworded here yet:
--- a/profiles/arch/riscv/package.use.mask
+++ b/profiles/arch/riscv/package.use.mask
@@ -6,6 +6,7 @@
 sci-libs/fftw mpi
 media-sound/alsa-utils libsamplerate
 dev-python/ipython doc matplotlib nbconvert notebook
+dev-python/matplotlib cairo excel tk wxwidgets 
 
 # Marek Szuba <marecki@gentoo.org> (2021-07-01)
 # Dependencies not keyworded here yet:
--- a/profiles/arch/riscv/package.use.mask
+++ b/profiles/arch/riscv/package.use.mask
@@ -7,6 +7,7 @@ sci-libs/fftw mpi
 media-sound/alsa-utils libsamplerate
 dev-python/ipython doc matplotlib nbconvert notebook
 dev-python/matplotlib cairo excel tk wxwidgets 
+app-text/texlive-core tk xetex
 
 # Marek Szuba <marecki@gentoo.org> (2021-07-01)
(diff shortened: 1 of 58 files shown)


mkszuba:
Strong NAK from me until at least the following two issues have been addressed:
 * masking USE=ssl on dev-python/pycurl forces this flag to off in net-misc/curl itself. This not only is IMHO a bad idea but also causes a tug of war over this USE flag when I try to emerge dev-python/ipython with tests enabled for all the packages you have modified;
 * masking tests is STRONGLY discouraged, and if it really cannot be avoided you should provide in your commit messages details of how exactly you have tested the relevant ebuilds.

caoye275:
> Strong NAK from me until at least the following two issues have been addressed:
> 
> * masking USE=ssl on dev-python/pycurl forces this flag to off in net-misc/curl itself. This not only is IMHO a bad idea but also causes a tug of war over this USE flag when I try to emerge dev-python/ipython with tests enabled for all the packages you have modified;
> * masking tests is STRONGLY discouraged, and if it really cannot be avoided you should provide in your commit messages details of how exactly you have tested the relevant ebuilds.

Thanks for your comments!! I'll try to work on these two issues, finding if I can unmask these to USE flags.

caoye275:
> Strong NAK from me until at least the following two issues have been addressed:
> 
> * masking USE=ssl on dev-python/pycurl forces this flag to off in net-misc/curl itself. This not only is IMHO a bad idea but also causes a tug of war over this USE flag when I try to emerge dev-python/ipython with tests enabled for all the packages you have modified;
> * masking tests is STRONGLY discouraged, and if it really cannot be avoided you should provide in your commit messages details of how exactly you have tested the relevant ebuilds.

There're still issues remained for USE=ssl on dev-python/pycurl. I've dropped the commits about this package this time.
Umask USE=test on dev-python/ipython as well as its dependent packages

caoye275:
> Strong NAK from me until at least the following two issues have been addressed:
> 
> * masking USE=ssl on dev-python/pycurl forces this flag to off in net-misc/curl itself. This not only is IMHO a bad idea but also causes a tug of war over this USE flag when I try to emerge dev-python/ipython with tests enabled for all the packages you have modified;
> * masking tests is STRONGLY discouraged, and if it really cannot be avoided you should provide in your commit messages details of how exactly you have tested the relevant ebuilds.

Fix these 2 issues. Thanks for you suggestions :)

profiles/arch/riscv/package.use.mask
```diff
@@ -1,6 +1,14 @@
 # Copyright 2019-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Ye Cao <ye.c@rioslab.org> (2021-06-30)
+# # Dependencies not keyworded, not tested
+sci-libs/fftw mpi
+media-sound/alsa-utils libsamplerate
+dev-python/ipython doc matplotlib nbconvert notebook
```
mkszuba, line 8: Why do you mask _doc_ and _matplotlib_? Your PR takes care of all the missing keywords on the dependencies pulled in by these two.

profiles/arch/riscv/package.use.mask
```diff
@@ -1,6 +1,14 @@
...
+dev-python/matplotlib cairo excel tk wxwidgets
```
mkszuba, line 9: Don't mask _tk_, the necessary dependencies got keyworded yesterday.

profiles/arch/riscv/package.use.mask
```diff
@@ -1,6 +1,14 @@
 # Copyright 2019-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Ye Cao <ye.c@rioslab.org> (2021-06-30)
+# # Dependencies not keyworded, not tested
+sci-libs/fftw mpi
+media-sound/alsa-utils libsamplerate
+dev-python/ipython doc matplotlib nbconvert notebook
+dev-python/matplotlib cairo excel tk wxwidgets 
+app-text/texlive-core tk xetex
```
mkszuba, line 10: Same here.

profiles/arch/riscv/package.use.mask
```diff
@@ -200,7 +208,7 @@ net-irc/irssi socks5
 net-libs/gnutls doc test-full
 net-libs/ldns doc python
 net-libs/libpcap rdma
-net-misc/curl brotli curl_ssl_mbedtls curl_ssl_nss rtmp ssh nss mbedtls
+net-misc/curl brotli curl_ssl_mbedtls rtmp ssh nss mbedtls
```
mkszuba, line 211: Please remove the _nss_ mask as well.

mkszuba (changes requested):
Much better, thank you! Just a few more mask tweaks and I think this will be ready for merging.

profiles/arch/riscv/package.use.mask
mkszuba: ...which in light of other changes here just committed to the tree means you should simply remove this line.

caoye275: Actually, I'm not sure if it's correct to remove all this USE flags. Simply remove this line might cause some issues in other places?

profiles/arch/riscv/package.use.mask
caoye275: I think I could unmask this in another commit after this PR?  It's quite hard for me to update my local branch to the latest changes. Solving this seperately after merging will be much easier.

profiles/arch/riscv/package.use.mask
caoye275: Thank you! I've fixed the issue just now.

caoye275: Thank you! fixed it.

profiles/arch/riscv/package.use.mask
caoye275: Thank you! fixed it.

profiles/arch/riscv/package.use.mask
caoye275: Thank you! fixed it.

profiles/arch/riscv/package.use.mask
caoye275: Thank you! fixed it.

mkszuba:
Looking good (cosmetics aside, I just had to propagate the keywords on dev-python/astroid and dev-python/pylint to the new versions which made their way to the tree between your PR submission and now), thank you very much for this huge batch of keywords!

caoye275:
> Looking good (cosmetics aside, I just had to propagate the keywords on dev-python/astroid and dev-python/pylint to the new versions which made their way to the tree between your PR submission and now), thank you very much for this huge batch of keywords!

You're welcome. Also very glad to make some contributions :)

PR #21475, opened 2021-06-29
dev-java/xpp2: Homepage, SRC_URI

vaukai:
`PullParser2.1.10.tgz ` is available on https://web.archive.org/web/20130904163102/http://www.extreme.indiana.edu/xgws/xsoap/xpp/download/PullParser2/.


fordfrog:
@vaukai thanks! though i'd rather get rid of it, but we can't atm...

vaukai:
> @vaukai thanks! though i'd rather get rid of it, but we can't atm...

It's very old but still needed for updating dom4j :-(

PR #21489, opened 2021-06-30
sys-cluster/charliecloud: Fix configure with dash shell

olifre:
The `configure` code contained a bashism, which broke `configure` when using `SHELL=/bin/dash` and `CONFIG_SHELL=/bin/dash`. 

Patch submitted to upstream at: https://github.com/hpc/charliecloud/pull/1111

Closes: https://bugs.gentoo.org/799377

### DIFF
--- a/sys-cluster/charliecloud/charliecloud-0.23.ebuild
+++ b/sys-cluster/charliecloud/charliecloud-0.23.ebuild
@@ -48,6 +48,7 @@ DEPEND="
 
 PATCHES=(
 	"${FILESDIR}"/${PN}-0.22-sphinx-4.patch
+	"${FILESDIR}"/${PN}-0.24-dash.patch
 )
 
 src_prepare() {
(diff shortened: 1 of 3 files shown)


ionenwks (approved):
Thanks for sending it upstream too.

I'll scrub the patch of metadata (the diff --git/index stuff) and add the upstream's link to it on merge, but otherwise no real changes.

olifre:
> I'll scrub the patch of metadata (the diff

@ionenwks Thanks :-). I was not aware the metadata should be scrubbed, but of course that is very sensible to save space in the tree. Noted down for future PRs.

ionenwks:
> > I'll scrub the patch of metadata (the diff
> 
> @ionenwks Thanks :-). I was not aware the metadata should be scrubbed, but of course that is very sensible to save space in the tree. Noted down for future PRs.

See
https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html

It's not a hard requirement, but appreciated to be noted for future PRs :)

PR #21490, opened 2021-06-30
Bitcoin 0.21.1

luke-jr:
Other than the typical bump, note this release changes the Bitcoin consensus rules. It is important that users explicitly opt-in to consensus rule changes. To accomplish this, a new `bitcoin_protocol_taproot` USE flag is added, unset by default, that must be set by users to install (`pkg_pretend` will `die` if it is unset). Not sure if this is the best approach or something else is desirable.


luke-jr:

candrews:
taproot locked - so why do we need to the user to do anything? My understanding is that the ship has sailed, taproot is happening, so to continue being a bitcoin node as of November taproot must be supported.

luke-jr:
Bitcoin is not a centralised network. The signalling only indicates miners are prepared early, but the actual protocol change depends on the community's acceptance.

candrews:
But the community has accepted - taproot is locked: https://taproot.watch/

luke-jr:
Miners are not the community.

Signalling shouldn't even begin until the community has decided, but that observation is not a reason to change rules on people without their own active consent. (This is one reason why we have intentionally never included any auto-upgrade in the software itself.)

Perhaps an argument can be made to forego the consent, but "miners say so" certainly is not that.

candrews:
How about an `einfo` message instead of a use flag?

Also, can you please fix the issues reported by CI?

dev-util/bitcoin-tx/bitcoin-tx-0.21.1.ebuild
```diff
@@ -0,0 +1,104 @@
...
+IUSE="knots"
+
+DEPEND="
+	>=dev-libs/boost-1.58.0:=[threads(+)]
+	>dev-libs/libsecp256k1-0.1_pre20200911:=[recovery,schnorr]
+	>=dev-libs/univalue-1.0.4:=
+"
+RDEPEND="${DEPEND}"
+BDEPEND="
+	>=sys-devel/autoconf-2.69
```
thesamesam, line 31: I don't think this BDEPEND is needed at all, but if it is, there are eclass variables for this from autotools.

net-libs/libbitcoinconsensus/libbitcoinconsensus-0.21.1.ebuild
```diff
@@ -0,0 +1,111 @@
...
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~mips ~ppc ~ppc64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="+asm bitcoin_protocol_taproot knots"
+
+DEPEND="
+	>dev-libs/libsecp256k1-0.1_pre20200911:=[recovery,schnorr]
+"
+RDEPEND="${DEPEND}"
+BDEPEND="
```
thesamesam, line 28: Same BDEPEND comments.

net-libs/libbitcoinconsensus/libbitcoinconsensus-0.21.1.ebuild
```diff
@@ -0,0 +1,111 @@
...
+	elog
+	ewarn "CAUTION: BITCOIN PROTOCOL CHANGE INCLUDED"
+	ewarn "This release adds enforcement of the Taproot protocol change to the Bitcoin"
+	ewarn "rules, beginning in November. Protocol changes require user consent to be"
+	ewarn "effective, and if enforced inconsistently within the community may compromise"
+	ewarn "your security or others! If you do not know what you are doing, learn more"
+	ewarn "before November. (You must make a decision either way - simply not upgrading"
+	ewarn "is insecure in all scenarios.)"
+	ewarn "To learn more, see https://bitcointaproot.cc"
+	if ! use bitcoin_protocol_taproot; then
```
thesamesam, line 56: I don't think this is the right approach (but I _do_ agree with ensuring users are informed):

- You mentioned upstream doesn't do auto-upgrades, but does it warn at all on installation/upgrade 'manually'?

- (Additionally, I don't really want to go beyond what upstream did or didn't do. Gentoo isn't really the place to lobby or impose what you wish they'd done upstream. Not saying that's what happening here, I just want to be careful of the line, because I know it's tempting sometimes.)

- I'd much prefer a news item for people who have this set of packages installed or a pkg_postinst message. Note that in general, we assume users read elogs because they often
contain important information. (People won't be running the new binary until they explicitly restart e.g. bitcoind anyway).

- In terms of reasons against pkg_pretend dying, the main issue is that it hurts automation (the package manager can only detect failure after dependency resolution and so on), and we already have REQUIRED_USE for things like this.

A REQUIRED_USE setting which isn't satisfied by default would trigger a QA warning which we _could_ possibly waive, but I'm still more inclined to go for the news item or pkg_postinst here. Those are the two official methods for letting users know about upcoming changes.

What do you reckon?

(cc @candrews)

thesamesam (changes requested):

net-libs/libbitcoinconsensus/libbitcoinconsensus-0.21.1.ebuild
Flowdalic: FWIW, it appears to me that this is nothing that users should be notified *via* gentoo mechanisms. I know that this is probably a delicate topic for some (as it's bitcoin). But ultimately a use flag that does nothing besides preventing the package from being emerged if not set, is not something we should introduce nor encourage in ::gentoo.

thesamesam: Upon reflection, I agree with @Flowdalic. Users should be subscribed to upstream's news notifications (mailing lists, whatever). We should really only be using Gentoo news items (and arguably pkg_postinst) for critical breaking changes.

This kind of thing is quite natural for Bitcoin. I think we should therefore be doing a pkg_postinst message _at most_.

luke-jr: This *is* a critical breaking change, one that can potentially cost the user their life savings.

I brought the subject up at today's Bitcoin development meeting, and there didn't seem to be much agreement for foregoing explicitly user interaction.

thesamesam: Then shouldn't **upstream** be doing something about it? Make bitcoind or whatever scream if a config option hasn't been set? Gentoo, being downstream, isn't the place to lobby for things you wish were done upstream?

There still is user interaction. People should not be doing unattended upgrades because people have to read pkg_postinst messages anyway (see e.g. PAM upgrades, OpenSSH, ...). If they are, they're going to have other problems.

Let's go a bit further too. I don't want to be the only people doing this. What's happening with Debian/Ubuntu .debs? Fedora/* RPMs?

luke-jr: >Then shouldn't upstream be doing something about it? 

We don't have any automatic updating upstream, intentionally.

>Make bitcoind or whatever scream if a config option hasn't been set?

That was also discussed, but it won't be here for Taproot.

>I also don't see anything on https://bitcoin.org/en/download mentioning this.

It probably should. https://bitcoincore.org/ is pretty explicit about it.

> Can you walk me through what happens when I run Bitcoin Core on say, Windows, and there's an update available?

Nothing. You have to manually seek it out and choose to upgrade.

>People should not be doing unattended upgrades because people have to read pkg_postinst messages anyway (see e.g. PAM upgrades, OpenSSH, ...).

Telling the user after the fact is kinda too late?

>Let's go a bit further too, I don't want to be the only people doing this.. what's happening with Debian/Ubuntu .debs? Fedora/* RPMs?

We asked them not to package at all. I maintain the PPA, and plan to do something similar (debconf has a way to insist on explicit user interaction)

>Separately: I'm wondering why Bitcoin isn't available for download in the UK.

It is. What gave you the impression it isn't?

Flowdalic: > I brought the subject up at today's Bitcoin development meeting, and there didn't seem to be much agreement for foregoing explicitly user interaction.

I read said meeting and agree that this is caused by using bitcoin via a package manager that performs auto updates *and* by bitcoin core enabling new consensus rules on new updates by default. I really appreciate your effort here.

That said, I think we best we can do is educate Gentoo users about this situation, that every net-p2p/bitcoind update potentially comes with new defaults that they need to research about. But `if ! use new-consensus-rule; then die; fi` seems excessive, especially since it may aborts an upgrade process (if `--keep-going` is not enabled).

I am currently leaning towards a short warning message on every update, but in case of taproot, a news item seems also sensible. Other means could be discussed as well, like a guard file, that if not existent prevents bitcoind from starting.

net-libs/libbitcoinconsensus/libbitcoinconsensus-0.21.1.ebuild
```diff
@@ -0,0 +1,111 @@
...
+		elog "https://bitcoincore.org/en/2021/05/01/release-${PV}/"
+	fi
+	elog
+	ewarn "CAUTION: BITCOIN PROTOCOL CHANGE INCLUDED"
+	ewarn "This release adds enforcement of the Taproot protocol change to the Bitcoin"
+	ewarn "rules, beginning in November. Protocol changes require user consent to be"
+	ewarn "effective, and if enforced inconsistently within the community may compromise"
+	ewarn "your security or others! If you do not know what you are doing, learn more"
+	ewarn "before November. (You must make a decision either way - simply not upgrading"
+	ewarn "is insecure in all scenarios.)"
```
Flowdalic, line 54: That kinda implies that it is possible to upgrade without using taproot. But if so, why do we have to `die` if taproot is not enabled?

net-libs/libbitcoinconsensus/libbitcoinconsensus-0.21.1.ebuild
luke-jr: `REQUIRED_USE` seems to solve the problem of aborting other updates.

Another possibility is an additional dummy LICENSE which isn't part of the default-accepted lists?

net-libs/libbitcoinconsensus/libbitcoinconsensus-0.21.1.ebuild
luke-jr: If people actually reject Taproot, they can indeed reject it. But doing so would involve writing code, getting others to adopt it, etc - it's not something we can just support as-is today until someone puts the work into it.

candrews: > I am currently leaning towards a short warning message on every update, but in case of taproot, a news item seems also sensible. Other means could be discussed as well, like a guard file, that if not existent prevents bitcoind from starting.

I'm on board with a log message and/or news item, but I won't merge with the blocking use flag approach.

I'd really like to get this Bitcoin version bump in soon...

net-libs/libbitcoinconsensus/libbitcoinconsensus-0.21.1.ebuild
thesamesam: >>People should not be doing unattended upgrades because people have to read pkg_postinst messages anyway (see e.g. PAM upgrades, OpenSSH, ...).

>Telling the user after the fact is kinda too late?

But it's not after the fact. Users shouldn't be doing unattended upgrades. They will see the log and downgrade if necessary?

>>Separately: I'm wondering why Bitcoin isn't available for download in the UK.

>It is. What gave you the impression it isn't?

https://bitcoin.org/en/download says "(This software is presently not available for download in the UK, and download links will not work if you are located within the UK.)" and the links (as it says) don't work.

----
I'm okay with a one-off news item as @Flowdalic suggests to indicate to users they should _always_ check on major upgrades and we'll do an elog too. In future, just an elog. How about that?

net-libs/libbitcoinconsensus/libbitcoinconsensus-0.21.1.ebuild
thesamesam: In future, I'd really prefer it if a solution was figured out upstream and we didn't need to have these debates. But yeah, agreed. See https://github.com/gentoo/gentoo/pull/21490#discussion_r664209806.

dev-util/bitcoin-tx/bitcoin-tx-0.21.1.ebuild
luke-jr: ???

net-libs/libbitcoinconsensus/libbitcoinconsensus-0.21.1.ebuild
luke-jr: >But it's not after the fact. Users shouldn't be doing unattended upgrades. They will see the log and downgrade if necessary?

It is after the fact, by definition. PAM/OpenSSH notices are things to do after the update, not reasons one might not consent to the upgrade.

>https://bitcoin.org/en/download says "(This software is presently not available for download in the UK, and download links will not work if you are located within the UK.)" and the links (as it says) don't work.

This is bitcoin.org's owner playing political games. There isn't any actual problem.

>I'm okay with a one-off news item as @Flowdalic suggests to indicate to users they should always check on major upgrades and we'll do an elog too. In future, just an elog. How about that?

That doesn't ensure the user has consented. This also isn't generally recognisable as a major upgrade (0.21.0 -> 0.21.1).

dev-util/bitcoin-tx/bitcoin-tx-0.21.1.ebuild
ionenwks: > ???

autoconf BDEPEND for autotools.eclass is already handled by the eclass (by default), so you don't really need this.

From autotools.eclass:
```
# @ECLASS-VARIABLE: AUTOTOOLS_AUTO_DEPEND
# @PRE_INHERIT
# @DESCRIPTION:
# Set to 'no' to disable automatically adding to DEPEND.  This lets
# ebuilds form conditional depends by using ${AUTOTOOLS_DEPEND} in
# their own DEPEND string.
: ${AUTOTOOLS_AUTO_DEPEND:=yes}
if [[ ${AUTOTOOLS_AUTO_DEPEND} != "no" ]] ; then
    case ${EAPI} in
        5|6) DEPEND=${AUTOTOOLS_DEPEND} ;;
        7) BDEPEND=${AUTOTOOLS_DEPEND} ;;
```

candrews:
I'm going to merge this PR with a couple changes:
* The BDEPEND removal as pointed out in https://github.com/gentoo/gentoo/pull/21490#discussion_r661999752
* The taproot USE flag removal. I'll leave in the warnings, but changed them so they only appear on package upgrade. I think that approach notifies the user (notably _before_ the package is installed because the warning is in the `pkg_pretend` phase) and also aligns with Gentoo packaging practices.

luke-jr:
Please revert. That is unethical and I did NOT sign-off on it.

gyakovlev:
I had to revert it due to @luke-jr reaching out and claiming statement above.

Please reach consensus of some kind.
I know nothing about bitcoin, it's a decision based solely on signoff complaint.
I do understand that's it's a sensitive topic, so reach out to wider audience if needed to reach consensus.
Peace.

luke-jr:
Thanks.

I'm not sure what the rush is - this upgrade is not important until November. We do have 4 months to find a clean and ethical solution (though hopefully it won't take nearly that long).

What concerns are there with using REQUIRED_USE or a dummy LICENSE?

gyakovlev:
we do have `PROPERTIES+=" interactive"` mechanism.
it also breaks unattended upgrades, but at least a user can decline to accept/decline the property via configuration.

```
--accept-properties=ACCEPT_PROPERTIES
    This option temporarily overrides the ACCEPT_PROPERTIES variable. The ACCEPT_PROPERTIES variable is incremental, which means that the specified setting is appended to the existing value from your configuration. The special -* token can be used to discard the existing configuration value and start fresh. See the MASKED PACKAGES section and make.conf(5) for more information about ACCEPT_PROPERTIES. A typical usage example for this option would be to use --accept-properties=-interactive to temporarily mask interactive packages. With default configuration, this would result in an effective ACCEPT_PROPERTIES value of "* -interactive". 
```

gyakovlev:
but idk if ^ is not deprecated. ask QA in doubt.

luke-jr:
> we do have PROPERTIES+=" interactive" mechanism.
>it also breaks unattended upgrades, but at least a user can decline to accept/decline the property via configuration.

Interesting. From what I'm reading, the purpose of the PROPERTIES value is to mask the package during unattended upgrades, so it sounds okay. We could maybe also include a suggestion to touch a file in /etc/bitcoin/ if one wants to accept non-interactively?

gyakovlev:
PROPERTIES=interactive  signals  portage that ebuild may ask for user input, afaik.
but it's up to you to implement reading input from user. check `cdrom.eclass`

candrews:
Emailed the gentoo-dev list requested further input: https://archives.gentoo.org/gentoo-dev/message/b4e73ddcb2409e71c66ad314d51de0f7

luke-jr:
A few corrections and clarifications to that email (inline):

On Saturday 10 July 2021 15:42:28 Craig Andrews wrote:
> * This is a minor version bump (assuming semver, this is the "patch"
> level version change in bitcoin), indicating that upstream does not
> consider this to be a major/breaking change.

It does not indicate that. Everyone else upstream that has commented on the matter, agrees user opt-in must be maintained at this time (probably until activation in November).

The reason it is a patch-level version change, is that
1) We do not wish to tie it to other improvements, thereby putting inappropriate pressure on users to enforce Taproot simply because they want some other feature; and
2) It is a change to Bitcoin, not Bitcoin Core. If Bitcoin changes, it is a security flaw in old versions of Bitcoin node software. Therefore, it is considered a security bugfix. But note this is assuming Bitcoin changes, which is true only by the consent of users. (Bitcoin itself does not have a well-defined version number, but the current protocol semantic versioning documentation has Taproot as a change from the current 2.4.2 version to 2.5.0.)

> * Upstream does not have a mechanism for notifying users or requiring
> them to opt-in to this change

We intentionally do not support automatic upgrades to avoid the issue, and ensure opt-in. Users explicitly opt-in when they go to our website, see the release (explicitly mentioning the addition of Taproot), and choose to actively download and install it. What Gentoo needs here, is an equivalent.

> * Upstream does not have a mechanism to opt out of this change. The
> users only option is to develop their own fork of the bitcoin software
> or never upgrade the package if they want to avoid taproot.

Assuming other users do upgrade, non-upgrading is not a viable way to reject Taproot. Only creating a fork will do the job.

If users in general, however, reject Taproot, Bitcoin Core (and others) will have no choice but to back out the changes for Taproot, in which case there would be a 0.21.2+ without it.

> * Taproot was locked by miners, so the network will be upgrading [2]

Miners do not determine Bitcoin's consensus rules at all.
Their signalling and lock-in process is not relevant to this.

> Therefore, I have a few questions for the fellow Gentoo developers:
> 1) Should we require users to explicitly opt-in to this upgrade beyond
> the usual?

Anything other than this, at this time, would be unethical.
If anyone remembers the issue years ago where a mere policy change was not well-documented enough, forgoing this opt-in would be MUCH worse.

This isn't just a policy. This is consensus rules the user is imposing on every other user. This has direct real-world financial risk. There may (or may not) be a legal argument for fiduciary liability if such a change is made without user consent. Doing so also compromises the security model of the Bitcoin network, which is based on the assumption that each user makes the decision for himself.

williamh:
@luke-jr  I sent a couple of emails on this, but after reading here, i definitely think the best answer for this is to not have anything unusual in the ebuild, but to commit the bump under package.mask along with a mask message explaining the issue. If you want a wider announcement, you can publish a newsitem after the package is committed behind a mask.
The way people would opt in in that case is add the package to `/etc/portage/package.unmask` and possibly `/etc/portage/package.accept_keywords`.

luke-jr:
@williamh That sounds acceptable, though maybe a bit stronger than would be ideal.

Reading your email, I want to note that:

1. Old versions don't just stop functioning - they remain functional but without security (security is the main reason to use these packages to begin with).
2. It isn't clear when we will know for sure the network has adopted the softfork. IMO activation is a good assumption for Taproot, but this may not be a usable precedent for less apparently-unanimous changes.

I don't think these change the conclusions you reached, however.

Shall I go ahead and make the changes proposed for the package.mask solution?

candrews:
> @williamh That sounds acceptable, though maybe a bit stronger than would be ideal.
> 
> Reading your email, I want to note that:
> 
>     1. Old versions don't just stop functioning - they remain functional but without security (security is the main reason to use these packages to begin with).
> 
>     2. It isn't clear when we will know for sure the network has adopted the softfork. IMO activation is a good assumption for Taproot, but this may not be a usable precedent for less apparently-unanimous changes.
> 
> 
> I don't think these change the conclusions you reached, however.
> 
> Shall I go ahead and make the changes proposed for the package.mask solution?

It seems we're both working on this at the same time :)

Here's a PR with has package.mask: https://github.com/gentoo/gentoo/pull/21587 Hopefully everyone finds it agreeable. Thanks to all for your assistance with this issue.

luke-jr:
Closing this in favour of #21587

PR #21492, opened 2021-07-01
dev-libs/libtpms: Bump to 0.8.4

salahcoronya:


dev-libs/libtpms/libtpms-0.8.4.ebuild
```diff
@@ -0,0 +1,46 @@
...
+)
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	econf \
+		--disable-static \
```
ionenwks: autotools.eclass received EAPI-8 support few days ago, meaning this ebuild can be bumped to EAPI-8 now and it would allow to remove this --disable-static

https://mgorny.pl/articles/the-ultimate-guide-to-eapi-8.html#new-econf-passed-options

dev-libs/libtpms/libtpms-0.8.4.ebuild
```diff
@@ -0,0 +1,46 @@
...
+		--with-openssl \
+		--with-tpm2
+}
+
+src_install() {
+	default
+	find "${D}" -name '*.la' -delete || die
```
ionenwks: trivial given both works but normally use `${ED}` here like the [policy-guide example](https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0303).

good to check if it's still needed now and then too, perhaps .la files aren't installed anymore (haven't checked myself)

dev-libs/libtpms/libtpms-0.8.4.ebuild
```diff
@@ -0,0 +1,46 @@
...
+}
+
+pkg_postinst() {
+	elog "Versions of libtpms prior to 0.8.0 generate weaker than expected TPM 2.0 RSA"
+	elog "keys due to a flawed key creation algorithm. Because fixing this would render"
+	elog "existing sealed data inaccessible, to use the corrected algorithm, the old"
+	elog "TPM state file must be deleted and a new TPM state file created. Data still"
+	elog "sealed using the old state file will be permanently inaccessible. For the"
+	elog "details see https://github.com/stefanberger/libtpms/issues/183"
```
ionenwks: Is it still necessary to display this elog every single install? i.e. even if <0.8.0 is not on the system anymore?

Not that I'm familiar with the issue, so if there's user action needed it may be a different story.

dev-libs/libtpms/Manifest
```diff
@@ -1,2 +1 @@
-DIST libtpms-0.8.3.tar.gz 1254474 BLAKE2B c71f52d6475d4328e93a35f88557de3400308984ec086f7e6b3c32ebc266e078b3a9864396546ba11fc310c1572efa1521bdac89075c5eba18cb35eef883e706 SHA512 07cf78c2a5b03f3c9179b041db57d5ec049dc0085c09e4fd5c9b54a247cf81a3b9982f0fb1cd571cd6b21b3d9824b061f8f38508fc25b6a23a6a6b06dea7115b
```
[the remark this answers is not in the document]
salahcoronya: 0.8.4 is a fairly lightly patched 0.8.3:

>  stefanberger/libtpms@a594c46 (tag: v0.8.4) rpm/debian: Add 0.8.4 entry in changelog
>  stefanberger/libtpms@dc176a5 CHANGES: Updated CHANGES document for 0.8.4
>  stefanberger/libtpms@5cc98a6 tpm2: Restore original value if unmarshalled value was illegal
>  stefanberger/libtpms@3ef9b26 tpm2: Add maxSize parameter to TPM2B_Marshal for sanity checks
>  stefanberger/libtpms@f16250b tpm2: Reset TPM2B buffer sizes after test fails for valid buffer size
>  stefanberger/libtpms@2d412c9 build-sys: Run autoupdate on configure.ac for autconf 2.71
>  stefanberger/libtpms@d67dffc build-sys: Build libtpms v0.8.4

I have no opposition to keeping 0.8.3,but its basically just a few autoconf fixes and a buffer fix, its probably not worth it this time.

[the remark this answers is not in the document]
ionenwks: It's okay then, just wanted to be sure it been considered.

dev-libs/libtpms/libtpms-0.8.4.ebuild
salahcoronya: Yes, it still install .la, but I have updated it to look like the example

dev-libs/libtpms/libtpms-0.8.4.ebuild
salahcoronya: I did not know about the REPLACING_VERSIONS variable, so I have implemented that

dev-libs/libtpms/libtpms-0.8.4.ebuild
```diff
@@ -0,0 +1,53 @@
...
+src_install() {
+	default
+	find "${ED}" -name '*.la' -delete || die
+}
+
+pkg_postinst() {
+	for v in ${REPLACING_VERSIONS}
+	do
+		if ver_test "${v}" -lt 0.8.0
+		then
```
ionenwks, line 43: `v` is lacking `local` and, while it's just style, in ebuilds we usually don't put `do` and `then` on their own line.

But nevermind `v`. While this method works, it's not necessary if the package isn't slotted, i.e. REPLACING_VERSIONS will only ever be either empty, or 1 single version with no spaces.

So you can:
`if [[ ${REPLACING_VERSIONS} ]] && ver_test ${REPLACING_VERSIONS} -lt 0.8.0; then`

dev-libs/libtpms/libtpms-0.8.4.ebuild
ionenwks: I see new push is still EAPI-7 using --disable-static, I assume you want to keep it that way for now? (that's fine too if you prefer)

salahcoronya: For now, sticking with EAPI 7. EAPI 8 is still a little too new. I don;t know how well automated tests works with it yet, and I know packages.gentoo.org is having problems with it: https://bugs.gentoo.org/802039 .

ionenwks (approved):
I've noticed there's a keywording request filled against libtpms-0.8.3, for the sake of being able to compare (there's apparently issues with swtpm on ppc64 for one), I think older should be kept just-in-case for now.

So I'll drop the remove old on merge, hope you don't mind.

Other than that, bump looks good and tests pass.

PR #21493, opened 2021-07-01
sys-libs/libxcrypt: switch to pre-generated autotools tarballs

thesamesam:
There are actually *two* circular dependencies involving Perl:

1) Use self-generated (for now) `make dist` tarballs to avoid a circular
   dependency with libxcrypt->automake->perl->libxcrypt.

   (Thanks juippis and floppym! We noticed this because juippis hit an
   interesting edge case when using binpkgs.)

2) We need to pre-generate the results of a Perl tool called
   during `./configure` in order to avoid unconditionally needing
   Perl.

   We can do this because the input is constant for all of the
   Gentoo build variants - for now.

   (Thanks mattst88! We noticed this while digging into suggestions
   for upstream.)

(Both best explained within the comments of the ebuild.)

Reported-by: Joonas Niilola <juippis@gentoo.org>
Reported-by: Mike Gilbert <floppym@gentoo.org>
Reported-by: Matt Turner <mattst88@gentoo.org>


sys-libs/libxcrypt/libxcrypt-4.4.23.ebuild
```diff
@@ -2,17 +2,27 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
-PYTHON_COMPAT=( python3_{7..10} )
-inherit autotools multibuild python-any-r1 multilib-minimal
+
+PYTHON_COMPAT=( python3_{7..9} )
```
Arfrever, line 6: Mistakenly removed `python3_10`.

sys-libs/libxcrypt/libxcrypt-4.4.23.ebuild
```diff
@@ -22,13 +32,12 @@ RDEPEND="${DEPEND}"
...
+	# See below comment/notes in src_prepare
+	BDEPEND+=" dev-lang/perl"
+fi
 
-REQUIRED_USE="split-usr? ( system )"
-
-PATCHES=(
-	"${FILESDIR}/libxcrypt-4.4.19-multibuild.patch"
-)
+RESTRICT="!test? ( test )"
```
Arfrever, line 40: This variable often is set directly after `IUSE` and `REQUIRED_USE`.

stefson:
I had problems with bootstrapping a cross compiler toolchain the other day, because there was USE="-crypt" in the hosts make.conf (for reasons I forgot), and so the cross-glibc was compiled without crypt support. That made stage2-gcc fail when it tried to build its sanitizers. This was before I read the news today, so I'm writing this to just to leave you some feedback.

sys-libs/libxcrypt/files/libxcrypt-4.4.20-read-hashes_selected-from-injected-file.patch
```diff
@@ -0,0 +1,29 @@
+From 6cc88caf1aac5418832af25fe02823a256fe5aa4 Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: I just realised: we should hack out the PROG_PERL stuff too.

thesamesam:
@stefson Thanks for bringing this to my attention! Let's discuss on the [bug](https://bugs.gentoo.org/799707) (yours?).

PR #21495, opened 2021-07-01
media-gfx/hydrus: Add 445, drop 441-r1

vaartis:
Some patches have been accepted upstream, so 445 doesn't need them anymore. 
Older version 441-r1 has been dropped.


media-gfx/hydrus/hydrus-445.ebuild
```diff
@@ -10,13 +10,21 @@ inherit python-single-r1 desktop optfeature
 
 DESCRIPTION="A booru-like media organizer for the desktop"
 HOMEPAGE="https://hydrusnetwork.github.io/hydrus/ https://github.com/hydrusnetwork/hydrus"
-SRC_URI="https://github.com/hydrusnetwork/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+if [ "${PV}" == "9999" ]; then
```
thesamesam, line 14: Bash [tests](https://devmanual.gentoo.org/tools-reference/bash/#single-versus-double-brackets-in-bash)!

vaartis: Fixed for all versions, plus updated 9999.

PR #21502, opened 2021-07-01
app-arch/makeself: Bump to 2.4.5

00-matt:


ionenwks (approved):
re: dropping 2.4.4, will merge as-is but do keep in mind to decide on a potential stable candidates and not always drop the previous

ionenwks:
I've additionally dropped:
- files/makeself-2.2.0-help-header.patch
- files/makeself-unpack

As they were only used by the dropped 2.2.0, hope that was alright.

00-matt:
That's great. Thank you!

PR #21503, opened 2021-07-01
Rework the eselect-fontconfig dependency

dm0-:
Currently `app-eselect/eselect-fontconfig` is installed via `PDEPEND`, which always installs it in `$ROOT`.  It needs to installed in `$BROOT` to be natively executable on the build system, e.g. to configure a staging root when cross-compiling.

It seems that nothing calls `eselect fontconfig` anywhere, so maybe the dependency could just be dropped, and users could install it as needed since it's only intended to be manually called.

If it needs to continue being installed automatically, it probably belongs in `IDEPEND` so it is available in `$BROOT` at install time so `$ROOT` can be configured on the build system.  Since `autotools.eclass` is blocking most of the distro from using EAPI 8, it would have to use the EAPI 7 convention of `RDEPEND`+`BDEPEND` to get close enough.  That's what this PR does.

See also [bug #799758](https://bugs.gentoo.org/799758).


thesamesam:
Thanks. Please see https://bugs.gentoo.org/799758#c1 and I can merge that too then.

PR #21504, opened 2021-07-01
x11-terms/kitty and kitty-terminfo: bump to 0.21.2

end222:
x11-terms/kitty and kitty-terminfo: bump to 0.21.2


x11-terms/kitty/kitty-9999.ebuild
```diff
@@ -57,8 +57,8 @@ BDEPEND="virtual/pkgconfig"
 	$(python_gen_cond_dep '>=dev-python/sphinx-1.7[${PYTHON_MULTI_USEDEP}]')"
 
 PATCHES=(
-	"${FILESDIR}"/${P}-flags.patch
-	"${FILESDIR}"/${P}-remove-terminfo.patch
+	"${FILESDIR}"/${PN}-0.21.2-flags.patch
+	"${FILESDIR}"/${PN}-0.21.2-remove-terminfo.patch
 	"${FILESDIR}"/${PN}-0.14.4-svg-icon.patch
```
juippis, line 62: This doesn't apply anymore on -9999. Can we just remove these -9999 versions? Rest looks good to me.

juippis (approved):

x11-terms/kitty/kitty-9999.ebuild
end222: Hello! Do you mean to remove the whole -9999 ebuild?

ionenwks: With proxy-maint it can be especially a pain to maintain a -9999 version, notably if they need patches likely to break (i.e. it's broken right now).

It's okay to have if you want them but if no strong feeling about it may be better to remove (whole) -9999.

juippis: > Hello! Do you mean to remove the whole -9999 ebuild?

Yes. It seems to be broken everytime I try to build it. And you have to do some patch messing around with it all the time too.
I think it's not worth having around.

end222: Ok, thank you very much.
I will remove it, it's really not worth it because it is constantly changing and breaking

PR #21507, opened 2021-07-02
net-misc/r8168: added use flags

Karlson2k:
* Added USE flags to control drivers defaults
Upstream has mechanism to control module defaults at compile-time. Added USE flag with same defaults as upstream.
* Added USE flag to enable firmware loading support
New compile-time parameter required to enable firmware loading support. Tested on my minicomputer with kernel dyn-debug and firmware loaded fine if USE flag is enabled. May help on old hardware or cheap hardware with outdated firmware.
* Demoted kernel config check from error to warning
If kernel parameter is enabled and module is compiled, nothing is failed. If both modules are enabled, kernel will load built-in module automatically. So instead of preventing of compilation of this module, simply warn user about required kernel configuration.

### DIFF
--- a/net-misc/r8168/metadata.xml
+++ b/net-misc/r8168/metadata.xml
@@ -8,4 +8,13 @@
 	<longdescription>Official Realtek r8168 linux driver. The following cards are currently supported:
 RTL8111B RTL8168B RTL8111 RTL8168 RTL8111C RTL8111CP RTL8111D(L) RTL8168C
 RTL8111DP RTL8111E RTL8168E RTL8111F RTL8411</longdescription>
+	<use>
+		<flag name="aspm">Enable ASPM by default</flag>
+		<flag name="aspm-dynamic">Enable dynimic ASPM by default</flag>
+		<flag name="enable-eee">Enable Energy-Efficient Ethernet by default</flag>
+		<flag name="hw-acceleration">Enable use of hardware acceleration by default. HW acceleration may cause problems on Realtek NICs!</flag>
+		<flag name="s5-keep-mac">Keep set MAC address for S5 by default</flag>
+		<flag name="s5-wol">Enable wake from S5 by WOL by default</flag>
+		<flag name="use-firmware">Enable support for automatic firmware loading</flag>
+	</use>
 </pkgmetadata>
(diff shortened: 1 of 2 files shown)


Karlson2k:
I see that ebuild was updated to the same upstream version in 2 days after this PR.
I've updated this PR leaving only new functionality in it. 
On my PC I enabled `use-firmware` flag and use this driver on the daily basis.
@pacho2 @thesamesam Could you check this PR?

Karlson2k:
The `use-firmware` is especially important. Other flags just set drivers default which can be overridden by module parameters.
The `use-firmware` flag is the only way to enable support of firmware loading.

pacho2:
Hello,

Why do you prefer to get that defaults changed at build time instead of the user passing the options when loading the module like for all the other kernel modules? I mean, for example, by a conf file in /etc/modprobe.d

Karlson2k:
Hello Pancho,

The main reason for this patch is enabling of firmware loading support. There is no way to enable it after compiling.

Other parameters are added because:
* they are supported directly by the upstream;
* they give users more flexibility.

One may prefer to set in module parameter in modprobe.d, another may prefer to use build-time defaults, someone may use both ways.

For example, I have several similar computers used in similar way. I installed and configured everything on the first one and then copied main configuration files to others. For sure I copied `etc/portage/packge.use/*`. With ability to use build-time driver defaults, I can skip copying some configuration files.

pacho2:
For the firmware loading support maybe we could enable it unconditionally

For the other options... While I was reviewing the upstream makefiles and also Arch package:
https://github.com/archlinux/svntogit-community/blob/packages/r8168/trunk/PKGBUILD

I wondered if maybe we could enable all the options unconditionally as they do.

Regarding allowing its build time configuration... I think no other kernel module is doing that, they all seem to rely on the administrator changing the options in modprobe.d files... but maybe someone from QA or Kernel teams could suggest something on that

Karlson2k:
We should not enable all options unconditionally. I don't see any reason to redefine option defined in Makefile by upstream with another value if this value is static.

This PR reflect upstream defaults to USE flags. Firmware support is disabled by default (maybe is not tested enough).
You need admin rights to change either modporobe.d or ebuild. Hiding these options doesn't improve anything (security, for example) as use may modify source code or .ebuild file.

If use change drive defaults at compile time, parameters still can be overridden by modprobe.d.

pacho2:
I still think we should simply build with defaults and rely on people setting the options they want in modprobe.d files as all the other kernel modules do

Karlson2k:
I still think that people should have freedom to choose and additional USE flags do not introduce any complexity for maintainability  and for deployment.
However, I've created alternative PR #21739 with firmware flag only.

Karlson2k:
Superseded by https://github.com/gentoo/gentoo/pull/39268

PR #21508, opened 2021-07-02
sys-apps/bat: Version updated to 0.18.1

jpds:
* Added BUILD_DIR hack to select a single output directory for zsh
  completions.


toshokan (approved):

scardracs:
Rust is not available for armv4, armv4t and armv5te profiles so you may want to modify the package.mask for these profiles or it will fails everytime. See, for eg, https://github.com/gentoo/gentoo/blob/master/profiles/arch/arm/armv4/package.mask

scardracs:
Please add a comment too (like the one I've made on it) explaining why the package is masked

juippis (approved):
Gonna add few missing `|| die`s and fix the DESCRIPTION back to what it was. Works great, thanks!

juippis (commented):
Note though: you+me can't add an arm keyword for this package. This needs to be requested through bugzilla. I'll add the package.masks but please head to bugzilla to request arm keyword. Let me know if you need help how to do it.

PR #21511, opened 2021-07-03
*/*: Update from deprecated l10n.eclass to plocale.eclass

ulm:
Let's do this in one go, which might be less time consuming than leaving it to package maintainers.

This should be a simple renaming with no change in functionality.


trofi:
LGTM for dolphin

Pesa:
ACK for app-misc/brewtarget and all @gentoo/qt packages.

SoapGentoo (approved):

PR #21512, opened 2021-07-03
Use installed gstreamer plugin libs to build split plugins packages

i-garrison:


media-plugins/gst-plugins-cdparanoia/files/gst-plugins-base-1.18.4-use-system-libs.patch
```diff
@@ -0,0 +1,58 @@
...
+
+diff --git a/ext/meson.build b/ext/meson.build
+index d984b5087..178b5169b 100644
+--- a/ext/meson.build
++++ b/ext/meson.build
+@@ -1,6 +1,6 @@
+ subdir('alsa')
+ subdir('cdparanoia')
+-subdir('gl')
++#subdir('gl')
```
thesamesam, line 15: From the [devmanual](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#how):
>Removed lines should not appear in the patch because they are commented — just remove them entirely. Patches show removed lines by prefixing them with a -, so no information is lost by simply deleting lines rather than commenting them out (which adds noise). This makes the patch shorter and more readable.

I assume this is Gentoo specific and can't be upstreamed too? May be worth putting a bit more info at the top of the patches.

thesamesam (changes requested):
Please also do one commit per package (if we're going to do what you did, it'd be media-libs/gstreamer*, not just media-libs).

i-garrison:
Closing in favor of https://github.com/gentoo/gentoo/pull/21541

PR #21514, opened 2021-07-04
Remove circular gcc deps when building packages with clang

lyssdod:
This allows to get completely gcc-free `libcxx`, `libcxxabi` and [LLVM] `libunwind` libraries when trying to use Clang as a main system compiler. Since "proper" Clang setup is kind of tricky now, additional USE flag shouldn't do much harm.

```
$ ldd /usr/lib/libc++_shared.so 
	ldd (0x7f9d390b6000)
	libunwind.so.1 => /usr/lib/libunwind.so.1 (0x7f9d38fcf000)
	libc.so => ldd (0x7f9d390b6000)
	libc++abi.so.1 => /usr/lib/libc++abi.so.1 (0x7f9d38f8e000)
```

```
$ ldd /usr/lib/libunwind.so 
	ldd (0x7f18a9723000)
	libc.so => ldd (0x7f18a9723000)
```

Without this change, mentioned libs link either to `libgcc_s.so`, `libatomic.so`, or both.


lyssdod:
Trying to setup system with Clang only according to wiki, USE for toolchain is:

```
sys-devel/clang compiler-rt default-compiler-rt default-lld default-libcxx llvm-libunwind
sys-devel/clang-runtime compiler-rt libcxx
sys-libs/compiler-rt clang
sys-libs/compiler-rt-sanitizers clang
sys-libs/libcxx libunwind libcxxabi
sys-libs/libcxxabi libunwind
sys-libs/llvm-libunwind compiler-rt
```

PR #21518, opened 2021-07-04
media-libs/mesa: Fix vulkan overlay mesa flags

DarkDefender:
These flags were changed to be in a list variable instead of individual
meson flags.

Mesa commit:
https://gitlab.freedesktop.org/mesa/mesa/-/commit/54fe5b0482df0f066384b274796d4081c2a1968c

Live and 21.1.x versions affected.


mattst88:
Thank you very much!

PR #21524, opened 2021-07-04
virtual/dev-manager: remove obsolete virtual

mm1ke:
Hi,

This PR removes `virtual/dev-manager` from the gentoo tree. The virtual isn't used by any package in tree and seems to exists for no particular reason. I guess it was used by packages in the past. For now i've directly removed it but I can update this PR to only mask it first if wanted.

Since this is not the only dead virtual in tree, i also have some general questions regarding virtuals:
* is there a reason to keep unused virtuals around? I haven't found any rules about this. I could remove many other virtuals too (though, i wouldn't touch the perl virtuals) but i'm afraid i'm missing something.
* when is it useful to add/have a virtual? Besides many unused virtuals there are also many who have only 1,2 or 3 customers. I think some of them could be replaced by a simple `|| ( ... )` (for example `virtual/ooo` only used by `app-office/unoconv` ). Any thoughts on this?

A full list of virutals and their usage can be found here: https://gentooqa.levelnine.at/results/gentoo/stats/ebuild_virtual_use_statistics/sort-by-filter/
Note: The first row (QTY) shows how many ebuilds (not packages) use a particular virtual.

@SoapGentoo I guess this might be interesting for you :)


ionenwks:
This is quite the opposite, this is used by everyone except prefix users. You should have it installed as well.

It's in `profiles/base/packages` making it part of the base system set to ensure everyone has a dev-manager

mm1ke:
Damn, you're right, i haven't check the obvious. Of course, I have it installed too..
Thanks for the hints, i'll have an closer look on this.

PR #21525, opened 2021-07-04
media-gfx/kuickshow: bump to 0.10.2

chrgerloff:
Fixes the issue in https://bugs.gentoo.org/790200.
Also addresses a number of other deprecated API calls.


juippis (approved):
In general I dislike how the version requirements is handled... since it seems useless for a long time already. But works and doesn't really cause harm, just a visual annoyance. Thanks!

PR #21528, opened 2021-07-04
selinux: ~arm keywords

jpds:


thesamesam:
One commit per package next time! I can share a script if it's useful.

PR #21530, opened 2021-07-04
dev-util/android-tools: New upstream, version bump

mid-kid:


mid-kid:
Just noticed that this current build fails with:
```
$ adb start-server
* daemon not running; starting now at tcp:5037
ADB server didn't ACK
Full server startup log: /tmp/adb.1000.log
Server had pid: 2629
--- adb starting (pid 2629) ---
adb I 07-05 02:26:18  2629  2629 main.cpp:62] Android Debug Bridge version 1.0.41
adb I 07-05 02:26:18  2629  2629 main.cpp:62] Version 31.0.0p1-android-tools
adb I 07-05 02:26:18  2629  2629 main.cpp:62] Installed as /usr/bin/adb
adb I 07-05 02:26:18  2629  2629 main.cpp:62] 
adb I 07-05 02:26:18  2629  2629 auth.cpp:417] adb_auth_init...
adb: symbol lookup error: adb: undefined symbol: OPENSSL_free

* failed to start daemon
error: cannot connect to daemon
```

I don't exactly get why this happens, since a manual build works just fine. Is it `-Wl,--as-needed`?

mid-kid:
Fixed the issue. Turns out boringssl wasn't being linked statically because of cmake.eclass' default `-DBUILD_SHARED_LIBS=ON`.

zmedico (changes requested):
Thank you! Ultimately, I want any patches to be acquired via SRC_URI rather than $FILESDIR.

mid-kid:
Two of those patches are mine, one can be fetched from a PR, and the big one can be fetched off of the relevant commit but doesn't apply cleanly and requires extra processing...
Should I rehost these patches or something?

zmedico:
> Two of those patches are mine, one can be fetched from a PR, and the big one can be fetched off of the relevant commit but doesn't apply cleanly and requires extra processing... I found this significantly cleaner...
> Should I rehost these patches or something?

I was planning to do something like this but if you can do it then that would be great. The commits can refer to a gentoo branch in your fork:

https://github.com/nmeum/android-tools/commit/df7bb55525aa93831874b2be8d7b21a5101d1466.patch -> ${P}-boringssl-fix-compilation-with-gcc-11.X.patch

mid-kid:
@zmedico I hope this is acceptable

zmedico (approved):
Perfect, thank you!

PR #21535, opened 2021-07-05
sys-libs/compiler-rt-sanitizers: fix building with clang on musl

lyssdod:
Not sure if this should go here or in https://github.com/gentoo/musl ?


lyssdod:
Closing as this seems to belong to `musl` overlay.

PR #21540, opened 2021-07-05
app-i18n/transifex-client: add new version 0.14.3

robert7k:


juippis (approved):
```
 *   expected: (unset)
```

PR #21544, opened 2021-07-06
app-emulation/wine-vanilla: revert winegcc '=' parameter handling

trofi:
LDFLAGS=-Wl,--defsym=__gentoo_check_ldflags__=0 causes
wine-vanilla-6.12 build failures in form of:

```
tools/winegcc/winegcc -o dlls/acledit/acledit.dll.so ... \
    -Wl,--defsym=__gentoo_check_ldflags__=0

ld:--defsym:0: syntax error
winegcc: /usr/lib/ccache/bin/x86_64-pc-linux-gnu-gcc failed
```

The change is a revert of upstream commit fcda0afd
("winegcc: Support -Wl,foo=... style linker options.")

Bug: https://bugs.winehq.org/show_bug.cgi?id=51413
Closes: https://bugs.gentoo.org/800809
Signed-off-by: Sergei Trofimovich <slyfox@gentoo.org>

### DIFF
--- a/app-emulation/wine-vanilla/files/wine-vanilla-6.12-winegcc-equals-args.patch
+++ b/app-emulation/wine-vanilla/files/wine-vanilla-6.12-winegcc-equals-args.patch
@@ -1,0 +1,13 @@
+https://bugs.winehq.org/show_bug.cgi?id=51413
+https://bugs.gentoo.org/800809
+--- a/tools/winegcc/winegcc.c
++++ b/tools/winegcc/winegcc.c
+@@ -1982,7 +1982,7 @@ int main(int argc, char **argv)
+                     if (strncmp("-Wl,", opts.args->base[i], 4) == 0)
+ 		    {
+                         unsigned int j;
+-                        strarray* Wl = strarray_fromstring(opts.args->base[i] + 4, ",=");
++                        strarray* Wl = strarray_fromstring(opts.args->base[i] + 4, ",");
+                         for (j = 0; j < Wl->size; j++)
+                         {
+                             if (!strcmp(Wl->base[j], "--image-base") && j < Wl->size - 1)
--- a/app-emulation/wine-vanilla/wine-vanilla-6.12-r1.ebuild
+++ b/app-emulation/wine-vanilla/wine-vanilla-6.12-r1.ebuild
@@ -1,0 +1,579 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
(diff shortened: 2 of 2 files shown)


app-emulation/wine-vanilla/files/wine-vanilla-6.12-winegcc-equals-args.patch
```diff
@@ -0,0 +1,13 @@
+https://bugs.winehq.org/show_bug.cgi?id=51413
```
sarnex, line 1: Do we need to do this for app-emulation/wine-staging-6.12 too?

app-emulation/wine-vanilla/files/wine-vanilla-6.12-winegcc-equals-args.patch
```diff
@@ -0,0 +1,13 @@
...
+https://bugs.gentoo.org/800809
```
sarnex, line 2: We need to keep in sync with ::wine, so please make sure this change makes it to both repos

sarnex (approved):

app-emulation/wine-vanilla/files/wine-vanilla-6.12-winegcc-equals-args.patch
mgorny: Yes.

mgorny:
I can confirm that adding the revert locally fixes the build of wine-staging for me.

trofi:
Looks like `wine-staging` lacks a few `wine-vanilla` changes like `AR` propagation which makes it fail to build on my system. I'm not sure how (or when) ebuilds need to be synced to `::wine`. Closing the request to let it get sorted by maintainer.

sarnex:
@trofi I'm happy to deal with the ::wine stuff, but what do you want me to do here? Should I push this for wine-vanilla and we should have a separate bug to track the wine-staging work?

trofi:
I think it would help having slightly clearer contribution guideline and expectation. Here is my understanding:
1. Prepare a patch against `::gentoo`'s `wine-vanilla` package (and against `::gentoo`'s `wine-staging` package if applies).
2. Get it reviewed by https://wiki.gentoo.org/wiki/Project:Wine, then push to `::gentoo`.
3. Sync the reviewed change to `::wine` by pushing the same change to https://gitweb.gentoo.org/repo/proj/wine.git/. (Do ACLs allow push by any gentoo dev?)

If it sounds about right I can add it to https://wiki.gentoo.org/wiki/Project:Wine and resend patches.

sarnex:
Thanks, yeah we should have a formal contribution guideline

Slight reordering:

1. Prepare a patch against ::wine's wine-vanilla package (and against ::wine's wine-staging package if applies).
2. Get it reviewed by https://wiki.gentoo.org/wiki/Project:Wine, then push to ::wine.
3. Sync the reviewed change to ::gentoo by pushing the same change to gentoo.git (No idea about the ACL, probably any dev)

I've added this to https://wiki.gentoo.org/wiki/Project:Wine/Policies_and_Procedures#User_contributions

trofi:
Trying on https://github.com/gentoo/gentoo/pull/21609 and https://github.com/gentoo/gentoo/pull/21618

PR #21545, opened 2021-07-07
Add Pillow zlib dependency for freedoom ebuilds

vilhelmgray:


ionenwks:
Headers are still showing 2020, be good to use pkgcheck/repoman even when doing trivial fixes.

I was going to fix that on merge, but while we're here, does python 3.10 work? Be nice to add it and remove 7.

ionenwks:
On a side-note, xdg_src_prepare already runs eapply_user, no need to run it twice. Thankfully doesn't fail given there's guard in places not to apply user patches twice.

vilhelmgray:
> Headers are still showing 2020, be good to use pkgcheck/repoman even when doing trivial fixes.
> 
> I was going to fix that on merge, but while we're here, does python 3.10 work? Be nice to add it and remove 7.

Sure, I can test if python 3.10 works. How can I tell portage to use 3.10 when I emerge a package?

ionenwks:
> Sure, I can test if python 3.10 works. How can I tell portage to use 3.10 when I emerge a package?

`PYTHON_TARGETS="python3_9 python3_10" emerge -1 freedoom-data` would be enough if 10 is in COMPAT, it'll pick the most recent version in a situation where any works. Alternatively, may want to permanently add 10 to your targets, messing with this is not recommended for users but maintainers need to test 10 relatively often.

Should see: ` * Using python3.10 to build` when building to confirm.

vilhelmgray:
Thanks, looks like 3.10 works fine so I went ahead and updated the ebuilds accordingly.

ionenwks (approved):
Thanks, since eapply_user isn't worth its own commit (like removing a blank line), I've squashed it into the python3.10 one with a short note. Hope you don't mind.

PR #21548, opened 2021-07-07
sys-power/thermald: add doc USE flag, make dep on dev-util/gtk-doc optional

nephros:
On one of my systems, this is the only package that unconditionally depends on dev-util/gtk-doc.

This makes this dependency optional by way of a `doc` USE flag.

Signed-off-by: Peter Gantner <gentoo@nephros.org>

### DIFF
--- a/sys-power/thermald/thermald-2.4.5-r1.ebuild
+++ b/sys-power/thermald/thermald-2.4.5-r1.ebuild
@@ -12,7 +12,7 @@ SRC_URI="https://github.com/intel/thermal_daemon/archive/v${PV}.tar.gz -> ${P}.t
 LICENSE="GPL-2+"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE=""
+IUSE="-doc"
 
 RDEPEND="
 	dev-libs/dbus-glib:=
@@ -22,16 +22,16 @@ RDEPEND="
 	sys-power/upower
 	sys-apps/dbus:="
 DEPEND="${RDEPEND}
-	dev-util/gtk-doc
+	doc? ( dev-util/gtk-doc )
 	dev-util/glib-utils"
 
 S=${WORKDIR}/thermal_daemon-${PV}
 DOCS=( thermal_daemon_usage.txt README.txt )
-
 src_prepare() {
 	sed -i -e "/group=/s/power/wheel/g" \
 		data/org.freedesktop.thermald.conf || die
 
+	use doc || PATCHES+=("${FILESDIR}/${PN}-no-docs.patch")
 	default
 	eautoreconf
 }
--- a/sys-power/thermald/files/thermald-no-docs.patch
+++ b/sys-power/thermald/files/thermald-no-docs.patch
@@ -1,0 +1,41 @@
+--- a/Makefile.am	2021-07-07 10:08:41.193127800 +0200
++++ b/Makefile.am	2021-07-07 10:08:49.336489915 +0200
+@@ -1,6 +1,6 @@
+ include $(GLIB_MAKEFILE)
+ 
+-SUBDIRS = . docs data
++SUBDIRS = . data
+ 
+ ACLOCAL_AMFLAGS =
+ 
+--- a/autogen.sh	2021-07-07 10:12:43.853933437 +0200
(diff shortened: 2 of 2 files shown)


dlan17:
I'm fine with adding USE=doc support here..
but, could you ping upstream? see if they willing to accept the conditional doc option patch?
it's always good to send/merge it upstream..

nephros:
> I'm fine with adding USE=doc support here..
> but, could you ping upstream? see if they willing to accept the conditional doc option patch?
> it's always good to send/merge it upstream..

I can try, but I doubt that it will be accepted (also I don't speak autotools really).

The dep isn't really a problem in a traditional distro where BDeps and RDeps are resolved on separate systems, resulting in the classic package + package-devel + package-docs split from the same build result.

dlan17:
> > I'm fine with adding USE=doc support here..
> > but, could you ping upstream? see if they willing to accept the conditional doc option patch?
> > it's always good to send/merge it upstream..
> 
> I can try, but I doubt that it will be accepted (also I don't speak autotools really).
> 
you introduce a patch here : sys-power/thermald/files/thermald-no-docs.patch 
It's better to let thermald upstream accept/merge it, so we won't carry it downstream forever..

> The dep isn't really a problem in a traditional distro where BDeps and RDeps are resolved on separate systems, resulting in the classic package + package-devel + package-docs split from the same build result, and build deps don't matter much.
> 
> Or are you talking about this ebuild?
> 
No, the change to the ebuild is just fine, or let's worry about it later when we move to EAPI=8

sys-power/thermald/thermald-2.4.5-r1.ebuild
```diff
@@ -22,16 +22,16 @@ RDEPEND="
 	sys-power/upower
 	sys-apps/dbus:="
 DEPEND="${RDEPEND}
-	dev-util/gtk-doc
+	doc? ( dev-util/gtk-doc )
```
dlan17, line 25: sorry, I didn't review the patch carefully..

you should introduce a "--enable-doc" option for upstream, and also avoid the conditional patch here

sys-power/thermald/thermald-2.4.5-r1.ebuild
```diff
@@ -22,16 +22,16 @@ RDEPEND="
...
 	dev-util/glib-utils"
 
 S=${WORKDIR}/thermal_daemon-${PV}
 DOCS=( thermal_daemon_usage.txt README.txt )
-
 src_prepare() {
 	sed -i -e "/group=/s/power/wheel/g" \
 		data/org.freedesktop.thermald.conf || die
 
+	use doc || PATCHES+=("${FILESDIR}/${PN}-no-docs.patch")
```
dlan17, line 34: sorry, I didn't review the patch carefully..

you should introduce a "--enable-doc" option for upstream, and also avoid the conditional patch here

nephros: > sorry, I didn't review the patch carefully..
> 
> you should introduce a "--enable-doc" option for upstream, and also avoid the conditional patch here

Thank you for your review and suggestion. After some more consideration I'm afraid I'll have to NACK that.

I actually don't know autotools and friends to a noteworthy degree, I don't think upstream will be interested, and my goal was mainly to get rid of that dependency on my Gentoo installations (it seemed most if not all other users of gtk-doc are conditional).

So, feel free to close/reject this PR, I'm happy to maintain it in my local overlays as-is.

dlan17: > > sorry, I didn't review the patch carefully..
> > you should introduce a "--enable-doc" option for upstream, and also avoid the conditional patch here
> 
> Thank you for your review and suggestion. After some more consideration I'm afraid I'll have to NACK that.
> 
> I actually don't know autotools and friends to a noteworthy degree, I don't think upstream will be interested, and my goal was mainly to get rid of that dependency on my Gentoo installations (it seemed most if not all other users of gtk-doc are conditional).
> 
I'll ping upstream first about the idea of adding optional 'doc' configuration

> So, feel free to close/reject this PR, I'm happy to maintain it in my local overlays as-is.
generally, trying to maintain patches downstream is kind of painful, it will easily become broken..
for now, please feel free to maintain it as in your local overlay, thanks

PR #21549, opened 2021-07-07
kernel-install.eclass: print uname() info after test

gyakovlev:
I thought it'd be nice to see version string and some info once it's booted in `src_test()`

stripped size on my system is 501k

### DIFF
--- a/eclass/kernel-install.eclass
+++ b/eclass/kernel-install.eclass
@@ -184,13 +184,26 @@ kernel-install_create_init() {
 
 	cat <<-_EOF_ >"${T}/init.c" || die
 		#include <stdio.h>
+		#include <sys/utsname.h>
+
 		int main() {
+			struct utsname u;
+			int ret = uname(&u);
+			if (ret != 0) {
+				printf("uname() failed, but that's fine\n");
+			}
+			else {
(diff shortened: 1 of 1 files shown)


Zlogene (approved):

eclass/kernel-install.eclass
```diff
@@ -184,13 +184,28 @@ kernel-install_create_init() {
...
 		#include <stdio.h>
+		#include <sys/utsname.h>
+
+		int uname(struct utsname *buf);
+
 		int main() {
+			struct utsname u;
+			int ret = uname(&u);
+			if (ret != 0) {
+				printf("uname() failed, but that's fine");
```
mgorny, line 195:
```suggestion
				printf("uname() failed, but that's fine\n");
```

eclass/kernel-install.eclass
```diff
@@ -184,13 +184,28 @@ kernel-install_create_init() {
 
 	cat <<-_EOF_ >"${T}/init.c" || die
 		#include <stdio.h>
+		#include <sys/utsname.h>
+
+		int uname(struct utsname *buf);
```
mgorny, line 189: Why do you need to declare it if you included `sys/utsname.h` already?

gyakovlev: no need indeed.removed.

mgorny (approved):
LGTM

PR #21551, opened 2021-07-07
sci-physics/thepeg: min java 1.8

vaukai:


AndrewAmmerlaan:
Thanks, I also bumped to the latest version when merging to fix a compile failure with gcc11

PR #21556, opened 2021-07-07
games-sports/stormbaancoureur: bump EAPI, fix build

DarthGandalf:


games-sports/stormbaancoureur/files/stormbaancoureur-2.1.6-clang.patch
```diff
@@ -0,0 +1,52 @@
+diff --git a/carpettrack.h b/carpettrack.h
+index 4bcd7b8..50dfdc9 100644
+--- a/carpettrack.h
```
ionenwks, line 3: Be good to get into the habit of cleaner patches, i.e. strip metadata (like these diff --git and index lines) and at least have some short description and the bgo and/or upstream links in the patch header.

For the strip metadata bit, the sed in the devmanual is pretty useful
https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#clean-patch-howto

DarthGandalf: Done.

ionenwks (approved):
Thanks.

And thanks for that `tc-export CXX` too, I tried the old one first but failed on not finding c++, and then seen you've already fixed that :+1:

PR #21558, opened 2021-07-08
IUSE compression flags simplification

mattst88:
Seems silly to me to have to explicitly enable support for things you already have on your system: `app-arch/bzip2`, `app-arch/xz-utils`, and `app-arch/zstd`.

Enable them by default, and rename some stray IUSE `xz` → `lzma` and `gzip` → `zlib` so they're also handled.


dev-util/diffoscope/diffoscope-177.ebuild
```diff
@@ -17,9 +17,9 @@ LICENSE="GPL-3+"
 SLOT="0"
 KEYWORDS="~amd64 ~ppc64 ~x86"
 IUSE="acl binutils bzip2 libcaca colord cpio +diff docx dtc e2fsprogs file
```
thesamesam, line 19: Honestly, this needs an optfeature conversion, but let's do this for now.

profiles/default/linux/make.defaults
```diff
@@ -11,7 +11,7 @@
 
 
 # Default starting set of USE flags for all default/linux profiles.
-USE="crypt ipv6 ncurses nls pam readline ssl tcpd zlib"
+USE="bzip2 crypt ipv6 lzma ncurses nls pam readline ssl tcpd zlib zstd"
```
thesamesam, line 14: I think I'm overall happy with this but I'd like to, at some point, audit the packages with these 3 new flags and figure out if it makes sense to have them and +IUSE them if necessary. Then drop it again.

Maybe we should put a TODO above it and/or a note to consider flipping IUSE defaults instead for future changes and link to the thread for some (historical) context.

thesamesam (approved):

profiles/default/linux/make.defaults
thesamesam: I'm also a little bit interested in whether you can actually get a built system (like with --root) without these, but it doesn't really affect the decision making for this change.

mattst88: Agreed, `optfeature` would be even better.

I think it's impossible to have a Gentoo system without `sys-libs/zlib`. It's an unconditional dependency of `sys-devel/binutils`, `sys-devel/gcc`, and `sys-devel/llvm`.

dev-util/diffoscope/diffoscope-177.ebuild
mattst88: Agreed, `optfeature` would be even better.

mattst88:
There, I've removed the one-line `make.defaults` change, per @orlitzky's suggestion and replaced it with 166 commits.

Maintainers: Please provide feedback on whether you think your packages' `bzip2`, `lzma`, `zstd`, and `zlib` USE flags should not be enabled by default.

The conversion was automated, except for `media-video/ffmpeg` and `net-mail/dovecot`, with:

     sed -i -e '/IUSE/ s/\([" ]\)\(bzip2\|lzma\|zstd\|zlib\)/\1+\2/g' \
        $(git grep -l 'IUSE.*[" ]\(bzip2\|lzma\|zstd\|zlib\)')

inserting the `+` characters and

    for x in $(git diff --stat=160 | grep ebuild | cut -d/ -f1,2  | sort -u);
    do
        (cd $x; git commit -s -m "$(cat-pkg): Enable compression USE by default" .)
    done

making the git commits.

thesamesam:
(Can you do the please reassign magic in the title which the bot mentions so it picks up the new packages?)

mattst88:
> (Can you do the please reassign magic in the title which the bot mentions so it picks up the new packages?)

I removed the `assigned` label — I hope that does the trick.

mattst88:
> > (Can you do the please reassign magic in the title which the bot mentions so it picks up the new packages?)
> 
> I removed the `assigned` label — I hope that does the trick.

Oh, `[please reassign]`. Neat.

thesamesam:
> > (Can you do the please reassign magic in the title which the bot mentions so it picks up the new packages?)
> 
> I removed the `assigned` label — I hope that does the trick.

If it doesn’t (it might, I’m on mobile so can’t check), try:
> In order to force reassignment and/or bug reference scan, please append [please reassign] to the pull request title.

thesamesam:
Manual ping: @tonyvroon @Polynomial-C

ceamac:
I checked for `media-libs/netpbm`.  The `zlib` USE flag is basically useless on its own.  It can be enabled by default for now, and I'll remove it with the next version bump and fix the dependency.  `libz` is required only for certain USE flags, not independently.

trofi:
LGTM for lynx

grobian:
ok for carbon-c-relay

tastytea:
LGTM for games-simulation/simutrans

Polynomial-C (approved):

amadio:
Ok for dev-util/perf. However, why not enable compression USE flags at the profile level instead? I usually enable compression flags globally rather than per package.

mattst88:
> Ok for dev-util/perf. However, why not enable compression USE flags at the profile level instead? I usually enable compression flags globally rather than per package.

I had the same thought, so I sent [this patch](https://archives.gentoo.org/gentoo-dev/message/2a12edf26c36dada663fac7f3e2a7188) to gentoo-dev@. Just see for yourself the replies...

wraeth:
Is there an actual problem this is fixing?

ulm:
*Not* o.k. for app-editors/emacs. This belongs in profiles (system, or user's).

mkszuba:
OK for _app-emulation/fuse-utils_, _app-emulation/libspectrum_, _dev-cpp/prometheus-cpp_ and _media-gfx/gmic_ - while in general I agree with ulm, having just checked respective sources the default upstream behaviour of all four packages is to enable this.

profiles/default/linux/make.defaults
ulm: > I think it's impossible to have a Gentoo system without `sys-libs/zlib`. It's an unconditional dependency of `sys-devel/binutils`, `sys-devel/gcc`, and `sys-devel/llvm`.

So enable it in profiles, not in hundreds of ebuilds. This isn't package specific, so it doesn't belong in IUSE defaults.

orlitzky:
> Is there an actual problem this is fixing?

Read the thread on -dev. There's no reason at all to enable this, and the mass pull request is just an attempt to annoy everyone as punishment for asking "what are you trying to fix?"

NACK for any of my packages. Shit's dumb and this is childish.

Whissi:
NACK for dev-libs/openssl, enabling compression creates security problems (CRIME vulnerability; TLS compression was already dropped in any major browser and shouldn't be used). We are already looking for a way to find a way that this won't get enabled by users by accident via global USE=zlib.

wraeth:
> Read the thread on -dev.

I've read the thread on -dev.

I'm asking the author to provide justification beyond "deps are present so it should probably be enabled".  I'm not keen on unnecessary churn (and rebuilds for users) because someone decided some flags should be enabled for convenience but I'd rather they explain why they think it's necessary, if there's actually a problem that I'm not aware of.

epsilon-0:
ok for gui-apps/waypipe.

mattst88:
> > Is there an actual problem this is fixing?
> 
> Read the thread on -dev. There's no reason at all to enable this, and the mass pull request is just an attempt to annoy everyone as punishment for asking "what are you trying to fix?"
> 
> NACK for any of my packages. Shit's dumb and this is childish.

This is not the case at all, and I'd appreciate the benefit of the doubt rather than accusations like this.

As you can see, there have been *lots* of maintainers responding by saying that that these flags should be enabled by default in their packages. This is exactly what you suggested. I'll happily drop patches to packages that where the maintainers think the flags should not be enabled.

> NACK for dev-libs/openssl, enabling compression creates security problems (CRIME vulnerability; TLS compression was already dropped in any major browser and shouldn't be used). We are already looking for a way to find a way that this won't get enabled by users by accident via global USE=zlib.

Thanks! Will drop this patch.

(And for @orlitzy: I think we should consider this a win as well)

> _Not_ o.k. for app-editors/emacs. This belongs in profiles (system, or user's).

@ulm: I did it this way precisely because @orlitzky was very against putting this in `make.defaults` (and IIUC is very against global USE flags in general).

But `zlib` is *already* in `make.defaults`, so as is this commit is literally a no-op, but perhaps in the future @orlitzky proposes we won't have `zlib` in `make.defaults`.

In any case, I'll drop the patch to emacs.
 
> I'm asking the author to provide justification beyond "deps are present so it should probably be enabled". I'm not keen on unnecessary churn (and rebuilds for users) because someone decided some flags should be enabled for convenience but I'd rather they explain why they think it's necessary, if there's actually a problem that I'm not aware of.

There's a general problem that happens in Gentoo where something doesn't work that works on another distro. You spend a bunch of time debugging it, only to discover that it's because some dumb USE flag was disabled — something like the problem with `kmod` not supporting xz'd kernel modules without `USE=lzma` (which is was fixed by enabling `USE=lzma` by default in the ebuild).

@orlitzky suggested that maintainers should be the ones to decide whether the flags are enabled by default, since they have a better idea of what they do. That is a reasonable position. To that end, I've speculatively done the commits, and I'm simply asking maintainers what they think. Sorry for the spam; I don't really know a better way.

orlitzky:
To anyone at risk of believing this retarded victim act: this is not what I suggested.

orlitzky:
> 
> But `zlib` is _already_ in `make.defaults`, so as is this commit is literally a no-op, but perhaps in the future @orlitzky proposes we won't have `zlib` in `make.defaults`.
> 

It's not a no-op, because the way the variables are stacked, setting `USE="-zlib"` in `make.conf` will no longer disable zlib. (Which you would know, if you had any idea what you're trying to do.)

floppym:
> It's not a no-op, because the way the variables are stacked, setting USE="-zlib" in make.conf will no longer disable zlib. (Which you would know, if you had any idea what you're trying to do.)

I think you are mistaken. With the default USE_ORDER setting, setting `USE="-zlib"` in /etc/portage/make.conf overrides both make.defaults in profiles and IUSE defaults in ebuilds.

orlitzky:
> > It's not a no-op, because the way the variables are stacked, setting USE="-zlib" in make.conf will no longer disable zlib. (Which you would know, if you had any idea what you're trying to do.)
> 
> I think you are mistaken. With the default USE_ORDER setting, setting `USE="-zlib"` in /etc/portage/make.conf overrides both make.defaults in profiles and IUSE defaults in ebuilds.

Ok, you're right. (It's bad form to be both obnoxious and wrong.) It does still increase the priority, however, meaning that you can no longer override the defaults with profiles like many of us do at work. And it's most definitely not a no-op.

wraeth:
> There's a general problem that happens in Gentoo where something doesn't work that works on another distro. You spend a bunch of time debugging it, only to discover that it's because some dumb USE flag was disabled — something like the problem with `kmod` not supporting xz'd kernel modules without `USE=lzma` (which is was fixed by enabling `USE=lzma` by default in the ebuild).

So it _is_ just a convenience thing - I'd consider the kmod[lzma] issue to be a special case.  I think this sort of thing belongs in a profile, and specifically I think it's suited for the desktop profile.

If, however, the consensus is for it not to be in a profile, I think a better way to go about making this change would be to file a bug for each package you want updated.  That way maintainers can adjust flags as part of a version bump, instead of needlessly causing changed-use rebuilds for users on otherwise-working systems.

mattst88:
I pushed the commits that rename some stray IUSE flags:

````
commit 948123ac1484eb76e20e13d38b8bbd9138880b2d
Author: Matt Turner <mattst88@gentoo.org>
Date:   Wed Jul 7 22:10:25 2021 -0700

    sys-fs/squashfs-tools-ng: Rename IUSE=xz->lzma
    
    Signed-off-by: Matt Turner <mattst88@gentoo.org>

commit 20662810f676778313c4654393fff47e13a2a397
Author: Matt Turner <mattst88@gentoo.org>
Date:   Wed Jul 7 22:08:30 2021 -0700

    sci-physics/pythia: Rename IUSE=gzip->zlib
    
    Signed-off-by: Matt Turner <mattst88@gentoo.org>

commit f21f1d5b81b47ceb5b371497d25cb99fc4a7f045
Author: Matt Turner <mattst88@gentoo.org>
Date:   Wed Jul 7 22:04:21 2021 -0700

    dev-util/archdiff: Rename IUSE=gzip->zlib
    
    Signed-off-by: Matt Turner <mattst88@gentoo.org>

commit 0c35580f8e9a728965508cd9c0824e2b54c02df5
Author: Matt Turner <mattst88@gentoo.org>
Date:   Wed Jul 7 22:16:11 2021 -0700

    stardict.eclass: Rename IUSE=gzip->zlib
    
    The flag actually controls a dependency on app-arch/gzip, but you
    already have that too. Name the flag IUSE=zlib so it'll be enabled by
    default.
    
    Signed-off-by: Matt Turner <mattst88@gentoo.org>
    Closes: https://github.com/gentoo/gentoo/pull/21558
    Signed-off-by: Matt Turner <mattst88@gentoo.org>
````

I'm leaving the rest. It's not sufficiently valuable to deal with the abuse from @orlitzky.

orlitzky:
Don't forget to revbump all those packages where you just changed `RDEPEND`!

PR #21563, opened 2021-07-08
media-gfx/hydrus: Add 446, drop 442 

vaartis:
No changes to the ebuild. Drop 442 to keep 4 newer versions in the tree.

@thesamesam


vaartis:
ping

PR #21566, opened 2021-07-08
media-libs/glfw: rename wayland useflag to wayland-only

Alessandro-Barbieri:
Closes: https://bugs.gentoo.org/616968

### DIFF
--- a/media-libs/glfw/glfw-3.3.3.ebuild
+++ b/media-libs/glfw/glfw-3.3.3.ebuild
@@ -12,11 +12,11 @@ SRC_URI="https://github.com/glfw/glfw/archive/${PV}.tar.gz -> ${P}.tar.gz"
 LICENSE="ZLIB"
 SLOT="0"
 KEYWORDS="amd64 ~arm ~arm64 ~hppa ~ppc64 x86"
-IUSE="wayland"
+IUSE="wayland-only"
 
 RDEPEND="
 	x11-libs/libxkbcommon
-	!wayland? (
+	!wayland-only? (
 		virtual/opengl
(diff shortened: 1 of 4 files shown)


pinkflames:
FWIW, I support this.

mattst88:
I think this is a good solution. We made a similar change with `gles2` → `gles2-only` a while back.

chewi:
Ahem, I like the idea but I don't think you tested this. Looks like your `usex` change is the wrong way round.

```
>>> Configuring source in /var/tmp/portage/media-libs/glfw-3.3.4/work/glfw-3.3.4 ...
 * ERROR: media-libs/glfw-3.3.4::gentoo failed (configure phase):
 *   USE Flag 'wayland' not in IUSE for media-libs/glfw-3.3.4
 * 
 * Call stack:
 *          ebuild.sh, line  127:  Called src_configure
 *        environment, line 2103:  Called usex 'wayland' 'wayland-only'
 *   phase-helpers.sh, line  213:  Called use 'wayland'
 *   phase-helpers.sh, line  252:  Called die
 *   				die "USE Flag '${u}' not in IUSE for ${CATEGORY}/${PF}"
```

ionenwks:
I was pondering what to do with this (other distros have done things such as wrappers to load the right library), I guess could potentially have used a eselect to let users switch, but I don't feel it's overly needed as long as it's clear what users are getting. So I'm okay with this as well.

ionenwks (approved):
Guess I'll do this now and egl(+) at same time.

I'll take only the changes to 3.3.4-r1, this is due to be stabled soon'ish either way.

PR #21570, opened 2021-07-08
sys-fs/bees: remove quotes from PREFIX

Alessandro-Barbieri:
the quotes will end up in the systemd unit
causing it to fail to start


ionenwks:
Oh, thanks. Didn't think to look at the systemd file when I suggested that, I'll have a quick look and merge this soon.

ionenwks:
Went for a different fix given this still fails with spaces if remove quotes (not to say prefix support is very important here).

But thanks for bringing this to my attention, tell me if anything still wrong.

PR #21571, opened 2021-07-09
sys-apps/iproute2: handle "bpf" USE flag for supported arches

Polynomial-C:
This is needed so we can finally restore keywords for >=sys-apps/iproute2-5.12

### DIFF
--- a/profiles/arch/base/package.use.mask
+++ b/profiles/arch/base/package.use.mask
@@ -1,6 +1,10 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Lars Wendler <polynomial-c@gentoo.org> (2021-07-09)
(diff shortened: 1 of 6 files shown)


profiles/arch/base/package.use.mask
```diff
@@ -1,6 +1,10 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Lars Wendler <polynomial-c@gentoo.org> (2021-07-09)
+# dev-libs/libbpf lacks keywords for most arches
```
thesamesam, line 5: It's keyworded in most places now, just now alpha. Could we just mask in certain arches accordingly because they're in the minority?

thesamesam (changes requested):

Polynomial-C:
Oh, right. I didn't see latest versions have many more keywords...

PR #21573, opened 2021-07-09
media-gfx/freecad: add slot-operator for sci-libs/vtk

waebbl:
Following PR #21149, I added subslots for the sci-libs/vtk package.
This patch adds the subslot operator for 0.19.2 and 9999 ebuilds.

Also drops old 0.19.1-r1 ebuild


ionenwks:
Need a revbump when adding :=, this was recently added in the devmanual on this page:

Which, while right, I admit is kind of annoying on large packages like freecad :(

PR #21574, opened 2021-07-09
dev-libs/openssl: rename USE=zlib to USE=tls-compression

Whissi:
USE=zlib is a global USE flag and probably enabled by most users.
However, in OpenSSL, zlib is used for TLS compression which allows
for attacks like CRIME vulnerability. Its usage is discouraged and
banned in TLS 1.3+.

Renaming the USE flag allows us to opt-out from globally set USE=zlib.


SoapGentoo:
can we finally completely disable this for 3? TLS compression was a historic mistake.

mattst88:
Thanks!

mattst88:
So, push this? What is this waiting on?

PR #21578, opened 2021-07-10
sci-libs/libsigrokdecode: Added support for Python 3.9.

jyelloz:
This just imports a patch from https://github.com/sigrokproject/libsigrokdecode/commit/9b0ad5177bd692f7556a4756bdbd2da81d9c34ce
Bug: https://bugs.gentoo.org/801196


thesamesam:
Thanks! This will work, but see [here](https://bugs.gentoo.org/794592#c4). It seems like the more future proof fix is to force usage of 'Python' and let the eclasses do the rest. Or make it read ${EPYTHON} from the environment.

thesamesam:
Thank you! See https://bugs.gentoo.org/801196#c10.

PR #21579, opened 2021-07-10
media-libs/netpbm: bump to v10.86.23

ceamac:
Remove zlib USE flag and enable zlib when needed (for png, postscript or xml)
Do not drop the previous version (v10.86.22) yet, because if I resolve the 2 blockers I can request stabilization


ceamac:
I've created 2 bugs for the existing problems:
- https://bugs.gentoo.org/801442
- https://bugs.gentoo.org/801445

Will convert this PR to draft until they are resolved.

ceamac:
No longer needed, already bumped to v10.86.24.

PR #21581, opened 2021-07-10
media-video/oxine: use a dependency default flag

ceamac:
media-libs/netpbm is dropping zlib USE flag, so add it with (+).


thesamesam:
Thank you for handling this!

PR #21584, opened 2021-07-10
autotools.eclass: rename configure.in, EAPI 8

thesamesam:
automake is "eventually" going to fail when it detects configure.in.

Don't do this immediately - instead only on new EAPIs to avoid
breaking existing ebuilds.

We give an eqawarn (which is triggered by e.g. PORTAGE_ELOG_CLASSES="qa" and
is opt-in) for maintainers to upstream these where possible, but this isn't
always the case for software ehich otherwise works but is unmaintained upstream.

Bug: https://bugs.gentoo.org/426262

### DIFF
--- a/eclass/autotools.eclass
+++ b/eclass/autotools.eclass
@@ -4,7 +4,7 @@
 # @ECLASS: autotools.eclass
 # @MAINTAINER:
 # base-system@gentoo.org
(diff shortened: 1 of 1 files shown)


thesamesam:
@Polynomial-C

PR #21586, opened 2021-07-10
games-mud/tkmoo: EAPI8, ebuild improvements

mm1ke:

@ionenwks

### DIFF
--- a/games-mud/tkmoo/tkmoo-0.3.32-r1.ebuild
+++ b/games-mud/tkmoo/tkmoo-0.3.32-r1.ebuild
@@ -1,45 +1,49 @@
-# Copyright 1999-2016 Gentoo Foundation
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
-inherit eutils
+EAPI=8
 
-MY_PN=${PN/moo/MOO-light}
-MY_P=${P/moo/MOO-light}
+inherit desktop
+
+MY_P="tkMOO-light-${PV}"
+
 DESCRIPTION="MOO Client written in Tcl/Tk"
 HOMEPAGE="http://www.awns.com/tkMOO-light/"
 SRC_URI="http://www.awns.com/tkMOO-light/Source/${MY_P}.tar.gz"
+S="${WORKDIR}/${MY_P}"
 
 LICENSE="tkMOO"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE=""
 
 DEPEND="
-	>=dev-lang/tcl-8.3.3:0=
-	>=dev-lang/tk-8.3.3:0="
-RDEPEND=${DEPEND}
-
-S=${WORKDIR}/${MY_P}
+	dev-lang/tcl
+	dev-lang/tk
+"
+RDEPEND="${DEPEND}"
 
 PATCHES=(
 	"${FILESDIR}/${PV}-Makefile-noclean.patch"
-		"${FILESDIR}/${PV}-keys-workaround.patch"
+	"${FILESDIR}/${PV}-keys-workaround.patch"
 )
 
 src_compile() {
-	emake \
-		WISH="$(type -P wish)" \
-		TKMOO_LIB_DIR="/usr/$(get_libdir)/${MY_PN}" \
-		TKMOO_BIN_DIR=/usr/bin
+	local myemakeargs=(
+		WISH="$(type -P wish)"
+		TKMOO_LIB_DIR="${EPREFIX}/usr/share/${PN}"
+		TKMOO_BIN_DIR="${EPREFIX}/usr/bin"
+	)
+	emake "${myemakeargs[@]}"
 }
 
 src_install() {
-	emake \
-		TKMOO_LIB_DIR="${D}/usr/$(get_libdir)/${MY_PN}" \
-		TKMOO_BIN_DIR="${D}/usr/bin" \
-		install
+	local emakeargs=(
+		TKMOO_LIB_DIR="${ED}/usr/share/${PN}"
+		TKMOO_BIN_DIR="${ED}/usr/bin"
+	)
+	emake "${emakeargs[@]}" install
+
 	dodoc README dot.tkmoolightrc bugsmail.txt
 	dosym tkMOO-lite /usr/bin/tkmoo
-	make_desktop_entry tkmoo "tkMOO"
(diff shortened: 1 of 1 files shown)


games-mud/tkmoo/tkmoo-0.3.32-r1.ebuild
```diff
@@ -1,44 +1,45 @@
...
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE=""
 
 DEPEND="
-	>=dev-lang/tcl-8.3.3:0=
-	>=dev-lang/tk-8.3.3:0="
-RDEPEND=${DEPEND}
+	dev-lang/tcl:0=
+	dev-lang/tk:0="
```
ionenwks, line 20: The 0 slot isn't needed anymore either.

games-mud/tkmoo/tkmoo-0.3.32-r1.ebuild
```diff
@@ -1,44 +1,45 @@
...
+	"${FILESDIR}/${PV}-keys-workaround.patch"
 )
 
 src_compile() {
 	emake \
 		WISH="$(type -P wish)" \
-		TKMOO_LIB_DIR="/usr/$(get_libdir)/${MY_PN}" \
-		TKMOO_BIN_DIR=/usr/bin
+		TKMOO_LIB_DIR="${EPREFIX}/usr/$(get_libdir)/${MY_PN}" \
+		TKMOO_BIN_DIR="${EPREFIX}/usr/bin"
```
ionenwks, line 34: I personally came to prefer replacing those backslash `\` mess by a variable even if it's just for 2-3 lines. i.e.
```
local emakeargs=(
	VAR=val
)
emake "${emakeargs[@]}"
```
`\` is also more error prone, if you forget a backlash it doesn't necessarily fail and bash just sets the value in the environment.

games-mud/tkmoo/tkmoo-0.3.32-r1.ebuild
```diff
@@ -1,44 +1,45 @@
...
 	emake \
-		TKMOO_LIB_DIR="${D}/usr/$(get_libdir)/${MY_PN}" \
-		TKMOO_BIN_DIR="${D}/usr/bin" \
+		TKMOO_LIB_DIR="${ED}/usr/$(get_libdir)/${MY_PN}" \
+		TKMOO_BIN_DIR="${ED}/usr/bin" \
 		install
+
 	dodoc README dot.tkmoolightrc bugsmail.txt
 	dosym tkMOO-lite /usr/bin/tkmoo
 	make_desktop_entry tkmoo "tkMOO"
```
ionenwks, line 45: Does this install an icon? (didn't check)

ionenwks (commented):
Just a quick look (didn't mean to send previous as a single comment)

I was actually going to do this one tomorrow (next in my list), but hey, now it's one less to do.

games-mud/tkmoo/tkmoo-0.3.32-r1.ebuild
mm1ke: Unfortunately it has no icon. I've checked the homepage for something but didn't find anything there either.

ionenwks: Just use `applications-games` then, better than using an icon that doesn't exist or making something up from nothing.

games-mud/tkmoo/tkmoo-0.3.32-r1.ebuild
ionenwks: On a side-note, I meant to say only the 0 while keeping :=, however upon closer look this doesn't link with the libraries so the := really isn't needed and this is fine :)

games-mud/tkmoo/tkmoo-0.3.32-r1.ebuild
```diff
@@ -1,44 +1,48 @@
...
-inherit eutils
+EAPI=8
 
-MY_PN=${PN/moo/MOO-light}
-MY_P=${P/moo/MOO-light}
+inherit desktop
+
+MY_PN="${PN/moo/MOO-light}"
+MY_P="${P/moo/MOO-light}"
 DESCRIPTION="MOO Client written in Tcl/Tk"
```
ionenwks, line 10: Tiny nitpick but I prefer a blank line after variables before description.

Also, I find the substitution kind of silly, and done twice for nothing, e.g. /could/
```
MY_PN="tkMOO-light"
MY_P="${MY_PN}-${PV}"
```

However see my next comment, I think `MY_P="tkMOO-light-${PV}"` is all we need here.

games-mud/tkmoo/tkmoo-0.3.32-r1.ebuild
```diff
@@ -1,44 +1,48 @@
...
 )
 
 src_compile() {
-	emake \
-		WISH="$(type -P wish)" \
-		TKMOO_LIB_DIR="/usr/$(get_libdir)/${MY_PN}" \
-		TKMOO_BIN_DIR=/usr/bin
+	local myemakeargs=(
+		WISH="$(type -P wish)"
+		TKMOO_LIB_DIR="${EPREFIX}/usr/$(get_libdir)/${MY_PN}"
```
ionenwks, line 33: Maybe it'd make more sense to use `${EPREFIX}/usr/share/${PN}`, this is just tcl files. Same for src_install.

No real need to use MY_PN here either, should use the package's gentoo name for its directories if possible.

mm1ke: Nice suggestion, I've changed the ebuild and got rid of `MY_PN` completely. :)

games-mud/tkmoo/tkmoo-0.3.32-r1.ebuild
mm1ke: Oh i see, good to know.

games-mud/tkmoo/tkmoo-0.3.32-r1.ebuild
ionenwks: Still keeping `$(get_libdir)`? I didn't see much point in having tcl script files in lib64, not byte-compiled or anything like that either, it's just arch-independent plain text.

Last thing worth a mention, rest looks good.

mm1ke: Completely oversaw this :D 
Fixed now.

ionenwks (approved):
Thanks, merging as-is.

PR #21587, opened 2021-07-10
Bitcoin 0.21.1 with package mask

candrews:
Bitcoin 0.21.1 with package.mask message suggested by @williamh on the gentoo-dev list at https://archives.gentoo.org/gentoo-dev/message/822d30ced6084cca00696eefce1c54ee

:crossed_fingers: addresses concerns raised in #21490 - please let me know.

CC and thank you to @luke-jr @thesamesam @Flowdalic @gyakovlev @williamh


net-p2p/bitcoin-qt/bitcoin-qt-0.21.1.ebuild
```diff
@@ -0,0 +1,206 @@
...
+	else
+		elog "You are building ${PN} from Bitcoin Core."
+		elog "For more information, see:"
+		elog "https://bitcoincore.org/en/2021/05/01/release-${PV}/"
+	fi
+	elog
+	elog "Replace By Fee policy is now always enabled by default: Your node will"
+	elog "preferentially mine and relay transactions paying the highest fee, regardless"
+	elog "of receive order. To disable RBF, set mempoolreplacement=never in bitcoin.conf"
+	if has_version "<${CATEGORY}/${PN}-0.21.1" ; then
```
luke-jr, line 91: Do we want a blank elog line here?

net-p2p/bitcoind/bitcoind-0.21.1.ebuild
```diff
@@ -0,0 +1,190 @@
...
+	>=dev-libs/univalue-1.0.4:=
+	system-leveldb? ( virtual/bitcoin-leveldb )
+	sqlite? ( >=dev-db/sqlite-3.7.17:= )
+	upnp? ( >=net-libs/miniupnpc-1.9.20150916:= )
+	berkdb? ( sys-libs/db:$(db_ver_to_slot "${DB_VER}")=[cxx] )
+	zeromq? ( net-libs/zeromq:= )
+"
+RDEPEND="${DEPEND}"
+BDEPEND="
+	>=sys-devel/autoconf-2.69
```
luke-jr, line 46: @thesamesam wanted this dropped I think

profiles/package.mask
```diff
@@ -32,6 +32,19 @@
 
 #--- END OF EXAMPLES ---
 
+# Craig Andrews <candrews@integralblue.com> (2021-07-10)
+# Masked until 2021-11-01
```
luke-jr, line 36: The unmask date isn't clear yet.
```suggestion
# Masked until at least 2021-11-12
```

profiles/package.mask
```diff
@@ -32,6 +32,19 @@
...
+# This release adds enforcement of the Taproot protocol change to the Bitcoin
+# rules, beginning in November. Protocol changes require user consent to be
+# effective, and if enforced inconsistently within the community may compromise
+# your security or others! If you do not know what you are doing, learn more
+# before November. (You must make a decision either way - simply not upgrading
+# is insecure in all scenarios.)
+# To learn more, see https://bitcointaproot.cc
+>=net-p2p/bitcoin-cli-0.21.1
```
luke-jr, line 44: No reason to mask bitcoin-cli. It's just a JSON-RPC client.

luke-jr (changes requested):
Also missing libbitcoinconsensus (also affected by Taproot change) and bitcoin-tx (not affected)

net-p2p/bitcoin-qt/bitcoin-qt-0.21.1.ebuild
candrews: IMHO, it doesn't seem necessary. The output looks fine as-is:
image

net-p2p/bitcoind/bitcoind-0.21.1.ebuild
candrews: thanks!

profiles/package.mask
candrews: Updated to simply say:
> Masked until November, 2021

profiles/package.mask
candrews: Good point

candrews:
> Also missing libbitcoinconsensus (also affected by Taproot change) and bitcoin-tx (not affected)

I've now added those :+1:

williamh:
From @luke-jr 's comment on the other pr, the mask should be `=` not `>=`. I would also make the mask the first commit in the series. Also, p.mask entries should include your gentoo.org email.

candrews:
> From @luke-jr 's comment on the other pr, the mask should be `=` not `>=`.

If the mask is an `=` then bitcoin 0.21.2 or even 0.21.1-r1 is released, then the mask wouldn't cover them and the user wouldn't have opt in, which is the whole issue at hand... right? (sorry, I can't find the relevant comment on the other PR to see the reasoning behind the `=` idea)

> Also, p.mask entries should include your gentoo.org email.

Ugh, stupid mistake... Thank you for pointing it out.

luke-jr:
It's possible that 0.21.2 would revert Taproot activation, but only if the community rejects Taproot. *Most likely*, 0.21.2 will retain Taproot because everyone supports it. IMO, that's good enough reason to use `>=`. Either way, if the prediction is wrong, it can be revised when 0.21.2 is added.

candrews:
> It's possible that 0.21.2 would revert Taproot activation, but only if the community rejects Taproot. _Most likely_, 0.21.2 will retain Taproot because everyone (presumably) supports it. IMO, that's good enough reason to use `>=`. Either way, if the prediction is wrong, it can be revised when 0.21.2 is added.

Great points, I completely agree. If taproot or another consensus change occurs in the future, we can adjust the package.mask appropriately. So I'll leave the package.mask as `>=` - we can change it later should the situation change.

candrews:
I'd like to merge this tomorrow, around 12 hours from now.
If there are any other suggestions, please let me know - again, thank you all for the collaboration that brought us to this point.

ionenwks:
> If the mask is an `=` then bitcoin 0.21.2 or even 0.21.1-r1 is released [...]

You can use `~cat/pkg-ver` for profile masks as well (to ignore revision). I feel = should be avoided unless it's really something specific to a particular revision. Not that this says much for 0.21.2.

PR #21588, opened 2021-07-10
sys-devel/gcc: add USE=cet to metadata.xml && toolchain.eclass

xmhd:
As per title - add USE=cet to gcc via toolchain.eclass.

Optionally fcf-protection (CET) can be enabled by default in gcc via a patch similar to stack-clash-protection and link_now for USE=hardened, see https://pastebin.com/raw/q9HFsLuc

I've been using this in production for months now with no issues to note. Ubuntu are using the same/similar patch for their gcc.

Signed-off-by: Dave Hughes <davidhughes205@gmail.com>

@zorry @thesamesam 

edit: This is amd64 only, so will require masking by default and then unmasked on amd64.

### DIFF
--- a/eclass/toolchain.eclass
+++ b/eclass/toolchain.eclass
@@ -188,6 +188,7 @@ if [[ ${PN} != "kgcc64" && ${PN} != gcc-* ]] ; then
 		IUSE+=" systemtap" TC_FEATURES+=(systemtap)
 	tc_version_is_at_least 9.0 && IUSE+=" d"
 	tc_version_is_at_least 9.1 && IUSE+=" lto"
+	tc_version_is_at_least 10 && IUSE+=" cet"
 	tc_version_is_at_least 10 && IUSE+=" zstd" TC_FEATURES+=(zstd)
 	tc_version_is_at_least 11 && IUSE+=" valgrind" TC_FEATURES+=(valgrind)
 	tc_version_is_at_least 11 && IUSE+=" custom-cflags"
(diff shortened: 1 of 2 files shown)


zorry:
The patch for gcc need to be like the ssp/pie check/enable

trofi:
Currently we mask `cet` use flag for other packages:

```
profiles/base/package.use.mask:sys-devel/binutils cet
profiles/base/package.use.mask:sys-devel/binutils-hppa64 cet
profiles/base/package.use.mask:sys-devel/gdb cet
profiles/base/package.use.mask:sys-libs/binutils-libs cet
profiles/base/package.use.mask:sys-libs/glibc cet
```

Can you mask it for gcc for now as well?

thesamesam:
I should be able to test this now I have a CPU capable of CET.

Inlining the patch (pastebin) so it doesn't get lost:
```
diff -Naur a/gcc/gcc.c b/gcc/gcc.c
--- a/gcc/gcc.c	2021-05-11 16:24:51.478873303 -0400
+++ b/gcc/gcc.c	2021-05-11 16:31:15.722859325 -0400
@@ -880,6 +880,12 @@
 #define NO_SSP_SPEC ""
 #endif
 
+#ifdef EXTRA_OPTIONS
+#define CF_PROTECTION_SPEC "%{!fcf-protection*:%{!fno-cf-protection:-fcf-protection}}"
+#else
+#define CF_PROTECTION_SPEC ""
+#endif
+
 #ifndef LINK_SSP_SPEC
 #ifdef TARGET_LIBC_PROVIDES_SSP
 #define LINK_SSP_SPEC "%{fstack-protector|fstack-protector-all" \
```

I don't see a need to worry about polishing the patch yet until we've seen if we want to even pursue it (test it on hardened profiles for a while).

PR #21591, opened 2021-07-11
sys-boot/woeusb: bump to 5.1.2

scardracs:
This new version drops the GUI interface, so I will maintain
both old version (v3) and new (v5+).


sys-boot/woeusb/woeusb-5.1.2.ebuild
```diff
@@ -0,0 +1,32 @@
...
+KEYWORDS="~amd64 ~x86"
+RESTRICT="test" # No test suite
+
+RDEPEND="
+	app-arch/wimlib
+	sys-apps/util-linux
+	sys-block/parted
+	sys-fs/dosfstools
+	sys-fs/ntfs3g
+	sys-boot/grub:2[grub_platforms_pc]
```
juippis, line 23: Please sort in alphabetical order.

sys-boot/woeusb/woeusb-5.1.2.ebuild
```diff
@@ -0,0 +1,32 @@
...
+"
+DEPEND="${RDEPEND}"
+
+S="${WORKDIR}/WoeUSB-${PV}"
+
+src_prepare() {
+	default
+	find . -type f -print0 | xargs -0 sed -i "s/@@WOEUSB_VERSION@@/${PV}/" || die
```
juippis, line 31: `find -exec sed`

sys-boot/woeusb/woeusb-5.1.2.ebuild
```diff
@@ -0,0 +1,32 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
juippis, line 4: 8 should already work here.

app-arch/wimlib/wimlib-1.13.4.ebuild
```diff
@@ -0,0 +1,65 @@
...
+"
+RDEPEND="
+	dev-libs/libxml2:2
+	ntfs? ( sys-fs/ntfs3g )
+	fuse? ( sys-fs/fuse:0 )
+	iso? (
+		app-arch/cabextract
+		app-cdr/cdrtools
+	)
+	openssl? ( dev-libs/openssl:0= )
```
juippis, line 35: There's no other issues here as I see, but sorting these would be nice too while at it :)

app-arch/wimlib/wimlib-1.13.4.ebuild
```diff
@@ -0,0 +1,65 @@
...
+inherit autotools pax-utils
+
+DESCRIPTION="The open source Windows Imaging (WIM) library"
+HOMEPAGE="https://wimlib.net"
+SRC_URI="https://wimlib.net/downloads/${P}.tar.gz"
+
+KEYWORDS="~amd64 ~x86"
+LICENSE="|| ( GPL-3+ LGPL-3+ ) CC0-1.0"
+SLOT="0"
+IUSE="cpu_flags_x86_ssse3 fuse iso ntfs openssl test threads yasm"
```
juippis, line 15: Failed to mention but we basically use `ssl` to enable SSL/TLS.

app-arch/wimlib/wimlib-1.13.4.ebuild
```diff
@@ -0,0 +1,65 @@
...
+REQUIRED_USE="cpu_flags_x86_ssse3? ( !openssl )"
```
juippis, line 16: This is rather annoying. Could you construct something in `src_configure` where the `ssse3` gets disabled if `ssl` is enabled? And print some `elog` to users.

We try to avoid `REQUIRED_USE` as much as possible.

juippis (approved):

PR #21593, opened 2021-07-11
dev-libs/rocr-runtime: fix for gentoo prefix

littlewu2508:
The current ebuild cannot find bitcode during cmake configuration in
gentoo prefix. This commit provide a fix.

version >= 4 are modified, while 3.x are older versions which remain
original states.

Closes: https://bugs.gentoo.org/801562


heroxbd:
Is this a duplication of https://github.com/gentoo/gentoo/pull/21593 ?

PR #21603, opened 2021-07-11
net-analyzer/iplog: EAPI8 revbump, ebuild improvements

mm1ke:

### DIFF
--- a/net-analyzer/iplog/iplog-2.2.3-r3.ebuild
+++ b/net-analyzer/iplog/iplog-2.2.3-r3.ebuild
@@ -1,0 +1,31 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit flag-o-matic toolchain-funcs
+
+DESCRIPTION="TCP/IP traffic logger"
+HOMEPAGE="http://ojnk.sourceforge.net/"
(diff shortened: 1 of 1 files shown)


net-analyzer/iplog/iplog-2.2.3-r3.ebuild
```diff
@@ -0,0 +1,30 @@
...
+
+DEPEND="net-libs/libpcap"
+RDEPEND="${DEPEND}"
+
+PATCHES=( "${FILESDIR}"/${PV}-DLT_LINUX_SSL.patch )
+
+DOCS=( AUTHORS NEWS README TODO example-iplog.conf )
+
+src_compile() {
+	emake CC="$(tc-getCC)" CFLAGS="${CFLAGS} -D_REENTRANT" all
```
thesamesam, line 24: Defines like this (-DFOO) belong in CPPFLAGS if the build system respects it. Can you try append-cppflags? You may have to pass in CPPFLAGS like you are now for CFLAGS.

thesamesam (changes requested):

net-analyzer/iplog/iplog-2.2.3-r3.ebuild
mm1ke: Updated. As you suggested i had to pass `CPPFLAGS` so that it works.

PR #21606, opened 2021-07-12
OpenMW Lua

DarthGandalf:


dev-lua/sol2/sol2-3.2.2.ebuild
```diff
@@ -0,0 +1,86 @@
...
+)
+
+src_prepare() {
+	sed -i -e 's/-Werror//' \
+		tests/runtime_tests/CMakeLists.txt \
+		tests/regression_tests/1011/CMakeLists.txt \
+		tests/config_tests/function_pointers/CMakeLists.txt \
+		examples/customization/CMakeLists.txt \
+		examples/CMakeLists.txt || die
+	grep -rn Werror && die "Werror is not removed from somewhere"
```
ionenwks, line 46: That's some nasty Werror proliferation.

If it's in that many files and it's safe to sed them all, may be worth just using:
`find . -name CMakeLists.txt -exec sed -i 's/-Werror//' {} + || die`
It won't miss new ones either so the grep won't be particularly useful.

Downside of find is that it won't tell us if no files were found, but I'd like to think we'll notice if it migrated away from cmake.

games-engines/openmw/files/openmw-0.48.0-lua.patch
```diff
@@ -0,0 +1,20 @@
+Use the lua version selected by the lua-single eclass
+
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -400,14 +400,11 @@ if(USE_LUAJIT)
+    set(LUA_INCLUDE_DIR ${LuaJit_INCLUDE_DIR})
+    set(LUA_LIBRARIES ${LuaJit_LIBRARIES})
+ else(USE_LUAJIT)
+-   find_package(Lua REQUIRED)
++   find_package(Lua ${GENTOO_LUA_VERSION} EXACT REQUIRED)
```
ionenwks, line 10: Can this bit be avoided? There's the chewi method first introduced in [slade](https://github.com/gentoo/gentoo/blob/eb34fab783d361ce64883e7bafc7a261dfadbb63/games-util/slade/slade-3.2.0_beta2.ebuild#L68) and I copied it minus the temporary variable in [meandmyshadow](https://github.com/gentoo/gentoo/blob/master/games-puzzle/meandmyshadow/meandmyshadow-0.5a.ebuild#L40)

Haven't looked closely, but it's okay if it makes things simpler considering seem still need a patch for the bit below either way. It may be easier to maintain if upstream shifts things around though.

games-engines/openmw/openmw-9999.ebuild
```diff
@@ -19,13 +20,15 @@ fi
 
 LICENSE="GPL-3 MIT BitstreamVera ZLIB"
 SLOT="0"
+
 IUSE="doc devtools +osg-fork test +qt5"
```
ionenwks, line 24: Same here, we don't usually do this.

ionenwks (commented):
Just a quick ebuild look, not super familiar with lua ebuilds so I'll have a closer look later.

dev-lua/sol2/sol2-3.2.2.ebuild
```diff
@@ -0,0 +1,86 @@
...
+
+DESCRIPTION="C++ <-> Lua API wrapper"
+HOMEPAGE="https://github.com/ThePhD/sol2"
+SRC_URI="https://github.com/ThePhD/sol2/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+IUSE="test"
```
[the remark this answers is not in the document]
DarthGandalf: Previously I was told I have too long blocks without lines. So now I randomly put several.
Removed.

dev-lua/sol2/sol2-3.2.2.ebuild
DarthGandalf: They already removed all these Werrors since the release. So for the next release this sed should become much better (removed also, together with grep)

games-engines/openmw/files/openmw-0.48.0-lua.patch
DarthGandalf: Used that method.

dev-lua/sol2/sol2-3.2.2.ebuild
ionenwks: When have that knowledge, it's nice to leave note in the ebuild in case of an event where someone else did the bump.

DarthGandalf: I actually tried to package the latest git snapshot :)
They also renamed the CMakeLists.txt directories, so this sed line is failing there already. That's hard to miss while doing the bump.

ionenwks: Why keep the grep in this current ebuild then, the only way it made sense to me is "detect changes on bump".

DarthGandalf: True. Fixed.

dev-lua/sol2/sol2-3.2.2.ebuild
```diff
@@ -0,0 +1,85 @@
...
+"
+BDEPEND="
+	test? (
+		virtual/pkgconfig
+	)
+"
+
+PATCHES=(
+	"${FILESDIR}"/${P}-luajit-pkgconf.patch
+	"${FILESDIR}"/${P}-catch-depend.patch
```
ionenwks, line 35: Was just about to give this a properly look+test, but this patch is missing (I'll have another look after it's there, assuming it's needed).

DarthGandalf: Oops. Added.

dev-lua/sol2/sol2-3.2.2.ebuild
```diff
@@ -0,0 +1,85 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="test"
+REQUIRED_USE="${LUA_REQUIRED_USE}"
+RESTRICT="!test? ( test )"
+
+RDEPEND="${LUA_DEPS}"
+DEPEND="
+	test? (
+		dev-cpp/catch:1
```
ionenwks, line 23: Should this really be catch:1?

regardless of gcc11 or 10, I get this:
```
/tmp/portage/dev-lua/sol2-3.2.2/work/sol2-3.2.2/tests/runtime_tests/source/utility.cpp: At global scope:
/tmp/portage/dev-lua/sol2-3.2.2/work/sol2-3.2.2/tests/runtime_tests/source/utility.cpp:126:24: error: expected constructor, destructor, or type conversion before '(' token
  126 | TEMPLATE_LIST_TEST_CASE("utility/optional-conversion", "test that regular optional will properly catch certain types", OptionalsToTest) {
      |                        ^
```

But solved if I replace this by catch:0 (aka catch2), and adjust the include dir to catch2.

dev-lua/sol2/sol2-3.2.2.ebuild
```diff
@@ -0,0 +1,85 @@
...
+			echo "Skipping test due to https://github.com/ThePhD/sol2/issues/1221"
+		else
+			cmake_src_test
+		fi
+	}
+	lua_foreach_impl test_wrapper
+}
+
+src_install() {
+	lua_foreach_impl cmake_src_install
```
ionenwks, line 79: One more thing...

Haven't tried but is foreach needed here? Considering this doesn't actually install any lua files, it's C++ headers.

It's actually making the whole LUA_TARGETS thing a bit awkward outside of tests. If anything I feel like lua targets (required_use, lua deps, etc..) could be made optional behind `test?`. Not that I really explored the idea.

ionenwks: Perhaps the category is even wrong, dev-cpp may make more sense.

games-engines/openmw/openmw-9999.ebuild
```diff
@@ -46,6 +48,9 @@ RDEPEND="
 		dev-qt/qtopengl:5
 		dev-qt/qtwidgets:5
 	)
+	$(lua_gen_cond_dep '
+		dev-lua/sol2[${LUA_USEDEP}]
+	')
```
ionenwks, line 53: Given last comment, made me check here again. Being a header package, this doesn't need RDEPEND, and likely doesn't need `${LUA_USEDEP}` either if make this test-only.

OpenMW itself is looking into which lua to use and depend on, some single-target c++ headers don't mean much in this.

dev-lua/sol2/sol2-3.2.2.ebuild
DarthGandalf: It works for me with both. But they used version 1, so I did the same. Updated.

dev-lua/sol2/sol2-3.2.2.ebuild
DarthGandalf: > Haven't tried but is foreach needed here?

cmake_src_install doesn't work by itself, because the `_build` directory is empty - cmake was never run there.

> could be made optional behind test?

That would require something like...
```
if test; then
    lua_foreach_impl cmake_src_install
else
    cmake_src_install
fi
```
which looks very strange

I would really like review from marecki@. Same for "Perhaps the category is even wrong"

ionenwks: At this point I can't see this as a dev-lua ebuild at all, it's a dev-cpp headers package with lua-based tests.

ionenwks: I'd just be about separating the two, this include compile-time so use test forearch should avoidable.

ionenwks (approved):

PR #21608, opened 2021-07-12
sys-apps/hwloc: cuda useflag now required for nvml

Alessandro-Barbieri:
Closes: https://bugs.gentoo.org/792477

### DIFF
--- a/sys-apps/hwloc/hwloc-2.3.0.ebuild
+++ b/sys-apps/hwloc/hwloc-2.3.0.ebuild
@@ -21,17 +21,20 @@ IUSE="cairo +cpuid cuda debug gl netloc nvml +pci static-libs svg udev xml X"
 
 RDEPEND=">=sys-libs/ncurses-5.9-r3:0[${MULTILIB_USEDEP}]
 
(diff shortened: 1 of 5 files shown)


sys-apps/hwloc/hwloc-2.3.0.ebuild
```diff
@@ -40,6 +40,7 @@ BDEPEND="
 	virtual/pkgconfig
 "
 
+REQUIRED_USE="nvml? ( cuda )"
```
ionenwks, line 43: I'm not familiar with this package, but I don't really see why the nvml flag needs to exist if it requires cuda, seems redundant. cuda also ensures the drivers are installed, although /if/ it directly links with nvidia-drivers libraries (haven't checked) it may need a := on drivers for safety.

thesamesam:
CI failure needs fixing?

Alessandro-Barbieri:
Ping

thesamesam:
> Ping

You haven’t replied to ionen?

sys-apps/hwloc/hwloc-2.3.0.ebuild
Alessandro-Barbieri: In a future, cuda may work on amd hardware

PR #21610, opened 2021-07-12
distutils-r1.eclass: check that EPYTHON is set

Flowdalic:

### DIFF
--- a/eclass/distutils-r1.eclass
+++ b/eclass/distutils-r1.eclass
@@ -460,7 +460,7 @@ distutils_enable_tests() {
 esetup.py() {
 	debug-print-function ${FUNCNAME} "${@}"
 
-	[[ -n ${EPYTHON} ]] || die "EPYTHON unset, invalid call context"
+	_python_check_EPYTHON
 
 	[[ ${BUILD_DIR} ]] && _distutils-r1_create_setup_cfg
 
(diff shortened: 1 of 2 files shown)


eclass/distutils-r1.eclass
```diff
@@ -820,6 +820,10 @@ distutils-r1_python_test() {
 		die "${FUNCNAME} can be only used after calling distutils_enable_tests"
 	fi
 
+	if [[ -n "${PYTHONPATH}" && "${PYTHONPATH@a}" != *x* ]]; then
```
mgorny, line 823: Could you try checking if `EPYTHON` is set instead? That's the primary variable used to indicate we're in ebuild env.

Flowdalic: good point. done

eclass/distutils-r1.eclass
```diff
@@ -820,6 +820,10 @@ distutils-r1_python_test() {
 		die "${FUNCNAME} can be only used after calling distutils_enable_tests"
 	fi
 
+	if [[ -z ${EPYTHON} ]]; then
```
mgorny, line 823: There are some other python* eclass functions that have similar check, please sync it to use the same syntax.

eclass/distutils-r1.eclass
mgorny: Cool. Could you cover other similar functions too? (i.e. `distutils-r1_python_compile` and so on)

eclass/distutils-r1.eclass
Flowdalic: Yep, I've saw those, but I decided to mimic the `if not-set; then die` pattern that is used in the function. Which is also the pattern I prefer. But I can change it back to `[[ -n ${EPYTHON} ]] || die "EPYTHON unset, invalid call context"`.

PR #21612, opened 2021-07-12
media-tv/plex-media-server version bump to 1.23.4.4805

ephemer0l:
Signed-off-by: Scott Martin <om@organizedmagnetism.com>


media-tv/plex-media-server/plex-media-server-1.23.4.4805.ebuild
```diff
@@ -0,0 +1,76 @@
...
+)
+
+src_install() {
+	# Remove Debian specific files
+	rm -r "usr/share/doc" || die
+
+	# Remove shipped openssl library
+	if use system-openssl; then
+		rm usr/lib/plexmediaserver/lib/libssl.so.1.1 || die
+	fi
```
juippis, line 45: With this, I couldn't start plex-media-server (at least with start_pms command) - are you sure this is supposed to work?
```
# start_pms 
Error relocating /usr/lib64/libssl.so.1.1: ENGINE_load_ssl_client_cert: symbol not found
Error relocating /usr/lib64/libssl.so.1.1: ENGINE_finish: symbol not found
Error relocating /usr/lib64/libssl.so.1.1: ENGINE_init: symbol not found
Error relocating /usr/lib64/libssl.so.1.1: ENGINE_get_ssl_client_cert_function: symbol not found
```
Works if I install this with `-system-openssl`.

ionenwks: Hm, I see libcrypto.so.1.1 is not being unbundled too, I thought it could be related (i.e. mismatch) but it also doesn't like the system's either:
```
Error relocating /usr/lib64/libcrypto.so.1.1: makecontext: symbol not found
Error relocating /usr/lib64/libcrypto.so.1.1: setcontext: symbol not found
```
As much as I dislike it, unsure if this IUSE can be supported at the moment.

Did this work before? I guess it's a good thing it's not enabled by default anyway.

ephemer0l: It did work with older version. It hasn't for the last two or three, though. I have removed the USE, and added a settings override file '/etc/default/plexmediaserver' so a user can keep changes through updates. Thanks for your time looking this over.

media-tv/plex-media-server/plex-media-server-1.23.4.4805.ebuild
```diff
@@ -0,0 +1,71 @@
...
+
+LICENSE="Plex"
+SLOT="0"
+KEYWORDS="-* ~amd64 ~x86"
+RESTRICT="mirror bindist"
+
+DEPEND="
+	acct-group/plex
+	acct-user/plex"
+
```
ionenwks, line 26: Seem to have lost `RDEPEND` entirely by removing openssl, still need `RDEPEND="${DEPEND}"` for those acct-* at runtime.

On a side-note, maybe add a remove old commit too? I don't think need 4 versions at once, keeping 2 should be enough.

ephemer0l: Fixed, thanks :-)

PR #21619, opened 2021-07-12
dev-cpp/asio: version 1.19.1

DarthGandalf:


dev-cpp/asio/asio-1.19.1.ebuild
```diff
@@ -0,0 +1,51 @@
...
+
+src_install() {
+	use doc && local HTML_DOCS=( doc/. )
+	default
+
+	if use examples; then
+		# Get rid of the object files
+		emake clean
+		dodoc -r src/examples
+		docompress -x /usr/share/doc/${PF}/examples
```
juippis, line 49: From a quick look, this installs stuff worth less than 2 MB. Since nothing gets built and no extra deps pulled, I'd say you can just install them unconditionally.

juippis (approved):

dev-cpp/asio/asio-1.19.1.ebuild
ionenwks: Never been fond of IUSE=examples either, and typically remove it unless it's over the top -- I see it as just more documentation without dependencies, and we always install those if no deps normally.

If a user really doesn't want them they can INSTALL_MASK /usr/share/doc/*/examples, also get removed by nodoc if don't want those either.

PR #21622, opened 2021-07-13
mail-filter/bogofilter: removed wrong license

scardracs:

### DIFF
--- a/mail-filter/bogofilter/bogofilter-1.2.5.ebuild
+++ b/mail-filter/bogofilter/bogofilter-1.2.5.ebuild
@@ -9,7 +9,7 @@ DESCRIPTION="Bayesian spam filter designed with fast algorithms, and tuned for s
 HOMEPAGE="http://bogofilter.sourceforge.net/"
 SRC_URI="mirror://sourceforge/${PN}/${P}.tar.xz"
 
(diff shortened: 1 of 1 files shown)


mail-filter/bogofilter/bogofilter-1.2.5-r1.ebuild
```diff
@@ -0,0 +1,129 @@
...
+
+EAPI=7
+
+inherit autotools db-use flag-o-matic toolchain-funcs
+
+DESCRIPTION="Bayesian spam filter designed with fast algorithms, and tuned for speed"
+HOMEPAGE="http://bogofilter.sourceforge.net/"
+SRC_URI="mirror://sourceforge/${PN}/${P}.tar.xz"
+
+LICENSE="GPL-3+ CC-BY-ND-3.0"
```
thesamesam: This still has the wrong licence? Anyway, I’d just git mv the bad ebuild and fix it in one commit?

thesamesam (changes requested):

juippis (commented):
I don't think a revbump is required for this change...

scardracs:
> I don't think a revbump is required for this change...

Done

juippis (approved):
Thanks!

PR #21623, opened 2021-07-13
dev-util/tup: upgrade to 0.7.10

pinicarus:
Build of 0.7.8 fails since `fuse` pkgconfig is now called `fuse3`.

Signed-off-by: Pierre-Nicolas Clauss <pinicarus@protonmail.com>

Closes: https://bugs.gentoo.org/704990


pinicarus:
I set the keyword to unstable amd64 to satisfy repoman but all my local tests proved the package quite stable

dev-util/tup/tup-0.7.10.ebuild
```diff
@@ -0,0 +1,103 @@
...
+	inherit git-r3
+else
+	SRC_URI="https://github.com/gittup/tup/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64"
+fi
+
+DEPEND="
+	dev-db/sqlite:=
+	dev-libs/libpcre:=
+	sys-fs/fuse:=
```
juippis, line 30: Does this work with both fuse's slots? You may need to restritct it with `:0=` or `:3=`.

dev-util/tup/tup-0.7.10.ebuild
```diff
@@ -0,0 +1,103 @@
...
+
+CONFIG_CHECK="~FUSE_FS ~NAMESPACES"
+WARNING_FUSE_FS="CONFIG_FUSE_FS is required for tup to work"
+WARNING_NAMESPACES="CONFIG_NAMESPACES is required for tup to work as intended (workaround: set TUP_NO_NAMESPACING env var when running tup)"
+
+src_prepare() {
+	# Use our toolchain
+	sed -i Tuprules.tup \
+		-e "s:CC = gcc:CC = $(tc-getCC) ${CFLAGS} ${LDFLAGS}:" \
+		-e "s:ar crs:$(tc-getAR) crs:"
```
juippis, line 42: You shouldn't use `:` as a delimiter when modifying `CFLAGS` with `sed`, 
https://bugs.gentoo.org/685160.

Also add `|| die` at the end of this command, 
https://devmanual.gentoo.org/ebuild-writing/error-handling/index.html

dev-util/tup/tup-0.7.10.ebuild
```diff
@@ -0,0 +1,103 @@
...
+
+	if [[ ${PV} != 9999 ]]; then
+		# Avoid invoking `git` to find version, use ours
+		sed -i src/tup/link.sh \
+			-e 's;`git describe`;v'"${PV};"
```
juippis, line 47: Here too.

dev-util/tup/tup-0.7.10.ebuild
```diff
@@ -0,0 +1,103 @@
...
+	fi
+
+	default
+}
+
+src_configure() {
+	echo "CONFIG_TUP_USE_SYSTEM_SQLITE=y" >> tup.config
```
juippis, line 54: You can put this in `src_prepapre`?

dev-util/tup/tup-0.7.10.ebuild
```diff
@@ -0,0 +1,103 @@
...
+	addwrite /dev/fuse
+
+	# Disabling namespacing because it leads to accessing /proc/<pid>/setgroups
+	# which violates sandboxing.
+	export TUP_NO_NAMESPACING=1
+
+	# Skip tests which require namespacing or root privileges.
+	pushd test || die
+	rm -v ./*full-deps*.sh
+	SKIPPED_TESTS=(
```
juippis, line 84: Strictly speaking we name local variables with small case lettering, but this doesn't bother me too much... However is this necessary? You can just call ` rm`on this list.

juippis (commented):
All that being said, have you tried running the test phase? It seems to hang for me, 30 minutes and nothing is happening.

dev-util/tup/tup-0.7.10.ebuild
pinicarus: Done

dev-util/tup/tup-0.7.10.ebuild
pinicarus: You're right, it needs to be restricted to `:3=`

dev-util/tup/tup-0.7.10.ebuild
pinicarus: Done

dev-util/tup/tup-0.7.10.ebuild
pinicarus: Done

dev-util/tup/tup-0.7.10.ebuild
pinicarus: Done

pinicarus:
> All that being said, have you tried running the test phase? It seems to hang for me, 30 minutes and nothing is happening.

You're right.
Locally it works fine but some more tests seem not to work within the emerge sandbox.

I see some errors during testing : `ERROR: ld.so: object 'libsandbox.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.`. I tried to re-emerge `sys-apps/sandbox` but it didn't help.

dev-util/tup/tup-0.7.10.ebuild
```diff
@@ -0,0 +1,102 @@
...
+src_prepare() {
+	# Use our toolchain
+	sed -i Tuprules.tup \
+		-e "s|CC = gcc|CC = $(tc-getCC) ${CFLAGS} ${LDFLAGS}|" \
+		-e "s|ar crs|$(tc-getAR) crs|" || die
+
+	if [[ ${PV} != 9999 ]]; then
+		# Avoid invoking `git` to find version, use ours
+		sed -i src/tup/link.sh \
+			-e 's|`git describe`|v'"${PV}|" || die
```
juippis, line 47: This doesn't seem to do anything.

juippis (approved):
All works, thanks!

PR #21624, opened 2021-07-13
net-analyzer/{net-snmp,bmon,traceroute-nanog, ...etc}: ~riscv keyword request

alexfanqi:


profiles/arch/riscv/package.use.mask
```diff
@@ -1,6 +1,10 @@
 # Copyright 2019-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Alex Fan <alexfanqi@yahoo.com> (2021-07-13)
+# dependencies not keyworded/ use flag not tested
+net-analyzer/net-snmp lm-sensors rpm smux
+
```
dlan17, line 7: strictly, I'd prefer p.use.mask one package first, then keyword it
so this commit should go before keywording net-analyzer/net-snmp

ZeroChaos-:
I lack the ability to test any of this, but if the arch team is okay with it then so am I.

profiles/arch/riscv/package.use.mask
alexfanqi: Thanks! I have reordered them

PR #21631, opened 2021-07-14
net-misc/sipcalc: move stable keywords, update DESCRIPTION 

mm1ke:
@thesamesam 
I've moved the stable keywords from the -r0 to the -r1. I know this is not way to go but since there is basically no difference between both ebuilds, there is clearly no reason to keep them both.
Actually my idea was to open a stable request for the -r1 at some point and get rid of the `EAPI=5` version, but you got ahead when bumping the -r0 directly to `EAPI=7` in 50d778f49f96ec5b1de41d862ea510811e9d37ab ;) (i guess since it's a really simple tool it would have been a bit of overkill to start a stable request for a simple eapi bump)

I've also updated the description, guess it fits a bit better and also doesn't include the package name.


thesamesam:
Thanks, that sounds good to me!

PR #21632, opened 2021-07-14
Fixing deprecated plugins and enforcing ayatana

antenore:
At upstream we removed the ST, NX and XDMCP plugins, Telepathy is not supported anymore since more than one year and now Ayatana (or appindicator) is a requirement.

This PR should amend these changes, please verify as I cannot test.


benkohler:
I recently changed the libappindicator dep to non-optional, but I will make the other changes you suggest

PR #21637, opened 2021-07-14
app-editors/vscode & app-editors/vscodium: version bump to 1.58.1

AdelKS:
Hello @juippis @zmedico !

A little version bump, nothing interesting to report.

Adel


juippis (approved):
In future there's no need to ping me directly, I get these notices via the proxy-maint tag. Works for me, thanks!

AdelKS:
> In future there's no need to ping me directly, I get these notices via the proxy-maint tag

Alright! Sorry for the spam x) Should I stop pinging you too @zmedico ?

PR #21645, opened 2021-07-14
profiles/features/musl/package.mask: mask net-misc/netkit-rsh, bug #715898

hlein:
Closes: https://bugs.gentoo.org/715898

### DIFF
--- a/profiles/features/musl/package.mask
+++ b/profiles/features/musl/package.mask
@@ -65,3 +65,7 @@ www-client/firefox-bin
 # Mikle Kolyada <zlogene@gentoo.org> (2020-03-20)
 #  No source builds for musl
 app-emulation/firecracker
+
+# Hank Leininger <hlein@korelogic.com> (2021-07-14)
+# Uses glibc-specific rexec(3) function, bug #715898
+net-misc/netkit-rsh


profiles/features/musl/package.mask
```diff
@@ -65,3 +65,7 @@ www-client/firefox-bin
 # Mikle Kolyada <zlogene@gentoo.org> (2020-03-20)
 #  No source builds for musl
 app-emulation/firecracker
+
+# Hank Leininger <hlein@korelogic.com> (2021-07-14)
```
thesamesam, line 69: Nit: shift to the top of the file (below the comment block, so above the most-recent entry)

hlein: Thanks! I could not tell because existing entries seemed haphazardly ordered, and my google-fu failed for finding package.mask update guidance (which doesn't come up often, after all). Noted for the future.

thesamesam: There's this [page](https://devmanual.gentoo.org/profiles/package.mask/index.html) which of course doesn't mention it yet (will soon)!

thesamesam: https://github.com/gentoo/devmanual/pull/236

PR #21647, opened 2021-07-14
net-im/profanity: Version updated to 0.11.0

jpds:
* Added dep on virtual/libcrypt.

Tested on `amd64`.


andrey-utkin:
Thanks @jpds @thesamesam

PR #21649, opened 2021-07-15
net-misc/ena-driver: Add upstream 2.5.0 version

gkelly:
- Copied from `2.2.3` ebuild.
- Tested by building with `linux-5.10.49-gentoo-r1` kernel.
- Removes `dodoc` line from previous ebuild, those files don't appear to be present.


gkelly:
Just realized that this has an `RDEPEND` on unzip, while it doesn't use a zip file. I'll take that out.

sormy:
The README file is presented but was renamed to `README.rst`

sormy:
Also, might want to update `KEYWORDS` and include `~arm64`. It is supported and buildable on arm64 as well.

AndrewAmmerlaan:

PR #21652, opened 2021-07-15
dev-util/spirv-tools: fix gentoo prefix cmake configure issue

littlewu2508:
during cmake configuration, SPIRV-Headers_SOURCE_DIR sets to /usr
causing cmake cannot find /usr/include/spirv/unified1/spirv.core.grammar.json

Closes: https://bugs.gentoo.org/802240

### DIFF
--- a/dev-util/spirv-tools/spirv-tools-2021.1.ebuild
+++ b/dev-util/spirv-tools/spirv-tools-2021.1.ebuild
@@ -34,7 +34,7 @@ BDEPEND="${PYTHON_DEPS}
 
 multilib_src_configure() {
 	local mycmakeargs=(
-		"-DSPIRV-Headers_SOURCE_DIR=/usr/"
+		"-DSPIRV-Headers_SOURCE_DIR=${ESYSROOT}/usr/"
 		"-DSPIRV_WERROR=OFF"
 		"-DSPIRV_TOOLS_BUILD_STATIC=OFF"
 		"-DBUILD_SHARED_LIBS=ON"


sarnex (approved):
Thanks, do you need me to push this?

littlewu2508:
> Thanks, do you need me to push this?

Sure. Thanks for the quick response!

dev-util/spirv-tools/spirv-tools-2021.1.ebuild
```diff
@@ -34,7 +34,7 @@ BDEPEND="${PYTHON_DEPS}
 
 multilib_src_configure() {
 	local mycmakeargs=(
-		"-DSPIRV-Headers_SOURCE_DIR=/usr/"
+		"-DSPIRV-Headers_SOURCE_DIR=${EPREFIX}/usr/"
```
thesamesam, line 37: This should ideally be ESYSROOT because that's [where](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-72002r2) DEPEND-class dependencies are satisfied:
```suggestion
		"-DSPIRV-Headers_SOURCE_DIR=${ESYSROOT}/usr/"
```

littlewu2508: Thanks for pointing out. I had no experience on cross-compiling before, but after briefly reading some documents I get a bit clearer. I think you're right about using ESYSROOT instead of EPREFIX.

littlewu2508: Should I change it on my branch or you just push the correct one to the master in a new commit?

ionenwks: Preferable to just force-push to this branch with fixes, then could use your commit as-is.

littlewu2508: Force pushing done.

PR #21664, opened 2021-07-15
dev-lang/julia: version bump to 1.6.2

anaveragehuman:
Might want to wait a bit on this since 1.6.3 and 1.7 are about to be released.


jsmolic:
Thanks for the PR, and sorry that this was unnoticed. I see that 1.6.3 was released in september. Do you know if it is compatible with llvm-12 or higher? I can have a look and test out, but ideally we'd get somebody familiar with julia to have a look because I'm not really familiar with it.

anaveragehuman:
>I see that 1.6.3 was released in september. Do you know if it is compatible with llvm-12 or higher? 

I don't remember, but upstream is often behind in adopting new LLVM versions. I usually set LLVM to the highest value it can take without the build crashing, and it is unlikely upstream fixed it in a minor version update. I will check again and update this PR with 1.6.3 and 1.7 when the latter is released.

PR #21666, opened 2021-07-15
app-emulation/virt-viwer: Add ebuild for virt-viewer-10.0.

sklv:


app-emulation/virt-viewer/virt-viewer-10.0.ebuild
```diff
@@ -0,0 +1,45 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+inherit meson xdg
```
juippis, line 5: We usually have newline in between these.

juippis (commented):
I'm getting, 
```
WARNING: Unknown options: "gtk-vnc, spice-gtk"
```
could you check and verify the source for correct meson options?

sklv:
@juippis - I've fixed the option names (listed [here](https://gitlab.com/virt-viewer/virt-viewer/-/blob/master/meson_options.txt)) and checked that toggling them works correctly.

juippis (approved):

PR #21671, opened 2021-07-16
dev-db/sqlcl-bin-21.2.0.174.2245 version bump

fedeliallalinea:
Closes: https://bugs.gentoo.org/802399


dev-db/sqlcl-bin/sqlcl-bin-21.2.0.174.2245.ebuild
```diff
@@ -0,0 +1,33 @@
...
+
+MY_PN="${PN/-bin}"
+MY_P="${MY_PN}-${PV}"
+
+inherit java-pkg-2
+
+DESCRIPTION="Oracle SQLcl is the new SQL*Plus"
+HOMEPAGE="https://www.oracle.com/database/technologies/appdev/sqlcl.html"
+SRC_URI="https://download.oracle.com/otn_software/java/sqldeveloper/${MY_P}.zip"
+RESTRICT="mirror"
```
ionenwks, line 14: Nice that fetch was dropped, but is only mirror right? No bindist is a bit unusual.

Not that I can make much sense out of this license, but looking into it still make sound it shouldn't be allowed.

ionenwks (approved):

PR #21674, opened 2021-07-16
net-im/telegram-desktop: Bump to 2.8.10

mid-kid:


mid-kid:
@gyakovlev Finished building the interesting USE flag combos. Feel free to merge whenever.

mid-kid:
I made the grave mistake of not waiting a day after the release. 2.8.10 is a single-line fix.

gyakovlev:
Hi.
testing.

I have a gut feeling that `jemalloc` should be enabled by default, and users can opt-out ( I read your patch explanation and reasoning). But if upstream defaults to it - we better to, unless it breaks something. jemalloc is not an uncommon dependency, not a big deal.
Thoughts?
If you agree, I can default-enable it in separate commit, so you don't have to re-push PR.

mid-kid:
I was basing it off of `eix -c --installed-without-use jemalloc` on my system, which currently reports:
```
[I] app-emulation/qemu (6.0.0@18/6/21): QEMU + Kernel-based Virtual Machine userland tools
[I] media-gfx/inkscape (1.0.2-r1@2/6/21): SVG based generic vector-drawing program
[I] net-libs/nghttp2 (1.41.0-r2(0/1.14)@1/7/21): HTTP/2 C Library
[I] net-misc/aria2 (1.35.0@1/3/21): A download utility with segmented downloading with BitTorrent support
```
I also didn't have jemalloc installed in the first place.

That said, none of these apps enable jemalloc by default as far as I can tell, and they provide upstream switches to control it, unlike telegram-desktop which had to be patched to work around that.

I guess it makes more sense to enable it by default, so I'll do that and adapt the patch message.

gyakovlev:
yeah you know upstream is stubborn and can be hard to reason with.
I've lost count amount of times pule/alsa/wayland flags were flipped around or removed/added.

at least you are not taking choice from users who need it.

thanks!
tested both builds, looks fine. re-added `~ppc64` keyword as well.

mid-kid:
Oops, did I remove the ppc64 keyword? I must've not noticed, sorry about that.

gyakovlev:
no problem at all. I keep an eye on it anyway and will catch it.

PR #21678, opened 2021-07-16
net-wireless/rtl88x2bu new ebuild

ephemer0l:
Signed-off-by: Scott Martin <om@organizedmagnetism.com>

Tested on amd64. Could potentially support arm arm64 and x86.


ZeroChaos-:
802.11r is broken in this version of the driver, but the entropiceffect fork which I've been using doesn't even build against 5.12.13 so I'm kind of unsure which to proceed with.  802.11r isn't in "wide" use, so I suspect "builds" is better than "has 802.11r support but doesn't build"  Lemme poke at this and I'll merge something.

ephemer0l:
Thanks for your time.

ZeroChaos-:
even the maintainer of the driver says "don't use this driver", I'm not sure we want this
https://github.com/morrownr/88x2bu/commit/cc94acc06fb7fb85ec27f4983953bef1fe35a474

ephemer0l:
Is a working driver better then none?

ephemer0l:
Updated to add to @module-rebuild set. I still cant find a way to get the ebuild to sign the kernel module.

ZeroChaos-:
> Updated to add to @module-rebuild set. I still cant find a way to get the ebuild to sign the kernel module.

this is how I do it
https://github.com/pentoo/pentoo-overlay/blob/master/profiles/pentoo/base/profile.bashrc#L26-L49

ephemer0l:
I've added it to the README. Any other critiques?

thesamesam:
Closing per Z_C's comments and issues with past maintenance.

PR #21679, opened 2021-07-16
net-dns/getdns: assign myself as maintainer

CaseOf:
Hello,
I did previously maintain net-dns/getdns package.
I gave up when they decided to drop definitely LibreSSL support.
Since that, Gentoo also decided to drop it, and I’ve started to switch some of my machines to OpenSSL.
I still keep some of them on LibreSSL to maintain the libressl overlay.
Now, I’m ready to maintain it again, if you agree with that._

### DIFF
--- a/net-dns/getdns/metadata.xml
+++ b/net-dns/getdns/metadata.xml
@@ -5,6 +5,10 @@
 		<email>blueness@gentoo.org</email>
 		<name>Anthony G. Basile</name>
 	</maintainer>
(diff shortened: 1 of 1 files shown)


thesamesam:
As I explained on the [bugs](https://bugs.gentoo.org/719868#c2) (and I think on IRC), it'd be really appreciated if you could fix issues in the PR re-adopting. This is the rule we apply in general.

I've actually gone through and cleaned up all of them now.

PR #21680, opened 2021-07-16
sys-libs/glibc: add preserve-libs for libcrypt

thesamesam:
Bug: https://bugs.gentoo.org/802207

### DIFF
--- a/sys-libs/glibc/glibc-2.33-r3.ebuild
+++ b/sys-libs/glibc/glibc-2.33-r3.ebuild
@@ -1,0 +1,1534 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8,9,10} )
+TMPFILES_OPTIONAL=1
+
+inherit python-any-r1 prefix preserve-libs toolchain-funcs flag-o-matic gnuconfig \
+	multilib systemd multiprocessing tmpfiles
+
+DESCRIPTION="GNU libc C library"
+HOMEPAGE="https://www.gnu.org/software/libc/"
+LICENSE="LGPL-2.1+ BSD HPND ISC inner-net rc PCRE"
+SLOT="2.2"
+
+EMULTILIB_PKG="true"
+
+# Gentoo patchset (ignored for live ebuilds)
+PATCH_VER=4
+PATCH_DEV=dilfridge
+
+if [[ ${PV} == 9999* ]]; then
+	inherit git-r3
+else
+	KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86"
+	SRC_URI="mirror://gnu/glibc/${P}.tar.xz"
+	SRC_URI+=" https://dev.gentoo.org/~${PATCH_DEV}/distfiles/${P}-patches-${PATCH_VER}.tar.xz"
+fi
+
+RELEASE_VER=${PV}
+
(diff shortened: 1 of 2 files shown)


thesamesam:
```
--- libxcrypt-4.4.23-r1.ebuild	2021-07-16 07:13:39.236333200 +0100
+++ libxcrypt-4.4.23-r2.ebuild	2021-07-16 23:15:00.062239332 +0100
@@ -26,7 +26,10 @@
 RESTRICT="!test? ( test )"

 DEPEND="system? (
-		elibc_glibc? ( sys-libs/glibc[-crypt(+)] )
+		elibc_glibc? (
+			sys-libs/glibc[-crypt(+)]
+			compat? ( !sys-libs/glibc[crypt(-)] )
+		)
 		!sys-libs/musl
 	)"
 RDEPEND="${DEPEND}"
```

```
--- glibc-2.33-r2.ebuild	2021-07-15 00:55:20.103656808 +0100
+++ glibc-2.33-r3.ebuild	2021-07-16 23:14:53.638944897 +0100
@@ -6,7 +6,7 @@
 PYTHON_COMPAT=( python3_{7,8,9,10} )
 TMPFILES_OPTIONAL=1

-inherit python-any-r1 prefix toolchain-funcs flag-o-matic gnuconfig \
+inherit python-any-r1 prefix preserve-libs toolchain-funcs flag-o-matic gnuconfig \
 	multilib systemd multiprocessing tmpfiles

 DESCRIPTION="GNU libc C library"
@@ -1488,6 +1488,13 @@
 		# https://bugs.gentoo.org/753740
 		rm "${EROOT}"/usr/lib/locale || die
 	fi
+
+	# Keep around libcrypt so that Perl doesn't break when merging libxcrypt
+	# (libxcrypt is the new provider for now of libcrypt.so.{1,2}).
+	# bug #802207
+	if use crypt; then
+		preserve_old_lib /$(get_libdir)/libcrypt$(get_libname 1)
+	fi
 }

 pkg_postinst() {
@@ -1517,4 +1524,7 @@
 			fi
 		done
 	fi
+
+	# No need for 'use crypt' guard here as we check if the library already exists on the system
+	preserve_old_lib_notify /$(get_libdir)/libcrypt$(get_libname 1)
 }
```

sys-libs/glibc/glibc-2.33-r3.ebuild
```diff
@@ -0,0 +1,1530 @@
...
+				ewarn "removed from glibc and is now provided by the package"
+				ewarn "  sys-auth/libnss-nis"
+				ewarn "Install it now to keep your NIS setup working."
+				ewarn ""
+			fi
+		done
+	fi
+
+	# No need for 'use crypt' guard here as we check if the library already exists on the system
+	preserve_old_lib_notify /$(get_libdir)/libcrypt$(get_libname 1)
```
thesamesam, line 1529: Oh, this is wrong, I need to keep a variable in preinst and check it here.

sys-libs/glibc/glibc-2.33-r3.ebuild
```diff
@@ -0,0 +1,1534 @@
...
+	if [[ -L ${EROOT}/usr/lib/locale ]]; then
+		# Help portage migrate this to a directory
+		# https://bugs.gentoo.org/753740
+		rm "${EROOT}"/usr/lib/locale || die
+	fi
+
+	# Keep around libcrypt so that Perl doesn't break when merging libxcrypt
+	# (libxcrypt is the new provider for now of libcrypt.so.{1,2}).
+	# bug #802207
+	if use crypt; then
```
Arfrever, line 1495: This `use crypt` is called in `pkg_preinst()` phase, so it will return result for newly installed instance of glibc, not previously installed instance.
You should use:
```
	if has_version "${CATEGORY}/${PN}[crypt]"; then
```

PR #21683, opened 2021-07-17
net-fs/samba: conditionalise dependency on dev-python/subunit

thesamesam:
Samba seems to configure + build + install fine without subunit
installed.

Revbumping to allow users to depclean subunit itself but also
a bunch of assorted Python dependencies. Also, to ensure that
users upgrading without dynamic deps enabled don't get stuck
on the "EAPI 8" trap discussed in the bug.

Closes: https://bugs.gentoo.org/801637

### DIFF
--- a/net-fs/samba/samba-4.13.10-r1.ebuild
+++ b/net-fs/samba/samba-4.13.10-r1.ebuild
@@ -68,24 +68,19 @@ COMMON_DEPEND="
 	>=net-libs/gnutls-3.4.7[${MULTILIB_USEDEP}]
 	net-libs/libnsl:=[${MULTILIB_USEDEP}]
 	sys-libs/e2fsprogs-libs[${MULTILIB_USEDEP}]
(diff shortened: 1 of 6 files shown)


thesamesam:
@Polynomial-C

net-fs/samba/samba-4.13.10-r1.ebuild
```diff
@@ -120,6 +119,7 @@ DEPEND="${COMMON_DEPEND}
 	)
 	spotlight? ( dev-libs/glib )
 	test? (
+		$(python_gen_cond_dep "dev-python/subunit[\${PYTHON_USEDEP},${MULTILIB_USEDEP}]" )
```
SoapGentoo, line 122:
```suggestion
		$(python_gen_cond_dep 'dev-python/subunit[${PYTHON_USEDEP}]' )
```

SoapGentoo (changes requested):

PR #21684, opened 2021-07-17
app-text/xapers: Python 3.9 support

WPettersson:
Also update copyright. Note that Python 3.10 support can't yet be tested as
xapers depends on xapian-bindings


WPettersson:
Forgot sign-off in first commit, fixed with force push (also rebased to newer master)

juippis (approved):
I noticed an interesting scenario with python and `src_test`, I'll make a bug and explain it clearly there.

PR #21686, opened 2021-07-17
net-proxy/mitmproxy: Bump to 7.0.0

00-matt:


net-proxy/mitmproxy/mitmproxy-7.0.0.ebuild
```diff
@@ -0,0 +1,75 @@
...
+		>=dev-python/parver-0.1[${PYTHON_USEDEP}]
+		>=dev-python/pytest-3.3[${PYTHON_USEDEP}]
+		>=dev-python/pytest-asyncio-0.10.0[${PYTHON_USEDEP}]
+		>=dev-python/requests-2.9.1[${PYTHON_USEDEP}]
+	)
+"
+
+RESTRICT="!test? ( test )"
+
+distutils_enable_tests pytest
```
ionenwks, line 62: Trivial but distutils_enable_tests handles adding "test" to IUSE, sets RESTRICT's `test? ( test )`, and adds pytest to BDEPEND. So ebuild is duplicating 3 of these things that could be removed.

See function description for other things it does:
https://devmanual.gentoo.org/eclass-reference/distutils-r1.eclass/index.html

Also, normally RESTRICT would be put straight below IUSE (or REQUIRED_USE) rather than down here to match skel.ebuild ordering. Not that this matters if removing it. distutils function call is fine being located there though.

net-proxy/mitmproxy/mitmproxy-7.0.0.ebuild
```diff
@@ -0,0 +1,75 @@
...
+	>=www-servers/tornado-4.3[${PYTHON_USEDEP}]
+	>=dev-python/urwid-2.1.1[${PYTHON_USEDEP}]
+	>=dev-python/wsproto-1.0.0[${PYTHON_USEDEP}]
+	>=dev-python/zstandard-0.11.0[${PYTHON_USEDEP}]
+	dev-libs/openssl:0
+"
+
+DEPEND="
+	${RDEPEND}
+	test? (
```
ionenwks, line 51: In most cases test? should be BDEPEND with python so it can be executed on the build machine. DEPEND would be for if needed to link with libraries, use arch-specific headers, and so on.

ionenwks (commented):
Looks fine, tests pass for me too. Maybe just small ebuild adjustments be nice.

net-proxy/mitmproxy/mitmproxy-7.0.0.ebuild
00-matt: Ah thanks for catching that.

Also I need to write a script that checks the orderering of things because I always seem to get it wrong :P

net-proxy/mitmproxy/mitmproxy-7.0.0.ebuild
ionenwks: I might add that I think DEPEND="${RDEPEND}" isn't needed, see distutils_enable_tests' function description about this again.

00-matt: That makes sense. Not sure how it ended up in DEPEND to begin with

ionenwks (approved):

PR #21687, opened 2021-07-17
Cannadic cleanup

SoapGentoo:


ulm:
Sad to see yc go, no issues with that package.

That said, ACK from GNU Emacs team for removal of app-emacs/yc.

PR #21688, opened 2021-07-17
net-im/telegram-desktop: Drop jemalloc patch

mid-kid:
Upstream clarified; The glibc allocator causes telegram to use excessive
amounts of memory, and never free it. That's a bug.

See-also: https://github.com/telegramdesktop/tdesktop/issues/16084
See-also: https://github.com/desktop-app/cmake_helpers/pull/91#issuecomment-881788003

@gyakovlev sorry for the rapid-fire succession of this PR...
What do you think? Should I keep the patch for the hardened/musl users that don't have this issue, and just add an `ewarn`?


gyakovlev:
you don't need a revbump to drop that useflag tbh.
pushed, but removed revbump.
thanks!

gyakovlev:
as for alternative to jmalloc - let's see feedback.
I don't think telegram is considered secure among users who use hardened malloc =)
did they ever update webrtc code for example? I think not and it still uses code at the time it was forked.

PR #21689, opened 2021-07-17
dev-qt/qtdeclarative: riscv: fix missing atomic library

dlan17:
this will fix the build error in ARCH=riscv, tested with gcc-11

Closes: https://bugs.gentoo.org/790689
Signed-off-by: Yixun Lan <dlan@gentoo.org>


dlan17:
@Pesa hi, could you help review and take this patch? it's a major block for the Qt keyword job for riscv, for the patch itself, it only modify for the ARCH=riscv, won't affect other ARCHs, so should be safe.

We may still work with Qt upstream to have a better solution, but may take sometime

Pesa (approved):
https://bugs.gentoo.org/790689#c6

PR #21705, opened 2021-07-18
www-servers/nginx: migrate to GLEP 81

ConiKost:
@Whissi Please check, if you are okay with changes, since this is a major used package.


Whissi:
Thank you but I plan to combine this migration with some other rewrites which are currently pending.

ConiKost:
> Thank you but I plan to combine this migration with some other rewrites which are currently pending.

Thanks for the update! Will close here and make a note on the bug ticket.

SoapGentoo:
@Whissi, given the bug has been open since late November 2019, and we're making really great progress on the GLEP 81 migration, I don't really want to hold that up on possible "pending" work.

We've had this sort of problem with e.g. `kernel-2.eclass` in the past. Ultimately, we're at the point where this should be finished, and the idea of batching with other fixes hasn't been a great approach overall in Gentoo given the migration progress until ConiKost's efforts began.

I'd like to ask you to merge this (possibly with other fixes if you wish) before 1st Aug; I'll handle it if not by then.

PR #21713, opened 2021-07-19
app-editors/featherpad: New package, import from ::guru

hansfn:


app-editors/featherpad/featherpad-0.18.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+
+inherit xdg qmake-utils
+
+MY_P="${P/featherpad/FeatherPad}"
+
+DESCRIPTION="Lightweight Qt5 Plain-Text Editor for Linux"
+HOMEPAGE="https://github.com/tsujan/FeatherPad"
+SRC_URI="https://github.com/tsujan/FeatherPad/archive/V${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3"
```
juippis, line 14: Looking at the source files, this seems to be `GPL-3+`.

app-editors/featherpad/featherpad-0.18.0.ebuild
```diff
@@ -0,0 +1,37 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit xdg qmake-utils
```
juippis, line 6: Since *qmake-utils.eclass* doesn't import any functions, you can just sort this list alphabetically and it should work.

app-editors/featherpad/featherpad-0.18.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+
+MY_P="${P/featherpad/FeatherPad}"
```
juippis, line 8: Since this variable is only used once in this ebuild, you can just plaintext it in the part where it's used. We usually define these kind of variables when they get used multiple times in an ebuild.

app-editors/featherpad/featherpad-0.18.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+RDEPEND="app-text/hunspell
+	dev-qt/qtcore:5
+	dev-qt/qtgui:5
+	dev-qt/qtnetwork:5
+	dev-qt/qtprintsupport:5
+	dev-qt/qtsvg:5
+	dev-qt/qtwidgets:5
+	dev-qt/qtx11extras:5"
+DEPEND="${RDEPEND}
+	dev-qt/linguist-tools:5"
```
juippis, line 27: I have a feeling linguist-tools should be a BDEPEND instead, so,
```suggestion
DEPEND="${RDEPEND}"
BDEPEND="dev-qt/linguist-tools:5"
```

app-editors/featherpad/featherpad-0.18.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="app-text/hunspell
+	dev-qt/qtcore:5
+	dev-qt/qtgui:5
+	dev-qt/qtnetwork:5
+	dev-qt/qtprintsupport:5
+	dev-qt/qtsvg:5
+	dev-qt/qtwidgets:5
+	dev-qt/qtx11extras:5"
```
juippis, line 25: Also add this dep:
```
+x11-libs/libX11
```

app-editors/featherpad/featherpad-0.18.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+DEPEND="${RDEPEND}
+	dev-qt/linguist-tools:5"
+
+S="${WORKDIR}/${MY_P}"
+
+src_configure() {
+	eqmake5
+}
```
juippis, line 33: I have a feeling this should be `src_compile` phase instead.

ionenwks: This one's fine in src_configure, it's doing Makefile configuration to be run by normal emake in src_compile/install

That aside, i see this package has a CMakeLists.txt (unsure if it's in a working state) -- if cmake is usable it'd be better than the qmake mess.

hansfn: Thx for checking, but qmake just compiles the Qt project file and generates the Makefile that is used during `the src_compile `phase.

app-editors/featherpad/featherpad-0.18.0.ebuild
hansfn: Will do.

app-editors/featherpad/featherpad-0.18.0.ebuild
hansfn: But then I have to maintain the version number in the ebuild file (since the unpacked tar.gz file contains the version number). I think I'll keep it. Or is there a better way to resolve this?

app-editors/featherpad/featherpad-0.18.0.ebuild
hansfn: You are 100% correct. I just looked at the COPYING file, not the actual source code.

hansfn:
I broke the ebuild when I accepted your dev-qt/linguist-tools suggestion, @juippis. Will fix shortly.

PS! Do you squashed these PRs? I'm just wondering which message format to use for update / commits to the PR branch.

app-editors/featherpad/featherpad-0.18.0.ebuild
juippis: `S="${WORKDIR}/FeatherPad-${PV}"`

app-editors/featherpad/featherpad-0.18.0.ebuild
juippis: Good to know, never wrote a qmake ebuild - it just looked like an alternative to `emake` instead of `econf` :)

juippis:
You should probably fix them locally, then squash and force-push over yeah. No idea how the Github's "merge suggestion" really works...

app-editors/featherpad/featherpad-0.18.0.ebuild
ionenwks: It's a common issue, the license file doesn't really say which it is. Good to check source files for headers either way, they take precedence over these files in case of other licenses too.

hansfn:
OK, I'm not used to force pushing, but will try to remember to do so. Have switched to CMake.

juippis (approved):
I'm happy with it, thank you both!

PR #21714, opened 2021-07-19
sys-cluster/hpx: Bump to v1.7.0

shifty91:
Bump version to v1.7.0. Changes included:

- Restrict `test`, as tests don't build
- Remove `doc`, because the doc generation segfaults
- Close bugs related to tests and compile errors in respect to gcc 11

Thanks, Kurt


juippis (approved):
You're missing some dependencies, but I can add them.

shifty91:
@juippis Thanks! I realized that `asio` is missing, but completely forgot to update the PR...

PR #21718, opened 2021-07-19
net-im/telegram-desktop: Bump to 2.8.11

mid-kid:
@gyakovlev

I've not been able to reproduce the bug this fixes, but one affected person has told me the patch works, so here goes nothing.


gyakovlev:
thanks. I'm one of those who can't reproduce, still looks fine on my system.

PR #21725, opened 2021-07-20
net-p2p/retroshare: Version bump

G10h4ck:
Fixes a bunch of bugs fixed upstream in newer version and disabling
  components that are been obsoleted upstream in older versions

Bug: https://bugs.gentoo.org/785964
Bug: https://bugs.gentoo.org/779838
Bug: https://bugs.gentoo.org/798048
Bug: https://bugs.gentoo.org/798099

Signed-off-by: Gioacchino Mazzurco <gio@altermundi.net>


net-p2p/retroshare/retroshare-0.6.5-r3.ebuild
```diff
@@ -0,0 +1,143 @@
+# Copyright 1999-2021 Gentoo Authors
```
ionenwks, line 1: I see updating the old ebuild too, is there a reason to? i.e. are you planning to stabilize -r3 over 0.6.6 or something? Or is there a problem with 0.6.6 that will make users want to keep using 0.6.5?

I don't use this, so I wouldn't know.

ionenwks (commented):
Be nice if the ebuild could receive some modernizing.

Small review for that if interested in a few suggestions (haven't really looked into building/runtime/bugs side of things).

net-p2p/retroshare/retroshare-0.6.5-r3.ebuild
G10h4ck: I plan to stabilize 0.6.6 but there might be problems with new version so I also wanted to do a minimal maintenance of 0.6.5 in case someone need to stick to that retroshare version

G10h4ck:
Thanks for suggestions @ionenwks I have updated 0.6.6 ebuild according to them, can this be merged now?

net-p2p/retroshare/retroshare-0.6.6.ebuild
```diff
@@ -0,0 +1,124 @@
...
+	if use gui; then
+		make_desktop_entry retroshare
+
+		for i in 24 48 64 128 ; do
+			doicon -s ${i} "data/${i}x${i}/apps/retroshare.png"
+		done
+	fi
+}
+
+pkg_preinst() {
```
ionenwks, line 100: No longer need to define the postinst/postrm with xdg.eclass, but it has a preinst and being overwritten here.
```suggestion
pkg_preinst() {
	xdg_pkg_preinst

```

Or else it leads to:
```
 * QA Notice: new icons were found installed but icon cache
 * has not been updated:
 *   /usr/share/icons/hicolor/64x64/apps/retroshare.png
```

net-p2p/retroshare/retroshare-0.6.6.ebuild
```diff
@@ -0,0 +1,124 @@
...
+		ewarn "You have disabled GXS database encryption, ${PN} will use SQLite"
+		ewarn "instead of SQLCipher for GXS databases."
+		ewarn "Builds using SQLite and builds using SQLCipher have incompatible"
+		ewarn "database format, so you will need to manually delete GXS"
+		ewarn "database (loosing all your GXS data and identities) when you"
+		ewarn "toggle sqlcipher USE flag."
+	fi
+
+	if [[ ${REPLACING_VERSIONS} ]]; then
+		if ver_test ${REPLACING_VERSIONS} -lt 0.5.9999; then
```
ionenwks, line 111: Could make this `-lt 0.6`

ionenwks (commented):
> Thanks for suggestions @ionenwks I have updated 0.6.6 ebuild according to them, can this be merged now?

Thanks, changes look good but still a few things.

Keeping 0.6.5-r3 is fine but I forgot to mention that it should be added in a separate commit. Version bumps are normally done in their own commit to allow easy reversal and could have a message for changes to specific to 0.6.5.

May also want to change `Bug:` tags to `Closes:`, if it's fixed it can be closed (unless you prefer to wait for stabilization to do that, up to you how to handle your bugs -- personally tend to close when fixed in ~arch unless it's affecting many users that may need to find the bug).

But more importantly, this time I tried to build and seem to get getting build issues with gcc11 for 0.6.6 with any USE combinations:
```
util/rsdir.cc:280:38: error: 'create_directories' is not a member of 'std::filesystem'
  280 |                 if(!std::filesystem::create_directories(dest_dir))
```
Builds fine with `append-cxxflags -std=c++14` (that's because gcc11 defaults to c++17), but that's more of a temporary workaround than a fix, preferable would be to patch rsdir.cc to `#include <filesystem>` and send that upstream (if haven't yet, do check the [clean patch howto](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#clean-patch-howto)).

Also, noticed a few other minor things.

G10h4ck:
Thanks for the careful review @ionenwks all your remarks should be addressed ;)
Is it ok to merge this now?

ionenwks (approved):
> Thanks for the careful review @ionenwks all your remarks should be addressed ;)
> Is it ok to merge this now?

Indeed, looks good. I see cleaned the edos2unix too, nice -- and hadn't realized you were the upstream, guess I didn't need to say anything about upstreaming patches :) Thanks for your work.

> May also want to change `Bug:` tags to `Closes:`, if it's fixed it can be closed (unless you prefer to wait for stabilization to do that, [...]

Haven't heard back of your preference on this point, assume it was overlooked considering there's no reason not to close the version bump one at least. I'll go ahead and switch to `Closes:` myself where relevant so this doesn't have to wait longer.

G10h4ck:
Thanks!

PR #21728, opened 2021-07-20
dev-util/codespell: new package v2.1.0

waebbl:
Upon request from @juippis, move the package from ::guru to ::gentoo.

Closes: https://bugs.gentoo.org/667830


waebbl:
16 tests are skipped. They need aspell-en, but even when adding aspell to BDEPEND, the tests are being skipped.

dev-util/codespell/codespell-2.1.0.ebuild
```diff
@@ -0,0 +1,36 @@
...
+DESCRIPTION="Check text files for common misspellings"
+HOMEPAGE="https://github.com/codespell-project/codespell"
+SRC_URI="https://github.com/${PN}-project/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+# Code licensed under GPL-2
+# Dictionary licensed under CC-BY-SA-3.0
+LICENSE="GPL-2 CC-BY-SA-3.0"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+RESTRICT="!test? ( test )"
```
ionenwks, line 20: `distutils_enable_tests pytest` already sets this for you, so it's not needed.
https://devmanual.gentoo.org/eclass-reference/distutils-r1.eclass/index.html

dev-util/codespell/codespell-2.1.0.ebuild
```diff
@@ -0,0 +1,36 @@
...
+EAPI=8
+
+PYTHON_COMPAT=( python3_{8..10} )
+DISTUTILS_USE_SETUPTOOLS=rdepend
+
+inherit distutils-r1
+
+DESCRIPTION="Check text files for common misspellings"
+HOMEPAGE="https://github.com/codespell-project/codespell"
+SRC_URI="https://github.com/${PN}-project/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
```
ionenwks, line 13: The first two `${PN}` could be plain texted, see:

dev-util/codespell/codespell-2.1.0.ebuild
```diff
@@ -0,0 +1,36 @@
...
+
+	# This will fail if the package itself
+	# is not yet installed
+	sed -e 's:test_command:_&:' \
+		-i codespell_lib/tests/test_basic.py || die
+
+	distutils-r1_python_prepare_all
+}
+
+distutils_enable_tests pytest
```
ionenwks, line 36: Most python ebuilds in the tree put this after *DEPEND variables rather than under `python_*()`. This is used to setup global stuff (iuse/restrict/deps) rather than being the src_test phase itself (that normally you'd sort there).

dev-util/codespell/codespell-2.1.0.ebuild
```diff
@@ -0,0 +1,36 @@
...
+python_prepare_all() {
+	# do not depend on pytest-cov
+	sed -e '/addopts/d' -i setup.cfg || die
+
+	# This will fail if the package itself
+	# is not yet installed
+	sed -e 's:test_command:_&:' \
+		-i codespell_lib/tests/test_basic.py || die
+
+	distutils-r1_python_prepare_all
```
ionenwks, line 33: Preferable to run the function that will apply patches before seds and other changes. Patches are typically expected applied against the unchanged source.

dev-util/codespell/codespell-2.1.0.ebuild
waebbl: Oh ok, so this was the reason why it wasn't in the file :)

ionenwks:
Was doing some final checks (all good), but noticed this has a man page -- wouldn't hurt to install it. The Makefile can generate it (dependency is kind of trivial and already on most systems, so unconditional sounds fine).

dev-util/codespell/codespell-2.1.0.ebuild
```diff
@@ -0,0 +1,49 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+PYTHON_COMPAT=( python3_{8..10} )
+DISTUTILS_USE_SETUPTOOLS=rdepend
```
juippis, line 7: I get
```
```

dev-util/codespell/codespell-2.1.0.ebuild
```diff
@@ -0,0 +1,49 @@
...
+python_prepare_all() {
+	distutils-r1_python_prepare_all
+
+	# do not depend on pytest-cov
+	sed -e '/addopts/d' -i setup.cfg || die
+
+	# This will fail if the package itself
+	# is not yet installed
+	sed -e 's:test_command:_&:' \
+		-i codespell_lib/tests/test_basic.py || die
```
juippis, line 37: I wonder if we could use this here:
https://dev.gentoo.org/~mgorny/python-guide/distutils.html#installing-the-package-before-testing

Basically try `distutils_enable_tests --install pytest`

dev-util/codespell/codespell-2.1.0.ebuild
ionenwks: Probably be good to mention both need `DISTUTILS_STRICT_ENTRY_POINTS=y` and "qa" in `PORTAGE_ELOG_CLASSES` to get this message.

waebbl: I was wondering, because I didn't get the warning. Didn't know about the `DISTUTILS_STRICT_ENTRY_POINTS=1` flag and only had qa in `PORTAGE_ELOG_CLASSES`.

dev-util/codespell/codespell-2.1.0.ebuild
waebbl: works like a charm

juippis (approved):
Thanks! Weirdly test case still tries to use aspell, but it doesn't work even if aspell is installed. I wonder if aspell should install some python modules...?

waebbl:
> Thanks! Weirdly test case still tries to use aspell, but it doesn't work even if aspell is installed. I wonder if aspell should install some python modules...?

> 16 tests are skipped. They need aspell-en, but even when adding aspell to BDEPEND, the tests are being skipped.

I already noticed this. I haven't checked the aspell ebuild. Currently it doesn't have any python support, but I'm not aware if it could have it.

Thanks for merging it. I'm gonna drop the ::guru package.

PR #21732, opened 2021-07-20
app-text/xapers: Use python_test() for testing

WPettersson:
Changing src_test() to python_test() so the correct interpreter is used
for tests, instead of just using the latest selected python interpreter.
Also addresses QA notice by adding DISTUTILS_USE_SETUPTOOLS=rdepend

Closes: https://bugs.gentoo.org/803023

### DIFF
--- a/app-text/xapers/xapers-0.9.0.ebuild
+++ b/app-text/xapers/xapers-0.9.0.ebuild
@@ -18,6 +18,8 @@ LICENSE="GPL-3"
 SLOT="0"
 KEYWORDS="~amd64"
 
+DISTUTILS_USE_SETUPTOOLS=rdepend
+
 RDEPEND="app-text/poppler[utils]
 	dev-libs/xapian-bindings[python,${PYTHON_USEDEP}]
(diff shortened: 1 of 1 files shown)


app-text/xapers/xapers-0.9.0.ebuild
```diff
@@ -18,6 +18,8 @@ LICENSE="GPL-3"
 SLOT="0"
 KEYWORDS="~amd64"
 
+DISTUTILS_USE_SETUPTOOLS=rdepend
```
ionenwks, line 21: This is a '@PRE-INHERIT` variable, you can see a mention if it in [eclass docs](https://devmanual.gentoo.org/eclass-reference/distutils-r1.eclass/index.html) or by running pkgcheck:
```
MisplacedEclassVar: version 0.9.0: invalid pre-inherit placement, line 21: 'DISTUTILS_USE_SETUPTOOLS=rdepend'
```

...**however**, if you look at the top, you'll see it's already set. It may be good to re-read what the QA notice said about this.

juippis (approved):

PR #21733, opened 2021-07-20
eclass/vim-plugin.eclass: delete if has_version condition.

monsieurp:
delete if has_version condition and replace it with a test against the REPLACING_VERSIONS variable. See https://projects.gentoo.org/pms/8/pms.html#x1-10900011.1.

Courtesy of Arfrever.


CyberTailor (approved):
It works.

PR #21739, opened 2021-07-21
net-misc/r8168: added USE flag

Karlson2k:
* Added USE flag to enable firmware loading support
* Demoted kernel config check from error to warning

This is an alternative version of PR #21507 without module default flags.

### DIFF
--- a/net-misc/r8168/metadata.xml
+++ b/net-misc/r8168/metadata.xml
@@ -8,4 +8,7 @@
 	<longdescription>Official Realtek r8168 linux driver. The following cards are currently supported:
 RTL8111B RTL8168B RTL8111 RTL8168 RTL8111C RTL8111CP RTL8111D(L) RTL8168C
 RTL8111DP RTL8111E RTL8168E RTL8111F RTL8411</longdescription>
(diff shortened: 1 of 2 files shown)


pacho2:
What is the drawback of enabling firmware loading support always?

Karlson2k:
> What is the drawback of enabling firmware loading support always?

1. It is not enabled by the upstream, so they don't consider it as a "stable".
2. Realtek NICs sometimes works unpredictable. There is no way to disable firmware load during run-time, so disabling firmware loading may help someone.

I think we should follow the upstream and leave this flag by default as "disabled" until it will be enabled by the upstream.

Karlson2k:
> What is the drawback of enabling firmware loading support always?

@pacho2 I sporadically experience system hung on resume with firmware enabled and no problems with firmware disabled.

pacho2:
That would need to be forwarded to upstream I think to get it handled in future releases I hope :/

Karlson2k:
@pacho2 Sure, that would be nice, but unfortunately Realtek doesn't have any official repo to report.

PR #21742, opened 2021-07-22
Drop media-libs/mesa's IUSE=egl and IUSE=gbm

mattst88:
Neither USE flag adds a dependency and both are small (the E in EGL stands for "Embedded"!)

I'd prefer to avoid revbumps for all of these packages to add `(+)` to their `RDEPEND`s, so this is really just a heads-up for maintainers: Please add `(+)` to `media-libs/mesa[egl]` and `media-libs/mesa[gbm]` in your packages on the next revbump.

When you do, please leave a comment here and I'll drop the patch to your package.

### DIFF
--- a/media-libs/mesa/mesa-9999.ebuild
+++ b/media-libs/mesa/mesa-9999.ebuild
@@ -33,18 +33,16 @@ for card in ${VIDEO_CARDS}; do
 done
 
 IUSE="${IUSE_VIDEO_CARDS}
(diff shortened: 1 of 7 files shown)


chewi (approved):
The games stuff looks fine to me.

If others are wondering, EGL shouldn't be confused with GLES. The latter doesn't have additional dependencies either but is probably a bit heavier.

mattst88:
@Pesa

Pesa (approved):

a17r:
Current ~arch dev-qt/qtwebengine already gets rid of the dep entirely and is pending stabilisation.

PR #21744, opened 2021-07-22
dev-util/mh: initial import

oz123:


dev-util/mh/files/Makefile.patch
```diff
@@ -0,0 +1,21 @@
+diff --git a/Makefile b/Makefile
+index 01023ca..f87ab56 100644
+--- a/Makefile
++++ b/Makefile
+@@ -14,6 +14,7 @@ endif
+ 
+ DEBUG = -g -fsanitize=address
+ CFLAGS = -Wall -lpcre2-8 #$(DEBUG)
```
SoapGentoo, line 8: 1. wrong, `-lpcre2-8` should be in `LIBS`
2. what about user `CFLAGS`?

dev-util/mh/files/Makefile.patch
```diff
@@ -0,0 +1,21 @@
...
++LDFLAGS ?=
```
SoapGentoo, line 9: omit this definition here, adds no value

dev-util/mh/files/Makefile.patch
```diff
@@ -0,0 +1,21 @@
...
+ OPTS = -D COLOROUTPUT
+ PREFIX = /usr/local
+ MANPATH = /man/man
+@@ -27,7 +28,7 @@ clean:
+ 	rm -fv $(PROGNAME)
+ 
+ mh:  ## compile this software
+-	gcc -o $(PROGNAME) -D VERSION=\"$(VERSION)\" $(OBJECTS) $(CFLAGS) $(OPTS)
++	gcc -o $(PROGNAME) -D VERSION=\"$(VERSION)\" $(OBJECTS) $(CFLAGS) $(LDFLAGS) $(OPTS)
```
SoapGentoo, line 18: what if I want to use clang?

dev-util/mh/mh-1.0.1.ebuild
```diff
@@ -0,0 +1,30 @@
...
+
+EAPI=7
+
+DESCRIPTION="Generate help for Makefile"
+HOMEPAGE="https://github.com/oz123/mh"
+SRC_URI="https://github.com/oz123/mh/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz"
+LICENSE="BSD"
+SLOT="0"
+IUSE="test"
+KEYWORDS="~amd64"
```
SoapGentoo, line 12: metadata block needs cleaning up

dev-util/mh/mh-1.0.1.ebuild
```diff
@@ -0,0 +1,30 @@
...
+
+RESTRICT="!test? ( test )"
+
+RDEPEND="dev-libs/libpcre2"
```
SoapGentoo, line 16: so you're linking `libpcre2`, but not pulling it in a build time?

SoapGentoo (changes requested):

dev-util/mh/mh-1.0.1.ebuild
```diff
@@ -0,0 +1,30 @@
...
+
+BDEPEND="test? ( dev-util/cmocka )"
+
+PATCHES="${FILESDIR}/Makefile.patch"
+
+src_compile() {
+	emake mh VERSION="${PV}"
+}
+
+src_install() {
```
thesamesam, line 26: We can’t use your own install target?

is this because we need to fix the Makefile?

oz123: Yes, I think so. I can fix the makefile upstream and try again.
@SoapGentoo @thesamesam thanks for the review.

dev-util/mh/mh-1.0.1.ebuild
oz123: Can you be specific? What should be removed?

dev-util/mh/mh-1.0.1.ebuild
oz123: My bad. Will be fixed.

dev-util/mh/mh-1.0.1.ebuild
oz123: @thesamesam using my own target, the binaries get installed to `/usr/local` ...

oz123: This is now fixed. I had to use PREFIX and DESTDIR (with some changes to the Makefile too).

dev-util/mh/files/Makefile.patch
oz123: Removed.

dev-util/mh/files/Makefile.patch
oz123: Fixed in the Makefile in version 1.0.3.

dev-util/mh/files/Makefile.patch
oz123: Fixed, now the makefile uses $(CC) instead of  gcc.

dev-util/mh/mh-1.0.1.ebuild
SoapGentoo: @oz123 I mistook this for another request to add a package to `::gentoo`. In general, we don't accept new packages to `::gentoo` anymore that are purely proxy maintained, for a number of reasons. Why not add it to `::guru`?

oz123: I can do that. But I don't contribute the guru project. That means I will need to go throuh the process of gaining access before this gets useful.
Quite a bummer. Out of scope here. But the. Direct engagement with Gentoo developers is really cool. Having another layer of abstraction is perhaps reducing the load on Gentoo developers but it's making it harder to recruit new developers.

oz123:
added to guru.

PR #21747, opened 2021-07-22
games-engines/openmw: collada support

DarthGandalf:


dev-games/openscenegraph-openmw/files/openscenegraph-3.6.5-use_boost_asio.patch
```diff
@@ -0,0 +1,274 @@
+diff -ruN OpenSceneGraph-OpenSceneGraph-3.6.5/CMakeLists.txt osg-my/CMakeLists.txt
+--- OpenSceneGraph-OpenSceneGraph-3.6.5/CMakeLists.txt	2020-01-31 12:03:07.000000000 +0100
++++ osg-my/CMakeLists.txt	2020-09-26 18:45:41.191099844 +0200
```
SoapGentoo, line 3: please clean up these patches

dev-games/openscenegraph-openmw/openscenegraph-openmw-3.6_p20210610.ebuild
```diff
@@ -0,0 +1,160 @@
...
+		x11-libs/cairo
+	)
+	tiff? ( media-libs/tiff:0 )
+	truetype? ( media-libs/freetype:2 )
+	vnc? ( net-libs/libvncserver )
+	xrandr? ( x11-libs/libXrandr )
+	zlib? ( sys-libs/zlib )
+"
+DEPEND="${RDEPEND}
+	>=dev-libs/boost-1.37.0:*
```
SoapGentoo, line 80: not relevant anymore? (the lower bound and slot specifier)

SoapGentoo (changes requested):

dev-games/openscenegraph-openmw/files/openscenegraph-3.6.5-use_boost_asio.patch
```diff
@@ -0,0 +1,274 @@
...
+@@ -764,7 +764,7 @@
+     FIND_PACKAGE(DirectInput)
+     FIND_PACKAGE(NVTT)
+     IF (NOT WIN32)
+-        FIND_PACKAGE(Asio)
++        FIND_PACKAGE(Boost)
+     ENDIF()
+     FIND_PACKAGE(ZeroConf)
+ 
+diff -ruN OpenSceneGraph-OpenSceneGraph-3.6.5/CMakeModules/FindAsio.cmake osg-my/CMakeModules/FindAsio.cmake
```
thesamesam, line 13: Do we need all this diff slime?

dev-games/openscenegraph-openmw/files/openscenegraph-3.6.5-use_boost_asio.patch
```diff
@@ -0,0 +1,274 @@
...
+--- OpenSceneGraph-OpenSceneGraph-3.6.5/CMakeModules/FindAsio.cmake	2020-01-31 12:03:07.000000000 +0100
```
thesamesam, line 14: Not using the same base.

dev-games/openscenegraph-openmw/openscenegraph-openmw-3.6_p20210610.ebuild
thesamesam, line 80: :*? And such an old lower bound?

dev-games/openscenegraph-openmw/openscenegraph-openmw-3.6_p20210610.ebuild
```diff
@@ -0,0 +1,160 @@
...
+	)
+	if use examples; then
+		mycmakeargs+=(
+			$(cmake_use_find_package fltk FLTK)
+			$(cmake_use_find_package fox FOX)
+			$(cmake_use_find_package glut GLUT)
+			$(cmake_use_find_package wxwidgets wxWidgets)
+		)
+	fi
+	if use lua; then
```
thesamesam, line 148:
```suggestion

	if use lua; then
```

dev-games/openscenegraph-openmw/openscenegraph-openmw-3.6_p20210610.ebuild
```diff
@@ -0,0 +1,160 @@
...
+		$(cmake_use_find_package sdl2 SDL2)
+		$(cmake_use_find_package svg RSVG)
+		$(cmake_use_find_package tiff TIFF)
+		$(cmake_use_find_package truetype Freetype)
+		$(cmake_use_find_package vnc LibVNCServer)
+		-DOSGVIEWER_USE_XRANDR=$(usex xrandr)
+		$(cmake_use_find_package zlib ZLIB)
+		-DOSG_USE_LOCAL_LUA_SOURCE=OFF
+	)
+	if use examples; then
```
thesamesam, line 140:
```suggestion

	if use examples; then
```

games-engines/openmw/openmw-9999.ebuild
```diff
@@ -38,8 +38,8 @@ RDEPEND="${LUA_DEPS}
 	media-video/ffmpeg:=
 	>=sci-physics/bullet-2.86:=[double-precision]
 	virtual/opengl
-	osg-fork? ( dev-games/openscenegraph-openmw:=[ffmpeg,jpeg,png,sdl,svg,truetype,zlib] )
-	!osg-fork? ( >=dev-games/openscenegraph-3.5.5:=[ffmpeg,jpeg,png,sdl,svg,truetype,zlib] )
+	osg-fork? ( >=dev-games/openscenegraph-openmw-3.6:=[collada(-),ffmpeg,jpeg,png,sdl,svg,truetype,zlib] )
```
thesamesam, line 41: Can you fix the append-cxxflags below to use ESYSROOT and probably cppflags instead?

thesamesam (changes requested):

dev-games/openscenegraph-openmw/files/openscenegraph-3.6.5-use_boost_asio.patch
DarthGandalf: This was copied as is from openscenegraph ebuild.

I'll clean up there first in a separate commit.

games-engines/openmw/openmw-9999.ebuild
DarthGandalf: Yep. For now I skipped updating 0.96 to avoid revbump and rebuild for a rare case; will remove that ebuild soon after 0.97 is released anyway

mattst88:
Thanks for including the mesa/egl bits!

dev-games/openscenegraph-openmw/metadata.xml
```diff
@@ -16,10 +16,16 @@
 	<longdescription lang="en">
 		OpenSceneGraph on steroids. This fork was created to experiment with performance enhancements for OpenMW that are too controversial to be included in the general purpose OSG project - either for design reasons or backward compatibility reasons. OpenMW still works with "vanilla" OpenSceneGraph, but it is recommended to use the fork for better performance (around 10-15% on most devices). Note that the improvements are very specific to OpenMW and its animation code in particular; other OpenSceneGraph applications are not likely to see the same speed-up, or any speed-up at all if they are already GPU limited.
 	</longdescription>
+	<upstream>
+		<remote-id type="github">OpenMW/osg</remote-id>
+	</upstream>
```
ionenwks, line 21: fwiw non-openmw openscenegraph put this below `</use>`. That's also what I usually do but it doesn't really matter, just saying given seem you want to keep those in sync where possible.

dev-games/openscenegraph-openmw/openscenegraph-openmw-3.6_p20210610.ebuild
```diff
@@ -0,0 +1,162 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+LUA_COMPAT=( lua5-{1,2} )
```
ionenwks, line 6: lua5-2 is dead, so you can remove it like with sol2. No need to update old ebuilds but on a bump it's nice to be up to date.

I do wonder if 3 or 4 works, but CMakeLists.txt does indicate it wants 5-2 or 5-1, so it may not be safe.

dev-games/openscenegraph-openmw/openscenegraph-openmw-3.6_p20210610.ebuild
```diff
@@ -0,0 +1,162 @@
...
+
+WX_GTK_VER="3.0-gtk3"
+inherit cmake flag-o-matic lua-single wxwidgets
+
+DESCRIPTION="OpenMW-specific fork of OpenSceneGraph"
+HOMEPAGE="https://github.com/OpenMW/osg"
+MY_COMMIT="b02abe200c4847e73b887b064a89ea1758a5b733"
```
ionenwks, line 13: Just style but so variables-to-check/edit-on-bump aren't buried, I usually do MY_ stuff like:
```
inherit ...

MY_COMMIT=
MY_P=
...

DESCRIPTION=
```
Still not all the way at the top so it's not pre-inherit pollution.

dev-games/openscenegraph-openmw/openscenegraph-openmw-3.6_p20210610.ebuild
```diff
@@ -0,0 +1,162 @@
...
+SRC_URI="https://github.com/OpenMW/osg/archive/${MY_COMMIT}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/osg-${MY_COMMIT}"
+
+LICENSE="wxWinLL-3 LGPL-2.1"
+SLOT="0/162" # NOTE: CHECK WHEN BUMPING! Subslot is SOVERSION
+KEYWORDS="~amd64 ~x86"
+IUSE="collada curl dicom debug doc egl examples ffmpeg fltk fox gdal
+gif glut gstreamer jpeg las lua openexr openinventor osgapps pdf png
+sdl sdl2 svg tiff truetype vnc wxwidgets xrandr +zlib"
```
ionenwks, line 22: Kind of unusual so see this like this, you can indent like other variables with tabs.

dev-games/openscenegraph-openmw/openscenegraph-openmw-3.6_p20210610.ebuild
```diff
@@ -0,0 +1,162 @@
...
+DEPEND="${RDEPEND}
+	dev-libs/boost
+	x11-base/xorg-proto
+"
+
+PATCHES=(
+	"${FILESDIR}"/openscenegraph-3.6.3-cmake.patch
+	"${FILESDIR}"/openscenegraph-3.6.3-docdir.patch
+	"${FILESDIR}"/openscenegraph-3.6.5-use_boost_asio.patch
+	"${FILESDIR}"/openscenegraph-3.6.5-cmake_lua_version.patch
```
ionenwks, line 88: Could optionally migrate away from this patch using same method used in openmw-9999, not that it's as important given it's not moving -9999 target.

As chewi mentioned, these patches were done before the -D method was noticed.

dev-games/openscenegraph/openscenegraph-3.6.5-r101.ebuild
```diff
@@ -0,0 +1,162 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+LUA_COMPAT=( lua5-{1,2} )
+
+MY_PN="OpenSceneGraph"
+MY_P=${MY_PN}-${PV}
```
ionenwks, line 9: See what I said about MY_COMMIT wrt pre-inherit pollution, usually would only have variables marked as `@PRE_INHERIT` in eclasses there (like LUA_COMPAT and WX_GTK_VER). Not that it's a problem per-se with MY_ namespace, but well.

Also most things I said about the other ebuild apply here, so won't repeat everything.

ionenwks (commented):
Just quick ebuild look, will run build/runtime tests later but assuming nothing comes up with that I'm not seeing real non-style issues.

dev-games/openscenegraph-openmw/openscenegraph-openmw-3.6_p20210610.ebuild
DarthGandalf: The patch looks simple enough, and they are using a custom FindLua52 module anyway, instead of FindLua

ionenwks (approved):
Didn't think of that during review, so I'll amend revbump number on merge and update old ebuilds due to `egl(+)`. This should be done for all ebuilds at once rather than just latest, still with a revbump due to rdepend change.

i.e.
openscenegraph-3.6.5-r100 -> r101 (straight-to-stable with `egl(+)`)
openscenegraph-3.6.5-r110 (collada support in ~arch from this PR)

Otherwise would have to wait for new revision to be stable and old one removed to cleanup mesa's egl. Going -r110 given collada-dom isn't stable yet and don't know how long this will take for all arches, so +10 is a precaution to have a space (guess could've went -r200 but meh, I feel 10 is enough in most cases).

Anyhow looks good, I'll just do some more build checks (bit slow) and get this in soon'ish.

PR #21749, opened 2021-07-23
media-libs/libsdl: Add 1.2.52 and 9999 targeting libsdl12-compat

lanodan:
This allows to use `media-libs/libsdl` in a more modern environment, mainly wayland.


chewi:
We are interested in libsdl-compat but I don't think this is the way we want to use it. I'll see what the others think.

lanodan:
Yeah, I guess it could also be done like `dev-libs/openssl-compat` or with a slot, but unlike openssl that would require to create an eselect module for the `libSDL.so` symlinks.

chewi:
I'm not a fan of eselect, as a general rule. It can be problematic at build time. In practise, you'd either want just one or the other, or you'd only want libsdl-compat for certain games. In the latter case, manually switching with eselect would be tedious, even if it's just at runtime. There are several options here.

* It could be a separate package that blocks the old one.
  * The two packages could form a virtual package.
  * All the revdeps could explicitly allow either with `|| ( media-lib/libsdl-compat media-lib/libsdl )`.
* It could just replace the old package entirely, saving us from having to fix all the revdeps.
* I was going to say it could be a separate package that installs to a different directory, allowing `LD_PRELOAD`, but then how would you build against it? Modifying `sdl.pc` and `sdl-config` to add `-I` and `-L` flags would cover some bases but not all.

Replacing the old package is by far the simplest option but it really depends on how well it works. I haven't had any issues with it but I've only briefly tried it.

chewi:
Another idea I had was that either package could be the default that you build against if you set a USE flag such as `default-libsdl` and each package would block the other using this flag. For example:

### media-libs/libsdl-compat
```
default-libsdl? ( !media-libs/libsdl[default-sdl(+)] )
```

### media-libs/libsdl
```
default-libsdl? ( !media-libs/libsdl-compat[default-sdl] )
```

Admittedly this is a bit weird though and we'd still need to update all the revdeps.

ionenwks:
I'd rather it be it's own package so it's clear what it is, libsdl12-compat package (or libsdl-compat, albeit upstreams calls it libsdl12).

And my impression is that shouldn't rush with integration in rdeps before it's even released and we know how much it breaks,

For now, maybe take a similar route as zlib-ng, i.e. it added the separate zlib-ng package to the tree but anyone that want to try it "now" needs to use package.provided given rdeps aren't updated. This could be useful to test and decide on route to take.

chewi:
I'm not counting on it being released. It's been around for years already.

ionenwks:
Also, does switching between sdl12-compat and sdl1 require a rebuild? Deps would need to be handled differently if it does.

chewi:
A rebuild isn't necessary. It's designed to have the same ABI. One game I tested it with was Psychonauts.

ionenwks:
I see, that's good then.

For long term, if at all possible, I'd like to see libsdl removed from the tree in favor of libsdl12. So I'd rather we don't put too many workarounds just to have both at once, could simply avoid updating the rdep when it's known broken with sdl12-compat until it's not (users can revert to real libsdl if they really want these).

On a different note, wonder how want the optfeature-like fake USE on the short and long term. If we look at say.. pygame2, it has a dependency on sdl2-image but consumer ebuilds need to depend on sdl2-image[png,gif] separately if they use those images, otherwise about only bmp works.

A) have libsdl12-compat handle them bit like in this current PR
B) have the virtual handle them, e.g.
```RDEPEND="
	|| (
		libsdl[flag?,flag?] (on top because probably shouldn't be default until more tested)
		(
			libsdl12-compat
			libsdl2[flag?,flag?]
		)

```

A is simple, but it's also "wrong" by causing rebuild of sdl12-compat when it's only libsdl2 that need rebuild on flag change. 

In case of B, if we remove libsdl and the virtual, ebuilds will need to depend on libsdl2 themselves for extra flags like with pygame2.

Also, I take it sdl12-compat always provide things like SDL_Joystick? Guess we'll only know if it _need_ joystick at runtime rather than build time when building with libsdl12-compat.

lanodan:
I've personally tested a lot of the games we already have in tree against sdl12-compat and it worked quite well. <https://hacktivis.me/git/blog/commit/73fcf9bb73b2cdc4a72187fab87ecee6980ad4c0.html>

To me if there is a split to be done it should be with slots to avoid tweaking many ebuilds, specially ones which aren't into `::gentoo` or `::guru`, as probably still typical for games, and their maintainers likely aren't going to notice the change.
At least I don't think there is a value in entirely splitting the ebuild.

> I'm not counting on it being released. It's been around for years already.

I'm not sure what was said before about it (nothing?) but <https://github.com/libsdl-org/sdl12-compat/issues/53> says it should be released soon and as seen in <https://github.com/libsdl-org/sdl12-compat/issues/78> they have been packaging+testing it into Fedora.

chewi:
I'd like to move ahead with this now.

I've been convinced that sdl12-compat works well enough now that it can simply be a new version of the existing package. I see that it reports itself as 1.2.50 so we can make use of that. There still hasn't been a release yet, and I don't know what version scheme a release would have, but I propose we go with 1.2.50_p20211010 for now.

We can keep the original around for a while, in case any issues crop up. Hopefully they can be fixed relatively quickly.

Of @ionenwks' two suggestions, I much prefer A. I understand that it's wrong, but rebuilds are a small price to pay to avoid some heavy complexity. sdl12-compat builds very quickly, after all.

I think all the USE flags that don't also apply to SDL2 can be dropped. I understand they were added for compatibility with existing ebuilds, but the ebuilds in ::gentoo don't reference any of these. It's unlikely any overlays do either.

I don't agree with allowing `sdl12-compat[X]` to depend on `libsdl2[wayland]`. I suspect that almost all the existing `libsdl[X]` dependencies (there aren't very many) are wrong and should be `libsdl[video]` instead. If X11 truly is required then swapping in Wayland here won't work. Running these under Wayland may work, but it would still require the X11 driver, along with XWayland.

I'll start putting something together based on this.

ionenwks:
> I much prefer A

Sounds fine to me.

> suspect that almost all the existing `libsdl[X]` dependencies (there aren't very many) are wrong

Yeah, it's typically something I been cleaning/replacing when I see it (as I've been going through old EAPI5-6 ebuilds, but still a lot I haven't checked) unless the package is already using X11 libraries/headers directly as well implying it probably can't work without xwayland and so the [X] made enough sense.

> I'll start putting something together based on this.

Looking forward to it, that would need to start using it sooner or later was inevitable. Haven't given it that much thought on how set it up wrt old comments and don't necessarily agree with myself -- so do what you think is best.

chewi:
Just want to say I haven't forgotten about this. There's something I need to do first and it should happen quite soon.

fpemud:
Any new progress?

hartwork (changes requested):
Hi everyone, happy new year! :tada:

From a look at libsdl12-compat sources and Debian packaging, I found that:
- Out of the box, libsdl and libsdl12-compat share 99% of the file locations, e.g. both provide `/usr/bin/sdl-config` and `/usr/lib64/libSDL-1.2.so`.
- pkg-config file `sdl12_compat.pc` makes it possible to target library "sdl12_compat" explicitly while also satisfying requests to plain "sdl" via its line `Provides: sdl = 1.2.61`.
- [Debian packaging](https://packages.debian.org/source/sid/sdl12-compat) forbids installation of both packages at the same time.

I would personally consider putting libsdl12-compat content in `media-libs/libsdl` too much  risk, and I suppose @chewi had the same feeling [above](https://github.com/gentoo/gentoo/pull/21749#issuecomment-885444222).

I believe the best way forward would be a new package `media-libs/libsdl12-compat` (with a blocker on `media-libs/libsdl`) for a start, and then a new virtual `virtual/libsdl12` or `virtual/libsdl` in round two.  I can jump in to help as needed, but I first wanted to make sure I'm not stepping on anyone's toes.

What do you think about this way forward?

Thanks and best, Sebastian

chewi:
You missed what I said later.

> I've been convinced that sdl12-compat works well enough now that it can simply be a new version of the existing package.

I'd rather keep it simple. The only reason I haven't done this already is I was trying to kill off certain USE flags like `alsa`, but there were one or two packages that needed to be fixed up first. One was Serious Sam, which now has an SDL2 port, but it needs more work, and I got sidetracked with that. I shouldn't have let that get in the way, sorry. I'll dig up what I have and try to turn this around by next week.

hartwork:
@chewi you're right, I was too quick to reply, only read some things after replying. My understanding is that if you put sdl12-comput in `media-libs/libsdl`, `games-fps/serious-sam-tse[alsa]` will continue to work with the latest ebuild of `media-libs/libsdl` that still has `IUSE="[..] alsa [..]"` around, so probably not a blocker. Same for the `abi_*` family. Let me know if you need help.

chewi:
I've now pushed the 1.2.60 release. I've kept the `alsa` USE flag for the time being. We can always drop it later.

I tried all the SDL1-based packages I already have installed. Only xrick doesn't work properly. I'll file an upstream bug about that tomorrow.

hartwork:
@chewi thank you! :+1:  [xrick looks like fun](http://www.bigorno.net/xrick/shots.html). Amazing how much more FPS I get with sdl12-compat over true SDL 1. On a non-recent i5 ThinkPad X220, e.g. my toy SDL code demo https://github.com/hartwork/sdl_video_demo goes from …

```console
# ./sdl1_video_demo 
[SDL1]  21 FPS at 1920x1022x32
[SDL1]  20 FPS at 1920x1022x32
[SDL1]  18 FPS at 1920x1022x32
…
```

… to …

```console
# ./sdl1_video_demo 
[SDL1]  56 FPS at 1920x1022x32
[SDL1]  54 FPS at 1920x1022x32
[SDL1]  56 FPS at 1920x1022x32
```
… for free!

PR #21750, opened 2021-07-23
riscv: keyword some dev-qt packages

alexfanqi:


dlan17:
@Pesa do you have any comment on this, if this is a complete package list or anything we missing. or say, we could definitely work as progress to bring in more packages..

btw, we currently aim to bring lxqt as target WM, and I did a basic test which works fine, @alexfanqi may give more tests on this.

PR #21754, opened 2021-07-23
app-arch/7zip: new package

Karlson2k:
[7-zip](https://www.7-zip.org/) project finally released sources with Linux/macOS support.
Compared to [p7zip project](https://sourceforge.net/projects/p7zip/files/p7zip/), which hasn't been updated for 5 years, the latest version of 7-zip offers better performance and some bugs fixed.
Despite being such outdated, p7zip sources are still downloaded 4k-6k times per week.

The new .ebuild make full use of user's CC, CXX, CFLAGS, CXXFLAGS, and LDFLAGS, with full support of upstream-defined warning control flags, depending on used compiler type and version.

As upstream publish [sources packed as .7z only](https://sourceforge.net/p/sevenzip/discussion/45797/thread/7fe6c21efa/), Debian sources mirror was used instead to avoid circular dependency.

Comments, suggestions are welcome.


Karlson2k:
Commented out suggestion of myself as proxied maintainer as it breaks QA checks.

Karlson2k:
@prometheanfire Could you check it, please?

Karlson2k:
Added testing, similar to debian testing of 7zip package.

thesamesam:
Ah, I missed this. I've imported `app-arch/7zip` from ::guru. Feel free to send PRs to that. Sorry & thanks.

PR #21761, opened 2021-07-24
net-libs/pjproject: Version 2.12.1 (security release).

jkroonza:
Whilst I submitted version yesterday back-porting the security fixes,
and official release has now been made available, bump to it instead.

Bug: https://bugs.gentoo.org/803614


ionenwks (commented):
I can't seem to be able build asterisk[pjproject] with this version (2.10 is fine)
```
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: cannot find -lpjsua2
make[1]: *** [/tmp/portage/net-misc/asterisk-18.5.1/work/asterisk-18.5.1/Makefile.rules:193: chan_pjsip.so] Error 1
```
And indeed there's no libpjsua2.so installed with 2.11, but it's there with 2.10.

Also (not a new issue, this could wait for another time):
```
 * /usr/lib64/libg7221codec.so.2
 * /usr/lib64/libilbccodec.so.2
 * /usr/lib64/libwebrtc.so.2
 * /usr/lib64/libyuv.so.2
```

jkroonza:
> I can't seem to be able build asterisk[pjproject] with this version (2.10 is fine)
> 
> ```
> /usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: cannot find -lpjsua2
> collect2: error: ld returned 1 exit status
> make[1]: *** [/tmp/portage/net-misc/asterisk-18.5.1/work/asterisk-18.5.1/Makefile.rules:193: chan_pjsip.so] Error 1
> ```
> 
> And indeed there's no libpjsua2.so installed with 2.11, but it's there with 2.10.
> 
> Also (not a new issue, this could wait for another time):
> 
> ```
>  * QA Notice: Files built without respecting LDFLAGS have been detected
>  *  Please include the following list of files in your report:
>  * /usr/lib64/libg7221codec.so.2
>  * /usr/lib64/libilbccodec.so.2
>  * /usr/lib64/libwebrtc.so.2
>  * /usr/lib64/libyuv.so.2
> ```

Upstream parallel build issue I've not been able to track down.  Reference https://bugs.gentoo.org/765799

I'm not convinced just using -j1 is any such case is the right fix as it's dependent on the way in which make resolves dependencies (tomorrow it could decide to build in a different order then the issue is back).

Two (three) options:

1.  I can force non-parallel build (emake -j1);
2.  This being a security bump we ignore the issue for now; or
3.  I need help tracking the missing dependency issue (I really dislike these "distibuted Makefile" type makefiles using recursive builds to achieve things ... in my experience it causes more problems than it solves).

ionenwks:
> Upstream parallel build issue I've not been able to track down. Reference https://bugs.gentoo.org/765799

The log snipplet I shown was from asterisk with USE=pjproject, not pjproject itself. pjproject built fine, but one library is missing causing asterisk to fail.

ionenwks:
i.e.
```
$ qlist -e =pjproject-2.10-r2 | grep 'libpjsua'
/usr/lib64/libpjsua2.so.2
/usr/lib64/libpjsua2.so
/usr/lib64/libpjsua.so.2
/usr/lib64/libpjsua.so

$ qlist -e =pjproject-2.11.1 | grep 'libpjsua'
/usr/lib64/libpjsua.so.2
/usr/lib64/libpjsua.so
```
Where did libpjsua2.so go? If it breaks its only rdep there's probably not much sense in merging this as-is.

jkroonza:
I'm going to take this to draft, doesn't look like asterisk will currently compile with 2.11, I did however back-port the asterisk-affecting security patch to 2.10-r2 in 93f6d97e4bd66daa168e1790f8cb3b8086854bd1. (After an initial round on -r1 which we should probably at this point remove)

jkroonza:
Known problem:

``` * QA Notice: Files built without respecting LDFLAGS have been detected
 * /usr/lib64/libg7221codec.so.2
 * /usr/lib64/libilbccodec.so.2
 * /usr/lib64/libresample.so.2
 * /usr/lib64/libwebrtc.so.2
 * /usr/lib64/libyuv.so.2
```

Previously these were all just silent failure not working at all, so in my opinion still an improvement, need some sleep now though.

PR #21763, opened 2021-07-24
media-tv/plex media server

ephemer0l:
media-tv/plex media server version bump, and removal of one old version.


ionenwks (approved):
I'll fix on merge, but do double-check commits. Can use a simple "git log --stat" to see changed files.

Currently the version bump commit is only changing the manifest, and the remove old is doing the version bump.

Also still missing the `:` i.e.
`media-tv/plex-media-server version...`
should be
`media-tv/plex-media-server: version ...`

PR #21768, opened 2021-07-24
net-p2p/go-ipfs-bin: remove unused file

mm1ke:


kakra:
This package is scheduled for removal anyways... (replaced by the non-bin version)

mm1ke:
Ah, you're right, in that case it's clearly not worth it. I'll close it.

PR #21770, opened 2021-07-24

arthurzam:
@mgorny , well I'm back to bumping to python 3.10, so if you have the time, I will gladly continue.

In this PR I wanted to bump one circular dependency around `dev-python/aiohttp`. Therefore the first 3 commits can't be separated, and I recommend to test using all commits together.
Also, I try to make every change as a separate commit, so if you or the maintainer are against a change, it is easy to remove.

- For `dev-python/async_timeout`, it was fixed upstream, but by many partial commits. I have ran `git diff tests/` to tagged version and got this smallest fixing patch. It changes only tests, therefore should be safe to change even in stable ebuild.
- For `dev-python/pytest-aiohttp` a trivial bump. This one doesn't have test suite, so it was tested by testing `dev-python/aiohttp` which uses it.
- For `dev-python/aiohttp` imported a patch from [here](https://github.com/aio-libs/aiohttp/issues/5905).
- For `dev-python/aiohttp-jinja2` it was trivial bump, and then I tried to improve the doc generation
- For `dev-python/uvloop` it was also trivial bump, I tried to to improve the `setup.cfg` generation (make it more readable, and tested for correct result of course). I wanted to improve doc generation but it failed so haven't touched.

All those changes were tested using `ebuild * test` on a lxd container test machine, and on my main machine.

### DIFF
--- a/dev-python/aiohttp-jinja2/aiohttp-jinja2-1.4.2.ebuild
+++ b/dev-python/aiohttp-jinja2/aiohttp-jinja2-1.4.2.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7,8,9} )
+PYTHON_COMPAT=( python3_{8..10} )
 
 inherit distutils-r1
 
--- a/dev-python/aiohttp-jinja2/aiohttp-jinja2-1.4.2.ebuild
+++ b/dev-python/aiohttp-jinja2/aiohttp-jinja2-1.4.2.ebuild
@@ -14,25 +14,19 @@ SRC_URI="mirror://pypi/${P:0:1}/${PN}/${P}.tar.gz"
 LICENSE="Apache-2.0"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE="doc test"
-RESTRICT="!test? ( test )"
 
 RDEPEND="
 	>=dev-python/aiohttp-2.3.9[${PYTHON_USEDEP}]
 	>=dev-python/jinja-2.10[${PYTHON_USEDEP}]
 "
-DEPEND="
-	${RDEPEND}
-	doc? (
-		>=dev-python/alabaster-0.6.2[${PYTHON_USEDEP}]
-		dev-python/sphinx[${PYTHON_USEDEP}]
-	)
+BDEPEND="
 	test? ( >=dev-python/pytest-aiohttp-0.3.0[${PYTHON_USEDEP}] )
 "
 
 DOCS=( README.rst )
 
 distutils_enable_tests pytest
+distutils_enable_sphinx docs dev-python/alabaster
 
 src_prepare() {
 	default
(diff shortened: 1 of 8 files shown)


sbraz:
> For dev-python/aiohttp imported a patch from here.

Please bear in mind I'm not 100% sure it's correct or I'd have pushed it to Gentoo already :smile:

arthurzam:
> 
> 
> > For dev-python/aiohttp imported a patch from here.
> 
> Please bear in mind I'm not 100% sure it's correct or I'd have pushed it to Gentoo already 😄

Well, I have taken your caution to notice and decided to test it.
First of all, what is removed in py3.10 and is important to us is the `loop` argument to various `asyncio` calls. Because the type of argument it was, only by using passing `loop=...` or using `**kw` it could be set - so lets check all those places (did a `grep -nr . -e loop`). Found a lot of matches.
Scanning every one of them, I checked if we somehow miss and pass it by mistake to a call which would call finally a function which now doesn't accept `loop` argument. All were fixed and all uses of `loop` where correct.
Could we miss a place? Theoretically it is possible, but I'm nearly sure at 99.9% you fixed it flawlessly.

Also, the test suite of `aiohttp` is very good and collects a lot of cases, so I am nearly sure if we have missed a place, it would throw an error during test phase. The fact the test passes is a very good sign we can more sure.

At that stage, I think we can be 100% sure you fixed it correctly.

But of course, we could wait as you have done. In that case, the whole PR is on hold, as without merging `aiohttp`, this PR is useless.

mgorny:
This is a life saver! I've just been blocked from bumping jeepney by missing py3.10 support in async_timeout. Good that I've decided to check your pending PRs before I started porting it myself ;-). Lemme just do some testing.

mgorny (changes requested):
I've rebased your commits and extended uvloop changes to the -r1 ebuild.

However, you seem to have included too many changes in async_timeout patch:

```
============================================================== FAILURES ===============================================================
_______________________________________________________ test_timeout_remaining ________________________________________________________

    @pytest.mark.asyncio
    async def test_timeout_remaining():
        with timeout(None) as cm:
            assert cm.remaining is None
        assert cm.remaining is None
    
        t = timeout(None)
        assert t.remaining is None
    
        t = timeout(1.0)
        assert t.remaining is None
    
        with timeout(1.0) as cm:
            await asyncio.sleep(0.1)
            assert cm.remaining < 1.0
        r = cm.remaining
        time.sleep(0.1)
>       assert r == cm.remaining
E       assert 0.8998326490000181 == 0.7997212050004237
E        +  where 0.7997212050004237 = <async_timeout.timeout object at 0x7f3883629030>.remaining

tests/test_timeout.py:237: AssertionError
________________________________________________________ test_timeout_elapsed _________________________________________________________

    @pytest.mark.asyncio
    async def test_timeout_elapsed():
        t = timeout(None)
>       assert t.elapsed == 0.0
E       AttributeError: 'timeout' object has no attribute 'elapsed'

tests/test_timeout.py:255: AttributeError
```

arthurzam:
About `dev-python/async_timeout`:

1. Fixed `elapsed` wrong patch import
2. Relaxed a little the time delta for `test_timeout_remaining` (looks a little like the last commits upstream to relax time diff).
3. I also added test dependency `dev-python/pytest-asyncio` (didn't build without it and didn't install it when ran in clean new container - so should be added, and fixed).

PR #21772, opened 2021-07-24
media-libs/sdl-net: EAPI8 bump, minor ebuild improvements

mm1ke:

I've looked into https://bugs.gentoo.org/730872 too and i think the problem was because it called `AS` directly. Please have a look. In the `config.status` file it now uses the correct tool, but i couldn't see anywhere in the logs that it was used.

### DIFF
--- a/media-libs/sdl-net/sdl-net-1.2.8-r2.ebuild
+++ b/media-libs/sdl-net/sdl-net-1.2.8-r2.ebuild
@@ -1,0 +1,29 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit multilib-minimal
+
+MY_P="${P/sdl-/SDL_}"
+
+DESCRIPTION="Simple Direct Media Layer Network Support Library"
+HOMEPAGE="https://www.libsdl.org/projects/SDL_net/"
+SRC_URI="https://www.libsdl.org/projects/SDL_net/release/${MY_P}.tar.gz"
+S="${WORKDIR}/${MY_P}"
+
+LICENSE="ZLIB"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~hppa ~ia64 ~ppc ~ppc64 ~sparc ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x86-solaris"
+
+RDEPEND="media-libs/libsdl[${MULTILIB_USEDEP}]"
+DEPEND="${RDEPEND}"
+
+multilib_src_configure() {
+	ECONF_SOURCE="${S}" econf --disable-gui
+}
+
+multilib_src_install_all() {
+	einstalldocs
+	find "${ED}" -name '*.la' -delete || die
+}


media-libs/sdl-net/sdl-net-1.2.8-r2.ebuild
```diff
@@ -0,0 +1,33 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit multilib-minimal toolchain-funcs
+
+MY_P="${P/sdl-/SDL_}"
+DESCRIPTION="Simple Direct Media Layer Network Support Library"
```
ionenwks, line 9: Prefer a clear cut for the description/homepage block, so I usually have a blank line between these.

media-libs/sdl-net/sdl-net-1.2.8-r2.ebuild
```diff
@@ -0,0 +1,33 @@
...
+	AS="$(tc-getAS)" ECONF_SOURCE="${S}" econf \
+		--disable-gui \
+		$(use_enable static-libs static)
+}
+
+multilib_src_install_all() {
+	dodoc CHANGES README
+	if ! use static-libs ; then
+		find "${D}" -type f -name '*.la' -exec rm {} + || die
+	fi
```
ionenwks, line 32: Make deletion unconditional but use the find statement from the policy guide:
https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0303

media-libs/sdl-net/sdl-net-1.2.8-r2.ebuild
```diff
@@ -0,0 +1,33 @@
...
+DEPEND="${RDEPEND}"
+
+multilib_src_configure() {
+	AS="$(tc-getAS)" ECONF_SOURCE="${S}" econf \
+		--disable-gui \
+		$(use_enable static-libs static)
+}
+
+multilib_src_install_all() {
+	dodoc CHANGES README
```
ionenwks, line 29: CHANGES and README are part of einstalldocs' defaults, so should just use that.

media-libs/sdl-net/sdl-net-1.2.8-r2.ebuild
```diff
@@ -0,0 +1,33 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit multilib-minimal toolchain-funcs
+
+MY_P="${P/sdl-/SDL_}"
+DESCRIPTION="Simple Direct Media Layer Network Support Library"
+HOMEPAGE="https://www.libsdl.org/projects/SDL_net/index.html"
```
ionenwks, line 10: Seem can drop the index.html bit.

media-libs/sdl-net/sdl-net-1.2.8-r2.ebuild
```diff
@@ -0,0 +1,33 @@
...
+LICENSE="ZLIB"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~hppa ~ia64 ~ppc ~ppc64 ~sparc ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~x86-solaris"
+IUSE="static-libs"
+
+RDEPEND="media-libs/libsdl[${MULTILIB_USEDEP}]"
+DEPEND="${RDEPEND}"
+
+multilib_src_configure() {
+	AS="$(tc-getAS)" ECONF_SOURCE="${S}" econf \
```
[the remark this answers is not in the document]
mm1ke: (Un)fortunately i couldn't reproduce it either.

What i found out was that after `configuring` the package, the generated `libtool` script contains a line with `as=AS`. However, i couldn't find anywhere in the build that `AS` was used actually.
I also saw that the llvm toolchain was used and was thinking this bug happens in combination with llvm (thus my initial note that i think i fixed it)..
And with setting `AS="$(tc-getAS)"` i could fix the `libtool` file to set it to the correct system variable..

However, since you also couldn't reproduce the problem with the llvm toolchain, I'm happily remove the workaround - and sorry that i haven't explained it more in detail.

media-libs/sdl-net/sdl-net-1.2.8-r2.ebuild
```diff
@@ -0,0 +1,33 @@
...
+		--disable-gui \
+		$(use_enable static-libs static)
```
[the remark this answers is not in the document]
mm1ke: Ok, i've also removed it now from the smpeg{,2} ebuilds in the other PR. I've checked the reverse-dependencies and it's not used anywhere.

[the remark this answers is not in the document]
ionenwks: Nice, indeed I don't see a need for it on smpeg either (hadn't looked yet).

All these libraries tend to need more attention and if going to revbump into ~arch want to get more done at once, part of why nobody bothered to touch them yet -- but this had to be done sooner or later.

ionenwks (approved):

PR #21775, opened 2021-07-24
net-firewall/nftables fixes to #789306 and #790014

klondi:
These two commits fix the issues on bug 789306 and provide a workaround of 790014 (until upstream fixes parsing for rulesets saved using -n).


klondi:
It may be better to instead backport the upstream fix for #790014 I'm colsing for now so I can prepare a new PR instead.

klondi:
Backported Pablo's patch instead to 0.9.9 this should close the bug (until we get a 1.0.0).

PR #21777, opened 2021-07-25
riscv: keyword lxqt-base/lxqt-meta, x11-misc/lightdm, x11-misc/sddm and a few qt depedencies

alexfanqi:
Compilation and all tests have passed. I also tested lxqt-meta via tigervnc and lightdm via x11vnc.


alexfanqi:
USE=iso is masked for kde-frameworks/solid. I don't have an Apple device/ can't test it physically.

PR #21780, opened 2021-07-25
media-video/binkplayer: EAPI8 bump, minor ebuild improvements

mm1ke:

### DIFF
--- a/media-video/binkplayer/binkplayer-1.99w.ebuild
+++ b/media-video/binkplayer/binkplayer-1.99w.ebuild
@@ -1,25 +1,24 @@
-# Copyright 1999-2015 Gentoo Foundation
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=5
+EAPI=8
+
 DESCRIPTION="Bink Video! Player"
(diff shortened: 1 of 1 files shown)


media-video/binkplayer/binkplayer-1.99w-r1.ebuild
```diff
@@ -0,0 +1,28 @@
...
+HOMEPAGE="http://www.radgametools.com/default.htm"
+# No version on the archives and upstream has said they are not
+# interested in providing versioned archives.
+SRC_URI="mirror://gentoo/${P}.zip"
+S="${WORKDIR}"
+
+# distributable per http://www.radgametools.com/binkfaq.htm
+LICENSE="freedist"
+SLOT="0"
+KEYWORDS="-* ~amd64 ~x86"
```
ionenwks, line 16: Technically you did the right thing here, but I think revbump + dropping to ~arch can be skipped here. This has no revdeps, seem scarcely used, changes are trivial (hard for a single dobin to fail), and RDEPEND changes only remove ancient constraints.

I wouldn't say that for libraries like smpeg (on that note, I'll be checking the libraries PRs a bit later assuming no one else does first, I haven't missed them).

ionenwks (approved):
Thanks :)

PR #21785, opened 2021-07-25
games-arcade/ascii-invaders: EAPI8, version bump, remove old

mm1ke:
Hi,

Seems like `ascii-invaders` got a new home and a new release. (from 2016)
I've updated the ebuild and added the new release. Since the changes are minor, i've directly removed the old version too.


games-arcade/ascii-invaders/ascii-invaders-1.0.1.ebuild
```diff
@@ -0,0 +1,33 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit toolchain-funcs
+
+DESCRIPTION="Space invaders clone, using ncurses library"
+HOMEPAGE="https://github.com/macdice/ascii-invaders"
```
ionenwks, line 9: When you see something moved to github/gitlab and such, nice to have a separate commit to update metadata.xml like commit 445c58728bba63b0bb937416a1cfec699ffe5f82

Feel free to adjust indenting at same time to be more like the above commit (other devs may have their own preferences but that's at least how games@ normally does it).

And you can remove the longdescription, this doesn't really add anything over the short description. That this is compiling for Linux sure isn't news :)

games-arcade/ascii-invaders/Manifest
```diff
@@ -1,2 +1 @@
 DIST ascii-invaders-1.0.1.tar.gz 37578 BLAKE2B 181509c07b1d9ae54f9147f7425c3d62a57f9bff9561fca6af3f3cc15b7fc96f62c809576bdbbe0ba6a2801947941e259e788f99980174848efedbc71c8a969c SHA512 326708ffc2c277e04575920c58f6a267e0bf46f8d2a3dbca9b5c317fb01006d4f673ab35f92f292549cc52dbee4400a84c85478eb81b9bd7703689a705547bc7
-DIST invaders0.1b.tgz 14597 BLAKE2B ecccca6636c4d8dffbc57b8cec82471efc2fd1a256648b9ac46fd256ef99d884205f1c86ff2bf0aedb7bb5996a59a7af3f301339944c6a4753355a8a4db23190 SHA512 3ac9dee0c532384e2849fdc8910bce885e3fdef362b629e8637df1cc0d82eba52c02a836a4925d811b339e9232f20caa7c761dbc1feeddc74e9b6f09e97046b8
```
ionenwks, line 1: I don't disagree that it's trivial and can be removed, but unless it was flat out broken I usually still give it some time (time will vary depending on how trivial it is, can be a week just to see if there's tinderbox bugs -- on that note, there was one for openglad because premake4 generated makefiles with race condition issues :-1: ).

Assuming I don't forget to, I make a note of bumps I did and do a batch later on, so I'll handle removing that later and it won't just sit there forever for nothing. If you were the maintainer, I would expect you to remove it so it'd be a different story.

games-arcade/ascii-invaders/ascii-invaders-1.0.1.ebuild
```diff
@@ -0,0 +1,33 @@
...
+BDEPEND="virtual/pkgconfig"
+
+src_prepare() {
+	default
+	rm -f Makefile
+}
+
+src_compile() {
+	tc-export CC
+	emake LDLIBS="$($(tc-getPKG_CONFIG) ncurses --libs)" invaders
```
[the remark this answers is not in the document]
mm1ke: Changed the ebuild as you suggested. Thanks again. :)
I didn't know about the pkg-config issue but it's clearly good to know.

games-arcade/ascii-invaders/ascii-invaders-1.0.1.ebuild
mm1ke: Ahh, yeah, you're right. Checking the `metadata.xml` file is something i easy forget sometimes... :D
I've updated it and removed the longdescription.

games-arcade/ascii-invaders/Manifest
mm1ke: No problem, i've removed the commit. I know how small changes sometimes can make a big difference, so better save then sorry.. ;)

ionenwks (approved):
Thanks, I'll just squash the `upd` commit, guess overlooked it (not a problem).

PR #21787, opened 2021-07-25
media-plugins/gst-plugins-qt5: new package

deepbluev7:
Nheko uses this to implement video calls. Since I was using this locally, I thought I might as well contribute it. (I was testing how well gstreamer 1.18.4 in the tree works with the video calls in Nheko).

The patch was needed to make the plugin build and is taken from the gstreamer upstream. Not quite sure about the use flags. The meson.build is a bit confusing: https://gitlab.freedesktop.org/gstreamer/gst-plugins-good/-/blob/1.18.4/ext/qt/meson.build

I've also recently seen a lot of gstreamer_system_libraries and similar calls land in gstreamer ebuilds, but not quite sure, when and why to use them.

qt5 might also be the wrong name, but from what I can tell, that is what upstream calls this. No idea, if that will change with qt6.


deepbluev7:
I have updated the PR with a link to the [bug](https://bugs.gentoo.org/810814) now and removed the mask, since gstreamer as a whole got unmasked. (I'm assuming the rebase was fine, since none looked at it yet)

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
```diff
@@ -0,0 +1,36 @@
...
+	wayland? ( dev-qt/qtwayland[${MULTILIB_USEDEP}] )
+	X? ( dev-qt/qtx11extras[${MULTILIB_USEDEP}] )
+"
+DEPEND="${RDEPEND}"
+
+PATCHES=(
+	"${FILESDIR}/gst-dont-use-volatile-to-mean-atomic-1.18.4.patch"
+)
+
+GST_PLUGINS_ENABLED="qt5"
```
lanodan, line 35: This is the default value for split ebuilds (ones where `PN != GST_ORG_MODULE`).

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
```diff
@@ -0,0 +1,36 @@
...
+
+DESCRIPTION="Qt5 QML video sink plugin"
+KEYWORDS="~amd64 ~arm ~arm64 ~x86"
+IUSE="+egl wayland +X orc"
+
+REQUIRED_USE="
+	wayland? ( egl )
+"
+
+GST_REQ="${PV}"
```
lanodan, line 17: Not really sure why you would define this variable.

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
```diff
@@ -0,0 +1,36 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+GST_ORG_MODULE=gst-plugins-good
+
+inherit gstreamer-meson
+
+DESCRIPTION="Qt5 QML video sink plugin"
```
lanodan, line 9:
```suggestion
DESCRIPTION="Qt5 QML video sink plugin for GStreamer"
```

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
deepbluev7: I have no idea, I copied that from another gstreamer ebuild. :D

deepbluev7:
I should have applied all feedback. Not sure if amending is the right approach, but other PRs also only seem to have one commit. Thanks a lot for the review, @lanodan !

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
deepbluev7: Removing this line seems to not work, did you mean to use the literal `${PN/gst-plugins-/}`?

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
deepbluev7: This was used in the dependency below, but I could just replace it by using the PV there directly.

lanodan:
>  Status: x broken
> https://qa-reports.gentoo.org/output/gentoo-ci/76c08021ed/output.html#media-plugins/gst-plugins-qt5

Looks like qt isn't made to be multilib… that's going to be a weird one to solve.

deepbluev7:
Oh, so that is what that warning is about... that will be fun indeed.

deepbluev7:
I masked the multilib useflags on x86 and amd64 for this ebuild now. I think that should be enough?

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
```diff
@@ -0,0 +1,34 @@
...
+RDEPEND="
+	>=media-libs/gst-plugins-base-${PV}:${SLOT}[egl?,opengl,wayland?,X?]
+	media-libs/mesa[egl?,X?,${MULTILIB_USEDEP}]
+
+	dev-qt/qtcore
+	dev-qt/qtgui
+	egl? ( dev-qt/qtgui[eglfs] )
+	dev-qt/qtdeclarative
+	wayland? ( dev-qt/qtwayland )
+	X? ( dev-qt/qtx11extras )
```
a17r, line 26: What if Qt6 comes around?

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
```diff
@@ -0,0 +1,34 @@
...
+IUSE="+egl wayland +X orc"
+
+REQUIRED_USE="
+	wayland? ( egl )
+"
+
+RDEPEND="
+	>=media-libs/gst-plugins-base-${PV}:${SLOT}[egl?,opengl,wayland?,X?]
+	media-libs/mesa[egl?,X?,${MULTILIB_USEDEP}]
+
```
a17r, line 20: superfluous empty line

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
```diff
@@ -0,0 +1,34 @@
...
+	dev-qt/qtcore
+	dev-qt/qtgui
+	egl? ( dev-qt/qtgui[eglfs] )
+	dev-qt/qtdeclarative
```
a17r, line 24: please keep dependencies sorted, with unconditional dependencies ahead of conditional block.

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
```diff
@@ -0,0 +1,34 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+GST_ORG_MODULE=gst-plugins-good
+
+inherit gstreamer-meson
+
+DESCRIPTION="Qt5 QML video sink plugin for GStreamer"
+KEYWORDS="~amd64 ~arm ~arm64 ~x86"
```
a17r, line 10: did you test all those arches?

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
```diff
@@ -0,0 +1,34 @@
...
+IUSE="+egl wayland +X orc"
+
+REQUIRED_USE="
+	wayland? ( egl )
+"
+
+RDEPEND="
+	>=media-libs/gst-plugins-base-${PV}:${SLOT}[egl?,opengl,wayland?,X?]
+	media-libs/mesa[egl?,X?,${MULTILIB_USEDEP}]
```
a17r, line 19: `${MULTILIB_USEDEP}` doesn't make sense if this can't be multilib anyway due to Qt5.

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
```diff
@@ -0,0 +1,34 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
a17r, line 4: empty line after EAPI please.

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
```diff
@@ -0,0 +1,34 @@
...
+GST_ORG_MODULE=gst-plugins-good
+
+inherit gstreamer-meson
+
+DESCRIPTION="Qt5 QML video sink plugin for GStreamer"
+KEYWORDS="~amd64 ~arm ~arm64 ~x86"
+IUSE="+egl wayland +X orc"
```
a17r, line 11: please keep them sorted

a17r (changes requested):

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
deepbluev7: Right, good point, will fix.

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
deepbluev7: I haven't tested the ebuild on all of those architectures, but I have built that specific plugin on those platforms and used it. Would it be best to just have "~amd64" in that list and leave the others out until someone tests the ebuild on those platforms?

deepbluev7: I have removed everything but amd64 for now

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
deepbluev7: Added. Format was copied from another GStreamer ebuild.

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
deepbluev7: Should be sorted now. Should I have added an empty line before the conditionals? Not sure what is used more often in that case?

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
leio: Absolutely all have no newline, so neither should it have it here

profiles/arch/x86/package.use.mask
```diff
@@ -1,6 +1,10 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Nicolas Werner <nicolas.werner@hotmail.de> (2021-09-05)
+# Qt5 packages don't support multilib, so disable it for this plugin.
+media-plugins/gst-plugins-qt5 abi_x86_64
```
leio, line 6: x86_64 is really exposed for x86 to need this? Also it's not keyworded x86, so I don't think this mask was added based on repoman complaining

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
```diff
@@ -0,0 +1,34 @@
...
+
+EAPI=7
+
+GST_ORG_MODULE=gst-plugins-good
+
+inherit gstreamer-meson
+
+DESCRIPTION="Qt5 QML video sink plugin for GStreamer"
+KEYWORDS="~amd64"
+IUSE="+egl orc wayland +X"
```
leio, line 12: orc should be default enabled, BUT qt plugins don't use orc, so it shouldn't be in the IUSE at all. The QA warning is a false positive from a wrong check in new eclasses [that's yet unsolved](https://bugs.gentoo.org/810417)

profiles/arch/amd64/package.use.mask
```diff
@@ -17,6 +17,10 @@
 
 #--- END OF EXAMPLES ---
 
+# Nicolas Werner <nicolas.werner@hotmail.de> (2021-09-05)
+# Qt5 packages don't support multilib, so disable it for this plugin.
+media-plugins/gst-plugins-qt5 abi_x86_32
```
leio, line 22: I feel a bit uneasy about an exception here, when we've had it all work for everything else. I guess it's a bit special, but maybe it'd be possible to not add the multilib stuff to it in the first place. However that's probably too much eclass churn..

leio (changes requested):

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
leio: Deps should be in the order they appear in meson.build, not alphabetically. This is a style difference between a17r and gnome/gstreamer teams. @a17r shouldn't be requesting such changes (in this way that implies it has to get sorted alphabetically) for gstreamer/gnome packages, but I guess technically this was no-ones yet :)

a17r: It's a style difference between gnome/gstreamer and everyone else.

profiles/arch/x86/package.use.mask
deepbluev7: Yep, totally not needed, my mind just went to automode there!

profiles/arch/amd64/package.use.mask
deepbluev7: I think this is the easiest solution for now. If we made the multilib optional in the eclass and only had one consumer using that, it would not get much testing and add a bit of complexity. I'd argue changing the eclass, if there is a second or third package with this issue makes the most sense?

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
deepbluev7: Tbh, I think the order it is in now does make sense. The mesa dependency is just needed to pass through the use flags, so it isn't explicitly in the meson.build and apart from that there is only the qt dependency. So having the 2 use deps on the top and then the meson.build deps after that seems reasonable and is alphabetically sorted as well as pretty much matches the meson.build. Or should I still change anything? I did really appreciate @a17r review, especially as they seem to be involved with qt packages and this package kinda falls in between and it helped resolve quite a few issues!

deepbluev7:
Sorry it took me a bit to respond to the feedback, I was travelling for work. I think I addressed everything I could, the only items left are mostly discussions or stuff I don't know how to resolve. Thanks a lot for the reviews!

media-plugins/gst-plugins-qt5/gst-plugins-qt5-1.18.4.ebuild
a17r: Another style difference between gnome/gstreamer and everyone else.

deepbluev7:
Rebased to get rid of the mask conflicts. Seems like this is in guru now too, so that voip in Nheko works, but I feel like it would be reasonable to have all the gstreamer plugins in the main tree :3

deepbluev7:
Closing, since it is in guru now and there seems to be little interest in merging this to ::gentoo. I do think it would be less error prone there because of the multilib issue, but qt6 is right around the corner and my packages will probably switch to that.

PR #21788, opened 2021-07-26
Cinnamon Updates (07.24)

SparkyBluefang:
Latest Cinnamon 5.0 releases.


juippis (approved):
There was something weird with cinnamon's gtk-doc, first time it failed in src_install, but passed everytime after. Might be some issue in parallelelization, just in case you get a bug notification about it.

Thanks!

PR #21792, opened 2021-07-26
dev-lang/uasm: new package

Karlson2k:
UASM is JWasm fork and replacement.
It uses different binary names so can be used alongside with JWasm.

JWasm hasn't been updated for serveral years and lacks support for new CPU commands.

MASM-compatible assembler is used in 7-zip package.
See https://sourceforge.net/p/sevenzip/discussion/45797/thread/7fe6c21efa/
and https://github.com/gentoo/gentoo/pull/21754


arthurzam:
Ah, I missed this. It have been imported from ::guru. Feel free to send PRs to that. Sorry & thanks.

PR #21796, opened 2021-07-26
app-emulation/wine-mono: add shared

Hello71:


sarnex:
Can you please check the contribution steps here? The PR should be to ::wine

https://wiki.gentoo.org/wiki/Project:Wine/Policies_and_Procedures#Contribution_steps:

Also, we need a signed-of-by

Hello71:
merged in ::wine

PR #21797, opened 2021-07-26
app-emulation/wine-gecko: add shared

Hello71:
skip 2.47 and 2.47.1, because poor support for shared upstream in those
versions, and they are also not used by any current gentoo wines


sarnex:
Same, see comment in https://github.com/gentoo/gentoo/pull/21796#issuecomment-887157729

Hello71:
merged in ::wine

PR #21800, opened 2021-07-26
games-puzzle/tanglet: bump to 1.6.1.1 fixing compilation with >=qt*-5.14

kobboi:
No new dependencies found in the binary. Existing patch still applies.


games-puzzle/tanglet/tanglet-1.6.1.1.ebuild
```diff
@@ -0,0 +1,37 @@
...
+inherit qmake-utils xdg
+
+DESCRIPTION="Single player word finding game based on Boggle"
+HOMEPAGE="https://gottcode.org/tanglet/"
+SRC_URI="https://gottcode.org/${PN}/${P}-src.tar.bz2"
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
```
ionenwks, line 15: No need for empty IUSE, remove the line.

I know it's just a copy of the old ebuild, but most games@ ebuild aren't a good reference, nice to cleanup things like this on bumps (I've been trying to improve games@ ebuilds slowly but there's 777 packages and hadn't gotten to tanglet yet, help is appreciated especially if want these fixed earlier).

games-puzzle/tanglet/tanglet-1.6.1.1.ebuild
```diff
@@ -0,0 +1,37 @@
+# Copyright 1999-2019 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit qmake-utils xdg
+
+DESCRIPTION="Single player word finding game based on Boggle"
+HOMEPAGE="https://gottcode.org/tanglet/"
+SRC_URI="https://gottcode.org/${PN}/${P}-src.tar.bz2"
```
ionenwks, line 10: Trivial but `${PN}` be nicer plaintexted

ionenwks (commented):
I kind of wanted to suggest migration to cmake, but seems the CMakeLists.txt is QT6-only. I guess Qt5 is on its last legs for this package.

More importantly, is this really running as expected for you? I just get a "Unable to read dice from file". Or at least with a blank state, maybe you have old settings that let it work for you?

From a quick look at strace it tried:
```
29748 access("/usr/bin/data/en/dice", F_OK) = -1 ENOENT (No such file or directory)
29748 access("/usr/share/data/en/dice", F_OK) = -1 ENOENT (No such file or directory)
29748 access("/usr/Resources/data/en/dice", F_OK) = -1 ENOENT (No such file or directory)
29748 openat(AT_FDCWD, "tanglet:en/dice", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
```
These paths don't look right, it should be at /usr/share/tanglet/data/en/dice

kobboi:
@ionenwks  Thanks for your feedback. I tested with ebuild, not emerge. I will re-investigate and post an update

kobboi:
Pretty sure that what happened is: I tested with the patch disabled, then enabled it when I noticed the /usr/local paths and did not properly re-test. I force-pushed a new commit that addresses your points and this time the game is working properly. Note that at the time the compilation (unknown version) was still OK, I remember the game also not working, with similar error messages as you describe above. Likely the patch never worked, or at least not with 1.5.3 (maybe I did with 1.5.2, as per the filename)

kobboi:
Force-pushed the above suggestion and adapted the commit message accordingly.

ionenwks (approved):
Thanks.

PR #21801, opened 2021-07-26
mail-filter/spamassassin: Add USE="bayes"

Philippe23:
People are confused now that berkdb is no longer enabled by default.  So
I'm adding a bayes use flag that requires some type of storage for any
learning features.

Closes: https://bugs.gentoo.org/802189
Signed-off-by: Philippe Chaintreuil <gentoo_bugs_peep@parallaxshift.com>

---

ebuild diff for easy reference:
```diff
--- spamassassin-3.4.6-r1.ebuild        2021-07-26 16:56:15.643784404 -0400
+++ spamassassin-3.4.6-r2.ebuild        2021-07-26 17:42:27.434323495 -0400
@@ -14,8 +14,9 @@
 LICENSE="Apache-2.0 GPL-2"
 SLOT="0"
 KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~ppc ~ppc64 ~s390 ~sparc ~x86 ~amd64-linux ~x86-linux"
-IUSE="berkdb cron ipv6 ldap mysql postgres qmail sqlite ssl test"
+IUSE="+bayes berkdb cron ipv6 ldap mysql postgres qmail sqlite ssl test"
 RESTRICT="!test? ( test )"
+REQUIRED_USE="bayes? ( || ( berkdb mysql postgres sqlite ) )"

 # The Makefile.PL script checks for dependencies, but only fails if a
 # required (i.e. not optional) dependency is missing. We therefore
```
(See commit diffs for changes to `metadata.xml`.)


grobian:
fwiw lgtm

juippis (commented):
What exactly does bayes do here? I don't see any configure option for it in this PR.

grobian:
it's always included, and by default enabled.  It's an auto-learn classifier, see also https://cwiki.apache.org/confluence/display/spamassassin/BayesInSpamAssassin

grobian:
and in addition, check https://cwiki.apache.org/confluence/display/SPAMASSASSIN/BasicConfiguration for the autolearn option

juippis:
Well if it can be toggled, then it should be toggled with this USE flag option. 

I don't think adding a `REQUIRED_USE` for this makes much sense. 
You could: 
1: auto-default to berkdb or any other db (why was it removed?),
2: inform the user with elog messages if they have / have not any db, and therefore beyes enabled,
3: possibly even toggle the beyes option with these db options.

grobian:
Required-use makes a lot of sense, but I'm fine without as well.  We shouldn't default to berkdb (read the -dev ML), which is what required_use would actually help with.

Toggling it, was my suggestion on the bugzie bug https://bugs.gentoo.org/802189

PR #21802, opened 2021-07-27
app-crypt/trousers: Fix permissions and ownership of /etc/tcsd.conf

salahcoronya:
tcsd will not start as-is until the permissions and ownership is changed on /etc/tcsd.conf, this fixes the ebuild to set the permissions and ownership correctly so it works "out of the box"

### DIFF
--- a/app-crypt/trousers/trousers-0.3.14-r3.ebuild
+++ b/app-crypt/trousers/trousers-0.3.14-r3.ebuild
@@ -64,5 +64,7 @@ src_install() {
 	systemd_dounit "${FILESDIR}"/tcsd.service
 	udev_dorules "${FILESDIR}"/61-trousers.rules
 	fowners tss:tss /var/lib/tpm
+	fowners :tss /etc/tcsd.conf
+	fperms 640 /etc/tcsd.conf
 	readme.gentoo_create_doc
 }
--- a/app-crypt/trousers/trousers-0.3.15.ebuild
+++ b/app-crypt/trousers/trousers-0.3.15.ebuild
@@ -61,5 +61,7 @@ src_install() {
 	systemd_dounit "${FILESDIR}"/tcsd.service
 	udev_dorules "${FILESDIR}"/61-trousers.rules
 	fowners tss:tss /var/lib/tpm
+	fowners :tss /etc/tcsd.conf
+	fperms 640 /etc/tcsd.conf
 	readme.gentoo_create_doc
 }


app-crypt/trousers/trousers-0.3.14-r3.ebuild
```diff
@@ -64,5 +64,7 @@ src_install() {
 	systemd_dounit "${FILESDIR}"/tcsd.service
 	udev_dorules "${FILESDIR}"/61-trousers.rules
 	fowners tss:tss /var/lib/tpm
+	fowners root:tss /etc/tcsd.conf
```
ionenwks, line 67:
```suggestion
	fowners :tss /etc/tcsd.conf
```
Same for other ebuild.

ionenwks (commented):
When applying the same fix to multiple versions, it's typically done in a single commit and so be nice to squash.

app-crypt/trousers/trousers-0.3.14-r3.ebuild
```diff
@@ -64,5 +64,7 @@ src_install() {
 	systemd_dounit "${FILESDIR}"/tcsd.service
 	udev_dorules "${FILESDIR}"/61-trousers.rules
 	fowners tss:tss /var/lib/tpm
+	fowners :tss /etc/tcsd.conf
+	fperms 640 /etc/tcsd.conf
```
ionenwks, line 68: Actually, just realized something. I feel like this bug was reported about 0.3.14-r3 only, so when I checked 0.3.15 I /thought/ root was default.

Without those two lines, 1.3.15 gives me a fine-looking:
```
-rw-r----- 1 root tss 7046 Jul 26 22:41 /etc/tcsd.conf
```

1.3.14 (still without those lines) however gives me:
```
-rw------- 1 tss tss 7046 Jul 26 22:44 /etc/tcsd.conf
```
So rather than fowners :tss or root:tss, need `fowners root` + 640

Please double-check and if this really starts fine given changing a stable ebuild.

Also a revbump may be good, in the event the config file is unmodified the permissions will be auto-updated.

ionenwks: Alternatively, a fix could be to simply stable 0.3.15.. if it's fixed, and there's nothing holding this back, then this PR may not be needed.

ConiKost: I would like also to propose, why just not adjust init script and call checkpath to adjust permissions during startup?

salahcoronya: Adjusting the initscript won't help systemd users. trousers-0.3.15 doesn't have this problem (it installs with the correct permissions, so lets just stabilize it.

salahcoronya:
app-crypt/trousers-0.3.15 seems to fix the problem, so I'm going to close this PR and request stabilization

PR #21805, opened 2021-07-27
net-p2p/go-ipfs: verbump to 0.9.1

stkw0:


kakra:
@stkw0 Maybe remove the 0.7 version, especially since it is affected by a security issue?

kakra:
You could then add a `Closes: https://bugs.gentoo.org/778182`

stkw0:
sorry for the delay, will do it now

juippis (approved):

juippis:
Oh right, we can't close security bugs with commits as they need to be checked by the security team, so just reference a security bug with `Bug:` instead in future.

PR #21806, opened 2021-07-27
elastic: security bump

hydrapolic:
Thank you!


juippis (approved):
at least some of these have EAPI-8 available, but... next time!

hydrapolic:
Thanks @juippis, yeah, we'll migrate to eapi 8, good point.

PR #21809, opened 2021-07-27
games-arcade/ceferino: EAPI8, minor improvements

mm1ke:

Simple EAPI8 bump with some minor cleanups. Some notes:
* when starting the game i'll getting an error like:
```
+ Loading sound files: ............. [ERROR]
Can't load 'v�=��U': Mix_LoadWAV_RW with NULL src
```
This error is present regardless of how `sdl-mixer` is built (I've tried it with sdl-mixer[mod,modplug] or sdl-mixer[mod,mikmod]) This isn't a regression since the original has this problem too. The game is also playable but it seems it misses some audio effects - though, there is still a general music playing in the background.

* I've also removed the `pkg_postinst`. I don't understand the reasoning behind this anyway. If sound should be optional, then at least a `optfeature` notice should be made. But i guess sound in games is general something the typical user would like to have...


games-arcade/ceferino/ceferino-0.97.8-r2.ebuild
```diff
@@ -1,34 +1,34 @@
...
 
-EAPI=6
+EAPI=8
+
 inherit autotools desktop
 
 DESCRIPTION="Super-Pang clone (destroy bouncing balloons with your grapnel)"
 HOMEPAGE="http://www.losersjuegos.com.ar/juegos/ceferino"
 SRC_URI="mirror://debian/pool/main/c/ceferino/${PN}_${PV}+svn37.orig.tar.gz"
+S="${WORKDIR}/${P}+svn37"
```
ionenwks, line 11:
```suggestion
S+="+svn37"
```
It's not done often, but this is safe to do.

ionenwks (commented):
>     * I've also removed the `pkg_postinst`. I don't understand the reasoning behind this anyway. If sound should be optional, then at least a `optfeature` notice should be made. But i guess sound in games is general something the typical user would like to have...

That was likely made before optfeature.eclass existed (it's not that old), so people used to do manual has_version checks + message. That aside, I'm generally in agreement that music/audio is expected to work on a game. Plus, when support is missing, games tend to throw a lot of errors even if they "work" and it just doesn't seem right. If the game had a --disable-sound that properly disable support + skip a dependency I'd respect that and add a IUSE="+sound" but this doesn't apply here. Anyhow, what you did is fine.

> ```
> + Loading sound files: ............. [ERROR]
> Can't load 'v�=��U': Mix_LoadWAV_RW with NULL src
> ```

That's easy to fix, compiler gave a hint:
```
audio.cc:208:30: warning: iteration 11 invokes undefined behavior [-Waggressive-loop-optimizations]
  208 |                 Mix_FreeChunk(sonidos[i]);
```

There's 12 wav files, all loops are set for this, but then in audio.h:
```
Mix_Chunk *sonidos[11];
```

Maybe patch it? :)

I'm amazed this went unnoticed for about 5-6 years (likely a bad pick from this +svn release). Well, maybe it worked with older compilers, not called undefined behavior for nothing.

games-arcade/ceferino/ceferino-0.97.8-r2.ebuild
```diff
@@ -1,34 +1,34 @@
...
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
 IUSE="nls"
 
 RDEPEND="
-	>=media-libs/libsdl-1.2[video]
-	>=media-libs/sdl-image-1.2
-	>=media-libs/sdl-mixer-1.2
+	media-libs/libsdl[video]
+	media-libs/sdl-image
```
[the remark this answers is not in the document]
mm1ke: Thanks for checking these, I'll try to keep an eye on these :)

mm1ke:
@ionenwks 
Now it seems to work as expected. Thanks for the hint, after reading your comment and looking further into it I got it :D

games-arcade/ceferino/ceferino-0.97.8-r2.ebuild
```diff
@@ -1,35 +1,34 @@
...
 KEYWORDS="~amd64 ~x86"
 IUSE="nls"
 
 RDEPEND="
-	>=media-libs/libsdl-1.2[video]
-	>=media-libs/sdl-image-1.2
-	>=media-libs/sdl-mixer-1.2
+	media-libs/libsdl[sound,video]
+	media-libs/sdl-image[jpeg,png]
+	media-libs/sdl-mixer[mod]
```
ionenwks, line 21: Similarly to other PRs, means needs a revbump.

games-arcade/ceferino/ceferino-0.97.8-r2.ebuild
```diff
@@ -1,35 +1,34 @@
...
 	nls? ( virtual/libintl )
 "
 DEPEND="${RDEPEND}
 	nls? ( sys-devel/gettext )
 "
 
-S="${WORKDIR}/${P}+svn37"
+PATCHES=( "${FILESDIR}"/${P}-fix-audio.patch )
```
ionenwks, line 28: Unimportant but, even for a single file, it's typically preferred to:
```
PATCHES=(
    file
)
```
Makes it easier to add/change patches without restyling every time.

games-arcade/ceferino/files/ceferino-0.97.8-fix-audio.patch
```diff
@@ -0,0 +1,11 @@
+--- a/src/audio.h	2013-08-05 16:29:49.000000000 +0200
++++ b/src/audio.h	2021-07-29 18:51:16.668397338 +0200
```
ionenwks, line 2: Believe already know what to do with these timestamps now :)

Normally be good to have a short description on top too, "fix audio" already says most of it but maybe add something like "12 .wav files for an array with 11 elements invokes undefined behavior."

On as side-note, for simple patch like this you could also use a smaller unified context, using diff's -U 2 or -U 1. Not that I think many devs do this beside me.

mm1ke: Updated the patch with `diff -U2` and removed the timestamp :)

ionenwks (approved):

ionenwks:
I'll also move gettext to BDEPEND, forgot about that before.

games-arcade/ceferino/ceferino-0.97.8-r2.ebuild
ionenwks: > It's not done often, but this is safe to do.

Need to correct myself here, while (currently) safe *with portage* it's actually not per PMS which allows for `S` to be undefined in global scope.
https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-10900011.1

So please forget what I said here. I've reverted this in commit 3dd5de66b4400f7b37ace36c4cb006b65f20a739

mm1ke: Got it. Thanks for letting me know :)

PR #21812, opened 2021-07-27
Libffi compat 7

trofi:
Allow libffi-compat:7 alternative

Allow libffi.so.8 from dev-libs/libffi and libffi.so.7 from
libffi-compat:7 to coexist in a single system.

Otherwise this binary package install attempts to downgrade libffi.

Bug: https://bugs.gentoo.org/804660

### DIFF
--- a/games-action/minecraft-launcher/minecraft-launcher-928-r3.ebuild
+++ b/games-action/minecraft-launcher/minecraft-launcher-928-r3.ebuild
@@ -21,7 +21,7 @@ RDEPEND="
 	app-crypt/libsecret
 	dev-libs/nss
 	dev-libs/libbsd
-	dev-libs/libffi:0/7
+	|| ( dev-libs/libffi:0/7 dev-libs/libffi-compat:7 )
 	dev-libs/libpcre
 	media-libs/alsa-lib
 	media-libs/openal
--- a/app-office/wps-office/wps-office-11.1.0.10161-r2.ebuild
+++ b/app-office/wps-office/wps-office-11.1.0.10161-r2.ebuild
@@ -31,7 +31,7 @@ RDEPEND="
 	dev-libs/expat
 	dev-libs/glib:2
 	dev-libs/libbsd
-	dev-libs/libffi:0/7
+	|| ( dev-libs/libffi:0/7 dev-libs/libffi-compat:7 )
 	dev-libs/libgcrypt:0
 	dev-libs/libgpg-error
 	dev-libs/libpcre:3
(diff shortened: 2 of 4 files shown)


games-action/minecraft-launcher/minecraft-launcher-928-r3.ebuild
```diff
@@ -21,7 +21,7 @@ RDEPEND="
 	app-crypt/libsecret
 	dev-libs/nss
 	dev-libs/libbsd
-	dev-libs/libffi:0/7
+	|| ( dev-libs/libffi:0/7 dev-libs/libffi-compat:7 )
```
thesamesam, line 24: I suggest we just depend on libffi-compat:7 to avoid backtracking and possibly confusing conflicts for Portage.

What do you think?

thesamesam (changes requested):

games-action/minecraft-launcher/minecraft-launcher-928-r3.ebuild
AndrewAmmerlaan: So the thing with `games-action/minecraft-launcher` is that it updates similarly to `net-im/discord-bin`, i.a.w. the installed binary is just a small application that fetches the actual launcher and keeps it up to date. Meaning that at some point the launcher will be updated to `dev-libs/libffi:0/8`, probably without the release of a new version of the fetching binary. 

We should probably indeed switch this to `dev-libs/libffi-compat:7` now, but at some point we are going to have to revbump it again to switch it back to  `dev-libs/libffi:0/8`, and because of the closed-source nature of this package and the lack of communication or proper changelog from upstream we are not going to know exactly when this is going to happen. It's slightly annoying, but nothing we can do about it.

Just out of curiosity, what is the reason these old versions of some libs are kept in a separate `dev-libs/...-compat` package?

trofi: To install `libffi.so.7` along with `libffi.so.8` we need a way to install them in parallel. For libraries that usually means that headers for both should not be installed into the same location to avoid conflicts. If build system does not make it easy to install everything into different subdirectories the easiest solution for binary-only packages is to install library only. libffi has a special headers story: https://bugs.gentoo.org/695788

Gentoo used to use non-0 SLOT for that, but that makes slots have very different meaning (present/absent headers). Now Gentoo uses `-compat` suffix instead to have clearer signal that it only provides a library for binary applications.

trofi: If the change does not look reasonable I can cut a bug against each maintainer and let them deal with it on package by package basis. There are many ways to do it depending on what you optimize:

- stable set of packages (current) vs unstable (altrnative)
- minimize amount of libffis installed (current) vs not (alternative)
- allow gradual stabilization of dependencies and `libffi`/`libff-compat` (current) or do it in a lockstep
- optimize for portage unmasking behaviour (i have no idea what portage does in this case and don't think we should optimize for such a low level detail)
- optimize for minimum amount of ebuild revbumps (alternative) or do another cleanup when libffi.so.7 will be provided only by `-compat` package (current)
- ${many other options unlisted}

PR #21813, opened 2021-07-27
games-emulation/dolphin: added optional libmgba support

xarblu:
Dolphin added integrated GBA input support via libmgba which would be enabled by default so emerging the package would fail because cmake couldn't find `Externals/mGBA`. Because the whole GBA input is optional I made it a new USE-flag.


trofi:
Is this pull request ready for review? I see it has a few commits that amend one another. Those should normally be squashed into one to be ready for inclusion.

A few notes on top of it:
1. https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=ec76e3439fd84be67bcd5b1e6591ccd24253b0d2 added missing mgba submodule to `-9999` ebuild today. If you think disabling this USE has some use for users we can proceed. It will need tiny rationaly why in commit message.
2. `IUSE` list should be alphabetically sorted.

Otherwise looks good.

xarblu:
Sorry, didn't know those things. 
For why I think that mgba should be a USE flag:
Only relatively few games support gba input ([only 63](https://wiki.dolphin-emu.org/index.php?title=Category:Game_Boy_Advance_(Input_supported))). From those 63 most don't really require it but have it more as a 'nice bonus' eg. for maps, bonus items or sending game-data from gba to gamecube ([as seen here](https://en.m.wikipedia.org/wiki/GameCube_%E2%80%93_Game_Boy_Advance_link_cable#Games)). That's why I believe most users wouldn't need this feature and if they did they could still enable it.

trofi:
Sounds good. If you update the pull request to make `mgba` optional i'll merge it as is.

xarblu:
I think I sort of messed things up in my fork of the repo. Will close this for now and open another pull request with all changes the way you described in an earlier message.

trofi:
Sounds good.

PR #21816, opened 2021-07-28
dev-java/zstd-jni: Allow CMake to find and use JNI without AWT

Leo3418:
More details can be found in the commit message.  The modified ebuilds are installable with `dev-java/openjdk-bin-8.292_p10dev-java/openjdk-bin-8.292_p10` with USE flag `headless-awt` enabled as well as disabled.

Tests: https://github.com/Leo3418/gentoo/actions/runs/1075517154


fordfrog:
@Leo3418 thanks for the fix! i also wondered why it should need awt when it's just about jni... :-)

PR #21818, opened 2021-07-28
x11-misc/slim: remove unused patch

mm1ke:


ionenwks (approved):
I'm amazed slim hasn't been last rited yet, so many patches and open bugs. I used to use it long ago but pretty done with that.

PR #21819, opened 2021-07-29
games-fps/freedoom-data: Requires games-util/deutex[png]

vilhelmgray:
This adds the `png` USE flag requirement to the `games-util/deutex` dependency for `games-fps/freedoom-data` and `games-fps/freedm-data`; the `games-util/deutex` is updated to default enable PNG support.

In addition, I noticed the `games-fps/freedoom` and `games-fps/freedm` ebuilds were missing `games-fps/chocolate-doom` as an optional source port, so I've updated them accordingly.

### DIFF
--- a/games-fps/freedm-data/freedm-data-0.12.1-r1.ebuild
+++ b/games-fps/freedm-data/freedm-data-0.12.1-r1.ebuild
@@ -18,7 +18,7 @@ KEYWORDS="~amd64 ~arm ~x86"
 BDEPEND="
 	$(python_gen_any_dep 'dev-python/pillow[${PYTHON_USEDEP},zlib]')
 	app-text/asciidoc
-	games-util/deutex"
+	games-util/deutex[png]"
 
 S="${WORKDIR}/freedoom-${PV}"
 
@@ -44,7 +44,7 @@ src_install() {
 	emake install-freedm \
 		prefix="${ED}/usr/" \
 		bindir="bin/" \
-		docdir="share/doc/${P}" \
+		docdir="share/doc/${PF}" \
 		mandir="share/man/" \
 		waddir="${DOOMWADPATH}/"
 }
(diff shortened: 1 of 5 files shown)


ionenwks:
The bug mentions that it still built/installed without it, meaning there could be people with a broken install.

Normally do a revbump to force reinstall in these cases.

vilhelmgray:
Good point, I'll update these to be revbumps.

games-fps/freedm-data/freedm-data-0.12.1-r1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+	emake wads/freedm.wad \
+		freedm.6 \
+		{NEWS,README}.html
+}
+
+src_install() {
+	emake install-freedm \
+		prefix="${ED}/usr/" \
+		bindir="bin/" \
+		docdir="share/doc/${P}" \
```
ionenwks, line 47: Normally should use `${PF}` rather than `${P}` for doc dirs.

With the new revision, leads to:
```
 * 
 *   /usr/share/doc/freedm-data-0.12.1
```

games-fps/freedoom-data/freedoom-data-0.12.1-r1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+	emake wads/freedoom{1,2}.wad \
+		freedoom{1,2}.6 \
+		{NEWS,README}.html
+}
+
+src_install() {
+	emake install-freedoom \
+		prefix="${ED}/usr/" \
+		bindir="bin/" \
+		docdir="share/doc/${P}" \
```
ionenwks, line 47: here too

ionenwks (commented):
This exposed a small issue.

Also, I assume don't mean to keep the old revision? Small fix like this is okay to drop in same commit (aka rename / git mv).

ionenwks (approved):

PR #21820, opened 2021-07-29
Mbedtls 3

juippis:
See the linked bug for description.

### DIFF
--- a/net-libs/mbedtls/mbedtls-3.0.0-r1.ebuild
+++ b/net-libs/mbedtls/mbedtls-3.0.0-r1.ebuild
@@ -11,7 +11,7 @@ SRC_URI="https://github.com/ARMmbed/mbedtls/archive/${P}.tar.gz"
 S=${WORKDIR}/${PN}-${P}
 
 LICENSE="Apache-2.0"
-SLOT="0/6.13.1" # ffmpeg subslot naming: SONAME tuple of {libmbedcrypto.so,libmbedtls.so,libmbedx509.so}
+SLOT="3/10.16.4" # ffmpeg subslot naming: SONAME tuple of {libmbedcrypto.so,libmbedtls.so,libmbedx509.so}
 KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86"
 IUSE="cpu_flags_x86_sse2 doc programs static-libs test threads"
 RESTRICT="!test? ( test )"
--- a/app-crypt/tpm2-tss/tpm2-tss-3.1.0-r1.ebuild
+++ b/app-crypt/tpm2-tss/tpm2-tss-3.1.0-r1.ebuild
@@ -23,7 +23,7 @@ RDEPEND="acct-group/tss
 	acct-user/tss
 	fapi? ( dev-libs/json-c
 		net-misc/curl )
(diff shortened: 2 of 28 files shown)


juippis:
@blueness 
@ionenwks 
@gentoo/qa

net-libs/mbedtls/mbedtls-3.0.0-r1.ebuild
```diff
@@ -11,7 +11,7 @@ SRC_URI="https://github.com/ARMmbed/mbedtls/archive/${P}.tar.gz"
 S=${WORKDIR}/${PN}-${P}
 
 LICENSE="Apache-2.0"
-SLOT="0/6.13.1" # ffmpeg subslot naming: SONAME tuple of {libmbedcrypto.so,libmbedtls.so,libmbedx509.so}
+SLOT="3/10.16.4" # ffmpeg subslot naming: SONAME tuple of {libmbedcrypto.so,libmbedtls.so,libmbedx509.so}
```
ulm, line 14: Have you actually tested this? Because I think that slotting won't work for this package. You'll get file collisions between different slots, e.g. for several of the header files in `/usr/include/mbedtls/`.

Files also cannot easily be move to another (not colliding) directory, because the package doesn't support pkgconfig.

juippis: Oh yea, forgot to add the blocker against :0. I also don't think these were meant to co-exist.

ionenwks: If slots block each others I don't think there's much meaning to slots :)

I did hope it would be (easily) slottable, but indeed I see nothing in this to discover paths so it'd mean needing to invent our own thing. Guess should just wait to see how this goes, and how upstreams are going to handle this. I'd hope the 2.x branch will be supported for some time still while keep this masked.

juippis: So the other alternative I see here, is to revbump the rdeps and make them depend on `<3.0.0`. I personally find the slotting method a bit cleaner overall, but should the rdeps be fixed like this instead?

SoapGentoo: this is pretty much equivalent to the openssl-1.1 situation, which we worked around with a very long `p.mask` and slowly transitioned consumers.

ulm: Slots with blocks between each other are useless. It's even worse than without slotting because it is confusing for users.

juippis: So what's the suggested course of action here? Identify every rdep, bump them while adding to package.mask one-by-one, and unmask... but unmask when?

blueness: @SoapGentoo pretty much said it.  This is going to be a long p.mask while consumers are updated.  We keep the mask on anything >= 3.0.0 so we can bump and test new versions in the 3.x.y branch against its rdeps.  You lift the mask when most consumers build against 3.x.y and you make sure the *DEPEND list for those packages have the right subslots.

juippis: Heh, sure, I was perhaps thinking more along the short-term solutions. I still do find my own suggestions reasonable, too, going forward.

But as this is contained now, we can take it slowly from here.

PR #21822, opened 2021-07-29
app-text/sigil: bump v1.7.0

arthurzam:
@mgorny A small version bump to my package

bump the version, and drop old stable version (1.5.0 remains as stable)
Version 1.7.0 works quite well, so I think we can push it to out users.

Note to myself: upstream might change python executable argument handling in CMake, so I might need to fix the passing of out selected single python.


app-text/sigil/sigil-1.7.0.ebuild
```diff
@@ -0,0 +1,82 @@
...
+		dev-python/pillow[${PYTHON_USEDEP}]
+		dev-python/regex[${PYTHON_USEDEP}]') )
+	system-mathjax? ( dev-libs/mathjax )
+"
+DEPEND="${RDEPEND}"
+
+BDEPEND="
+	dev-qt/linguist-tools:5
+	virtual/pkgconfig
+"
```
ionenwks: `${PYTHON_DEPS}` should be in BDEPEND too, don't know if build system needs it but the eclass runs python at build time and expects you to handle adding deps.

app-text/sigil/sigil-1.7.0.ebuild
```diff
@@ -0,0 +1,82 @@
...
+		dev-python/html5lib[${PYTHON_USEDEP}]
+		dev-python/pillow[${PYTHON_USEDEP}]
+		dev-python/regex[${PYTHON_USEDEP}]') )
+	system-mathjax? ( dev-libs/mathjax )
+"
+DEPEND="${RDEPEND}"
+
+BDEPEND="
+	dev-qt/linguist-tools:5
+	virtual/pkgconfig
```
ionenwks: Unimportant but can technically skip pkgconfig with cmake, it's a RDEPEND of cmake and not something you need to worry about (unless you are calling pkg-config yourself, like in the ebuild or so -- just to be proper given it wouldn't be cmake using it).

You'd need this when it's a ./configure using it though.

ionenwks (commented):
Seems to run fine, but maybe time for some small ebuild improvements while here?

arthurzam:
> Seems to run fine, but maybe time for some small ebuild improvements while here?

Thanks for the review - it was indeed time to do major cleanup.

app-text/sigil/sigil-1.7.0.ebuild
```diff
@@ -0,0 +1,82 @@
...
+"
+
+S="${WORKDIR}/Sigil-${PV}"
```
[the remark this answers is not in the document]
arthurzam: Thanks for this trick - quite useful!

ionenwks (approved):
Thanks, looks great.

PR #21823, opened 2021-07-29
dev-java/jansi: Version bump to 1.13

Leo3418:
The original slot is preserved for tests of all reverse dependencies are still [passing](https://github.com/Leo3418/gentoo/runs/3194845570?check_suite_focus=true).

Some notable changes:
- This ebuild does not use Ant; it uses `java-pkg-simple.eclass` instead.
- `HOMEPAGE` is updated to an accessible HTTPS URL.


fordfrog:
@Leo3418 merged, thanks!

PR #21826, opened 2021-07-29
media-gfx/hydrus: Add 448, drop 444 and now-unused patches

vaartis:
No changes to ebuild, just a new version again. The patches are not needed now so they have been dropped.

@thesamesam


media-gfx/hydrus/hydrus-448.ebuild
```diff
@@ -0,0 +1,141 @@
...
+	SRC_URI="https://github.com/hydrusnetwork/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+	KEYWORDS="~amd64"
+fi
+
+# hydrus itself is WTFPL
+# icons included are CC-BY-2.5
+LICENSE="WTFPL-2 CC-BY-2.5"
+SLOT="0"
+IUSE="+mpv +ffmpeg +lz4 socks +cloudscraper charts test"
```
juippis: This block could be sorted alphabetically. Besides, you have many use flags with no function in the ebuild, you should put these into `optfeature`, 

media-gfx/hydrus/hydrus-448.ebuild
```diff
@@ -0,0 +1,141 @@
...
+		mpv? (
+			media-video/mpv[libmpv,${PYTHON_USEDEP}]
+			dev-python/python-mpv[${PYTHON_USEDEP}]
+		)
+		socks? (
+			|| ( dev-python/requests[socks5,${PYTHON_USEDEP}]
+				dev-python/PySocks[${PYTHON_USEDEP}] )
+		)
+	')
+"
```
juippis: How is this block sorted? Could you add a descriptive comment about it on top of RDEPEND?

media-gfx/hydrus/hydrus-448.ebuild
```diff
@@ -0,0 +1,141 @@
...
+	rm -r "${DOCS[@]}" "${HTML_DOCS[@]}" || die
+	# The program expects to find documentation here, so add a symlink to doc
+	ln -s "${doc}/html/help" help || die
+
+	insinto /opt/hydrus
+	doins -r "${S}"/.
+
+	exeinto /usr/bin
+	python_newexe - hydrus-server < <(sed "s/python/${EPYTHON}/" "${FILESDIR}/hydrus-server")
+	python_newexe - hydrus-client < <(sed "s/python/${EPYTHON}/" "${FILESDIR}/hydrus-client")
```
juippis:
```suggestion
	python_newexe - hydrus-server < <(sed "s/python/${EPYTHON}/" "${FILESDIR}/hydrus-server" || die)
	python_newexe - hydrus-client < <(sed "s/python/${EPYTHON}/" "${FILESDIR}/hydrus-client" || die)

File: media-gfx/hydrus/hydrus-448.ebuild [hunk 4]
```diff
@@ -0,0 +1,141 @@
...
+	einstalldocs
+
+	# Files only needed for testing
+	rm test.py hydrus/hydrus_test.py || die
+	rm -r hydrus/test/ static/testing/ || die
+
+	# These files are copied into doc
+	rm -r "${DOCS[@]}" "${HTML_DOCS[@]}" || die
+	# The program expects to find documentation here, so add a symlink to doc
+	ln -s "${doc}/html/help" help || die
```
juippis: `dosym`?

media-gfx/hydrus/hydrus-448.ebuild
vaartis: This was discussed on IRC before, I removed one of use flags that was really optional and probably unpopular but these are more of things that should probably be installed by default but can be disabled. Besides it's a package that doesn't need to compile anything, so re-installing is quickly. Those marked with + certainly should be enabled by default, those without seem like useful features but require multiple packages with certain use flags to be installed, which is somewhat inconventient.

media-gfx/hydrus/hydrus-448.ebuild
vaartis: They're grouped together by type (no specific requirement, slot/version requirement, specfic use flags) and then sorted, I'll add some comments.

media-gfx/hydrus/hydrus-448.ebuild
juippis: NNope, the policy does not have exclusion list for this reason. In fact, all these do are just be deceiful to the users. What if I do want to disable a feature, but have the dep installed? Toggling the USE flag does not help. At least optfeature doesn't lie about it. It was built for cases like this, so it should be used in cases like this.

I'll be updating the ebuilds at some point migrating them to `optfeature`. Can I clean out some older ebuilds while at it?

vaartis: I see. In case you don't have the time until the next week, I'll do it myself on next release. I try to keep 4 ebuilds in the tree because they come out ~every week or so, so it's just about a month of updates. I could scale it down of course. The idea was to have some versions stable, but the dependencies are still unstable so can't put anything in stable still.

juippis: I'd much rather have you agree with the decision and update the ebuild. We try to be consistent, and this is one rule that has allowed us to clean numerous ebuilds before. 

I don't know whom you've talked with in IRC, but if the QA policy should allow "special cases" for python and perl, it should clearly state it. The rules should be same for everyone. But I doubt such exception can be made, since solution like this fights against the whole principle of Gentoo by giving an illusion you have a choice. Toggling USE flags should have a working function, both ways.

PR #21828, opened 2021-07-29
dev-python/pipenv: remove vendored jinja2 and attr.

oz123:
Since pipenv bundles tons of packages, these will be
removed in a slow pace. If no bugs are reported, the
removal of bundled packages and addition of dependencies
can continue.

Bug: https://bugs.gentoo.org/717666

### DIFF
--- a/dev-python/pipenv/pipenv-2021.5.29-r1.ebuild
+++ b/dev-python/pipenv/pipenv-2021.5.29-r1.ebuild
@@ -1,0 +1,54 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DISTUTILS_USE_SETUPTOOLS=rdepend
+PYTHON_COMPAT=( python3_{7,8,9} )
+
+inherit distutils-r1
+
+MY_PV=${PV/_beta/b}
+DESCRIPTION="Python Development Workflow for Humans"
+HOMEPAGE="https://github.com/pypa/pipenv https://pypi.org/project/pipenv/"
+SRC_URI="https://github.com/pypa/pipenv/archive/v${MY_PV}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}"/${PN}-${MY_PV}
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="test"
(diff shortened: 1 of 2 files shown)


dev-python/pipenv/pipenv-2021.5.29-r1.ebuild
```diff
@@ -0,0 +1,57 @@
...
+src_prepare() {
+	# remove vendored version of PyYAML that is backported to Python2
+	rm -vR "${S}/${PN}/patched/yaml2/" || die
+	# remove vendored versions
+	# see https://bugs.gentoo.org/717666
+	rm -vR "${S}/${PN}/vendor/attr/" || die
+	rm -vR "${S}/${PN}/vendor/jinja2/" || die
+	# disable running of unittests in parallel with xdist
+	sed -i 's/addopts = -ra -n auto/;&/g' setup.cfg || die
+	sed -i 's/plugins = xdist/;&/g' setup.cfg || die
```
juippis, line 51:
```
 * QA: sed did not cause any changes
 * (sed -i s/addopts = -ra -n auto/;&/g setup.cfg)
 * QA: sed did not cause any changes
 * (sed -i s/plugins = xdist/;&/g setup.cfg)
```
could you verify these seds work the way you intended?

juippis (approved):
Rest LGTM.

dev-python/pipenv/pipenv-2021.5.29-r1.ebuild
oz123: Well, upstream remove those also, so it's not needed for this version.
How do I set up my system, so I see this QA warning? I don't see it when I install with emerge or directly with ebuild.

ionenwks: Preferably it's good to always audit seds on bumps for your own ebuilds, if you'd rather not then use patches instead since they'll more normally fail once irrelevant.

As for the "QA" there's currently no "normal" way to see this, it's a dirty helper script I made that use /etc/portage/bashrc to compare files before and after. Haven't really worked on it much nor properly distribute it anywhere -- it also print a lot of noise and not really intended for normal use (but we check a lot of ebuilds from many different people and it'd be a lot to verify every sed manually :)

oz123: OK. Thanks for that.

juippis (approved):

PR #21829, opened 2021-07-29
games-emulation/dolphin: added USE=mgba

xarblu:
Here's the new pull request I mentioned [here](https://github.com/gentoo/gentoo/pull/21813#issuecomment-889357466) for making mgba optional. 

The reasons are the same as in the previous pull request:
Only relatively few games support gba input ([only 63](https://wiki.dolphin-emu.org/index.php?title=Category:Game_Boy_Advance_(Input_supported))). From those 63 most don't really require it but have it more as a 'nice bonus' eg. for maps, bonus items or sending game-data from gba to gamecube ([as seen here](https://en.m.wikipedia.org/wiki/GameCube_%E2%80%93_Game_Boy_Advance_link_cable#Games)). That's why I believe most users wouldn't need this feature and if they did they could still enable it.

### DIFF
--- a/games-emulation/dolphin/dolphin-9999.ebuild
+++ b/games-emulation/dolphin/dolphin-9999.ebuild
@@ -25,7 +25,7 @@ HOMEPAGE="https://www.dolphin-emu.org/"
 
 LICENSE="GPL-2"
 SLOT="0"
-IUSE="alsa bluetooth discord-presence doc +evdev ffmpeg log lto profile pulseaudio +qt5 systemd upnp vulkan"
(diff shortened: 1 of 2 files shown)


trofi:
Pulled. Thank you!

A few notes for next time:
1. Commit rationale should be put into commit message description itself, not the pull request description: Gentoo applies individual commits and does not merge requests. Thus merge request descriptions get lost.
2. Each commit should should start from `cat/pkg: ...`
3. Each commit should have your `Signed-off-by: real name` to trace the contribution.

https://www.gentoo.org/glep/glep-0066.html and https://www.gentoo.org/glep/glep-0076.html have a bit more detail.

PR #21834, opened 2021-07-30
dev-python/*: enable pypy3 support for dev-python/async_timeout

arthurzam:
@mgorny Here are my results trying to add `pypy3` support to `dev-python/async_timeout`. Just to summarize, it was a fun ride.

I will start with `dev-python/regex`, which needed a lot of work. As you noted in the bug you reported [here](https://bitbucket.org/mrabarnett/mrab-regex/issues/404/), there are problems with PyPy. I needed to view the internal implementation of CPython and PyPy3 of array module to understand and fix.
I have managed to create two patches (they are inside the `files/` dir), a fix for every test failure. The first one, changes the C code only for PyPy3 (explanation in the patch header), and the second one relaxes a little the python test for correct type.
In the worst case that my patches are incorrect, they implicit only PyPy, so there shouldn't be regressions for CPython.
**Please do extra review to the first commit**, just to be sure.

Then a trivial bump to `dev-python/re-assert` and test pass.
Then bumped for `app-arch/brotli`. The testsuite froze multiple times (which I guess why it is restricted), so I decided to check it manually using some examples and the next testsuite of aiohttp.

Then bump the circular dependency of `dev-python/async_timeout`, `dev-python/pytest-aiohttp` and `dev-python/aiohttp`.
For `aiohttp` I needed to disable one test only for `pypy3` which fails because of an unawaited coroutine and failing with Runtime Warning. I think it happens because of some raise expect, but it was too hard to understand the error log from pytest. I think just skipping this test is fine, but it is your decision.

After that, I just bumped the reverse dependencies of the circle, and bumped to `pypy3` support. At one point I decided to stop so the PR isn't to big for a review.


dev-python/pytest-aiohttp/pytest-aiohttp-0.3.0-r1.ebuild
```diff
@@ -14,7 +14,6 @@ SRC_URI="mirror://pypi/${P:0:1}/${PN}/${P}.tar.gz"
 LICENSE="Apache-2.0"
 SLOT="0"
 KEYWORDS="~alpha amd64 arm arm64 hppa ~ia64 ppc ppc64 ~riscv sparc x86 ~x64-macos"
-IUSE=""
 
 RDEPEND="
 	>=dev-python/pytest-5.4[${PYTHON_USEDEP}]
```
thesamesam, line 19: Is the indentation for the below line wrong, or is it just the editor?

Also, no tests for this?

dev-python/regex/files/regex-2021.4.4-pypy3-fix-test_empty_array.patch
```diff
@@ -0,0 +1,35 @@
+From: Arthur Zamarin <arthurzam@gmail.com>
```
thesamesam, line 1: Upstreamable? If so, pop a link in.

Good work, by the way!

thesamesam (changes requested):
Some minor comments. BTW, are you still checking if we can add ALLARCHES as you go? (Not saying any candidates are in this PR).

dev-python/pytest-aiohttp/pytest-aiohttp-0.3.0-r1.ebuild
arthurzam: About indentation, I even checked using `hexdump -C` and it uses `\t` - so I guess it is Ok?

About tests - this is a very trivial package that adds `aiohttp`'s pytest support as a first class. Except of an import is does nothing.
The testsuite is tested in `aiohttp` package.

dev-python/regex/files/regex-2021.4.4-pypy3-fix-test_empty_array.patch
arthurzam: Well, I posted the PR [here](https://bitbucket.org/mrabarnett/mrab-regex/pull-requests/5), lets see their response...

mgorny (approved):
Great work, thanks!

PR #21836, opened 2021-07-30
dev-lang/nasm: Adopt package

00-matt:
Also update upstream remote, bump to EAPI 8, and drop unused eclass.

Signed-off-by: Matt Smith <matt@offtopica.uk>

### DIFF
--- a/dev-lang/nasm/nasm-2.15.05.ebuild
+++ b/dev-lang/nasm/nasm-2.15.05.ebuild
@@ -1,21 +1,18 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
-inherit flag-o-matic
-
 DESCRIPTION="groovy little assembler"
 HOMEPAGE="https://www.nasm.us/"
 SRC_URI="https://www.nasm.us/pub/nasm/releasebuilds/${PV/_}/${P/_}.tar.xz"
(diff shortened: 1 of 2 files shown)


dev-lang/nasm/nasm-2.15.05.ebuild
```diff
@@ -1,9 +1,7 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
```
DarthGandalf, line 1: 1. If RDEPEND/DEPEND are empty, you can drop them.
2. S= needs quotes; and please move it closer to SRC_URI
3. Perhaps install needs ${ED} instead of ${D}?

00-matt: Thanks for the suggestions!

I should probably write a pre-commit hook to check the order of everything because I always seem to forget...

dev-lang/nasm/nasm-2.15.05.ebuild
```diff
@@ -1,21 +1,18 @@
...
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
-
-inherit flag-o-matic
+EAPI=8
```
juippis, line 4: Generally we revbump the ebuild when bumping EAPI, but I don't see a need to do this.

00-matt: I've reverted the revbump as discussed on irc :)

juippis (approved):
Thanks!

PR #21838, opened 2021-07-30

arthurzam:
@mgorny Another **low-priority** batch. It might look big, but it is only 8 packages.

On the way, did a small cleanup for some of them (mainly handling `--deselect` for pytest).
Also marked `ALLARCHES` (rechecked again that there is just python bytecode compiling and no real code compiling).
Bumped to `EAPI=8` (those are marked in the commit title - I like being informative).
Of course, ran test on cleaned container and all passed.

Just for `dev-python/leather` needed to port patch from upstream commit.


mgorny (approved):
Great work!

PR #21843, opened 2021-07-31
media-plugins/gst-plugins-jack: Fix jack2 dependency

lanodan:
Closes: https://bugs.gentoo.org/804957

### DIFF
--- a/media-plugins/gst-plugins-jack/gst-plugins-jack-1.18.4-r1.ebuild
+++ b/media-plugins/gst-plugins-jack/gst-plugins-jack-1.18.4-r1.ebuild
@@ -7,8 +7,12 @@ GST_ORG_MODULE=gst-plugins-good
 inherit gstreamer-meson
 
 DESCRIPION="JACK audio server source/sink plugin for GStreamer"
-KEYWORDS="~amd64 ~arm ~arm64 ~ia64 ~ppc ~ppc64 ~riscv ~sparc ~x86"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~riscv ~x86"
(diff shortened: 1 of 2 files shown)


media-plugins/gst-plugins-jack/gst-plugins-jack-1.18.4-r1.ebuild
```diff
@@ -10,5 +10,5 @@ DESCRIPION="JACK audio server source/sink plugin for GStreamer"
 KEYWORDS="~amd64 ~arm ~arm64 ~ia64 ~ppc ~ppc64 ~sparc ~x86"
 IUSE=""
 
-RDEPEND="virtual/jack[${MULTILIB_USEDEP}]"
+RDEPEND="media-sound/jack2[${MULTILIB_USEDEP}]"
```
pinkflames, line 13: This would exclude media-video/pipewire[jack-sdk] which has the same headers as JACK2 and while not all functions are implemented, is likely to work, too. Therefore please consider using this instead:

        || (
                media-sound/jack2[${MULTILIB_USEDEP}]
                media-video/pipewire[${MULTILIB_USEDEP},jack-sdk(-)]
        )

pinkflames (changes requested):

pinkflames:
BTW, I noticed that someone on the linked bugs.gentoo.org report has added a patch for the build system - for better or worse it seems that the Gentoo preferred method is to fix the build system version check if the project itself builds fine against JACK1. Not that I think it's a good idea to keep dragging JACK1 around but who am I to decide that?

pinkflames:
Thanks for fixing PipeWire support. ;)

PR #21847, opened 2021-07-31
dev-util/shards: bump to EAPI=8, new maintainer

toshokan:


juippis (commented):
Same question here. We don't usually let people take over maintenance without prior contributions to a package, and EAPI-7 -> 8 bump doesn't seem to be doing anything here. 

You can still do version bumps without becoming the maintainer.

PR #21848, opened 2021-07-31
dev-lang/crystal: bump to 1.1.1, new maintainer

toshokan:


dev-lang/crystal/crystal-1.1.1.ebuild
```diff
@@ -0,0 +1,123 @@
...
+	dev-libs/libatomic_ops
+	dev-libs/libevent
+	dev-libs/libpcre
+	sys-libs/libunwind
+	dev-libs/pcl
+	dev-libs/gmp:0
+"
+RDEPEND="${DEPEND}
+	xml? ( dev-libs/libxml2 )
+	yaml? ( dev-libs/libyaml )
```
juippis, line 39: These don't seem to be having any function in this ebuild, 

if they're somehow automatically detected by the build system, that behaviour should be patched out.

dev-lang/crystal/crystal-1.1.1.ebuild
```diff
@@ -0,0 +1,123 @@
...
+
+DEPEND="
+	sys-devel/llvm:${LLVM_MAX_SLOT}
+	dev-libs/boehm-gc[static-libs,threads]
+	dev-libs/libatomic_ops
+	dev-libs/libevent
+	dev-libs/libpcre
+	sys-libs/libunwind
+	dev-libs/pcl
+	dev-libs/gmp:0
```
juippis, line 35: Please sort this block alphabetically for readability.

dev-lang/crystal/crystal-1.1.1.ebuild
```diff
@@ -0,0 +1,123 @@
...
+	x86? ( https://github.com/crystal-lang/crystal/releases/download/${BV/-*}/crystal-${BV_X86}.tar.gz )"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="doc debug examples blocking-stdio-hack +xml +yaml"
+
+RESTRICT=test # not stable for day-to-day runs
+
+LLVM_MAX_SLOT=11
```
juippis, line 25: Is it somehow confirmed 12 won't work?

dev-lang/crystal/crystal-1.1.1.ebuild
```diff
@@ -0,0 +1,123 @@
...
+SRC_URI="https://github.com/crystal-lang/crystal/archive/${PV}.tar.gz -> ${P}.tar.gz
+	amd64? ( https://github.com/crystal-lang/crystal/releases/download/${BV/-*}/crystal-${BV_AMD64}.tar.gz )
+	x86? ( https://github.com/crystal-lang/crystal/releases/download/${BV/-*}/crystal-${BV_X86}.tar.gz )"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="doc debug examples blocking-stdio-hack +xml +yaml"
+
+RESTRICT=test # not stable for day-to-day runs
```
juippis, line 23: What exactly does this mean? Could you attempt to get tests to work?

dev-lang/crystal/crystal-1.1.1.ebuild
```diff
@@ -0,0 +1,123 @@
...
+DESCRIPTION="The Crystal Programming Language"
+HOMEPAGE="https://crystal-lang.org"
+SRC_URI="https://github.com/crystal-lang/crystal/archive/${PV}.tar.gz -> ${P}.tar.gz
+	amd64? ( https://github.com/crystal-lang/crystal/releases/download/${BV/-*}/crystal-${BV_AMD64}.tar.gz )
+	x86? ( https://github.com/crystal-lang/crystal/releases/download/${BV/-*}/crystal-${BV_X86}.tar.gz )"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="doc debug examples blocking-stdio-hack +xml +yaml"
```
juippis, line 21: Also please sort this alphabetically. What does `blocking-stdio-hack` do...

dev-lang/crystal/crystal-1.1.1.ebuild
```diff
@@ -0,0 +1,123 @@
...
+
+	# crystal uses 'LLVM_TARGETS' to override default list of targets
+	unset LLVM_TARGETS
+	emake \
+		$(usex debug "" release=1) \
+		progress=true \
+		stats=1 \
+		threads=$(makeopts_jobs) \
+		verbose=1 \
+		\
```
thesamesam, line 72: Please use a Bash array for this rather than using \ as it doesn't help readability, e.g. [Botan](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/botan/botan-2.18.1-r1.ebuild#n104).

thesamesam: (Same for src_test, may want to consider putting the emake with all the boilerplate into its own function).

dev-lang/crystal/crystal-1.1.1.ebuild
thesamesam, line 35: Please subscribe with := to be rebuilt if subslot of dependencies change if they have one (gmp does, I think other packages do here too) and if we're linked against it.

Does Crystal link against e.g. gmp?

thesamesam (changes requested):

dev-lang/crystal/crystal-1.1.1.ebuild
ionenwks: Alternatively to a function could share the array for what's in common, a non-local array defined in src_configure will be available in src_test (on side-note, for non-local variables usually use uppercase names for distinction).

dev-lang/crystal/crystal-1.1.1.ebuild
toshokan: They're mandatory dependencies
https://github.com/crystal-lang/crystal/blob/55ebfbfe75e93ffdc9373acf649a473b00b24526/src/yaml/lib_yaml.cr#L4
https://github.com/crystal-lang/crystal/blob/55ebfbfe75e93ffdc9373acf649a473b00b24526/src/xml/libxml2.cr#L8

dev-lang/crystal/crystal-1.1.1.ebuild
toshokan: Yes 
https://github.com/crystal-lang/crystal/blob/55ebfbfe75e93ffdc9373acf649a473b00b24526/src/big/lib_gmp.cr#L1

dev-lang/crystal/crystal-1.1.1.ebuild
toshokan: Yes crystal-lang/crystal#10434
I'll add a comment to the ebuild

dev-lang/crystal/crystal-1.1.1.ebuild
toshokan: The upstream test suite isn't very reliable (their own CI fails often)
Some tests also make assumptions about the permissions of the user running the test suite or capability to use the network, but aren't tagged to allow them to be disabled. I'll see what can be done to resolve this upstream.

dev-lang/crystal/crystal-1.1.1.ebuild
toshokan: No longer needed, apparently :)
It was a patch to work around a bug in 0.22 that has been fixed

toshokan:
See report above, I'm not sure how best to handle dropping the `yaml` USE flag in the `dev-util/shards` ebuild.
Should I depend on either `crystal[yaml]` or `crystal >= 1.1.1`?

ionenwks:
> See report above, I'm not sure how best to handle dropping the `yaml` USE flag here in the `dev-util/shards` ebuild.
> (I'm dropping the USE flag because it really is not optional)
> Should I depend on either `crystal[yaml]` or `crystal >= 1.1.1`?

Haven't looked in depth at what's being done here, but crystal[yaml(+)] would mean "if flag is missing, assume it's enabled"
https://devmanual.gentoo.org/general-concepts/dependencies/#use-dependency-defaults

toshokan:
We will likely be able to remove the `RESTRICT=test` line soon, see crystal-lang/crystal#11048
More tests might need to be tagged first, but progress is being made!

dev-lang/crystal/crystal-1.1.1.ebuild
```diff
@@ -0,0 +1,107 @@
...
+
+src_install() {
+	insinto /usr/$(get_libdir)/crystal
+	doins -r src/.
+	dobin .build/crystal
+
+	insinto /usr/share/zsh/site-functions
+	newins etc/completion.zsh _crystal
+
+	use examples && dodoc -r samples
```
juippis, line 99: This dir seems to be worth 300-400 kb, so always install it unconditionally.

juippis (commented):
Also a bug to fix in shards ebuild!
```
 * /usr/bin/shards
```

dev-util/shards/shards-0.15.0.ebuild
```diff
@@ -20,7 +20,7 @@ SLOT="0"
 KEYWORDS="~amd64"
 
 DEPEND="
-	>dev-lang/crystal-0.11.1[yaml]
+	>dev-lang/crystal-0.11.1[yaml(+)]
```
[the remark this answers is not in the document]
toshokan: revbump, ok;
I'm not sure adding one of crystal's dependencies to the shards ebuild is the best way forward. 
It wouldn't be very clear what that dependency is doing, since shards itself isn't directly using or linking to libyaml, only using it through crystal's interface.
This dependency should communicate "crystal with the ability to use the yaml feature"

dev-lang/crystal/crystal-1.1.1.ebuild
toshokan: Do you recommend dropping the `examples` flag entirely?

dev-util/shards/shards-0.15.0.ebuild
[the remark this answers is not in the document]
juippis: No matter how you look at it, the dep is then 
```
	dev-lang/crystal
	dev-libs/libyaml
```
and that's correct.

dev-lang/crystal/crystal-1.1.1.ebuild
juippis: Yes, that's what I'd recommend.

toshokan:
@juippis Do I drop dev-lang/shards-0.15 now that I've revbumped to a version that depends on libyaml?

juippis:
Yep you can. I guess `git mv` is fine here, but can do a separate commit too if you don't want to rebase.

PR #21849, opened 2021-07-31
dev-lang/elixir: bump to EAPI=8

toshokan:


juippis (commented):
Hey, 

EAPI-7 is still perfectly fine. Does EAPI-8 bring something you need for this ebuild? I don't see anything changing.

You should instead perhaps focus on fixing existing bugs, stabilize 1.12.2,...

PR #21854, opened 2021-08-01
dev-python/PyQt5: updated build system to sip-build bug #805965

chriscpritchard:
updated the build system used by dev-python/PyQt5 to sip-build instead
of configure.py. This necessitated changing of some build time options.

Signed-off-by: Chris Pritchard <chris@christopherpritchard.co.uk>
Closes: https://bugs.gentoo.org/805965


chriscpritchard:
A few issues with this:

1. the `ssl` use flag now does nothing - it seems the option to build without SSL was removed, but I don't have the capacity to change all the packages that depend on this package built with the SSL use flag - not sure of the best way to deprecate / remove this flag without breaking other packages
2. I've not tested the `gles2-only` use flag and I had to make a change to how this was handled as my system isn't set up with this
3. sip-build doesn't appear to support python 3.10 at the moment so I've removed it from the supported python versions

On the plus side it seems like we no longer need to `emake -j1` as it installs fine with `-j8` in my makeopts and `-j1` removed.

Pesa:
Copying my comment from the linked bug:

I've actually been (slowly) working on this for several weeks in a local branch. I have a working ebuild for PyQt5 but still haven't found much time to test the reverse deps. So I propose the following: I will go ahead and commit my ebuild (masked) and then you can submit any changes/improvements you may have by rebasing your PR. Also, testing of reverse deps would be greatly appreciated.

chriscpritchard:
That sounds perfect

Pesa:
Pushed

chriscpritchard:
I've added a 5.15.4-r2 ebuild based on yours and added the latest released sip version (which works with python below 3.10, snapshot's needed for python 3.10). I've also amended PyQtWebEngine to use the new build system. If I get a chance tomorrow I'll look at other reverse deps.

Pesa:
Thanks. I'm not sure I fully understand certain pieces of this PR though, see questions below.

> I've added a 5.15.4-r2 ebuild based on yours

Why? given that an `-r2` would also need to be masked, why don't you just use `5.15.5_pre*`?

> and added the latest released sip version (which works with python below 3.10, snapshot's needed for python 3.10).

I don't see the point. Just use `6.2.0_pre*` which does support py3.10, it simplifies dependencies a lot.

> I've also amended PyQtWebEngine to use the new build system.

That's good, PyQtWebEngine was the last major piece of the puzzle. I'll take a closer look at the ebuild tomorrow.

chriscpritchard:
Honestly, I added them (and also updated package.mask) because I wasn't sure that people would necessarily want to install pre-released versions if they do become unmasked.

That said, you make some good points so can easily remove them, if that's what you'd prefer.

Pesa:
I merged your PyQtWebEngine commit (with simplified deps) and bumped it to a 5.15.5_pre snapshot. I'm NAK'ing the rest as it seems superfluous to me.

Thanks!

PR #21855, opened 2021-08-01
sci-libs/vtk: fix auto-detection of CUDA arch

waebbl:
In some cases, the CUDA architecture is not successfully detected, when
'native' is passed to the build system. For this, a function get_cuda_arch
is defined, which checks for the environment variable CUDA_ARCH to be
exported and passes that value from the environment instead, if it's a
valid CUDA architecture name.
The function ignores the 'fermi' architecture, as it's no longer supported
by current dev-util/nvidia-cuda-toolkit-11* versions.

Closes: https://bugs.gentoo.org/803818

### DIFF
--- a/sci-libs/vtk/vtk-9.0.1.ebuild
+++ b/sci-libs/vtk/vtk-9.0.1.ebuild
@@ -41,7 +41,7 @@ RESTRICT="!test? ( test )"
 
 REQUIRED_USE="
 	all-modules? ( boost ffmpeg gdal imaging mysql odbc postgres qt5 rendering theora views )
-	cuda? ( X )
+	cuda? ( X video_cards_nvidia )
 	java? ( rendering )
 	python? ( ${PYTHON_REQUIRED_USE} )
 	qt5? ( X rendering )
@@ -294,7 +294,28 @@ src_configure() {
 	fi
 
 	if use cuda; then
-		mycmakeargs+=( -DVTKm_CUDA_Architecture="native" )
+		local cuda_arch=
+		case ${VTK_CUDA_ARCH:-native} in
+			# we ignore fermi arch, because current nvidia-cuda-toolkit-11*
+			# no longer supports it
+			kepler|maxwell|pascal|volta|turing|ampere|all)
+				cuda_arch=${VTK_CUDA_ARCH}
+				;;
+			native)
+				ewarn "If auto detection fails for you, please try and export the"
+				ewarn "VTK_CUDA_ARCH environment variable to one of the common arch"
+				ewarn "names: kepler, maxwell, pascal, volta, turing, ampere or all."
+				cuda_arch=native
+				;;
+			*)
+				eerror "Please properly set the VTK_CUDA_ARCH environment variable to"
+				eerror "one of: kepler, maxwell, pascal, volta, turing, ampere, all"
+				die "Invalid CUDA architecture given: '${VTK_CUDA_ARCH}'!"
+				;;
+		esac
+		ewarn "Using CUDA architecture '${cuda_arch}'"
+
+		mycmakeargs+=( -DVTKm_CUDA_Architecture=${cuda_arch} )
 	fi
 
 	if use ffmpeg; then


sci-libs/vtk/vtk-9.0.1.ebuild
```diff
@@ -139,6 +139,28 @@ DOCS=( CONTRIBUTING.md README.md )
...
+
+get_cuda_arch() {
+	if [[ -z ${CUDA_ARCH} ]]; then
+		MY_CUDA_ARCH=native
+		return
+	else
+		# we ignore fermi arch, because current nvidia-cuda-toolkit-11*
+		# no longer supports it
+		case ${CUDA_ARCH} in
+			kepler|maxwell|pascal|volta|turing|ampere|all|none)
```
ionenwks, line 152: Any chance to use `all` as default?

Automatic hardware detection is something that really should be avoided, no guarantee even being built on the system this will run on and so, unless user manually specify it through CUDA_ARCH, `all` seem like the only valid default.

I get that it'll make compilation take longer though (not that I tried it yet to see how bad it is).

sci-libs/vtk/vtk-9.0.1.ebuild
```diff
@@ -139,6 +139,28 @@ DOCS=( CONTRIBUTING.md README.md )
 
 CHECKREQS_DISK_BUILD="3G"
 
+MY_CUDA_ARCH=
+
+get_cuda_arch() {
```
ionenwks, line 144: I assume you're using a function with the idea that the cuda eclass will replace it eventually, but global scope functions should bit more private namespacing like vtk_ or my_

Also, functions used only in a single phase are normally defined inside the phase itself to have less in global scope. There's even been talk to make this a pkgcheck warning.

...that aside, personally feel this shouldn't be a function if called only once, not like it'll be hard to delete the block and replace by a cuda.eclass function call if it ever exists.

sci-libs/vtk/vtk-9.0.1.ebuild
```diff
@@ -209,6 +231,11 @@ src_prepare() {
 	if use cuda; then
 		cuda_add_sandbox -w
 		cuda_src_prepare
+
+		get_cuda_arch
+		einfo "Using CUDA architecture ${MY_CUDA_ARCH}"
+		einfo "If auto detection fails for you, please try and export the CUDA_ARCH"
+		einfo "environment variable to one of the common arch names"
```
ionenwks, line 238: Just to be pedantic, this seem like src_configure material rather than src_prepare (it's just picking a configure option).

If use src_configure and no get_cuda_arch function, could use a local variable rather than MY_CUDA_ARCH too, all contained within src_configure's cuda block.

sci-libs/vtk/vtk-9.0.1.ebuild
```diff
@@ -139,6 +139,28 @@ DOCS=( CONTRIBUTING.md README.md )
 
 CHECKREQS_DISK_BUILD="3G"
 
+MY_CUDA_ARCH=
+
+get_cuda_arch() {
+	if [[ -z ${CUDA_ARCH} ]]; then
+		MY_CUDA_ARCH=native
+		return
+	else
```
ionenwks, line 148: This if/else feels kind of redundant, you're setting =native in two spots.

Default variable assignment may be of use, e.g. `${CUDA_ARCH:-all}` would return `all` if unset, otherwise `${CUDA_ARCH}`

sci-libs/vtk/vtk-9.0.1.ebuild
```diff
@@ -139,6 +139,28 @@ DOCS=( CONTRIBUTING.md README.md )
 
 CHECKREQS_DISK_BUILD="3G"
 
+MY_CUDA_ARCH=
+
+get_cuda_arch() {
+	if [[ -z ${CUDA_ARCH} ]]; then
```
ionenwks, line 145: One last thing, if we ever implement this in the eclass, there's no telling how we'll be using CUDA_ARCH variable exactly and it may conflict with vtk's implementation.

So VTK_CUDA_ARCH may make more sense until we have a real centralized implementation.

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: Build times will seriously hurt with `all`. While developing the ebuild, I had a build time of much more than a day

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: For one, I'm using a function with the idea of extending the cuda eclass.
Second, it makes it more modular. I like my ebuilds clean and do not like the idea of having a if-clause containing a switch clause, all within an optional `if use cuda; then` statement, deeply nested inside a phase function.

sci-libs/vtk/vtk-9.0.1.ebuild
ionenwks: Oh well. Alright then, keep native until a better solution -- at least have a einfo now so people are warned of this.

Not that einfo is very visible (won't show if people aren't looking at build.log), I'd say this is getting more important and maybe should be upgraded to ewarn if VTK_CUDA_ARCH is unset.

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: I mainly used src_prepare to fail as early as possible if anything goes wrong with this and to give the user a hint early on.

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: fair enough

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: I was thinking on setting the MY_CUDA_ARCH to an initial value of `native` to avoid this path.
Good point about the default assignment

sci-libs/vtk/vtk-9.0.1.ebuild
ionenwks: Why would get_cuda_arch fail? It's just checking variables and giving info.

Also src_prepare makes no real difference, if you want to fail early it'd be pkg_pretend.

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: What do you think of not respecting the `fermi` arch? We still have nvidia-cuda-toolkit v9 and v10 in the tree, which both support this arch? Is it better to support it, due to being outdated and probably not used with real cuda applications?

ionenwks: Fermi needs LTS nvidia-drivers-390.xx and all in-tree cuda-toolkit currently have a dependency on DRIVER_PV >= 396.xx, meaning they can't install any of them as things stands, as such no sense in vtk worrying about fermi unless that's changed.

390 is also on its last legs, losing features after features (nvenc, uvm on x86, egl-wayland is broken too), and official end of support is due end of next year. I'll keep it in the tree for longer with minimal support but once something goes really wrong it'll be the end of it.

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: On a closer look... together with `test -z` the test would never be true, because, if `CUDA_ARCH` is unset or null, the default value will be assigned, thus giving the string a length.

sci-libs/vtk/vtk-9.0.1.ebuild
```diff
@@ -294,7 +294,25 @@ src_configure() {
 	fi
 
 	if use cuda; then
-		mycmakeargs+=( -DVTKm_CUDA_Architecture="native" )
+		local MY_CUDA_ARCH=native
+		if [[ -n ${VTK_CUDA_ARCH} ]]; then
+			# we ignore fermi arch, because current nvidia-cuda-toolkit-11*
+			# no longer supports it
+			case ${VTK_CUDA_ARCH} in
```
ionenwks, line 301: What I had in mind for default assignment was to potentially use it here, i.e. remove "if" entirely, and then add native to the list of valid values. Not that it changes much beside less indenting, I'm okay with it either way (better than before).

sci-libs/vtk/vtk-9.0.1.ebuild
```diff
@@ -294,7 +294,25 @@ src_configure() {
...
+				kepler|maxwell|pascal|volta|turing|ampere|all|none)
```
ionenwks, line 302: Is there a point to "none"? Sounds like it'd be like USE=-cuda, not that I checked what it does. If users want to disable it, it's the wrong place for this.

sci-libs/vtk/vtk-9.0.1.ebuild
```diff
@@ -294,7 +294,25 @@ src_configure() {
...
+					MY_CUDA_ARCH=${VTK_CUDA_ARCH}
+					;;
+				*)
+					# use default value of MY_CUDA_ARCH, ignore any other values
```
ionenwks, line 306: Could `die` with arch's value in the message to inform users they set an unexpected value rather than fallback to native against the user's wish.

May allow to simplify too given this case statement would be a pure sanity check and not need to set anything.

It's unlikely so even falling late doesn't seem so bad.

sci-libs/vtk/vtk-9.0.1.ebuild
```diff
@@ -294,7 +294,25 @@ src_configure() {
...
+					;;
+			esac
+		fi
+		ewarn "Using CUDA architecture ${MY_CUDA_ARCH}"
+		ewarn "If auto detection fails for you, please try and export the"
+		ewarn "VTK_CUDA_ARCH environment variable to one of the common arch"
+		ewarn "names"
```
ionenwks, line 313: Think users may potentially not know what this refers to nvidia's naming schemes, maybe an example arch or two? The displayed "native" default doesn't say much about what's expected here.

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: Haven't checked it yet. Sounds like it is like you're saying. May it be there's some code which builds for cuda without building for specific device?

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: I strongly disagree with your last statement. It was way better before. Easier to read, easier to follow the code flow, the structure more clean.

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: agreed

sci-libs/vtk/vtk-9.0.1.ebuild
ionenwks: I don't really get why a simple sanity check need to be complicated, but okay.

ionenwks: I was also wondering if this check is really useful, i.e. could just pass the value if the build system knows to error out on unknown values.

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl:
```
# The option 'none' is provided so that when being built as part of another
# project, its own custom flags can be used.
```
I suppose `none` is not relevant for our use case.

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: Build system doesn't fail during configuration, it just accepts invalid values. I just tried with `VTK_CUDA_ARCH=-DDEBUG ebuild vtk-9.0.1 clean configure`. The src_configure phase succeeds, the relevant values are
```
artus # grep VTKm_CUDA /var/tmp/portage/sci-libs/vtk-9.0.1/work/vtk-9.0.1_build/CMakeCache.txt 
VTKm_CUDA_Architecture:STRING=-DDEBUG
//STRINGS property for variable: VTKm_CUDA_Architecture
VTKm_CUDA_Architecture-STRINGS:INTERNAL=native;fermi;kepler;maxwell;pascal;volta;turing;all;none
```
It will probably fail during src_compile phase.

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: > Automatic hardware detection is something that really should be avoided, no guarantee even being built on the system this will run on and so, unless user manually specify it through CUDA_ARCH, `all` seem like the only valid default.

What's wrong with introspection? We're not in the 1990's anymore. This should work in most of the cases.

ionenwks: Notice how we also have CPU_FLAGS_X86, VIDEO_CARDS, etc... rather than auto-detection. The entire ecosystem is built around the user specifying what they can or can't use. This is a bit like a build system passing -march=native by default and later failing on everything but the current hardware.

sci-libs/vtk/vtk-9.0.1.ebuild
ionenwks: Given it could take a while to fail, might as well keep the check then -- better off having `die` do it.

sci-libs/vtk/vtk-9.0.1.ebuild
waebbl: Your examples are different. We don't have a USE_EXPAND for CUDA architectures yet. I agree, that it's better to choose from a pre-defined set of values, yet we don't have such values. There's nothing wrong with using -march=native, I just have to be aware of the consequences and can not expect a binary built with it, to be able to run on another hardware. But I fully trust the compiler to produce valid code which can run on my machine. The same is true for the GPU built into my machine.

ionenwks: I thought we had already concluded that was okay going with native due to build times? I'm not sure why you brought this thread back up.

We could keep arguing about how this goes against Gentoo philosophy but this doesn't change the ebuild unless you're trying to get convinced to go back on this.

waebbl: > I'm not sure why you brought this thread back up.

Gonna write you in private about this.

ionenwks (approved):
Thanks because I'll go ahead and make use of this myself. Native unsurprisingly always failed to build in my containers, but built fine with pascal just now :)

ionenwks:
I've slightly updated the commit message to reflect the current state of things, e.g. VTK_CUDA_ARCH, and no function.

Hope you didn't mind.

waebbl:
> Thanks because I'll go ahead and make use of this myself. Native unsurprisingly always failed to build in my containers, but built fine with pascal just now :)

Thank you for reviewing. I too have issues in my test chroots occassionally with native. But it never failed on my desktop.

waebbl:
> I've slightly updated the commit message to reflect the current state of things, e.g. VTK_CUDA_ARCH, and no function.
> 
> Hope you didn't mind.

Never mind. Good catch. I forgot to update the comment

PR #21860, opened 2021-08-02
Xboxdrv cmake

juippis:
- upstream has moved to cmake, but no new release since 2015.
 - build works, but runtime doesn't, the error is still the same as
   reported in 2019:
     https://gitlab.com/xboxdrv/xboxdrv/-/issues/258
 - throwing my cmake'd ebuild here to continue when upstream has
   fixed the runtime issue.


dev-libs/argparser/argparser-1.0.0.ebuild
```diff
@@ -0,0 +1,31 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit cmake
+
+DESCRIPTION="A Command Line Argument Parser for C++"
+HOMEPAGE="https://gitlab.com/argparser/argparser"
+SRC_URI="https://gitlab.com/argparser/argparser/-/archive/v1.0.0/argparser-v1.0.0.tar.bz2"
```
juippis, line 10:
```suggestion
SRC_URI="https://gitlab.com/argparser/argparser/-/archive/v${PV}/argparser-v${PV}.tar.bz2"
```

dev-libs/argparser/argparser-1.0.0.ebuild
```diff
@@ -0,0 +1,31 @@
...
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="amd64 ~x86"
```
juippis, line 14:
```suggestion
KEYWORDS="~amd64 ~x86"
```

dev-libs/argparser/metadata.xml
```diff
@@ -0,0 +1,13 @@
...
+<pkgmetadata>
+  <maintainer type="person">
+    <email>juippis@gentoo.org</email>
+    <name>Joonas Niilola</name>
+  </maintainer>
+  <stabilize-allarches/>
+  <upstream>
+    <remote-id type="pypi">capyt</remote-id>
+    <remote-id type="github">lbatalha/capyt</remote-id>
+  </upstream>
```
juippis, line 12:
```suggestion
```

games-util/xboxdrv/xboxdrv-0.8.8_p20200909.ebuild
```diff
@@ -0,0 +1,88 @@
...
+	dev-libs/glib:2
+	dev-libs/libfmt
+	net-wireless/bluez
+	sys-apps/dbus
+	virtual/libudev:=
+	>=virtual/libusb-1:=
+	x11-libs/libX11
+	x11-libs/cairo
+	x11-libs/gdk-pixbuf
+	x11-libs/gtk+:3"
```
juippis, line 31:
```suggestion
	x11-libs/cairo
	x11-libs/gdk-pixbuf
	x11-libs/gtk+:3
	x11-libs/libX11"
```

ionenwks:
fwiw same runtime issue here (I used to use this long ago but not anymore), it kinda start with --ui-clear but it's also useless, looks like a case of halfway code revamping that likely never worked right and may never. I'd say just updating the old branch was the right move (next to last rites anyway).

juippis:
Thanks. Yeah, I haven't actually used my xbox gamepad in years, but I remember needing this so the LED didn't blink nonstop. But then again I have a vague memory of seeing something kernel-related for that behaviour too. 

I should just really unplug the controller, but before that, check if this can be resolved by a kernel update. Wonder if anyone else needs this indeed...

dev-libs/argparser/argparser-1.0.0.ebuild
```diff
@@ -0,0 +1,31 @@
...
+IUSE="test"
+
+RDEPEND="dev-libs/libfmt"
```
thesamesam, line 17: := or not at all in RDEPEND (but I think it should be there)

dev-libs/argparser/argparser-1.0.0.ebuild
```diff
@@ -0,0 +1,31 @@
...
+DEPEND="${DEPEND}"
+BDEPEND="test? ( dev-cpp/gtest )"
+
+S="${WORKDIR}/argparser-v${PV}"
+
+PATCHES=( "${FILESDIR}"/argparser-1.0.0-enable-shared-libs.patch )
+
+src_configure() {
+	local mycmakeargs=(
+		-DBUILD_TESTS=$(usex test true false)
```
thesamesam, line 27: Definitely need true/false explicitly?

dev-libs/argparser/metadata.xml
thesamesam: Yeah, doesn’t seem allarches

games-util/xboxdrv/xboxdrv-0.8.8_p20200909.ebuild
```diff
@@ -0,0 +1,88 @@
...
+
+src_prepare() {
+	cmake_src_prepare
+
+	# Don't build bundled libs
+	sed -i '/subdirectory/d' CMakeLists.txt || die
+}
+
+src_configure() {
+	local mycmakeargs=(
```
thesamesam, line 62: True false needed?

games-util/xboxdrv/xboxdrv-0.8.8_p20200909.ebuild
```diff
@@ -0,0 +1,88 @@
...
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="test"
+
+DEPEND="dev-libs/argparser
+	dev-libs/dbus-glib
+	dev-libs/glib:2
+	dev-libs/libfmt
```
thesamesam, line 23: := or just DEPEND but I don’t think it’s header only

PR #21862, opened 2021-08-02
sys-fs/udev: enable musl

SoapGentoo:


sys-fs/udev/udev-249-r2.ebuild
```diff
@@ -22,6 +22,9 @@ else
 	KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~s390 ~sparc ~x86"
 fi
 
+SRC_URI+="
+	elibc_musl? ( https://dev.gentoo.org/~soap/distfiles/${P}-musl-patches.tar.xz )"
```
floppym, line 26: Could you add a comment indicating where these patches came from?

profiles/features/musl/package.mask
```diff
@@ -56,7 +56,7 @@ sys-libs/glibc
 # We use eudev which is tested on both uclibc and musl
 =virtual/libudev-232-r1
 sys-apps/systemd
-sys-fs/udev
+<sys-fs/udev-249-r2
```
thesamesam, line 60: Going to think about either keeping the mask or masking newer versions of the virtual for now to avoid changing defaults for new installs yet.

sys-fs/udev/udev-249-r2.ebuild
gyakovlev: I wonder how different it to the tmpfiles patchset.
looking at tar archive yours it a bit smaller.
    # http://cgit.openembedded.org/openembedded-core/tree/meta/recipes-core/systemd/systemd
    # check SRC_URI_MUSL in systemd_${PV}.bb file for exact list of musl patches
^ I just extract those patches without much filtering.

floppym (approved):

gyakovlev (approved):
add `app-text/tree` as test depend and good to go.
passes all test on ppc64-be-musl, except 2 disk related irrelevant tests.

sys-fs/udev/files/249-sys-include-posix.patch
```diff
@@ -0,0 +1,62 @@
+From f8d54f7810aeea5ff27a5db03e1aab7ea54c8268 Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: No link to upstream PR? ;)

thesamesam (approved):

PR #21865, opened 2021-08-03
media-gfx/hydrus: Fix wrong dosym, drop use flags for optfeature

vaartis:
@juippis 

Since I just noticed the dosym error I fixed it as soon as possible plus dropped the USE flags for optfeatures as discussed. Also made some default USE flag dependencies non-optional since the functinality is restricted somewhat without them, hence why they were default-enabled.


vaartis:
ping

juippis:
Thanks for the fast work! I'll be finishing this today. Had no access to my pc in a week.

Currently digging deeper into why pyside2 fails to compile...

juippis (approved):
Tested latest and 445 without any optfeature packages installed, seems to work fine! 
Thanks!

PR #21867, opened 2021-08-03
net-proxy/mitmproxy: Bump to 7.0.2

00-matt:


net-proxy/mitmproxy/mitmproxy-7.0.2.ebuild
```diff
@@ -0,0 +1,70 @@
...
+	>=dev-python/wsproto-1.0.0[${PYTHON_USEDEP}]
+	>=dev-python/zstandard-0.11.0[${PYTHON_USEDEP}]
+	dev-libs/openssl:0
+"
+
+BDEPEND="
+	test? (
+		>=dev-python/hypothesis-5.8[${PYTHON_USEDEP}]
+		>=dev-python/parver-0.1[${PYTHON_USEDEP}]
+		>=dev-python/pytest-6.1.0[${PYTHON_USEDEP}]
```
ionenwks, line 51: Why add this? Is it for the version check?

`distutils_enable_tests pytest` already adds `test? ( >=pytest-4.5.0... )` to BDEPEND, and pytest-asyncio will further request >=5.4.0 (mitmproxy tests passed fine with pytest-5.4.3 for me, so >=6.1.0 doesn't seem to matter -- plus 6.1 been stable since Nov 2020 and very likely to be used either way).

00-matt: Yes I added it because it's in mitmproxy's `setup.py`.  Although it has required pytest-6.1.0 since version 5.3.0 and seems to work fine with older pytest versions... I'll just drop it

ionenwks (approved):
Thanks, typical of these having overzealous restrictions.

PR #21871, opened 2021-08-03
net-im/telegram-desktop: Bug fixes and version bump

mid-kid:
@gyakovlev


gyakovlev:
thanks, merged.

I had to revbump 2.8.11, because both changes you have introduced required a revbump.
changing RDEPEND generally requires revbump, except some edge cases.
patching runtime behavior requires revbump, no exceptions.

and any users who already updated to stable ( via https://bugs.gentoo.org/802537) will get a patched up `-r1` version now.

mid-kid:
You're right, I forgot. Sorry.

gyakovlev:
no problem at all =) that's my job too to catch those things. you do heavy lifting, I catch policy stuff, seems like good collaboration ;-)

mid-kid:
@gyakovlev I re-read the commits and I noticed I messed up with usex. Can you merge this commit real quick? https://github.com/mid-kid/gentoo/commit/0846f5a92f8d8bf5375fddb895fd8352e335b291

gyakovlev:
merged, sorry for missing it.
I always hated negative logic in those options.

PR #21872, opened 2021-08-04
ejabberd: Package updates for 21.07

jpds:
Tested on an `amd64` system.


Flowdalic:
Thanks for your contribution.

I have a few things that I'd like to change in ejabberd's (and its associated ebuilds). By priority those are
1. Move away from EAPI 6 (not sure if straight to 8, or "just" to 7)
2. I think the ejabberd ebuild should declare most of its dependencies as fixed version dependencies (minus revisions). That is, instead of `>=dev-erlang/eimp-1.0.20`, you have `~dev-erlang/eimp-1.0.20`. This is in accordance with ejabberd's rebar configuration, which also declares fixed version dependencies (even though newer versions may work, as I was told by an ejabberd developer).
3. Switch away from `net-im/jabber-base`. I already talked with @ConiKost (the maintianer of prosody in ::gentoo), and we both don't see any point in `jabber-base`. Especially running all jabber servers (prosody, ejabberd, …) under the same user seems more like a security risk than net gain. We already [reserved a UID/GID pair for the ejabberd user](https://github.com/gentoo/api-gentoo-org/blob/82bda8d73304b53621ed7ceb69db4fb926da3743/files/uid-gid.txt#L170). Note that this requires an migration phase in the ebuild, where the ownership of ejabberd's configuration files is migrated from the old `jabber` user to the new `ejabberd` user (and group).
4. Remove the TLS certificate handling in ejabberd's ebuild. It accounts for a large part of the ebuild's complexity and is mostly obsolete due to Let's Encrypt (and `dev-erlang/p1_acme`).

I started working on this a while ago. You will find [e.g. ebuilds for `acct-user/ejabberd` in my overlay](https://gitlab.com/Flow/flow-s-ebuilds/-/tree/master/acct-user/ejabberd) (note that this was before the 21.07 release of ejabberd).

Unfortunately, I am currently battling a deadline until mid august (and will be on vocation shortly after that), and due the amount of planned changes, my processes on the updated ejabberd ebuilds is not as fast as I hoped.

Flowdalic (changes requested):
Should address the points I've mentioned.

Flowdalic:
This probably has become obsolete with ejabberd 21.12 in ::gentoo now. Hence I am closing it. Feel free to issue a new PR if there are things of this PR that you believe should still be merge.d

PR #21875, opened 2021-08-04
dev-cpp/asio: version 1.19.2

DarthGandalf:


dev-cpp/asio/asio-1.19.2.ebuild
```diff
@@ -0,0 +1,51 @@
...
+IUSE="doc examples ssl test"
+RESTRICT="!test? ( test )"
+# test searches for libssl during ./configure, and REQUIRED_USE is easier than
+# patching configure to not search for it with USE=-ssl
+REQUIRED_USE="test? ( ssl )"
+
+RDEPEND="dev-libs/boost:=
+	ssl? (
+		dev-libs/openssl:0=
+	)"
```
ionenwks: Why does this have a RDEPEND? Or especially binding `:=` operators when it doesn't install anything linked against these libraries (meaningless rebuilds). Setting this and binding operators is something asio consumers need to do, not asio itself.

Both openssl and boost seem like test-only deps thanks to the dummied Makefile, could just have a DEPEND="test? ( ... )" with both (and no :=/:0=, don't need to keep the 0 anymore either), and remove the IUSE=ssl and REQUIRED_USE

ionenwks: Then again, galera has a meaningless `dev-cpp/asio[ssl]` (with its own dependency on openssl:=), so that would need to be updated if wanted to drop it.

ionenwks: Oh wait, that one only uses <asio-1.12.0 so it at least doesn't matter for latest.

DarthGandalf: Because to use asio with ssl user doesn't need to #include openssl in user code, including asio/ssl.hpp is enough; but that won't work without openssl being also installed

ionenwks: But ultimately consumers still need to add the openssl dependency either way given they'll be linking against it and need the binding operator too (which can't be handled by asio, meaningless there).

DarthGandalf: The same applies to dependency on boost. Do users need to know all the recursive dependencies of a library to be able to depend on it?

SoapGentoo: this is where public dependencies become messy - in theory yes.

ionenwks (approved):
Yep, looks good to me now. Merging as-is.

PR #21877, opened 2021-08-04
sys-auth/seatd: Add builtin and server IUSE flags

greg904:
I wanted to use the `builtin` backend in libseat but I saw that it was disabled at compile time. This PR adds an USE flag to enable it.

Also the standalone server can be disabled if it's not needed so while we're at it, I added a flag for that too. Let me know if that's not needed.

### DIFF
--- a/sys-auth/seatd/seatd-0.5.0-r1.ebuild
+++ b/sys-auth/seatd/seatd-0.5.0-r1.ebuild
@@ -11,7 +11,7 @@ if [[ ${PV} == 9999 ]]; then
 	inherit git-r3
 	EGIT_REPO_URI="https://git.sr.ht/~kennylevinsen/seatd"
 else
-	KEYWORDS="~amd64 ~arm64 ~ppc64 ~x86"
+	KEYWORDS="amd64 arm64 ~ppc64 ~riscv x86"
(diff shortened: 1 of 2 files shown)


lanodan:
- `seatd-0.5.0` has been stabilized so you should move it's modifications to `seatd-0.5.0-r1`
- `newinitd` and `systemd_dounit` commands in `src_install` should be into a `if use server; then … fi`

lanodan (approved):

sys-auth/seatd/seatd-0.5.0-r1.ebuild
```diff
@@ -0,0 +1,53 @@
...
+
+inherit meson systemd
+
+DESCRIPTION="Minimal seat management daemon and universal library"
+HOMEPAGE="https://sr.ht/~kennylevinsen/seatd"
+if [[ ${PV} == 9999 ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://git.sr.ht/~kennylevinsen/seatd"
+else
+	KEYWORDS="amd64 arm64 ~ppc64 ~riscv x86"
```
ionenwks, line 14: I'll drop to ~arch on merge.

ionenwks (approved):

PR #21881, opened 2021-08-04
media-plugins/mythplugins: bump snapshot 20210731, add Python 3.9 target

thebitpit:


thebitpit:
Bump the plugins version and source snapshot to match mythtv because they share source and versions must match.
The use of dev-qt/qtwebkit:5 has not been fixed in the fixes/31 branch yet.

juippis (approved):

PR #21884, opened 2021-08-04
media-libs/openexr: bump to 3.1.1

waebbl:


media-libs/openexr/openexr-3.1.1.ebuild
```diff
@@ -0,0 +1,78 @@
...
+	if use doc; then
+		DOCS+=( docs/*.pdf )
+	fi
+	use examples && docompress -x /usr/share/doc/${PF}/examples
+	cmake_src_install
+
+	cat > "${T}"/99${PN}3 <<-EOF || die
+	LDPATH=/usr/$(get_libdir)/${MY_P}
+	EOF
+	doenvd "${T}"/99${PN}3
```
juippis, line 77: There's a better syntax for this, 
```suggestion
	newenvd - 99${PN}3 <<-EOF
		LDPATH=/usr/$(get_libdir)/${MY_P}
	EOF
```

juippis (approved):
Don't know if it's related, but upon 6 random rdeps I compiled, kde-frameworks/kimageformats-5.82.0 was broken. But quickly looking through it looked more like imath problem.

juippis (commented):
Also is this a mistake or...?
`DroppedKeywords: version 3.1.1: arm, arm64, hppa, ppc, ppc64, riscv, sparc`?

waebbl:
> Also is this a mistake or...?
> `DroppedKeywords: version 3.1.1: arm, arm64, hppa, ppc, ppc64, riscv, sparc`?

https://bugs.gentoo.org/788235

When I asked back when 3.0.1 was released, someone said, I should remove those keywords which I can't test for slot 3.

media-libs/openexr/openexr-3.1.1.ebuild
waebbl: Tried this, but didn't work out. Most probably because I had some unidentified syntax error in the command.

PR #21891, opened 2021-08-05
net-im/telegram-desktop: Drop vulnerable versions

mid-kid:
@gyakovlev Doing this will allow security to close the bug, right?


gyakovlev:
yeah, I updated the bug too.
it's up to security now.

PR #21893, opened 2021-08-05
dev-embedded/stlink: bump to v1.7.0

dawn-minion:
Bumps stlink to v1.7.0. First release with full support for V3 programmers, as well as new chips. No new deps, works well here with the new ebuild.


dawn-minion:
@alexxy Are you able to look at this change? Would be nice to have the latest version in and not build from git to be able to use st-link V3s. Happy to be a proxy maintainer here too if it helps, I use this package extensively as part of my normal day to day work.

thesamesam:
> @alexxy Are you able to look at this change? Would be nice to have the latest version in and not build from git to be able to use st-link V3s. Happy to be a proxy maintainer here too if it helps, I use this package extensively as part of my normal day to day work.

That sounds pretty good! If you get no reply, it's fine to ping us in e.g. #gentoo-proxy-maint.

This version seems to be in tree now: https://github.com/gentoo/gentoo/pull/21893#issuecomment-901358342.

PR #21894, opened 2021-08-05
profiles: remove pcl mask of ~media-gfx/freecad-0.19

waebbl:
sci-libs/pcl-1.12.0 now builds against vtk-9


ionenwks (approved):
I happened to see it, but be good to ping proxy-maint on profile-only changes given there's no assignment, without a notification it may not get in members' queue and go missed.

PR #21901, opened 2021-08-06
dev-db/mydumper: bump to 0.10.7.2

hydrapolic:
Tested on ~amd64.


dev-db/mydumper/mydumper-0.10.7.2.ebuild
```diff
@@ -0,0 +1,49 @@
...
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="doc"
+
+RDEPEND="app-arch/zstd
+	dev-db/mysql-connector-c:=
+	dev-libs/glib:=
+	dev-libs/libpcre:=
+	dev-libs/openssl:0=
```
ionenwks: General rule of thumb for adding := is when dependency has a non-zero subslot defined (albeit there are exceptions, like qt deps are special), and not do it otherwise.

And when a package define a non-zero slot (like glib), should also either request it directly or use `:*` if there is multiple slots and they are all known to work.
```suggestion
RDEPEND="app-arch/zstd:=
	dev-db/mysql-connector-c:=
	dev-libs/glib:2
	dev-libs/libpcre
	dev-libs/openssl:=
```
As for openssl, its slots are a thing of a distant past.. not that cleaning the 0 really matters so you'll still see a lot of ebuilds with it.

dev-db/mydumper/mydumper-0.10.7.2.ebuild
```diff
@@ -0,0 +1,49 @@
...
+	sys-libs/zlib:="
+DEPEND="${RDEPEND}"
+BDEPEND="virtual/pkgconfig
+	doc? ( dev-python/sphinx )"
+
+PATCHES=(
+	"${FILESDIR}/${PN}-0.10.7-atomic.patch" #654314
+)
+
+S="${WORKDIR}/${MY_P}"
```
ionenwks, line 33: nitpick but usually this is put right below SRC_URI without a blank line (related to it, closer to where MY_P is set, and matches skel.ebuild ordering)

dev-db/mydumper/mydumper-0.10.7.2.ebuild
```diff
@@ -0,0 +1,49 @@
...
+	dev-libs/glib:=
+	dev-libs/libpcre:=
+	dev-libs/openssl:0=
+	sys-libs/zlib:="
+DEPEND="${RDEPEND}"
+BDEPEND="virtual/pkgconfig
+	doc? ( dev-python/sphinx )"
+
+PATCHES=(
+	"${FILESDIR}/${PN}-0.10.7-atomic.patch" #654314
```
ionenwks, line 30: This patch is identical to 0.10.5-atomic.patch, just use that rather than duplicate it (version doesn't need to match and is just an indicator of when first added).

dev-db/mydumper/mydumper-0.10.7.2.ebuild
```diff
@@ -0,0 +1,49 @@
...
+	sed -i -e 's|-Werror -O3 -g|${CMAKE_C_FLAGS}|' CMakeLists.txt || die
+
+	# fix doc install path
+	sed -i -e "s|share/doc/mydumper|share/doc/${PF}|" docs/CMakeLists.txt || die
+
+	cmake_src_prepare
+}
+
+src_configure() {
+	local mycmakeargs=("-DBUILD_DOCS=$(usex doc)")
```
ionenwks, line 46: These quotes aren't needed.

dev-db/mydumper/mydumper-0.10.7.2.ebuild
```diff
@@ -0,0 +1,49 @@
...
+
+EAPI=7
+
+inherit cmake
+
+MY_P="${PN}-$(ver_rs 3 -)"
+
+DESCRIPTION="A high-performance multi-threaded backup (and restore) toolset for MySQL"
+HOMEPAGE="https://github.com/maxbube/mydumper"
+SRC_URI="https://github.com/maxbube/${PN}/archive/v${MY_PV}.tar.gz -> ${MY_P}.tar.gz"
```
ionenwks, line 12: `${MY_PV}`?
```
>>> Downloading 'https://github.com/maxbube/mydumper/archive/v.tar.gz'
2021-08-07 05:39:27 ERROR 404: Not Found.
```

dev-db/mydumper/mydumper-0.10.7.2.ebuild
```diff
@@ -0,0 +1,49 @@
...
+
+LICENSE="GPL-3"
```
ionenwks, line 14: Seeing the later clause in mydumper.h
```suggestion
LICENSE="GPL-3+"
```

dev-db/mydumper/mydumper-0.10.7.2.ebuild
hydrapolic: Thanks, I've missed https://github.com/gentoo/gentoo/commit/0f936ec81f1dc20fa58ba40bc17989ee44810214

dev-db/mydumper/mydumper-0.10.7.2.ebuild
hydrapolic: Seems like I tried the wrong patch (there are 2 in filesdir, no idea why I picked the oldest, duh...) when building, because I got an error when applying it, so I thought something changed in the cmake file and the patch needs to be regenerated.

dev-db/mydumper/mydumper-0.10.7.2.ebuild
hydrapolic: Seems like I had MY_PV in the ebuild before and removed just by accident and since the file was downloaded the building passed, oh, stupid mistake. MY_PV is needed as version is 0.10.7-2.

Also haven't noticed the PN plaintexting, thanks for that.

dev-db/mydumper/mydumper-0.10.7.2.ebuild
hydrapolic: Good catch, thanks.

hydrapolic:
Thanks @ionenwks for the review, sorry for the stupid mistake with MY_PN. Hopefully it's ok now.

ionenwks (approved):
> Thanks @ionenwks for the review and tips, sorry for the stupid mistake with MY_PN. Hopefully it's ok now.

No problem, it happens. And yes, all ok :)

Also thanks for these drop old, I see still had some cmake-utils in there.

hydrapolic:
Thanks

PR #21912, opened 2021-08-08
dev-libs/leveldb: bump to v1.23

arthurzam:
@Zlogene Here is my assignment

For this ebuild a lot have changed, and we can view it as a new ebuild.

1. Upstream moved to CMake, making building it much cleaner.
2. Removed old HOPEPAGE which doesn't exist (the current official one is the GitHub repo). Did it also for the old ebuild.
3. Removed the "a " prefix from the DESCRIPTION
4. There is no need for manual OS targeting (Linux vs FreeBSD) thanks to CMake move.
5. Added a new USE flag and dependency for `crc32c`, which enables accelerated crc32 computation
6. Added test dependencies `dev-cpp/{gtest,benchmark}`, which I unbundled using a self written patch to use the system installed one. The patch isn't upstreamable as upstream prefers the current way.
7. Removed possible `-Werror` compile flag for clang.
8. For testing, set some extra environment variables so all testing happens in `${T}` without artifacts on `/tmp`. Had quite a hard time to understand why sometimes the test suite fails, until I found that I had those temp artifacts left after a test run as root.

Now the problems that rose from the bump:

1. We are unable to support static libs generation. CMake enables choosing only one (shared/static) and installing only static library isn't the Gentoo way. I have checked all reverse dependencies (using `grep -nr . -e dev-libs/leveldb` and found no such dependencies on the use flag).
2. Because of the addition of new dependencies, I needed to drop many keywords. I have opened [bug](https://bugs.gentoo.org/807025), which would change to a rekeyword one after merge (and yes, I missed some keywords in the previous request, will be fixed).

I guess after becoming an official developer, I will take maintenance on this package. Actually might be useful for one of my projects :smile:


dev-libs/leveldb/files/leveldb-1.23-unbundle-gtest-benchmark.patch
```diff
@@ -0,0 +1,40 @@
...
+
+Signed-off-by: Arthur Zamarin <arthurzam@gmail.com>
+
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -289,27 +289,8 @@ target_link_libraries(leveldbutil leveldb)
+ if(LEVELDB_BUILD_TESTS)
+   enable_testing()
+ 
+-  # Prevent overriding the parent project's compiler/linker settings on Windows.
```
mgorny, line 15: You can generally make the patch smaller by using `if (FALSE)` or alike. Or even better, by adding an `option()` to use system gtest/benchmark and making an upstreamable patch with it.

dev-libs/leveldb/metadata.xml
```diff
@@ -2,6 +2,9 @@
 <!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
 <pkgmetadata>
   <!-- maintainer-needed -->
+  <use>
+    <flag name="crc32c">Enable accelerated crc32 computation using <pkg>dev-libs/crc32c</pkg></flag>
```
mgorny, line 6: Any reason not to enable that unconditionally? It sounds a bit like `USE=-crc32c` uses bundled CRC32 routines.

dev-libs/leveldb/files/leveldb-1.23-unbundle-gtest-benchmark.patch
arthurzam: Well, smart minds think a like, and found out that the idea was already thought about before, https://github.com/google/leveldb/pull/912.
I guess Google don't want this idea (and I can understand why actually, theirs reason is quite good).

Is it better a smaller patch (with the `if` clause) or use the one from the mentioned PR, whose patch is even bigger than mine (but better).

dev-libs/leveldb/metadata.xml
arthurzam: Was thinking it as an easier mask if there would be a need for an arch around `dev-libs/crc32c`. I saw that there is a basic, not hardware accelerated version of crc32c, bundled inside leveldb.

But after extra investigation of `dev-libs/crc32c`, I found it also has the same portable code, meaning it should work on other arches without big hiccups (I hope), after the required keywording.

Fixed this point.

dev-libs/leveldb/files/leveldb-1.23-unbundle-gtest-benchmark.patch
mgorny: Oh, right, it's gtest. In that case you don't even need to care to patch it. From my experience, using external gtest may cause real issues (e.g. because it heavily relies on matching flags).

arthurzam: As obviously you are more experienced, I changed the way I use `gtest` and `benchmark`.
Now I don't unbundle them at all, and download the version specified as a submodule.
I decided to use the SRC_URI which downloads this commit, and then uses best matching `PV`, to have best chances to match other packages using the same commit and format.

mgorny: Argv, I am sorry, I didn't know that they aren't included in the tarball. I'd generally suggest doing whatever is easier, so up to you whether you prefer to fetch them or unbundle them.

mgorny: s/is easier/& and works/

arthurzam: Then lets stay with current version (download `SRC_URI` and link) as it enables us to test like upstream, and like you said, no problems with mismatching flags. It also means less patch handling on Gentoo's side.

arthurzam:
Closing the PR, as most of it was merged in [this](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=aad4b3e0fe8379f6fa34c7b48b4584590025760c) commit.
Also, I'm taking maintenance on this package.

PR #21918, opened 2021-08-08
sci-electronics/kicad: Fixup ebuild to support opencascade-7.5.2

zpuskas:


sci-electronics/kicad/kicad-5.1.10-r2.ebuild
```diff
@@ -114,7 +114,12 @@ src_configure() {
 		-DPYTHON_LIBRARY="$(python_get_library_path)"
 	)
 	if use occ; then
-		if has_version ">=sci-libs/opencascade-7.5"; then
+		if has_version ">=sci-libs/opencascade-7.5.2"; then
```
ionenwks, line 117: These `>=` don't mean much if it only works for "this" version. You can use `~` (aka `~cat/pkg-7.5.2`) to say "this version at any revision", and then limit in *DEPEND or else it will break again next opencascade version rather than say "don't upgrade opencascade yet".

For RDEPEND, given all slots are fine, feels kind of bad but guess can do:
`occ? ( <sci-libs/opencascade-7.5.3:=[vtk(+)] )` (not <= to ensure it pickups potential 7.5.2 revisions)
Given there's a := operator and >=7.5.2 couldn't build, a revbump shouldn't be necessary despite RDEPEND change.

I'd like to think there's a better way to do this, but haven't explored opencascade/kicad stuff to suggest anything, for now it's better than nothing anyway.

zpuskas: Thanks for the suggestions. I'll apply them and revert back to -r1.

OpenCascade has this issue, I'm not sure if any of the packages using it found a better solution to this so far.

PR #21919, opened 2021-08-08
app-misc/piper: Remove flake8 test

kwhat:
Closes: https://bugs.gentoo.org/807205
Signed-off-by: Alexander Barker <alex@1stleg.com>

### DIFF
--- a/app-misc/piper/files/piper-0.5.1-disable-flake8-linting.patch
+++ b/app-misc/piper/files/piper-0.5.1-disable-flake8-linting.patch
@@ -1,0 +1,17 @@
+Remove linting test that is broken upstream.
+Bug: https://bugs.gentoo.org/807205
+
+--- piper-0.5.1/meson.build
++++ piper-0.5.1/meson.build
+@@ -119,11 +119,3 @@
+                configuration: config_piper_devel)
+ 
+ meson.add_install_script('meson_install.sh')
+-
+-flake8 = find_program('flake8', required: false)
+-if flake8.found()
+-    test('flake8', flake8,
+-         args: ['--ignore=E501,W504',
+-                join_paths(meson.source_root(), 'piper'),
+-                join_paths(meson.source_root(), 'piper.in')])
+-endif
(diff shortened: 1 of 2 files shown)


app-misc/piper/piper-0.5.1-r2.ebuild
```diff
@@ -40,6 +40,8 @@ DEPEND="
 	virtual/libudev
 "
 
+PATCHES=( "${FILESDIR}"/${P}-disable-flake8-linting.patch )
```
ionenwks, line 43: nitpick but even for 1 patch I tend to prefer patch on its own line (makes it quicker to add/remove patches without restyling everytime), aka:
```suggestion
PATCHES=(
	"${FILESDIR}"/${P}-disable-flake8-linting.patch
)
```

app-misc/piper/files/piper-0.5.1-disable-flake8-linting.patch
```diff
@@ -0,0 +1,14 @@
+--- piper-0.5.1/meson.build	2020-06-22 01:07:53.000000000 -0700
++++ piper-0.5.1/meson.build	2021-08-08 11:46:32.984756474 -0700
+@@ -119,11 +119,3 @@
+                configuration: config_piper_devel)
+ 
+ meson.add_install_script('meson_install.sh')
+-
+-flake8 = find_program('flake8', required: false)
+-if flake8.found()
```
ionenwks, line 9: If making patches that are never going to be upstreamed, you can (optionally) cheat to have shorter patches by (for example) using `if false`

But it's fine if you prefer it like that too.

ionenwks (commented):
Commit message seem to have everything (including close/sign-off) is inside the summary instead of the message body.

app-misc/piper/files/piper-0.5.1-disable-flake8-linting.patch
kwhat: This only results in one line saved for this particular patch, however, I will keep it in mind for larger patches.

PR #21924, opened 2021-08-09
dev-python/treq: enable py3.10, enable tests

arthurzam:
@mgorny Enabling testsuite for `treq` and enable python 3.10 on the way.

Still need to cleanup the ebuild (I want to clean the dependencies) and also need to fix `doc` generation (for now it depends on pre merged self package, but I think I can fix it).

Just in case, there is a need to test that it really works without network (the comment is misleading).

Signed-off-by: Arthur Zamarin <arthurzam@gmail.com>

### DIFF
--- a/dev-python/treq/treq-21.5.0.ebuild
+++ b/dev-python/treq/treq-21.5.0.ebuild
@@ -25,20 +25,16 @@ RDEPEND="
 	>=dev-python/twisted-18.7.0[crypt,${PYTHON_USEDEP}]
 "
 BDEPEND="
-	doc? ( dev-python/sphinx
-		${RDEPEND} )
 	test? (
 		dev-python/httpbin[${PYTHON_USEDEP}]
 	)"
 
-python_compile_all() {
-	use doc && emake -C "${S}/docs" html
-}
+distutils_enable_sphinx docs
 
-python_install_all() {
-	use doc && HTML_DOCS=( docs/_build/html/ )
-
-	distutils-r1_python_install_all
+src_prepare() {
+	# fix relative path for docs generation
+	sed -e "s@('..')@('../src')@" -i docs/conf.py || die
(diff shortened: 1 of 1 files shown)


dev-python/treq/treq-21.5.0.ebuild
```diff
@@ -47,17 +47,7 @@ python_install_all() {
...
-	ewarn "as a user for each of the python versions it is installed to"
-	ewarn "to correctly test this package."
-}
-
 python_test() {
-	# Tests fail when run via emerge
-	# they need proper network access
-	test_instructions
+	distutils_install_for_testing
+	"${EPYTHON}" trial treq || die "Tests failed with ${EPYTHON}"
```
mgorny, line 52:
```suggestion
	"${EPYTHON}" -m trial treq || die "Tests failed with ${EPYTHON}"
```

?

mgorny: Nah, i am wrong.

dev-python/treq/treq-21.5.0.ebuild
```diff
@@ -17,47 +17,28 @@ KEYWORDS="~amd64 ~arm64 ~x86"
...
-	>=dev-python/twisted-18.7.0[crypt,${PYTHON_USEDEP}]
 	>=dev-python/requests-2.1.0[${PYTHON_USEDEP}]
-	dev-python/attrs[${PYTHON_USEDEP}]
+	>=dev-python/twisted-18.7.0[crypt,${PYTHON_USEDEP}]
 "
 
-DEPEND="${COMMON_DEPEND}
-	doc? ( dev-python/sphinx
-		${RDEPEND} )
+DEPEND="
```
mgorny, line 27:
```suggestion
BDEPEND="
```

PR #21925, opened 2021-08-09
dev-python/pipenv: fixed vendored packages

oz123:
* Fix missing deps on jinja and attr
 * Remove vendored colorama, add dep on system colorama
 * Enable Python-3.10

Bug: https://bugs.gentoo.org/show_bug.cgi?id=717666


dev-python/pipenv/pipenv-2021.5.29-r1.ebuild
```diff
@@ -4,7 +4,7 @@
 EAPI=7
 
 DISTUTILS_USE_SETUPTOOLS=rdepend
-PYTHON_COMPAT=( python3_{7,8,9} )
+PYTHON_COMPAT=( python3_{7,8,9,10} )
```
ionenwks, line 7: If going to touch this, remove 7 at same time. It's dead code that does nothing (not recognized by eclasses).

If adding 10, I assume you've ran the test suite with it?

dev-python/pipenv/pipenv-2021.5.29-r1.ebuild
```diff
@@ -20,11 +20,17 @@ KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
...
 
-PATCHES="${FILESDIR}/${PN}-${PV//./-}-${PR}-remove-attr-vendor-import.patch"
+PATCHES=( 
+	"${FILESDIR}/${PN}-${PV//./-}-${PR}-remove-attr-vendor-import.patch"
+	"${FILESDIR}/${PN}-${PV//./-}-${PR}-remove-colorama-vendor-import.patch"
+	)
 
 RDEPEND="
 	${PYTHON_DEPS}
+	dev-python/attr[${PYTHON_USEDEP}]
```
ionenwks, line 30: As CI say, we have no `dev-python/attr` in the tree.

But we do have a `dev-python/attrs`, I think it's the right thing? Please double check.

dev-python/pipenv/pipenv-2021.5.29-r1.ebuild
```diff
@@ -20,11 +20,17 @@ KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
...
 	dev-python/certifi[${PYTHON_USEDEP}]
+	>=dev-python/colorama-0.4.4[${PYTHON_USEDEP}]
+	>=dev-python/jinja-3.0.1[${PYTHON_USEDEP}]
```
ionenwks, line 33: Adding new RDEPEND requires a revbump.

oz123: OK, will change this, to be a new version.

dev-python/pipenv/pipenv-2021.5.29-r1.ebuild
oz123: Darn, the package name is different from the module imported. Sorry about it.

dev-python/pipenv/pipenv-2021.5.29-r1.ebuild
oz123: What do you mean by that? I don't understand.
Is Python3_7 dead or Python3_10?

oz123: I ran the tests with Python3_10, they work.

ionenwks: 3_7 is dead as far as eclasses are concerned, so can cleanup

3_10 I'm just asking if you've tested it or just adding it like that

mgorny, line 7:
```suggestion
PYTHON_COMPAT=( python3_{8..10} )
```

The `..` syntax is not obligatory but IMO it's nicer.

oz123: I agree :-)

oz123: Removed Python3.7

juippis (approved):

PR #21929, opened 2021-08-09
Mask net-print/libgnomecups for removal

a17r:
Bug: https://bugs.gentoo.org/352952


leio (approved):
I guess in theory you need a bug to track this, where any objections can be raised?

a17r:
> I guess in theory you need a bug to track this, where any objections can be raised?

...the old one will suffice...

PR #21932, opened 2021-08-09
media-gfx/gscan2pdf: version bump to 2.12.2 and EAPI 8 and remove old 2.11.1

cjmayo:
* 2.12 needs a newer dev-perl/PDF-Builder
* ultimate guide to EAPI 8 showed dev-util/desktop-file-utils needed for xdg_desktop_database_update


media-gfx/gscan2pdf/gscan2pdf-2.12.2.ebuild
```diff
@@ -0,0 +1,112 @@
...
+		app-text/djvu[jpeg,tiff]
+		app-text/poppler[utils]
+		app-text/tesseract[-opencl,osd(+),png,tiff]
+		app-text/unpaper
+		media-gfx/imagemagick[djvu,jpeg,png,tiff,perl,postscript]
+		media-gfx/sane-backends[sane_backends_test]
+		media-gfx/sane-frontends
+	)"
+
+IDEPEND="dev-util/desktop-file-utils"
```
ionenwks, line 68: This is actually not *really* needed and ebuilds normally don't add these deps unless needed for something else like tests.

xdg-utils is designed to do nothing if the command isn't found and that's actually fine because that means a cache update is not needed "yet" -- it will be updated later given cache is done "all at once" and will include this.

It's a simple example of what IDEPEND could be used for though, thus the note in the guide.

media-gfx/gscan2pdf/gscan2pdf-2.12.2.ebuild
```diff
@@ -0,0 +1,112 @@
...
+	optfeature "scan post-processing" app-text/unpaper
+	optfeature "automatic document feeder support" media-gfx/sane-frontends
+	optfeature "sending PDFs as email attachments" x11-misc/xdg-utils
+}
+
+pkg_postrm() {
+	xdg_desktop_database_update
+}
+
+src_test() {
```
ionenwks: It's nice to respect phase ordering in ebuilds, aka same order that they're ran in by emerge.
https://devmanual.gentoo.org/ebuild-writing/functions/index.html

So this src_test should be before pkg_postinst in this ebuild.

media-gfx/gscan2pdf/gscan2pdf-2.12.2.ebuild
```diff
@@ -0,0 +1,112 @@
...
+	echo "Using:"
+	echo "  $(best_version app-text/djvu)"
+	echo "  $(best_version app-text/poppler)"
+	echo "  $(best_version app-text/tesseract)"
+	echo "  $(best_version dev-perl/Gtk3-ImageView)"
+	echo "  $(best_version dev-perl/Image-Sane)"
+	echo "  $(best_version dev-perl/PDF-Builder)"
+	echo "  $(best_version media-gfx/imagemagick)"
+	echo "  $(best_version media-gfx/sane-backends)"
+	echo "  $(best_version media-libs/tiff)"
```
ionenwks: Be nice to use `einfo` over echo here. einfo exists to indicate a message is coming from the ebuild, yet still without spamming user logs (unlike ewarn/elog).

media-gfx/gscan2pdf/gscan2pdf-2.12.2.ebuild
```diff
@@ -0,0 +1,112 @@
...
+	optfeature "adding to an existing PDF" app-text/poppler[utils]
+	optfeature "Optical Character Recognition" app-text/tesseract[tiff]
+	optfeature "scan post-processing" app-text/unpaper
+	optfeature "automatic document feeder support" media-gfx/sane-frontends
+	optfeature "sending PDFs as email attachments" x11-misc/xdg-utils
+}
+
+pkg_postrm() {
+	xdg_desktop_database_update
+}
```
ionenwks: I'd like to say to just use xdg.eclass instead of xdg-utils so can remove things like this, but xdg.eclass is still EAPI-7 only at the moment (the bump is undergoing review on dev ML right now).

Just making a note of it, unless want to go back to EAPI-7 for now.

media-gfx/gscan2pdf/gscan2pdf-2.12.2.ebuild
cjmayo: I don't follow: xdg.eclass would add a DEPEND on desktop-file-utils and this update would always get done?

media-gfx/gscan2pdf/gscan2pdf-2.12.2.ebuild
cjmayo: Changed to einfo.

media-gfx/gscan2pdf/gscan2pdf-2.12.2.ebuild
cjmayo: Moved src_test up.

media-gfx/gscan2pdf/gscan2pdf-2.12.2.ebuild
ionenwks: These updates aren't package specific, i.e. when a package updates cache it updates any missing entries.

And if the tools aren't installed yet, then nothing is using this cache and updating it doesn't matter. Once it is, cache will be updated (and will include this package).

So xdg.eclass' DEPEND was never really relevant, and there's especially no need to go out of your way to add this dependency manually.

cjmayo: Finally I get it. Thanks for pointing out and explaining. Removed in a new patch (can squash if preferred).

PR #21934, opened 2021-08-09
net-misc/xmrig: Move optional rdep to optfeature

00-matt:


candrews:
The systemd service would also have to be updated to make msr-tools usage optional.

00-matt:
Thanks for catching that! I didn't look too deeply into the XMRig service because I don't use it.

Do you think it's patching `randomx_boost.sh` to check for the existence of `/usr/sbin/rdmsr` and `/usr/sbin/wrmsr` before running would be enough?

candrews:
Perhaps?

But what's the motivation? What benefit does the user get by having this dependency be optional? I feel like all users would benefit from this dependency.

00-matt:
Sorry, I mis-remembered the devmanual.  I thought it said to never put optional runtime deps in RDEPENDS, but it actually just says not to make USE flags for them.

PR #21939, opened 2021-08-10
sys-libs/musl: exit early in ldconfig if nothing has changed

yemouu:
This fixes an access violation when compiling sys-libs/ncurses
and /usr/bin/ld is not ld.bfd.

Closes: https://bugs.gentoo.org/719330
Signed-off-by: Kofi Hannam <meeyou@tuta.io>


yemouu:
Whoops, I linked the wrong bug on accident.

PR #21940, opened 2021-08-10
app-editors/vscode: version bump: 1.59.0

Kangie:
Signed-off-by: Matt Jolly <Matt.Jolly@footclan.ninja>


AdelKS:
Hello! Thank you for your help, can you bump vscodium too ? I am away from my computer for one more week and will not be able to do it untill then.

juippis:
I've heard some trouble with sandbox and either vscode, or vscodium, or perhaps both. We do 
`fperms u+s /opt/program/chrome-sandbox` on many of these electron-based apps. Can you check whether that binary exists, and use the fperms above?

AdelKS:
Yes that binary does exist and currently it has `fperms +x /opt/${PN}/chrome-sandbox`, so you suggest putting `fperms u+s /opt/${PN}/chrome-sandbox` ? I did get some feedback about this indeed. I still don't understand why it works on my machine. Can you try this @Kangie ?

Thank you!

Kangie:
@AdelKS All seems good here; both launch and behave as I expect.

AdelKS:
Awesome thanks!

PR #21941, opened 2021-08-10
media-gfx/imv: Remove obsolete libOpenGL modification

lanodan:
Related: https://github.com/eXeC64/imv/pull/285


ionenwks:
I am aware of that this was merged, but it's automagic, e.g. it uses libGL if it's found rather than libOpenGL and that'd require me to depend on libglvnd[X] even with -X. By forcing libOpenGL that's not a problem.

PR #21942, opened 2021-08-10
dev-python/cloudscraper: Add nodejs as a test dependency

vaartis:
Discovered by tinderbox: https://bugs.gentoo.org/807454


sbraz:
Hi,
Can you please remove $RDEPEND from $BDEPEND? It is only required for tests and distutils_enable_tests adds it automatically.

sbraz:
And you need to add a line to `arch/amd64/x32/package.use.mask`.

sbraz:
Thanks, I took the liberty of merging both commits and adding the mask amongst the other similar masks.

PR #21946, opened 2021-08-10
media-tv/xmltv: Add missing depend

thebitpit:
Closes: https://bugs.gentoo.org/807577


thebitpit:
This PR also removes obsolete xmltv-0.5.70.ebuild that repoman does not like.

thesamesam (approved):
Thanks! I've dropped -r0 (0.6.1) and pushed this straight to stable as EAPI 7 because we want everyone to pick up the fix.

PR #21951, opened 2021-08-11
media-fonts/terminus-font: fix double prefix

xdch47:


mkszuba:
Good catch, thanks.

PR #21955, opened 2021-08-12
riscv: keyword kde-plasma/plasma-meta net-firewall/ufw

alexfanqi:
tested on qemu riscv64, sifive unleash board. works okay

Note: kde requires opengl support, otherwise there will be segment fault.


a17r (commented):
keywording `media-libs/phonon` without either `media-libs/phonon-vlc` or `media-libs/phonon-gstreamer` makes no sense. `media-libs/libcanberra` is already keyworded for riscv and would have sufficed for functionality by masking `phonon` IUSE.

PR #21961, opened 2021-08-12
net-im/telegram-desktop: Fix bugs and version bump

mid-kid:
@gyakovlev

This change is a bit bigger than usual, given I'm updating the `*DEPEND`, adding some USE flags and disabling the audio backends of tg\_owt.

I hope this gets rid of ABI compatibility issues and whatnot.


gyakovlev:
merged, thanks. tested new version, seems to be ok.

PR #21965, opened 2021-08-12
media-tv/xmltv: bump xmltv-1.0.0

thebitpit:


thebitpit:
There were not a lot of changes, mostly deletions to match upstream. 
Bump to EAPI=8, but will go back to 7 if requested.

Please review and merge if approved.

thebitpit:
Added a warning about required Perl module not in portage.
Can't make xmltv-1.0.0.ebuild depend on the CPAN Linux::DVB module.
Tried g-cpan, but it fails and is probably the wrong way to do it.
The canary indicated if may not be compatible with the version of Perl on my system.

juippis (approved):
You can add the missing deps if you see them worthy.

PR #21968, opened 2021-08-12
sci-libs/coinor-clp: update EAPI 7 -> 8

Alessandro-Barbieri:
Restore the old mumps/MPI behaviour
https://github.com/coin-or/Clp/issues/199

### DIFF
--- a/sci-libs/coinor-clp/coinor-clp-1.17.6-r1.ebuild
+++ b/sci-libs/coinor-clp/coinor-clp-1.17.6-r1.ebuild
@@ -1,0 +1,143 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit toolchain-funcs
+
+MY_PN=Clp
+
+DESCRIPTION="COIN-OR linear programming solver"
+HOMEPAGE="https://github.com/coin-or/Clp"
+SRC_URI="https://github.com/coin-or/Clp/archive/releases/${PV}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/${MY_PN}-releases-${PV}/${MY_PN}"
+
+LICENSE="EPL-1.0"
+SLOT="0/1" # major soname component
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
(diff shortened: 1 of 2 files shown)


sci-libs/coinor-clp/coinor-clp-1.17.6-r1.ebuild
```diff
@@ -0,0 +1,141 @@
...
+
+RESTRICT="!test? ( test )"
+
+src_prepare() {
+	# Needed to make the --with-coin-instdir in src_configure happy.
+	dodir /usr
+
+	mumpslibs="-lmumps_common -ldmumps -lzmumps -lsmumps -lcmumps"
+
+	if has_version sci-libs/mumps[-mpi]; then
```
SoapGentoo, line 49: if you flip `USE=mpi` on `sci-libs/mumps`, then this package will break?

Alessandro-Barbieri: I've tried to test this thing in a clean chroot and I discovered that mumps is severely underlinked. We need to fix it before trying again

Alessandro-Barbieri: If you change the mpi use on mumps you also need to rebuild coinor-clp.
Can this be achieved by using `mumps:=`?

ionenwks: USE-based subslot rebuilds been discussed but aren't a thing right now.

Haven't looked at this package's situation, but the idea would generally be to add a USE=mpi, and depend on `mumps[mpi?]`, then check for `use mpi` rather than do has_version checks.

ionenwks: Or actually, I think you may need mumps[mpi=] to enforce match.

Alessandro-Barbieri: > add a USE=mpi, and depend on `mumps[mpi?]`, then check for `use mpi` rather than do has_version checks.

I like this idea

Alessandro-Barbieri:
Ping

AndrewAmmerlaan:
Thanks :+1:

PR #21970, opened 2021-08-13
*/*: bump to python 3.10

arthurzam:
@mgorny Small list of bumps toward python 3.10

1. `app-text/htmlmin` - bumped and passed tests. Also cleaned `DISTUTILS_USE_SETUPTOOLS` and put `ALLARCHES` (tested with and without cython, ran multiple tests and scanned the build system).
2. `dev-python/plumbum` - changed `SRC_URI` to use pypi before redirection (cleaner) and tested multiple times and by removing distfile, that the checksum is the same. Then did a py3.10 bump and dep cleaning, and also cleanup of the deselection of tests (so now we can remove the patch).
3. `dev-util/flawfinder` - bumped version and py3.10, and passed tests and put `ALLARCHES`.
4. `app-text/recode` - bumped and passed tests. Also fixed the license field (based on upstream commit change and scanning of source file headers).


mgorny (approved):
Thanks!

PR #21971, opened 2021-08-13
net-mail/mu: Bump to 1.6.3

00-matt:


net-mail/mu/mu-1.6.3.ebuild
```diff
@@ -0,0 +1,64 @@
...
+	dev-libs/glib:2=
+	dev-libs/gmime:3.0=
+	>=dev-libs/xapian-1.4:=
+	emacs? ( >=app-editors/emacs-25.3:* )
+	guile? ( >=dev-scheme/guile-2.2:* )
+	mug? (
+		net-libs/webkit-gtk:4=
+		x11-libs/gtk+:3=
+	)
+	readline? ( sys-libs/readline )"
```
thesamesam, line 27: If you link to it and it has a subslot, in general, you should subscribe to it

net-mail/mu/mu-1.6.3.ebuild
```diff
@@ -0,0 +1,64 @@
...
+HOMEPAGE="https://www.djcbsoftware.nl/code/mu/ https://github.com/djcb/mu"
+SRC_URI="https://github.com/djcb/mu/releases/download/${PV}/mu-${PV}.tar.xz"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~x86 ~x64-macos"
+IUSE="emacs guile mug readline"
+
+DEPEND="
+	dev-libs/glib:2=
```
thesamesam, line 18: Neither glib nor gmime define subslots, so let's drop the = part.

PR #21980, opened 2021-08-13
sci-* EAPI-5 cleanup

jsmolic:


sci-chemistry/bodr/bodr-10.ebuild
```diff
@@ -10,14 +10,11 @@ SRC_URI="mirror://sourceforge/${PN}/${P}.tar.bz2"
...
-DEPEND="dev-libs/libxslt"
-RDEPEND=""
-
-DOCS=( "THANKS" "AUTHORS" "README" "ChangeLog" )
+BDEPEND="dev-libs/libxslt"
 
 src_prepare() {
+	default
 	sed -i -e "s/COPYING//g" Makefile.* || die
 }
```
thesamesam, line 20: We can drop the [docdir](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-128001r8) in newer EAPIs (yay!)

sci-libs/qrupdate/qrupdate-1.1.2-r2.ebuild
```diff
@@ -0,0 +1,50 @@
...
+}
+
+src_compile() {
+	emake solib
+}
+
+src_install() {
+	emake DESTDIR="${D}" install-shlib
+	dosym libqrupdate.so.$(ver_cut 1) /usr/$(get_libdir)/libqrupdate.so
+	dodoc README ChangeLog
```
SoapGentoo, line 49: check whether `einstalldocs` maybe also installs additional stuff?

SoapGentoo (approved):

sci-libs/qrupdate/qrupdate-1.1.2-r2.ebuild
jsmolic: nope, only readme and changelog

sci-chemistry/bodr/bodr-10.ebuild
jsmolic: Perfect :smiley:

PR #21982, opened 2021-08-13
app-emulation/firecracker-bin: Bump to 0.25.0, drop old versions

s-hamann:
Trivial version bump to firecracker 0.25.0. I also dropped all but the latest version of the 0.24.x branch.


juippis (approved):

juippis:
But in future please separate bumping and removing version to their own respective commits.

PR #21984, opened 2021-08-13
sys-libs/ncurses: Use `BUILD_CC` as `CC` when compiling tic.

vaibhav-rustagi:
Not having `CC` explictly set creates a problem in case of
cross-compilation as it can use host `CC` for tic.

Signed-off-by: Vaibhav Rustagi <vaibhavrustagi@google.com>


SoapGentoo:
can you rebase this on the latest revbump?

thesamesam:
Thanks! I’ve only added it to the newer version as we’ll be stabling it shortly anyway and I was reluctant to change the existing stable version.

PR #21985, opened 2021-08-14
app-crypt/gcr: rebuild pkg on wayland use change

jayofdoom:
If this package is not rebuilt after a GTK+ is rebuilt with a wayland
use flag change, packages that depend on it will fail to build.

This is related to the larger issue tracked in bug #624960.

Bug: https://bugs.gentoo.org/624960
Closes: https://bugs.gentoo.org/808081
Signed-off-by: Jay Faulkner <jay@jvf.cc>

### DIFF
--- a/app-crypt/gcr/gcr-3.40.0.ebuild
+++ b/app-crypt/gcr/gcr-3.40.0.ebuild
@@ -13,7 +13,7 @@ HOMEPAGE="https://gitlab.gnome.org/GNOME/gcr"
 LICENSE="GPL-2+ LGPL-2+"
 SLOT="0/1" # subslot = suffix of libgcr-base-3 and co
 
-IUSE="gtk gtk-doc +introspection test +vala"
+IUSE="gtk gtk-doc +introspection test +vala wayland"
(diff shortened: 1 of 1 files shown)


jayofdoom:
@thesamesam I believe this is the fix we talked about in IRC. Question though -- I wasn't sure if I should change this ebuild inline or create a new separate ebuild and add `-r1` to the version.

leio:
Requiring USE=wayland on gtk with USE=-wayland on gcr is wrong, they don't have to be locked like this. Bug 624960 is mostly unsolved because there's no good way to solve it. Doing it this way just introduces other kind of problems and wrongly forces a matching wayland USE flag match. If you do this to everything that could potentially need it, it's going to be near impossible to toggle it during an upgrade cleanly, etc.

jayofdoom:
Would you consider a PR adding an elog to gui-libs/gtk, warning of the perils of changing the state of the wayland use flag? A short blurb and a link back to the bug would likely save someone troubleshooting time it the future -- it would've for me.

PR #21986, opened 2021-08-14
dev-python/uvicorn: bump w/ new deps [WIP]

mgorny:
CC @arthurzam, if you find a minute, this needs pydantic py3.10 + figuring out test failures in uvicorn (might be tmpfs-related).


arthurzam:
@mgorny Managed to bump pydantic (using snapshot) and added missing test dependencies and enabled most of tests.
I have cherry picked your commits here, and the next two packages passes (`dev-python/pytest-toolbox` needed extra dependency, and `dev-python/watchgod` just passes).

I'm still unable to pass `dev-python/uvicorn`, and I'm testing where is the issue. We do need to deselect the following test because of too long path for unix socket:
`tests/test_config.py::test_bind_unix_socket_works_with_reload_or_workers`
I think the issue arises around handling websockets, and I think somehow around the event loop of `asyncio`. Will investigate more tomorrow.

For now, you can find all my commits (can't push to your branch as I'm not a member) in the following branch:
https://github.com/arthurzam/gentoo/tree/py3.10-unicorn

arthurzam:
All have been merged/fixed in tree.

PR #21987, opened 2021-08-14
sys-fs/cryptsetup: wait for source to appear

Cynerd:
Devices that are primarily hotplug are commonly initialized little bit
later when init is already started. This can be either because they
require module or because they are just slow to initialize or because
they are just last in the line. No matter the reason the expectation
that source is always immediately available to open does not apply.

Real live example is all time plugged in SD card on some embedded
devices serving as extension storage. It is removable and thus it is
highly desirable to encrypt it but it is initialized asynchronously thus
it is not available immediately after boot.

This adds new argument `wait`. It is in default set to 5 seconds. It can
be set to zero for devices that are not expected to be present during
the boot process (although such usage seems to me like a hack). Any
other device should be present and thus wait is skipped or it benefits
from wait.

Signed-off-by: Karel Kočí <cynerd@email.cz>

### DIFF
--- a/sys-fs/cryptsetup/cryptsetup-2.4.0_rc1-r100.ebuild
+++ b/sys-fs/cryptsetup/cryptsetup-2.4.0_rc1-r100.ebuild
@@ -1,0 +1,133 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools linux-info tmpfiles
+
+DESCRIPTION="Tool to setup encrypted devices with dm-crypt"
+HOMEPAGE="https://gitlab.com/cryptsetup/cryptsetup/blob/master/README.md"
+SRC_URI="https://www.kernel.org/pub/linux/utils/${PN}/v$(ver_cut 1-2)/${P/_/-}.tar.xz"
+
+LICENSE="GPL-2+"
+SLOT="0/12" # libcryptsetup.so version
+[[ ${PV} != *_rc* ]] && \
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86"
+CRYPTO_BACKENDS="gcrypt kernel nettle +openssl"
+# we don't support nss since it doesn't allow cryptsetup to be built statically
+# and it's missing ripemd160 support so it can't provide full backward compatibility
+IUSE="${CRYPTO_BACKENDS} +argon2 nls pwquality reencrypt ssh static static-libs +udev urandom"
+REQUIRED_USE="^^ ( ${CRYPTO_BACKENDS//+/} )
+	static? ( !gcrypt )" #496612
+
+LIB_DEPEND="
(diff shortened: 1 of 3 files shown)


thesamesam:
@Polynomial-C @robbat2

floppym:
Here's the current diff for easier review:
```diff
--- 1.6.7-dmcrypt.confd 2015-08-09 01:39:39.419164109 -0400
+++ 2.4.0-dmcrypt.confd 2021-08-16 11:16:38.893125736 -0400
@@ -52,6 +52,7 @@
 #                                    be looked up automatically.
 # pre_mount='cmds'                   == commands to execute before mounting partition.
 # post_mount='cmds'                  == commands to execute after mounting partition.
+# wait=5                             == wait given amount of seconds for source to appear
 #-----------
 # Supported Modes
 # gpg                                  == decrypt and pipe key into cryptsetup.
--- 1.6.7-dmcrypt.rc    2017-02-28 16:28:17.451673539 -0500
+++ 2.4.0-dmcrypt.rc    2021-08-16 11:16:38.893125736 -0400
@@ -3,6 +3,7 @@
 # Distributed under the terms of the GNU General Public License v2

 depend() {
+       use modules
        before checkfs fsck

        if grep -qs ^swap= "${conf_file}" ; then
@@ -21,7 +22,7 @@
 # Setup mappings for an individual target/swap
 # Note: This relies on variables localized in the main body below.
 dm_crypt_execute() {
-       local dev ret mode foo
+       local dev ret mode foo source_dev

        if [ -z "${target}" -a -z "${swap}" ] ; then
                return
@@ -31,6 +32,7 @@
        : ${dmcrypt_key_timeout:=1}
        : ${dmcrypt_max_timeout:=300}
        : ${dmcrypt_retries:=5}
+       : ${wait:=5}

        # Handle automatic look up of the source path.
        if [ -z "${source}" -a -n "${loop_file}" ] ; then
@@ -38,7 +40,16 @@
        fi
        case ${source} in
        *=*)
-               source=$(blkid -l -t "${source}" -o device)
+               i=0
+               while [ ${i} -lt ${wait} ]; do
+                       if source_dev="$(blkid -l -t "${source}" -o device)"; then
+                               source="${source_dev}"
+                               break
+                       fi
+                       : $((i += 1))
+                       einfo "waiting for source \"${source}\" for ${target}..."
+                       sleep 1
+               done
                ;;
        esac
        if [ -z "${source}" ] || [ ! -e "${source}" ] ; then
@@ -266,10 +277,10 @@
                        dm_crypt_execute

                        # Prepare for the next target/swap by resetting variables
-                       unset gpg_options key loop_file target options pre_mount post_mount source swap remdev
+                       unset gpg_options key loop_file target options pre_mount post_mount source swap remdev wait
                        ;;

-               gpg_options=*|remdev=*|key=*|loop_file=*|options=*|pre_mount=*|post_mount=*|source=*)
+               gpg_options=*|remdev=*|key=*|loop_file=*|options=*|pre_mount=*|post_mount=*|wait=*|source=*)
                        if [ -z "${target}${swap}" ] ; then
                                ewarn "Ignoring setting outside target/swap section: ${targetline}"
                                continue

```

PR #21990, opened 2021-08-14
Remove last-rited ruby packages

jsmolic:


SoapGentoo:
@graaff ok do go ahead?

ajakk:
Thank you for doing this!

graaff:
This pull request seems to mix up different things. Removal of last-rited packages should not be in the same pull request as removing ruby25 keywords. 

As for the last-rites removal: please exclude dev-lang/ruby, virtual/rubygems, and similar packages. These (may) need additional handling in terms of profile or eclass changes, so this is not a straightforward removal.

As for the ruby25 keyword removal: this brings no benefit that I can see and make things harder for people who are last to upgrade, so I won't merge such a change.

jsmolic:
@graaff Ok, done, I hope I didn't cause a lot of confusion by putting `ruby25` here, I'm just used to combining multiple things in single PR (makes it easier for devs that merge PRs).
Anyway, last-rites are here and as for removing old ruby versions I don't know how it's usually done  so I will leave it to you but I'm always happy to help out if needed

PR #21992, opened 2021-08-14
media-plugins/vdr-*: pkgcheck fixes on misc. vdr plugins

madmartin:
fix some packages where pkgcheck reports `HttpsUrlAvailable`, and update their copyright headers


juippis (approved):
You should definitely start looking into bumping those EAPI-5 ebuilds to modern age.

PR #21993, opened 2021-08-15
Odamex 0.9.4

vilhelmgray:
Version bump to 0.9.4 and drop old 0.9.2

Closes: https://bugs.gentoo.org/808255


games-engines/odamex/odamex-0.9.4.ebuild
```diff
@@ -21,12 +21,12 @@ RDEPEND="
 		media-libs/libpng:0=
 		media-libs/libsdl2[joystick,sound,video]
 		media-libs/sdl2-mixer
+		net-misc/curl
 		odalaunch? ( x11-libs/wxGTK:${WX_GTK_VER}[X] )
 		portmidi? ( media-libs/portmidi )
 		X? ( x11-libs/libX11 )
 	)
 	server? (
-		dev-libs/jsoncpp:=
```
chewi, line 29: What's the deal here? They're still bundling it but they've changed it to an external CMake project. Is it custom-patched now?

ionenwks: It bundles ancient jsoncpp from 2010, and unless start rewriting code it won't be able to use the system's (too new).

And the dep was just wrong before (unused).

ionenwks: Well, if really wanted to, could add a new jsoncpp slot to the tree for deprecated jsoncpp-0.10.2, but not convinced it's worth it for 1 consumer and been leaving this one alone.

PR #21994, opened 2021-08-15
[media-gfx/blender, media-libs/oidn] Bump versions and provide glibc fix for older oidn versions.

DarkDefender:
This bumps the blender/odin version and provides a backported glibc fix for the 1.3.0 release.
Closes: https://bugs.gentoo.org/806514


media-gfx/blender/blender-2.83.17.ebuild
```diff
@@ -0,0 +1,380 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8} )
```
thesamesam, line 6: Python 3.7 isn't supported by the eclass anymore, so we can drop this (it's a noop).

media-gfx/blender/blender-2.83.17.ebuild
```diff
@@ -0,0 +1,380 @@
...
+src_unpack() {
+	if [[ ${PV} = *9999* ]] ; then
+		git-r3_src_unpack
+	else
+		default
+	fi
+
+	if use test; then
+		mkdir -p lib || die
+		mv "${WORKDIR}"/blender-${TEST_TARBALL_VERSION}-tests/tests lib || die
```
thesamesam, line 158: We still need to do something like moving this into ${T} instead IIRC, I think, because Portage will become stricter in future.

media-gfx/blender/blender-2.83.17.ebuild
```diff
@@ -0,0 +1,380 @@
...
+		-DWITH_SDL=$(usex sdl)
+		-DWITH_STATIC_LIBS=OFF
+		-DWITH_SYSTEM_EIGEN3=ON
+		-DWITH_SYSTEM_GLEW=ON
+		-DWITH_SYSTEM_LZO=ON
+		-DWITH_TBB=$(usex tbb)
+		-DWITH_USD=OFF
+	)
+	if ! use debug ; then
+		append-flags  -DNDEBUG
```
thesamesam, line 255:
```suggestion
		append-flags -DNDEBUG
```

media-gfx/blender/blender-2.83.17.ebuild
```diff
@@ -0,0 +1,380 @@
...
+	else
+		append-flags  -DDEBUG
```
thesamesam, line 257:
```suggestion
		append-flags -DDEBUG
```

... but instead, I'd do:
```suggestion
	append-cppflags $(usex debug '-DDEBUG' '-DNDEBUG')
```

thesamesam (commented):
Minor comments but not blockers.

media-libs/oidn/oidn-1.4.1.ebuild
```diff
@@ -0,0 +1,39 @@
...
+IUSE=""
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+RDEPEND="
+	${PYTHON_DEPS}
+	dev-cpp/tbb
+	dev-lang/ispc"
+DEPEND="
+	${RDEPEND}
+	dev-util/cmake"
```
thesamesam, line 33: BDEPEND, as we execute it on CBUILD.

media-libs/oidn/oidn-1.4.1.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+CMAKE_BUILD_TYPE=Release
```
thesamesam, line 35: We usually prefer not to have to set this (instead patch the CMakeLists.txt file(s) to be tolerant of other release types) because the Gentoo build type sets a bunch of important variables.

media-libs/oidn/oidn-1.4.1.ebuild
```diff
@@ -0,0 +1,39 @@
...
+	EGIT_REPO_URI="https://github.com/OpenImageDenoise/oidn.git"
+	EGIT_BRANCH="master"
+else
+	SRC_URI="https://github.com/OpenImageDenoise/${PN}/releases/download/v${PV}/${P}.src.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64 ~x86"
+fi
+
+LICENSE="Apache-2.0"
+SLOT="0"
+IUSE=""
```
thesamesam, line 24: We can drop [this](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-620007.3):
```suggestion
```

media-libs/oidn/oidn-1.4.1.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python2_7 python3_{6,7,8,9} )
+
+inherit cmake python-single-r1 eutils
```
thesamesam, line 8: eutils doesn't do anything in EAPI 7+ (you should use the eclasses replacing it) and is obsolete.

```suggestion
inherit cmake python-single-r1
```

media-libs/oidn/oidn-1.4.1.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python2_7 python3_{6,7,8,9} )
```
thesamesam, line 6: Lot of obsolete implementations here (which the eclass ignores anyway):
```suggestion
PYTHON_COMPAT=( python_{8,9} )
```

media-libs/oidn/oidn-1.3.0.ebuild
```diff
@@ -34,6 +34,8 @@ DEPEND="
 
 CMAKE_BUILD_TYPE=Release
 
+PATCHES="${FILESDIR}/glibc-1.3.0.patch"
```
thesamesam, line 37: Always treat PATCHES as an array:

```suggestion
PATCHES=( "${FILESDIR}"/glibc-1.3.0.patch )
```

Note that we normally name patches as ${PN}-versionAppliedToFirst-description.patch.

Looks good to me, although I'll have to mirror the patch on my end as it's a bit large (not your fault! It's best if I do that for now.)

media-libs/oidn/files/glibc-1.3.0.patch
```diff
@@ -0,0 +1,613 @@
+From 4180502f326dccd5e46ab67176b53cc800e32f7b Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: Missing link at the top to the relevant bug in Gentoo + upstream commit. See [here](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#clean-patch-howto).

media-gfx/blender/blender-2.93.2.ebuild
```diff
@@ -0,0 +1,390 @@
...
+	)
+	osl? ( >=media-libs/osl-1.11.10.0 )
+	pdf? ( media-libs/libharu )
+	potrace? ( media-gfx/potrace )
+	pugixml? ( dev-libs/pugixml )
+	pulseaudio? ( media-sound/pulseaudio )
+	sdl? ( media-libs/libsdl2[sound,joystick] )
+	sndfile? ( media-libs/libsndfile )
+	tbb? ( dev-cpp/tbb )
+	test? ( dev-vcs/subversion )
```
thesamesam, line 103: Hm, is the code for this gone now? It's not relevant for non-9999, but in 9999, we should fetch directly from svn, right? I don't see the svn stuff in the github viewer but might be missing something.

(We can conditionally inherit _based on ebuild version_ because it's constant. We can use the same templating for all ebuilds.)

thesamesam (changes requested):

media-gfx/blender/blender-2.83.17.ebuild
DarkDefender: If that is the case, then we need to do more patching of where blender looks for these files.

I guess we can do that when the restriction actually happens?
It sounds like things are still not set in stone, or am I misinterpreting things?

media-libs/oidn/oidn-1.3.0.ebuild
DarkDefender: Renamed the file with the suggested naming scheme :)

media-libs/oidn/oidn-1.4.1.ebuild
DarkDefender: I think I looked into this in the past and gave up because probably at that point it was a PITA to patch this out.
I don't remember the specifics though :C

DarkDefender:
Fixed all comments besides the two more cumbersome ones.
How do you want to proceed?

media-libs/oidn/oidn-1.4.1.ebuild
thesamesam: Look into it next time? 😁

media-gfx/blender/blender-2.83.17.ebuild
thesamesam: Set in stone (the specification leaves them specifically undefined in these phases). Perhaps we could check if it does/does not exist and copy only from ${T} if necessary?

We can still use the same locations, just we'll use an intermediate location.

PR #21995, opened 2021-08-15
Make systemd optional and version bump to 6.44

Role92:
profile-sync-daemon can still be used entirely without systemd, by calling the profile-sync-daemon binary directly instead of using the provided systemd service files:

`profile-sync-daemon sync        # on login`
`profile-sync-daemon resync     # periodically, e.g. via cronjob`
`profile-sync-daemon unsync    # on logout`

The makefile even provides a separate install-target for the systemd service-files.
Thus, I think the ebuild(s) should not have systemd as a hard dependency.

I also added an ebuild for the most recent version 6.44.


www-misc/profile-sync-daemon/profile-sync-daemon-6.35.ebuild
```diff
@@ -11,15 +11,17 @@ KEYWORDS="~amd64 ~x86"
...
 	app-shells/bash
 	net-misc/rsync[xattr]
-	sys-apps/systemd"
+	systemd? ( sys-apps/systemd )"
 
 src_install() {
-	emake DESTDIR="${D}" COMPRESS_MAN=0 install
+	emake DESTDIR="${D}" COMPRESS_MAN=0 install-bin
+	emake DESTDIR="${D}" COMPRESS_MAN=0 install-man
+		"$(usex systemd "install-systemd")"
```
SoapGentoo, line 24: so the only difference is installing systemd unit files?

Role92: Not exactly. I made the systemd dependency optional by adding a USE flag for it. Beforehand it was not possible to install profile-sync-daemon on non-systemd systems. And since the makefile has a separate target for the systemd unit files, I figured those should also be only installed, if the USE=systemd is set.

anaveragehuman: [Small files should be installed unconditionally.](https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0301) So we could remove the dependency on systemd altogether and not need this `usex` line. It would be nice if you could add an example cron file or OpenRC service script.

AndrewAmmerlaan:
Closing as obsolete, latest version in ::gentoo is 6.50

PR #22000, opened 2021-08-15
media-sound/pavucontrol: Bump to version 5.0

i-garrison:
Signed-off-by: Igor V. Kovalenko <igor.v.kovalenko@gmail.com>


media-sound/pavucontrol/pavucontrol-5.0.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+inherit flag-o-matic
```
thesamesam, line 5: This can go now too (because append-cxxflags is gone), I'll drop it for you.

PR #22001, opened 2021-08-15
media-radio/qsstv: Allow apulse as pulseaudio replacement

hamarituc:
This patch allows to build QSSTV with USE="-pulseaudio".

Since I'm not the maintainer of the package I would like @dl1jbe to review and hopefully approve this patch.


ZeroChaos-:
I realize I've been sitting on this for a while, and I apologize for that.  I just had some time to review, and I'm fine with this except pulse should be first instead of second.  Firefox also has pulse first and apulse second so it seems to fit.

If you don't mind making the change I'll pull your commit, otherwise I can make the modification myself and commit it.

ZeroChaos-:
pulled, thanks!

PR #22004, opened 2021-08-16
gui-apps/waybar: Add support for sndio

lanodan:


JonasToth:
Thank you for the Patch!

Please adjust the whitespace in the diff to be consistent.

ionenwks:
> Please adjust the whitespace in the diff to be consistent.

Hm, was just looking over the differences and I don't see what you're referring to. This is only using tabs and they're matching indenting.

JonasToth:
Weird, maybe its a Bug from github on Smartphone



Checked with my laptop and I can't see anything bad either, sorry for the noise

ionenwks (approved):
Ah I see. Well, if that's all there was I assume that means you're fine with it then?

Patch looks simple and builds fine, so I'll go ahead with merging.

PR #22005, opened 2021-08-16
net-wireless/hostapd: Remove vulnerable 2.9-r5; Corretly detect libnl when cross-compiling

jsmolic:
* Build system tries to find libnl only on the host system, which
  breaks cross-compilation as it cannot find it:

```
Package libnl-3.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `libnl-3.0.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libnl-3.0', required by 'virtual:world', not found
```

* We can solve this using pkg-config to pass additional include
  for the cross-compilation libnl to the CFLAGS

* Also, let's remove vulterable 2.9-r5 version for CVE-2021-30004 (https://bugs.gentoo.org/780135)


net-wireless/hostapd/hostapd-2.9-r6.ebuild
```diff
@@ -193,7 +194,11 @@ src_configure() {
 	# Removed for now, since the 3.2 version is broken, and we don't
 	# support it.
 	if has_version ">=dev-libs/libnl-3.2"; then
+		local libnl_include
+
 		echo "CONFIG_LIBNL32=y" >> ${CONFIG} || die
+		libnl_include=$($(tc-getPKG_CONFIG) --cflags libnl-3.0)
+		echo "CFLAGS += ${libnl_include}" >> ${CONFIG} || die
```
thesamesam, line 201: I would've thought we could just do append-cflags?

jsmolic: Sure, let me just test it quickly

jsmolic: Works perfectly :)

thesamesam: Thanks for checking so quickly!

Just so you can internalise the thought process: the idea was, we're modifying CFLAGS, but surely we respect CFLAGS in this package, so let's append it in the normal way and see what breaks. That way, we could just fix the "real" problem (not respecting CFLAGS) if necessary. But tbh, this is one of those packages where the build system is scary and it messes wih your head.

PR #22008, opened 2021-08-17
games-engines/devilutionx: use dynamic linking (fixes libsodium build)

thesamesam:
```-DDIST``` controls whether devilutionx links dynamically to libraries
(only SDL 2 and glibc are linked dynamically otherwise).

For later versions of libsodium, FindSodium.cmake malfunctions and
does not correctly populate all the needed variables. For Gentoo's
purposes, we can ignore this problem, and instead force dynamic
linking (which is policy anyway).

A fix to use a supported method to detect libsodium (pkg-config)
is being pursued upstream.

Bug: https://github.com/diasurgical/devilutionX/issues/2615
Closes: https://bugs.gentoo.org/791031


thesamesam:
@Polynomial-C

PR #22012, opened 2021-08-17
net-misc/tmate-ssh-server: new package

alexfanqi:
Signed-off-by: Alex Fan <alexfanqi@yahoo.com>

It works for me on amd64 and riscv.

It includes systemd and openrc service file. But it will require user to set specificly certain config variables to work.

In pkg_preinst phase, it will generate RSA and ED25519 keys in /etc/tmate-ssh-server

There is one optional dependency on utempter. The build system will use it if it is found. 
I am not sure how to make it a USE flag to optionally hide utempter from the build system though.

### DIFF
--- a/games-roguelike/angband/angband-4.2.2-r1.ebuild
+++ b/games-roguelike/angband/angband-4.2.2-r1.ebuild
@@ -12,7 +12,7 @@ SRC_URI="https://github.com/angband/angband/archive/refs/tags/${PV}.tar.gz -> ${
 
 LICENSE="GPL-2"
 SLOT="0"
-KEYWORDS="~amd64 ~x86"
+KEYWORDS="~amd64 ~riscv ~x86"
 IUSE="+ncurses sdl sound +X"
 
 REQUIRED_USE="sound? ( sdl )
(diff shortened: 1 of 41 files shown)


dlan17:
1) check QA failure, eg missing metadata.xml, 
2) also I'd suggest not pushing the live ebuild
  * live ebuild isn't constant and will become broken easily
  * this is a rather low volume commit package which not worth a live version

dlan17:
> Signed-off-by: Alex Fan [alexfanqi@yahoo.com](mailto:alexfanqi@yahoo.com)
> 
> It works for me on amd64 and riscv.
> 
> It includes systemd and openrc service file. But it will require user to set specificly certain config variables to work.
> 
> In pkg_preinst phase, it will generate RSA and ED25519 keys in /etc/tmate-ssh-server
> 
I'd rather prefer not creating keys in pkg_preinst(), downside is, it will become part of files of this package.
either do it like ssh as creating at script runtime,  or as pkg_config() like mysql (use emerge --config )


> There is one optional dependency on utempter. The build system will use it if it is found.
> I am not sure how to make it a USE flag to optionally hide utempter from the build system though.
I'd suggest
1) explicitly depend on utempter, just make it a hard dependency (which is current upstream status)
2) or patch the build script (configure.ac) to have an option, may something like --enable-utempter to enable/disable it at build time

PR #22014, opened 2021-08-17
sys-cluster/drbd-utils: Add missing tmpfiles_process call, fix drbadm path in init script 02124fe 

jsmolic:


ultrabug:
thank you!

PR #22018, opened 2021-08-18
net-voip/sipvicious: version 0.3.4

jkroonza:


net-voip/sipvicious/sipvicious-0.3.4.ebuild
```diff
@@ -0,0 +1,30 @@
+# Copyright 2020-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7..10} )
```
ionenwks, line 6: 7 is dead code (not recognized by eclasses anymore), good to cleanup at same time if you're going to change this line.

net-voip/sipvicious/sipvicious-0.3.4.ebuild
```diff
@@ -0,0 +1,30 @@
+# Copyright 2020-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
ionenwks, line 4: maybe time for EAPI-8

net-voip/sipvicious/sipvicious-0.3.4.ebuild
```diff
@@ -0,0 +1,30 @@
...
+PYTHON_COMPAT=( python3_{7..10} )
+DISTUTILS_USE_SETUPTOOLS=rdepend
+DISTUTILS_SINGLE_IMPL=1
+
+inherit distutils-r1
+
+DESCRIPTION="SIPVicious is a SIP security package"
+HOMEPAGE="https://github.com/EnableSecurity/sipvicious/wiki"
+SRC_URI="https://github.com/EnableSecurity/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+KEYWORDS="~amd64 ~x86"
```
ionenwks, line 15: Just style but, to match skel.ebuild ordering, KEYWORDS normally would go straight below SLOT

net-voip/sipvicious/sipvicious-0.3.4.ebuild
```diff
@@ -0,0 +1,30 @@
...
+
+LICENSE="GPL-3+"
+SLOT="0"
+
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
```
ionenwks, line 20: With distutils-r1 you don't need to define this, only needed with python-* eclasses (or if you set DISTUTILS_OPTIONAL, but it's not here)

net-voip/sipvicious/sipvicious-0.3.4.ebuild
```diff
@@ -0,0 +1,30 @@
...
+
+RDEPEND="$(python_gen_cond_dep '
+		dev-python/dnspython[${PYTHON_USEDEP}]
+		dev-python/reportlab[${PYTHON_USEDEP}]
+	')"
+
+src_prepare() {
+	default
+	sed -re 's:man/man1:share/man/man1:' -i setup.py || die "Error updating man page installation location."
```
ionenwks, line 29: Optional but personally think an elaborate `die` message is not needed for a self-explanatory sed (the line will be displayed on error), so just `|| die` is fine.

This also doesn't use extended regex, so -r is superfluous.

net-voip/sipvicious/sipvicious-0.3.4.ebuild
```diff
@@ -0,0 +1,30 @@
+# Copyright 2020-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7..10} )
+DISTUTILS_USE_SETUPTOOLS=rdepend
```
ionenwks, line 7: Can remove this
```
 *   expected: (unset)
```

net-voip/sipvicious/sipvicious-0.3.4.ebuild
```diff
@@ -0,0 +1,30 @@
...
+DISTUTILS_SINGLE_IMPL=1
+
+inherit distutils-r1
+
+DESCRIPTION="SIPVicious is a SIP security package"
+HOMEPAGE="https://github.com/EnableSecurity/sipvicious/wiki"
+SRC_URI="https://github.com/EnableSecurity/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
```
ionenwks, line 14: Could plain-text this ${PN}, see:

net-voip/sipvicious/sipvicious-0.3.4.ebuild
ionenwks: Any update on addressing some of the other points I mentioned like this one? (there are others I haven't heard back from)

There's even a bug opened for this now
https://bugs.gentoo.org/809473

net-voip/sipvicious/sipvicious-0.3.4.ebuild
jkroonza: I still prefer keeping the text, what you consider self-explanatory may not be for others.

net-voip/sipvicious/sipvicious-0.3.4.ebuild
jkroonza: And plaintexting this makes additional requirement for change on version to version ... so how do you balance these conflicting ideals?  emerge -pf is useful and easy for the case described given in the referenced link, even if not plaintexted here.

ionenwks: This was only about ${PN}, when does ${PN} change?

ionenwks (approved):

PR #22022, opened 2021-08-19
sys-apps/plocate: add version 1.1.10

ArsenArsen:
This version is probably stable already, there were only 136 (non-context) lines of diff, and in my lax audit (which so far includes no test suites), I see no glaring issues or significant changes compared to last version.

Signed-off-by: Arsen Arsenović <arsen@aarsen.me>


thesamesam:
Thank you!

sys-apps/plocate/files/plocate-1.1.9-no-systemd-check.patch
```diff
@@ -0,0 +1,38 @@
...
+
+diff --git a/meson.build b/meson.build
+index dc14eb1..9a3b3fe 100644
+--- a/meson.build
++++ b/meson.build
+@@ -76,18 +76,9 @@ updatedb_conf_man = configure_file(input: 'updatedb.conf.5.in',
+                configuration: conf_data)
+ install_man(updatedb_conf_man)
+ 
+-if get_option('install_systemd')
```
Flowdalic, line 18: Why is this patch necessary, given that `install_systemd` defaults to true and the ebuild not modifying the meson option? Is it because `systemunitdir` may be not autodetected if sytemd is not installed? If so, why not set this variable using systemd.eclass' `systemd_get_systemunitdir`, which also "works" if systemd is not installed? I think we could then also drop the `systemd_dounit` in the ebuild's `src_install` (but I have to check first if there is a gentoo policy that `systemd_dounit` has to be used, or not).

sys-apps/plocate/plocate-1.1.9.ebuild
```diff
@@ -0,0 +1,82 @@
...
+ERROR_IO_URING="required for USE=io-uring"
+
+PATCHES=(
+	"${FILESDIR}"/${PN}-1.1.9-no-systemd-check.patch
+)
+
+src_prepare() {
+	# Pretend liburing dep doesn't exist when USE flag off
+	if ! use io-uring; then
+		sed -i "s/dependency('liburing/dependency('/" meson.build || die
```
Flowdalic, line 37: At some point, we should try to upstream a patch that allows overriding this "automagic dependency" so this sed can be dropped. Or was this already tried and rejected by upstream?

ArsenArsen: I didn't attempt to upstream it. I can do that today, I presume it's preferable to add an option here to just ignore the dep?

Flowdalic: Yes, if a software enables or disables a feature based on the availability of a dependency, then this called an "automagic feature". While those may be nice under certain circumstances, they are frowned upon in source based distributions (see also https://devmanual.gentoo.org/general-concepts/use-flags/). The reason is that the distribution, Gentoo here, wants to control if a feature is enabled or not.

So you usually want a tri-state setting, where one option is the current behavior ('automatic', or 'if-available'), the other two being 'enabled' and 'disabled'.

sys-apps/plocate/files/plocate-1.1.9-no-systemd-check.patch
ArsenArsen: The systemd dep was unconditionally checked before, it can be removed now.

sys-apps/plocate/plocate-1.1.9.ebuild
Flowdalic: See also https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies

ArsenArsen: I've emailed in a patch

ArsenArsen:
> Possibly related to GCC 11?

I cannot reproduce this even with GCC11. I installed it, set it as default with gcc-config, and tried to ebuild again in a new shell, but it merged fine. Are you also on glibc?

juippis:
The build fails with `-io-uring` and passes with `+io-uring`. Now unfortunately I forgot to test whether it works with `-io-uring` if you happen to have liburing installed.

ArsenArsen:
good catch! I'll write up a patch and send it to upstream when I figure the issue out

ArsenArsen:
Updated! I got some of the patches upstreamed, and it does compile with and without `io-uring`, at least if `ebuild ... clean package` is to be believed.

PR #22026, opened 2021-08-19
Update delve 1.7.1 Drop 1.6.1

oz123:


dev-go/delve/delve-1.7.1.ebuild
```diff
@@ -0,0 +1,35 @@
...
+
+EGO_PN="github.com/go-delve/delve"
+SRC_URI="https://${EGO_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+KEYWORDS="~amd64"
+
+LICENSE="MIT BSD BSD-2 Apache-2.0"
+SLOT="0"
+IUSE="test"
+FEATURES=network-sandbox
```
juippis, line 19: NNope, this won't do. Can't you somehow add `test? ( ... )` to `SRC_URI`?
You could also try adding `PROPERTIES="test_network"` here and maybe `go test` in `src_test` then works on its own?

dev-go/delve/delve-1.7.1.ebuild
```diff
@@ -0,0 +1,35 @@
...
+
+EAPI=7
+
+inherit go-module
+
+DESCRIPTION="A source-level debugger for the Go programming language"
+HOMEPAGE="https://github.com/go-delve/delve"
+
+EGO_PN="github.com/go-delve/delve"
+SRC_URI="https://${EGO_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
```
juippis:
```suggestion
SRC_URI="https://github.com/go-delve/delve/archive/v${PV}.tar.gz -> ${P}.tar.gz"
```

dev-go/delve/delve-1.7.1.ebuild
oz123: Unfortunately, `go test` needs network access to download more packages. I am still not familiar with go ebuilds. 
I will have to check how to specify these packages as test dependency and allow them to install.

juippis: Just try it with the `PROPERTIES="test_network" ` I mentioned above, it was created for situations like this.

But yes, overall packaging go is a bit messy.

oz123: I got a good help in the IRC channel and got everything working without networking. Just need to commit and push.

oz123:
@juippis I believe that the ebuild is now in a much better state than it was before.
I have learned how to use `dev-go/get-ego-vendor` and the tests are now working without network.
The only test that needs network is disabled. However, I know how to fix it, and will do in the next
release of delve. 
I appreciate your constant watch and feedback!

oz123:
Accidentally closed the PR by deleting the branch .

oz123:
Closing as newer delve version should be merged #22468 .

PR #22040, opened 2021-08-19
sys-cluster/hpx: Bump version to v1.7.1

shifty91:
Update to current HPX release. Thanks!


juippis (approved):

juippis (approved):
FYI you don't have to rename the patch, you can also use the previous one from PATCHEs. We'd prefer to have the original version number when it was introduced, and it should be upstreamed if possible.

PR #22044, opened 2021-08-19
net-irc/quasselgrep: Port to EAPI 8, fix DISTUTILS_USE_SETUPTOOLS

jsmolic:
Closes: https://bugs.gentoo.org/809128

### DIFF
--- a/net-irc/quasselgrep/quasselgrep-0_p20200830.ebuild
+++ b/net-irc/quasselgrep/quasselgrep-0_p20200830.ebuild
@@ -1,11 +1,10 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
 
-PYTHON_COMPAT=( python3_{7..9} )
-PYTHON_REQ_USE=( sqlite )
-DISTUTILS_USE_SETUPTOOLS=rdepend
+PYTHON_COMPAT=( python3_{8..9} )
+PYTHON_REQ_USE="sqlite"
 
 inherit distutils-r1 optfeature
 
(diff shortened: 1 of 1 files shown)


net-irc/quasselgrep/quasselgrep-0_p20200830.ebuild
```diff
@@ -1,28 +1,28 @@
...
 PYTHON_REQ_USE=( sqlite )
-DISTUTILS_USE_SETUPTOOLS=rdepend
 
 inherit distutils-r1 optfeature
 
+MY_COMMIT="9b6b0bc1252daa6e574363d87d04eebd981215a5"
 DESCRIPTION="Tool for searching quassel logs from the commandline"
 HOMEPAGE="https://github.com/fish-face/quasselgrep"
-
-MY_COMMIT="9b6b0bc1252daa6e574363d87d04eebd981215a5"
```
a17r, line 15: Imo, either move `MY_COMMIT` before `inherit` where it is usually placed, or keep it here since it is related to `SRC_URI` more than the block above.

jsmolic: Well I usually see it placed between `inherit` and `DESCRIPTION`, but its fine for me either way

net-irc/quasselgrep/quasselgrep-0_p20200830.ebuild
```diff
@@ -1,11 +1,10 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
 
 PYTHON_COMPAT=( python3_{7..9} )
```
SoapGentoo, line 6:
```suggestion
PYTHON_COMPAT=( python3_{8..9} )
```

net-irc/quasselgrep/quasselgrep-0_p20200830.ebuild
```diff
@@ -1,11 +1,10 @@
...
 PYTHON_REQ_USE=( sqlite )
```
SoapGentoo, line 7: shouldn't this be a string and not an array? :thinking:

SoapGentoo (changes requested):

net-irc/quasselgrep/quasselgrep-0_p20200830.ebuild
jsmolic: Done

net-irc/quasselgrep/quasselgrep-0_p20200830.ebuild
jsmolic: You're right, it should be string

SoapGentoo (approved):

PR #22045, opened 2021-08-20
app-editors/vscod* bump to 1.59.1

AdelKS:
Hello,

Small version bump, the two new versions work on my machine :tm: 

Adel


app-editors/vscode/vscode-1.59.1.ebuild
```diff
@@ -88,6 +88,7 @@ src_install() {
 	doins -r *
 	fperms +x /opt/${PN}/{,bin/}code
 	fperms 4711 /opt/${PN}/chrome-sandbox
+	fperms 755 /opt/${PN}/resources/app/extensions/git/dist/askpass*.sh
```
juippis, line 91: The glob `*` doesn't work with `fperms` :\ 

it passes if you have the files already installed, but on new installs it fails with:
```
>>> Install app-editors/vscode-1.59.1 into /var/tmp/portage/app-editors/vscode-1.59.1/image
chmod: cannot access '/var/tmp/portage/app-editors/vscode-1.59.1/image/opt/vscode/resources/app/extensions/git/dist/askpass*.sh': No such file or directory
 * ERROR: app-editors/vscode-1.59.1::gentoo failed (install phase):
 *   fperms failed
```
been meaning to try and fix fperms for that but haven't looked it yet.

Same thing with vscodium.

AdelKS: Alright thanks! let me fix that.

juippis:
> 
> > Although you may want to check your VDB's PROVIDES file if it understands what is bundled within vscode and vscodium.
> 
> I am not sure I understand this, where is this file ? (a quick googling didn't give me the answer x) )
> 

Check your local */var/db/pkg/* browse to *app-editors/vscode* and check for files *REQUIRES*, *NEEDED*, ...

> 
> Yes it can work on pure wayland [with extra command line arguments](https://wiki.archlinux.org/title/Visual_Studio_Code#Running_natively_under_Wayland). My point was to use USE flags to not pull unneeded dependencies when vscode is run on a pure wayland desktop. I suppose some of those `x11-libs` packages are not needed when running on pure wayland.

I guess you can just make a custom .desktop file, add -wayland to its name and include those parameters in it? This can be installed unconditionally imho.

AdelKS:
> Check your local /var/db/pkg/ browse to app-editors/vscode and check for files REQUIRES, NEEDED, ...

Alright thanks!

> I guess you can just make a custom .desktop file, add -wayland to its name and include those parameters in it? This can be installed unconditionally imho.

I agree, I will do that for the next PR, a new version got released, I am just waiting for vscodium to follow.

> If you test it in a pure wayland system then sure. I just have a feeling in the end something will end up pulling those anyway...

Probably, but theoretically not `vscode` and we'll contribute to making pure-wayland a better place :P I won't do it in the next PR, but if one day I have nothing else to do, I may do it. Or when I switch myself to pure wayland xD

PR #22046, opened 2021-08-20
games-fps/prboom-plus: Bump to version 2.6.1

vilhelmgray:
Closes: https://bugs.gentoo.org/808797


ionenwks:
Getting link failure with `USE="-opengl"`
```
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: src/CMakeFiles/prboom-plus.dir/d_main.c.o: in function `D_Display':
/tmp/portage/games-fps/prboom-plus-2.6.1/work/prboom-plus-2.6.1um/prboom2/src/d_main.c:399: undefined reference to `gld_ProcessExtraAlpha'
```
Tell me if you need more info

vilhelmgray:
> Getting link failure with `USE="-opengl"`
> 
> ```
> /usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: src/CMakeFiles/prboom-plus.dir/d_main.c.o: in function `D_Display':
> /tmp/portage/games-fps/prboom-plus-2.6.1/work/prboom-plus-2.6.1um/prboom2/src/d_main.c:399: undefined reference to `gld_ProcessExtraAlpha'
> collect2: error: ld returned 1 exit status
> ```
> 
> Tell me if you need more info

Would you be able to verify that `BUILD_GL` is disabled and CMake is not checking for the existence of OpenGL? The line of code referenced in the error message is guarded by the `GL_DOOM` preprocessor define which should not exist if OpenGL is not found by CMake: https://github.com/coelckers/prboom-plus/blob/master/prboom2/src/CMakeLists.txt#L391-393

ionenwks:
> Would you be able to verify that `BUILD_GL` is disabled and CMake is not checking for the existence of OpenGL? The line of code referenced in the error message is guarded by the `GL_DOOM` preprocessor define which should only exist if OpenGL is found by CMake: https://github.com/coelckers/prboom-plus/blob/master/prboom2/src/CMakeLists.txt#L391

I see the `-DBUILD_GL=no` and no opengl checks. But from a quick look it doesn't seem guarded by GL_DOOM, there's a `#ifdef GL_DOOM` higher up but there's an `#endif` right after.
https://github.com/coelckers/prboom-plus/blob/master/prboom2/src/d_main.c#L284

prboom-plus-2.6 doesn't seem affected, looks this this introduced the regression:
https://github.com/coelckers/prboom-plus/commit/87fef9d2aa3d4eea1625f86ce9d33cac5cdce832

vilhelmgray:
> > Would you be able to verify that `BUILD_GL` is disabled and CMake is not checking for the existence of OpenGL? The line of code referenced in the error message is guarded by the `GL_DOOM` preprocessor define which should only exist if OpenGL is found by CMake: https://github.com/coelckers/prboom-plus/blob/master/prboom2/src/CMakeLists.txt#L391
> 
> I see the `-DBUILD_GL=no` and no opengl checks. But from a quick look it doesn't seem guarded by GL_DOOM, there's a `#ifdef GL_DOOM` higher up but there's an `#endif` right after.
> https://github.com/coelckers/prboom-plus/blob/master/prboom2/src/d_main.c#L284

Ah you're right, I was looking at the wrong line.

> prboom-plus-2.6 doesn't seem affected, looks this this introduced the regression:
> [coelckers/prboom-plus@87fef9d](https://github.com/coelckers/prboom-plus/commit/87fef9d2aa3d4eea1625f86ce9d33cac5cdce832)

Okay, looks like this is an upstream regression. I can write a patch to guard this new line and submit it to upstream as well.

PR #22047, opened 2021-08-20
app-backup/restic: verbump to 0.12.1

stkw0:


juippis (commented):
Please split adding a new version and removing an old one to their separate commits.

Also, 	"github.com/coreos/go-systemd/v22 v22.3.2/go.mod" heh

stkw0:
What is the problem with go-systemd? It's an indirect dependency, restic does not need systemd to run

PR #22051, opened 2021-08-20
[QA] riscv bluetooth

mkszuba:


ConiKost:
Is this still needed and can it be closed?

dlan17:
ya, bluetooth works fine, tested on unmatched board using Intel 3165NGW module

thanks for your concern

PR #22052, opened 2021-08-20
mongodb: fix building with glibc-2.34

hydrapolic:
Compile tested on ~amd64.


ultrabug:
What can I say... thanks mate! I don't even have newer glibc on my box, it's so dying I don't update any more :(

hydrapolic:
Thanks @ultrabug

PR #22059, opened 2021-08-21
media-libs/netpbm: Fix test failure with USE -png

ceamac:
Skip `png-roundtrip*` tests when not compiling with *png* USE flag

### DIFF
--- a/media-libs/netpbm/netpbm-10.86.22.ebuild
+++ b/media-libs/netpbm/netpbm-10.86.22.ebuild
@@ -112,6 +112,7 @@ src_prepare() {
 			-e 's:(pamrgbatopng|pngtopnm).*::' \
 			test/legacy-names.{ok,test} || die
 		sed -i -e '/^$/d' test/legacy-names.ok || die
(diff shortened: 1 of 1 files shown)


media-libs/netpbm/netpbm-10.86.22.ebuild
```diff
@@ -112,6 +112,7 @@ src_prepare() {
 			-e 's:(pamrgbatopng|pngtopnm).*::' \
 			test/legacy-names.{ok,test} || die
 		sed -i -e '/^$/d' test/legacy-names.ok || die
+		sed -i -e 's:png-roundtrip.*::' test/Test-Order || die
```
juippis, line 115: `if ! use png; then`? 
Just let me know if it's supposed to work **when** png is enabled.

ceamac: Without `png` USE flag, tools like `pamtopng` and `pngtopam` are not built, so we need to disable the tests using them.  So `if ! use png` I remove them with `sed`.  If png is enabled, the tests run fine.

juippis (approved):

PR #22060, opened 2021-08-21
sys-devel/llvm-roc: add hip-location.patch back

littlewu2508:
As https://bugs.gentoo.org/809392 descriped, clang from llvm-roc-4.3.0 throws error
during compilation of rocm packages for GPU devices (e.g. rocBLAS). The
missing of $ROCM_PATH and deprecation of hip-location.patch together
causes in this situation.

This commit update the hip-location.patch so it can be used again. Also,
it sets ROCM_PATH during src_configure to let cmake aware of, as double
insurance.

Closes: https://bugs.gentoo.org/809392


heroxbd:
> As https://bugs.gentoo.org/809392 descriped, clang from llvm-roc-4.3.0 throws error
> during compilation of rocm packages for GPU devices (e.g. rocBLAS). The
> missing of $ROCM_PATH and deprecation of hip-location.patch together
> causes in this situation.
> 
> This commit update the hip-location.patch so it can be used again. Also,
> it sets ROCM_PATH during src_configure to let cmake aware of, as double
> insurance.
> 
> Closes: https://bugs.gentoo.org/809392
> Package-Manager: Portage-3.0.20, Repoman-3.0.3
> Signed-off-by: Yiyang Wu [xgreenlandforwyy@gmail.com](mailto:xgreenlandforwyy@gmail.com)

I discourage "double insurance".  If it works we don't need it.  Otherwise we should fix what's broken.

littlewu2508:
> I discourage "double insurance". If it works we don't need it. Otherwise we should fix what's broken.

The "double insurance" is adding one line: "export ROCM_PATH=${EPREFIX}/usr". After checking, this will only affect mlir component, currently not installed with llvm-roc, so it is safe not to use it. However mlir is closely related to tensorflow, I'mnot sure whether this line will be needed in the future.

heroxbd:
> > I discourage "double insurance". If it works we don't need it. Otherwise we should fix what's broken.
> 
> The "double insurance" is adding one line: "export ROCM_PATH=${EPREFIX}/usr". After checking, this will only affect mlir component, currently not installed with llvm-roc, so it is safe not to use it. However mlir is closely related to tensorflow, I'mnot sure whether this line will be needed in the future.

We shall add it back if it is needed in the future.

heroxbd:
LGTM

PR #22064, opened 2021-08-21
app-misc/radeontop: fit for gentoo prefix

littlewu2508:
performing hprefixify on Makefile, so the install location is in
${EPREFIX}


mm1ke:
LGTM

app-misc/radeontop/radeontop-9999.ebuild
```diff
@@ -31,6 +31,11 @@ DEPEND="${RDEPEND}
 "
 BDEPEND="virtual/pkgconfig"
 
+src_prepare() {
+	hprefixify Makefile
+	default
```
ionenwks, line 36: It's good to make file changes after `default` to not mess with user patches, but that aside...

Also good avoid doing file changes if it can be avoided, and in this case the Makefile has `PREFIX ?= /usr`, ?= means it'll use PREFIX from the environment allowing for a simple:
`export PREFIX="${EPREFIX}"/usr` in src_configure.

Maybe use that instead? Unless there's something I'm missing. For -9999 it'll also remove the need to define src_prepare.

PR #22067, opened 2021-08-21
QA Fixes: Variable Scope

00-matt:


app-text/bibutils/bibutils-6.10.ebuild
```diff
@@ -30,6 +33,12 @@ src_compile() {
 		LDFLAGSIN="${LDFLAGS}"
 }
 
+src_install() {
+	export EINSTALLDIR="${ED}/usr/bin"
```
thesamesam, line 37: Can’t we make it repeat DESTDIR instead (and libdir if necessary, but that should be passed in configure _ideally_)

app-text/catdoc/catdoc-0.95-r1.ebuild
```diff
@@ -31,8 +31,7 @@ src_prepare() {
 }
 
 src_configure() {
-	econf --with-install-root="${D}" \
-		$(use_with tk wish "${EPREFIX}"/usr/bin/wish) \
+	econf $(use_with tk wish "${EPREFIX}"/usr/bin/wish) \
```
thesamesam, line 34:
```suggestion
	econf \
	(tab)$(use_with tk wish "${EPREFIX}"/usr/bin/wish) \
```

net-misc/netkit-talk/netkit-talk-0.17-r7.ebuild
```diff
@@ -0,0 +1,57 @@
...
+SRC_URI="ftp://ftp.uk.linux.org/pub/linux/Networking/netkit/${MY_P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~hppa ~ia64 ~mips ~ppc ~ppc64 ~sparc ~x86"
+IUSE="ipv6"
+
+COMMON_DEPEND=">=sys-libs/ncurses-5.2:="
+DEPEND="
+	${COMMON_DEPEND}
```
thesamesam, line 22: DEPEND? 🤨

net-misc/netkit-talk/netkit-talk-0.17-r7.ebuild
```diff
@@ -0,0 +1,57 @@
...
+DESCRIPTION="Netkit - talkd"
+HOMEPAGE="ftp://ftp.uk.linux.org/pub/linux/Networking/netkit/"
+SRC_URI="ftp://ftp.uk.linux.org/pub/linux/Networking/netkit/${MY_P}.tar.gz"
+
+LICENSE="BSD"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~hppa ~ia64 ~mips ~ppc ~ppc64 ~sparc ~x86"
+IUSE="ipv6"
+
+COMMON_DEPEND=">=sys-libs/ncurses-5.2:="
```
thesamesam, line 20: We can usually dump the idea of COMMON_DEPEND

sys-fs/copyfs/copyfs-1.0.1-r1.ebuild
```diff
@@ -30,13 +30,14 @@ PATCHES=(
 
 src_prepare() {
 	default
+	sed -i '/@/d' Makefile.in || die
```
thesamesam, line 33: Explain!

app-text/bibutils/bibutils-6.10.ebuild
```diff
@@ -30,6 +33,12 @@ src_compile() {
...
+	export ELIBINSTALLDIR="${ED}/usr/$(get_libdir)"
+	emake install
+}
+
 src_test() {
```
thesamesam, line 42: define in order of execution

sys-fs/copyfs/copyfs-1.0.1-r1.ebuild
00-matt: Whoops, that wasn't supposed to make it in!  It stopped the configure script from changing lines like `bindir=@bindir@`, but that doesn't matter as I just set `bindir` anyway when I called make.

app-text/bibutils/bibutils-6.10.ebuild
00-matt: I am not totally sure what you meant, but I made the patches a bit less gross and just prepended DESTDIR to the install destinations.

net-misc/netkit-talk/netkit-talk-0.17-r7.ebuild
00-matt:
```
/me does a quiz and writes an answer about all of the different DEPENDs, when they're present and which host they're for
/me leaves ncurses in DEPEND
```

net-misc/netkit-talk/netkit-talk-0.17-r7.ebuild
```diff
@@ -0,0 +1,54 @@
...
+RDEPEND="
+	${DEPEND}
+	virtual/inetd
+"
+
+PATCHES=( "${FILESDIR}"/${P}-time.patch )
+
+src_prepare() {
+	default
+	use ipv6 && eapply "${FILESDIR}"/${P}-ipv6.diff
```
thesamesam, line 31: 1. .diff -> .patch
2. Conditional patching is something we try to avoid, but I don't expect you to fix this in some abandoned codebase.
3. USE=ipv6 has dubious value in most cases (just something to know).

00-matt: Do you think it's worth dropping the USE flag and unconditionally applying the ipv6 fix?  To be honest I didn't look at the contents of the patch, just made sure that it built.

thesamesam: It really depends on if how the patch works. But I wouldn't mess with it in this case given the fragility of the netkit stuff.

(That's fine by the way, part of why these fixes are great is because it gives me an opportunity to soapbox about random stuff to ensure you know it!)

app-text/bibutils/bibutils-6.10.ebuild
thesamesam: Oh, much better. Nice!

net-misc/netkit-talk/netkit-talk-0.17-r7.ebuild
thesamesam: OK, checked the patch. Too janky for me. In theory, it'd be fine to unconditionally apply it, but please don't -- it's a bit scary with the bodged indentation and the rest of it.

thesamesam:
Thanks! Unrevbumped netkit-talk though. Revbumps for EAPI changes are usually an unnecessary use of arch testers' time, although OK in some situation, I usually don't bother.

sys-apps/pcmciautils/pcmciautils-018_p8-r3.ebuild
```diff
@@ -44,7 +44,6 @@ pkg_setup() {
 		STRIP=true
 		RANLIB="$(tc-getRANLIB)"
 		OPTIMIZATION="${CFLAGS} ${CPPFLAGS}"
```
thesamesam, line 46: actually: I didn’t notice this earlier but configure arguments really shouldn’t be in pkg_setup anyway!

PR #22069, opened 2021-08-21
app-text/cmark: update to 0.30.1

winterheart:
Update to 0.30.1, remove 0.30.0.


app-text/cmark/cmark-0.30.1.ebuild
```diff
@@ -0,0 +1,34 @@
...
+PYTHON_COMPAT=( python3_{7..10} )
+
+inherit cmake python-any-r1
+
+DESCRIPTION="CommonMark parsing and rendering library and program in C"
+HOMEPAGE="https://github.com/commonmark/cmark"
+SRC_URI="https://github.com/commonmark/cmark/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="BSD-2"
+SLOT="0/0.30.0"
```
ionenwks: shouldn't this be bumped?
```
 * QA: comparing =app-text/cmark-0.30.0 with app-text/cmark-0.30.1/image
 * SONAME:-libcmark.so.0.30.0(64)
 * SONAME:+libcmark.so.0.30.1(64)
```
Maybe a good occasion to just make this `SLOT="0/${PV}"` if it's like that every time rather than risk forgetting.

PR #19001, opened 2021-01-09
mail-filter/rspamd: bump version, sync live, clan old

arkamar:
Here is the bump to the newest version 2.7. I synced the live ebuild as well, and removed old version 2.5 because 2.6 is currently stable.

PR #22070, opened 2021-08-21
mail-filter/rspamd: add 3.0, sync live and drop 2.6

arkamar:
This is a version bump to 3.0 and live ebuild sync. There are few changes from previous versions:
- `zstd` unbundle patch is not necessary anymore, new `-DSYSTEM_ZSTD=ON` option is used instead.
- New `libfmt` dependency is used from system with `-DSYSTEM_FMT=ON` option, which is already present in upstream and I beckported it to 3.0 via patch.
- The `src_test` function is updated in order to make it work properly, see https://github.com/gentoo/gentoo/pull/19001#issuecomment-757142647. Tests work only with `luajit`, therefore, I added `test` to `IUSE`, made `luajit` required if the `test` is enabled and set `RESTRICT="!test? ( test )"`.

I removed 2.6 version, because 2.7 is already for half year stable without any know issue.


mail-filter/rspamd/rspamd-3.0.ebuild
```diff
@@ -76,6 +80,10 @@ src_configure() {
 		-DRUNDIR=/var/run/rspamd
 		-DDBDIR=/var/lib/rspamd
 		-DLOGDIR=/var/log/rspamd
+
+		-DSYSTEM_FMT=ON
```
thesamesam, line 84: Thank you for working to upstream this stuff btw. Really appreciated.

PR #22073, opened 2021-08-22
www-apps/gitea: Version bump 1.15.2.

tastytea:
I removed the `build-client` useflag, because it tries to download stuff and access `/etc/npm`. I don't know how to fix that.


holgersson32644:
Hi,

this sounds like https://github.com/go-gitea/gitea/issues/16769.

Instead of removing the option I would suggest to either bundle the files from the git repo and let a dev mirror it or simply wait.
I bumped gitea locally for myself and actually didn't check the nodejs-Part. However, I've made two other changes that you might want to incorporate:

1. Hide file-based capabilities behind a USE flag. Disabling filecaps improves  the security a little bit for everyone running gitea behind reverse proxy  like nginx while keeping the option to run it standalone easily.
2. Fix the copyright year inside the ebuild header. Gitea was introcuded in 2016  as version 0.9.97_p20161219 in 4f33c6397807b7840db72e5f16f9860a79f87ff.

Kind regards,
Nils

tastytea:
I'm not convinced the `build-client` useflag is worth the effort. `fcaps.eclass` already adds the `filecaps` useflag automatically (by the way, it is also needed for the inbuilt SSH server). I'll fix the ebuild header.

holgersson32644:
> I'm not convinced the `build-client` useflag is worth the effort. `fcaps.eclass` already adds the `filecaps` useflag automatically (by the way, it is also needed for the inbuilt SSH server). I'll fix the ebuild header.

You're right, my mistake!

VTimofeenko:
JFYI, 'USE="-fcaps"' may break gitea because of suid on gitea binary, see [bug #750218](https://bugs.gentoo.org/750218)

tastytea:
> JFYI, 'USE="-fcaps"' may break gitea because of suid on gitea binary, see [bug #750218](https://bugs.gentoo.org/750218)

Thanks, fixed it in this ebuild. This seems to be a [bug in fcaps.eclass](https://bugs.gentoo.org/811105).

ionenwks:
> > JFYI, 'USE="-fcaps"' may break gitea because of suid on gitea binary, see [bug #750218](https://bugs.gentoo.org/750218)
> 
> Thanks, fixed it in this ebuild. This seems to be a [bug in fcaps.eclass](https://bugs.gentoo.org/811105).

See `games-emulation/pcsx2` ebuild for setting a different fallback, e.g. do nothing. setuid is there to "make this work anyway".

tastytea:
> FWIW it is intended behavior rather than a bug

I don't think it is intended by the eclass, since it doesn't happen with default `pkg_postinst`. it only happens when I override `pkg_postinst` and call `fcaps_pkg_postinst` from there.

ionenwks:
> > FWIW it is intended behavior rather than a bug
> 
> I don't think it is intended by the eclass, since it doesn't happen with default `pkg_postinst`. it only happens when I override `pkg_postinst` and call `fcaps_pkg_postinst` from there.

You have two inherits exporting pkg_postinst, fcaps and go-module, and the latter one gets used. So it wasn't getting called when it should.

ionenwks:
Although as seen on gentoo-dev ML, go-module will lose its pkg_postinst soon.

VTimofeenko:
Looks like [1.15.1](https://github.com/go-gitea/gitea/releases/tag/v1.15.1) is available as of several hours ago

hydrapolic (approved):
Upgraded from 1.14.6 to 1.15.2, works like a charm.

tastytea:
I'm abandoning this PR because it doesn't seem that it will get merged; and I don't want to “officially” maintain it in ::gentoo because I know too little about go and JavaScript to fix the open and future bugs.

However, I will continue to maintain it in [::tastytea](https://schlomp.space/tastytea/overlay), without any guarantees.

juippis (approved):
Sorry, no idea how this was missed. 

Seems to work, thanks!

hydrapolic:
Thanks @juippis

PR #22076, opened 2021-08-23
net-misc/frr: Adding net-dns/c-ares in COMMON_DEPEND

alarig:
Closes: https://bugs.gentoo.org/809728
Signed-off-by: Alarig Le Lay <alarig@swordarmor.fr

### DIFF
--- a/net-misc/frr/files/frr-8.0-c-ares-vtysh.patch
+++ b/net-misc/frr/files/frr-8.0-c-ares-vtysh.patch
@@ -1,4 +1,5 @@
-https://bugs.gentoo.org/809728
+diff --git a/lib/subdir.am b/lib/subdir.am
+index 75f3d91b9..05910349e 100644
 --- a/lib/subdir.am
 +++ b/lib/subdir.am
 @@ -144,7 +144,6 @@ vtysh_scan += \
--- a/net-misc/frr/frr-8.0.ebuild
+++ b/net-misc/frr/frr-8.0.ebuild
@@ -4,7 +4,7 @@
 EAPI=7
 
 PYTHON_COMPAT=( python3_{8..10} )
-inherit autotools pam python-single-r1 systemd
+inherit autotools optfeature pam python-single-r1 systemd
 
 DESCRIPTION="The FRRouting Protocol Suite"
 HOMEPAGE="https://frrouting.org/"
@@ -23,7 +23,6 @@ COMMON_DEPEND="
 	${PYTHON_DEPS}
 	acct-user/frr
 	dev-libs/json-c:0=
-	net-dns/c-ares:=
 	>=net-libs/libyang-2.0.0
 	sys-libs/libcap
 	sys-libs/readline:0=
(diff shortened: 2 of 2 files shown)


net-misc/frr/frr-8.0.ebuild
```diff
@@ -23,6 +23,7 @@ COMMON_DEPEND="
 	${PYTHON_DEPS}
 	acct-user/frr
 	dev-libs/json-c:0=
+	net-dns/c-ares:=
```
jsmolic, line 26: I don't think we need to pull in c-ares unconditionally.
See https://github.com/FRRouting/frr/pull/9415

alarig: It’s 5 days old, the 8.0 is a month old

jsmolic: Yeah but I would like to backport that patch to 8.0

alarig: I added the patch and the optfeature in my last commit

net-misc/frr/files/frr-8.0-c-ares-vtysh.patch
```diff
@@ -1,4 +1,5 @@
-https://bugs.gentoo.org/809728
+diff --git a/lib/subdir.am b/lib/subdir.am
+index 75f3d91b9..05910349e 100644
```
jsmolic, line 2: The patch is already added to the package so this change can be removed

net-misc/frr/frr-8.0.ebuild
```diff
@@ -33,16 +33,19 @@ COMMON_DEPEND="
 	rpki? ( >=net-libs/rtrlib-0.6.3[ssh] )
 	snmp? ( net-analyzer/net-snmp:= )
 "
+
```
jsmolic, line 36: Let's not add these empty lines, they unnecessarily increase vertical whitespace

net-misc/frr/frr-8.0.ebuild
```diff
@@ -51,7 +54,7 @@ RDEPEND="
 
 PATCHES=(
 	"${FILESDIR}"/${PN}-7.5-ipctl-forwarding.patch
-	"${FILESDIR}"/${P}-c-ares.patch
+	"${FILESDIR}/${P}-c-ares-vtysh.patch"
```
jsmolic, line 57: The patch name can also be left unchanged

net-misc/frr/frr-8.0.ebuild
```diff
@@ -148,3 +151,7 @@ src_install() {
 	# Conflict files, installed by net-libs/libsmi, bug #758383
 	rm "${ED}"/usr/share/yang/ietf-interfaces.yang || die
 }
+
+pkg_postinst() {
+	optfeature "dns" net-dns/c-ares
```
jsmolic, line 156: nhrpd requires c-ares so this couldn't be an optfeature

net-misc/frr/frr-8.0.ebuild
```diff
@@ -33,16 +33,19 @@ COMMON_DEPEND="
 	rpki? ( >=net-libs/rtrlib-0.6.3[ssh] )
 	snmp? ( net-analyzer/net-snmp:= )
 "
+
 BDEPEND="
-	>=dev-util/clippy-${PV}
+	~dev-util/clippy-"${PV}"
```
jsmolic, line 38: This change is okay, but `${PV}` doesn't need to be in quotations

jsmolic (changes requested):

jsmolic:
@alarig Thanks for the PR, the bug was already closed with https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=295a86b2494085c4125761afddc962d6a66ebc09

alarig:
Thanks for the fix!

PR #22084, opened 2021-08-23
media-plugins/mythplugins: Remove oauth2 dependency

thebitpit:

### DIFF
--- a/media-plugins/mythplugins/mythplugins-31.0-r2.ebuild
+++ b/media-plugins/mythplugins/mythplugins-31.0-r2.ebuild
@@ -3,27 +3,21 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7,8,9} )
+# Grab only the major version number.
+MAJOR_PV=${PV%%.*}
+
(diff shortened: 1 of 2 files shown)


thebitpit:
Remove oauth2 dependency from mythnetvision option as it has been blocked from building by upstream for the mythplugins versions in gentoo. It has been fixed on the upstream master, but not yet released. This revision is done to help with dev-python/oauth2 clean up.

sbraz:
Hi, thanks for the PR. I still don't really understand why `oauth2` would have been needed as I don't see anything that would reference it [here](https://github.com/MythTV/mythtv/search?q=oauth2).

I don't like the idea that we are dropping a dependency that might be needed if that plugin is fixed. If you confirm that even if the plugin is fixed, this dependency is not really required (as I was thinking), I will merge this.

sbraz:
Hi @thebitpit did you see my previous message?

thebitpit:
> Hi @thebitpit did you see my previous message?

Yes! I guess I failed to hit the Comment button. I am sure it no longer needs oauth2 and the version in portage builds without it.

PR #22087, opened 2021-08-24
dev-util/bats: disable failing test and minor output fix

henning-schild:


dev-util/bats/bats-1.3.0.ebuild
```diff
@@ -13,12 +13,21 @@ SRC_URI="https://github.com/${MY_PN}/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.
 LICENSE="MIT"
 SLOT="0"
 KEYWORDS="amd64 ~arm ~arm64 ppc64 ~riscv x86"
+IUSE="test"
```
Flowdalic, line 16: Declaring the 'test' use flag is not necessary if there are no test dependencies. See also https://devmanual.gentoo.org/ebuild-writing/functions/src_test/index.html#supporting-test-suites-in-packages

dev-util/bats/bats-1.4.1.ebuild
```diff
@@ -13,15 +13,24 @@ SRC_URI="https://github.com/${MY_PN}/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.
 LICENSE="MIT"
 SLOT="0"
 KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~riscv ~x86"
+IUSE="test"
```
Flowdalic, line 16: Declaring the 'test' use flag is not necessary if there are no test dependencies. See also https://devmanual.gentoo.org/ebuild-writing/functions/src_test/index.html#supporting-test-suites-in-packages

dev-util/bats/bats-1.3.0.ebuild
```diff
@@ -13,12 +13,21 @@ SRC_URI="https://github.com/${MY_PN}/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.
...
 DEPEND="app-shells/bash:*"
 RDEPEND="${DEPEND}"
 
 S="${WORKDIR}/${MY_PN}-${PV}"
 
+src_prepare() {
+	default
+	if use test ; then
+		# "parallelity factor is met exactly" fails in portage, disable
+		sed -i -e 's/@test "parallelity factor is met exactly" {$/@test "parallelity factor is met exactly" { skip "gentoo"/' test/parallel.bats
```
Flowdalic, line 27: Not a blocker, but it would be nice if the comment would indicate why the test is failing.

dev-util/bats/bats-1.4.1.ebuild
```diff
@@ -13,15 +13,24 @@ SRC_URI="https://github.com/${MY_PN}/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.
...
+	if use test ; then
+		# "parallelity factor is met exactly" fails in portage, disable
+		sed -i -e 's/@test "parallelity factor is met exactly" {$/@test "parallelity factor is met exactly" { skip "gentoo"/' test/parallel.bats
+	fi
+}
+
 src_test() {
 	local my_jobs=$(makeopts_jobs)
-	if ! command -v parallel; then
+	if ! command -v parallel >/dev/null; then
```
Flowdalic, line 33: Why not simply have sys-process/parallel as test DEPEND? (Then IUSE="test" would be required again)

Flowdalic (changes requested):

dev-util/bats/bats-1.4.1.ebuild
henning-schild: duplicate

dev-util/bats/bats-1.3.0.ebuild
henning-schild: unless i patch unconditionally this is needed because of the `if use test`

dev-util/bats/bats-1.4.1.ebuild
henning-schild: this commit is not related to the bug, its just a drive-by fix

parallel is an optdepend, and depending on whether or not it is installed testing needs to be done differently

dev-util/bats/bats-1.3.0.ebuild
henning-schild: as said in the commit message ... the full reason is unfortunately not clear to me, i did debug for a while and also modified the test to potentially write a patch to upstream. But after investing about 2h without a real reason ... i decided to patch it out. Dirty but imho acceptable.

henning-schild:
@Flowdalic thanks for the review. I resolved threads with comments, do not see a reason to change the code. Feel free to unresolve threads in case my answers are not satisfying

dev-util/bats/bats-1.3.0.ebuild
henning-schild, line 16: this is needed because of the `if use test` in `src_prepare` ... not sure whether this one change will need a `-r1`, so far the versions stay untouched.

Flowdalic: Why is the sed even done conditionaly? You could probably simply do it in src_test. Also the sed is missing a `|| die`.

dev-util/bats/bats-1.4.1.ebuild
Flowdalic: I get that testing needs to be done if it's not installed, but I wonder why not simply have `DEPEND=test? ( sys-process/parallel )`?

Flowdalic:
> @Flowdalic thanks for the review. I resolved threads with comments, do not see a reason to change the code. Feel free to unresolve threads in case my answers are not satisfying

Thanks for the response. It appears I am not able to unmark threads as resolved. But ultimately those are not really big issues. But I personally would make sys-process/parallel a test dependency, then you could get rid of a few lines of code in src_test and I would disable the test(s) in src_test and not in src_prepare, then you could get rid of src_prepare.

Please also note the missing `|| die` after the `sed` invocation.

dev-util/bats/bats-1.3.0.ebuild
henning-schild: i first had it in `src_test` but since it is a "patch" i opted for `src_prepare` since that seems much cleaner ... so `src_test` just runs the tests but does not alter the sources ... which should probably be done by `src_prepare` as a general rule.

the `|| die` is a very good point which i will adress, not sure unconditional is a good idea ... will try that and wait for further reviews

henning-schild: did add the `|| die` and unconditional `sed`

PR #22088, opened 2021-08-24
Remove last-rited packages

jsmolic:


acct-group/minbif/metadata.xml
```diff
@@ -1,5 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
```
SoapGentoo, line 2: don't remove any of the `acct-*` packages yet, the policy and discussions around these aren't finalised yet

SoapGentoo (changes requested):

acct-group/minbif/metadata.xml
jsmolic: Okay, I've removed acct packages

SoapGentoo (approved):

PR #22089, opened 2021-08-24
media-video/rav1e: add USE=static-libs

Hello71:
tested libheif[rav1e], libavif[rav1e] still working


mgorny:
What's the purpose of supporting static libs in the first place?

Hello71:
i don't know that much about rust, but i think possibly the ABI is not stable? anyways, i don't want to break someone's spacebar heating if the USE flag is easy to add. i don't feel strongly about the option, just don't want them installed by default.

thesamesam:
Would you mind changing to just [never](https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0302) install static libs, then I'll merge? Thanks!

thesamesam:
Thanks!

PR #22093, opened 2021-08-24
media-libs/tg_owt: Fix building without YASM

mid-kid:
@gyakovlev

The YASM objects aren't used, so it really shouldn't affect build output at all. Avoiding revbump.


gyakovlev:
thanks.

PR #22095, opened 2021-08-24
[DO NOT MERGE] lxde-base/* removal

jsmolic:
LXDE develpoment is discontinued upstream in favor of LXQt, so let's last-rite everything we can while keeping 
`x11-misc/pcmanfm`and its core dependencies in the tree.
Bug: https://bugs.gentoo.org/810091


jsmolic:
Note that only last-rite commit can be pushed first (https://github.com/gentoo/gentoo/pull/22095/commits/2728a4f60143f46ba9c6bcf7bce2d2fe1a715a65) while others can be done along with package removals in 30 days.

Since we're keeping `x11-misc/pcmanfm{,-qt}` in the tree, we also need `lxde-base/menu-{cache,data}` ( we also need those for some `lxqt-base` packages) and `x11-libs/libfm{,-extra,-qt}`. Rest of `lxde-base` packages can be safely removed from the tree.

jsmolic:
- A full list of packages that will be removed from the tree:
```
media-sound/lxmusic
lxde-base/lxappearance
lxde-base/lxappearance-obconf
lxde-base/lxde-common
lxde-base/lxde-icon-theme
lxde-base/lxde-meta
lxde-base/lxinput
lxde-base/lxlauncher
lxde-base/lxpanel
lxde-base/lxrandr
lxde-base/lxsession
lxde-base/lxtask
lxde-base/lxterminal
```
- Packages that will have to stay because they have reverse dependencies:
```
lxde-base/lxmenu-data
lxde-base/menu-cache
```

Additionaly, we have `lxde-base/lxdm` whose maintainer wants to keep it in the tree for a while longer and will be moved to another category. `media-gfx/gpicview` also has a maintainer who wants to keep it in the tree, so those two packages will be untouched for now.

If we decide to go through with this:
- Step 1 will be to last-rite the packages from the list.
- Step 2 will be done in 30 days :

After removal we will only have 2 packages left in the `lxde-base` category, so it might be worth considering removing the whole category. If we decide to do that, we will have to move those 2 packages to a different category, add pkgmove entries in `profiles/updates` and fix all references to the old package name and update them (dependencies in other ebuilds, profile or metadata.xml entries or `<pkg>` tags, change open bug summaries, etc.).
Additionaly we will also have to remove `lxde-base` from the list in `profiles/categories`

There are also some reverse dependencies which will have to be fixed after removal, which includes for instance removing `lxde-base/lxsession` from dependencies in `sys-auth/polkit` and removing  `lxde-base/lxde-icon-theme` from `virtual/freedesktop-icon-theme`.

If we leave `lxde-base/lxdm` in the tree there will be no profile entries to cleanup, otherwise we would have to remove an entry from `features/prefix/package.use.mask`.

All of this will be double checked in  30 days when removing the packages for, so I will again go through all reverse dependencies, look for leftover profile entries, metadata.xml references to lxde packages, etc.


Wiki pages are also going to have to be updated, explaining that LXDE packages are no longer in the Gentoo tree.
(https://wiki.gentoo.org/wiki/LXDE)

juippis (commented):
As said, I have been thinking about lxde's removal ever since I disbanded the project +1 year ago. It seems some faults have been fixed meanwhile, and I just tested, and lxde still works. It's unfortunate it has no maintainer in Gentoo currently, but I don't see a reason to force its removal in its current state. 

I'll write this .patch as a reference for future if/when it breaks down, again, and receives no maintainer in Gentoo.

juippis:
Thank you for the effort!

PR #22096, opened 2021-08-24
sci-libs/med: fix build against hdf5-1.12

waebbl:
Thanks to Alexandre Ferreira for providing the patch.

Closes: https://bugs.gentoo.org/809008

profiles/base/package.use.mask: mask mpi for sci-libs/med

Package depends on sci-libs/hdf5[mpi] which currently has test issues
and has USE=mpi masked.

### DIFF
--- a/profiles/base/package.use.mask
+++ b/profiles/base/package.use.mask
@@ -6,6 +6,11 @@
 # This file is only for generic masks. For arch-specific masks (i.e.
 # mask everywhere, unmask on arch/*) use arch/base.
 
+# Bernd Waibel <waebbl-gentoo@posteo.net> (2021-08-24)
+# Needs sci-libs/hdf5[mpi], which is currently masked.
+# Mask until hdf5 builds with MPI support.
+sci-libs/med mpi
+
 # Marek Szuba <marecki@gentoo.org> (2021-07-14)
 # Even on a clean install, 8 phdf5 tests fail (Bug #808612)
 # Mask for further study.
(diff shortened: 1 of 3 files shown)


AndrewAmmerlaan:
Thanks :+1:

waebbl:
@AndrewAmmerlaan Andrew, can you revert this? I just got an email that the mask breaks gmsh. Haven't tested the revdeps on this.

AndrewAmmerlaan:
> @AndrewAmmerlaan Andrew, can you revert this? I just got an email that the mask breaks gmsh. Haven't tested the revdeps on this.

Reverted, I'm slightly confused though, because the qa-bot said it was fine

waebbl:
> > @AndrewAmmerlaan Andrew, can you revert this? I just got an email that the mask breaks gmsh. Haven't tested the revdeps on this.
> 
> Reverted, I'm slightly confused though, because the qa-bot said it was fine

Does it check for revdeps? Just looked up latest gmsh and noticed it depends on med[mpi] unconditionally and has a `REQUIRED_USE=med? ( mpi )` constraint. Looks, like we need to mask the med USE flag on gmsh as well.

Need to figure out how to re-track the branch again. Already deleted it locally :/

AndrewAmmerlaan:
> Does it check for revdeps? Just looked up latest gmsh and noticed it depends on med[mpi] unconditionally and has a REQUIRED_USE=med? ( mpi ) constraint. Looks, like we need to mask the med USE flag on gmsh as well.

It should, it checks the whole tree. Possibly your local tree was significantly behind the main tree?

waebbl:
> It should, it checks the whole tree. Possibly your local tree was significantly behind the main tree?

No, I don't think so. I always rebase before pushing, so the local tree was up to date.

waebbl:
Just checked the gmsh-4.8.4 build file. From my understanding the packages doesn't need to have mpi support for med enabled unconditionally. So instead of masking gmsh[med] as well, I suggest to change the dependency on sci-libs/med of all gmsh versions starting 4.6.0 and above from `sci-libs/med[mpi]` to `sci-libs/med[mpi=]` which should resolve the break.

PR #22100, opened 2021-08-25
net-im/webex: version bump to 41.9.0.20091

mijaros:
Closes: https://bugs.gentoo.org/794700

Signed-off-by: Miroslav Jaros <mirek@mijaros.cz>


mijaros:
One question though, does it even make sense to have the previous ebuilds in the repository? Since it is impossible to download 
 a specific version of the rpm I think that the other ebuilds can be safely removed.

mschilli87:
Meanwhile, upstream appears to have moved on to `41.10.0.20280`.
@akhuettel: Would it make sense for @mijaros to update this PR (or someone else to open a new one superceeding this one) or will this remain stale anyways? :wink:

---
*Off-topic:* Also, if a version bump is affectively only a rename aynways, why not create a `-9999` ebuild and symlink to it?
IMHO, this would make the situtation more transparent.

thesamesam:
> Meanwhile, upstream appears to have moved on to `41.10.0.20280`. @akhuettel: Would it make sense for @mijaros to update this PR (or someone else to open a new one superceeding this one) or will this remain stale anyways? 😉
> 

akhuettel dropped himself from webex so this is m-n now. Needs somebody to either update this PR or do a new one (you? 😄)

> _Off-topic:_ Also, if a version bump is affectively only a rename aynways, why not create a `-9999` ebuild and symlink to it? IMHO, this would make the situtation more transparent.

The hard part would be knowing when you can get an update I guess, but we could do PROPERTIES="live", yes. Symlinks aren't allowed though.

mijaros:
@thesamesam I will look into this later today if there is a possibility to do so. I'm not that strong with ebuilds but let this be a great opportunity to learn some more :)

net-im/webex/webex-41.9.0.20091.1.ebuild
```diff
@@ -0,0 +1,67 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit desktop xdg rpm
+
+DESCRIPTION="Cisco video conferencing and online meeting software"
+HOMEPAGE="https://www.webex.com/"
+SRC_URI="https://binaries.webex.com/WebexDesktop-CentOS-Official-Package/Webex.rpm -> ${P}.rpm"
```
juippis: This is very unfortunate. Isn't there anywhere where this package can be downloaded with a fixed version? For example, I can't even test this due to:
```
!!! Fetched file: webex-41.9.0.20091.1.rpm VERIFY FAILED!
!!! Got:      239953332
!!! Expected: 232153632
```
The `RESTRICT="mirror"` makes it impossible for us to merge this PR like this. And it also makes it impossible for any individual to host it.

I see this package is last-rited due to this reason, not much we can do, other than blame upstream...

juippis (commented):
Well there are other nitpicks with this PR but I don't see it getting merged due to the reasons listed.

AndrewAmmerlaan:
Closing due to the removal of net-im/webex.

PR #22118, opened 2021-08-26
games-strategy/0ad: Bump to alpha25b and EAPI=8

holgersson32644:
Upstream bumped 0ad with some fixes, quoting upstream:

- Ranked matches did not longer grant points in the lobby.
- Mod installation no longer require restart.
- Mod installation no longer prints errors when installing mods
  by double-clicking
- Mod installation now overwrites the existing mods to allow
  for easier updates.
- Fishes are now easier to spot on lowest settings with
  the addition of seagulls.

For their full announcement see:
https://play0ad.com/re-release-of-0-a-d-alpha-25-yauna/

This bump hops to EAPI=8 as all used eclasses support it now and drops
an forgotten comment about the bundled premake5.
Additionally it adds a hint about distfile signature verification using
app-crypt/minisign as noted upstream: https://play0ad.com/download/source/


thesamesam:
Thank you!

PR #22126, opened 2021-08-27
dev-util/bazel: remove hack for Gentoo prefix

littlewu2508:
In bazel-3.2 prefix and static-libs use flag is added concerning bug #773982

For bazel-3.7 the issue is gone and specifying static link flags causes build error.

### DIFF
--- a/dev-util/bazel/bazel-3.7.2.ebuild
+++ b/dev-util/bazel/bazel-3.7.2.ebuild
@@ -13,8 +13,7 @@ SRC_URI="https://github.com/bazelbuild/bazel/releases/download/${PV}/${P}-dist.z
 LICENSE="Apache-2.0"
 SLOT="0"
 KEYWORDS="~amd64"
-IUSE="examples tools prefix static-libs"
-REQUIRED_USE="prefix? ( static-libs )"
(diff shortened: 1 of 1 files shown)


heroxbd:
> > (I would use a Bug tag in the first commit for historical references in future.)
> 
> OK I'll split them into 2 PRs. And for the first one I'll reference a resolved bug.

@thesamesam was telling you to specify the bug URL here https://github.com/gentoo/gentoo/pull/22126#issue-981049807

heroxbd:
https://bugs.gentoo.org/808053 is a duplication of https://bugs.gentoo.org/787248 , which is resolved.  Is this PR still relevant?

littlewu2508:
> https://bugs.gentoo.org/808053 is a duplication of https://bugs.gentoo.org/787248 . Both are resolved. Is this PR still relevant?

This PR is for gentoo prefix. Since the gcc-11 issue is fixed, I'll drop my local change that fix gcc-11 issue and only preserve the commit for gentoo prefix.

heroxbd:
Please remember that when you change USE flags, a revision bump is needed.

PR #22136, opened 2021-08-28
dev-libs/spdlog: fix dependency version

stkw0:
Closes: https://bugs.gentoo.org/810496

### DIFF
--- a/dev-libs/spdlog/spdlog-1.9.2.ebuild
+++ b/dev-libs/spdlog/spdlog-1.9.2.ebuild
@@ -25,7 +25,7 @@ BDEPEND="
 	virtual/pkgconfig
 "
 DEPEND="
-	>=dev-libs/libfmt-6.1.2:=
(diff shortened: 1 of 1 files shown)


dev-libs/spdlog/spdlog-1.9.2.ebuild
```diff
@@ -25,7 +25,7 @@ BDEPEND="
 	virtual/pkgconfig
 "
 DEPEND="
-	>=dev-libs/libfmt-6.1.2:=
+	>=dev-libs/libfmt-8.0.0:=
```
ionenwks, line 28: Normally would need a revbump for a rdepend change, but given it's a build failure and there's a binding operator to prevent downgrading, it's all good (just wanted to make a note of it).

ionenwks (approved):
Thanks, I'll additionally sync this change in the live ebuild.

PR #20546, opened 2021-04-26
dev-java/hppc-template-processor: new package 

vaukai:
```
 * Complete command:
 * javac -source 1.8 -target 1.8 -d target/classes -encoding UTF-8 -classpath //usr/share/randomized-runner/lib/randomized-runner.jar://usr/share/assertj-core-2/lib/assertj-core.jar://usr/share/junit-4/lib/junit.jar://usr/share/hamcrest-core-1.3/lib/hamcrest-core.jar://usr/share/cglib-3/lib/cglib.jar://usr/share/ant-core/lib/ant.jar://usr/share/ant-core/lib/ant-bootstrap.jar://usr/share/ant-core/lib/ant-launcher.jar://usr/share/asm-9/lib/asm.jar://usr/share/guava-20/lib/guava.jar://usr/share/junit-4/lib/junit.jar://usr/share/antlr-4/lib/antlr-runtime.jar://usr/share/antlr-4/lib/antlr-tool.jar://usr/share/velocity/lib/velocity.jar://usr/share/commons-io-1/lib/commons-io.jar://usr/share/commons-lang-3.6/lib/commons-lang.jar://usr/share/slf4j-api/lib/slf4j-api.jar://usr/share/animal-sniffer-annotations/lib/animal-sniffer-annotations.jar://usr/share/error-prone-annotations/lib/error-prone-annotations.jar://usr/share/jsr305/lib/jsr305.jar://usr/share/hamcrest-core-1.3/lib/hamcrest-core.jar://usr/share/antlr-3.5/lib/antlr-runtime.jar://usr/share/stringtemplate-4/lib/stringtemplate.jar://usr/share/treelayout/lib/treelayout.jar @sources.lst
 * Compiling ...
warning: [options] bootstrap class path not set in conjunction with -source 8
src/main/java/com/carrotsearch/hppc/generator/parser/SignatureReplacementVisitor.java:21: error: package com.carrotsearch.hppc.generator.parser.Java7Parser does not exist
import com.carrotsearch.hppc.generator.parser.Java7Parser.ClassDeclarationContext;
```
What's there instead is
```
./src/main/antlr4/com/carrotsearch/hppc/generator/parser/Java7Parser.g4
```
How to find out what needs to be done?

PR #22139, opened 2021-08-28
dev-java/commons-lang: restrict to java:1.8

vaukai:
Closes: https://bugs.gentoo.org/779472

Setting SOURCE and TARGET here since it compiled fine using openjdk:8
JAVA_PKG_WANT_SOURCE="1.4"
JAVA_PKG_WANT_TARGET="1.4"

### DIFF
--- a/dev-java/commons-lang/commons-lang-2.6-r2.ebuild
+++ b/dev-java/commons-lang/commons-lang-2.6-r2.ebuild
@@ -1,0 +1,39 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+JAVA_PKG_IUSE="doc source test"
+MAVEN_ID="commons-lang:commons-lang:2.6"
+JAVA_PKG_WANT_SOURCE="1.4"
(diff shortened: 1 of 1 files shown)


fordfrog:
what's the progress on `app-text/languagetool` wrt this version of `commons-lang`? i'd like to see `dev-java/commons-lang:2.1` gone completely...

vaukai:
> what's the progress on `app-text/languagetool` wrt this version of `commons-lang`?

It's the huge dependency tree where I was stuck at https://github.com/gentoo/gentoo/pull/20546.   
But re-thinking removal of `slot 2.1`, why not let it survive as long it's working with `jdk:8` and has no security issues?

`app-text/languagetool` however seems to have never had a stable version and issues on that package IMHO should be handled independent from and not block [unmasking java 11](https://bugs.gentoo.org/810613).

fordfrog:
thanks for the info. so i merged your commit straight to stable, dropped the old version (ad6f15d1fbfe8aa862cfa12cbbe6c5e7d5b139cc), and also removed this old version from deprecation (67450ec4dc95524632d2f3ae461b0532225d7c47).

PR #22140, opened 2021-08-28
mail-filter/rspamd: correct LuaBitOp dependency rule

arkamar:
Rspamd requires LuaBitOp dependency for all lua targets.

### DIFF
--- a/mail-filter/rspamd/rspamd-3.0-r1.ebuild
+++ b/mail-filter/rspamd/rspamd-3.0-r1.ebuild
@@ -30,7 +30,7 @@ REQUIRED_USE="${LUA_REQUIRED_USE}
 RDEPEND="${LUA_DEPS}
 	$(lua_gen_cond_dep '
 		dev-lua/LuaBitOp[${LUA_USEDEP}]
-	' lua5-{1,2})
+	')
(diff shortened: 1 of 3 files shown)


mail-filter/rspamd/rspamd-3.0-r1.ebuild
```diff
@@ -30,7 +30,7 @@ REQUIRED_USE="${LUA_REQUIRED_USE}
 RDEPEND="${LUA_DEPS}
```
thesamesam, line 30: I'd revbump for such a change, but I can do this on merge.

arkamar:
Thanks!

Dessa:
are you sure thats correct? upstream doesn't include LuaBitOp in the luajit case, unless im missing something

arkamar:
Yes, I am sure. `luajit` implements `bit` and this is the reason why upstream doesn't build it in again when `luajit` is used. However, the `LuaBitOp` package is done in such a way that it does not install any `bit.so` for `luajit`, therefore, it serves just like some sort of placeholder in that case and `luajit` will use its internal implementation anyway. I decided to change it this way because I wanted to have only one place where supported Lua versions are set.

PR #22153, opened 2021-08-30
package.mask: last-rite games-puzzle/gnudoku

DarthGandalf:
Bug: https://bugs.gentoo.org/711344


chewi:
It would be easily fixed and the GTK2 removal makes me uneasy in general, but there are other sudoku games and we're almost the only distro still carrying this.

PR #22158, opened 2021-08-30
Fixed dependencies in app-admin/perl-cleaner package.

wikjas:
The defined runtime dependencies: sys-apps/portage app-portage/portage-utils sys-apps/pkgcore are not needed.
It was tested while the package was cross-compiled.


mattst88:
perl-cleaner definitely depends on these packages at runtime. It calls them directly.

PR #22163, opened 2021-08-30
x11-libs/libdrm: add "udev" USE flag

Karlson2k:
`udev` support switches internal logic and does not require any external libs.

The main difference is [here](https://gitlab.freedesktop.org/mesa/drm/-/blob/libdrm-2.4.107/xf86drm.c#L740). Without `udev` libdrm may try to create by itself missing device node when device is opened, while with `udev` libdrm may wait up to 1 millisecond for device node to appear.


mattst88:
Why would a user want this code enabled vs disabled? Should it be a USE flag, or should it always be enabled?

PR #22164, opened 2021-08-30
perl-core/*: remove last-rited pkgs

jsmolic:


profiles/package.mask
```diff
@@ -260,18 +260,6 @@ app-text/fb2edit
...
-# Andreas K. Hüttel <dilfridge@gentoo.org> (2021-07-31)
-# Obsolete; all versions in current Perl core distributions
-# are newer, and no virtuals currently pull these packages.
-# Removal in 30 days.
-perl-core/ExtUtils-MakeMaker
-perl-core/ExtUtils-Manifest
-perl-core/File-Path
-perl-core/Getopt-Long
-perl-core/HTTP-Tiny
-perl-core/JSON-PP
```
jsmolic, line 272: I see perl-core/JSON-PP and perl-core/ExtUtils-MakeMaker also appear in perl ebuilds (https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-lang/perl/perl-5.34.0-r2.ebuild#n93), do we need to remove them from there as well?

akhuettel: No, we need to keem them there. It's not impossible the perl-core packages will come back at some point in the future.

akhuettel (approved):

PR #22165, opened 2021-08-30
games-puzzle/pipewalker: fix deps, bump eapi

DarthGandalf:
Closes: https://bugs.gentoo.org/739316

### DIFF
--- a/games-puzzle/pipewalker/pipewalker-0.9.4-r2.ebuild
+++ b/games-puzzle/pipewalker/pipewalker-0.9.4-r2.ebuild
@@ -1,8 +1,8 @@
-# Copyright 1999-2018 Gentoo Foundation
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
-inherit desktop flag-o-matic
+EAPI=8
+inherit autotools desktop flag-o-matic toolchain-funcs
 
 DESCRIPTION="Rotating pieces puzzle game"
 HOMEPAGE="http://pipewalker.sourceforge.net/"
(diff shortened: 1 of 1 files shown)


games-puzzle/pipewalker/pipewalker-0.9.4-r2.ebuild
```diff
@@ -11,16 +11,22 @@ SRC_URI="mirror://sourceforge/pipewalker/${P}.tar.gz"
...
 
+src_prepare() {
+	default
+	eautoreconf
+}
+
 src_configure() {
-	append-flags $(sdl-config --cflags)
-	econf --datadir="/usr/share"
+	append-cflags $(sdl-config --cflags)
```
SoapGentoo, line 28: check exactly what `sdl-config --cflags` returns here, because my feeling is this might need to be `append-cppflags`

games-puzzle/pipewalker/pipewalker-0.9.4-r2.ebuild
```diff
@@ -11,16 +11,22 @@ SRC_URI="mirror://sourceforge/pipewalker/${P}.tar.gz"
...
+	econf
```
SoapGentoo, line 29: change to `default`

SoapGentoo (changes requested):

games-puzzle/pipewalker/pipewalker-0.9.4-r2.ebuild
chewi: If you're needing to get flags manually like this then I'd much prefer you use `pkg-config` (`tc-getPKG_CONFIG`) instead.

ionenwks: econf is probably using sdl-config --libs /still/, but pkg-config is still more right even if it mismatches. Optimally I wish sdl-config wasn't used at all.

That aside, it's often overlooked but sdl-config/pkg-config calls should get `|| die` as well, I had a case where build succeeded despite pkg-config failed and resulted in an underlinked shared library causing issues for other packages.

Usually I do something [like this](https://github.com/gentoo/gentoo/blob/cf72416b29d1fc39f978044b7da63d909d90bd24/games-rpg/openglad/openglad-0.98_p20200907.ebuild#L51) anyhow.

DarthGandalf: Done

PR #22167, opened 2021-08-31
dev-lang/perl: append required cflags when x-compiling

jepio:
Bug: https://bugs.gentoo.org/811237
Reported-by: Github User DRKV333
Signed-off-by: Jeremi Piotrowski <jpiotrowski@microsoft.com>


thesamesam:
Thank you for the patch and explanation!

PR #22168, opened 2021-08-31
app-arch/lz4: Add USE=lib-only

jsmolic:
* Add lib-only flag which provides a minimal lz4 build
  with only library and without other binaries. This is
  useful in e.g. embedded or stripped down systems where
  one might want just the core functionality of liblz4.

Signed-off-by: Jakov Smolić <jsmolic@gentoo.org>

### DIFF
--- a/app-arch/lz4/lz4-1.9.3-r2.ebuild
+++ b/app-arch/lz4/lz4-1.9.3-r2.ebuild
@@ -14,7 +14,7 @@ LICENSE="BSD-2 GPL-2"
 # https://abi-laboratory.pro/tracker/timeline/lz4/
 SLOT="0/r132"
 KEYWORDS="~alpha amd64 arm arm64 hppa ~ia64 ~m68k ~mips ppc ppc64 ~riscv ~s390 sparc x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~sparc-solaris ~sparc64-solaris ~x64-solaris ~x86-solaris"
-IUSE="static-libs"
+IUSE="lib-only static-libs"
 
 CMAKE_USE_DIR=${S}/build/cmake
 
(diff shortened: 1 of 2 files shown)


app-arch/lz4/lz4-1.9.3-r2.ebuild
```diff
@@ -14,7 +14,7 @@ LICENSE="BSD-2 GPL-2"
 # https://abi-laboratory.pro/tracker/timeline/lz4/
 SLOT="0/r132"
 KEYWORDS="~alpha amd64 arm arm64 hppa ~ia64 ~m68k ~mips ppc ppc64 ~riscv ~s390 sparc x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~sparc-solaris ~sparc64-solaris ~x64-solaris ~x86-solaris"
-IUSE="static-libs"
+IUSE="lib-only static-libs"
```
amadio, line 17: I'd consider `IUSE=minimal` instead, as that's an existing global use flag used by many packages.

mgorny:
What's the end goal here? Given that lz4 executable is 80k in size (+3k for manpage), this sounds like a serious overkill. For a start, it's going to mean that we're going to have to verify all revdeps for whether they're using the library or the executable, and force `[-lib-only(-)]` on all of them.

This really looks like a case for `INSTALL_MASK` rather than USE flags. Though in either case you're asking for trouble if you're installing an archiver without its CLI.

jsmolic:
@mgorny Yeah, I'm aware about the size, although it still matters when you're fighting for every kb but I admit difference is small, that's why I was unsure and wanted to hear your opinion. I do have a use case for this, but at the same time I don't want to create more unnecessary work given that difference here is very small and I don't know if we want to go through updating all revdeps just for this. In any case if you're not up for accepting this I can opt for a different solution and avoid changing `::gentoo` ebuild

app-arch/lz4/lz4-1.9.3-r2.ebuild
jsmolic: I'm okay with `minimal` as well

mgorny:
Yes, I'd prefer not having this in `::gentoo`. It feels like microoptimizing to me.

jsmolic:
Okay, thanks for your input

PR #22169, opened 2021-08-31
media-libs/libaom: Bump to 3.1.2

novomesk:
This release contains a fix which will be needed by future libavif.

https://bugs.gentoo.org/814002


novomesk:
@Whissi Maybe you can check this again like the last time?

PR #22170, opened 2021-08-31
cmake.eclass, ecm.eclass: Support EAPI-8

a17r:


eclass/cmake.eclass
```diff
@@ -9,38 +9,52 @@
...
 # It provides all inherited features (DOCS, HTML_DOCS, PATCHES) along with
 # out-of-source builds (default), in-source builds and an implementation of the
 # well-known use_enable function for CMake.
 
+case ${EAPI} in
+	7|8) ;;
+	*) die "EAPI=${EAPI:-0} is not supported" ;;
+esac
+
+EXPORT_FUNCTIONS src_prepare src_configure src_compile src_test src_install
```
mgorny, line 25: Please move EXPORT_FUNCTIONS after inherit, it turns out it results in inconsistent behavior (read: random breakage) since Portage changed its behavior in 2009.

(yes, Gentoo's gonna be full of QA warnings about that in 1-2 days)

eclass/ecm.eclass
```diff
@@ -30,6 +30,7 @@ if [[ -v KDE_GCC_MINIMAL ]]; then
 fi
 
 EXPORT_FUNCTIONS pkg_setup src_prepare src_configure src_test pkg_preinst pkg_postinst pkg_postrm
```
mgorny, line 32: Also here. This is the eclass we've actually gotten bugreport against.

eclass/cmake.eclass
```diff
@@ -343,25 +370,36 @@ cmake_src_prepare() {
...
+	if [[ ${EAPI} == 7 ]]; then
+		popd > /dev/null || die
+	fi
 
-	# make ${S} read-only in order to detect broken build-systems
+	# Make ${CMAKE_USE_DIR} (in EAPI-7: ${S}) read-only in order to detect
+	# broken build systems.
 	if [[ ${CMAKE_QA_SRC_DIR_READONLY} && ! ${CMAKE_IN_SOURCE_BUILD} ]]; then
-		chmod -R a-w "${S}"
+		if [[ ${EAPI} == 7 ]]; then
```
DerDakon, line 398: I wonder if it wouldn't make sense to do this ```if``` once at the beginning of the function and assign it to a local variable and just use that everywhere throughout the function without any further EAPI switching.

eclass/cmake.eclass
```diff
@@ -9,38 +9,52 @@
...
-# You usually do *NOT* want nor need to set it as it pulls CMake default
-# build-type specific compiler flags overriding make.conf.
-: ${CMAKE_BUILD_TYPE:=Gentoo}
+#
+# The default is RelWithDebInfo as that is least likely to append undesirable
+# flags. However, you may still need to sed CMake files or choose a different
+# build type to achieve desirable results.
+#
+# In EAPI 7, the default was non-standard build type of Gentoo.
+: ${CMAKE_BUILD_TYPE:=RelWithDebInfo}
```
thesamesam, line 57: I thought I'd replied to mgorny on the [ML](https://archives.gentoo.org/gentoo-dev/message/b9b1f9afc0abd07bba6c5aa62c7443ab) but apparently not. While I wasn't completely sure about changing, using the Gentoo build type doesn't remove the need for care and checking for undesired behaviour in the buidl system -- so I don't have an objection.

thesamesam (approved):

eclass/cmake.eclass
a17r: I don't see much benefit there.

PR #22173, opened 2021-08-31
dev-libs/elfutils: backport patches fixing clang build

10ne1:
Latest elfutils mainline development branch contains the
last two patches required for building elfutils with clang.

This backports the two patches onto v0.185 so we can get
it built with clang and should be dropped when Gentoo
eventually upgrades to v0.186 or later.


thesamesam:
Thanks!

PR #22175, opened 2021-09-01
app-editors/featherpad: Bump to 1.0.0

hansfn:
This is really a minor / patch update. From changelog:

Bumped the version to 1.0.0 (some users may have mistaken
"0.X.Y" for "unstable").


app-editors/featherpad/featherpad-1.0.0.ebuild
```diff
@@ -0,0 +1,28 @@
...
+	dev-qt/qtnetwork:5
+	dev-qt/qtprintsupport:5
+	dev-qt/qtsvg:5
+	dev-qt/qtwidgets:5
+	dev-qt/qtx11extras:5
+	x11-libs/libX11"
+DEPEND="${RDEPEND}"
+BDEPEND="dev-qt/linguist-tools:5"
+
+S="${WORKDIR}/FeatherPad-${PV}"
```
arthurzam, line 28: Please put the variable `S=` a line after `SRC_URI` (without extra line between them) as they are connected.

app-editors/featherpad/featherpad-1.0.0.ebuild
```diff
@@ -0,0 +1,28 @@
...
+
+RDEPEND="app-text/hunspell
+	dev-qt/qtcore:5
+	dev-qt/qtgui:5
+	dev-qt/qtnetwork:5
+	dev-qt/qtprintsupport:5
+	dev-qt/qtsvg:5
+	dev-qt/qtwidgets:5
+	dev-qt/qtx11extras:5
+	x11-libs/libX11"
```
arthurzam, line 24: Looking upstream [here](https://github.com/tsujan/FeatherPad/blob/master/CMakeLists.txt#L18) and [here](https://github.com/tsujan/FeatherPad/blob/master/featherpad/CMakeLists.txt#L26-L29) it looks like we can make those optional dependencies with an `IUSE="X"`.

arthurzam (changes requested):
Thank you for the Pull Request.
Looks good, but some small changes

app-editors/featherpad/featherpad-1.0.0.ebuild
hansfn: Sure, I can do that. Is it something new? All other ebuilds under app-editors seems to do it similar to what I have used.

app-editors/featherpad/featherpad-1.0.0.ebuild
hansfn: I'm aware of this - it is documented in the INSTALL file:

> *******************************
> *   Compilation without X11   *
> *******************************
> 
> If, for whatever reason, you do not want the X11 support, you could first turn on the option "WITHOUT_X11" with cmake:

I can add the USE flag. Will update the PR.

app-editors/featherpad/featherpad-1.0.0.ebuild
arthurzam: You can always reference the `skel.ebuild` file located at the root of the tree, to know the basic order of variables.
The change itself is quite old (have been done [here](https://github.com/gentoo/gentoo/commit/0f936ec81f1dc20fa58ba40bc17989ee44810214)).

About the other ebuilds, I guess they just haven't been corrected since.

app-editors/featherpad/featherpad-1.0.0.ebuild
```diff
@@ -0,0 +1,38 @@
...
+DESCRIPTION="Lightweight Qt5 Plain-Text Editor for Linux"
+HOMEPAGE="https://github.com/tsujan/FeatherPad"
+SRC_URI="https://github.com/tsujan/FeatherPad/archive/V${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="+X"
+
+RDEPEND="app-text/hunspell
```
arthurzam, line 17:
```suggestion
RDEPEND="app-text/hunspell:=
```

`app-text/hunspell` has a subslot, meaning we want to auto rebuild `featherpad` when the subslot changes.
For more info, you can read the [devmanual](https://devmanual.gentoo.org/general-concepts/slotting/index.html#sub-slots).

app-editors/featherpad/featherpad-1.0.0.ebuild
hansfn: Thank you for the `skel.ebuild` reference.

app-editors/featherpad/featherpad-1.0.0.ebuild
```diff
@@ -0,0 +1,38 @@
...
+
+DESCRIPTION="Lightweight Qt5 Plain-Text Editor for Linux"
+HOMEPAGE="https://github.com/tsujan/FeatherPad"
+SRC_URI="https://github.com/tsujan/FeatherPad/archive/V${PV}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/FeatherPad-${PV}"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="+X"
```
juippis, line 16: Just a quick note, we try to use `gui` USE flag for this going to the future.

hansfn: Thx, but this flag isn't about GUI or not. From the INSTALL file (and from my own testing):

> If, for whatever reason, you do not want the X11 support, you could first turn on the option "WITHOUT_X11" with cmake:
> [...]
> The result will have all features except for virtual desktop awareness and tab DND.

I think using "gui" as use flag will be confusing.

arthurzam: Lets remain with `X` USE flag, but add an explanation what it does.
Please add to the `metadata.xml` an explanation on the USE flag, so it won't confuse.
You can look at the documentation [here](https://devmanual.gentoo.org/ebuild-writing/misc-files/metadata/index.html), and an example [here](https://github.com/gentoo/gentoo/blob/master/app-text/sigil/metadata.xml#L8-L11).

juippis: Yeah, sorry, mistook it toggled GUI with a configure option `--with-X` like many other programs do. Glad you checked upstream!

hansfn: Have added use flag description.

juippis (approved):
Thanks a lot to both of you! 

@arthurzam I'll do the honours since I'm around at the moment and ready to push in a minute :)

PR #22177, opened 2021-09-01
net-im/telegram-desktop: Bump to 3.0.0

mid-kid:
@gyakovlev


gyakovlev:
thanks, works fine on ppc64 here.

PR #22183, opened 2021-09-01
app-admin/puppet-agent: symlink fix

pdemonaco:
Both 7.9 and 7.10 installed a broken symlink to the libcrypt.so.1
library which cause the agent to fail. There may be some systems where
the appropriate library still appears first in the system search path,
however, on my system the glibc version is selected and it does not
contain `XCRYPT_2.0` which ruby apparently requires.

Closes: https://bugs.gentoo.org/809263
Signed-off-by: Philip DeMonaco <phil@demona.co>

### DIFF
--- a/app-admin/puppet-agent/puppet-agent-6.19.1-r1.ebuild
+++ b/app-admin/puppet-agent/puppet-agent-6.19.1-r1.ebuild
@@ -72,7 +72,16 @@ src_install() {
 	dosym ../../opt/puppetlabs/bin/hiera /usr/bin/hiera
 	dosym ../../opt/puppetlabs/bin/puppet /usr/bin/puppet
 	dosym ../../opt/puppetlabs/puppet/bin/virt-what /usr/bin/virt-what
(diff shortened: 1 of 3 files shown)


thesamesam:
This isn't quite right. The location of the libxcrypt libraries depends on USE=system on it. There's currently a [migration](https://www.gentoo.org/support/news-items/2021-07-23-libxcrypt-migration.html) underway in ~arch.

I guess you could have a USE flag which requires system to be enabled on libxcrypt, and in that case, use the old symlink path (current, I mean), and another which requires it to be off which uses the "hidden" libxcrypt folder.

pdemonaco:
@thesamesam good point - I've built logic to detect the presence of the crypt flag on sys-libs/glibc based on your suggestion. What do you think?

prometheanfire:
merged this, thanks for the work here and sorry I didn't see it earlier

pdemonaco:
@prometheanfire you're welcome - the "good" fix wasn't pushed until today anyway so by some metrics you've got a sub 2 hour turn-around!

PR #22184, opened 2021-09-01
Dev util/cppunit: EAPI version bump to 8

donny-nyc:
Closes: https://bugs.gentoo.org/show_bug.cgi?id=811345

Trivial Update: bumps EAPI version to 8 for all ebuilds, and volunteers myself as maintainer


donny-nyc:
I had mistyped the name tag for the maintainer. Just pushed a patch to correct.

juippis (commented):
First of all thanks for your contribution and interest in maintaining packages in Gentoo! And here comes the but...

We're quite reluctant in adding "unknown" contributors as maintainers. That won't stop you from contributing, we'll gladly accept all your work towards maintainer-needed packages, and some projects also accept user contributions (but not all - recommend to start with maintainer-needed packages and go from there).

Then to this particular case. Normally we'd revbump when EAPI-bumping since EAPI might bring new configure options, or provide updated ways of handling `src_install` essentially changing the list of installed files. Some developers do EAPI-bump in place though, **if** they can be sure it won't break anything. Looking at this package, it has rather scary reverse dependencies, so I wouldn't like to EAPI-bump without revbumping it. Besides you went from 6->8 without making EAPI-7 relevant changes. 

Here's my suggestion and commits in logical order: 
 - revbump 1.15.1-r2 to 1.15.1-r3 and this revbump includes steps:
      * bumping EAPI to 8, 
      * reset KEYWORDS to ~ with `ekeyword ~all cppunit-1.15.1-r3.ebuild`, 
      * clean the ebuild - take out -9999 parts, take out useless DEPEND="${RDEPEND}" and RDEPEND="", check how much space installing examples take and if it's few kilobytes, install them unconditionally.
 - drop -9999 ebuild altogether if CI allows, 
 - drop cppunit-1.14.0.ebuild if CI allows, 
 - don't become a maintainer yet
 - since nothing seems to depend on cppunit[static-libs], try dropping the 'static-libs' IUSE from 1.15.1-r3 in a separate commit so the commit can be dropped if it breaks stuff.

EDIT: Github formatting

donny-nyc:
Thanks for all the great feedback @juippis, I'll make sure to include all of your suggestions (the breakdown is helpful).

I understand why unknown developers wouldn't be added as maintainers right away, but am excited to learn more and work towards that role. This is a great project, and i"m excited to find ways to contribute.

donny-nyc:
@juippis I had to keep the 1.14.0 ebuild to satisfy the dependency issues listed above.

The compiled examples added up to just under 1MB, so I kept the conditional statement:
```
879K examples/simple/ExampleTestCase.o
63K examples/simple/Main.o
```

No failures in the test suite:
```
Testsuite summary for cppunit 1.15.1
# TOTAL: 1
# PASS:  1
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

Please let me know if any additional changes are needed.

juippis (approved):
Should've been clearer, I meant to make a new file -r3 based on -r2, so -r2 could keep the stable keywords. That's probably why 1.14 couldn't be dropped. 

Thanks, I can work with this, I'll attempt the cleanup post-merge!

PR #22191, opened 2021-09-02
app-crypt/hashcat: update to 6.2.4 and fix compilation without brain

Miezhiko:


app-crypt/hashcat/files/hashcat-6.2.4-brainless.patch
```diff
@@ -0,0 +1,40 @@
+From 0d5aed883b45cf3c218cb71916b08d56ae3d25a3 Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: Wow, what a coincidence. I was looking at this last night.

Can you report + submit this upstream please if you haven't already, and pop the link at the top of the patch?

Also, please keep the old version in tree (or at least drop the old one in a new commit) as hashcat is sometimes fragile.

Miezhiko: https://github.com/hashcat/hashcat/pull/2961

Miezhiko: @thesamesam I'll message maintainer if/when my PR will be merged upstream, that was required.

Miezhiko: @thesamesam  done, also PR was merged to upstream.

thesamesam:
Merged as 1632d007342a9cf30f0b201fb2085e9bdc2e96c2, thanks!

PR #22193, opened 2021-09-02
media-gfx/hydrus: Drop 448-r1, add 453

vaartis:
No changes to the ebuild 

@juippis


juippis (approved):
Thanks! The tests hang for me, but program works fine. 

FYI you don't have to ping us separately, we get the notifications already through the @gentoo-bot

vaartis:
> Thanks! The tests hang for me, but program works fine.
> 
> FYI you don't have to ping us separately, we get the notifications already through the @gentoo-bot

Ah, sorry.. I believe the tests hang sometimes for no apparent reason. I thought it was because the program is running, but they haven't been hanging for me for a long time, even with a running program.

PR #22198, opened 2021-09-02
app-editors/kakoune: bump 2021.08.28

mujo-hash:
Bump to latest release. Drop old unstable ebuilds.
I've upstreamed the gentoo ebuild filetype detection patch and this release builds with gcc 11 so I dropped the old patches.
There was a new unrelated bug I've discovered and have an on PR regarding upstream, so I've included the patch with the ebuild as it is too late to make it into this release.


app-editors/kakoune/kakoune-2021.08.28.ebuild
```diff
@@ -0,0 +1,46 @@
...
+DESCRIPTION="Modal editor inspired by vim"
+HOMEPAGE="http://kakoune.org/ https://github.com/mawww/kakoune"
+SRC_URI="https://github.com/mawww/kakoune/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="Unlicense"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~x86"
+
+DEPEND=""
+RDEPEND="${DEPEND}"
```
juippis: This can be removed as it's useless. I'll do it so no need to re-push.

juippis (approved):
Thanks!

PR #22200, opened 2021-09-03
net-analyzer/suricata: Added hyperscan USE flag.

jpds:
Tested on `amd64`.

### DIFF
--- a/net-analyzer/suricata/metadata.xml
+++ b/net-analyzer/suricata/metadata.xml
@@ -12,6 +12,7 @@
     <flag name="control-socket">Enable unix socket</flag>
     <flag name="cuda">Enable NVIDIA Cuda computations support</flag>
     <flag name="detection">Enable detection modules</flag>
+    <flag name="hyperscan">Enable high-performance regex matching with Hyperscan</flag>
     <flag name="lz4">Enable support for compressed pcap logging using the LZ4 algorithm</flag>
     <flag name="nflog">Enable libnetfilter_log support</flag>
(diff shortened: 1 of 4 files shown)


net-analyzer/suricata/metadata.xml
```diff
@@ -12,6 +12,7 @@
     <flag name="control-socket">Enable unix socket</flag>
     <flag name="cuda">Enable NVIDIA Cuda computations support</flag>
     <flag name="detection">Enable detection modules</flag>
+    <flag name="hyperscan">Enable Hyperscan support</flag>
```
mkszuba, line 15: I would very much prefer a more detailed description here. Yes, I know some of the existing ones are very terse too.

profiles/arch/riscv/use.mask
```diff
@@ -15,6 +15,10 @@
 -gtk3
 -introspection
 
+# Jonathan Davies (2021-09-03)
+# No arch support
+hyperscan
```
mkszuba, line 20: This should be a suricata-specific mask in package.use.mask, global masks of local USE flags are frowned upon. It would be nice to have a more detailed explanation (_i.e._ that Hyperscan only works on CPUs with SSSE3) as well.

mkszuba (changes requested):
In addition to the specific comments inline, please increment revisions in both cases - USE=hyperscan might be off by default but it still does change RDEPEND.

mkszuba:
LGTM now! Hyperscan support is currently automagical but it's not a problem with your PR, will take care of it later myself.

PR #22203, opened 2021-09-03
Colm/ragel shared libs

juippis:


juippis:
I dropped prefix keywords because looking at the ragel patch, no clue whether it works on prefix or not. And I got no way to test it either on prefix. But on ~unstable seems to work.

lu-zero:
It looks fine to me.

PR #22207, opened 2021-09-03
games-puzzle/arrows: EAPI8 bump, ebuild improvements

mm1ke:
@ionenwks 

New revision for `games-puzzle/arrows` since lots of dependencies were missing.
I've also added a desktop entry because why not :) This also fixes calling `cc` directly.


games-puzzle/arrows/arrows-0.6-r2.ebuild
```diff
@@ -0,0 +1,59 @@
...
+RDEPEND="
+	dev-libs/atk
+	dev-libs/glib
+	media-libs/fontconfig
+	media-libs/freetype
+	media-libs/harfbuzz
+	x11-libs/cairo
+	x11-libs/gdk-pixbuf
+	x11-libs/gtk+:2
+	x11-libs/pango"
```
ionenwks, line 25: Yeah, the missing deps is a bad habit that "gtk+ is there and depends on these, so all these other deps don't need to be listed" that some people have, so this is fine :)

Just to be picky, usually define the slot of a few of these too and harfbuzz is missing :=, my experimental `app-portage/iwdevtools` is giving:
```
 * QA: comparing =games-puzzle/arrows-0.6-r1 with games-puzzle/arrows-0.6-r2/image
 *  FILES:+usr/share/applications/arrows-arrows.desktop
 * ------> FILES(+1)
 * QA: detected possibly incorrect RDEPEND (games-puzzle/arrows-0.6-r2)
 * dev-libs/glib       | dev-libs/glib:2
 * media-libs/harfbuzz | media-libs/harfbuzz:=
 * x11-libs/gdk-pixbuf | x11-libs/gdk-pixbuf:2
```
And yes, good on adding that .desktop :) All desktop applications should have them even if not always perfect.

games-puzzle/arrows/arrows-0.6-r2.ebuild
```diff
@@ -0,0 +1,59 @@
...
+		game.c || die
+	sed -i \
+		-e '/^CC /d' \
+		-e '/CCLIBS/s|$| $(LDFLAGS)|' \
+		Makefile || die
+}
+
+src_compile() {
+	tc-export CC
+	make clean || die
```
ionenwks, line 47:
```suggestion
	emake clean
```

games-puzzle/arrows/arrows-0.6-r2.ebuild
mm1ke: Thanks for the review. I saw your `app-portage/iwdevtools` bumps recently but didn't looked into it what it does, but your output got me interested. Going to have a look at them :)

ionenwks: I should've realized, /before/ the `-Wl, --as-needed` ordering was wrong and now that using LINKOPTS it's allowing it to work as it should (issue is that libraries need to be listed last, but the former sed was adding LDFLAGS after CCLIBS which is entirely wrong).

Now we have:
```
$ scanelf -n /usr/bin/arrows 
 TYPE   NEEDED FILE 
ET_DYN libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libpango-1.0.so.0,libgobject-2.0.so.0,libglib-2.0.so.0,libc.so.6 /usr/bin/arrows
```

And my scripts now switched to (the `<` is suggesting these may be possible to remove)
```
 * QA: detected possibly incorrect RDEPEND (games-puzzle/arrows-0.6-r2)
 * dev-libs/atk          < 
 * media-libs/fontconfig < 
 * media-libs/freetype   < 
 * media-libs/harfbuzz:= < 
 * x11-libs/cairo        < 
 * x11-libs/gdk-pixbuf:2 < 
```
If it's no longer actually linked with those (i.e. only injected by gtk and then not used), in these case it's right not to list them, i.e. even if gtk stopped using one of these and it was depcleaned, it wouldn't break this (neither runtime nor build-time give gtk's pkg-config file wouldn't try to use it anymore).

Some of the ones you added still need to stay though, like glib:2 and pango. I went ahead and gave it the finishing touches -- it's just a few lines to remove :)

PR #22208, opened 2021-09-03
x11-misc/shutter: version 0.99

DarthGandalf:


x11-misc/shutter/shutter-0.99.ebuild
```diff
@@ -0,0 +1,81 @@
...
+	doins share/appdata/shutter.appdata.xml
+
+	# .po doesn't belong to installed system, only .mo
+	rm -r "${ED}"/usr/share/shutter/resources/po || die
+
+	# shutter executes perl scripts as standalone scripts, and after that "require"s them.
+	find "${ED}"/usr/share/shutter/resources/system/plugins/ -type f ! -name '*.*' -exec chmod 755 {} \; \
+		|| die "failed to make plugins executables"
+	find "${ED}"/usr/share/shutter/resources/system/upload_plugins/upload -type f \
+		-name "*.pm" -exec chmod 755 {} \; || die "failed to make upload plugins executables"
```
ionenwks, line 73: `{} \;` has a bit of a problem, e.g. try to run:
```
find /some/dir/with/files -exec false {} \; || echo failed
```
It won't echo failed, but:
```
find /some/dir/with/files -exec false {} + || echo failed
```
will.

Differences is that + will run the command only once (pass all arguments to it), then give the return code. `\;` calls it for as many files there is and ignores return codes of each. So, unless can only take 1 argument, `\;` should never be used (it's also inefficient to do all these calls either way).

On a related note, adding a check for if `find` matched no files is on my TODO for iwdevtools ;p (given it doesn't die). Albeit will be purely informational given it's sometime used for this purpose in the first place. For this ebuild, finding no files would be unexpected though.

While here, may want to use `chmod +x` rather than the full 755 too.

x11-misc/shutter/shutter-0.99.ebuild
```diff
@@ -0,0 +1,81 @@
...
+
+EAPI=8
+
+inherit desktop optfeature prefix xdg
+
+DESCRIPTION="Feature-rich screenshot program"
+HOMEPAGE="https://shutter-project.org/"
+SRC_URI="https://github.com/shutter-project/shutter/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3"
```
ionenwks, line 12:
```
#  the Free Software Foundation; either version 3 of the License, or
#  (at your option) any later version.
```

DarthGandalf:
Fixed. Also added a missing BDEPEND

PR #22209, opened 2021-09-04
media-fonts/kanjistrokeorders: Bump to 4.004

ShadowsFriend:
Update homepage and SRC_URI.

Signed-off-by: Christian Widmer <cwidmer@umbrox.de>


media-fonts/kanjistrokeorders/kanjistrokeorders-4.004.ebuild
```diff
@@ -0,0 +1,23 @@
+# Copyright 2009-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit font
+
+DESCRIPTION="Font for viewing stroke order diagrams for kanji, kana and other characters"
+HOMEPAGE="https://www.nihilist.org.uk/"
+SRC_URI="https://drive.google.com/uc?export=download&id=1snpD-IQmT6fGGQjEePHdDzE2aiwuKrz4 -> ${P}.zip"
```
amadio, line 10: Is this really the only official download link?

ShadowsFriend: It is the one listed on https://www.nihilist.org.uk/ as the official download link, which makes me reasonably assume that. The page also lists an alternative download link on dropbox.
And while https://sites.google.com/site/nihilistorguk/, which is where HOMEPAGE in the v4.002 package from the official tree is pointing to, is still working, the readme from that package lists http://www.nihilist.org.uk/.

PR #22210, opened 2021-09-04
Remove dev-util/pkgconfig

jsmolic:


profiles/package.mask
```diff
@@ -175,16 +175,6 @@ dev-python/pydiff
...
-# David Seifert <soap@gentoo.org> (2021-08-04)
-# Last release over 4 years ago, upstream pretty much dead, the
-# ecosystem has switched to dev-util/pkgconf, which is alive. Testing
-# and prefix bugs, blocks WANT_AUTOMAKE=1.12 removal.
-# (If you hit blockers, please run:
-# $ emerge --deselect dev-util/pkgconfig
-# This is necessary as your world file should not contain dependencies.)
-# Bug #245228, #632124, #691268, #767853, removal in 30 days.
-dev-util/pkgconfig
-
```
SoapGentoo, line 187: I would leave the blocker, as it helps people with `dev-util/pkgconfig` in their world file

SoapGentoo (changes requested):

profiles/package.mask
jsmolic: makes sense, done ;)

leio:
It doesn't close https://bugs.gentoo.org/245228 - the problem still exists with pkgconf; so it's more about virtual/pkgconfig there, but when the bug was filed, pkgconf probably didn't exist yet.

PR #22211, opened 2021-09-04
app-editors/scite: version bump to 5.1.1

dermoench42:
Upstream Bugfixes extracted lexer lib,
compatible with gcc-11 (bug #786447)

Bug: https://bugs.gentoo.org/786447
Signed-off-by: Ervin Peters coder@ervnet.de


juippis (approved):
Thanks, EAPI-8 should work, so I'll test and bump it!

PR #22217, opened 2021-09-05
Drop libudev static libs support

floppym:
Upstream really isn't interested in supporting this.

https://github.com/systemd/systemd/issues/20600


tamiko (approved):

gyakovlev:
ack on zfs

PR #22221, opened 2021-09-05
dev-lang/rust: fix build for libcxx/musl.

ahmedcharles:
Signed-off-by: Ahmed Charles <me@ahmedcharles.com>

### DIFF
--- a/dev-lang/rust/rust-1.61.0-r2.ebuild
+++ b/dev-lang/rust/rust-1.61.0-r2.ebuild
@@ -41,7 +41,7 @@ LLVM_TARGET_USEDEPS=${ALL_LLVM_TARGETS[@]/%/(-)?}
 
 LICENSE="|| ( MIT Apache-2.0 ) BSD-1 BSD-2 BSD-4 UoI-NCSA"
 
-IUSE="clippy cpu_flags_x86_sse2 debug dist doc miri nightly parallel-compiler profiler rls rustfmt rust-src system-bootstrap system-llvm test wasm ${ALL_LLVM_TARGETS[*]}"
+IUSE="clippy cpu_flags_x86_sse2 debug dist doc llvm-libunwind miri nightly parallel-compiler profiler rls rustfmt rust-src system-bootstrap system-llvm test wasm ${ALL_LLVM_TARGETS[*]}"
 
 # Please keep the LLVM dependency block separate. Since LLVM is slotted,
 # we need to *really* make sure we're not pulling more than one slot
@@ -105,7 +105,10 @@ DEPEND="
 	net-misc/curl:=[http2,ssl]
 	sys-libs/zlib:=
 	dev-libs/openssl:0=
-	elibc_musl? ( sys-libs/libunwind:= )
+	elibc_musl? (
+		llvm-libunwind? ( sys-libs/llvm-libunwind:= )
+		!llvm-libunwind? ( sys-libs/libunwind:= )
+	)
 	system-llvm? ( ${LLVM_DEPEND} )
 "
 
@@ -397,6 +400,7 @@ src_configure() {
 		dist-src = false
 		remap-debuginfo = true
 		lld = $(usex system-llvm false $(toml_usex wasm))
+		llvm-libunwind = "$(usex elibc_musl system no)"
 		# only deny warnings if doc+wasm are NOT requested, documenting stage0 wasm std fails without it
 		# https://github.com/rust-lang/rust/issues/74976
 		# https://github.com/rust-lang/rust/issues/76526
--- a/dev-lang/rust/rust-1.62.0.ebuild
+++ b/dev-lang/rust/rust-1.62.0.ebuild
@@ -41,7 +41,7 @@ LLVM_TARGET_USEDEPS=${ALL_LLVM_TARGETS[@]/%/(-)?}
 
 LICENSE="|| ( MIT Apache-2.0 ) BSD-1 BSD-2 BSD-4 UoI-NCSA"
 
-IUSE="clippy cpu_flags_x86_sse2 debug dist doc miri nightly parallel-compiler profiler rls rustfmt rust-src system-bootstrap system-llvm test wasm ${ALL_LLVM_TARGETS[*]}"
+IUSE="clippy cpu_flags_x86_sse2 debug dist doc llvm-libunwind miri nightly parallel-compiler profiler rls rustfmt rust-src system-bootstrap system-llvm test wasm ${ALL_LLVM_TARGETS[*]}"
 
 # Please keep the LLVM dependency block separate. Since LLVM is slotted,
 # we need to *really* make sure we're not pulling more than one slot
@@ -105,7 +105,10 @@ DEPEND="
 	net-misc/curl:=[http2,ssl]
 	sys-libs/zlib:=
 	dev-libs/openssl:0=
-	elibc_musl? ( sys-libs/libunwind:= )
+	elibc_musl? (
+		llvm-libunwind? ( sys-libs/llvm-libunwind:= )
+		!llvm-libunwind? ( sys-libs/libunwind:= )
+	)
 	system-llvm? ( ${LLVM_DEPEND} )
 "
 
@@ -401,6 +404,7 @@ src_configure() {
 		dist-src = false
 		remap-debuginfo = true
 		lld = $(usex system-llvm false $(toml_usex wasm))
+		llvm-libunwind = "$(usex elibc_musl system no)"
 		# only deny warnings if doc+wasm are NOT requested, documenting stage0 wasm std fails without it
 		# https://github.com/rust-lang/rust/issues/74976
 		# https://github.com/rust-lang/rust/issues/76526
--- a/dev-lang/rust/rust-1.62.1.ebuild
+++ b/dev-lang/rust/rust-1.62.1.ebuild
@@ -41,7 +41,7 @@ LLVM_TARGET_USEDEPS=${ALL_LLVM_TARGETS[@]/%/(-)?}
 
 LICENSE="|| ( MIT Apache-2.0 ) BSD-1 BSD-2 BSD-4 UoI-NCSA"
 
-IUSE="clippy cpu_flags_x86_sse2 debug dist doc miri nightly parallel-compiler profiler rls rustfmt rust-src system-bootstrap system-llvm test wasm ${ALL_LLVM_TARGETS[*]}"
+IUSE="clippy cpu_flags_x86_sse2 debug dist doc llvm-libunwind miri nightly parallel-compiler profiler rls rustfmt rust-src system-bootstrap system-llvm test wasm ${ALL_LLVM_TARGETS[*]}"
 
 # Please keep the LLVM dependency block separate. Since LLVM is slotted,
 # we need to *really* make sure we're not pulling more than one slot
@@ -105,7 +105,10 @@ DEPEND="
 	net-misc/curl:=[http2,ssl]
 	sys-libs/zlib:=
 	dev-libs/openssl:0=
-	elibc_musl? ( sys-libs/libunwind:= )
+	elibc_musl? (
+		llvm-libunwind? ( sys-libs/llvm-libunwind:= )
+		!llvm-libunwind? ( sys-libs/libunwind:= )
+	)
 	system-llvm? ( ${LLVM_DEPEND} )
 "
 
@@ -401,6 +404,7 @@ src_configure() {
 		dist-src = false
 		remap-debuginfo = true
 		lld = $(usex system-llvm false $(toml_usex wasm))
+		llvm-libunwind = "$(usex elibc_musl system no)"
 		# only deny warnings if doc+wasm are NOT requested, documenting stage0 wasm std fails without it
 		# https://github.com/rust-lang/rust/issues/74976
 		# https://github.com/rust-lang/rust/issues/76526


dev-lang/rust/rust-1.54.0.ebuild
```diff
@@ -303,6 +306,7 @@ src_configure() {
 		ninja = true
 		targets = "${LLVM_TARGETS// /;}"
 		experimental-targets = ""
+		use-libcxx = $(usex libcxx true false)
```
gyakovlev, line 309: there should be a dependency on libcxx, right?

dev-lang/rust/rust-1.54.0.ebuild
```diff
@@ -105,7 +105,10 @@ DEPEND="
 	net-misc/curl:=[http2,ssl]
 	sys-libs/zlib:=
 	dev-libs/openssl:0=
-	elibc_musl? ( sys-libs/libunwind:= )
+	elibc_musl? (
+		llvm-libunwind? ( sys-libs/llvm-libunwind:= )
```
gyakovlev, line 109: is it really musl-only thing? why is it in this block?

dev-lang/rust/rust-1.54.0.ebuild
ahmedcharles: Done.

dev-lang/rust/rust-1.54.0.ebuild
ahmedcharles: It makes sense to me but I'm not entirely sure. Every rust ebuild only depends on libunwind for musl.

The config.toml.example has this comment:

```
# Use LLVM libunwind as the implementation for Rust's unwinder.
# Accepted values are 'in-tree' (formerly true), 'system' or 'no' (formerly false).
# This option only applies for Linux and Fuchsia targets.
# On Linux target, if crt-static is not enabled, 'no' means dynamic link to
# `libgcc_s.so`, 'in-tree' means static link to the in-tree build of llvm libunwind
# and 'system' means dynamic link to `libunwind.so`. If crt-static is enabled,
# the behavior is depend on the libc. On musl target, 'no' and 'in-tree' both
# means static link to the in-tree build of llvm libunwind, and 'system' means
# static link to `libunwind.a` provided by system. Due to the limitation of glibc,
# it must link to `libgcc_eh.a` to get a working output, and this option have no effect.
```

Gentoo never seems to use crt-static, so systems with musl could link to libgcc_s.so and systems without musl could dynamically link to libunwind.so, whether via llvm or not. Though, if crt-static were used, things would get more complicated, since static musl and static glibc never overlap.

My goal was to get rust working with clang/libcxx/llvm-libunwind, which this change enables, though there are other possibilities.

Suggestions welcome.

gyakovlev:
> Is it possible to limit the libcxx use to only amd64, x86, arm and arm64?

yeah, you have to mask the useflag via `profiles/arch/base/package.use.mask` (that masks for all arches)
and unmask via `profiles/arch/amd64/package.use.mask`

ahmedcharles:
I lost track of whether I need to resolve anything else? Can you let me know, thanks? @gyakovlev

dev-lang/rust/rust-1.55.0.ebuild
```diff
@@ -41,7 +41,7 @@ LLVM_TARGET_USEDEPS=${ALL_LLVM_TARGETS[@]/%/(-)?}
 
 LICENSE="|| ( MIT Apache-2.0 ) BSD-1 BSD-2 BSD-4 UoI-NCSA"
 
-IUSE="clippy cpu_flags_x86_sse2 debug doc miri nightly parallel-compiler rls rustfmt system-bootstrap system-llvm test wasm ${ALL_LLVM_TARGETS[*]}"
+IUSE="clippy cpu_flags_x86_sse2 debug doc libcxx llvm-libunwind miri nightly parallel-compiler rls rustfmt system-bootstrap system-llvm test wasm ${ALL_LLVM_TARGETS[*]}"
```
Arfrever, line 44: This `USE=libcxx` here...

dev-lang/rust/rust-1.55.0.ebuild
```diff
@@ -105,7 +105,11 @@ DEPEND="
 	net-misc/curl:=[http2,ssl]
 	sys-libs/zlib:=
 	dev-libs/openssl:0=
-	elibc_musl? ( sys-libs/libunwind:= )
+	elibc_musl? (
+		llvm-libunwind? ( sys-libs/llvm-libunwind:= )
+		!llvm-libunwind? ( sys-libs/libunwind:= )
+	)
+	libcxx? ( sys-libs/libcxx )
```
Arfrever, line 112: ... and `USE=libcxx` here look unnecessary.
Current policies are to not usually specify explicitly dependencies on C / C++ standard library, and `sys-libs/libcxx` is one of such implementations of C++ standard library.
Users should select libc++ as C++ standard library by `-stdlib=libc++` compiler option or `USE=default-libcxx` for `sys-devel/clang`.

dev-lang/rust/rust-1.55.0.ebuild
```diff
@@ -302,6 +306,7 @@ src_configure() {
 		ninja = true
 		targets = "${LLVM_TARGETS// /;}"
 		experimental-targets = ""
+		use-libcxx = $(usex libcxx true false)
```
Arfrever, line 309: Before this `cat`, I suggest:
```
local libcxx=false
$(tc-getCXX) ${CXXFLAGS} ${LDFLAGS} $(no-as-needed) -o "${T}/test-c++-std-lib" -x c++ - <<<"int main() { return 0; }"
if type -p scanelf > /dev/null; then
	local lib
	while IFS= read -d , lib; do
		if [[ ${lib} =~ ^"libc++.so"(\..)?$ ]]; then
			libcxx=true
		fi
	done <<<"$(scanelf -qF "#F%n" "${T}/test-c++-std-lib")"
fi
```
And then here:
```
use-libcxx = ${libcxx}
```

gyakovlev:
I see you updated the pr.
clang-gang asked to merge this sooner so let's finally handle it.

to me it looks like llvm-libunwind will work on glibc too, isn't it the case?
so I'm still not sure why it's musl specific in this PR.


https://github.com/rust-lang/rust/blob/f90b06d7fb00470177516c5881bcec0432cb4f24/config.toml.example#L600-L610

```
# Use LLVM libunwind as the implementation for Rust's unwinder.
# Accepted values are 'in-tree' (formerly true), 'system' or 'no' (formerly false).
# This option only applies for Linux and Fuchsia targets.
# On Linux target, if crt-static is not enabled, 'no' means dynamic link to
# `libgcc_s.so`, 'in-tree' means static link to the in-tree build of llvm libunwind
# and 'system' means dynamic link to `libunwind.so`. If crt-static is enabled,
# the behavior is depend on the libc. On musl target, 'no' and 'in-tree' both
# means static link to the in-tree build of llvm libunwind, and 'system' means
# static link to `libunwind.a` provided by system. Due to the limitation of glibc,
# it must link to `libgcc_eh.a` to get a working output, and this option have no effect.
#llvm-libunwind = 'no'
```
in the ebuild we disable `crt-static` for musl (it's disabled for glibc by default) so the 1st case of this explanation is used.
so we don't hit libc-dependent scenario.

WDYT?

dev-lang/rust/rust-1.58.1.ebuild
```diff
@@ -377,6 +382,7 @@ src_configure() {
 		dist-src = false
 		remap-debuginfo = true
 		lld = $(usex system-llvm false $(toml_usex wasm))
+		llvm-libunwind = "$(usex elibc_musl system no)"
```
gyakovlev, line 385: does not seem to be correct flag to toggle?
I think it should be `llvm-libunwind = "$(usex llvm-libunwind system no)"

gyakovlev:
how does it work with system-llvm and bundled llvm btw?

dev-lang/rust/rust-1.58.1.ebuild
```diff
@@ -105,7 +105,11 @@ DEPEND="
 	net-misc/curl:=[http2,ssl]
 	sys-libs/zlib:=
 	dev-libs/openssl:0=
-	elibc_musl? ( sys-libs/libunwind:= )
+	elibc_musl? (
+		llvm-libunwind? ( sys-libs/llvm-libunwind:= )
+		!llvm-libunwind? ( sys-libs/libunwind:= )
+	)
```
gyakovlev, line 111:
```
llvm-libunwind? ( sys-libs/llvm-libunwind:= )
!llvm-libunwind? ( elibc_musl?  ( sys-libs/libunwind:= ) )
```
this is what you probably want. may be also complicated by system-llvm, not sure yet.

dev-lang/rust/rust-1.58.1.ebuild
gyakovlev: or maybe even
`llvm-libunwind = "$(usex llvm-libunwind $(usex system-llvm system in-tree) no)" `

and not pul it as dep if system-llvm is not enabled

gyakovlev:
btw how does it survive subslot changes in llvm and deps?
dependency probably needs to match llvm subslot it's built against?
can it be used iet's say with system-llvm-13.0.1 and `llvm-libunwind-14.0.0_rc1` ?

Arfrever:
While e.g. `sys-devel/llvm` is slotted, `sys-libs/libcxx`, `sys-libs/libcxxabi` and `sys-libs/llvm-libunwind` are not, and they even never had subslot changes.
ABI compatibility should be normally preserved during upgrades of `sys-libs/libcxx`, `sys-libs/libcxxabi` and `sys-libs/llvm-libunwind`.

dev-lang/rust/rust-1.58.1.ebuild
Arfrever:
```
$ grep llvm-libunwind profiles/use.desc 
llvm-libunwind - Use sys-libs/llvm-libunwind instead of sys-libs/libunwind
```
Please do not try to use this USE flag with different meaning.

If you want to have a USE flag for switching between using any libunwind library (`sys-libs/llvm-libunwind`, `sys-libs/libunwind`) or none, then `libunwind` USE flag can be used (it exists as local USE flag for few packages, e.g. `sys-libs/libcxxabi`, `dev-cpp/glog`).

If a USE flag would switch between system libunwind library or bundled libunwind library, then it could be called `system-libunwind`.

gyakovlev:
what I'm trying to imagine is a scenario:
user has `dev-lang/rust[system-llvm,llvm-libunwind]`
user installs llvm-14, while rust and llvm-13 it's linked to remains installed too.
llvm-libunwind gets updated to 14.
will it break rust?
there are known cases of rust being unable to update itself with [system-llvm,system-boostrap] if llvm abi changes, for example if user changed LLVM_TARGETS.

mozilla is known to break if built with llvm13 and rust with llvm12 (as example) and mismatched lld version.
I suspect unwind can complicate things too, but not sure.

dev-lang/rust/rust-1.58.1.ebuild
gyakovlev: system-llvm logic already covers it. don't see a problem with it or need to introduce yet another flag.
please stop offering unsolicited advice, not helping.

ahmedcharles:
> to me it looks like llvm-libunwind will work on glibc too, isn't it the case? so I'm still not sure why it's musl specific in this PR.
> 
> WDYT?

The reasons it's musl specific are:

1) I didn't want to change the behavior on non-musl systems, just to get this combination of options that, apparently, no one uses to work.
2) If I change the behavior to always use llvm-libunwind, then every gentoo system using the defaults will start building llvm-libunwind and therefore, have two unwind libraries on their system.

> how does it work with system-llvm and bundled llvm btw?

I didn't test this scenario. (I suppose I can.)

> can it be used iet's say with system-llvm-13.0.1 and `llvm-libunwind-14.0.0_rc1` ?

Given that llvm-libunwind is a drop in replacement for both libgcc_s and libunwind, it necessarily can't have abi changes across versions without failing that premise.

dev-lang/rust/rust-1.58.1.ebuild
ahmedcharles: The `llvm-libunwind = ` option in the config doesn't mean the same thing as the `llvm-libunwind` gentoo use flag. The `no` option on non-musl systems preserves the default for glibc systems (I really don't want to change what they use) and the `system` option dynamically links to either `llvm-libunwind` or `libunwind` based on whichever is installed in the system (which is what the use flag controls). The in-tree version was never used on gentoo before and it is still never used with this change. That would be a completely different PR/change.

dev-lang/rust/rust-1.58.1.ebuild
ahmedcharles: This would change behavior on glibc systems, which is something I'd rather not do.

gyakovlev:
> > to me it looks like llvm-libunwind will work on glibc too, isn't it the case? so I'm still not sure why it's musl specific in this PR.
> > WDYT?
> 
> The reasons it's musl specific are:
> 
>     1. I didn't want to change the behavior on non-musl systems, just to get this combination of options that, apparently, no one uses to work.
> 
>     2. If I change the behavior to always use llvm-libunwind, then every gentoo system using the defaults will start building llvm-libunwind and therefore, have two unwind libraries on their system.
> 
> 
> > how does it work with system-llvm and bundled llvm btw?
> 
> I didn't test this scenario. (I suppose I can.)
> 
> > can it be used iet's say with system-llvm-13.0.1 and `llvm-libunwind-14.0.0_rc1` ?
> 
> Given that llvm-libunwind is a drop in replacement for both libgcc_s and libunwind, it necessarily can't have abi changes across versions without failing that premise.

this sounds broken.
so with current state of PR:
* glibc users get a new flag that does exactly nothing but still can be flipped. noop flag is no good. and ` llvm-libunwind` is a valid usecase for glibc too.
* musl users unconditionally get switched to llvm-libundind option, which can link to any library, which is controlled by separate dependency and flag.
* and I haven't looked at `libcxx` yet, I believe it should be separate PR after we deal with libunwind.

I suggest doing the following:

1. introduce ` llvm-libunwind` flag that toggles usage of it regardless of libc. we can mask or stable-mask it for starters if you want to limit possible breakage. I may be able to test that if you don't want to deal with glibc.
2. make sure ` llvm-libunwind` flag uses system version of ` llvm-libunwind` if `system-llvm` is set, and bundled otherwise, on both libcs. basically overload the flag function and make it do both things. that way we get default upstream behavior if flag is untouched, and custom behavior that matches our flag description if flag is set.

an option named ` llvm-libunwind` that has a description `# Use LLVM libunwind as the implementation for Rust's unwinder.`  which can work with non-llvm `libunwind` is quite surprising.

thoughts?

Arfrever:
> so with current state of PR:
>     * musl users unconditionally get switched to llvm-libundind option, which can link to any library, which is controlled by separate dependency and flag.

This looks correct to me, if it works...


@ahmedcharles
To better illustrate situation, could you show what error happens (on musl system) without changes from this pull-request?
(Possibly file bug in Gentoo bugzilla and attach full build log.)

ahmedcharles:
You get something similar to:

```

Calculating dependencies... done!
[ebuild  N     ] sys-libs/libunwind-1.5.0-r1:0/8::gentoo  USE="zlib -debug -debug-frame -doc -libatomic -lzma -static-libs" 858 KiB
[ebuild   R    ] dev-lang/rust-1.58.1:stable/1.58::musl [1.58.1:stable/1.58::local] USE="clippy rust-src rustfmt -debug -dist -doc -miri -nightly (-parallel-compiler) -rls -system-bootstrap -system-llvm -test -verify-sig -wasm (-libcxx%*) (-llvm-libunwind%*)" CPU_FLAGS_X86="sse2" LLVM_TARGETS="(X86) -AArch64 -AMDGPU -ARM -AVR -BPF -Hexagon -Lanai -MSP430 -Mips -NVPTX -PowerPC -RISCV -Sparc -SystemZ -WebAssembly -XCore" 0 KiB
[blocks B      ] sys-libs/libunwind ("sys-libs/libunwind" is soft blocking sys-libs/llvm-libunwind-13.0.0)

Total: 2 packages (1 new, 1 reinstall), Size of downloads: 858 KiB
Conflict: 1 block (1 unsatisfied)

!!! Multiple package instances within a single package slot have been pulled
!!! into the dependency graph, resulting in a slot conflict:

dev-lang/rust:stable

  (dev-lang/rust-1.58.1:stable/1.58::musl, ebuild scheduled for merge) USE="clippy rust-src rustfmt -debug -dist -doc -miri -nightly (-parallel-compiler) -rls -system-bootstrap -system-llvm -test -verify-sig -wasm" CPU_FLAGS_X86="sse2" LLVM_TARGETS="(X86) -AArch64 -AMDGPU -ARM -AVR -BPF -Hexagon -Lanai -MSP430 -Mips -NVPTX -PowerPC -RISCV -Sparc -SystemZ -WebAssembly -XCore" pulled in by
    dev-lang/rust (Argument)

  (dev-lang/rust-1.58.1:stable/1.58::local, installed) USE="clippy libcxx llvm-libunwind rust-src rustfmt -debug -dist -doc -miri -nightly (-parallel-compiler) -rls -system-bootstrap -system-llvm -test -verify-sig -wasm" CPU_FLAGS_X86="sse2" LLVM_TARGETS="(X86) -AArch64 -AMDGPU -ARM -AVR -BPF -Hexagon -Lanai -MSP430 -Mips -NVPTX -PowerPC -RISCV -Sparc -SystemZ -WebAssembly -XCore" pulled in by
    (no parents that aren't satisfied by other packages in this slot)


!!! Enabling --newuse and --update might solve this conflict.
!!! If not, it might help emerge to give a more specific suggestion.


 * Error: The above package list contains packages which cannot be
 * installed at the same time on the same system.

  (sys-libs/libunwind-1.5.0-r1:0/8::gentoo, ebuild scheduled for merge) pulled in by
    >=sys-libs/libunwind-1.0.1-r1[static-libs?,abi_x86_32(-)?,abi_x86_64(-)?,abi_x86_x32(-)?,abi_mips_n32(-)?,abi_mips_n64(-)?,abi_mips_o32(-)?,abi_s390_32(-)?,abi_s390_64(-)?] (>=sys-libs/libunwind-1.0.1-r1[abi_x86_64(-)]) required by (sys-libs/libcxxabi-13.0.0:0/0::gentoo, installed) USE="libunwind -static-libs -test"
    sys-libs/libunwind:= required by (dev-lang/rust-1.58.1:stable/1.58::musl, ebuild scheduled for merge) USE="clippy rust-src rustfmt -debug -dist -doc -miri -nightly (-parallel-compiler) -rls -system-bootstrap -system-llvm -test -verify-sig -wasm" CPU_FLAGS_X86="sse2" LLVM_TARGETS="(X86) -AArch64 -AMDGPU -ARM -AVR -BPF -Hexagon -Lanai -MSP430 -Mips -NVPTX -PowerPC -RISCV -Sparc -SystemZ -WebAssembly -XCore"

  (sys-libs/llvm-libunwind-13.0.0:0/0::gentoo, installed) pulled in by
    sys-libs/llvm-libunwind required by (sys-devel/clang-13.0.0:13/13::gentoo, installed) USE="default-compiler-rt default-libcxx default-lld llvm-libunwind static-analyzer -debug -doc -test -xml" LLVM_TARGETS="BPF (X86) -AArch64 -AMDGPU (-ARC) -ARM -AVR (-CSKY) -Hexagon -Lanai (-M68k) -MSP430 -Mips -NVPTX -PowerPC -RISCV -Sparc -SystemZ (-VE) -WebAssembly -XCore" PYTHON_SINGLE_TARGET="python3_9 -python3_10 -python3_8"
    sys-libs/llvm-libunwind:0/0= required by (dev-lang/rust-1.58.1:stable/1.58::local, installed) USE="clippy libcxx llvm-libunwind rust-src rustfmt -debug -dist -doc -miri -nightly (-parallel-compiler) -rls -system-bootstrap -system-llvm -test -verify-sig -wasm" CPU_FLAGS_X86="sse2" LLVM_TARGETS="(X86) -AArch64 -AMDGPU -ARM -AVR -BPF -Hexagon -Lanai -MSP430
 -Mips -NVPTX -PowerPC -RISCV -Sparc -SystemZ -WebAssembly -XCore"


For more information about Blocked Packages, please refer to the following
section of the Gentoo Linux x86 Handbook (architecture is irrelevant):

https://wiki.gentoo.org/wiki/Handbook:X86/Working/Portage#Blocked_packages
```

I'm not entirely sure how to use/access Gentoo's bugzilla, so it's easier to just post it here for now.

As far as what I think, I just wanted to fix this specific issue so that I could install rust using my specific configuration. Gentoo's use flag system is complicated and I'm not sure I understand it well enough to go through the process of resolving all of these questions.

gyakovlev:
my version builds fine on glibc system
```shell-session
lddtree /usr/bin/rustc
rustc => /usr/bin/rustc (interpreter => /lib64/ld-linux-x86-64.so.2)
    librustc_driver-f72aad5e707f7f15.so => /usr/lib/rust/lib/librustc_driver-f72aad5e707f7f15.so
        libLLVM-14libcxx.so => /usr/lib/llvm/14/lib64/libLLVM-14libcxx.so
            libffi.so.8 => /usr/lib64/libffi.so.8
            libz.so.1 => /lib64/libz.so.1
            libtinfo.so.6 => /lib64/libtinfo.so.6
        libc++.so.1 => /usr/lib64/libc++.so.1
        libc++abi.so.1 => /usr/lib64/libc++abi.so.1
        libunwind.so.1 => /usr/lib64/libunwind.so.1
...
```

currently testing it in different scenarios

gyakovlev:
alternatively, can you just test my version on your system?
it's in ::gentoo now, 1.63.0-r1

ahmedcharles:
Sure, I'll take a look when I have some time. Thanks.

gyakovlev:
we actually managed to bootstrap it with upstream tarball and some MAGIC =D
so it just works out of the box!

https://github.com/gentoo/gentoo/commit/941cc3c5faae01bff0045d92e62da0a9689a0358

I believe this PR can be closed now, thanks for initial work, it's now fixed and is bootstrap-able on gcc-less system without any fiddling or custom tarballs.

PR #22224, opened 2021-09-05
sys-fs/zfs-kmod: set CROSS_COMPILE in the environment

floppym:
This allows it to be overriden in a local Makefile.

Closes: https://bugs.gentoo.org/811600
Signed-off-by: Mike Gilbert <floppym@gentoo.org>


gyakovlev (approved):
lgtm

PR #22228, opened 2021-09-06
games-engines/scummvm: fix build, deps, bump eapi

DarthGandalf:
Closes: https://bugs.gentoo.org/744421

### DIFF
--- a/games-engines/scummvm-tools/files/scummvm-tools-2.2.0-strings.patch
+++ b/games-engines/scummvm-tools/files/scummvm-tools-2.2.0-strings.patch
@@ -1,0 +1,17 @@
+Fix build without native symlinks
+https://bugs.gentoo.org/744424
+
+--- a/configure
++++ b/configure
+@@ -1026,9 +1026,9 @@ void _ebcdic() { char* s = (char*) ebcdi
+ int main() { _ascii (); _ebcdic (); return 0; }
+ EOF
+ $CXX $CXXFLAGS -c -o $TMPO.o tmp_endianness_check.cpp
+-if strings $TMPO.o | grep BIGenDianSyS >/dev/null; then
++if ${STRINGS} $TMPO.o | grep BIGenDianSyS >/dev/null; then
+ 	_endian=big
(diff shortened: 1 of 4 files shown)


ionenwks:
Looks fine, but mind to do a bit more work while here?
- If revbumping, it'd make sense to fix https://bugs.gentoo.org/801229 too
- there's a very similar bug at https://bugs.gentoo.org/744424
- since there's a -9999, putting this together with a "sync live" commit be nice.

DarthGandalf:
Done, except for -tools one, which is a different package

games-engines/scummvm-tools/scummvm-tools-2.2.0-r1.ebuild
```diff
@@ -0,0 +1,71 @@
...
+		$(use_enable iconv) \
+		$(use_enable iconv freetype2) \
+		$(use_enable mad) \
+		$(use_enable png) \
+		$(use_enable vorbis) || die
+}
+
+src_compile() {
+	emake
+}
```
ionenwks, line 67: Can drop this.

Also, not seeing a reason to revbump, this only fixed a build failure. EAPI bump may sometime need one to avoid changing stable version or so, but this doesn't apply here. I'd imagine you may have heard the opposite at times though.

games-engines/scummvm-tools/scummvm-tools-2.2.0-r1.ebuild
```diff
@@ -0,0 +1,71 @@
...
+SRC_URI="http://scummvm.org/frs/scummvm-tools/${PV}/${P}.tar.xz"
+
+LICENSE="GPL-2+"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc ~ppc64 ~x86"
+IUSE="flac iconv mad png vorbis"
+RESTRICT="test" # some tests require external files
+
+RDEPEND="
+	>=dev-libs/boost-1.32:=
```
ionenwks, line 20: Can side-drop the version range, boost.1.32 been gone from the tree since 2005. Ranges are good to keep in general but at some point it gets silly.

PR #22230, opened 2021-09-06
mail-filter/rspamd: backport fixes to 3.0 and bump live

arkamar:
- backport of null pointer dereference patch to fix some segfaults, I decided to do it in separate commit because I didn't want to mix it with `doctest` fix,
- backport patch which allows to use system `doctest`,
- new ebuild for `dev-libs/doctest`, I will take care of it, I referenced related bug [809752](https://bugs.gentoo.org/show_bug.cgi?id=809752) as well,
- sync live ebuild.


mail-filter/rspamd/rspamd-3.0-r3.ebuild
```diff
@@ -35,6 +35,7 @@ RDEPEND="${LUA_DEPS}
 	acct-user/rspamd
 	app-arch/zstd:=
 	dev-db/sqlite:3
+	dev-libs/doctest:=
```
thesamesam, line 38: You're not exposing a subslot in doctest?

mail-filter/rspamd/files/rspamd-3.0-system-doctest.patch
```diff
@@ -0,0 +1,72 @@
+From d37a3764e9889e6de0a7341ba3195ee8a314de23 Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: I've said it before but I'd like to just say again that you're doing great work, and it is really appreciated. Thanks.

dev-libs/doctest/doctest-2.4.6.ebuild
```diff
@@ -0,0 +1,27 @@
...
+HOMEPAGE="https://github.com/onqtam/doctest"
+SRC_URI="https://github.com/onqtam/doctest/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="test"
+RESTRICT="!test? ( test )"
+
+DEPEND=""
```
thesamesam, line 18: We can [drop](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-620007.3) all 3 of these lines.

dev-libs/doctest/doctest-2.4.6.ebuild
```diff
@@ -0,0 +1,27 @@
...
+RDEPEND="${DEPEND}"
+BDEPEND=""
+
+src_configure() {
+	local mycmakeargs=(
+		-DDOCTEST_WITH_TESTS=$(usex test ON OFF)
```
thesamesam, line 24: Should work:
```suggestion
		-DDOCTEST_WITH_TESTS=$(usex test)
```

mail-filter/rspamd/files/rspamd-3.0-system-doctest.patch
arkamar: Thanks :)

mail-filter/rspamd/rspamd-3.0-r3.ebuild
arkamar: Yeah, the thing with this is that I wasn't sure how I should expose the subslot (how many version numbers), but I forgot to mention it in PR. It is a single-header C++ library with a lot of templates, therefore, I decided `0/${PV}` should be there, similar to boost.

dev-libs/doctest/doctest-2.4.6.ebuild
arkamar: thanks

dev-libs/doctest/doctest-2.4.6.ebuild
arkamar: Unbelievable how many things could be improved in such a small ebuild :D Thanks

arkamar:
FP with review fixes in `dev-libs/doctest`.

mail-filter/rspamd/rspamd-3.0-r3.ebuild
thesamesam: If it's header-only, we should be fine _not_ to (the problem with Boost is that Boost has some parts which must be linked against, and figuring out if you do/don't need them is tricky).

I'll drop it for now, I think.

thesamesam: Actually, this makes me think about putting it in dev-cpp.

arkamar: Ok, thanks! Anyway, my idea behind slotting was that in case of some bug it is necessary to rebuild packages depending on `doctest`. But we can introduce it if it really will be needed in future.

PR #22234, opened 2021-09-06
media-plugins/vdr-imonlcd: version bump to 1.0.3

madmartin:


media-plugins/vdr-imonlcd/vdr-imonlcd-1.0.3.ebuild
```diff
@@ -0,0 +1,51 @@
...
+	vdr-plugin-2_pkg_setup
+}
+
+src_configure() {
+	tc-export PKG_CONFIG
+	default
+}
+
+src_install() {
+	rm -f README.git || die
```
juippis, line 43: Does this exist or not? You should be able to predict it. So drop `-f`, rest LGTM.

madmartin: done

juippis (approved):

PR #22236, opened 2021-09-06
x11-misc/qt5ct bump 1.3-r1 to fix USE=-dbus to actually work

mmogilvi:
Changes developed by Nuno Silva for 1.2 (see bug);
tweaked for 1.3 by mmogilvi.

Closes: https://bugs.gentoo.org/802828
Signed-off-by: Matthew Ogilvie <mmogilvi+gnto@zoho.com>

### DIFF
--- a/x11-misc/qt5ct/files/qt5ct-1.3-cmake-dbus.patch
+++ b/x11-misc/qt5ct/files/qt5ct-1.3-cmake-dbus.patch
@@ -1,0 +1,36 @@
+diff -ur a/CMakeLists.txt b/CMakeLists.txt
+--- a/CMakeLists.txt	2021-04-20 20:29:54.000000000 +0100
++++ b/CMakeLists.txt	2021-08-28 11:20:28.130637732 +0100
+@@ -21,10 +21,17 @@
+ set(CMAKE_AUTOUIC ON)
+ set(QT_MINIMUM_VERSION "5.12.0")
+ 
++option(DISABLE_DBUS "disable D-Bus support (global menu and D-Bus tray icon implementation)" OFF)
+ 
+ ADD_DEFINITIONS(-DQT_DISABLE_DEPRECATED_BEFORE=0x050C00 -DUSE_WIDGETS)
+ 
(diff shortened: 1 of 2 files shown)


x11-misc/qt5ct/files/qt5ct-1.3-cmake-dbus.patch
```diff
@@ -0,0 +1,36 @@
+diff -ur a/CMakeLists.txt b/CMakeLists.txt
```
mgorny, line 1: Has this been submitted upstream?

mmogilvi: I haven't submitted it myself.  Not sure about the patch's author (Nuna Silva), although I don't seen any obvious sign of it under the upstream home page.

I don't know cmake at all.  Is there a cleaner way to code for optional dependencies, perhaps involving some kind of additional variable and/or something more like a C '?:' operator on a single line?

mmogilvi:
> here is patch to `x11-misc/qt5ct/qt5ct-1.3.ebuild` without patching any sources

I'm not sure why the 1.1 -> 1.2 update switched to cmake, but I assumed there must be some kind of reason.  Otherwise, switching back to qmake makes sense to me.

mmogilvi:
@dm9pZCAq

I can't speak for the gentoo maintainer, but I would guess he would be more amenable to re-introducing the dbus USE flag if someone pushed through the necessary changes to the cmake infrastructure to make the dbus dependency optional, got it merged $upstream, and into a new release.  Perhaps something similar to Nuno Silva's patch like what is included as part of this pull request, except pushed upstream and officially released.  (Although I don't know if there is a cleaner way to structure the patch?  I don't know cmake.)

As for myself, I decided to completely uninstall qt5ct from the last machine that had it, today, since it was now blocking a system update.  (I barely used it at all, and I had been excluding it from updates ever since it started implicitly (and then later explicitly) requiring qtdbus without a way to disable it.  But now it requires more work than it is worth to me.)

dm9pZCAq:
@mmogilvi, you can still use it from [my overlay](https://github.com/gentoo-mirror/dm9pZCAq), here i maintain qt5ct without `dbus`

PR #22239, opened 2021-09-07
www-client/qutebrowser: version bump to 2.3.1

SineLaude:
Closes: https://bugs.gentoo.org/808648
Signed-off-by: Christian Schroeder <chris.schroeder.public@gmail.com>


crocket:
I'd like to see it merged soon.

ionenwks (approved):
Thanks, seems to work fine indeed.

I'll do some minor cleanups of my own at same time and merge this soon.

I'm curious to know if py3.10 works fine and pass tests (deps-wise, only need it to add 10 to adblock) but that should probably be left for whenever 2.4.0 releases, so I'll keep it at 9 for now.

PR #22241, opened 2021-09-08
app-crypt/tpm2-tools: Fix automagic depedency on libeifvar

salahcoronya:
This fixes some other problem as well, like eautoreconf being called before patching instead of the other way around.


app-crypt/tpm2-tools/tpm2-tools-5.1.1-r2.ebuild
```diff
@@ -15,9 +15,10 @@ SLOT="0"
 KEYWORDS="~amd64"
 IUSE="+fapi test"
 
-RDEPEND="net-misc/curl:=
-	>=app-crypt/tpm2-tss-3.0.1:=[fapi?]
+RDEPEND=">=app-crypt/tpm2-tss-3.0.1:=[fapi?]
 	dev-libs/openssl:=
+	net-misc/curl:=
+	sys-libs/efivar
```
thesamesam, line 21: It exposes a subslot and we link against it so I'll add :=.

thesamesam:
Thanks!

app-crypt/tpm2-tools/files/tpm2-tools-5.1.1-no-efivar-automagic.patch
```diff
@@ -0,0 +1,32 @@
+From 0cea7f0f78f1a9e8dca789eb5f2ece052e026bed Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: 1. Upstreamable? If so, chuck in a link.
2. Always good to chuck in a link to the bug anyway.

PR #22245, opened 2021-09-08
dev-python/mkdocs* : enable py3.10

arthurzam:
Quite a big PR for bumping all `dev-python/mkdocs*` packages to python 3.10

- All packages that have test already setup, was tested and tests pass.
- All packages without test support, upstream (github) was checked to see if a test suite appears there - only `dev-python/mkdocs-redirects` applied and added tests from github archive (sadly using a commit tag) - so please note id the change is acceptable?
- All other packages, were tested by using them as doc generation for other packages - most of those were here.
- Lastly, `dev-python/mkdocs-bootswatch` and `dev-python/mkdocs-bootstrap` were left - so tested by reviewing all python code upstream and seeing no problems.

As most of those packages are of @AndrewAmmerlaan , waiting for check, review if needed, and possible ACK. I can merge without problems :)


AndrewAmmerlaan (approved):
Very nice, Thank You! Feel free to merge this.

PR #22248, opened 2021-09-08
dev-python/elasticsearch-py: version bump to 7.14.1

parona-source:
Also includes an addition of py10 target to sci-libs/shapely and a new ebuild dev-python/unasync. Unasync is an optional dependency for dev-python/elasticsearch-py.

All ebuilds pass tests on python 3.9 and python 3.10.

Not quite sure on the maintainer assignment on dev-python/unasync so just put the python project and maintainer needed in its metadata.xml. Although id fine with taking it as a proxy maintainer if required.

The new version of dev-python/elasticsearch-py has  x86 keyword dropped due to me only having a amd64 system at my disposal, so cant vouch for x86.


parona-source:
Amended oversight in test dependencies.

parona-source:
The sci-libs/shapely python target bump currently doesn't make sense in this pull request due to me to forgetting it in. The reason for it being initially here was due me  trying add an another python ebuild which required it before realizing it wasn't necessary in that version of elasticsearch-py (it will be a dependency for some api tests in the next version).

dev-python/unasync/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>python@gentoo.org</email>
+	</maintainer>
+	<!-- maintainer-needed -->
```
hydrapolic, line 7: Probably better to replace the python project with you + proxy maint (if not consulted otherwise with the python team).

parona-source: Added myself as proxy maint and removed the python project as I haven't consulted anyone in the python team.

ferki:
@parona-source: thank you for the version bump!

(Mostly) due to not having an active use case for the Elastic stack on Gentoo anymore, I'm considering to step down as proxied maintainer. I'd be glad to have someone stepping up for dev-python/elasticsearch-py who is in a better position to keep maintaining it for Gentoo users. If you'd be willing to do that, please consider to add yourself to the `metadata.xml` for that package as well.

parona-source:
> @parona-source: thank you for the version bump!
> 
> (Mostly) due to not having an active use case for the Elastic stack on Gentoo anymore, I'm considering to step down as proxied maintainer. I'd be glad to have someone stepping up for dev-python/elasticsearch-py who is in a better position to keep maintaining it for Gentoo users. If you'd be willing to do that, please consider to add yourself to the `metadata.xml` for that package as well.

Added myself to metadata.xml.

ferki (changes requested):
> Also includes an addition of py10 target to sci-libs/shapely

Hmm, I don't see why that's in scope for a this dev-python/elasticsearch-py version bump. Is it in the dependency tree somehow? If it's unrelated, I'd expect Gentoo devs would prefer that change to be in a separate PR (fixme).

> The new version of dev-python/elasticsearch-py has x86 keyword dropped due to me only having a amd64 system at my disposal, so cant vouch for x86.

IIRC, I was previously told that it's fine to contain both `~amd64` and `~x86`, so I do that in my proxied packager ever since. Can't find a reference for that quickly though, so it would be nice to be confirmed by an actual Gentoo dev with commit rights :)

----

I've tried to test the elasticsearch-py ebuild with

```
# USE=test ebuild elasticsearch-py-7.14.1.ebuild test
```

but that failed with

```
/var/tmp/portage/dev-python/elasticsearch-py-7.14.1/temp/environment: line 2919: java-config: command not found
```

I didn't try to fix that, but perhaps including a java environment among the test dependencies could solve that, right?

juippis:
Could you please rebase this on top of fresh master, to account changes already done to shapely and elasticsearch-py in the ::gentoo repo? Let me know if you need help with it. 

Sorry it's taken a while, got my eyes here now.

parona-source:
@juippis Rebased. I havent figured out why it failed for @ferki though, as I to my knowledge virtual/jre should pull in dev-java/java-config with any of the jre and jdk implementations.

dev-python/elasticsearch-py/elasticsearch-py-7.14.1.ebuild
```diff
@@ -0,0 +1,185 @@
...
+
+src_prepare() {
+	default_src_prepare
+
+	if use test; then
+		local es_port="25124"
+
+		export ES_DIR="${WORKDIR}/elasticsearch-${PV}"
+		export ES_INSTANCE="gentoo-es-py-test"
+		export ES_JAVA_HOME=$(java-config -g JAVA_HOME || die)
```
juippis, line 62: This gives me the wrong directory, 
`/opt/openjdk-bin-8.302_p08/bin' which is empty. The right one is `/opt/openjdk-bin-8/bin`. Why this happens though, and why openjdk-bin installs empty directories, no idea. Sounds like opendjdk-bin bug. 11.0.12_p7 does not have this problem.

```
Future versions of Elasticsearch will require Java 11; your Java version from [/opt/openjdk-bin-8/jre]    does not meet this requirement. Consider switching to a distribution of Elasticsearch with a bundled JDK. If you are already using a distribution with a bundled JDK, ensure the JAVA_HOME environment variable is  not set.
```
So maybe we should restrict it to >=11

dev-python/elasticsearch-py/elasticsearch-py-7.14.1.ebuild
```diff
@@ -0,0 +1,185 @@
...
+DEPEND="
+	test? (
+		${RDEPEND}
+		>=dev-python/requests-2.0.0[${PYTHON_USEDEP}]
+		<dev-python/requests-3.0.0[${PYTHON_USEDEP}]
+		dev-python/mock[${PYTHON_USEDEP}]
+		dev-python/numpy[${PYTHON_USEDEP}]
+		dev-python/pandas[${PYTHON_USEDEP}]
+		dev-python/python-dateutil[${PYTHON_USEDEP}]
+		>=dev-python/pyyaml-5.4[${PYTHON_USEDEP}]
```
juippis, line 43: Needed to also add aiohttp.

juippis (approved):
Thanks a lot!

ferki:
> I havent figured out why it failed for @ferki though, as I to my knowledge virtual/jre should pull in dev-java/java-config with any of the jre and jdk implementations.

@parona-source : ah, yes, now I see it was my bad to not make sure the dependencies are installed. Thanks for taking over the maintenance!

PR #22250, opened 2021-09-09
app-admin/lnav: 0.10.0-r1 version bump

brianmego:
Closes: https://bugs.gentoo.org/812191


app-admin/lnav/lnav-0.10.0-r1.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 1999-2021 Gentoo Authors
```
bowlofeggs, line 1: I recommend renaming this file to `lnav-0.10.0.ebuild` since this is the first edition of the ebuild.

bowlofeggs (changes requested):
Some of the tests seem to fail:

```
make  check-TESTS                                                                                                      
make[2]: Entering directory '/var/tmp/portage/app-admin/lnav-0.10.0-r1/work/lnav-0.10.0/test'                          
make[3]: Entering directory '/var/tmp/portage/app-admin/lnav-0.10.0-r1/work/lnav-0.10.0/test'                          
PASS: test_abbrev                                                                                                      
PASS: test_auto_fd                                         
PASS: test_auto_mem                                        
PASS: test_date_time_scanner                               
PASS: test_ansi_scrubber                                   
PASS: test_bookmarks                                       
PASS: test_format_installer.sh
PASS: lnav_doctests                                        
FAIL: test_mvwattrline.sh                                  
FAIL: test_listview.sh                                     
PASS: test_grep_proc2                                      
PASS: test_cli.sh                                          
PASS: test_log_accel                                       
PASS: test_format_loader.sh                                
PASS: test_grep_proc.sh                                    
PASS: test_reltime                                         
PASS: test_scripts.sh                                      
PASS: test_shlexer.sh                                      
PASS: test_config.sh                                       
PASS: test_meta.sh                                         
PASS: test_sessions.sh                                     
PASS: test_sql_anno.sh                                     
PASS: test_sql_coll_func.sh                                
FAIL: test_remote.sh                                       
PASS: test_json_format.sh                                  
PASS: test_sql_xml_func.sh                                 
PASS: test_sql_fs_func.sh                                  
PASS: test_sql_time_func.sh                                
PASS: test_sql_json_func.sh                                
PASS: test_view_colors.sh                                  
FAIL: test_vt52_curses.sh                                  
PASS: test_tui.sh                                          
FAIL: test_sql.sh                                          
PASS: test_curl.sh                                         
PASS: test_pretty_print.sh                                 
PASS: test_sql_str_func.sh                                 
PASS: test_data_parser.sh                                  
FAIL: test_logfile.sh                                      
PASS: test_cmds.sh
Testsuite summary for lnav 0.10.0
# TOTAL: 39                                                
# PASS:  33                                                
# SKIP:  0                                                 
# XFAIL: 0                                                 
# FAIL:  6                                                 
# XPASS: 0                                                 
# ERROR: 0                                                 
See test/test-suite.log                                    
Please report to lnav@googlegroups.com

**Comment by bowlofeggs** [comment]
I wonder if some of the dropped patches might still be necessary. For example, `lnav-0.8.4-disable-tests.patch` looks like this, which seems like it might account for two of these failures:

```
--- lnav-0.8.4/test/Makefile.am
+++ lnav-0.8.4/test/Makefile.am
@@ -373,7 +373,6 @@
    test_concise \
    test_line_buffer2 \
    test_line_buffer.sh \
-   test_listview.sh \
    test_meta.sh \
    test_grep_proc.sh \
    test_grep_proc2 \
@@ -381,7 +380,6 @@
    test_json_op.sh \
    test_json_ptr_walk.sh \
    test_log_accel \
-   test_logfile.sh \
    test_pcrepp \
    test_reltime \
    test_scripts.sh \
```

**Comment by brianmego** [comment]
Hey, @bowlofeggs . Thanks for the review. The removed patch had diverged enough that it no longer matched. Happy to make another one that matches and does the same operation, though what did you run to get the tests to fail? The build worked ok for me, so maybe I'm missing an operation in my test cycle. I have only done a handful of these before.

**Comment by thesamesam** [comment]
(Did you run "ebuild .. clean test install merge" -- key bit being 'test'? It won't run the test suite by default, which is handy sometimes, and not others...)

**Comment by brianmego** [comment]
Nope! Totally missed test. Thanks. Will take another pass tomorrow.

**Comment by brianmego** [comment]
@bowlofeggs While I still get test failures, they are different depending on how I run this thing. I can certainly comment out the ones causing issues, but that feels like it'll just mask a deeper problem. What I've noticed:

1. If I run `ebuild lnav-0.10.0.ebuild clean prepare configure compile test` the test suite hangs
2. If I run this as root (prefixed with sudo) then I get 4 failures:
    -  test_listview.sh
    - test_remote.sh
    - test_sql.sh
    - test_logfile.sh
3. If I run `ebuild lnav-0.10.0.ebuild clean prepare configure compile` and then manually cd into the build directory and run `sudo make check` (which is a good proxy for ebuild test as far as I can tell) I get 3 failures:
    - test_remote.sh
    - test_sql.sh
    - test_logfile.sh

So the interesting points here summarized:
1. With ebuild, I get 4 errors not 6 as you do. That could be due to the "tests that have been fixed in the top of the tree" as  the author pointed out here: https://github.com/tstack/lnav/issues/911#issuecomment-916574933.
2. I get 3 errors when running at the location of the code, so something is different about how ebuild is doing it than the manual process I'm running. Maybe pathing?
3. I cannot get the test suite to complete if running as my unprivileged user. Is this something you can reproduce? Or something about my setup?
4. Is the right thing to do here indeed comment out the affected tests with a patch and log my findings with the original project?

**Comment by bowlofeggs** [comment]
> @bowlofeggs While I still get test failures, they are different depending on how I run this thing. I can certainly comment out the ones causing issues, but that feels like it'll just mask a deeper problem. What I've noticed:
> 
>     1. If I run `ebuild lnav-0.10.0.ebuild clean prepare configure compile test` the test suite hangs

I believe I observed this too, but iirc it does eventually finish after some time (maybe 10-20 minutes?).

>     2. If I run this as root (prefixed with sudo) then I get 4 failures:
>        
>        * test_listview.sh
>        * test_remote.sh

This one seems to assume the test will be able to ssh to localhost, which I don't think would or should be allowed during an ebuild. Disabling it makes sense to me.

>        * test_sql.sh
>        * test_logfile.sh
> 
>     3. If I run `ebuild lnav-0.10.0.ebuild clean prepare configure compile` and then manually cd into the build directory and run `sudo make check` (which is a good proxy for ebuild test as far as I can tell) I get 3 failures:
>        
>        * test_remote.sh
>        * test_sql.sh
>        * test_logfile.sh
> 
> 
> So the interesting points here summarized:
> 
>     1. With ebuild, I get 4 errors not 6 as you do. That could be due to the "tests that have been fixed in the top of the tree" as  the author pointed out here: [6 tests fail with 0.10.0 on Gentoo tstack/lnav#911 (comment)](https://github.com/tstack/lnav/issues/911#issuecomment-916574933).
> 
>     2. I get 3 errors when running at the location of the code, so something is different about how ebuild is doing it than the manual process I'm running. Maybe pathing?
> 
>     3. I cannot get the test suite to complete if running as my unprivileged user. Is this something you can reproduce? Or something about my setup?

As said above - try giving it longer to complete. If it takes more than 20 minutes then perhaps something is wrong.

>     4. Is the right thing to do here indeed comment out the affected tests with a patch and log my findings with the original project?

I'd say the right thing to do is to go through each one case by case and work with upstream to determine what to do. I do think the one that tries to ssh to localhost should just be disabled via a patch since that doesn't seem likely to be allowed in an ebuild.

**Comment by brianmego** [comment]
I'm still working with underlying repo to determine what's different about the gentoo build vs the build straight from the repo's source, but it probably makes sense to review this initial pass here as I believe it's working.

The difference appears to be between pulling the tar.gz version of the code (which already has a ./configure) from github vs git cloning it and running autogen.sh followed by configure/make. May take longer to track down, and not a great reason to hold this up.

**Comment by brianmego** [comment]
OK, made a bit of progress. The big change was to pull the repo from the tag rather than the release:

`SRC_URI="https://github.com/tstack/lnav/archive/refs/tags/v${PV}.tar.gz"`
vs
`SRC_URI="https://github.com/tstack/lnav/releases/download/v${PV}/${P}.tar.gz"`

This allows us to build our own `configure` script from autogen.sh during the `src_configure` step of the build, which fixes the test_sql.sh and test_logfile.sh tests for reasons I don't understand yet. But compiling from source seems more the gentoo way anyway, so I'm ok with it.

Secondly, this cleaned up the patches, removing individual tests rather than entire test files where possible. One of these was test_tui.sh which was causing the hang in the test suite.

Finally, I have been able to observe some flakiness in the test suite with the test_logfile.sh file. There appears to be a race condition in which sometimes the test file is read before data is flushed to it, resulting in a diff and therefore a test failure. It's ~20% of the test runs on my box, though I'm interested to see if this is reproducible with someone else. In the meantime, I've left it in.

I believe this is ready for review. Thanks for your patience.

**Review by bowlofeggs — CHANGES_REQUESTED** [review]
Hi again @brianmego!

Nice work! I still see one test failure: `test_mvwattrline.sh` (same as I had reported in https://github.com/tstack/lnav/issues/911). Does this one pass for you?

I haven't observed the log one failing, but if it's flaky I wouldn't be opposed to commenting it out too. It seems like some of these tests are really integration tests, and the Gentoo build environment isn't always appropriate for these kinds of tests in my opinion, so I don't think it's terrible to comment some tests out if we believe that they are due to the environment.

By the way, the action I am performing to trigger the tests is simply emerging it with `FEATURES="collision-protect ipc-sandbox network-sandbox sandbox split-log split-elog strict test userfetch userpriv usersandbox -buildpkg"`. Some of these could be the differences in failures. I use a container in podman to do my ebuild development, on amd64.

**Comment by bowlofeggs** [comment]
Another data point: When I run `ebuild lnav-0.10.0.ebuild clean prepare configure compile` I do not see the failure that I reported above, but I do see it when I emerge with the `FEATURE` flags I mentioned earlier. I wonder if the test is trying to do something forbidden by one of those features.

File: app-admin/lnav/lnav-0.10.0.ebuild [hunk 2]
```diff
@@ -0,0 +1,42 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools
+
+DESCRIPTION="A curses-based tool for viewing and analyzing log files"
+HOMEPAGE="https://lnav.org"
+SRC_URI="https://github.com/tstack/lnav/archive/refs/tags/v${PV}.tar.gz"
```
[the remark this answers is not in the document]
brianmego: The link you suggested did not work for me. I believe there was a redirect involved that ended up causing issues. With that said, I found another example in the codebase of a rename operation ( -> newname.tar.gz), which works great. Thanks for the nudge!

brianmego:
I am unable to get the same test failures you have, even with the FEATURES set the same way. I am just running on bare metal, also amd64. Not sure where to go next here.

bowlofeggs:
@thesamesam Would you have objections if we just patched out the two remaining tests? (One fails for me but not Brian, the other fails for Brian but not me.) Some of lnav's tests seem to require a particular environment, and the test environment doesn't seem to satisfy it.

bowlofeggs:
It is possible that it fails because I am running in a container environment too.

thesamesam:
> @thesamesam Would you have objections if we just patched out the two remaining tests? (One fails for me but not Brian, the other fails for Brian but not me.) Some of lnav's tests seem to require a particular environment, and the test environment doesn't seem to satisfy it.

Fine with me, just pop in a comment explaining it if you can. Cheers!

app-admin/lnav/files/lnav-0.10.0-disable-remote-tests.patch
```diff
@@ -0,0 +1,75 @@
+--- lnav-0.10.0/test/test_remote.sh
++++ lnav-0.10.0/test/test_remote.sh
+@@ -76,38 +76,38 @@
+ error: unable to open file: nonexistent-host: -- failed to ssh to host: ...
+ EOF
+
+-run_test ${lnav_test} -d /tmp/lnav.err -n \
+-    localhost:nonexistent-file
++# run_test ${lnav_test} -d /tmp/lnav.err -n \
```
thesamesam, line 9: I don't want to hold up the PR over this and it's not a big deal, but for future, we can actually just drop lines (no need to comment) because the diff will preserve their contents as - lines anyway.

Also, /tmp is suspicious, we should allow using ${T} from the ebuild environment.

app-admin/lnav/files/lnav-0.10.0-disable-tests.patch
```diff
@@ -0,0 +1,34 @@
...
+# mvattrline: Fails on some testers boxes. Documented here: https://github.com/tstack/lnav/issues/911
+# logfile: Flaky. Fails some times and not others. i.e. Race condition
+# tui: Can hang the build
+--- lnav-0.10.0/test/Makefile.am
++++ lnav-0.10.0/test/Makefile.am
+@@ -355,14 +355,14 @@
+ 	test_cli.sh \
+ 	test_cmds.sh \
+ 	test_config.sh \
+-	test_listview.sh \
```
thesamesam, line 12: See other comment about commenting.

thesamesam (approved):

PR #22251, opened 2021-09-09
net-irc/inspircd: Version bump

clinew:
Add 3.11.0
Drop 3.10.0-r1

(I should have submitted a stabilization request for 3.10.0-r1 but I forgot, sorry :()


net-irc/inspircd/inspircd-3.11.0.ebuild
```diff
@@ -0,0 +1,122 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="7"
```
juippis: 8 should work here. I'll test and add it if it does. 

Thanks!

juippis (approved):

net-irc/inspircd/inspircd-3.11.0.ebuild
clinew: Thanks juippis!

PR #22253, opened 2021-09-09
media-gfx/hydrus: Drop 450, add 454

vaartis:
No changes to the ebuild, still.


media-gfx/hydrus/hydrus-454.ebuild
```diff
@@ -0,0 +1,139 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
juippis: 8 should work here, some changes in `src_install` that might be affected. I'll test and add it if it does.

Thanks!

juippis (approved):

media-gfx/hydrus/hydrus-454.ebuild
vaartis: Somehow I didn't know EAPI 8 is already a thing at all.. Thank you.

PR #22254, opened 2021-09-09
media-sound/mpdscribble: add version 0.23

ArsenArsen:
Addresses https://bugs.gentoo.org/812275 and https://bugs.gentoo.org/784398

Keywords in `0.23` were adjusted to my testing abilities (and even with that might be inaccurate still, since I haven't explicitly tested on `x86` alone). I believe bug https://bugs.gentoo.org/605374 can also be addressed in this commit, with some further changes.


media-sound/mpdscribble/mpdscribble-0.23.ebuild
```diff
@@ -0,0 +1,46 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
juippis, line 4: 8 should work now.

media-sound/mpdscribble/mpdscribble-0.23.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+inherit meson systemd
+
+DESCRIPTION="An MPD client that submits information to Audioscrobbler"
+HOMEPAGE="https://www.musicpd.org/clients/mpdscribble/"
```
juippis, line 9: Could you also add the Github page as 2ndary HOMEPAGE here?

media-sound/mpdscribble/mpdscribble-0.23.ebuild
```diff
@@ -0,0 +1,46 @@
...
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64"
+
+RDEPEND="
+	media-libs/libmpdclient
+	net-misc/curl
+	dev-libs/boost
+	dev-libs/libgcrypt
+"
```
juippis, line 21: Please sort

media-sound/mpdscribble/files/mpdscribble-0.23-Don-t-install-AUTHORS-COPYING-NEWS-README.rst.patch
```diff
@@ -0,0 +1,29 @@
...
+-  'AUTHORS', 'COPYING', 'NEWS', 'README.rst',
+-  install_dir: join_paths(get_option('datadir'), 'doc', meson.project_name()),
+-)
+-
+ subdir('systemd')
+ 
+ subdir('doc')
+-- 
+2.32.0
+
```
juippis, line 29: I guess there's some sort of conflict with compressing these files properly?

media-sound/mpdscribble/mpdscribble-0.23.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+DOCS=( AUTHORS COPYING NEWS README.rst )
+
+DEPEND="${RDEPEND}"
```
juippis, line 25: Please move this below RDEPEND.

media-sound/mpdscribble/mpdscribble-0.23.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+PATCHES=(
+	"${FILESDIR}"/${PN}-0.23-Unconditionally-install-systemd-unit-files.patch
+	"${FILESDIR}"/${PN}-0.23-Don-t-install-AUTHORS-COPYING-NEWS-README.rst.patch
+)
+
+src_prepare() {
+	# we install these below via systemd_do{,user}unit
+	sed -i '/install_dir/d' -i systemd/*/meson.build
```
juippis, line 34: Don't you already introduce some sort of patch file for this? Why's the sed needed anymore? Besides, external commands (`sed`) needs to `|| die`.

juippis (commented):
Okay let's process this. Few notes.

media-sound/mpdscribble/mpdscribble-0.23.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64"
+
+RDEPEND="
+	media-libs/libmpdclient
+	net-misc/curl
+	dev-libs/boost
+	dev-libs/libgcrypt
```
juippis, line 20: Also if this gets linked, add a subslot binder, 
`dev-libs/libgcrypt:=`

media-sound/mpdscribble/files/mpdscribble-0.23-Don-t-install-AUTHORS-COPYING-NEWS-README.rst.patch
ArsenArsen: yes, a version with and without compression would be installed

ArsenArsen:
All that should be fixed, I believe.

BTW, is the sort order for fields defined anywhere? For future reference

juippis:
> BTW, is the sort order for fields defined anywhere? For future reference

Theory:
https://devmanual.gentoo.org/general-concepts/dependencies/index.html

Practical example:

FYI getting:
```
 * QA Notice: This ebuild installs into paths that should be created at runtime.
 *  To fix, simply do not install into these directories.  Instead, your package
 *  should create dirs on the fly at runtime as needed via init scripts/etc...
 * 
 *   var/cache
 *   var/cache/mpdscribble
```
Contribution looks fine, but I'd like to hear your opinion on this: Do you think you can modify the systemd service file & openrc init script to accomplish this? It's not a requirement, will merge if you don't want to put time for that now, but it'd be a great extra. Let me know.

ArsenArsen:
I did ask about this on IRC, since this is an old issue. This seems like a job for the *tmpfile implementations. I can add the configuration for those.

juippis (approved):
Well let's merge and fix the issues this PR was originally raised for. If you do end up fixing the init.d/service files, please ping me directly in Github so I see it (not part of the project that maintains this)

PR #22258, opened 2021-09-10
app-text/zathura-cb: add jpeg optfeature

N-R-K:
Adjusted the ebuild according to @juippis 's suggestion on https://github.com/gentoo/gentoo/pull/20965 (not sure how to reopen that). Please take a look.

Fixes: https://git.pwmt.org/pwmt/zathura/-/issues/213

### DIFF
--- a/app-text/zathura-cb/zathura-cb-0.1.8.ebuild
+++ b/app-text/zathura-cb/zathura-cb-0.1.8.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-inherit meson xdg-utils
+inherit meson xdg-utils optfeature
 
 if [[ ${PV} == *9999 ]]; then
 	inherit git-r3
@@ -31,6 +31,8 @@ BDEPEND="virtual/pkgconfig"
 
 pkg_postinst() {
 	xdg_desktop_database_update
+
+	optfeature "jpeg support" x11-libs/gdk-pixbuf:=[jpeg]
 }
 
 pkg_postrm() {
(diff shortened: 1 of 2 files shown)


app-text/zathura-cb/zathura-cb-0.1.8.ebuild
```diff
@@ -31,6 +31,8 @@ BDEPEND="virtual/pkgconfig"
 
 pkg_postinst() {
 	xdg_desktop_database_update
+
+	optfeature "jpeg support" x11-libs/gdk-pixbuf:=[jpeg]
```
juippis, line 35: `:=` doesn't do anything in here, it's a subslot binder and causes rebuilds on SONAME / ABI updates. Since this doesn't get linked in compile time, or it shouldn't, it doesn't work here either.

juippis (approved):
Thanks! 

BTW we're working on clarifying the "real name requirement" part, so please do continue contributing even without sign-offs. 
https://archives.gentoo.org/gentoo-project/message/b8d2bea78308fcb5fd6eca6f66e323d1

app-text/zathura-cb/zathura-cb-0.1.8.ebuild
N-R-K: Should I remove `:=` then?

juippis: I did before pushing, thought it's faster since I was going to push anyway.

PR #22259, opened 2021-09-10
llvm.org.eclass target improvements

mgorny:


thesamesam:
Merged a few days ago.

PR #22260, opened 2021-09-10
app-editors/vscod* : version bump + proper deps + wayland launch icons

AdelKS:
Hello,

On the menu in this PR:

- A little version bump
- The dependencies should be properly handled. Closes https://bugs.gentoo.org/811759
- Launch icon to run in native wayland. I had to rename `Visual Studio Code` to `VSCode` in the icon's name so `Wayland` shows.
- Additional symbolic links to the executable with upstream naming scheme `code` and `codium`, which partly solves https://bugs.gentoo.org/809401. But I do not know what else we can do about it except change the package name, which I am against. Although if a majority would like that name change to happen, let's do it.

Thanks!


RarogCmex:
I've installed vscode from that pr and I can confirm it works.

juippis (approved):
```
 * Messages for package app-editors/vscodium-1.60.0:
 * Log file: /var/tmp/portage/vbslogs/build/app-editors/vscodium-1.60.0:20210914-071226.log

 * Package:    app-editors/vscodium-1.60.0
 * Repository: gentoo
 * Maintainer: adel.ks@zegrapher.com proxy-maint@gentoo.org
 * Upstream:   https://github.com/VSCodium/vscodium/issues
 * USE:        abi_x86_64 amd64 elibc_glibc kernel_linux userland_GNU
 * FEATURES:   network-sandbox preserve-libs sandbox userpriv usersandbox
 * 
 * 	/opt/vscodium/codium: libcups.so.2
 * 
 * Final size of build directory: 293768 KiB (286.8 MiB)
 * Final size of installed tree:  294040 KiB (287.1 MiB)
 * QA: listing app-editors/vscodium-1.60.0/image info
 * SONAME: libEGL.so(64)
 * SONAME: libGLESv2.so(64)
 * SONAME: libffmpeg.so(64)
 * SONAME: libvk_swiftshader.so(64)
 * SONAME: libvulkan.so.1(64)
 * Updating .desktop files database ...
 * When compared to the regular VSCode, VSCodium has a few quirks
 * More information at: https://github.com/VSCodium/vscodium/blob/master/DOCS.md
 * >>> SetUID: [chmod go-r] /opt/vscodium/chrome-sandbox ...
```
```
tmp@my-test-container-snap-22260 /root $ vscodium 
/opt/vscodium/bin/../codium: error while loading shared libraries: libcups.so.2: cannot open shared object file: No such file or directory
```
Have to add cups as a dep here, I'll do it. Thanks!

AdelKS:
> Have to add cups as a dep here

Yea I saw that, and I find it not normal... `vscodium` adds no extra functionality over `vscode` and the latter doesn't need `cups`. Also, there is no printing feature in `vscode` as far as I know. So I decided to ignore `cups` as a dependency since I suppose some people don't wanna install `cups` just for this software that does not offer any printing capabilities.

ionenwks:
It's not uncommon that libcups is needed even if no printing support in binary-only packages, often they build dependencies with default settings and they get built against unnecessary libraries. And then if missing it won't won't work, so you need to list it anyway.

PR #22262, opened 2021-09-10
profiles/arch/sparc/package.use.stable.mask: drop obsolete masks

DerDakon:
mbedtls, nftables and xterm are stable on sparc meanwhile.


DerDakon:
@thesamesam @mattst88

PR #22263, opened 2021-09-10
Fix icon_cache_update in several ebuilds

parona-source:
There are several ebuilds at EAPI 7 which incorrectly use gnome2_icon_cache_update which isnt enabled for EAPI 7 https://github.com/gentoo/gentoo/commit/1b183b8fb6e5c0ce2265265556fe5d061b959f79. This pull request updates any that I found with `grep -lr "gnome2_icon_cache_update" | xargs grep -l "EAPI=[78]"`. With most I just went with using xdg.eclass and removing pkg_{preinst,post{inst,rm}} where I could as being redundant.

With zathura-djvu-0.2.9 i gave it same treatment as zathura-djvu-9999 even though it didnt use gnome2_icon_cache_update.

### DIFF
--- a/app-text/zathura-djvu/zathura-djvu-0.2.9.ebuild
+++ b/app-text/zathura-djvu/zathura-djvu-0.2.9.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-inherit meson xdg-utils
+inherit meson xdg
 
 if [[ ${PV} == *9999 ]]; then
 	inherit git-r3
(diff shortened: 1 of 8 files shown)


ionenwks (approved):
Had to deal with many of those in games-* ebuilds before (was widespread there). Hadn't bothered with sauerbraten given it was only the old version that's affected and, on that note, I'll drop the commit for that one on merge and delete the old version instead.

Thanks :)

PR #22267, opened 2021-09-11
**/metadata.xml: Replace http by https in DOCTYPE element

ulm:
Bug: https://bugs.gentoo.org/552720


mgorny:
Do you recall what was the issue when jlec last tried it? repoman hardcoding the URL?

ulm:
> Do you recall what was the issue when jlec last tried it? repoman hardcoding the URL?

See bug 552720 (linked above). Should be fixed with this commit:
https://gitweb.gentoo.org/proj/portage.git/commit/?id=3950d76df63cf0f3f6502014dd61b8aa9c260e36

mgorny (approved):
Let's do it then.

floppym:
If our tooling has been update, this seems fine to me. Viewing the file list locks up my web browser, so I can't actually approve this in github.

We should also update the templates in app-vim/gentoo-syntax.

ulm:
Merged.

PR #22268, opened 2021-09-11
LLVM 10 cleanup

mgorny:
Maintainers, please ACK cleanups. There's no need to do this immediately, I can do it all at once when everything is ready for the removal.

- [x] dev-python/llvmlite
- [ ] dev-libs/opencl-clang
- [ ] dev-util/spirv-llvm-translator
- [ ] media-gfx/blender
- [ ] media-libs/osl

Note that some of them depend on stablereqs that I've filed (and need your ACK too) but feel free to ACK earlier. I won't clean until the tracker is clean, i.e. everything's been stabilized or ported, or lastrited, or whatever.

Tracker: https://bugs.gentoo.org/812533


ConiKost:
As for intel-neo, we will have to wait until we can stabilize the new version. This will take a few days more, so I think, we can start stabilizing at the beginning of october and afterwards drop all old versions.

ConiKost:
Looks like LLVM10 is long gone, so closing here. Please reopen, if I am not correct.

PR #22269, opened 2021-09-11
dev-util/rebar-bin-3.17.0

00-matt:


dev-util/rebar-bin/rebar-bin-3.17.0.ebuild
```diff
@@ -0,0 +1,30 @@
...
+
+EAPI=8
+
+MY_PN=${P#-bin}
+MY_P=${MY_PN}-${PV}
+
+DESCRIPTION="A sophisticated build-tool for Erlang projects that follows OTP principles"
+HOMEPAGE="https://rebar3.org https://github.com/erlang/rebar3"
+
+SRC_URI="https://github.com/erlang/rebar3/releases/download/${PV}/rebar3 -> ${P}"
```
ionenwks, line 12: Just a nitpick but usually no blank line between HOMEPAGE and SRC_URI, and S= would usually be right after SRC_URI without a blank line to match new skel.ebuild ordering.

dev-util/rebar-bin/rebar-bin-3.17.0.ebuild
```diff
@@ -0,0 +1,30 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+MY_PN=${P#-bin}
+MY_P=${MY_PN}-${PV}
```
ionenwks, line 7: These aren't used anywhere.

ionenwks (commented):
Can ignore CI, verified that it was bogus.

PR #22271, opened 2021-09-11
app-shells/mcfly: version bump (0.5.9)

mm1ke:
Hi,

Simple version bump for mcfly.


app-shells/mcfly/mcfly-0.5.9.ebuild
```diff
@@ -0,0 +1,117 @@
...
+	$(cargo_crate_uris ${CRATES})"
+
+LICENSE="Apache-2.0 Apache-2.0-with-LLVM-exceptions BSD BSD-2 CC0-1.0 MIT Unlicense"
+SLOT="0"
+KEYWORDS="~amd64"
+
+DEPEND="dev-db/sqlite:3"
+RDEPEND="${DEPEND}"
+
+QA_FLAGS_IGNORED="/usr/bin/mcfly"
```
ionenwks: Not super-important but normally QA_* variables use relative paths, i.e. no leading /

Not that it's well documented, briefly mentioned under the variables in:
https://devmanual.gentoo.org/eclass-reference/ebuild/index.html

app-shells/mcfly/mcfly-0.5.9.ebuild
```diff
@@ -0,0 +1,117 @@
...
+	winapi-i686-pc-windows-gnu-0.4.0
+	winapi-x86_64-pc-windows-gnu-0.4.0
+"
+
+inherit cargo
+
+DESCRIPTION="Context-aware bash history search replacement (ctrl-r)"
+HOMEPAGE="https://github.com/cantino/mcfly"
+SRC_URI="https://github.com/cantino/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz
+	$(cargo_crate_uris ${CRATES})"
```
ionenwks: Change in July to cargo.eclass allows to omit `${CRATES}`, it's used by default if unspecified.

Also, fwiw

app-shells/mcfly/mcfly-0.5.9.ebuild
```diff
@@ -0,0 +1,117 @@
...
+	cargo_src_install
+
+	insinto /usr/share/${PN}
+	doins ${PN}.{bash,fish,zsh}
+
+	einstalldocs
+}
+
+pkg_postinst() {
+	elog "To start using ${PN}, add the following to your shell:"
```
ionenwks: That's a pretty huge wall of elog to show every time.

Have you considered making it a README.gentoo? Note that you can still use EPREFIX and the like to create the README.gentoo (like I do in e.g. nvidia-drivers, or quakeforge -- rather than a static file)

app-shells/mcfly/mcfly-0.5.9.ebuild
```diff
@@ -0,0 +1,117 @@
...
+HOMEPAGE="https://github.com/cantino/mcfly"
+SRC_URI="https://github.com/cantino/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz
+	$(cargo_crate_uris ${CRATES})"
+
+LICENSE="Apache-2.0 Apache-2.0-with-LLVM-exceptions BSD BSD-2 CC0-1.0 MIT Unlicense"
+SLOT="0"
+KEYWORDS="~amd64"
+
+DEPEND="dev-db/sqlite:3"
+RDEPEND="${DEPEND}"
```
ionenwks: This dependency unfortunately doesn't seem used.

The libsqlite3-sys crate is using bundled sqlite3

mm1ke:
@ionenwks 
Thanks again for your in-deep review. I've addressed your point and hope i haven't forget anything :)

app-shells/mcfly/mcfly-0.5.9.ebuild
```diff
@@ -0,0 +1,120 @@
...
+	doins ${PN}.{bash,fish,zsh}
+
+	# create README.gentoo
+	local DISABLE_AUTOFORMATTING="yes"
+	local DOC_CONTENTS=\
+"To start using ${PN}, add the following to your shell:
+
+~/.bashrc
+p='${EPREFIX}/usr/share/${PN}/${PN}.bash'
+[[ -f ${p} ]] && source ${p}
```
ionenwks, line 102: The three `p=` weren't part of the elog before, it was used to shorten having to copy/paste the path several times.

This ends up with a README.gentoo that:
```bash
p='/usr/share/mcfly/mcfly.bash'
[[ -f  ]] && source 
```
As you can see, the paths are missing.

There's a few possible routes to take here:
1. do similarly to the elog, either by splitting and appending to DOC_CONTENTS, or by creating three variables (p1,p2,p3 or so) set beforehand, but that sound needlessly messy.
2. no longer use p= and just embed the path everywhere needed
3. make p= actually part of the scripts by escaping `\${p}`, i.e. to result in
```zsh
p='/usr/share/mcfly/mcfly.bash'
[[ -f ${p} ]] && source "${p}"
```
(personally never really tried zsh/fish, so unsure how they want these)

4. As an extension to 2, using these `[[ -f ${p} ]] &&` may not be overly needed in the first place for an example *rc, removing would simplify by only needing to give paths once for sourcing.

mm1ke: Option 5: Update the Information according to upstream suggestions ;)
I've checked the upstream readme and saw there is a much easier way to use mcfly which i simply took over. Also doesn't require the path. Hope thats fine.

ionenwks (approved):
Nice, matching upstream with option 5 seems better indeed.

PR #22274, opened 2021-09-12
mail-filter/rspamd: mask jemalloc use-flag

arkamar:
- mask `jemalloc` due to the bug [#810337](https://bugs.gentoo.org/show_bug.cgi?id=810337)
- ~introduce `lto` use flag, it was disabled by default~

### DIFF
--- a/profiles/base/package.use.mask
+++ b/profiles/base/package.use.mask
@@ -6,6 +6,12 @@
 # This file is only for generic masks. For arch-specific masks (i.e.
 # mask everywhere, unmask on arch/*) use arch/base.
 
+# Petr Vanek <arkamar@atlas.cz> (2021-09-12)
+# rspamd-3.0 segfaults a lot with jemalloc enabled (Bug #810337)
+# It is caused by -Wl,--as-needed linker flag,
+# see also: https://github.com/rspamd/rspamd/issues/3871
+~mail-filter/rspamd-3.0 jemalloc
+
 # James Le Cuirot <chewi@gentoo.org> (2021-09-04)
 # Currently requires an unreleased version of FFmpeg. If you really want it,
 # unmask the flag and emerge ffmpeg-9999 with this environment variable set:


profiles/base/package.use.mask
```diff
@@ -6,6 +6,12 @@
 # This file is only for generic masks. For arch-specific masks (i.e.
 # mask everywhere, unmask on arch/*) use arch/base.
 
+# Petr Vanek <arkamar@atlas.cz> (2021-09-12)
+# rspamd-3.0 segfaults a lot with jemalloc enabled (Bug #810337)
+# It is caused by -Wl,--as-needed linker flag,
+# see also: https://github.com/rspamd/rspamd/issues/3871
+~mail-filter/rspamd-3.0 jemalloc
```
thesamesam, line 13: You changed your mind! No runtime configuration change needed, I guess?

mail-filter/rspamd/rspamd-3.0-r3.ebuild
```diff
@@ -92,6 +92,7 @@ src_configure() {
 		-DENABLE_HYPERSCAN=$(usex cpu_flags_x86_ssse3 ON OFF)
 		-DENABLE_JEMALLOC=$(usex jemalloc ON OFF)
 		-DENABLE_LUAJIT=$(usex lua_single_target_luajit ON OFF)
+		-DENABLE_LTO=$(usex lto ON OFF)
```
thesamesam, line 95: Does this just control -flto or something more? If so, we try to avoid USE=lto (although that said, I personally like advertising if an upstream supports it or not).

profiles/base/package.use.mask
arkamar: Yes I did :) And yes, you are right.

mail-filter/rspamd/rspamd-3.0-r3.ebuild
arkamar: It seems to just control `-flto` plus it appends `-fno-fat-lto-objects` as well.

For `-DENABLE_LTO=ON` it calls:
```cmake
	check_ipo_supported (RESULT SUPPORT_LTO OUTPUT LTO_DIAG )
		if (SUPPORT_LTO)
			set (CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
```

arkamar:
FP without `lto` change.

mail-filter/rspamd/rspamd-3.0-r3.ebuild
thesamesam: Hm, let's leave it out for now then (as you have).

PR #22278, opened 2021-09-12
Adding wireplumber

pinkflames:
[According to Wim Taymans](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1170#note_927579), the lead developer of PipeWire, [Fedora 35 is planned to switch to wireplumber](https://fedoraproject.org/wiki/Changes/WirePlumber#Switch_to_WirePlumber_as_the_PipeWire_session_manager) as the replacement for pipewire-media-session daemon, so Gentoo as a rolling and up to date distribution should also provide the option.

Supporting USE=doc would require dev-python/sphinx so for now API docs are omitted

TODO:
- [x] Add some instructions and/or support for OpenRC users


profiles/targets/desktop/package.use
```diff
@@ -1,6 +1,11 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Niklāvs Koļesņikovs <89q1r14hd@relay.firefox.com> (2021-09-13)
+# WirePlumber obligatory links and embeds either Lua 5.3 or 5.4
+# interpreter with the later being the upstream default.
```
thesamesam, line 6: Nit:
```suggestion
# interpreter with the latter being the upstream default.
```

media-video/wireplumber/wireplumber-9999.ebuild
```diff
@@ -0,0 +1,80 @@
...
+	dev-util/glib-utils
+"
+
+DEPEND="
+	${LUA_DEPS}
+	>=dev-libs/glib-2.62
+	>=media-video/pipewire-0.3.26
+	virtual/libc
+	systemd? ( sys-apps/systemd )
+
```
thesamesam, line 41:
```suggestion
```

media-video/wireplumber/wireplumber-9999.ebuild
```diff
@@ -0,0 +1,80 @@
...
+#	')
+RDEPEND="${DEPEND}"
+
+DOCS=( {NEWS,README}.rst )
+
+src_configure() {
+	local emesonargs=(
+		-Dintrospection=disabled # Likely only used for doc generation
+		-Dsystem-lua=true # We always unbundle everything we can
+		#-Dsystem-lua-version=<version number> # Some distros can't enforce deterministc system Lua themselves
```
thesamesam, line 56: Should we do:
$(lua_get_version)
```suggestion
		-Dsystem-lua-version=$(lua_get_version)
```
instead, just for future proofing? Or no need because it respects some environment variable the eclass sets?

media-video/wireplumber/wireplumber-9999.ebuild
```diff
@@ -0,0 +1,80 @@
...
+		$(meson_feature systemd)
+		-Dsystemd-system-service=false # Matches upstream
+		$(meson_use systemd systemd-user-service)
+		#-Dsystemd-system-unit-dir=""
+		#-Dsystemd-user-unit-dir=""
```
thesamesam, line 61:
```suggestion
		-Dsystemd-user-unit-dir=$(systemd_get_userunitdir)
```

media-video/wireplumber/wireplumber-9999.ebuild
```diff
@@ -0,0 +1,80 @@
...
+
+src_configure() {
+	local emesonargs=(
+		-Dintrospection=disabled # Likely only used for doc generation
+		-Dsystem-lua=true # We always unbundle everything we can
+		#-Dsystem-lua-version=<version number> # Some distros can't enforce deterministc system Lua themselves
+		$(meson_feature systemd)
+		-Dsystemd-system-service=false # Matches upstream
+		$(meson_use systemd systemd-user-service)
+		#-Dsystemd-system-unit-dir=""
```
thesamesam, line 60:
```suggestion
		-Dsystemd-system-unit-dir=$(systemd_get_systemunitdir)
```

media-video/wireplumber/wireplumber-9999.ebuild
```diff
@@ -0,0 +1,80 @@
...
+		#-Dsystemd-user-unit-dir=""
+	)
+
+	meson_src_configure
+}
+
+src_install() {
+	# No idea if this is needed with meson but eclass example has this for emake
+	LUA_VERSION="$(lua_get_version)" meson_src_install
```
thesamesam, line 69: You can test by installing both (5.3 and 5.4 which you said is preferred upstream) and then see if it uses 5.3 when you express a preference for it.

But LUA_VERSION is probably just an example for some random Makefile. Likely not necessary here.

media-video/wireplumber/wireplumber-9999.ebuild
```diff
@@ -0,0 +1,80 @@
...
+}
+
+pkg_postinst() {
+	elog "To replace media-session with WirePlumber when using systemd:"
```
thesamesam, line 73: if systemd_is_booted ;

media-video/wireplumber/wireplumber-9999.ebuild
```diff
@@ -0,0 +1,80 @@
...
+	elog "systemctl --user disable --now pipewire-media-session.service"
+	elog "systemctl --user enable --now wireplumber.service"
+	elog
+	elog "OpenRC users need to copy ${EROOT}/usr/share/pipewire/pipewire.conf"
```
thesamesam, line 77: else

media-video/wireplumber/wireplumber-9999.ebuild
```diff
@@ -0,0 +1,80 @@
...
+	elog "to ${EROOT}/etc/pipewire/pipewire.conf and in it replace"
+	elog "/usr/bin/pipewire-media-session with /usr/bin/wireplumber"
```
thesamesam, line 79: ${EROOT} both of these

media-video/wireplumber/wireplumber-9999.ebuild
pinkflames: If the eclass couldn't enforce it, then the SINGLE_LUA_TARGET would be a sham, so I just hand-waved it away by assuming that the  lua-single.eclass is better than certain other distros and can enforce the chosen version. :sweat_smile:

As you might be aware, Lua has no concept of consistent version naming, causing each distro to came up with their own pkg-config file naming scheme. Gentoo probably didn't want to be left out and for 5.4 implemented all(?) of them: `lua-5.4.pc lua5.4.pc lua54.pc`one of which we need pass to that Meson switch (minus the file extension).

For now I'll install 5.3 alongside 5.4 and do some configure runs to see if it reliably picks the version requested (I may or may not have done this at some point) in hopes of not having to bash `lua_get_version()` into one of the accepts formats.

thesamesam:
Thanks!

PR #22282, opened 2021-09-13
media-gfx/sxiv: add webp useflag

N-R-K:
since imlib2 supports webp, sxiv does too.
see: https://github.com/muennich/sxiv/pull/348#issuecomment-560167643

### DIFF
--- a/media-gfx/sxiv/sxiv-26.ebuild
+++ b/media-gfx/sxiv/sxiv-26.ebuild
@@ -18,12 +18,12 @@ HOMEPAGE="https://github.com/muennich/sxiv/"
 
 LICENSE="GPL-2"
 SLOT="0"
-IUSE="exif gif +jpeg +png"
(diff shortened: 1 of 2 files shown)


gyakovlev (approved):
looks reasonable, thanks!

gyakovlev:
the only thing I changed - is I did not enable it by default
https://github.com/gentoo/gentoo/commit/c280708f3cc44ab8fe55d9846da308d20bb948f2

N-R-K:
Just FYI, I had it enabled by default since webp was enabled by default on `imlib2`.

gyakovlev:
it does not really matter since flag only pulls dependency and does not actually enable anything in sxiv itself.

we don't usually allow RDEPEND only useflags, but I can make an exception for sxiv due to almost instant rebuilds, and In my mind typical user of this software would run pretty minimal useflags, so makes sense to have a clear way of enabling  webp.

PR #22283, opened 2021-09-13
dev-python/citeproc-py: add 0.6.0 with tests

parona-source:
Added python3.10 support to app-text/rnc2rng so would have to hold back dev-python/citeproc-py at python3.9 for no reason.

Both pass tests with python3.9 and python3.10, dropped x86 keyword as I don't have access to an x86 system at this moment.


djc:
I'm not interested in testing 3.10  as long as the GA version hasn't even been released yet.

parona-source:
> I'm not interested in testing 3.10 as long as the GA version hasn't even been released yet.
> 
> Once it has, please submit a PR upstream to add/test support for it.

Ill remove it then from the pull request.
But its just a month left for the final release and if there is the same pace in enabling python3.10 as the default as with python 3.9 last year, then there will be a two month window to try and get as much of the tree updated. So id think its preferable to try and get as much of tree already sorted out when python 3.10 is getting its finishing touches.

parona-source:
@djc https://www.python.org/downloads/release/python-3100/ Would you mind if I add python 3.10 support now in this pull request?

djc:
No, though this would probably go faster if you filed a separate PR for the rnc2rng changes (and ideally you'd also still submit a PR to rnc2rng upstream -- also me -- to add 3.10 in its CI workflows).

PR #22284, opened 2021-09-13

arthurzam:
Hi again @AndrewAmmerlaan 
In this PR, I have bumped to python 3.10 all packages needed for `dev-python/pytest-xvfb`.
Do note that there is a 4 package circular dependency here.
Also added the GitHub remote-id to the metadata.xml

---

Found why and how to run the test suite of `dev-python/vncdotool`, and must say it is one of the worst I saw till today. The package expects the examples from `net-libs/libvncserver`. But those aren't intended to be installed (CMake skips them from install). By manually configuring and manually calling ninja build example, and manually installing, the test suite can be ran and work. It isn't worth it to fix the test suite in ebuild.

---

For `dev-python/pyvirtualdisplay`, managed to run tests by defining:
``` 
EPYTEST_DESELECT=(
    tests/test_smart.py::test_empty
)
```
But I'm not sure on which system you got a hang and on which test, so preferred to not enable the tests.

---

As this quite a big bump, with packages that have not an excellent upstream test suite, wanted your ACK before merging.

### DIFF
--- a/dev-python/easyprocess/easyprocess-0.3-r1.ebuild
+++ b/dev-python/easyprocess/easyprocess-0.3-r1.ebuild
@@ -1,15 +1,16 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
 
-PYTHON_COMPAT=( python3_{7,8,9} )
+PYTHON_COMPAT=( python3_{8..10} )
 
 inherit distutils-r1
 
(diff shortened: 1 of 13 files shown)


AndrewAmmerlaan (approved):
Nice, Thanks :+1: 

Yeah these packages are a bit of a maze of circular dependencies and non-functional test suites. The test suites are written in such a way that they also test reverse dependencies, which sounds good but it actually isn't because it has the unpleasant side-effect of this circular dependency if USE="test".

PR #22285, opened 2021-09-13
net-vpn/wireguard: Added selinux USE flag and dependency.

jpds:


zx2c4:
I wasn't aware we had a selinux policy for wireguard. Neat! Will merge.

PR #22289, opened 2021-09-14
net-vpn/i2pd: version bump 2.39.0

khumarahn:
A regular bump.

Version bump bug: https://bugs.gentoo.org/812843
Added a temporary fix for PID file not being deleted with openrc on exit
https://bugs.gentoo.org/808845


PF4Public:
Please don't merge, untill @khumarahn sees my reply in https://bugs.gentoo.org/808845

PF4Public:
@khumarahn Shouldn't logrotate also be modified?
https://github.com/gentoo/gentoo/blob/86acbfdd28a0b69b0a3c744905713de91954e1e6/net-vpn/i2pd/files/i2pd-2.38.0-r1.logrotate#L8

khumarahn:
> @khumarahn Shouldn't logrotate also be modified?

a very good catch, thank you. I see now that it is not easy to modify the path for the pid file because systemd also uses it, and the logic in systemd is different: we have to rely on i2pd to create the pid file. So we still need `/run/i2pd`.... This is not a perfect solution, but I reverted to `/run/i2pd`, only now on the openrc side it is not managed by i2pd and belongs to root.

We only use the pid file for logrotate. Would be nice to have some other way.

Can you check if the openrc scripts work: the pid file is created, contains the correct pid, and the graceful shutdown works? That would be much appreciated.

khumarahn:
Uhm..... I made a container with openrc and did some tests. Apparently, `superwise-daemon` creates a pidfile for itself, not for the daemon it is supervising. So it gives us a wrong pid for i2pd, from the point of view of logrotate. Also I cannot find out if `superwise-daemon` supports forking daemons, where i2pd would write its own pid.

So I don't see how to make i2pd work well with both logrotate and supervise-daemon. I'll probably scrap the last commit, and we will stick with the simple workaround which for pid file removal. At least for the time being.

Ideally there would be some different way of truncating the logs, without sending a `HUP`.

PF4Public:
> I'll probably scrap the last commit, and we will stick with the simple workaround which for pid file removal. At least for the time being.

I tink that would be an acceptable solution, given that i2pd isn't expected to crash :)

juippis (approved):
Thanks! Can we in future split adding a new version and removing an old one to their own commits? Especially since this seems to introduce some noticeable changes. Just so if something doesn't work, users / us can easily restore the previous version.

khumarahn:
> Can we in future split adding a new version and removing an old one to their own commits?

No problem, can do. I used to keep last 3 versions in the tree, but someone has been cleaning it up, so it looked that past versions are not so important from the dev point of view.

PR #22297, opened 2021-09-14
Composer slot

jkroonza:
```
jkroon@plastiekpoot ~ $ PORTDIR=/home/jkroon/projects/gentoo equery list -ipo composer
 * Searching for composer ...
[-P-] [ ~] dev-php/composer-1.10.22:0
[IP-] [ ~] dev-php/composer-1.10.22-r1:1
[-P-] [  ] dev-php/composer-2.0.13:0
[-P-] [ ~] dev-php/composer-2.0.14:0
[IP-] [ ~] dev-php/composer-2.0.14-r1:2
[-P-] [ ~] dev-php/composer-2.1.1:0
[-P-] [ ~] dev-php/composer-2.1.2:0
[-P-] [ ~] dev-php/composer-2.1.3:0
[-P-] [ ~] dev-php/composer-2.1.5:0
[IP-] [ ~] dev-php/composer-2.1.6:0
[-P-] [ ~] dev-php/composer-2.1.6-r1:2
jkroon@plastiekpoot ~ $ composer --version
Composer version 2.1.6 2021-08-19 17:11:08
jkroon@plastiekpoot ~ $ composer2 --version
Composer version 2.0.14 2021-05-21 17:03:37
jkroon@plastiekpoot ~ $ composer1 --version
Composer version 1.10.22 2021-04-27 13:10:45
jkroon@plastiekpoot ~ $
```

So this gets us to the multiple versions installed stage.  Next up would be to create an eselect-composer (which I'm sure I can put under my own profile, but I'm not sure what Gentoo policy is here).

As it stands I believe this can be merged, however, it may be preferable to wait until such time as eselect-composer is ready too.


Whissi (commented):
I am sorry but like posted in bug https://bugs.gentoo.org/813045, I will not ACK this change.

The way we maintain PHP lib packages in Gentoo is like we maintain most other lib packages, for example Python libs: We can't have them multiple times (=no real slotting) because there is no way for libs and applications to understand slotting (which is causing all the drama when we have to phase out a new major Python version). Where we support slotting, there is always another helper mechanism in place like pkg-config combined with unique names allowing to differentiate between libfoo-1 and libfoo-2. But if libfoo would install same PC file for v1 and v2 we would be out of the game (sure, we could patch package and rename file on our own but all depending packages would need to learn to check for our custom name in addition or they would be stuck with the version installing the PC file they know).

Look at all the deps with their version pinning. In most cases this is not necessary and it is caused by the way how modern devs do their work today. But that's not the point here. Even if you find a composer-1, composer-2.0, composer-2.1 and composer-2.2 version which could co-exist **today**, there **will** be a future version which **will require a newer version** of a library breaking backward compatibility so that other depending applications/libraries can't use this dep anymore (and that happened in the past when 2.0 came out -- this is not just a theoretical problem).

Combined with the fact that most projects are no longer shipping tests ([tests are getting added to ignore list](https://github.com/composer/composer/blob/d64d1adf6101320b076085cb08f613854e757c2d/.gitattributes#L12)) because there is a spreading weird trend that tests shouldn't run downstream anymore (our own CI is better than yours and we don't like your incomplete bug reports we cannot reproduce due to 'unsupported' mix of deps) where we cannot workaround (only git cloning would work; creating our own release tarballs would be overkill) we are now unable to know if really everything is working like expected.

So no, I am sorry but I cannot ACK such a contribution:

- Please see the reasons I outlined above.
- Please always keep in mind that stuff we ship must have a chance to work the long road. I.e. introducing slotted composer because you find a version combination working today but killing slotted composer again in 1,2,3,4,5 weeks or months because an update would break is not an option.
- Please note that composer-1 is EOL and [access to packagist](https://blog.packagist.com/deprecating-composer-1-support/) is already limited for that version.
- And yes, I would still recommend using Phar version in your case instead (and to be honest, now that we cannot run tests anymore I am really questioning the point in maintaining all the deps without any other consumer vs. switching to Phar version in general; to keep the Gentoo way we would maybe create that Phar file on our own (so you could apply custom changes via patches) but more like we maintain Rust/Go packages these days and don't install all the deps (Rust crates or Go modules) as separate package -- but that's a different discussion).

jkroonza:
Whissi - I feel and share your frustration regarding the way "modern devs" operate.  In some cases these pins are required, but I tend to agree that frequently this is not the case.

I cannot argue with anything you've said, and will have to find a different solution, of the three options I've got this was the one I least disliked.

jkroonza:
Whissi - I feel and share your frustration regarding the way "modern devs" operate.  In some cases these pins are required, but I tend to agree that frequently this is not the case.

I cannot argue with anything you've said, and will have to find a different solution, of the three options I've got this was the one I least disliked.

PR #22298, opened 2021-09-14
dev-libs/libbpf: Port to EAPI 8 and Version bump to 0.5.0

jsmolic:
- Reorder variables per skel.ebuild
- Drop redundant COMMON_DEPEND
- Export CC, AR and other variables instead of passing them directly to emake
- Drop USE=static-libs, since the build-system doesn't offer a way to build only
  shared library, so let's just install both libraries unconditionally
- Update 0.3 and 0.4 versions to use 9999 patch since it's the same
  as 0.3 patch

Signed-off-by: Jakov Smolić <jsmolic@gentoo.org>


chutz:
> 
>     * Drop USE=static-libs, since the build-system doesn't offer a way to build only
>       shared library, so let's just install both libraries unconditionally
I think it might be better to keep `static-libs` around, and just remove the static library in `src_install`. Ideally we shouldn't install static libraries unless they are actually needed.
> 
>     * Update 0.3 and 0.4 versions to use 9999 patch since it's the same
>       as 0.3 patch
> 
> 
> Signed-off-by: Jakov Smolić [jsmolic@gentoo.org](mailto:jsmolic@gentoo.org)

dev-libs/libbpf/libbpf-9999.ebuild
```diff
@@ -10,44 +10,35 @@ if [[ ${PV} =~ [9]{4,} ]]; then
...
+	"${FILESDIR}"/libbpf-9999-paths.patch
 )
 
-src_compile() {
+src_configure() {
 	append-cflags -fPIC
-	emake \
-		BUILD_SHARED=y \
-		LIBSUBDIR="$(get_libdir)" \
-		$(usex static-libs 'BUILD_STATIC=y' '' '' '') \
```
thesamesam, line 39: Did they remove this option or did BUILD_STATIC=y never do what it looks like?

jsmolic: Yeah they removed it, now there is only `BUILD_STATIC_ONLY` so the static library gets built either way

jsmolic:
> > ```
> > * Drop USE=static-libs, since the build-system doesn't offer a way to build only
> >   shared library, so let's just install both libraries unconditionally
> > ```
> 
> I think it might be better to keep `static-libs` around, and just remove the static library in `src_install`. Ideally we shouldn't install static libraries unless they are actually needed.

Okay I pushed new version with the flag, ideally there would be an option to disable it at build time, I'll try to cook up a patch and send it upstream.

PR #22300, opened 2021-09-14
net-im/telegram-desktop: Add libdl dependency

mid-kid:
@gyakovlev

Only affects USE=webkit, which didn't build before. No revbump.


gyakovlev:
thanks!

PR #22303, opened 2021-09-15
media-tv/plex-media-server: version bump, and removal of an old version

ephemer0l:


ionenwks (approved):
I'll fix on merge (no problem) but note the ebuild is missing from the version bump commit and was added inside the remove old one.

PR #22305, opened 2021-09-16
sys-auth/seatd: Security bump, 0.6.0 to 0.6.2

lanodan:


ionenwks (approved):
Thanks.

I'll just change the bug tags on merge, security bugs aren't supposed to be closed by us so `Closes:` tags shouldn't be used, and `Bug:` should be used instead of `Related:`

PR #22309, opened 2021-09-16
games-puzzle/quadra: EAPI8, version bump (1.3.0)

mm1ke:

@ionenwks 
New Version of `games-puzzle/quadra`. This one runs on libsdl2.
I haven't added a cleanup commit for the old one, but if you want i can add it too.. :)


games-puzzle/quadra/quadra-1.3.0.ebuild
```diff
@@ -0,0 +1,41 @@
...
+	dobin ${PN}
+
+	insinto /usr/share/${PN}
+	doins ${PN}.res
+
+	doicon images/${PN}.xpm
+	make_desktop_entry ${PN} ${PN^}
+
+	einstalldocs
+}
```
ionenwks, line 41: You can replace this entire src_install by just "default + make_desktop_entry", default will even install the .xpm icon, i.e. the makefile's install target works just fine.

games-puzzle/quadra/quadra-1.3.0.ebuild
```diff
@@ -0,0 +1,41 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit autotools desktop
+
+DESCRIPTION="Tetris clone with multiplayer support"
+HOMEPAGE="https://github.com/quadra-game/quadra"
+SRC_URI="https://github.com/${PN}-game/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
```
ionenwks, line 10: fwiw

On a related note, since I see github, had a look at the metadata and seems the remote-id is missing. Feel free to add it, or I could later (can keep the old google-code one along with it).

games-puzzle/quadra/quadra-1.3.0.ebuild
```diff
@@ -0,0 +1,41 @@
...
+
+LICENSE="LGPL-2.1"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="
+	dev-libs/boost
+	media-libs/libpng:0=
+	media-libs/libsdl2[X,sound,video]
```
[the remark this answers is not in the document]
mm1ke: Hmm, you're right. :)
When i was testing the minimal USE flags required to run this game i saw that it only worked with `video sound X`.
However, turns out not `X` is required but `opengl`

BTW, it does run on wayland too (running on `sway` with libsdl2 build without `X`).

games-puzzle/quadra/quadra-1.3.0.ebuild
mm1ke: Nice, didn't expected that - makes it much cleaner.

games-puzzle/quadra/quadra-1.3.0.ebuild
mm1ke: i've added the github remote and remove google-code.

games-puzzle/quadra/quadra-1.3.0.ebuild
```diff
@@ -0,0 +1,41 @@
...
+	sys-libs/zlib"
+DEPEND="${RDEPEND}"
+
+src_prepare() {
+	default
+	eautoreconf
+	sed -i \
+		-e "/^datagamesdir/s|\/games|\/${PN}|" \
+		Makefile.in || die
```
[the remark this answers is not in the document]
mm1ke: Nice catch, updated the `sed` and gonna change the `.am` file now

games-puzzle/quadra/quadra-1.3.0.ebuild
```diff
@@ -0,0 +1,32 @@
...
+SRC_URI="https://github.com/quadra-game/quadra/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="LGPL-2.1"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="
+	dev-libs/boost:=
+	media-libs/libpng:=
+	media-libs/libsdl2[opengl,sound,video]
```
ionenwks, line 19: > However, turns out not `X` is required but `opengl`

I don't think it does, I have -opengl by default on my libsdl2 to notice these, and just tried again and it runs fine with -opengl. From a quick look at the source, I don't see any reference to GL either.

But ultimately it can't work with /just/ minimal flags and will still need either X or wayland, but as I said we typically don't check that /at least one/ backend is selected for neither audio nor video (need USE=wayland to work on wayland, USE=X to work on X, USE=pulseaudio to use pulse, etc...). If that's ever checked it'd be libsdl2 that should with a REQUIRED_USE if USE=video rather than this ebuild anyhow.

> BTW, it does run on wayland too (running on sway with libsdl2 build without X).

Nice :)

games-puzzle/quadra/metadata.xml
```diff
@@ -6,6 +6,6 @@
 		<name>Gentoo Games Project</name>
 	</maintainer>
 	<upstream>
-		<remote-id type="google-code">quadra</remote-id>
+		<remote-id type="github">quadra-game/qaudra</remote-id>
```
ionenwks, line 9: typo
```suggestion
		<remote-id type="github">quadra-game/quadra</remote-id>
```
Also, normally these in their own "add github remote-id" or so commits. Change doesn't need to be tied to a version bump, i.e. if we ever wanted to revert this bump (for whatever reason) the metadata change doesn't need to go along with it.

Personally thought it wouldn't hurt to keep google-code too (even if version is outdated), it still tries to connect to google at runtime for servers and it has additional info.
> On a related note, [snip] (can keep the old google-code one along with it).

Not that I particularly mind either way here, github is still the active one.

games-puzzle/quadra/quadra-1.3.0.ebuild
mm1ke: Thanks for the detailed explanation - I think understand now. 
I've removed the `opengl` flag again.

ionenwks (approved):
Thanks, I'll additionally fix the open clang bug while here (seems 1.3.0 is still affected and fix is trivial)

PR #22318, opened 2021-09-17
sys-fs/zfs: install compatibility.d

implr:
Added upstream in https://github.com/openzfs/zfs/pull/11468, didn't work on Gentoo as we remove the entirety of `/usr/share/zfs/`.
I admit the brace/glob expansion is a bit ugly.

### DIFF
--- a/sys-fs/zfs/zfs-2.1.1-r1.ebuild
+++ b/sys-fs/zfs/zfs-2.1.1-r1.ebuild
@@ -1,0 +1,288 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DISTUTILS_OPTIONAL=1
+DISTUTILS_USE_SETUPTOOLS=manual
+PYTHON_COMPAT=( python3_{8,9,10} )
+
+inherit autotools bash-completion-r1 dist-kernel-utils distutils-r1 flag-o-matic linux-info pam systemd udev usr-ldscript
+
+DESCRIPTION="Userland utilities for ZFS Linux kernel module"
+HOMEPAGE="https://github.com/openzfs/zfs"
+
+if [[ ${PV} == "9999" ]]; then
+	inherit git-r3 linux-mod
+	EGIT_REPO_URI="https://github.com/openzfs/zfs.git"
+else
+	VERIFY_SIG_OPENPGP_KEY_PATH=${BROOT}/usr/share/openpgp-keys/openzfs.asc
+	inherit verify-sig
+
+	MY_P="${P/_rc/-rc}"
+	SRC_URI="https://github.com/openzfs/${PN}/releases/download/${MY_P}/${MY_P}.tar.gz"
+	SRC_URI+=" verify-sig? ( https://github.com/openzfs/${PN}/releases/download/${MY_P}/${MY_P}.tar.gz.asc )"
+	S="${WORKDIR}/${P%_rc?}"
+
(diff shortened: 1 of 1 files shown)


sys-fs/zfs/zfs-2.1.1-r1.ebuild
```diff
@@ -0,0 +1,288 @@
...
+}
+
+src_install() {
+	default
+
+	gen_usr_ldscript -a nvpair uutil zfsbootenv zfs zfs_core zpool
+
+	use pam && { rm -rv "${ED}/unwanted_files" || die ; }
+
+	use test-suite || { rm -r "${ED}/usr/share/zfs/"{test-runner,zfs-tests,runfiles,*sh} || die ; }
```
gyakovlev, line 224: `find "${ED}/usr/share/zfs" -mindepth 1 -maxdepth 1 -not -name compatibility.d -exec rm -r '{}' \;`

this should work better =)

thesamesam: Problem is, the PR version makes it easier to not miss things in future.. hm

implr: Yeah I considered something like the find, but then when they add more files there we'll drop them again and not notice.
I'm not good with autotools; would it be feasible to instead patch out/disable installing the test files at an earlier stage?

gyakovlev:
```diff
diff --git a/sys-fs/zfs/zfs-2.1.1.ebuild b/sys-fs/zfs/zfs-2.1.1.ebuild
index cf525674e3ab..f141ae27d79e 100644
--- a/sys-fs/zfs/zfs-2.1.1.ebuild
+++ b/sys-fs/zfs/zfs-2.1.1.ebuild
@@ -154,6 +154,13 @@ src_prepare() {
        default
        libsoversion_check
 
+       # remove those before build, and before autoreconf
+       if ! use test-suite; then
+               rm -rf tests || die
+               sed -i '/tests\/.*Makefile/d' configure.ac || die
+               sed -i '/^SUBDIRS\ +=/ s/tests//' Makefile.am || die
+       fi
+
        # Run unconditionally (bug #792627)
        eautoreconf
 
@@ -221,8 +228,6 @@ src_install() {
 
        use pam && { rm -rv "${ED}/unwanted_files" || die ; }
 
-       use test-suite || { rm -r "${ED}/usr/share/zfs" || die ; }
-
        find "${ED}" -name '*.la' -delete || die
 
        dobashcomp contrib/bash_completion.d/zfs
```

this is pre-build deletion version messing with autofiles.
tbh looks fragile, and leaves some extra dangling files too.
I think the PR version is fine in the long run. so let's merge it.

thesamesam: I worry about *sh matching something it shouldn't, but at some point, we can try upstream a proper way to not install these. Let's add a comment and merge the PR version?

gyakovlev:
merged, restored riscv keyword (ebuild was keyworded meanwhile) and moved quotes a bit, just cosmetic.
thanks for great first contribution to gentoo ;-) really useful.

@sam feel free to add comment to ebuild ofc if you want.

gyakovlev:
and yeah, I agree that some kind of configure switch upstream would be nice. installing those files unconditionally does not make a lot of sense for end users.

implr:
Thanks. I'll bring that up upstream if I find some time later.

PR #22321, opened 2021-09-17
net-misc/{gns3-gui,gns3-server}: version bump (2.2.25)

mm1ke:
New versions of gns3-gui and gns3-server


net-misc/gns3-gui/gns3-gui-2.2.25.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+PYTHON_COMPAT=( python3_{8,9} )
+
```
ionenwks: Hm, seems like upstream added it to rdepend for gns3-gui too, so now we get:
```
 *   have:     DISTUTILS_USE_SETUPTOOLS=bdepend (or unset)
 *   expected: DISTUTILS_USE_SETUPTOOLS=rdepend
```
https://github.com/GNS3/gns3-gui/commit/67014965bee0355aaf66d8725c9e8819c86b99db

mm1ke: Thanks for pointing this out. Completely oversaw this. Should be fixed now.

ionenwks (approved):

PR #22327, opened 2021-09-18
sys-cluster/charliecloud: Bump to version 0.25., drop old.

olifre:
This unbundles `dev-python/lark-parser` which is now bundled upstream by default, and adds python 3.10 compatibility.


juippis (approved):
I can't build the -9999 version, oddly looks like a gcc issue. But 0.25 works, thanks!

olifre:
@juippis Thanks! 
The issue with the `-9999` version is fixed with:
https://github.com/hpc/charliecloud/pull/1181
It shows up with gcc 10 only, the fix is under review by upstream already :smile:.

reidpr:
Regarding the Lark bundling, our plan is to tighten supported versions of Lark significantly, perhaps down to a single version. (There are new features we want to use, and we want to remove workarounds for long-fixed bugs.) `--disable-bundled-lark` won't go away, but it might become harder to satisfy the Lark dependency externally. Happy to coordinate on this; we do want to minimize pain for packagers.

PR #22331, opened 2021-09-18
games-emulation/desmume: fix build with clang

DarthGandalf:
Closes: https://bugs.gentoo.org/739144


ionenwks (approved):
Thanks, it's something I should've fixed myself back then but forgot about.

Albeit it'll be fixed whenever do another snapshot from https://github.com/TASVideos/desmume/commit/92cb90f43322e64fb614f43a5073bd3a52f19e89

Upstream was semi-inactive for a while, and think I'll wait a bit longer before considering another snapshot -- so your simpler fix will do :)

PR #22333, opened 2021-09-19
sci-geosciences/grass: fix install version 9999 (changes from 7.9 dev version to 8.0)

tmszi:
[Bugfix](https://bugs.gentoo.org/813879).

### DIFF
--- a/sci-geosciences/grass/grass-9999.ebuild
+++ b/sci-geosciences/grass/grass-9999.ebuild
@@ -216,7 +216,7 @@ src_install() {
 		sed -i "s|${ED}|/|g" "${scriptMakeDir}/${file}" || die
 	done
 
-	mv ${D}/usr/bin/grass ${D}/usr/bin/${MY_PM}
+	mv ${D}/usr/bin/grass ${D}/usr/bin/${MY_PM} || die
 
 	# get proper folder for grass path in script
 	local gisbase=/usr/$(get_libdir)/${MY_PM}
--- a/sci-geosciences/grass/grass-9999.ebuild
+++ b/sci-geosciences/grass/grass-9999.ebuild
@@ -5,7 +5,7 @@ EAPI=7
 
 PYTHON_COMPAT=( python3_{8,9} )
 PYTHON_REQ_USE="sqlite"  # bug 572440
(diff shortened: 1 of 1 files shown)


sci-geosciences/grass/grass-9999.ebuild
```diff
@@ -216,6 +216,8 @@ src_install() {
 		sed -i "s|${ED}|/|g" "${scriptMakeDir}/${file}" || die
 	done
 
+	mv ${D}/usr/bin/grass ${D}/usr/bin/${MY_PM}
```
juippis, line 219: Should `|| die` here.

sci-geosciences/grass/grass-9999.ebuild
```diff
@@ -3,14 +3,14 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{6..9} )
+PYTHON_COMPAT=( python3_{8,9} )
 PYTHON_REQ_USE="sqlite"  # bug 572440
 WANT_AUTOCONF="2.1"
```
juippis, line 8: This fails with 
```
***** autoconf *****
***** PWD: /var/tmp/portage/sci-geosciences/grass-9999/work/grass-9999
***** autoconf

ac-wrapper: Since configure.ac is present, aclocal always use
    autoconf 2.59+, which conflicts with your choice and
    causes error. You have two options:
    1. Try execute command again after removing configure.ac
    2. Don't set WANT_AUTOCONF
```

sci-geosciences/grass/grass-9999.ebuild
tmszi: Thanks. I've already fixed it.

sci-geosciences/grass/grass-9999.ebuild
tmszi: Thanks I overlooked it. GRASS GIS upgrade [autoconf version to 2.69 ](https://github.com/OSGeo/grass/commit/dca9364c64d7dac5b4ae9262f016708a97c979d7) in last commits. I've already fixed it.

thesamesam:
Thanks! I fixed up the commit messages to have the right prefix.

PR #22334, opened 2021-09-19
games-puzzle/amoebax: EAPI8, minor improvements

mm1ke:
Package-Manager: Portage-3.0.23, Repoman-3.0.3

@ionenwks 
This should be a pretty simply `EAPI=8` conversation. 
However, there is a bug open for this game (bug #739286). I've looked into it and i _guess_ it's not so difficult to fix. However, I'm not a programmer, and I have no clue how to fix this correctly.

### DIFF
--- a/games-puzzle/amoebax/amoebax-0.2.1-r2.ebuild
+++ b/games-puzzle/amoebax/amoebax-0.2.1-r2.ebuild
@@ -1,40 +1,40 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=8
+
 inherit autotools
 
-DESCRIPTION="A cute and addictive action-puzzle game, similar to tetris"
-HOMEPAGE="http://www.emma-soft.com/games/amoebax/"
-SRC_URI="http://www.emma-soft.com/games/amoebax/download/${P}.tar.bz2"
+DESCRIPTION="Cute and addictive action-puzzle game, similar to tetris"
+HOMEPAGE="https://www.emma-soft.com/games/amoebax/"
+SRC_URI="https://www.emma-soft.com/games/amoebax/download/${P}.tar.bz2"
 
 LICENSE="Free-Art-1.2 GPL-2+"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE=""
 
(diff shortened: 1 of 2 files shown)


ionenwks:
> However, there is a bug open for this game (bug #739286). I've looked into it and i _guess_ it's not so difficult to fix. However, I'm not a programmer, and I have no clue how to fix this correctly.

I don't really do C++ myself but it's indeed simple. Most clang issues on old games tend to either missing return values (often caused by a function that should've been declared as void), or narrowing in { list-initialization } without explicit cast (in this case { toupper() } returns an integer being narrowed to a smaller char type) and, while gcc only warns about these poor practices, clang made this a hard failure.

That aside, when unsure, a good place to find already-made clang fixes is freebsd
https://www.freshports.org/games/amoebax
https://cgit.freebsd.org/ports/tree/games/amoebax/files/patch-src_NewHighScoreState.cxx

Feel free to add it :) Very trivial things like this don't need attribution too, -clang.patch with buglink is good enough.

mm1ke:
> > However, there is a bug open for this game (bug #739286). I've looked into it and i _guess_ it's not so difficult to fix. However, I'm not a programmer, and I have no clue how to fix this correctly.
> 
> I don't really do C++ myself but it's indeed simple. Most clang issues on old games tend to either missing return values (often caused by a function that should've been declared as void), or narrowing in { list-initialization } without explicit cast (in this case { toupper() } returns an integer being narrowed to a smaller char type) and, while gcc only warns about these poor practices, clang made this a hard failure.
> 
> That aside, when unsure, a good place to find already-made clang fixes is freebsd
> https://www.freshports.org/games/amoebax
> https://cgit.freebsd.org/ports/tree/games/amoebax/files/patch-src_NewHighScoreState.cxx
> 
> Feel free to add it :) Very trivial things like this don't need attribution too, -clang.patch with buglink is good enough.

Thanks for checking. Looking at other distros in general it probably a good idea. :)
I've added the pach, but i haven't checked it with clang. Can you maybe test it? (i really have to setup a vm with a clang enviroment.. )

ionenwks (approved):
> I've added the patch, but i haven't checked it with clang. Can you maybe test it? (i really have to setup a vm with a clang enviroment.. )

Yep, builds with clang++ just fine now, thanks.

PR #22335, opened 2021-09-19
gui-wm/sway: introduce grimshot: utility for taking screenshots

mm1ke:
Hi,

This PR introduces the `grimshot` utility which comes with the `sway` package. This wasn't installed yet so i've modified the ebuild to install it. It's a simply bash script which depends on a view other tools in order to work (which is why i didn't enabled it by default).
Unfortunately I also had to remove `ppc64 riscv x86` keywords because some of the tools miss keywords.

I've also checked the `meson.build` file in order to install the script via meson, but as far as I could see there is no option for `grimshot`.

Please review.

### DIFF
--- a/gui-wm/sway/sway-1.7-r1.ebuild
+++ b/gui-wm/sway/sway-1.7-r1.ebuild
@@ -14,7 +14,7 @@ if [[ ${PV} == 9999 ]]; then
 else
 	MY_PV=${PV/_rc/-rc}
 	SRC_URI="https://github.com/swaywm/${PN}/archive/${MY_PV}.tar.gz -> ${P}.tar.gz"
-	KEYWORDS="~amd64 ~arm64"
+	KEYWORDS="~amd64 ~arm64 ~loong ~ppc64 ~riscv ~x86"
 	S="${WORKDIR}/${PN}-${MY_PV}"
 fi
 
@@ -37,6 +37,11 @@ DEPEND="
 	swaybg? ( gui-apps/swaybg )
 	swayidle? ( gui-apps/swayidle )
 	swaylock? ( gui-apps/swaylock )
+	tray? ( || (
+		sys-apps/systemd
+		sys-auth/elogind
+		sys-libs/basu
+	) )
 	wallpapers? ( x11-libs/gdk-pixbuf:2[jpeg] )
 	X? ( x11-libs/libxcb:0= )
 "
@@ -69,6 +74,7 @@ if [[ ${PV} == 9999 ]]; then
 else
 	BDEPEND+="man? ( >=app-text/scdoc-1.9.3 )"
 fi
+REQUIRED_USE="grimshot? ( swaymsg )"
 
 src_configure() {
 	local emesonargs=(
@@ -82,7 +88,6 @@ src_configure() {
 		-Dfish-completions=true
 		-Dzsh-completions=true
 		-Dbash-completions=true
-		-Dwerror=false
 	)
 
 	meson_src_configure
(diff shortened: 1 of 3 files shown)


ticoombs:
Isn't this provided by `gui-apps/grim` ?

mm1ke:
> Isn't this provided by `gui-apps/grim` ?

Not exactly. 
`grimshot` is more like an wrapper to `grim` (thus the dependency on it). However `grimshot` has more possibilities, like creating an screenshot from only an area of the screen by simply using the mouse to mark the area you want.

ticoombs:
i utilise slurp with grim

```
set $ps2 Control+Print
bindsym $ps2 exec grim -g "$(slurp)" - | wl-copy
```
for screen based capture

gui-wm/sway/sway-9999.ebuild
```diff
@@ -14,13 +14,13 @@ if [[ ${PV} == 9999 ]]; then
 else
 	MY_PV=${PV/_rc/-rc}
 	SRC_URI="https://github.com/swaywm/${PN}/archive/${MY_PV}.tar.gz -> ${P}.tar.gz"
-	KEYWORDS="~amd64 ~arm64 ~ppc64 ~x86"
```
prometheanfire, line 17: keywords should remain the same unless there's a reason for removal (which should probably be it's own review)

prometheanfire (changes requested):
Should the version changed be 1.7-r1?

Also, keywords should be `~amd64 ~arm64 ~ppc64 ~riscv ~x86` unless there's a good reason for removal (which should probably be it's own commit if needed).

gui-wm/sway/sway-9999.ebuild
prometheanfire: oh, old because it's an old request, sorry for not noticing

mm1ke: Hi @prometheanfire 
No problem, i've rebased my PR and reverted the dropped `KEYWORDS` for the live ebuild.

gui-wm/sway/sway-1.7-r1.ebuild
```diff
@@ -0,0 +1,98 @@
...
+DESCRIPTION="i3-compatible Wayland window manager"
+HOMEPAGE="https://swaywm.org"
+
+if [[ ${PV} == 9999 ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/swaywm/${PN}.git"
+else
+	MY_PV=${PV/_rc/-rc}
+	SRC_URI="https://github.com/swaywm/${PN}/archive/${MY_PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64 ~arm64"
```
prometheanfire, line 17: Is there a reason for the dropped keywords?  We shouldn't drop keywords if we can help it.

mm1ke: Well, the new dependencies are not (yet) keyworded for these archs.
For example: `gui-apps/grim` is only `~amd64 ~arm64` keyworded. `gui-apps/slurp` is only keyworded for `~amd64 ~arm64 ~x86`

I guess a better approch would be to use.mask `grimshot` for these archs for the meantime?

prometheanfire: nah, better would be to get them keyworded so we can add them here. (just did so)

mm1ke: well that's of course even better :)
We still need the keywords for `gui-apps/slurp`, then i should be able to restore all keywords (until then, repoman won't be happy)

prometheanfire:
thanks for the update, merged

PR #22353, opened 2021-09-21
media-gfx/inkscape: bump to 1.1 & sync live

arkamar:
This PR synces live ebuild with upstream and bunps inkscape to 1.1, where I also backported patch from upstream in order to support musl.

I enabled testing in live ebuild, where it works for me, but it is still restricted for 1.1 because few tests fail and it would be necessary to backport multiple quite big commits to make it work, which didn't make sense to me.

I am not sure if https://bugs.gentoo.org/show_bug.cgi?id=776265 should be closed as well.


arkamar:
FP where `~riscv` is restored because it is present in `1.0.2-r2`, but I removed `~hppa` because `app-text/gspell` dependency does not have it as well.

arkamar:
FP with dropped `~ia64` keyword bacause of the same reason as above, `app-text/gspell` dependency does not have it.

media-gfx/inkscape/inkscape-1.1.ebuild
```diff
@@ -0,0 +1,162 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{8..9} )
```
mikhailkoliada, line 6: Do you really need an interval for only 2 python versions here?

mikhailkoliada (changes requested):

media-gfx/inkscape/inkscape-1.1.ebuild
arkamar: No, I don't need it. This used more often than `{8,9}` in `::gentoo`.

arkamar: Anyway, do you want me to change it?

arkamar: Changed to `{8,9}` in recent FP.

PR #22361, opened 2021-09-21
Add ISOBMFF build option via package use flag

goeland86:
Submitting PR to provide a cleaner way of tracing the submission of the updated exiv2 build with ISOBMFF for CR3, AVIF  and HEIF support option. Bugzilla ticket is here: https://bugs.gentoo.org/814212

### DIFF
--- a/media-gfx/exiv2/exiv2-0.27.4-r3.ebuild
+++ b/media-gfx/exiv2/exiv2-0.27.4-r3.ebuild
@@ -1,0 +1,116 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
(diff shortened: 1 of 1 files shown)


goeland86:
Oh boy. I was hoping using the Github editor to upload my ebuild file would be easier. But it really isn't. And I don't want to clone the entire project for one new file to add - I get the need for the sign-offs, but it's a bit convoluted to make them properly when you're in a hurry late at night.

thesamesam:
That's OK, I've got the bit I need (the sign off), so I can do the git mangling if necessary this time ;).

Thanks!

goeland86:
Much appreciated! I'll try to do better next time!

thesamesam:
Thanks! Done as https://github.com/gentoo/gentoo/commit/a0a701d1456974ba5e8ce8b69809889c2c569271.

PR #22363, opened 2021-09-22
app-emulation/xen-tools: fix building with gcc-11

hydrapolic:
The patch is not needed any more. Tested with gcc-10 and gcc-11.


juippis (approved):

hydrapolic:
Thanks @juippis

juippis:
As always, you too!

PR #22366, opened 2021-09-22
net-im/discord-bin: Version bump to 0.0.16

owenwastaken:
Old version does not launch and prompts the user to update


thesamesam:
Beaten to it in https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=e108c0e59ae70b6abe3bec5fd62f8d1b3fa1d519, but thank you!

PR #22368, opened 2021-09-23
dev-libs/jansson: version bump to 2.14

Zer0-One:
Also bumps EAPI to 8, and removes the sphinx patch.

Signed-off-by: David Zero <zero-one@zer0-one.net


thesamesam:
Needs a Closes tag for the bug and to bump (add) subslot.

Will need to handle all consumers too but I’ll do that side.

Zer0-One:
Added slotting and a `Closes` tag to the commit.

dev-libs/jansson/jansson-2.14.ebuild
```diff
@@ -0,0 +1,36 @@
...
+EAPI=8
+
+inherit multilib-minimal
+
+DESCRIPTION="C library for encoding, decoding and manipulating JSON data"
+HOMEPAGE="https://www.digip.org/jansson/"
+SRC_URI="https://github.com/akheron/jansson/releases/download/v${PV}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="2/4"
```
thesamesam, line 13: We need to add a [subslot](https://devmanual.gentoo.org/general-concepts/slotting/index.html#sub-slots) to force rebuilding, but no need to change the main slot (0->2).

PR #22370, opened 2021-09-23
media-video/shotcut: bump to 21.09.20

xdch47:


media-video/shotcut/shotcut-21.09.20.ebuild
```diff
@@ -0,0 +1,66 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
juippis: 8 should work, I'll add it. Thanks!

juippis (approved):

PR #22379, opened 2021-09-23
media-gfx/prusaslicer: version bump to 2.3.3

kobboi:
Dropping the gcc-11 patch, which is no longer necessary


kobboi:
@alexxy / @expeditioneer  FYI

PR #22381, opened 2021-09-24
app-crypt/coolkey: don't parallel install

kobboi:
Something fishy going on in docs/Makefile

	install-data-hook:
		rm -f $(DESTDIR)$(libdir)/libckyapplet.a
		rm -f $(DESTDIR)$(libdir)/libckyapplet.la

but agreed with sam that it's not worth it to debug further
(seeing that upstream is dead and the src_compile is being
done -j1 as well.)

- Override src_install with its default but using '-j1'
- Update the copyright


kobboi:
@thesamesam for your review/feedback

juippis (approved):
/var/tmp/portage/vbslogs/build/app-crypt/coolkey-1.1.0-r7:20210924-071419.log:./configure: line 19940: enable_pk11install: command not found

Might be worth checking.

PR #22387, opened 2021-09-24
app-crypt/coolkey: fix configure error and copyright

kobboi:
One of the patches in the patch tarball (0004-Bug-fixups.patch)
addresses an error in configure.in, but without a call to
autoreconf, the configure script remained unchanged.

Closes: https://bugs.gentoo.org/745285


kobboi:
@thesamesam FYI, this PR addresses the comment by @juippis in PR #22381

PR #22389, opened 2021-09-24
app-editors/scite: Version Bump to 5.1.3.

dermoench42:
Actual enhancements and upstream bugfixes. See
http://www.scintilla.org/ScintillaHistory.html

Signed-off-by: Ervin Peters coder@ervnet.de


app-editors/scite/scite-5.1.3.ebuild
```diff
@@ -0,0 +1,115 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+LUA_COMPAT=( lua5-1 )
```
juippis, line 6: I wonder if we can start testing other newer lua implementations?

juippis (approved):
Thanks!

PR #22392, opened 2021-09-24
app-admin/ngxtop - add py3.10

hydrapolic:
Tested on ~amd64.


app-admin/ngxtop/ngxtop-0.0.3_pre141201-r2.ebuild
```diff
@@ -0,0 +1,28 @@
...
+
+PYTHON_COMPAT=( python3_{7..10} )
+PYTHON_REQ_USE="sqlite"
+
+inherit distutils-r1
+
+DESCRIPTION="real-time metrics for nginx server (and others)"
+HOMEPAGE="https://github.com/lebinh/ngxtop"
+#SRC_URI="https://github.com/lebinh/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+SRC_URI="https://dev.gentoo.org/~jlec/distfiles/${P}.tar.xz"
```
juippis: Huh?

juippis (approved):
This doesn't really need a revbump, but I'll use the opportunity to bump this into EAPI-8, and drop python-3.7 support.

app-admin/ngxtop/ngxtop-0.0.3_pre141201-r2.ebuild
hydrapolic: Custom patches IIRC

hydrapolic:
Thanks @juippis

PR #22397, opened 2021-09-25
bzr.eclass

ulm:


arthurzam:
One serious issue we have is the support for sftp in `dev-vcs/breezy`:
I have set it as optfeature, meaning no PYTHON_TARGET checking is done on the ebuild level. Meaning if we want to correctly update the `BDEPEND` in `bzr.eclass` when an sftp BZR_URI given, we would need to use a lot of checking for python compatible target is found (we need a match).

As a cleaner and simpler solution, I can add `IUSE="sftp"` which depends on those deps, and just use `BDEPEND="dev-vcs/breezy[sftp]"` (as in old eclass).
The only issue is that we have created a pure runtime USE flag, which violated QA policy. The build time is quite low (around 2 minutes per python target, on an old PC), so not the worst case.

@ulm: So before I do this simple change (adding `sftp` USE flag) - I want to recheck - is it needed?

(When I wrote the ebuild, never thought it would return an eclass based on it, which affected my mindset how I wrote it. Thank you for all your work.)

---

On another note - should we try to keyword `dev-vcs/breezy` on main other arches, before committing this eclass? As of now, I just know it passes all tests on my main system and containers (amd64 & virtual x86), but no idea for other arches and setups.

ulm:
> One serious issue we have is the support for sftp in `dev-vcs/breezy`:
> I have set it as optfeature, meaning no PYTHON_TARGET checking is done on the ebuild level. Meaning if we want to correctly update the `BDEPEND` in `bzr.eclass` when an sftp BZR_URI given, we would need to use a lot of checking for python compatible target is found (we need a match).
> 
> As a cleaner and simpler solution, I can add `IUSE="sftp"` which depends on those deps, and just use `BDEPEND="dev-vcs/breezy[sftp]"` (as in old eclass).
> The only issue is that we have created a pure runtime USE flag, which violated QA policy. The build time is quite low (around 2 minutes per python target, on an old PC), so not the worst case.
> 
> @ulm: So before I do this simple change (adding `sftp` USE flag) - I want to recheck - is it needed?
> 
> (When I wrote the ebuild, never thought it would return an eclass based on it, which affected my mindset how I wrote it. Thank you for all your work.)

I believe that no action is needed for now. We can think about it when an ebuild will appear that needs it. TBH, I don't expect `bzr.eclass` to have more than a handful of ebuilds using it.

> On another note - should we try to keyword `dev-vcs/breezy` on main other arches, before committing this eclass? As of now, I just know it passes all tests on my main system and containers (amd64 & virtual x86), but no idea for other arches and setups.

IMHO that's in the category "nice to have" but not strictly necessary for now. Old `dev-vcs/bzr` had the following keywords, BTW. :smile:
```
KEYWORDS="~alpha amd64 arm arm64 hppa ~ia64 ~mips ppc ppc64 ~s390 sparc x86 ~amd64-linux ~x86-linux ~ppc-macos ~x64-macos ~x86-macos ~m68k-mint ~sparc-solaris"
```

PR #22400, opened 2021-09-25
dev-libs/pigpio: fix broken Makefile

github-k8n:
fix broken Makefile, missing linked lib

Signed-off-by: Daniel Kenzelmann <gentoo@k8n.de>


github-k8n:
Bug: https://bugs.gentoo.org/811150

juippis (approved):
Can this be upstreamed? 

(can't test btw, but I don't see anyone merging either)

Thanks!

github-k8n:
Title might be misleading, this is fixing the Makefile that was accidentally broken in previous commits (i didn't notice at first because i was re-emerging it to test and not unmerge it before thus it did not fail when including the library..
upstream was not affected ..

PR #22401, opened 2021-09-25
app-emulation/qemu: add PYTHON_USEDEP for sphinx bdeps

ajakk:
Bug: https://bugs.gentoo.org/814650
Signed-off-by: John Helmert III <ajak@gentoo.org>

I wasn't actually able to reproduce this bug, but this would seem to be the obvious solution to it.


juippis:
Why the revbump though?

ajakk:
Because it's early and I didn't fully think about this only affecting bdeps :P

juippis:
27688b840d6c3a8cbca0b6f644364fb1e3c51171
\o/

@tamiko you can use [pram](https://packages.gentoo.org/packages/app-portage/pram) to merge these PRs and it'll close them automatically :)

PR #22403, opened 2021-09-25
app-editors/kakoune: Update download location. Bump to 2021.08.28-r1

mujo-hash:
The release download contains an extra .version file in the source for inserting version information at runtime.
Instead of patching, just switch the download location.


ionenwks (approved):
Ah yes, it's better to use /releases/ tarballs when offered.

I was going to nag about needing to rename the distfile, but gz->bz2 solved that already :)

PR #22407, opened 2021-09-26
app-vim/fugitive: version bump to 3.4

zippy2:


zippy2:
CC @radhermit it would be nice if we could stabilize some newer version. The current one (v2.4) is ~3 years old.

PR #22416, opened 2021-09-26
media-gfx/qimgv: version bump to 1.0.2 and drop 0.9.1

kamiyaa:


media-gfx/qimgv/qimgv-1.0.1.ebuild
```diff
@@ -0,0 +1,58 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
juippis: 8 should work here.

media-gfx/qimgv/qimgv-1.0.1.ebuild
```diff
@@ -0,0 +1,58 @@
...
+
+pkg_postinst() {
+	xdg_icon_cache_update
+	xdg_desktop_database_update
+}
+
+pkg_postrm() {
+	xdg_icon_cache_update
+	xdg_desktop_database_update
+}
```
juippis: If you simply inherit *xdg.eclass* you can remove these parts.

media-gfx/qimgv/qimgv-1.0.1.ebuild
```diff
@@ -0,0 +1,58 @@
...
+
+EAPI=7
+
+inherit cmake xdg-utils
+
+DESCRIPTION="A cross-platform image viewer with webm support, written in qt5"
+HOMEPAGE="https://github.com/easymodo/qimgv"
+SRC_URI="https://github.com/easymodo/qimgv/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3"
```
juippis: Now I noticed some files are licensed under BSD... so is this GPL or BSD, or mixed? Could you investigate?
https://github.com/easymodo/qimgv/blob/master/qimgv/gui/flowlayout.cpp

kamiyaa: That was just some example code from qt examples.
The project owner said they would most likely rewrite it later on to make it completely GPL.
Does this still warrant a change to the license for this ebuild?

juippis: Yeah, let's add BSD for now. Also:
```
 * QA: following sed did not cause any changes
 *     sed -i -e '/set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -lstdc++fs -O3")/d' CMakeLists.txt || die
```

also:
```
CMake Error at /usr/lib64/cmake/Qt5/Qt5Config.cmake:28 (find_package):
  Could not find a package configuration file provided by "Qt5Svg" with any
  of the following names:
```

also:
```
CMake Error at /usr/lib64/cmake/Qt5/Qt5Config.cmake:28 (find_package):
  Could not find a package configuration file provided by "Qt5PrintSupport"
  with any of the following names:

    Qt5PrintSupportConfig.cmake
    qt5printsupport-config.cmake
```

kamiyaa:
- Added `xdg` class to inherit
  - removed `pkg_postinst` and `pkg_postrm`
- Update to EAPI 8
- Added BSD license
- Add `qtprintsupport` and `qtsvg` as dependencies
- Remove sed command

juippis (approved):
Thanks!

PR #22419, opened 2021-09-27
app-backup/dar: add v2.7.2

ceamac:
Minor update, bugfixes.

No change in ebuild, only rename and set keywords to unstable.


app-backup/dar/dar-2.7.2.ebuild
```diff
@@ -0,0 +1,90 @@
...
+		$(usex dar32 --enable-mode=32 '')
+		$(usex dar64 --enable-mode=64 '')
+		$(usex doc '' --disable-build-html)
+		$(usex gcrypt '' --disable-libgcrypt-linking)
+		$(usex gpg '' --disable-gpgme-linking)
+		$(usex lzo '' --disable-liblzo2-linking)
+		$(usex nls '' --disable-nls)
+		$(usex rsync '' --disable-librsync-linking)
+		$(usex threads '' --disable-threadar)
+		$(usex xattr '' --disable-ea-support)
```
ionenwks, line 72: Looking at this makes me want to ask you to bump this to EAPI-8 for its `usev` 2nd argument[1]
e.g.
`$(usex dar64 --enable-mode=64 '')` becomes `$(usev dar64 --enable-mode=64)`
`$(usex nls '' --disable-nls)` becomes `$(usev !nls --disable-nls)`

[1] https://mgorny.pl/articles/the-ultimate-guide-to-eapi-8.html#usev-now-accepts-a-second-argument

app-backup/dar/dar-2.7.2.ebuild
```diff
@@ -0,0 +1,90 @@
...
+
+BDEPEND="
+	doc? ( app-doc/doxygen )
+	nls? (
+		sys-devel/gettext
+		virtual/libintl
+	)
+"
+
+DOCS=( AUTHORS ChangeLog NEWS README THANKS TODO )
```
ionenwks, line 47: Don't need this line, the default DOCS is:
```
README* ChangeLog AUTHORS NEWS TODO CHANGES THANKS BUGS \
FAQ CREDITS CHANGELOG
```
And so will install all of these.

app-backup/dar/dar-2.7.2.ebuild
```diff
@@ -0,0 +1,90 @@
...
+	xattr? ( sys-apps/attr:= )
+"
+
+DEPEND="${RDEPEND}"
+
+BDEPEND="
+	doc? ( app-doc/doxygen )
+	nls? (
+		sys-devel/gettext
+		virtual/libintl
```
ionenwks, line 43: virtual/libintl doesn't make sense in BDEPEND, should be behind nls? in RDEPEND+DEPEND

Not that libintl means much in 95% of cases given it's builtin most libcs.

app-backup/dar/dar-2.7.2.ebuild
```diff
@@ -0,0 +1,90 @@
...
+	>=sys-libs/zlib-1.2.3:=
+	argon2? ( app-crypt/argon2 )
+	curl? ( net-misc/curl )
+	gcrypt? ( dev-libs/libgcrypt:0= )
+	gpg? ( app-crypt/gpgme )
+	lzo? ( dev-libs/lzo:= )
+	rsync? ( net-libs/librsync:= )
+	threads? ( dev-libs/libthreadar:= )
+	xattr? ( sys-apps/attr:= )
+"
```
ionenwks, line 35: My app-portage/iwdevtools are nagging
```
 * QA: detected possibly incorrect RDEPEND (app-backup/dar-2.7.2)
 * app-arch/xz-utils:=    | app-arch/xz-utils
 *                        > app-arch/zstd:=
 * app-crypt/argon2       | app-crypt/argon2:=
 * app-crypt/gpgme        | app-crypt/gpgme:1=
 *                        > dev-libs/libgpg-error
 * dev-libs/libthreadar:= | dev-libs/libthreadar
 * dev-libs/lzo:2=        | dev-libs/lzo:2
 * sys-apps/attr:=        | sys-apps/attr
```
Basically saying:
1. some of these := aren't necessary (no subslot), harmless but customary not to define if unneeded
2. directly linked with libzstd but no dependency, this looks like automagic so needs a USE for `--disable-libzstd-linking`
3. directly linked with libgpg-error but not explicitly listed, this is only with USE=gcrypt (not that it's a big issue considering gcrypt depends on this, but that's a rabbit hole better avoided).
4. argon2 and gpgme define a subslot but here the := is missing, the 1 in :1= isn't *essential* in this case but if want to be pedantic (given it defines a slot).

app-backup/dar/dar-2.7.2.ebuild
```diff
@@ -0,0 +1,90 @@
...
+
+EAPI=7
+
+inherit flag-o-matic
+
+DESCRIPTION="A full featured backup tool, aimed for disks"
+HOMEPAGE="http://dar.linux.free.fr/"
+SRC_URI="mirror://sourceforge/dar/${P}.tar.gz"
+
+LICENSE="GPL-2"
```
ionenwks, line 12: This looks like a `GPL-2+`
```
// as published by the Free Software Foundation; either version 2
// of the License, or (at your option) any later version.
```

app-backup/dar/dar-2.7.2.ebuild
```diff
@@ -0,0 +1,90 @@
...
+
+	econf "${myconf[@]}"
+}
+
+src_install() {
+	emake DESTDIR="${D}" pkgdatadir="${EPREFIX}"/usr/share/doc/${PF}/html install
+
+	einstalldocs
+
+	find "${ED}" \( -name "*.a" -o -name "*.la" \) -delete || die
```
ionenwks, line 86: I don't see any .a being installed even without this (likely thanks to --disable-static), but the .la remains.

Could simplify:
```suggestion
	find "${ED}" -name '*.la' -delete || die
```

app-backup/dar/dar-2.7.2.ebuild
```diff
@@ -0,0 +1,90 @@
...
+	# configure.ac is totally funked up regarding the AC_ARG_ENABLE
+	# logic.
+	# For example "--enable-dar-static" causes configure to DISABLE
+	# static builds of dar.
+	# Do _not_ use $(use_enable) until you have verified that the
+	# logic has been fixed by upstream.
+	local myconf=(
+		--disable-dar-static
+		--disable-python-binding
+		--disable-static
```
ionenwks, line 59: EAPI-8 pass this one by default[1], so you can remove after eapi bump

[1] https://mgorny.pl/articles/the-ultimate-guide-to-eapi-8.html#new-econf-passed-options

ionenwks (commented):
> Minor update, bugfixes.
> 
> No change in ebuild, only rename and set keywords to unstable.

Sounds like a nice and safe time to actually do changes :eyes:

app-backup/dar/dar-2.7.2.ebuild
ceamac: Nice tool, thanks.  I've used it and fixed the libs.  For gpg I also had to add `dev-libs/libassuan`, also a dependency of `app-crypt/gpgme`.

I also added **zstd** enabled by default, without a USE flag (it's been a pull request about this [here](https://github.com/gentoo/gentoo/pull/21558) closed and dropped by now, but it seemed like a good idea).  But I added a flag for **lz4** instead - it was automagically detected too.

ionenwks: Good thinking wrt zstd USE. I agree with it but wasn't really thinking about it when I suggested.

libgpg-error wasn't for gpg (well not only for it), as I mentioned gcrypt uses it:
```
[ebuild  N     ] app-backup/dar-2.7.2::gentoo  USE="gcrypt -argon2 -curl -dar32 -dar64 -doc -gpg -lz4 -lzo -nls -rsync -threads -xattr"
<snip>
 * QA: detected possibly incorrect RDEPEND (app-backup/dar-2.7.2)
 *  > dev-libs/libgpg-error
```
gpg may need it as well but given there's a REQUIRED_USE it only needs to be listed once behind `gcrypt?`

wrt libassuan, it doesn't actually use it and it's just a bad inject by gpgme discarded by -Wl,--as-needed, in these case I'd say it's okay to not list it. If say.. gpgme ever stopped using it and libassuan is depcleaned, it won't break dar. While for libgpg-error you'd get an unresolved library and wouldn't startup anymore without preserve-libs.
```
[ebuild   R    ] app-backup/dar-2.7.2::gentoo  USE="gcrypt gpg -argon2 -curl -dar32 -dar64 -doc -lz4 -lzo -nls -rsync -threads -xattr"
<snip>
 * QA: detected possibly incorrect RDEPEND (app-backup/dar-2.7.2)
 * dev-libs/libassuan < 
```
Or was it actually in NEEDED for you? I guess it's possible there's some other condition.

ceamac: I don't have --as-needed.  It looks like this:
`LDFLAGS="-Wl,--defsym=__gentoo_check_ldflags__=0"`

But if I added it libassuan went away.

I'll remove libassuan and move `dev-libs/libgpg-error` on `gcrypt?`

ionenwks: Ah I see, it'd be good to use LDFLAGS="${LDFLAGS} -Wl,--defsym=__gentoo_check_ldflags__=0" to keep the defaults.

ionenwks (approved):

PR #22425, opened 2021-09-27
mail-client/neomutt: Add use flags for header cache compression

jan-krieg:
Neomutt upstream supports header cache compression via lz4, zlib, and zstd since release 2020-03-13. However, the current ebuilds don't offer the relevant use flags to activate this feature. Adding lz4, zlib, and zstd use flags works well out-of-the-box on my system (tested with all three compression algorithms on amd64 for both 20210205-r1 and 99999999, using the gdbm backend).

release notes: https://neomutt.org/2020/03/13/release
docs: https://neomutt.org/guide/optionalfeatures.html#8-1-%C2%A0header-caching
upstream pull request: https://github.com/neomutt/neomutt/pull/1922

Closes: https://bugs.gentoo.org/815187
Signed-off-by: Jan Krieg <git@jankrieg.anonaddy.com>

### DIFF
--- a/mail-client/neomutt/metadata.xml
+++ b/mail-client/neomutt/metadata.xml
@@ -27,6 +27,9 @@
 			traditional/inline PGP</flag>
 		<flag name="smime-classic">Build classic-smime backend to support
 			S/MIME</flag>
+		<flag name="lz4">Add lz4 support for header cache compression</flag>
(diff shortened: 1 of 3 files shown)


nicolasbock:
Thanks @jan-krieg !

PR #22426, opened 2021-09-27
app-crypt/tpm2-pkcs11: Bump to 1.7.0

salahcoronya:
The tpm2_pool utility gains a new operation, Import but requires >=app-crypt/tpm2-tool-5.2 (Not released yet). 

I'd like to enable more of the tests, but there's a few problems doing so:

1) Requires dev-python/python-pkcs11 - not in portage 
2) Requires >=app-crypt/tpm2-tool-5.2 - not released yet
3) One of test downloads stuff off the Internet, so it should be patched out (or fixed, if possible)
4) But the main issue issue one of the tests is written in Java. None of the java eclasses support a build-only dependency to Java.


juippis (approved):
> Requires dev-python/python-pkcs11 - not in portage
> 

If you find it worth, you can add that when next modifying this package somehow.

> One of test downloads stuff off the Internet, so it should be patched out (or fixed, if possible)
>

Yes, if possible, you should pre-download the required files with `test? ( ... )` USE conditional.

> But the main issue issue one of the tests is written in Java. None of the java eclasses support a build-only dependency to Java.

I don't know java at all, but couldn't you just straight up DEPEND/BDEPEND on some java?

PR #22429, opened 2021-09-28
Security updates to www-apps/redmine

winterheart:
Addressed issues CVE-2021-31863, CVE-2021-31864, CVE-2021-31865, CVE-2021-31866, CVE-2021-37156

Bug: https://bugs.gentoo.org/786561
Bug: https://bugs.gentoo.org/806842


dev-ruby/rotp/metadata.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+<maintainer type="project">
```
thesamesam, line 4: We don't want to add ruby@ to stuff without their permission, so I'll drop that bit.

dev-ruby/rotp/rotp-6.2.0.ebuild
```diff
@@ -0,0 +1,31 @@
...
+
+DESCRIPTION="The Ruby One Time Password Library"
+HOMEPAGE="https://github.com/mdp/rotp"
+SRC_URI="https://github.com/mdp/rotp/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="dev-libs/openssl:0="
```
thesamesam, line 22: Needs to be in DEPEND too for := to be effective.

thesamesam:
Thank you!

PR #22430, opened 2021-09-28
net-irc/weechat: Bump to weechat-3.3

qiankehan:
Signed-off-by: Han Han <hhan@redhat.com>


gyakovlev:
this PR had issues, but I noticed it a bit later and committed my version of bump.

the patch failed to apply, so it needed a new version.
also upstream added new `typing` plugin, I've added that, default-enabled. it does not have any external deps.

thanks anyway and sorry for delay.

PR #22432, opened 2021-09-28
app-crypt/ima-evm-utils: Fix tests

salahcoronya:
This adds the missing test dependency ( app-editors/vim-core ).

The skipped test will remain skipped as it requires the IBM TSS2 tools which are not in Portage.


app-crypt/ima-evm-utils/ima-evm-utils-1.3.2.ebuild
```diff
@@ -12,15 +12,18 @@ SRC_URI="https://download.sourceforge.net/linux-ima/${P}.tar.gz"
 RDEPEND="
 	dev-libs/openssl:0=
 	sys-apps/keyutils:="
-DEPEND="${RDEPEND}"
+DEPEND="${RDEPEND}
+	test? ( app-editors/vim-core )"
```
thesamesam, line 16: We need to execute xxd -> BDEPEND

thesamesam:
Thanks! Switched to BDEPEND and synced live on my end as well.

PR #22433, opened 2021-09-28
games-puzzle/{concentration,gemdropx}: EAPI8, improvements

mm1ke:
@ionenwks 

Another two games updates. :)
`games-puzzle/concentration`: there is still a bug open about a file collision with `sci-biology/unafold` (https://bugs.gentoo.org/657602), i think a simply blocker would be fine. Otherwise i could try to rename this package.. any opinions about this?
`games-puzzle/gemdropx`: i could get rid of the `sed` command but I'm not really convinced about the `XTRA_FLAGS` variable since I've simply added `${LDFLAGS}` there. However the make call is still exactly the same.


ionenwks (commented):
>i think a simply blocker would be fine. Otherwise i could try to rename this package.. any opinions about this?

Sounds fine to me, this isn't the kind of thing people are likely to install at same time at all.

On a side-note usually blockers are sorted at the end of the dep list, and a one-way blocker needs a revbump to be active.

games-puzzle/gemdropx/gemdropx-0.9-r2.ebuild
```diff
@@ -1,47 +1,43 @@
...
-src_prepare() {
-	default
-	sed -i \
-		-e '/^CC/d' \
-		-e '/^CXX/d' \
-		-e 's/CXX/CC/' \
-		-e 's/-o/$(LDFLAGS) -o/' \
-		Makefile || die
-
-	find data/ -type d -name .xvpics -exec rm -rf \{\} +
```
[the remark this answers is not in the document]
mm1ke: Damn, i was specially looking for that file. I was thinking that it doesn't get installed...
I must oversaw that - fixed

games-puzzle/gemdropx/gemdropx-0.9-r2.ebuild
```diff
@@ -1,47 +1,43 @@
...
 
 src_compile() {
-	emake \
-		DATA_PREFIX="/usr/share/${PN}" \
-		XTRA_FLAGS="${CFLAGS}"
+	local myemakeargs=(
+		CC="$(tc-getCC)"
+		CXX="$(tc-getCXX)"
+		DATA_PREFIX="${EPREFIX}/usr/share/${PN}"
+		XTRA_FLAGS="${CFLAGS} ${LDFLAGS}"
```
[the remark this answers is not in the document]
mm1ke: Thank you, very good explanation and i didn't even know about these implicit make rules :)

games-puzzle/gemdropx/gemdropx-0.9-r2.ebuild
```diff
@@ -1,47 +1,43 @@
...
-}
+DOCS=( AUTHORS.txt CHANGES.txt ICON.txt README.txt TODO.txt )
 
 src_compile() {
-	emake \
-		DATA_PREFIX="/usr/share/${PN}" \
-		XTRA_FLAGS="${CFLAGS}"
+	local myemakeargs=(
+		CC="$(tc-getCC)"
+		CXX="$(tc-getCXX)"
```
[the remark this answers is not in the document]
mm1ke: Thanks for the detailed explanation (also with your previous comment).
I've changed the ebuild now as following:
```
src_prepare() {
	default
	sed -i \
		-e 's/$(CFLAGS) -o/$(LDFLAGS) -o/' \
		Makefile || die
	rm -r data/images/.xvpics || die
}

src_compile() {
	local myemakeargs=(
		CXX="$(tc-getCC)"
		DATA_PREFIX="${EPREFIX}/usr/share/${PN}"
		XTRA_FLAGS="${CFLAGS}"
	)
	emake "${myemakeargs[@]}"
}
```

However i still see that `gcc` is called directly once.
```
gcc -Wall -I/usr/include/SDL -D_GNU_SOURCE=1 -D_REENTRANT -DDATA_PREFIX=\"/usr/share/gemdropx\" -DSOUND -DJOYSTICK -DNOT_EMBEDDED -O2 -pipe -march=bdver2 -mprefer-avx128   -c -o gemdropx.o gemdropx.c
```

Adding `CC="$(tc-getCC)"` would fix it, but i'm actually not really sure where it even comes from.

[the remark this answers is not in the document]
ionenwks: I didn't mean to say should remove CC=, just update CXX to point to CC as well

The $(CC) is used by implicit rules.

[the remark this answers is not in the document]
mm1ke: Ahhh, ok, now i got it.
to recap: `$(CC) ..` is the implicit rule and thus used by the makefile for building the `.o` file (out of the `.c` file). `CXX` isn't really needed for linking (since no c++ files are present) and the previous `sed` simply removed the fixed `gcc` calls and replaced CXX by CC.
Now we don't replace `CXX` by `CC` but simply set `tc-getCC` to both. Once for the implicit `CC` variable and once for the `CXX` variable in the makefile.

I've updated the ebuild once again, and sorry for my stupidity. I really had to look a few times to understand it.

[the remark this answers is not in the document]
ionenwks: Don't worry about it, it's the Makefile itself that's silly/misleading -- I was almost tempted to suggest dropping it entirely (easy when it's a single .c file) but, along with all the -D that need to be set, felt it didn't simplify much so I dropped the idea.

games-puzzle/gemdropx/gemdropx-0.9-r3.ebuild
```diff
@@ -0,0 +1,50 @@
...
+SRC_URI="ftp://ftp.tuxpaint.org/unix/x/gemdropx/src/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="
+	media-libs/libsdl[joystick,sound,video]
+	media-libs/sdl-mixer[mod]
+	!!sci-biology/unafold"
```
ionenwks, line 19: Two things here.

1.
A weak-blocker (single `!`) is enough to prevent file-collisions, it tells portage it can ignore collisions and unmerge unafold afterwards.

A strong `!!` tells portage it /must/ unmerge first and this can sometime create unsolvable situations needing manual user interaction (users often end up asking for support not knowing what to do), it's best avoided where possible. Primary case where you'd use that is if the presence of the package would cause a build failure, but here we have no direct relation to unafold.

https://devmanual.gentoo.org/general-concepts/dependencies/#blockers

2.
This is supposed to be in cencentration, not gemdropx :)

ionenwks: Oh and don't forget to revbump concentration when adding that.

thesamesam: I didn't immediately spot that (essentially the only) example in the devmanual, but if it's not there, would you mind doing a PR to add it?

ionenwks: Spot what? The devmanual gives the weak blocker as its first example and mentions file collisions, so it's probably enough.

Revbump-wise, revbump is only needed blocker is one-sided (rather than added to both) -- but generally it's a RDEPEND change nonetheless.

thesamesam: Explicitly stating that a strong blocker is pretty much only for where something else being installed calls a failure to compile?

ionenwks: Oh I see, well it does say `if strictly necessary` -- that implies you're supposed to have a reason that goes beyond the aforementioned file collisions.

thesamesam: yep, just wanted to add a concrete example/essentially the only reason to actually do this given the misunderstandings

mm1ke: @ionenwks 
fixed and updated my PR. Thanks again for the explanation :)

ionenwks (approved):
thanks as always, looks good

PR #22434, opened 2021-09-28
app-crypt/tpm2-tools: Bump to 5.2, remove old

salahcoronya:


app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
...
+	test? (
+		app-crypt/swtpm
+		app-crypt/tpm2-abrmd
+		app-editors/vim-core
+		dev-tcltk/expect
+		dev-util/cmocka
+		dev-python/pyyaml
+	)"
+BDEPEND="virtual/pkgconfig
+	sys-devel/autoconf-archive"
```
ionenwks: `${PYTHON_DEPS}` should be here as well (executed at build time by python-single-r1, and things in non-B DEPEND can't be executed when cross-compiling).

For the same reason, believe /at least/ pyyaml, expect, and vim-core should be moved in BDEPEND as well (realistically we don't run tests when cross-compiling, but it doesn't hurt to put deps where they belong). Not sure for the others except cmocka (cmocka is linked with rather than executed) without looking deeper.

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
...
+
+RESTRICT="!test? ( test )"
+REQUIRED_USE=" ${PYTHON_REQUIRED_USE} "
```
ionenwks: Just style, but these are more commonly placed after IUSE given what they use is related and it's nice to be grouped together, e.g.
```
KEYWORDS=
IUSE=
REQUIRED_USE=
RESTRICT=
```
Also, no need for the spaces in REQUIRED_USE

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
...
+PATCHES=(
+	"${FILESDIR}/${PN}-5.1.1-no-efivar-automagic.patch"
+)
+
+src_prepare() {
+	sed -i \
+	"s/m4_esyscmd_s(\[git describe --tags --always --dirty\])/${PV}/" \
+	"${S}/configure.ac" || die
+	"${S}/scripts/utils/man_to_bashcompletion.sh"
+	default
```
ionenwks: It's preferable for patches to be applied before run sed commands or other things, i.e. after `default`. This sometime get messy otherwise.

Also,`${S}/` is the working directory in **src**_prepare, so it doesn't need to be specified.

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
...
+# https://github.com/tpm2-software/tpm2-tools/issues/2767
+PATCHES=(
+	"${FILESDIR}/${PN}-5.1.1-no-efivar-automagic.patch"
+)
+
+src_prepare() {
+	sed -i \
+	"s/m4_esyscmd_s(\[git describe --tags --always --dirty\])/${PV}/" \
+	"${S}/configure.ac" || die
+	"${S}/scripts/utils/man_to_bashcompletion.sh"
```
ionenwks: Missing `|| die`

Almost missed this because of the alignment + quotes made me feel like it was part of the sed (on that note, removing `${S}/` will allow to cleanup quotes too).

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
...
+# One of the tests fails without this patch. See
+# https://github.com/tpm2-software/tpm2-tools/issues/2767
+PATCHES=(
+	"${FILESDIR}/${PN}-5.1.1-no-efivar-automagic.patch"
+)
+
+src_prepare() {
+	sed -i \
+	"s/m4_esyscmd_s(\[git describe --tags --always --dirty\])/${PV}/" \
+	"${S}/configure.ac" || die
```
ionenwks: Be nice to have an extra tab for each line here, or reformat some other way as preferred.

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
...
+		$(use_enable test unit) \
+		--with-bashcompdir=$(get_bashcompdir) \
+		--enable-hardening
+}
+
+src_install() {
+	default
+
+	mv "${D}/$(get_bashcompdir)/tpm2_completion.bash" \
+	   "${D}/$(get_bashcompdir)/tpm2" || die
```
ionenwks: Even if unlikely to be installed on prefix, should still use `${ED}`

Could also optionally skip the double `get_bashcompdir` by using shell expansion:
```suggestion
	mv "${ED}"/$(get_bashcompdir)/tpm2{_completion.bash,} || die
```
This can feel a bit less clear in general but most people working on ebuilds are familiar with shell expansion and unbothered by it.

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
...
+	for B in "${D}"/usr/bin/tpm2_*
+	do
+		TPM2_UTILS="${TPM2_UTILS} $(basename ${B})"
+	done
+	bashcomp_alias tpm2 ${TPM2_UTILS}
```
ionenwks: Few things there
1. `${ED}` here too
2. `B` and `TMP2_UTILS` are missing a `local` declaration
3. just style, but usually we use lowercase naming for local variables, e.g. `b` or `util` instead to be more distinguishable
4. also just style but `for var in ...; do` without do/then on its own line is more common in ebuilds
5. `${B}` is missing quotes while `${D}` inside it could have spaces that get split as separate arguments to basename. I know it's strange but you can do quotes-in-quotes here because `$(...)` is its own context, i.e. `="${xvar} $(command "${yvar}")"` works -- but you could instead use `${var##*/}` to strip up to the last / without calling external basename, quotes also won't matter then.
6. could simplify a bit with `+=`, e.g. `TPM2_UTILS+=" ${var##*/}"`

But these aside, I have my own suggestion to simplify this without a loop if you like it:
```suggestion
	local utils=( "${ED}"/usr/bin/tpm2_* )
	bashcomp_alias tpm2 "${utils[@]##*/}"
```
In case not familiar with these, this will create an array of all paths using the glob, and then expand it as arguments like `econf "${myeconfargs[@]}"` would but will remove `*/` from each elements.

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
...
+	sys-libs/efivar:=
+	${PYTHON_DEPS}"
+DEPEND="${RDEPEND}
+	test? (
+		app-crypt/swtpm
+		app-crypt/tpm2-abrmd
+		app-editors/vim-core
+		dev-tcltk/expect
+		dev-util/cmocka
+		dev-python/pyyaml
```
ionenwks: Right now it'll install pyyaml for any python rather than the one requested to build this. For example if you try to test python3_10, there's nothing telling portage to install pyyaml for python3_10 first.
```suggestion
		$(python_gen_cond_dep 'dev-python/pyyaml[${PYTHON_USEDEP}]')
```
This will be evaluated by python-single-r1, so the single quotes are proper here.

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
...
+	sys-devel/autoconf-archive"
+
+RESTRICT="!test? ( test )"
+REQUIRED_USE=" ${PYTHON_REQUIRED_USE} "
+
+# One of the tests fails without this patch. See
+# https://github.com/tpm2-software/tpm2-tools/issues/2767
+PATCHES=(
+	"${FILESDIR}/${PN}-5.1.1-no-efivar-automagic.patch"
+)
```
ionenwks: Is that comment outdated? It doesn't seem related to this patch and the given link is closed.

Be nice to cleanup if so, had me confused.

ionenwks (commented):
Be nice to improve the ebuild a bit while here.

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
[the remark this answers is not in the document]
salahcoronya: Changed to 8.

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
...
+
+PYTHON_COMPAT=( python3_{8,9} )
+inherit autotools bash-completion-r1 python-single-r1
+
+DESCRIPTION="Tools for the TPM 2.0 TSS"
+HOMEPAGE="https://github.com/tpm2-software/tpm2-tools"
+SRC_URI="https://github.com/tpm2-software/${PN}/releases/download/${PV}/${P}.tar.gz"
```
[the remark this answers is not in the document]
salahcoronya: Plaintexted

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
salahcoronya: I'm a little unclear on BDEPEND. The way I understand it, stuff in BDEPEND is runs on $CBUILD and DEPEND/RDEPEND on $CHOST, however tests aren't run on cross compile, so should testing-only stuff always be on BDEPEND ? I'm not so sure now...

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
salahcoronya: Moved

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
salahcoronya: 2nd line indented

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
salahcoronya: Moved "default" to the beginning and ${S} removed

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
salahcoronya: indented / die added

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
salahcoronya: Done

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
salahcoronya: I like your version better, so I changed over your your version

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
salahcoronya: Comment is outdated, removed

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
```diff
@@ -0,0 +1,71 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{8,9} )
```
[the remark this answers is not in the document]
salahcoronya: Yup, works just fine under Python 10.

app-crypt/tpm2-tools/tpm2-tools-5.2.ebuild
ionenwks: Well there are cases where you /could/ run tests when cross-compiling (term used loosely), it's often possible with scripts, e.g. python/perl would have `BDEPEND="test? ( ${RDEPEND} )"` (typically handled by eclasses) to ensure their runtime deps are available for testing on CBUILD.

And then things like xxd (vim-core) would be executed and belong in CBUILD.

But as I said, this is more about pedantic and having dependencies in the right block, in cases where there's compiled code we won't be able to run tests anyway (so if you'd rather not bother, can leave this alone).

ionenwks:
Do the tests sometime hang for you? They passed but had to `killall swtpm` to finish. But that may just be due to my chroot setup (I'll try in a VM instead if you that's not usual).

Also, if I install libtpms-0.9.0 from the other PR I get a test failure (fine with 0.8.7), any idea what's up with that? I /was/ thinking to merge them together.
```
FAIL: test/integration/tests/testparms
======================================

creating simulator working dir: /tmp/portage/app-crypt/tpm2-tools-5.2/temp/tpm2_test_Byj9El
/tmp/portage/app-crypt/tpm2-tools-5.2/temp/tpm2_test_Byj9El /tmp/portage/app-crypt/tpm2-tools-5.2/work/tpm2-tools-5.2
Switched to CWD: /tmp/portage/app-crypt/tpm2-tools-5.2/temp/tpm2_test_Byj9El
Starting the simulator
Attempting to start simulator on port: 4584
LISTEN 0      0          127.0.0.1:4584       0.0.0.0:*    users:(("swtpm",pid=14251,fd=6))
LISTEN 0      0          127.0.0.1:4585       0.0.0.0:*    users:(("swtpm",pid=14251,fd=5))
Started simulator on port 4584 in dir "/tmp/portage/app-crypt/tpm2-tools-5.2/temp/tpm2_test_Byj9El"
tpm2tools_tcti="swtpm:host=localhost,port=4584"
Started the simulator
Starting tpm2-abrmd
TPM2ABRMD_TCTI is empty, configuring
TPM2ABRMD_TCTI="--session --dbus-name=com.intel.tss2.Tabrmd4584 --tcti=swtpm:port=4584"
tpm2-abrmd command: tpm2-abrmd --session --dbus-name=com.intel.tss2.Tabrmd4584 --tcti=swtpm:port=4584 --session --dbus-name=com.intel.tss2.Tabrmd4584 --tcti=swtpm:port=4584
tpm2tools_tcti="tabrmd:bus_type=session,bus_name=com.intel.tss2.Tabrmd4584"
TPM2TOOLS_TEST_TCTI=
TPM2TOOLS_TEST_TCTI not set, attempting to figure out default
export TPM2TOOLS_TCTI="tabrmd:bus_type=session,bus_name=com.intel.tss2.Tabrmd4584"

--- To recreate this test run the following: ---
#!/usr/bin/env bash
export TPM2_ABRMD="tpm2-abrmd"
export TPM2_SIM="swtpm"
export TPM2ABRMD_TCTI="--session --dbus-name=com.intel.tss2.Tabrmd4584 --tcti=swtpm:port=4584"
export TPM2_SIMPORT="4584"
export TPM2TOOLS_TEST_TCTI="tabrmd:bus_type=session,bus_name=com.intel.tss2.Tabrmd4584"
export TPM2TOOLS_TEST_PERSISTENT=""
export PATH="/tmp/portage/app-crypt/tpm2-tools-5.2/work/tpm2-tools-5.2/tools:/tmp/portage/app-crypt/tpm2-tools-5.2/work/tpm2-tools-5.2/tools/misc:/tmp/portage/app-crypt/tpm2-tools-5.2/work/tpm2-tools-5.2/test/integration:/tmp/portage/app-crypt/tpm2-tools-5.2/work/tpm2-tools-5.2/tools/fapi:/tmp/portage/app-crypt/tpm2-tools-5.2/temp/python3.9/bin:/usr/lib/portage/python3.9/ebuild-helpers/xattr:/usr/lib/portage/python3.9/ebuild-helpers:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/bin:/usr/lib/llvm/13/bin:/usr/lib/llvm/12/bin:/opt/cuda/bin"
export srcdir="."
export abs_srcdir="/tmp/portage/app-crypt/tpm2-tools-5.2/work/tpm2-tools-5.2"
export abs_builddir="/tmp/portage/app-crypt/tpm2-tools-5.2/work/tpm2-tools-5.2"

/tmp/portage/app-crypt/tpm2-tools-5.2/work/tpm2-tools-5.2/test/integration/tests/testparms.sh
--- EOF ---

run_startup: false
tpm2 testparms rsa:rsassa
tpm2 testparms rsa:rsapss
tpm2 testparms succeeded while it shouldn't or TPM failed
Shutting down
/tmp/portage/app-crypt/tpm2-tools-5.2/work/tpm2-tools-5.2
Removing sim dir: /tmp/portage/app-crypt/tpm2-tools-5.2/temp/tpm2_test_Byj9El
/tmp/portage/app-crypt/tpm2-tools-5.2/work/tpm2-tools-5.2/test/integration/helpers.sh: line 443: 14251 Killed                  "$TPM2_SIM" socket --tpm2 --server port="$tpm2_sim_port" --ctrl type=tcp,port="$tpm2_sim_cmd_port" --flags not-need-init --tpmstate dir="$PWD"  (wd: /tmp/portage/app-crypt/tpm2-tools-5.2/temp/tpm2_test_Byj9El)
/tmp/portage/app-crypt/tpm2-tools-5.2/work/tpm2-tools-5.2/test/integration/helpers.sh: line 443: 14857 Killed                  $TPM2_ABRMD $tpm2_tabrmd_opts $TPM2ABRMD_TCTI  (wd: /tmp/portage/app-crypt/tpm2-tools-5.2/temp/tpm2_test_Byj9El)
FAIL test/integration/tests/testparms.sh (exit status: 1)
```

ionenwks (approved):
> Do the tests sometime hang for you? They passed but had to `killall swtpm` to finish. But that may just be due to my chroot setup (I'll try in a VM instead if that's not usual).
> 
> Also, if I install libtpms-0.9.0 from #22442 I get a test failure (fine with 0.8.7), any idea what's up with that? I /was/ thinking to merge them together but I could just leave 0.9.0 alone for now.

Well, I get the same with tpm-tools-5.1.1 so it's not regressions either way. I'll go ahead and merge but leave 0.9.0 out for now.

PR #22442, opened 2021-09-29
dev-libs/libtpms: Bump to 0.9.0

salahcoronya:
Note the state file format has changed for 0.9.0, so its not possible to downgrade after updating to 0.9.0.


salahcoronya:
It turns out one of the tests in the tpm2-tool test suite had a bug in it that wasn't revealed until dev-libs/libtpms-0.9.0 . So this PR includes the patch to fix the test

ionenwks (approved):
> It turns out one of the tests in the tpm2-tool test suite had a bug in it that wasn't revealed until dev-libs/libtpms-0.9.0 . So this PR includes the patch to fix the test

Thanks for looking into it, that it's the test that was bad makes me less worried about merging this now.

PR #22443, opened 2021-09-29
media-video/handbrake: version bump to 1.4.1

SineLaude:
- Added libjpeg-turbo and zimg as new dependencies for 1.4.x
- Added modified "remove faac dependency" patch that works with 1.4.1 sources
- Removed unused gnome2-utils from inherited eclasses

Closes: https://bugs.gentoo.org/804891
Signed-off-by: Christian Schroeder <chris.schroeder.public@gmail.com>


thesamesam:
ping @SineLaude

PR #22449, opened 2021-09-30
dev-cpp/libstudxml: new package

pravorskyi:
Hi! libstudxml is a streaming XML pull parser and streaming XML serializer implementation for modern, standard C++.

This library uses the "build2" build system and I believe we could create an eclass for build2 in the future if more packages based on build2 are added to the tree.


pravorskyi:
I've added sign-off.

fkobi:
> I've added sign-off.

To let the bot know that the tags (including missing-signoff) need to be reassigned add `[please reassign]` to the PR title (see bot footer for details)

PR #22451, opened 2021-09-30
dev-texlive/texlive-xetex: Force run fc-cache to populate .uuid files

band-a-prend:
Force additional run of `fc-cache` in `pkg_postinst()` to populate `.uuid` files
in non-standard font dirs listed in `09-texlive.conf`.

Closes: https://bugs.gentoo.org/812401


thesamesam:
Thanks to both you and @juippis for testing too!

PR #22452, opened 2021-09-30
dev-scheme/racket: install boot files and static libraries 

xgqt:
Enable installing racket boot files and static library.
Both of those are needed for embedding Racket into programs (using C-interface).
Also, update configuration comments.


ulm:
Merged. Thank you.

PR #22453, opened 2021-10-01
App emacs/treemacs

Jturnerusa:
This PR will add treemacs as a new package. :smiley: 

I did want to ask about the patch to lv. It simply added require cl-lib to prevent warning when byte compiling. I could fetch a snapshot of the hydra repo and not have to patch it, I wasn't sure which would be preferred though.

I'm sure there are some things that could be cleaned up, let me know and I'll work on it! Thanks.


app-emacs/cfrs/cfrs-1.5.4.ebuild
```diff
@@ -0,0 +1,29 @@
...
+DESCRIPTION="A simple alternative to read-string that allows reading input via a small child-frame spawned at the position of the cursor."
+HOMEPAGE="https://github.com/Alexander-Miller/cfrs"
+SRC_URI="https://github.com/Alexander-Miller/cfrs/archive/refs/tags/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64"
+
+DEPEND=">=app-emacs/dash-2.11.0
+>=app-emacs/posframe-0.8.5
```
thesamesam, line 19: Indent

app-emacs/cfrs/cfrs-1.5.4.ebuild
```diff
@@ -0,0 +1,29 @@
...
+>=app-emacs/s-1.10.0"
+RDEPEND="${DEPEND}"
+
+SITEFILE="50cfrs-gentoo.el"
+DOCS=( README.org )
+
+src_compile() {
+	elisp-make-autoload-file "${S}/${PN}-autoload.el" "${S}/"
```
thesamesam, line 27: Nit: I'd say this is overquoting, and I'd do:
```suggestion
	elisp-make-autoload-file "${S}"/${PN}-autoload.el "${S}"/
```

... but it doesn't really matter.

app-emacs/ht/ht-2.3.ebuild
```diff
@@ -0,0 +1,22 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit elisp
+
+DESCRIPTION="The missing hash table library for Emacs."
```
thesamesam, line 8: [Avoid](https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide/Style_Guide#Descriptive_text) full stops at the end of DESCRIPTION:
```suggestion
DESCRIPTION="The missing hash table library for Emacs"
```

(Not sure if this is in the devmanual, but if it isn't, it probably should be. Let me know!)

app-emacs/ht/ht-2.3.ebuild
```diff
@@ -0,0 +1,22 @@
...
+HOMEPAGE="https://github.com/Wilfred/ht.el"
+SRC_URI="https://github.com/Wilfred/ht.el/archive/refs/tags/${PV}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/${PN}.el-${PV}"
```
thesamesam, line 11: Nit, see other quoting remark:
```suggestion
S="${WORKDIR}"/${PN}.el-${PV}
```

app-emacs/hydra/hydra-0.15.0.ebuild
```diff
@@ -0,0 +1,29 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+NEED_EMACS="24.3"
+
+inherit elisp
+
+DESCRIPTION="A package for GNU Emacs that can be used to tie related commands \
```
thesamesam, line 10: I'm surprised pkgcheck doesn't warn about this -- it's too [long](https://devmanual.gentoo.org/ebuild-writing/variables/index.html#ebuild-defined-variables)! If pkgcheck isn't warning about it, I think we should file a bug.

We can use <longdescription> in metadata.xml for longer stuff like this. I think it's discouraged / forbidden(?) to use line continuations in DESCRIPTION.

app-emacs/lv/files/lv-require-cl-lib.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/lv.el b/lv.el
```
thesamesam, line 1: Nit 1: I'd use [iwdevtools](https://github.com/ionenwks/iwdevtools)'s (app-portage/iwdevtools) scrub-patch to clean up the first two lines.

Nit 2: I'd name the file ```lv-${FIRST_VERSION_THIS_APPLIED_TO}-cl-lib.patch``` instead in case you have to rebase it in future.

Less-of-a-nit-and-matters-a-bit-more: Could you add a brief message at the top of the patch/file to explain why it's needed? See [here](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#patch-descriptions) in particular (but the whole page is pretty useful/worth reading if haven't already).

app-emacs/lv/lv-0.15.0.ebuild
```diff
@@ -0,0 +1,25 @@
...
+SRC_URI="https://github.com/abo-abo/hydra/archive/refs/tags/${PV}.tar.gz -> hydra-${PV}.tar.gz"
+S="${WORKDIR}/hydra-${PV}"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64"
+
+RESTRICT="test" # the tests target hydra and not lv, so we should skip them
+SITEFILE="50lv-gentoo.el"
+PATCHES=(
```
thesamesam, line 22: Nit:
```suggestion

PATCHES=(
```

app-emacs/lv/lv-0.15.0.ebuild
```diff
@@ -0,0 +1,25 @@
...
+	"${FILESDIR}/${PN}-require-cl-lib.patch"
+)
+ELISP_REMOVE="hydra-examples.el hydra-ox.el hydra-test.el hydra.el"
```
thesamesam, line 25:
```suggestion

ELISP_REMOVE="hydra-examples.el hydra-ox.el hydra-test.el hydra.el"
```

app-emacs/pfuture/pfuture-1.9.ebuild
```diff
@@ -0,0 +1,23 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit elisp
+
+DESCRIPTION="A set of simple functions wrapping Emacs’ existing process \
```
thesamesam, line 8: Too long.

app-emacs/posframe/posframe-1.0.4.ebuild
```diff
@@ -0,0 +1,29 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+NEED_EMACS="27"
+
+inherit elisp
+
+DESCRIPTION="Pop a postframe at point."
```
thesamesam, line 10:
```suggestion
DESCRIPTION="Pop a postframe at point"
```

app-emacs/posframe/posframe-1.0.4.ebuild
```diff
@@ -0,0 +1,29 @@
...
+HOMEPAGE="https://github.com/tumashu/posframe"
+COMMIT="74f06b77deeb770cd0a96977b1e6bdedb682487a"
```
thesamesam, line 12: _Personally_, I'd put this above HOMEPAGE and call it MY_COMMIT instead, but I don't think this is wrong as-is and some people have different opinions.

(Weak) rationale for each:
1. I like having the "real" standard variables together, then custom ones on top/below (not so much)
2. MY_COMMIT instead of COMMIT has an element of 'namespacing' in case something weird in the build system uses COMMIT or whatever.

app-emacs/treemacs/treemacs-2.9.ebuild
```diff
@@ -0,0 +1,57 @@
...
+inherit elisp
+
+DESCRIPTION="A tree layout file explorer for Emacs"
+HOMEPAGE="https://github.com/Alexander-Miller/treemacs"
+SRC_URI="https://github.com/Alexander-Miller/treemacs/archive/refs/tags/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="evil magit icons-dired" # some flags need packages not yet in gentoo
```
thesamesam, line 17: Which flags? Flags you want to add but aren't here? Or...?

I'd maybe do this instead:
```suggestion
# TODO: some possible extra flags need packages not yet in gentoo
# foo: would need app-emacs/bar
# ...
IUSE="evil magit icons-dired"
```

app-emacs/treemacs/treemacs-2.9.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	magit? ( app-emacs/magit )"
+DEPEND="${RDEPEND}"
+
+RESTRICT="test" # test requires cask and buttercup.el which are not packaged
+SITEFILE="50treemacs-gentoo.el"
+DOCS=( README.org Changelog.org Extensions.org )
+
+src_unpack() {
+	elisp_src_unpack
+	mv "${S}"/src/{elisp,scripts}/* "${S}/" || die
```
thesamesam, line 34: Nits:
```suggestion

	mv "${S}"/src/{elisp,scripts}/* "${S}/" || die
	
```

app-emacs/treemacs/treemacs-2.9.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	if use evil; then
+		mv "${S}/src/extra/treemacs-evil.el" "${S}/" || die
+	fi
+	if use magit; then
```
thesamesam, line 38:
```suggestion

	if use magit; then
```

app-emacs/treemacs/treemacs-2.9.ebuild
```diff
@@ -0,0 +1,57 @@
...
+		mv "${S}/src/extra/treemacs-magit.el" "${S}/" || die
+	fi
+	if use icons-dired; then
```
thesamesam, line 41:
```suggestion

	if use icons-dired; then
```

app-emacs/treemacs/treemacs-2.9.ebuild
```diff
@@ -0,0 +1,57 @@
...
+
+RESTRICT="test" # test requires cask and buttercup.el which are not packaged
+SITEFILE="50treemacs-gentoo.el"
+DOCS=( README.org Changelog.org Extensions.org )
+
+src_unpack() {
+	elisp_src_unpack
+	mv "${S}"/src/{elisp,scripts}/* "${S}/" || die
+	if use evil; then
+		mv "${S}/src/extra/treemacs-evil.el" "${S}/" || die
```
thesamesam, line 36: Nit:
```suggestion
		mv "${S}"/src/extra/treemacs-evil.el "${S}"/ || die
```
ditto the others ("overquoting")

app-emacs/treemacs/treemacs-2.9.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	fi
+}
+
+src_compile() {
+	elisp-make-autoload-file "${S}/${PN}-autoload.el" "${S}/"
+	elisp_src_compile
+}
+
+src_install() {
+	elisp_src_install
```
thesamesam, line 52:
```suggestion
	elisp_src_install
	
```

app-emacs/treemacs/treemacs-2.9.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	mkdir -p "${D}/usr/share/${PN}/icons/" || die
```
thesamesam, line 53: Can you use [dodir](https://devmanual.gentoo.org/function-reference/install-functions/) instead here?

(Also, ${ED}, not ${D}, to work on Prefix.)

app-emacs/treemacs/treemacs-2.9.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	cp -r "${S}/icons/default/" "${D}/usr/share/${PN}/icons/" || die
+	ln -s "/usr/share/${PN}/icons/" "${D}/${SITELISP}/${PN}/icons" || die
```
thesamesam, line 55: dosym, and ${ED} if you were going to use ln

app-emacs/treemacs/treemacs-2.9.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	cp "${S}"/*.py "${D}/${SITELISP}/${PN}/" || die
```
thesamesam, line 56: doins and ${ED} if you were going to stuck with cp

thesamesam: Also, *.py makes me wonder if we need to use a Python eclass..?

app-emacs/treemacs/treemacs-2.9.ebuild
```diff
@@ -0,0 +1,57 @@
...
+
+src_compile() {
+	elisp-make-autoload-file "${S}/${PN}-autoload.el" "${S}/"
+	elisp_src_compile
+}
+
+src_install() {
+	elisp_src_install
+	mkdir -p "${D}/usr/share/${PN}/icons/" || die
+	cp -r "${S}/icons/default/" "${D}/usr/share/${PN}/icons/" || die
```
thesamesam, line 54: What are these icons for?

Do we need to use [doicon](https://devmanual.gentoo.org/eclass-reference/desktop.eclass/index.html) from desktop.eclass?

app-emacs/ace-window/ace-window-0.10.0_p20200706.ebuild
```diff
@@ -0,0 +1,30 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+NEED_EMACS="24.1"
+
+inherit elisp
+
+DESCRIPTION="GNU Emacs package for selecting a window to switch to."
```
thesamesam, line 10:
```suggestion
DESCRIPTION="GNU Emacs package for selecting a window to switch to"
```

thesamesam, line 10: Random thought: if we're using a snapshot, maybe we should put a comment in the ebuild saying why, so people don't get confused.

app-emacs/treemacs/treemacs-2.9.ebuild
Jturnerusa: There are only 3 python files in src/scripts, and treemacs searches for in the directory that the elisp files get dumped in. Do you agree that a python eclass is overkill?

Byte compiling the python scripts also probably won't work because it searches for files with a .py in the directory it's installed into, but I'm not sure.

app-emacs/treemacs/treemacs-2.9.ebuild
Jturnerusa: What exactly is different here? It looks identical on my side lol!

thesamesam: New line after it :)

app-emacs/treemacs/treemacs-2.9.ebuild
Jturnerusa: The icons are used in the treemacs buffer. It shows various different icons depending on what type of file is at a certain path. For example it shows a folder icon for directories, a bandaid for patch files, a sheet of paper for plain text files, etc.

From what I can tell doicon is used for installing icons used for application menus that use desktop files right? I certainly agree that there may be a better way to do this and was very unsure of the original way when I wrote the ebuild! 

Also the symlink is needed because treemacs expects the icons to be in the directory where the elisp sources are. I don't know if it would be okay to just put them in ${SITELISP}/treemacs/icons/ or not, but that may not be terrible. 

Let me know what you think!

app-emacs/treemacs/treemacs-2.9.ebuild
thesamesam: I think [python-single-r1](https://dev.gentoo.org/~mgorny/python-guide/single.html) would be right here: if we're installing Python files, they might stop working with a future Python version, or need a minimum Python version => we should declare PYTHON_COMPAT.

Also, using a Python eclass means we can fix the shebang on the files to match the Python interpreter the user chose to install this with (via PYTHON_SINGLE_TARGET) rather than running whatever 'python' happens to be on the system.

Agree it feels like overkill but I think it's needed here?

app-emacs/treemacs/treemacs-2.9.ebuild
thesamesam: Ohh, duh, I see. Yeah, I think you're right here.

I like the /usr/share method better too with the symlink.

PR #22454, opened 2021-10-01
app-admin/lnav: Unbundle doctest

bowlofeggs:
Backport https://github.com/tstack/lnav/pull/915 so we can use
the system doctest while we wait on a new lnav upstream release.

Closes: https://bugs.gentoo.org/809752
Closes: https://bugs.gentoo.org/812353

### DIFF
--- a/app-admin/lnav/lnav-0.10.0-r2.ebuild
+++ b/app-admin/lnav/lnav-0.10.0-r2.ebuild
@@ -1,0 +1,62 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools
+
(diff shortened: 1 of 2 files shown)


app-admin/lnav/lnav-0.10.0-r2.ebuild
```diff
@@ -0,0 +1,62 @@
...
+LICENSE="BSD-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="unicode test"
+RESTRICT="!test? ( test )"
+
+RDEPEND="
+	app-arch/bzip2:0=
+	app-arch/libarchive:=
+	test? ( dev-cpp/doctest )
```
bowlofeggs, line 21: Right now, doctest is in testing so we should make sure to stabilize it before it comes time to stabilize an lnav with this dependency.

PR #22455, opened 2021-10-01
sci-libs/hdf5: added ROS3 flag

TheChymera:
Closes: https://bugs.gentoo.org/814530
Package-Manager: Portage-3.0.23, Repoman-3.0.3

### DIFF
--- a/sci-libs/hdf5/metadata.xml
+++ b/sci-libs/hdf5/metadata.xml
@@ -6,13 +6,14 @@
     <name>Gentoo Science Project</name>
   </maintainer>
   <longdescription>
-  HDF5 is a file format and library for storing scientific data. HDF5
-  was designed and implemented to address the deficiencies of HDF4.x. It
-  has a more powerful and flexible data model, supports files larger
-  than 2 GB, and supports parallel I/O.
-</longdescription>
+    HDF5 is a file format and library for storing scientific data. HDF5
+    was designed and implemented to address the deficiencies of HDF4.x. It
+    has a more powerful and flexible data model, supports files larger
+    than 2 GB, and supports parallel I/O.
+  </longdescription>
   <use>
     <flag name="hl">Enable high level API (https://support.hdfgroup.org/HDF5/doc/HL/index.html)</flag>
+    <flag name="ros3">Enable support for read-only S3 access (https://portal.hdfgroup.org/display/HDF5/Configuration+and+Setup+for+HDF5+Read+Only+S3+VFD)</flag>
     <flag name="unsupported">Enable unsupported combinations of configuration options</flag>
   </use>
 </pkgmetadata>
--- a/sci-libs/hdf5/hdf5-1.12.1-r1.ebuild
+++ b/sci-libs/hdf5/hdf5-1.12.1-r1.ebuild
@@ -1,0 +1,123 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+FORTRAN_NEEDED="fortran"
+
+inherit cmake flag-o-matic fortran-2
+
+MY_P="${PN}-${PV/_p/-patch}"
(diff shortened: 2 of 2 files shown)


sci-libs/hdf5/hdf5-1.12.1.ebuild
```diff
@@ -29,6 +29,10 @@ RESTRICT="!test? ( test )"
 
 RDEPEND="
 	mpi? ( virtual/mpi[romio] )
+	ros3? (
+		net-misc/curl
+		dev-libs/openssl
```
jsmolic, line 34: Does it link to curl and openssl? If yes, we need to add `:=`

sci-libs/hdf5/metadata.xml
```diff
@@ -6,13 +6,14 @@
...
-  than 2 GB, and supports parallel I/O.
-</longdescription>
+    HDF5 is a file format and library for storing scientific data. HDF5
+    was designed and implemented to address the deficiencies of HDF4.x. It
+    has a more powerful and flexible data model, supports files larger
+    than 2 GB, and supports parallel I/O.
+  </longdescription>
   <use>
     <flag name="hl">Enable high level API (https://support.hdfgroup.org/HDF5/doc/HL/index.html)</flag>
+    <flag name="ros3">Enable support for read-only S3 access (https://portal.hdfgroup.org/display/HDF5/Configuration+and+Setup+for+HDF5+Read+Only+S3+VFD)</flag>
```
jsmolic, line 16: I can't access this link?  
https://portal.hdfgroup.org/display/HDF5/Configuration+and+Setup+for+HDF5+Read+Only+S3+VFD

sci-libs/hdf5/hdf5-1.12.1.ebuild
```diff
@@ -17,7 +17,7 @@ SRC_URI="https://www.hdfgroup.org/ftp/HDF5/releases/${MAJOR_P}/${MY_P}/src/${MY_
 LICENSE="NCSA-HDF"
 SLOT="0/${PV%%_p*}"
 KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~ppc ~ppc64 ~riscv ~sparc ~x86 ~amd64-linux ~x86-linux"
-IUSE="cxx debug doc examples fortran +hl mpi szip test threads unsupported zlib"
+IUSE="cxx debug doc examples fortran +hl mpi ros3 szip test threads unsupported zlib"
```
jsmolic, line 20: You're adding a new USE flag and changing RDEPEND, so we need to revbump (https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html)

jsmolic (changes requested):

sci-libs/hdf5/metadata.xml
TheChymera: @jsmolic I tried again, and the link seems to be working...

sci-libs/hdf5/hdf5-1.12.1.ebuild
TheChymera: @jsmolic I am pretty sure it doesn't (some other packages which depend on hdf5[ros3] fail some tests in the network sandbox and the error is clearly ascribed to the libs and not to hdf5). But how would I know for sure?

sci-libs/hdf5/hdf5-1.12.1.ebuild
TheChymera: done :)

sci-libs/hdf5/metadata.xml
jsmolic: Yep, works for me now as well

sci-libs/hdf5/hdf5-1.12.1.ebuild
TheChymera: @jsmolic as per the discussion just now both are linked but only openssl has multiple slots

PR #22456, opened 2021-10-01
media-video/droidcam: support Clang and Clang+LTO built kernels OOTB

AdelKS:
Hello,

I use regularly droidcam but every time I want to use it I need to reboot into a GCC built kernel... I could not figure out the right environment variables to give the ebuild to make it successfully emerge on my Clang+LTO built kernels. Changing the ebuild with code from `nvidia-drivers` makes it work flawlessly (to basically switch to `clang` and `ld.lld` when needed).

Also, I removed `KERNELRELEASE="${KV_FULL}"` in the line 76, I believe it is not needed as the ebuild emerges fine without it, for any chosen kernel (not restricted to the running kernel). Although the ["linux-info" eclass](https://devmanual.gentoo.org/eclass-reference/linux-info.eclass/index.html) (that defines that variable) was not inherited, so maybe the variable was simply empty ?

Hopefully this will be useful to others.

Adel

### DIFF
--- a/media-video/droidcam/droidcam-1.8.0-r1.ebuild
+++ b/media-video/droidcam/droidcam-1.8.0-r1.ebuild
@@ -84,7 +84,7 @@ src_compile() {
 			fi
 		fi
 	fi
-	linux-mod_src_compile
+	KERNEL_DIR="/usr/src/linux" linux-mod_src_compile
 }
 
 src_test() {
--- a/media-video/droidcam/droidcam-1.8.0-r1.ebuild
+++ b/media-video/droidcam/droidcam-1.8.0-r1.ebuild
@@ -1,0 +1,155 @@
+# Copyright 2019-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit desktop linux-mod xdg
+
+DESCRIPTION="Use your phone or tablet as webcam with a v4l device driver and app"
+HOMEPAGE="https://www.dev47apps.com/droidcam/linux/"
+SRC_URI="https://github.com/dev47apps/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+KEYWORDS="~amd64"
+LICENSE="GPL-2"
+SLOT="0"
+
+IUSE="gtk"
+
+# Requires connection to phone/tablet
+RESTRICT="test"
+
+DEPEND="
+	app-pda/libplist
+	app-pda/libusbmuxd
+	dev-libs/glib
+	dev-libs/libappindicator:3
+	dev-libs/libxml2
+	dev-util/android-tools
+	media-libs/alsa-lib
+	media-libs/libjpeg-turbo
+	>=media-libs/speex-1.2.0-r1
+	media-video/ffmpeg
+	gtk? (
+		dev-cpp/gtkmm:3.0
+		x11-libs/gdk-pixbuf
+		x11-libs/gtk+:3
+		x11-libs/libX11
+		x11-libs/pango
+	)
+"
+RDEPEND="${DEPEND}"
+BDEPEND="virtual/pkgconfig"
+
+BUILD_TARGETS="all"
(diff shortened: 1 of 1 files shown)


AdelKS:
Oh nevermind, it still uses the running kernel, I need to check how to fix this.

AdelKS:
Okay, it works now with `KERNEL_DIR="/usr/src/linux" linux-mod_src_compile` but I think one needs to properly check that `/usr/src/linux` actually points towards correct kernel sources, maybe with ["check_kernel_built"](https://devmanual.gentoo.org/eclass-reference/linux-info.eclass/index.html) ?

AndrewAmmerlaan:
> Also, I removed KERNELRELEASE="${KV_FULL}" in the line 76, I believe it is not needed as the ebuild emerges fine without it, for any chosen kernel (not restricted to the running kernel). 

> Oh nevermind, it still uses the running kernel, I need to check how to fix this.

If I remember correctly `KERNELRELEASE="${KV_FULL}"` was added exactly for this purpose.

> Okay, it works now with KERNEL_DIR="/usr/src/linux" linux-mod_src_compile but I think one needs to properly check that /usr/src/linux actually points towards correct kernel sources, maybe with "check_kernel_built" ?

We don't need the kernel to be built completely for droidcam, but we do need to run at least `make modules_prepare` before trying to compile any module. (Most of the droidcam ebuild is heavily inspired on the v4l2loopback and virtualbox-modules ebuilds, which as far as I know also don't require the kernel to be built completely.)

I'm gonna take a proper look at this later, because I am rather busy with other things at the moment, sorry

AndrewAmmerlaan:
> > Also, I removed KERNELRELEASE="${KV_FULL}" in the line 76, I believe it is not needed as the ebuild emerges fine without it, for any chosen kernel (not restricted to the running kernel).
> 
> > Oh nevermind, it still uses the running kernel, I need to check how to fix this.
> 
> If I remember correctly `KERNELRELEASE="${KV_FULL}"` was added exactly for this purpose.

Upstream Makefile has:

```
KERNELRELEASE ?= $(shell uname -r)
KERNEL_DIR  ?= /lib/modules/$(KERNELRELEASE)/build
```

So either `KERNELRELEASE` or `KERNEL_DIR` should do the trick.

Though I see the linux-mod eclass already sets KERNEL_DIR, so now I'm a bit confused as to why this is necessary at all:
```
KERNEL_DIR ?= /usr/src/linux
    A string containing the directory of the target kernel sources. The default value is "/usr/src/linux" 
```

I wonder if `KERNEL_DIR=${KERNEL_DIR} linux-mod_src_compile` or `export KERNEL_DIR` would also work? To me this seems to be the cleanest way to do it.

AdelKS:
> If I remember correctly KERNELRELEASE="${KV_FULL}" was added exactly for this purpose.

Unfortunately when I tried this it made the ebuild use the running kernel to build for the target kernel... some kind of mix and made the ebuild fail. A good test bed is to try to build for a clang+LTO built kernel from a GCC kernel and conversely.

> linux-mod inherits linux-info, so we can use variables set in linux-info even if we don't inherit it explicitly in the ebuild.

Oh alright, did not know. It indeed looks like `KV_FULL` is filled from the kernel pointed by `/usr/src/linux`. 

> I wonder if KERNEL_DIR=${KERNEL_DIR} linux-mod_src_compile or export KERNEL_DIR would also work? To me this seems to be the cleanest way to do it.

I will try!

> We don't need the kernel to be built completely for droidcam, but we do need to run at least make modules_prepare before trying to compile any module.

Yes! But I do not know what is the fitting check for kernels that have at least `make modules_prepare`, I can check other ebuilds.

> I'm gonna take a proper look at this later, because I am rather busy with other things at the moment, sorry

Take all your time, good luck!

AdelKS:
Looks like the export method works!

AdelKS:
Happy to help! 

I opened a bug report for suggesting that we handle this at the `linux-mod` eclass level: https://bugs.gentoo.org/816024

PR #22457, opened 2021-10-01
app-emulation/docker-cli: use `${PV}` for CLI version

tormath1:
Bug: https://bugs.gentoo.org/815658


app-emulation/docker-cli/docker-cli-20.10.7.ebuild
```diff
@@ -37,7 +37,7 @@ src_compile() {
 	export CGO_LDFLAGS="-L${ESYSROOT}/usr/$(get_libdir)"
 		emake \
 		LDFLAGS="$(usex hardened '-extldflags -fno-PIC' '')" \
-		VERSION="$(cat VERSION)" \
+		VERSION="${PV}"  \
```
floppym, line 40: Get rid of the extra space between the quote and the backslash.

PR #22463, opened 2021-10-02
keyword ~riscv for certbot and dependencies.

ahmedcharles:


prometheanfire:
generally ok with this, is riscv seeking to have this be marked stable soon as well?

ahmedcharles:
I don't actually know. I'm fairly new to using gentoo, I'm just testing packages and keywording them as having been tested.

ahmedcharles:
Merged, thanks.

ionenwks:
> generally ok with this, is riscv seeking to have this be marked stable soon as well?

riscv is marked testing-only in arches.desc right now, so stabilization doesn't happen for anything yet

dlan17:
> > generally ok with this, is riscv seeking to have this be marked stable soon as well?
> 
> riscv is marked testing-only in arches.desc right now, so stabilization doesn't happen for anything yet

quite right, given current status (not too many developers working on this arch, not too many packages keyworded/tested),
we'd still keep it as testing-only .. thanks

PR #22464, opened 2021-10-02
games-fps/gzdoom: Version bump to 4.7.0

vilhelmgray:
The gles2, telemetry, and vulkan USE flags are introduced.

Closes: https://bugs.gentoo.org/815274


chewi:
I was initially confused by this CI result, then I realised that I caused it in 67e17930f5dc6ce0f781fb7f8defbf26471fa03a! I'll see whether Vulkan is still limited to amd64.

chewi:
Okay, if you rebase, that error should go away now.

vilhelmgray:
Just a side-note: in theory the OpenGL dependency can be conditional now, but due to the way GZDoom is coded (it's lacking a specific OPENGL preprocessor define), OpenGL would still remain at the very least an unconditional RDEPEND dependency. Because of this, I decided to keep things simple with this version and leave OpenGL as an unconditional dependency as it has been so far in order to simplify our maintenance of this ebuiild.

PR #22465, opened 2021-10-02
dev-qt/qtwayland: More fine-grained subslot

a17r:
If the observation in the bug is indeed correct, then at least this time we can get away with simply doing the same as in qtcore/qtgui/qtsql.

### DIFF
--- a/dev-qt/qtwayland/qtwayland-5.15.2-r14.ebuild
+++ b/dev-qt/qtwayland/qtwayland-5.15.2-r14.ebuild
@@ -7,6 +7,7 @@ KDE_ORG_COMMIT=efe6edcaf8eba601dff99ec6ad4457c8a4442f86
 inherit qt5-build
 
 DESCRIPTION="Wayland platform plugin for Qt"
+SLOT=5/${QT5_PV} # bug 815646
 
 if [[ ${QT5_BUILD_TYPE} == release ]]; then
 	KEYWORDS="amd64 arm arm64 ~hppa ~ppc ~ppc64 ~riscv ~sparc x86"


dev-qt/qtwayland/qtwayland-5.15.2-r14.ebuild
```diff
@@ -7,6 +7,7 @@ KDE_ORG_COMMIT=efe6edcaf8eba601dff99ec6ad4457c8a4442f86
 inherit qt5-build
 
 DESCRIPTION="Wayland platform plugin for Qt"
+SLOT=5/${QT5_PV} # bug 815646
```
Pesa, line 10: I'm not opposed to adding this, but the reporter said qtwayland was upgraded from 5.15.2-r1 to 5.15.2-r13, so this subslot definition won't help.

dev-qt/qtcore/qtcore-5.15.2-r10.ebuild
```diff
@@ -8,7 +8,7 @@ QT5_MODULE="qtbase"
 inherit linux-info qt5-build
 
 DESCRIPTION="Cross-platform application development framework"
-SLOT=5/$(ver_cut 1-3)
+SLOT=5/${QT5_PV}
```
Pesa, line 11: ACK on this commit

dev-qt/qtwayland/qtwayland-5.15.2-r14.ebuild
thesamesam: It should as a _one off_ (b/c changes from 5/0 -> 5/${QT5_PV}), but yeah, it won't help for future breaks in patches which get imported (if that happens). Bleh.

a17r: Yes this would be a one-time fix, for now.

a17r:
So, after having established that 3 revdeps are indeed using private headers from `dev-qt/qtwayland`, not having had the minor version subslot indeed looks like an omission.

List of revdeps:
``
>=kde-plasma/kwayland-integration-5.22.90
>=kde-plasma/plasma-workspace-5.22.90
>=kde-plasma/layer-shell-qt-5.22.0
``

On the other hand, all relatively young dependencies.

PR #22471, opened 2021-10-03
net-im/qtox: update live ebuild and add two new packages

Chiitoo:
These will most likely need more work, but I figured it's time to put them out here for picking them nits and such.

I guess at the very least the metadata might need another name or few for the maintainer bits...


Chiitoo:
All comments should have been implemented as appropriate, I hope.

I wonder a bit about the x86 keyword for the new packages due to the on-going conversation regarding them.  I guess in this case it is appropriate to add it to them, since they're dependencies of qTox that is already has it.

ionenwks:
> I wonder a bit about the x86 keyword for the new packages due to the on-going conversation regarding them. I guess in this case it is appropriate to add it to them, since they're dependencies of qTox that is already has it.

Yeah, once add next qtox release would've needed to keyword for ~x86 anyway -- dropping keywords tend to be more of a hassle and needs rekeywording requests. I'd just skip stabilizing x86 if go there, unless someone requests that too (there's a bug open right now asking for amd64 qtox).

That aside, skipping adding ~x86 on new ebuilds is more for if you want to skip having to test it yourself (i.e. usually at least in a x86 chroot), if you don't mind then it's harmless and just nice for people still using it (always supposed to try it at least once for keywording, but people likely often casually add ~x86 without ever trying on new packages -- I'll give these a quick try later on x86 anyhow, not that I expect issues).

Chiitoo:
Hm, at the very least I forgot to add '- Disable a couple of network related tests.' to the commit message.

PR #22472, opened 2021-10-03
net-im/telegram-desktop-bin: disable internal updater again

henning-schild:
Upstream silently dropped the "-externalupdater" switch. Instead we need
to create a file in /etc to disable the internal updater.

Closes: https://bugs.gentoo.org/814062

### DIFF
--- a/net-im/telegram-desktop-bin/telegram-desktop-bin-3.0.1-r1.ebuild
+++ b/net-im/telegram-desktop-bin/telegram-desktop-bin-3.0.1-r1.ebuild
@@ -16,13 +16,16 @@ LICENSE="GPL-3-with-openssl-exception"
 SLOT="0"
 KEYWORDS="-* ~amd64"
 
-QA_PREBUILT="usr/lib/${PN}/Telegram"
+QA_PREBUILT="usr/bin/telegram-desktop"
 
 RDEPEND="
 	dev-libs/glib:2
 	>=media-libs/fontconfig-2.13
 	media-libs/freetype:2
+	sys-libs/zlib
 	virtual/opengl
+	x11-libs/libSM
+	x11-libs/libdrm
 	x11-libs/libX11
 	>=x11-libs/libxcb-1.10[xkb]
 "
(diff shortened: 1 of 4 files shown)


net-im/telegram-desktop-bin/telegram-desktop-bin-3.0.1-r1.ebuild
```diff
@@ -33,9 +33,11 @@ RDEPEND="
...
 
 src_install() {
-	exeinto /usr/lib/${PN}
-	doexe "Telegram"
-	newbin "${FILESDIR}"/${PN} "telegram-desktop"
+	newbin "Telegram" "telegram-desktop"
+
+	insinto /etc/tdesktop/
+	echo "/usr/bin/telegram-desktop" > "${WORKDIR}/externalupdater"
+	doins "${WORKDIR}/externalupdater"
```
ionenwks, line 40: Missing EPREFIX, and echo that writes to a file would need `|| die`, but you can skip the latter by simplifying with:
```suggestion
	newins - externalupdater <<<"${EPREFIX}/usr/bin/telegram-desktop"
```

net-im/telegram-desktop-bin/telegram-desktop-bin-3.0.1-r1.ebuild
```diff
@@ -33,9 +33,11 @@ RDEPEND="
 S="${WORKDIR}/Telegram"
 
 src_install() {
-	exeinto /usr/lib/${PN}
-	doexe "Telegram"
-	newbin "${FILESDIR}"/${PN} "telegram-desktop"
+	newbin "Telegram" "telegram-desktop"
+
+	insinto /etc/tdesktop/
```
ionenwks, line 38: These quotes and trailing / aren't needed, but up to you (it's just style).

net-im/telegram-desktop-bin/telegram-desktop-bin-3.0.1-r1.ebuild
ionenwks: Oh and fwiw it wouldn't hurt to add a bugref to explain this, just a ` #814062` at the end of the line is enough.

net-im/telegram-desktop-bin/telegram-desktop-bin-3.0.1-r1.ebuild
henning-schild: removed quotes and /

net-im/telegram-desktop-bin/telegram-desktop-bin-3.0.1-r1.ebuild
henning-schild: applied that suggestion with stdin instead of echo, not sure the bug tag would be needed. Commit messages have it, do not feel like duplication.

PR #22489, opened 2021-10-04
games-strategy/warmux: fix build, deps, bump eapi

DarthGandalf:
Remove nls USE flag, because it caused build failure when disabled,
because in that case it defines its own ngettext macro, which conflicts
with that function defined in libintl.h which is transitively included
anyway.

Closes: https://bugs.gentoo.org/739400
Closes: https://bugs.gentoo.org/796914

### DIFF
--- a/games-strategy/warmux/warmux-11.04.1-r2.ebuild
+++ b/games-strategy/warmux/warmux-11.04.1-r2.ebuild
@@ -1,36 +1,46 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=8
 
 inherit autotools desktop
 
 DESCRIPTION="A free Worms clone"
-HOMEPAGE="http://gna.org/projects/warmux/"
-SRC_URI="http://download.gna.org/warmux/${P}.tar.bz2"
+HOMEPAGE="https://wiki.gentoo.org/wiki/No_homepage"
+SRC_URI="mirror://gentoo/${P}.tar.bz2"
 S="${WORKDIR}"/${PN}-11.04
 
-LICENSE="GPL-2"
+LICENSE="
+	GPL-2+
+	|| ( Apache-2.0 GPL-3 )
+	UbuntuFontLicense-1.0
+	vlgothic
+"
 SLOT="0"
 KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
-IUSE="debug nls unicode"
+IUSE="debug unicode"
 
 RDEPEND="
 	dev-libs/libxml2
+	media-libs/libpng:=
 	media-libs/libsdl[joystick,video,X]
 	media-libs/sdl-image[jpeg,png]
 	media-libs/sdl-mixer[vorbis]
 	media-libs/sdl-ttf
 	media-libs/sdl-net
-	media-libs/sdl-gfx
-	media-fonts/dejavu
+	media-libs/sdl-gfx:=
 	net-misc/curl
+	virtual/libintl
 	x11-libs/libX11
-	nls? ( virtual/libintl )
-	unicode? ( dev-libs/fribidi )"
+	unicode? ( dev-libs/fribidi )
+"
 DEPEND="${RDEPEND}
+	x11-base/xorg-proto
+"
+BDEPEND="
+	sys-devel/gettext
 	virtual/pkgconfig
-	nls? ( sys-devel/gettext )"
+"
 
 PATCHES=(
 	"${FILESDIR}"/${P}-gentoo.patch
@@ -40,6 +50,7 @@ PATCHES=(
 	"${FILESDIR}"/${P}-stat.patch
 	"${FILESDIR}"/${P}-fix-c++14.patch
 	"${FILESDIR}"/${P}-respect-AR.patch
+	"${FILESDIR}"/${P}-clang.patch
 )
 
 src_prepare() {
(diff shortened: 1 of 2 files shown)


games-strategy/warmux/warmux-11.04.1-r2.ebuild
```diff
@@ -52,15 +57,14 @@ src_configure() {
...
+		--enable-nls \
 		$(use_enable debug) \
-		$(use_enable nls) \
 		$(use_enable unicode fribidi)
 }
 
 src_install() {
 	default
 
-	rm -f "${ED%/}"/usr/share/${PN}/font/DejaVuSans.ttf || die
```
ionenwks, line 63: I know you're removing it because it's not installed, but it tips off the ebuild overlooked the change because of this -f.

It's installing 3 other fonts, and it wouldn't use dejavu if not for --with-font-path (we're essentially forcing a different font and installing a bundled one for nothing).

Could unbundle them if you want (imagine the two other non-font-path fonts would need workarounds), but I'm not picky about unbundling for fonts (not like we're talking outdated libraries here), so could alternatively just remove the dejavu dep along with --with-font-path, and add the relevant LICENSE for these fonts `|| ( Apache-2.0 GPL-3 ) UbuntuFontLicense-1.0 vlgothic`

games-strategy/warmux/warmux-11.04.1-r2.ebuild
```diff
@@ -13,24 +13,28 @@ S="${WORKDIR}"/${PN}-11.04
 LICENSE="GPL-2"
```
ionenwks, line 13: main.cpp:
```
 *  the Free Software Foundation; either version 2 of the License, or
 *  (at your option) any later version.
```

games-strategy/warmux/warmux-11.04.1-r2.ebuild
```diff
@@ -52,15 +57,14 @@ src_configure() {
 		--with-localedir-name="${EPREFIX}"/usr/share/locale \
 		--with-datadir-name="${EPREFIX}"/usr/share/${PN} \
```
ionenwks, line 58: Can remove these two.

Passing localedir notably tends to be a bad leftover from games.eclass because datadir=/usr/games made locales go in /usr/games/locale and the like, but no one cleaned them up in migration.

See defaults:
```
  --with-datadir-name=DIR specify where datas are installed (default:
                          ${prefix}/share/warmux)
  --with-localedir-name=DIR
                          specify where locales are installed (default:
                          ${prefix}/share/locale)
```

games-strategy/warmux/warmux-11.04.1-r2.ebuild
```diff
@@ -13,24 +13,28 @@ S="${WORKDIR}"/${PN}-11.04
...
 	media-fonts/dejavu
 	net-misc/curl
 	x11-libs/libX11
-	nls? ( virtual/libintl )
-	unicode? ( dev-libs/fribidi )"
-DEPEND="${RDEPEND}
+	virtual/libintl
+	unicode? ( dev-libs/fribidi )
+"
+DEPEND="${RDEPEND}"
```
ionenwks, line 33: Missing `x11-base/xorg-proto`

games-strategy/warmux/warmux-11.04.1-r2.ebuild
```diff
@@ -13,24 +13,28 @@ S="${WORKDIR}"/${PN}-11.04
...
+BDEPEND="
 	virtual/pkgconfig
-	nls? ( sys-devel/gettext )"
+	sys-devel/gettext
```
ionenwks, line 36: Can sort deps here and above.

ionenwks (commented):
+1 for removing nls, generally I don't feel it's a USE worth supporting -- and even less if it's going to cause issues

Be nice to "fix" https://bugs.gentoo.org/631434 as well. The forks are looking dubious/dead and there's no real replacement I can see, so I'd just use the sad:
```
HOMEPAGE="https://wiki.gentoo.org/wiki/No_homepage"
SRC_URI="mirror://gentoo/${P}.tar.bz2"
```

games-strategy/warmux/warmux-11.04.1-r2.ebuild
DarthGandalf: Yes. I tried to, but got confused with the order of letters in the alphabet.

games-strategy/warmux/warmux-11.04.1-r2.ebuild
DarthGandalf: Only build/android/jni needs it?

games-strategy/warmux/warmux-11.04.1-r2.ebuild
DarthGandalf: Where did you find license for these fonts?

DarthGandalf:
All done. Re https://bugs.gentoo.org/631434 - I looked for the open bugs of this packages, but missed this one

games-strategy/warmux/warmux-11.04.1-r2.ebuild
ionenwks: sdl[X] needs it, try to force remove xorg-proto and then emerge (which won't pull xorg-proto because it's a not a dep)

games-strategy/warmux/warmux-11.04.1-r2.ebuild
ionenwks: You can get them from their ebuilds:
media-fonts/wqy-microhei (although it's license is slightly wrong, it's one or the other `|| (  )`)
media-fonts/ubuntu-font-family
media-fonts/vlgothic (this has extra licenses but they don't apply to the used font as far as I can tell)

games-strategy/warmux/warmux-11.04.1-r2.ebuild
ionenwks: Personally I don't really think about it, in vim I just e.g. `:.,30!sort` to sort the range, I tend to prefer POSIX sorting too (capital X comes before a).

May want to do something similar in whichever editor you use.

ionenwks: Oh and by "and above" I was implying RDEPEND isn't sorted either.. but well, I'll fix that on merge :)

ionenwks (approved):
Thanks, 3 bugs in less.

games-strategy/warmux/warmux-11.04.1-r2.ebuild
DarthGandalf: I fixed one line in RDEPEND (which is above), but didn't notice the sdl* ones :(
Using the editor is a good trick, thanks

PR #22495, opened 2021-10-05
media-gfx/mandelbulber: Version bump to 2.26

miller-alex:
Existing ebuild is outdated, fails to build with USE=opencl; maintainer-needed.
Development moved to github.

Newer versions seem to build without problems, so just bump to latest.

Build tested on amd64 both with all USE flags disabled and USE="opencl openexr sndfile tiff".
Didn't test new USE=joystick, though.


thesamesam:
Thank you! I've kept the old version for now in case the new one has any regressions.

PR #22497, opened 2021-10-05
Add ebuild for Mixxx 2.3.1 latest release

geraldog:
Let's please have Mixxx 2.3.1 building with portage :)

Incorporated styling changes by @thesamesam

Ping @fordfrog


thesamesam:
Thanks! I'll let fordfrog review. Could you update the 9999 ebuild too if it needs it?

geraldog:
My pleasure. Yes, that's the plan, first getting this one right then backporting the necessary changes to 2.3.9999 ;)

I think 2.4.9999 deserves an ebuild too, but that's a little trickier.

fordfrog:
i just added request for keywording ``libkeyfinder`` (https://bugs.gentoo.org/816561)

otherwise, iirc, as @thesamesam already suggested, you should use ``repoman full`` and ``pkgcheck scan`` to verify that there are no issues with the ebuild. you can also see the issues in the "Pull request CI report". you can ignore the one related to ``libkeyfinder``, for that i created the bug.

media-sound/mixxx/files/mixxx-2.3.1-local-manuals.patch
```diff
@@ -0,0 +1,20 @@
+--- a/src/widget/wmainmenubar.cpp
++++ b/src/widget/wmainmenubar.cpp
+@@ -556,7 +556,7 @@ void WMainMenuBar::initialize() {
+     pHelpMenu->addAction(pHelpSupport);
+ 
+     // User Manual
+-    QUrl manualUrl = documentationUrl(m_pConfig->getResourcePath(),
++    QUrl manualUrl = documentationUrl("/usr/share/doc/mixxx-2.3.1",
```
fordfrog, line 8: can't we update the resource path itself instead of hardcoding these occurrences?

media-sound/mixxx/mixxx-2.3.1.ebuild
```diff
@@ -0,0 +1,126 @@
...
+		-DINSTALL_USER_UDEV_RULES=OFF
+		-DWAVPACK=$(usex wavpack on off)
+		-DQTKEYCHAIN=$(usex qtkeychain on off)
+		-DKEYFINDER=$(usex keyfinder on off)
+	)
+
+	if [[ "${PV}" == 9999 ]] ; then
+	local mycmakeargs+=(
+		-DENGINEPRIME="OFF"
+	)
```
fordfrog, line 109: please indent this block to improve readability

media-sound/mixxx/files/mixxx-2.3.1-local-manuals.patch
geraldog: I confirm we can't mess with the resource path. There's even a Mixxx command line argument for setting the resource path, which is where e.g. controller mappings reside.

By the way, this was the only request from the Mixxx side (upstream), that the manual PDF would properly open through the GUI

jospezial:
https://github.com/libusb/hidapi/releases/tag/hidapi-0.11.0
https://github.com/gentoo/gentoo/pull/17290

geraldog:
Hi @jospezial unfortunately I do not have access to any HID hardware to test with Mixxx, so I can't help much with bumping hidapi to 0.11.0

geraldog:
> > I had seen the issues already. I set up repoman and it's taking a loooong time to scour through the neighbourhood... I'll push some more fixes once that completes.
> 
> Please make sure you run it in the direcotry with this ebuild in, or it'll scan the whole tree/repo ;)

Sheesh, it's gonna hit Mars!

> 
> I also recommend you use pkgdev commit/repoman commit to ensure your commit message has the right prefix.

You are very kind Sam James, thank you for the guidance.

geraldog:
Sorry about the confusion, I'm able to use repoman now.

We need to discuss whether hardcoding the path to the docs is allowable. If it can't find the docs at the default location, it will always fallback to web.

Plus, should we add Engine Prime Export as a proper USE flag? This will download and build libdjinterop statically.

Lastly, not a block for merge, but we need to coordinate with someone who has HID pro gear in the near future and get hidapi 0.11 building with portage, with CMake and all.

geraldog:
Sorry everyone, I've been very busy and still am. I'll find the time to work more on the PR next week. Thanks!

jospezial:
Some months ago I invested time together with @fordfrog to create the cmake based mixxx live ebuilds.
These are universal to be used also for releases.
@geraldog Please use these as your base and add your changes to have the same universal ebuild content for the latest release and live ebuilds.
What is the right syntax?
```
-DFAAD="$(usex aac on off)"
or
-DFAAD=$(usex aac on off)
```

Thanks for your work.

jospezial:
related CMAKE_BUILD_TYPE:
https://bugs.gentoo.org/797235
https://github.com/mixxxdj/mixxx/pull/4112
https://gitweb.gentoo.org/repo/user/gerislay.git/tree/media-sound/mixxx/files/mixxx-9999-gentoo-build.patch

"media-sound/mixxx-9999 - disable mixxx-test"
https://bugs.gentoo.org/822357
https://gitweb.gentoo.org/repo/user/gerislay.git/commit/media-sound/mixxx/files?id=9003d67b4072e7879fedcf823a4e4855aa6b199c
https://gitweb.gentoo.org/repo/user/gerislay.git/tree/media-sound/mixxx/files/drop-mixxx-test.patch

media-sound/mixxx/mixxx-2.3.1.ebuild
```diff
@@ -0,0 +1,125 @@
...
+	ffmpeg? ( media-video/ffmpeg:0= )
+	hid? ( dev-libs/hidapi )
+	keyfinder? ( media-libs/libkeyfinder )
+	lv2? ( media-libs/lilv )
+	mp3? ( media-libs/libmad )
+	mp4? ( media-libs/libmp4v2:= )
+	opus? (	media-libs/opusfile )
+	qtkeychain? ( dev-libs/qtkeychain )
+	shout? ( >=media-libs/libshout-2.4.5 )
```
jospezial, line 68: libshout-idjc
https://github.com/mixxxdj/mixxx/blob/8acb633220024222504cddcd1f5ea26e659fbcc7/CMakeLists.txt#L2465
https://bugs.gentoo.org/show_bug.cgi?id=775443

geraldog:
@jospezial  thanks for the heads up I'll take it into account later when I push the necessary fixes. As to whether there's a guideline in regards to quoting I sincerely don't know but in my humble opinion it shouldn't be used without need.

Finally got some time for working on this.

geraldog:
One thing I have noticed is that when Upstream changes the Manual (which is often) we'll have to update the Manifest.

geraldog:
With all due respect to everybody invovled which is so far @jospezial, @fordfrog and @thesamesam but does Gentoo have a clever way of redownloading for example lthe Mixxx 2.3 Manual - which is always changing - and fixing the manifest for everybody, i.e. it will be changed when we do emerge --sync ? Maybe there's a script for that I'm unaware of?

Apart from that, I think the commit is almost clean now unless anybody has some major divergences to share?  I"m aware that @jospezial pointed out we need to work on packaging some key dependencies that are being statically built for now, and that changes need to backported to 2.3.9999 at least (honestly I think we need not only that but 2.4.9999 - main - building too). That will be worked on.

thesamesam:
We need to mirror the manual in some state (possibly to dev.gentoo.org) if there's no stable URL upstream.

geraldog:
Please realize I'm totally aware of the (in)security implications of re-downloading material over the open internet and updating perfectly good checksums with unknown quality ones.

Perhaps this can be solved by the gentoo mirroring infrastructure? So at least we don't break Portage compilations with checksum/invalid hash errors when upstream changes the manual? I'm asking in good faith

geraldog:
@thesamesam I must have read your mind 3 minutes too late

geraldog:
I think I messed up. I should probably start a new PR.

PR #22499, opened 2021-10-06
sys-libs/libcxxabi: fix abi_x86_32 build & test failures

telans:
I'm not sure if you would want this for all libcxxabi versions because I imagine tests fail there too, however the build fails for `13.0.0` without this.

`sys-libs/libcxxabi-13.0.0[abi_x86_32]` now builds and completes tests without errors for me.

Closes: https://bugs.gentoo.org/773604


mgorny (changes requested):
This doesn't look like the right solution. The sole fact that you're patching out a check blocking this indicates it's not correct.

PR #22521, opened 2021-10-08
net-mail/notmuch: don't unconditionally require gmime[crypt]

Hello71:
patch out the configure tests instead. gmime:3 <3.2.7 was removed over a
year ago, and gpgme <1.8.0 was removed many years ago, so we can assume
all gentoo users will have it already

Signed-off-by: Alex Xu (Hello71) <alex_y_xu@yahoo.ca>


thesamesam:
Not maintainer but bit reluctant about a large patch given that notmuch is sometimes fragile to bump anyway.

PR #22525, opened 2021-10-09
gui-apps/waybar: added popups as default use flag

tiotags1:
without 'popups' use flag waybar behaves in unexpected ways


JonasToth:
Thanks for your patch.

What are the unexpected ways? Is this a Bug for every system or a preference for you?
Maybe you could elaborate a bit more on this and/or reference a bug.

tiotags1:
if you're not using 'popups' use flag, popups show up in wrong places, some design elements don't show, the application name flickers when you hover on it, etc
I believe it should default to the more user friendly appearance

JonasToth:
I see.
That is ok for me and i do not have concerns with it. Not sure if it is against any gentoo policy to change the defaults of such a flag, but given this is not a stable package it seems fine.

The ebuild would need a revision bump though. It should become the `0.9.7-r2` version.

PR #22526, opened 2021-10-09
dev-libs/imath, media-libs/{ilmbase,openexr}: unkeyword hppa

waebbl:
drop hppa keyword from all versions.

Affects all imath and ilmbase versions and openexr:0 versions.

Bug: https://bugs.gentoo.org/788235


waebbl:
@thesamesam Several issues with revdeps having hppa support. So we can not easily drop the keyword for now.

waebbl:
Shall I leave this bug open?

PR #22530, opened 2021-10-09
dev-ml/findlib: work around findlib and ocamlbuild META collision

hlein:
Still has one remaining failure case, if findlib is installed before ocamlbuild.

Bug: https://bugs.gentoo.org/803275


atupone:
The following scenario does not work here with this patch

emerge findlib
emerge ocamlbuild
emerge findlib

It seems to me that findlib-1.9.1 (without this PR) shall be emerged before ocamlbuild.
After the first time, you can re-emerge findlib or ocamlbuild, there is no problem.

I don't know how to express that in the ebuild

PR #22531, opened 2021-10-09
net-irc/quassel: add spell support without full kde

arthurzam:
Add new use flag "spell" for enabling spell checking without full KDE integration. KF5Sonnet is categorized as "Optional KF5 tier1
component" and can be used without full KDE, if find_package finds it.
Using this new USE flag we disable automagic dependency on KF5Sonnet if it is installed during build. I have tested on my system, and if it is installed, it is linked into quassel.

Changed REQUIRED_USE so that kde depends on spell, and spell depends on some kind of gui.

I hope "spell" is a good name for this USE flag?


thesamesam:
Rebased, hopefully correctly, onto the _rc2/9999 changes I made. Thanks!

PR #22543, opened 2021-10-10
app-misc/broot: bump to 1.6.4

creideiki:
Add USE=clipboard for X11 clipboard support.

Install shell completions for Bash, Zsh and Fish.

Add libgit2 to RDEPEND.

Closes: https://bugs.gentoo.org/815826
Package-Manager: Portage-3.0.26, Repoman-3.0.3
Signed-off-by: Karl-Johan Karlsson <creideiki@lysator.liu.se>


creideiki:
The X11 clipboard functionality behaves weirdly. When I build it, I get some extra commands, and running those commands with a Wireshark between Broot and my X server shows that Broot is talking with the X server about selections, but neither the primary selection nor the clipboard are updated.

All my systems are KDE Plasma on X11, so there may be something specific to that setup. I've asked for tests on https://bugs.gentoo.org/815826 but the user who contributed the enabling of the clipboard functionality has not responded in a week. Could someone here please test it? To do so, install this package with `USE="clipboard"`, run `broot`, type `:copy_path` and see if the pathname appears in selection or clipboard.

PR #22548, opened 2021-10-10
net-libs/grpc: remove unused patch(es)

mm1ke:


gyakovlev:
@arthurzam seems you've committed a patch, but it's not used in ebuild. can you please check if it's correct?

non-working tests for `net-libs/grpc` is ok, it's complete PITA to make it work properly.

gyakovlev:
6808a5a7420dfa7355ed17d81b0f88d6889f3894 just for the record

arthurzam:
@gyakovlev Yes, my mistake - we can remove it.
I was trying to fix IUSE="test" build (do note that it fails for older versions also, but thankfully it is tests restricted so we hide it).
I thought the fix would be to use system gtest or manually bundle gtest (using manual download and link) - but both failed.

gyakovlev:
thanks for checking, merged.

PR #22558, opened 2021-10-11
media-libs/libsdl2: add USE to enable hidapi-libusb flag

Tatsh:

[Yuzu](https://yuzu-emu.org/) and other emulators need this flag enabled if SDL is to be used as a shared library. Without it, motion controls silently fail.

Yuzu is also in GURU but only builds from Git master and submodules for the time being.

[My Yuzu ebuild](https://github.com/Tatsh/tatsh-overlay/blob/master/games-emulation/yuzu)

### DIFF
--- a/media-libs/libsdl2/libsdl2-2.0.16.ebuild
+++ b/media-libs/libsdl2/libsdl2-2.0.16.ebuild
@@ -14,7 +14,7 @@ LICENSE="ZLIB"
 SLOT="0"
 KEYWORDS="~alpha amd64 arm arm64 ~hppa ~ia64 ppc ppc64 ~riscv sparc x86"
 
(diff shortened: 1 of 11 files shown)


Tatsh:
Yes. Happy to change it

There are two variants of HIDAPI usage: HIDAPI+libusb and HIDAPI without libusb. Do we want a USE flag for either of these? Currently without anything only HIDAPI without libusb support is built (based on auto detection).

I propose having `+hidapi` (`+` because most people would want it and it is the default; precedent: `app-misc/g810-led`) and then specify `USE="hidapi libusb"` to this variant.

tiagodfer:
Hey guys, what's the current state of this request? Is there anyway to have motion detected by SDL2?

Tatsh:
It's a pain to make this proper for Gentoo so it's on hold.

PR #22564, opened 2021-10-12
sci-libs/rocSPARSE: remove incomplete test suites which fail

littlewu2508:
test_csricsv.yaml does not have function parameters for testing
csricsv, so template function csricsv will be uninstantiated, which
causes gtest to FAIL it. This is also the cause for csrilusv.
Since these two yaml does not contain actual tests, just remove them at
this time.


littlewu2508:
The fix is incorrect. The root cause is misplaced matrix data file during src_prepare.

PR #22573, opened 2021-10-13
net-im/telegram-desktop: Apply openssl3 patch to 3.0.1

mid-kid:
@gyakovlev

Build-time exclusive fix, package will already be rebuilt when the
openssl version changes.

I want to stabilize this one soon, it's been too long.


gyakovlev:
thanks, merged.

PR #22574, opened 2021-10-13
meson.eclass: better handling of optimization...

netfab:
... and debug core options.

During configure phase, if an ebuild use a debug useflag in this way :
>	$(meson_use debug)

meson will produce the following warning :
> WARNING: Recommend using either -Dbuildtype or -Doptimization + -Ddebug.
> Using both is redundant since they override each other.
> See: https://mesonbuild.com/Builtin-options.html#build-type-options

For example, you can see this warning in the current portage tree if you do :
> $ ebuild "$(portageq get_repo_path / gentoo)"/games-emulation/dosbox-staging/dosbox-staging-0.77.0-r1.ebuild configure

Instead of using the --buildtype meson core option, we extract the level
optimization value from BUILD_{C,CPP,CXX,FC,LD}FLAGS, and we use the
-Doptimization meson core option. Then we strip the '-O*' option from
all *FLAGS.

{_}_meson_is-flagq and _meson_filter-{flags,var} functions are highly
inspired (or completely copied) from flag-o-matic.eclass.

Signed-off-by: Fabrice Delliaux <netbox253@gmail.com>

### DIFF
--- a/eclass/meson.eclass
+++ b/eclass/meson.eclass
@@ -43,7 +43,7 @@ if [[ -z ${_MESON_ECLASS} ]]; then
 _MESON_ECLASS=1
 
 [[ ${EAPI} == 6 ]] && inherit eapi7-ver
-inherit multiprocessing ninja-utils python-utils-r1 toolchain-funcs
+inherit flag-o-matic multiprocessing ninja-utils python-utils-r1 toolchain-funcs
 
 EXPORT_FUNCTIONS src_configure src_compile src_test src_install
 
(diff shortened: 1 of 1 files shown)


ionenwks:
Is the package treating the build type specially?

For the most parts, using this only adds -g and it's something users should set themselves rather than through USE=debug

floppym:
Calling `meson setup --buildtype=plain --debug` or `meson setup -Dbuildtype=plain -Ddebug=true` seems to work fine. The latter option simply overrides the former, which is probably what you want to happen anyway.

netfab:
> I want to see meson.build from your project.

Here it is : https://github.com/netfab/GLogiK/blob/dev/meson.build
You will not find anything special here, like I said :
> enable_debug = get_option('debug')

> Calling meson setup --buildtype=plain --debug or meson setup -Dbuildtype=plain -Ddebug=true seems to work fine.
> The latter option simply overrides the former, which is probably what you want to happen anyway.

True, but the eclass currently use --buildtype, and this cannot be overrided with -Dbuildtype \:
> $ meson --buildtype=plain -Dbuildtype=plain -Ddebug=true . /tmp/foo_build8
> ERROR: Got argument buildtype as both -Dbuildtype and --buildtype. Pick one.

This means that the eclass should use the -Dbuildtype form instead of --buildtype.

Mixing both options forms seem to not work well \:
> $ meson --buildtype=plain -Ddebug=true . /tmp/foo_build1 | grep Message
> Message: debug:false

> $ meson -Dbuildtype=plain -Ddebug=true . /tmp/foo_build2 | grep Message
> Message: debug:true

This would eliminate the need to use -Doptimization. Initially I focused on this warning because of this wrong debug behavior.

netfab:
Yes, it works. Thank you !

PR #22583, opened 2021-10-15
net-im/discord-bin: Update depedencies

hugglesfox:
net-im/discord-bin had a few dependencies that are no longer required so I've updated them to be equal to the dependencies in the deb provided from discord.

I've added dev-libs/libappindicator as an optional feature as it's only required if the user wishes to use the system tray and in that case, libappindicator is probably already present anyway. 

I've also added pulseaudio as an optional feature as whilst it's not in the dependencies listed in the deb, https://bugs.gentoo.org/722668 states it's required so I felt it probably doesn't hurt to at least mention it at install time. It could be added as a dependency but there are multiple packages which provides the pulse api (pulseaudio, apulse, pipewire, etc), and most of the time the user will have one of those installed anyway, so I feel it is cleaner just to have it as an optfeature.  

Lastly I also bumped the eapi to version 8.

If possible, I wouldn't mind proxy maintaining net-im/discord-bin going forward as it's something I rely on daily. This would be my first time proxy maintaining.

### DIFF
--- a/net-im/discord-bin/discord-bin-0.0.16-r1.ebuild
+++ b/net-im/discord-bin/discord-bin-0.0.16-r1.ebuild
@@ -3,14 +3,15 @@
 
 EAPI=8
 
-MY_PN=${PN/-bin/}
-MY_BIN="D${MY_PN/d/}"
+MY_PN="${PN%-bin}"
+MY_BIN="${MY_PN^}"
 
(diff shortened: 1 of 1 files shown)


hugglesfox:
Ah sorry I missed that bit about the gbm use flag. Thanks heaps for all the help and support.

PR #22587, opened 2021-10-15
www-apps/redmine: new versions 4.1.5 and 4.2.3

winterheart:
Fix security issue CVE-2021-42326
Bug: https://bugs.gentoo.org/817917


thesamesam:
Thanks!

PR #22595, opened 2021-10-16
dev-java/openjdk-bin: add musl support

yemouu:
prebuilt musl binaries are available for amd64 systems

Signed-off-by: yemou Hannam <meeyou@tuta.io>


gyakovlev:
I ended up doing it differently by abusing bash a bit.
`abi_uri` is already quite confusing.

thanks for contribution and inspiration.

gyakovlev:
fixed libc dep in next commit btw, missed at first.

PR #22600, opened 2021-10-16
x11-misc/shutter: limit dev-libs/glib version to <2.70

DarthGandalf:


ionenwks:
Downgrading glib after other applications have been built with the new one is something that caused a lot of random issues before, and would much rather avoid imposing this kind of restriction (this is safer on binary distros given they know which version things been built against)

Fortunately stable is not affected for the time being and it's not /completely/ broken in ~arch

PR #22605, opened 2021-10-16
sys-cluster/opencoarrays: New package. Move from ::guru

band-a-prend:
New package `sys-cluster/opencoarrays-2.9.3` provides Fortran 2018 standard Coarrays feature
for parallel programming for gfortran(`sys-devel/gcc[fortran]`) compiler.

Package depends on implementation of "MPI 3.1".

The sys-cluster/openmpi is preferred in `RDEPEND` over `sys-cluster/mpich`
like in `virtual/mpi`.

Tests fail with `FEATURES="network-sandbox"` for most versions of openmpi
and mpich it with error:
```
"No network interfaces were found for out-of-band communications.
We require at least one available network for out-of-band messaging."
```
Therefore the following EAPI-8 option is added: `PROPERTIES="test_network"`

### DIFF
--- a/sys-cluster/opencoarrays/opencoarrays-2.10.0.ebuild
+++ b/sys-cluster/opencoarrays/opencoarrays-2.10.0.ebuild
@@ -6,7 +6,7 @@ EAPI=8
 CMAKE_MAKEFILE_GENERATOR=emake
 FORTRAN_STANDARD="2003"
 
-inherit cmake fortran-2
+inherit cmake flag-o-matic fortran-2
 
 MY_PN="OpenCoarrays"
 
@@ -33,3 +33,9 @@ RDEPEND="
 DEPEND="
 	${RDEPEND}
 "
+
+src_configure() {
+	filter-lto # Bug 860765
+
+	cmake_src_configure
+}


sys-cluster/opencoarrays/opencoarrays-2.9.3.ebuild
```diff
@@ -0,0 +1,43 @@
...
+RESTRICT="!test? ( test )"
+
+RDEPEND="
+	|| ( >=sys-cluster/openmpi-4.1.2[fortran] >=sys-cluster/mpich-3.4.3[fortran,mpi-threads,threads] )
+"
+DEPEND="
+	${RDEPEND}
+"
+
+pkg_setup() {
```
thesamesam, line 37: cmake.eclass doesn't export a pkg_setup so you shouldn't need to put this (it should be by default)

sys-cluster/opencoarrays/opencoarrays-2.9.3.ebuild
```diff
@@ -0,0 +1,43 @@
...
+	fortran-2_pkg_setup
+}
+
+src_test() {
```
thesamesam, line 41: Do tests not work without this?

sys-cluster/opencoarrays/opencoarrays-2.9.3.ebuild
band-a-prend: It seems I misunderstood this.

Should I still to use `pkg_setup()` for calling `fortran-2_pkg_setup` explicitly to check `FORTRAN_STANDARD` variable compatibility?

sys-cluster/opencoarrays/opencoarrays-2.9.3.ebuild
band-a-prend: Yes, tests work without this.

sys-cluster/opencoarrays/opencoarrays-2.9.3.ebuild
band-a-prend: I will drop it on version bump as it seems the `fortran-2_pkg_setup` is exported by default.

band-a-prend:
Actually Fortran Coarrays feature requires compiler support of `Fortran 2008` standard therefore the `fortran-2.eclass` is require to be updated to check it.

sys-cluster/opencoarrays/opencoarrays-2.10.0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+
+# Tests fail with FEATURES="network-sandbox" for most versions of openmpi and mpich it with error:
+# "No network interfaces were found for out-of-band communications.
+#  We require at least one available network for out-of-band messaging."
+IUSE="test"
+PROPERTIES="test_network"
+RESTRICT="!test? ( test )"
+
+RDEPEND="
+	|| ( >=sys-cluster/openmpi-4.1.2[fortran] >=sys-cluster/mpich-3.4.3[fortran,mpi-threads,threads] )
```
Alessandro-Barbieri, line 31: can't you use virtual/mpi?

band-a-prend: No. I discussed it with one of gentoo developers and he offered to specify explicit implementation as MPI 3.1 support is required by package while not all virtual/mpi packages provides it.

Alessandro-Barbieri: maybe virtual/mpi can be versioned?

band-a-prend: Agree. But I need to recheck again what MPI 3.1 feature was required and test against nullmpi and nativempi.

band-a-prend: @Alessandro-Barbieri maybe `virtual/mpi[-nullmpi]` is suitable?

But I'm confused that `nullmpi? ( sys-cluster/native-mpi )` within virtual/mpi but not `nullmpi? ( sys-cluster/nullmpi )`.
@thesamesam is it alright?

But the minimal supported version of MPI by sys-cluster/native is still under question for me :(

band-a-prend: @Alessandro-Barbieri there is another one problem: package could require `sys-cluster/mpich[fortran,mpi-threads,threads]` while `virtual/mpi` doesn't provide `mpi-threads` USE-flag but without it the tests are broken.

I don't know the details why `treads` and `mpi-threads` are split in `sys-cluster/mpich`.

PR #22606, opened 2021-10-16
media-tv/plex-media-server: version bump, remove old

ephemer0l:
Changed to use the init that ships with plex, as well as a /etc/default/plexmediaserver override file.


thesamesam:
Thanks!

PR #22623, opened 2021-10-18
app-editors/featherpad: Bump to 1.0.1

hansfn:


app-editors/featherpad/featherpad-1.0.1.ebuild
```diff
@@ -0,0 +1,37 @@
...
+inherit xdg cmake
+
+DESCRIPTION="Lightweight Qt5 Plain-Text Editor for Linux"
+HOMEPAGE="https://github.com/tsujan/FeatherPad"
+SRC_URI="https://github.com/tsujan/FeatherPad/archive/V${PV}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/FeatherPad-${PV}"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
```
thesamesam, line 15: Why drop ~riscv? See https://devmanual.gentoo.org/keywording/index.html#keywording-on-upgrades.

Or is it because of a new dep and we'll file the rekeywording bug once this is merged?

hansfn: I didn't drop `~riscv`. It was added to the 1.0.0 ebuild (by somebody else) after I created this patch. I can update the PR, no problem.

thesamesam: np! I'll squash the commit.

thesamesam:
Merged as https://github.com/gentoo/gentoo/commit/f31ce7ecbd751a7ed7e9375f8627db9db5f72468!

PR #22633, opened 2021-10-19
sys-fs/erofs-utils: new ebuild

xen0n:
Tested on my amd64 box. This is my first upstreamed ebuild, and I hope I got the metadata right! (It seems all non-Gentoo-dev submissions are implicitly marked proxy-maint.)


xen0n:
Oops, the package is already in GURU :facepalm:

Should this be closed instead?

hsiangkao:
Hi @xen0n, there was a bug related to that --- https://bugs.gentoo.org/701284.
According to "https://bugs.gentoo.org/701284#c7", I think there may some other work to do?

> Alec Moskvin 2020-08-31 16:12:55 UTC
> (In reply to David Michael from comment #3)
>> If no developers want to add this, I can take it as a proxy maintainer

> Hi David,

> Have you seen the Proxy Maintainer guide?
> https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide

> It says should should either do a Github PR, send the patch to the list or have proxy-maint CC'd on the bug, and I don't think you've done either.

thesamesam:
Importing this from GURU would be fine but can you contact the GURU maintainer to see if they want to help or have thoughts?

PR #22641, opened 2021-10-20
net-im/gajim: fix nbxmpp dependency

Flowdalic:
Closes: https://bugs.gentoo.org/818976


thesamesam:
This is a pretty straightforward fix and we also want to stable this version shortly for the sec bug, so merging.

PR #22645, opened 2021-10-20
app-emulation/xen-tools: fix building with ocaml 4.12

hydrapolic:
Fix building with ocaml 4.12


hydrapolic:
Thank you @thesamesam

PR #22652, opened 2021-10-21
net-analyzer/hydra/hydra-9999999.ebuild

EmilienMottet:
Use README instead of README.md in dodoc

Closes: https://bugs.gentoo.org/776670
Signed-off-by: Emilien Mottet <emilien.mottet@grenoble-inp.org>


thesamesam:
It's ok to just force push to the same PR btw

PR #22663, opened 2021-10-21
media-sound/beets: bump to version 1.5.0

ArsenArsen:
I re-introduced some of the previously failing tests, they pass on my machine now. `test_completion` still appears to be broken, unfortunately. I don't have a clean system to run tests on, though. I also didn't check whether some of the test deps are now extraneous. If the changelog is detailed enough, there shouldn't be such cases.

Signed-off-by: Arsen Arsenović <arsen@aarsen.me>


thesamesam:
Thanks! Will file the rekeywording bug.

PR #22677, opened 2021-10-22
www-servers/apache: Fix runtime defines for proxy deps

klondi:
The proxy_http2 and proxy_uwsgi modules depend on the proxy module
which is masked by the PROXY runtime define. As a result, systems
compiled with either module would fail to run unless either the
appropriate IfDefine directives were added to http2.conf or the
-DPROXY directive was added on apache2.conf so that the proxy
module was loaded.

This patch fixes the MODULE_DEFINES variable to correctly keep
track of that. Since this only affects configuration files and
nobody has reported the issue since June when the uwsgi module
was added, the ebuild is not revbumped in this commit.

Affected users can either manually patch the http2.conf or
reemerge the ebuild and run etc-config to see the changes.

RepoMan-Options: --force
Signed-off-by: Francisco Blas Izquierdo Riera (klondike) <klondike@gentoo.org>

### DIFF
--- a/www-servers/apache/apache-2.4.51.ebuild
+++ b/www-servers/apache/apache-2.4.51.ebuild
@@ -112,7 +112,9 @@ MODULE_DEFINES="
 	proxy_ftp:PROXY
 	proxy_html:PROXY
 	proxy_http:PROXY
+	proxy_http2:PROXY
 	proxy_fcgi:PROXY
+	proxy_uwsgi:PROXY
 	proxy_scgi:PROXY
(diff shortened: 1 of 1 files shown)


jkroonza (changes requested):
This will need to also apply to -r1 (and potentially -r2 which I'm about to file a PR for).  I'll pre-emptively include this into -r2 so you can ignore that one.

jkroonza (commented):
Actually, kill this, this is outdated, I'll put into same PR for the other stuff I need to do currently.

PR #22689, opened 2021-10-24
x11-themes/papirus-icon-theme: skip pantheon variants by default

ixti:
ePapirus and ePapirus-Dark are variants designed for elementaryOS and Pantheon DE.

Signed-off-by: Alexey Zapparov <alexey@zapparov.com>

### DIFF
--- a/x11-themes/papirus-icon-theme/papirus-icon-theme-20211001-r1.ebuild
+++ b/x11-themes/papirus-icon-theme/papirus-icon-theme-20211001-r1.ebuild
@@ -1,0 +1,23 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit xdg
+
+DESCRIPTION="Free and open source SVG icon theme"
+HOMEPAGE="https://github.com/PapirusDevelopmentTeam/papirus-icon-theme"
+SRC_URI="https://github.com/PapirusDevelopmentTeam/papirus-icon-theme/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3"
+SLOT="0"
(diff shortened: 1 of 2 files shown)


x11-themes/papirus-icon-theme/papirus-icon-theme-20211001-r1.ebuild
```diff
@@ -0,0 +1,23 @@
...
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~hppa ~ia64 ~m68k ~mips ~ppc ~ppc64 ~riscv ~s390 ~sparc ~x86"
+IUSE="pantheon"
+
+src_compile() { :; }
+
+src_install() {
+	insinto /usr/share/icons
+	doins -r Papirus{,-Dark,-Light}
+	use pantheon && doins -r ePapirus{,-Dark}
```
juippis, line 22: In general we don't add USE flags to control installing few kb worth of stuff, 

but even without that, why would you not want to install these variants unconditionally? :P

ixti: I'm fine to install those unconditionally. I simply was always confused what exactly ePapirus theme is (it's not very obvious from its name), and according to README that flavour is sensible for the pantheon desktop environment only.

@juippis do you think we should remove use flag?

ixti: I've opened a new PR with new version of the package and unconditional theme installation (but added a comment in the ebuild what those themes are) :D

ixti: But, I think we need the use flag :P

ixti:
Closing in favour of https://github.com/gentoo/gentoo/pull/22893

PR #22699, opened 2021-10-24
app-arch/snappy: remove -fno-rtti option

winterheart:
Remove -fno-rtti option which may broke compilation of dependant packages. Migrate to EAPI=8, minor cleanups.
Closes: https://bugs.gentoo.org/819909


stha09:
Chromium does not depend on snappy anymore.

juippis (approved):

PR #22705, opened 2021-10-26
www-apps/gitea: security bump to 1.15.5

hydrapolic:
Tested on amd64, thanks!

https://github.com/go-gitea/gitea/releases/tag/v1.15.5


thesamesam:
Thank you!

thesamesam:
(Tempting to say file a security bug for stuff like this even though it's ~arch only, so do if you can, but it is a bit of a pain given we're not going to do anything with it now that cleanup is done.)

hydrapolic:
Thanks @thesamesam. Is there any point opening a security bug given we are not going to cleanup nor stabilize?

PR #22714, opened 2021-10-26
app-emulation/genymotion-bin: fix  preserved libs with dev-libs/hiredis-1.0.1

fedeliallalinea:
Closes: https://bugs.gentoo.org/820386

### DIFF
--- a/app-emulation/genymotion-bin/genymotion-bin-3.2.1-r1.ebuild
+++ b/app-emulation/genymotion-bin/genymotion-bin-3.2.1-r1.ebuild
@@ -1,0 +1,122 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit bash-completion-r1 desktop pax-utils
+
+MY_PN="${PN/-bin}"
+MY_P="${MY_PN}-${PV}"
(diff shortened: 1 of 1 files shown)


app-emulation/genymotion-bin/genymotion-bin-3.2.1-r1.ebuild
```diff
@@ -0,0 +1,122 @@
...
+	default
+
+	chmod +x ${BIN_ARCHIVE} || die "chmod failed"
+	yes | ./${BIN_ARCHIVE} > /dev/null || die "unpack failed"
+
+	# removed windows line for bashcompletion
+	sed -i "/complete -F _gmtool gmtool.exe/d" "${MY_PN}/completion/bash/gmtool.bash" || die "sed failed"
+
+	# patch to support hiredis
+	for i in genymotion genyshell gmtool player libcom.so.1.0.0 librendering.so.1.0.0 ; do
```
thesamesam, line 73: Hiredis didn't mean to bump SONAME here, so that's been fixed now in 1.0.2, but FWIW, this approach is rather dangerous and I'm not keen on doing it in general.

There's no real way to be sure it's safe and I'd rather pin deps on the specific subslot of the thing that broke ABI (given rebuilds aren't possible).

Thank you for the PR though!

PR #22718, opened 2021-10-26
net-misc/bfgminer: Upstream patch to fix build failure with USE=-cpumining

luke-jr:
Closes: https://bugs.gentoo.org/775560

No revbump because only changes anything in scenarios where the build failed previously.


thesamesam:
Thanks!

net-misc/bfgminer/files/bfgminer-5.5.0-cpus-undefined.patch
```diff
@@ -0,0 +1,21 @@
+commit 83f83d2cbc5e3044b314b914beb32dbb83a0055e
```
thesamesam, line 1: Pop a link in to the source in future at the top

PR #22724, opened 2021-10-27
x11-terms/kitty misc fixes and co-maintain

ionenwks:
@end222 If don't mind I'd want to co-maintain kitty since I started using it.

Revamped the ebuild a bit (notably handles the requested "make X optional" -- not that I needed that myself, but it works).

Merging this isn't urgent (until there's a new release), so posting this here in case want to take a look or have anything to suggest.


end222:
No problem from my part.
I've taken a look at the changes you've made and it looks good to me, thank you very much for your work!

PR #22726, opened 2021-10-27
sci-geosciences/qgis: Fix qtserialport dependency

doctaweeks:

### DIFF
--- a/sci-geosciences/qgis/qgis-3.20.0-r1.ebuild
+++ b/sci-geosciences/qgis/qgis-3.20.0-r1.ebuild
@@ -43,7 +43,6 @@ COMMON_DEPEND="
 	dev-qt/qtnetwork:5[ssl]
 	dev-qt/qtpositioning:5
 	dev-qt/qtprintsupport:5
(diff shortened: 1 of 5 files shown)


thesamesam:
Thanks! Revbumped though to avoid the useless dependency for people.

doctaweeks:
> Thanks! Revbumped though to avoid the useless dependency for people.

I thought by not bumping it would prevent unnecessary rebuilds since the actual linking *is* handled by the use flag. It would just avoid unnecessarily installing qtserialport or allow depcleaning it. (Binary packages aside...) Perhaps I still do not understand when/where dependencies are checked.

thesamesam:
Sorry, to be clear, yes, I just wanted to allow unnecessarily installing it + allow depcleaning.

doctaweeks:
I think I understand after reading: https://wiki.gentoo.org/wiki/Project:Portage/Dynamic_dependencies

I did not realize dynamic deps were not in PMS and all dependency changes still required a revbump.

PR #22736, opened 2021-10-28
Bump app-editors/kakoune to 2021.10.28

mujo-hash:
No patches needed for this release and the previous release hasn't stabilized so just drop it.


mujo-hash:
ping

PR #22737, opened 2021-10-28
metadata/install-qa-check.d: add 60libtool-la (check for unnecessary …

thesamesam:
….la files)

### DIFF
--- a/metadata/install-qa-check.d/60libtool-la
+++ b/metadata/install-qa-check.d/60libtool-la
@@ -1,0 +1,45 @@
+# Check if we're installing .la files unnecessarily
+# https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0303
+
+libtool_la_check() {
+	if [[ ${CATEGORY}/${PN} == dev-libs/libltdl ]] ; then
+		# bug #293921
+		return
+	fi
+
+	# Bail out if there aren't any .la files being installed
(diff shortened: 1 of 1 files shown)


mgorny:
Well, to be honest I think it'd be better if you check for files with matching names, i.e. complain for every `.la` that doesn't come with matching `.a`.

metadata/install-qa-check.d/60libtool-la
```diff
@@ -0,0 +1,35 @@
...
+	# Iterate over all the .la files we are installing to verify there's
+	# a corresponding .a file - they're pointless without a corresponding
+	# static library.
+	local file
+	local dir
+	local base
+	local bad_files=()
+	for file in "${files[@]}" ; do
+		dir=$(dirname ${file})
+		base=$(basename ${file})
```
mgorny, line 18: Don't call external tools when you can do it via variable substitution.

metadata/install-qa-check.d/60libtool-la
```diff
@@ -0,0 +1,35 @@
...
+		base=${base/.la/}
```
mgorny, line 19:
```suggestion
		base=${base%.la}
```

You don't expect `.la` in middle of the name, do you?

mgorny (changes requested):

thesamesam:
Merged a little while ago.

PR #22738, opened 2021-10-28
media-gfx/openvdb-{7.1.0-r2,8.0.1-r1}: find numpy

waebbl:
media-gfx/openvdb-8.0.1-r1:
- adds consistency in the find_package call for NumPy by adding the
	same components like in the find_package call for Python
- additionally pass -DPython_INCLUDE_DIR to cmake
- add an option for python unittests when USE=test is set
- remove flag-o-matic inherit, which isn't needed according to pkgcheck
- change the negation on the utils USE flag to build the utilities if
	the USE flag is set.

Bug: https://bugs.gentoo.org/788886

media-gfx/openvdb-7.1.0-r2:
- additionally pass -DPython_INCLUDE_DIR and -DPython_LIBRARY to help
	find numpy
- remove flag-o-matic inherit, which isn't needed according to pkgcheck
- change the negation of the utils USE flag use for building the utilities
	if the USE flag is set

Closes: https://bugs.gentoo.org/788886

@redchillipadi:
- If the option for python unittests in v8 was intentionally left out, please tell me, and I'm gonna remove it. I didn't checked whether the tests actually work, just saw the cmake variable in the cmakefile.
- Also, if I misunterstand the logic for the utility binaries and they work as expected with the negated setting, tell me and I'm gonna reset them.

### DIFF
--- a/media-gfx/openvdb/openvdb-7.1.0-r2.ebuild
+++ b/media-gfx/openvdb/openvdb-7.1.0-r2.ebuild
@@ -5,7 +5,7 @@ EAPI=7
 
 PYTHON_COMPAT=( python3_{7,8,9} )
 
-inherit cmake flag-o-matic python-single-r1
+inherit cmake python-single-r1
 
 DESCRIPTION="Library for the efficient manipulation of volumetric data"
 HOMEPAGE="https://www.openvdb.org"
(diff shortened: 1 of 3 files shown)


media-gfx/openvdb/openvdb-7.1.0-r2.ebuild
```diff
@@ -5,7 +5,7 @@ EAPI=7
```
thesamesam, line 5: We should revbump for this change as it fixes how/which tools get built.

waebbl: @thesamesam thanks for revbumping. I was on a vacation today, so couldn't react that fast.

PR #22748, opened 2021-10-28
net-vpn/protonvpn-cli: last version bump

ILMostro:
Closes: https://bugs.gentoo.org/809449
Closes: https://bugs.gentoo.org/806097
Signed-off-by: Amel Hodzic <ilmostro7@gmail.com>


thesamesam:
Could you clean up in a separate commit?

ILMostro:
> Could you clean up in a separate commit?

"clean up" the older version or the distutils issue? I imagine, the new version needs to be available before the old one can be cleaned.  So, make this commit for the version bump only, right?

thesamesam:
Right, new version in one commit, then cleanup in the next :)

PR #22754, opened 2021-10-29
 games-util/lutris: add version 0.5.9.1

AndrewAmmerlaan:
- We add a patch that avoids:
importError (Requiring namespace 'Gtk' version '3.0', but '4.0' is already loaded)

- And we remove the sed line that makes the application fail to find data_path during runtime, the test that this claims to fix works (for me) without this sed line anyway.
Bug: https://bugs.gentoo.org/811285

- Bumped to EAPI 8

- Also update 9999 accordingly

Signed-off-by: Andrew Ammerlaan <andrewammerlaan@gentoo.org>


AndrewAmmerlaan:
Patch submitted upstream here: https://github.com/lutris/lutris/pull/3790

thesamesam (approved):

PR #22757, opened 2021-10-30
sys-kernel/rt-sources: Stable Updates

shifty91:
Stable Updates for rt-sources excluding v4.4 due to missing tarball on kernel.org. Thanks!


shifty91:
Added v5.15-rt LTS and removed unsupported v5.14-rt.

PR #22759, opened 2021-10-30
net-news/liferea: add python 3.10 and use optfeature eclass

nE0sIghT:
Closes: https://bugs.gentoo.org/818748


thesamesam:
Thanks!

PR #22765, opened 2021-10-30
app-misc/broot: bump to 1.7.0 and remove old

creideiki:


app-misc/broot/broot-1.7.0.ebuild
```diff
@@ -0,0 +1,267 @@
...
+	dev-libs/libgit2:=
+	X? ( x11-libs/libxcb:= )
+"
+DEPEND="${RDEPEND}"
+BDEPEND="X? ( ${PYTHON_DEPS} )"
+
+QA_FLAGS_IGNORED="usr/bin/broot"
+
+pkg_setup() {
+	use X && python-any-r1_pkg_setup # Used by XCB crate
```
ionenwks: Well, so much for this... since the xcb crate was updated, I had a look at it... they removed the python script :eyes:
```
## [0.10.0] - 2021-09-19 - rtbo
### Changed
- build script is written in Rust. Python no longer needed. (#56, #62, #89, #99)
```
So all the python cruft can be removed (inherit, compat, pkg_setup, BDEPEND), builds fine even if python isn't found now.

creideiki: Neat. Pushed a new version with the Python stuff removed.

Broot's changelog also says "fix clipboard features not working on some recent linux distributions", and that functionality now works for me as well.

PR #22768, opened 2021-10-30
media-video/wireplumber: enable PW/WP by default on systemd

pinkflames:
As requested by @leio of Gentoo GNOME project and accepted by @thesamesam  who
maintains media-video/wireplumber, we're transitioning to have it
enabled by default. For now a 30-wireplumber.preset file is generated
and installed but it's possible that all user units are set to enabled
vendor state by default, which would make such a file unnecessary.

The minimum pipewire version was raised so that the updated user instructions in elog would be correct.

### DIFF
--- a/media-video/wireplumber/wireplumber-9999.ebuild
+++ b/media-video/wireplumber/wireplumber-9999.ebuild
@@ -40,7 +40,7 @@ BDEPEND="
 DEPEND="
 	${LUA_DEPS}
 	>=dev-libs/glib-2.62
-	>=media-video/pipewire-0.3.37
+	>=media-video/pipewire-0.3.39
 	virtual/libc
 	elogind? ( sys-auth/elogind )
 	systemd? ( sys-apps/systemd )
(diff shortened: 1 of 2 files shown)


media-video/wireplumber/wireplumber-9999.ebuild
```diff
@@ -54,6 +54,23 @@ RDEPEND="${DEPEND}"
 
 DOCS=( {NEWS,README}.rst )
 
+presetfile=30-${PN}.preset
+
+src_prepare() {
+	default
+
+	einfo "Generating ${presetfile}"
+	cat > ${presetfile} <<- EOF || die
```
floppym, line 63: Why are we generating what is essentially a static file at runtime? We could just ship it in FILESDIR instead.

media-video/wireplumber/wireplumber-9999.ebuild
```diff
@@ -54,6 +54,23 @@ RDEPEND="${DEPEND}"
...
+		# Start of ${presetfile} from ${P}
+
+		enable pipewire.socket
+		enable pipewire-pulse.socket
```
floppym, line 67: I feel like pipewire (not wireplumber) should install a preset for these, or you should create some separate ebuild that installs them.

floppym: I guess this doesn't really matter.

media-video/wireplumber/wireplumber-9999.ebuild
pinkflames: At the time this was done it was not clear if and what preset file we'd be installing in the long run. Or, as I mentioned to you on Libera/Matrix, which package would do that (since ultimately the intention is to get pipewire[-sound] operational). This alone means that I did not want to provide a file that then needs to be present in every git clone and every rsync mirror and on every Gentoo user repository when it may go away eventually. Worse yet it may possibly change multiple times over the next few versions, leading to multiple of such small files being there on every Gentoo system and mirror for weeks or months.

Finally the code in question is a common pattern in Gentoo media related packages (the old known example being sys-auth/realtime-base), so if this was somehow against Gentoo rules or guidelines, other packages would need fixing too.

Seeing how this is non-actionable suggestion, I will be marking this conversation as resolved but thank you for making me consider more use of FILESDIR in the future packaging work.

media-video/wireplumber/wireplumber-9999.ebuild
pinkflames: Another option would be to have PipeWire conditionally installing a higher priority (lower numbered) preset file that specifically disabled pipewire-pulse.socket. This PR was made before it was determined that the current Gentoo default is for user units to be vendor enabled out of the box, so concern was that it would be inappropriate to call `systemctl --global preset foo bar baz` when the actual preset was not provided (and there's little point in PipeWire ebuild enabling only its own units since they are non-functioning without a session manager).

pinkflames:
As brought up in https://bugs.gentoo.org/823104 it may be prudent to also use --force option, because there may be existing pipewire-session-manager.service alias symlink that may (or may not) block creating one for wireplumber.service instead.

PR #22769, opened 2021-10-31
sci-mathematics/petsc: addpredict /dev/kfd

Alessandro-Barbieri:
Closes: https://bugs.gentoo.org/810841


AndrewAmmerlaan:
I can't reproduce this issue (with USE="X fortran"), could you find exactly which USE flag introduces the problem and put the `addpredict` command behind that flag?

AndrewAmmerlaan:
> That's because you don't have rocm installed

Aah right, that makes sense

PR #22779, opened 2021-11-01
dev-python/cherrypy: Add the new CherryPy website to the metadata

webknjaz:
This patch resurrects a link to the official CherryPy website originally
deleted by @mgorny in https://github.com/gentoo/gentoo/commit/c8b3d98d.

It is now https://cherrypy.dev.
Ref: https://twitter.com/cherrypy/status/1435583865024823297

Signed-Off-By: Sviatoslav Sydorenko <webknjaz@redhat.com>

### DIFF
--- a/dev-python/cherrypy/cherrypy-18.6.1.ebuild
+++ b/dev-python/cherrypy/cherrypy-18.6.1.ebuild
@@ -9,7 +9,7 @@ inherit distutils-r1
 MY_P="CherryPy-${PV}"
 
 DESCRIPTION="CherryPy is a pythonic, object-oriented HTTP framework"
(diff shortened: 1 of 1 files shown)


arthurzam:
For including your commit in tree, please:

1.

arthurzam:
For including your commit in tree, please:

1. Use "Signed-off-by:" in commit (tools are case sensitive)
2. Use package prefix in commit title, for example `dev-python/cherrypy: update HOMEPAGE`

thesamesam:
(pkgdev commit or repoman commit will get those bits right btw, makes life a lot easier)

arthurzam:
Thank you very much for the contribution.

PR #22780, opened 2021-11-01
Add an ebuild for aiohttp v3.8.0

webknjaz:
This is currently a draft because it's untested and depends on the `frozenlist` library that used to be a part of aiohttp itself but is now a separate package and needs to be packaged too.

Signed-Off-By: Sviatoslav Sydorenko <webknjaz@redhat.com>


mgorny:
I'm working on it for a few hours already.

webknjaz:
Linking https://github.com/gentoo/gentoo/commit/320b8fba for history.

PR #22796, opened 2021-11-02
dev-libs/openssl: add `fips` support

tormath1:
`FIPS` provider is not enabled by default with OpenSSL version 3. Let's
make it optional by adding conditional `fips` internal useflag.

See also: https://github.com/openssl/openssl/blob/master/README-FIPS.md

Bug: https://bugs.gentoo.org/820173


SoapGentoo (approved):
yes let's do it

thesamesam:
Thank you!

PR #22797, opened 2021-11-02
x11-drivers/nvidia-drivers: revamp 495.44 ebuild (wip)

ionenwks:
Attempt at ~complicating~ simplifying things by replicating nvidia-installer and using the .manifest file rather a full manual install (plus some other small changes).

Just throwing this here to ponder whether to use this or not... feedback welcome if anyone happen to look.

pros:
- no need to maintain a list of files to install otherwise requiring close inspection every bumps
- no need to run scanelf on every libraries (removes pax-utils dep fwiw)
- more precise installation (no generic library symlinks)
- easier compatibility with x86/aarch64's .run, if something is amd64-only then it won't be in the .manifest
- mostly-in-one-place list of install paths (top of src_install along with arrays to skip files)
- no more multilib eclass or figuring which libraries have compat32 or not
- inherits fit on one line :eyes:

cons:
- feels like a bad hack and not ebuild-like
- ebuild only barely smaller
- still need some workarounds to handle .manifest inconsistencies that feel a bit fragile
- future maintainers may not know how to handle this whenever it'll need updates, may feel like they are trying to fix kernel-2.eclass
- if partially breaks, files could silently go missing (previously it had to be 1:1 with ebuild save for docs)

### DIFF
--- a/x11-drivers/nvidia-drivers/files/nvidia-settings-390.144-desktop.patch
+++ b/x11-drivers/nvidia-drivers/files/nvidia-settings-390.144-desktop.patch
@@ -1,0 +1,12 @@
+For icon: https://github.com/NVIDIA/nvidia-settings/pull/56
+--- a/nvidia-settings/doc/nvidia-settings.desktop
++++ b/nvidia-settings/doc/nvidia-settings.desktop
+@@ -5,5 +5,5 @@
+ Comment=Configure NVIDIA X Server Settings
+-Exec=__UTILS_PATH__/nvidia-settings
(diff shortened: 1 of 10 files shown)


ionenwks:
ended up using this ebuild type in commit 0023d464ce869d0d80fc7847dea9ed40a348d1d4, also present in 5c7482ce04446808e1d4fd698dcc1c6202c6145a

390.xx and 460.xx may receive a revbump at a later date if no new version soon, but they're both in need of fix for new xorg so I'm /hoping/ nvidia releases new versions.

PR #22801, opened 2021-11-03
dev-util/gn: fix build error on ARCH=riscv

dlan17:
Closes: https://bugs.gentoo.org/821400
Signed-off-by: Yixun Lan <dlan@gentoo.org>


dlan17:
fixed in version 0.1943, so close this

PR #22806, opened 2021-11-03
sci-libs/rocSPARSE: fix matrices dir for tests

littlewu2508:
Tests using data will be skipped, and several test suites which only
contains data driven tests will be classified as
UninstantiatedParameterizedTestSuite by gtest.

Closes: https://github.com/ROCmSoftwarePlatform/rocSPARSE/issues/227


heroxbd:
There is an -r1 ebuild in the repository, why are we changing the old one?

I don't understand `Tests using data will be skipped`.  You have added test data to `SRC_URI`, but they are skipped according to your `git commit`.  What's the rationale behind?

littlewu2508:
> 
> 
> There is an -r1 ebuild in the repository, why are we changing the old one?
> 
> I don't understand `Tests using data will be skipped`. You have added test data to `SRC_URI`, but they are skipped according to your `git commit`. What's the rationale behind?

SRC_URI have no problem -- the download and unpack stage is correct. But when running `src_prepare`, those testing data, which should be transformed by `deps/convert`, are converted to the wrong place, which is my mistake when writing the first version of rocSPARSE.

PR #22818, opened 2021-11-04
package.mask: re-mask bitcoin >=0.21.1

luke-jr:
Without a mask, automatic upgrades would trigger, which is unethical (at least at this time).

This reverts commit ae7251a476e1ea18a429bf012dcdf45ec59662a6.

### DIFF
--- a/profiles/package.mask
+++ b/profiles/package.mask
@@ -636,6 +636,18 @@ net-libs/libsoup:3.0
 # breaks if not all at least many revdeps. #805011 for tracker bug.
 >=net-libs/mbedtls-3.0.0
 
+# Luke Dashjr <luke-jr+gentoobugs@utopios.org> (2021-11-04)
+# This release adds enforcement of the Taproot protocol change to the Bitcoin
+# rules, beginning in November. Protocol changes require user consent to be
+# effective, and if enforced inconsistently within the community may compromise
+# your security or others! If you do not know what you are doing, learn more
+# before November. (You must make a decision either way - simply not upgrading
+# is insecure in all scenarios.)
+# To learn more, see https://bitcointaproot.cc
+>=net-p2p/bitcoind-0.21.1
+>=net-p2p/bitcoin-qt-0.21.1
+>=net-libs/libbitcoinconsensus-0.21.1
+
 # Lars Wendler <polynomial-c@gentoo.org> (2021-07-10)
 # Masked for testing
 # bug #802186


luke-jr:
Quick fix to avoid users being upgraded without actively consenting.

Later, we can discuss other solutions (see also #21490 for some prior discussion)

gyakovlev:
@candrews FYI

merged since sounds like a big deal.
please find consensus on this bitcoin stuff guys.

and just a disclaimer again: I have no opinion on this topic or cryptocurrency in general, 
just merging this PR as a request due to time-sensitivity.

candrews:
NACK.

@gyakovlev I _strongly_ disagree that this mask should be preserved.

Please see https://github.com/gentoo/gentoo/pull/21490 where it was agreed that the mask would be removed as November approached - I removed that mask as agreed.

Also, this is _not_ an emergency and this mask _should not have been re-added_ so quickly, especially since I'm listed as co-maintainer and had not yet responded.

Taproot is happening. There is no stopping it. And I'm not seeing any ethical argument here. The bitcoin community talked about taproot for over a year, it was scheduled for implementation through a community driven approval process, and _it is happening_. As far as I can tell, this is all unjustified FUD and it either it needs to stop, or real evidence needs to be cited and raised with the bitcoin project (and honestly, the media outlets).

candrews:
By blocking upgrades to taproot supporting version of bitcoin, _we are hurting our users_.

Taproot happens on November 16; there is no stopping it.

> What happens if I don't upgrade in time?
> If you have not upgraded before Taproot activates (around November 12th), your node will no longer be fully secure and you may become vulnerable to certain attacks. In such a case, your node would be trusting and relying on miners to protect you until you do upgrade.

-- https://bitcointaproot.cc/

luke-jr:
>Please see #21490 where it was agreed that the mask would be removed as November approached - I removed that mask as agreed.

That was not agreed on. In fact, I explicitly rejected your suggestion to do so Nov 1st and explained why it was unethical, and you changed it to imply it was not decided yet. Removing it should have been discussed and done only when ethical to do so (it isn't today).

>Also, this is not an emergency and this mask should not have been re-added so quickly, especially since I'm listed as co-maintainer and had not yet responded.

You're not... and I don't think I would be comfortable with adding you at this point. This is the second time you've done this.

>Taproot is happening. There is no stopping it.

This is false. If users do not choose to enforce it (which requires they make that choice), Taproot doesn't happen.

> By blocking upgrades to taproot supporting version of bitcoin, we are hurting our users.

The intent is not to block, but to ensure users actively consent. That is why they are advised to make a decision and unmask it themselves.

Better (IMO) alternatives to ensure user opt-in were suggested, but this is the method we ended up using this time. Next time, we should use something cleaner. But not getting user consent at all, is not ethical.

gyakovlev:
like I said - no opinion here from me on switch or not.
mask serves a purpose of a message to user, true.

also consider creating a news item. there has been a lot of points raised last time this discussion happened.

looks like there's still time.

floppym:
@candrews You are not named as a "co-maintainer" on any of these packages. You are a member of the proxy-maint project, but I don't think that really makes you a co-maintainer. If you really want to take responsibility for bitcoin, add yourself to metadata.xml please.

PR #22819, opened 2021-11-04
Bitcoin: Add warning to soon-to-be-obsolete versions

luke-jr:


juippis:
@candrews is this correct? I haven't followed this bitcoin mess at all.

luke-jr:
It was correct 13 days ago. Now it's superceded by #22954

PR #22820, opened 2021-11-04
Bump KiCad to 5.1.11

zpuskas:
- sci-electronics/kicad: Bump to 5.1.11
- sci-electronics/kicad-footprints: Bump to 5.1.11
- sci-electronics/kicad-i18n: Bump to 5.1.11
- sci-electronics/kicad-packages3d: Bump 5.1.11
- sci-electronics/kicad-symbols: Bump to 5.1.11
- sci-electronics/kicad-templates: Bump to 5.1.11
- app-doc/kicad-doc: Bump to 5.1.11
- sci-electronics/kicad-meta: Bump to 5.1.11


juippis (approved):
We really dislike `git mv`ing version bumps like this. If the old version has to be returned, reverting the ebuild would now also revert the version bumps. Also, many of these ebuilds, if not all, could already be migrated to EAPI-8. 

So please next time split adding a version and removing an old version to their separate commits. 

Everything seems to work so thanks for your time!

dflogeras:
Hi all, a little late to the party but have a look at:

https://www.kicad.org/blog/2021/11/KiCad-5.1.12-Release/

Especially "If you are wondering why there was no official 5.1.11 stable release, it’s because there was an issue with the 5.1.11 tagged code which some users downloaded before the actual release announcement. This required a fix and a new release tag."

zpuskas:
@juippis Thanks for the heads up, I'll drop old versions in separate commits in the future.
@dflogeras Thanks for the heads up, will submit a new PR soon with 5.1.12.

PR #22822, opened 2021-11-05
dev-libs/ntl: update bundled config.guess automatically

alexfanqi:
upstream version of config.guess and config.sub is a bit outdated. This makes it so that both files get updated to latest in src_unpack. I have tested on amd64 and riscv. 
Closes: https://bugs.gentoo.org/821061
Signed-off-by: Alex Fan <alexfanqi@yahoo.com>

### DIFF
--- a/dev-libs/ntl/ntl-11.4.4-r1.ebuild
+++ b/dev-libs/ntl/ntl-11.4.4-r1.ebuild
@@ -1,0 +1,71 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit toolchain-funcs gnuconfig
+
+DESCRIPTION="High-performance and portable C++ number theory library"
+HOMEPAGE="https://www.shoup.net/ntl/ https://github.com/libntl/ntl"
+SRC_URI="https://www.shoup.net/ntl/${P}.tar.gz"
+
+LICENSE="LGPL-2.1+"
+SLOT="0/43"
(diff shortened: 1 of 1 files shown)


dlan17:
this looks good to me.. @orlitzky can you review this?

dev-libs/ntl/ntl-11.4.4.ebuild
```diff
@@ -26,6 +26,11 @@ REQUIRED_USE="bindist? ( !cpu_flags_x86_avx2 )"
 
 DOCS=( "${WORKDIR}/${P}"/README )
 
+src_unpack() {
+	default
+	gnuconfig_update "$S/libtool-origin/"
```
AndrewAmmerlaan, line 31:
```suggestion
	gnuconfig_update "${S}/libtool-origin/"
```

dev-libs/ntl/ntl-11.4.4.ebuild
```diff
@@ -26,6 +26,11 @@ REQUIRED_USE="bindist? ( !cpu_flags_x86_avx2 )"
 
 DOCS=( "${WORKDIR}/${P}"/README )
```
AndrewAmmerlaan, line 27: This is probably unnecessary because `einstalldocs` automatically installs README

AndrewAmmerlaan: Aaah no it isn't, never mind

orlitzky (changes requested):
This ebuild is marked stable... can you do these changes in an -r1? Otherwise LGTM.

alexfanqi:
Thanks for taking time to review this. Should I mark all arches as ~ testing in the -r1 ebuild?

orlitzky:
Merged, thanks!

PR #22823, opened 2021-11-05
elastic: bump to 7.15.1

hydrapolic:
Tested on amd64.


app-admin/filebeat/files/filebeat-glibc-2.34.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/libbeat/common/seccomp/policy_linux_386.go b/libbeat/common/seccomp/policy_linux_386.go
```
thesamesam, line 1: When you can, pop a link to the upstream bug/commit(s) at the top of patches

hydrapolic:
Thank you @thesamesam

app-admin/filebeat/files/filebeat-glibc-2.34.patch
hydrapolic: I put it into the commit message, but it can be here too.

PR #22804, opened 2021-11-03
sci-libs/hipFFT: ROCm FFT marshalling library

littlewu2508:
hipFFT is the front end of rocFFT, and is dependency of ROCm
supported math/DL frameworks like pytorch.

Package-Manager: Portage-3.0.20, Repoman-3.0.3
Signed-off-by: Yiyang Wu <xgreenlandforwyy@gmail.com>

PR #22824, opened 2021-11-05
dev-util/rocm-clang-ocl: OpenCL compilation with clang compiler.

littlewu2508:
This small package migrated from science overlay is the dependency
of sci-libs/MIOpen (currently in ::science; prepared for ::gentoo)


heroxbd:
As for https://github.com/gentoo/gentoo/pull/22804 , better ask @justxi for permission and decide if you would also like to be one of the maintainers of this package.

justxi:
You can add the ebuild and you can add me as a maintainer.

PR #22825, opened 2021-11-05
dev-libs/half: Half-precision floating-point library

littlewu2508:
Migrated from science overlay, this package contains only
one header -- half.hpp.
sci-libs/miopen, AMD's Machine Intelligence Library for ROCm,
depend on this package.


heroxbd:
Same as https://github.com/gentoo/gentoo/pull/22804, need acknowledgement from @justxi .

Would you like to be one of the maintainers?

justxi:
I agree to add the ebuild and would like to become a maintainer.

dev-libs/half/half-1.12.0.ebuild
```diff
@@ -0,0 +1,17 @@
...
+DESCRIPTION="Half-precision floating-point library"
+HOMEPAGE="http://half.sourceforge.net/"
+SRC_URI="https://github.com/ROCmSoftwarePlatform/half/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+KEYWORDS="~amd64"
+SLOT="0"
+
+src_install() {
+	cd include
```
heroxbd, line 15: `|| die` here.

dev-libs/half/half-2.1.0.ebuild
```diff
@@ -0,0 +1,20 @@
...
+S="${WORKDIR}"
+
+LICENSE="MIT"
+KEYWORDS="~amd64"
+SLOT="0"
+
+BDEPEND="app-arch/unzip"
+
+src_install() {
+	cd include
```
heroxbd, line 18: `|| die` here.

heroxbd (changes requested):

dev-libs/half/half-2.1.0.ebuild
```diff
@@ -0,0 +1,20 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DESCRIPTION="Half-precision floating-point library"
+HOMEPAGE="http://half.sourceforge.net/"
+SRC_URI="https://downloads.sourceforge.net/project/half/half/${PV}/${P}.zip"
```
heroxbd, line 8: sourceforge `SRC_URI` should use `mirror://sourceforge/` instead.  cf. https://wiki.gentoo.org/wiki//etc/portage/mirrors and portage(5) manpage.

heroxbd (changes requested):
I want to understand the purpose of the 2 versions here.

One is from https://github.com/ROCmSoftwarePlatform/half. version 1.12.0.  The other is from sourceforge, version 2.1.0.  

Are there any ROCm specific changes needed in https://github.com/ROCmSoftwarePlatform/half?  Can the sourceforge half-2.1.0 meet the reverse dependencies?  Why do we need source code from 2 repositories?

littlewu2508:
> I want to understand the purpose of the 2 versions here.
> 
> One is from https://github.com/ROCmSoftwarePlatform/half. version 1.12.0. The other is from sourceforge, version 2.1.0.
> 
> Are there any ROCm specific changes needed in https://github.com/ROCmSoftwarePlatform/half? Can the sourceforge half-2.1.0 meet the reverse dependencies? Why do we need source code from 2 repositories?

The original half package on sourceforge is now version 2.x, but MIOpen requires version 1.12.0 (See https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/requirements.txt). Compilation error occurred when building miopen against half-2.1.

The ROCm forked version, half-1.12.0, contains two more commit compared to the original release: [one](https://github.com/ROCmSoftwarePlatform/half/commit/cc5beae0e03a27e9e8b68c9524bbc3f0cde772b0) for C++17, [the other](https://github.com/ROCmSoftwarePlatform/half/commit/87d7f25f7ba2c7d3b051f6c857031de0ecac5afd) for gcc.

littlewu2508:
I removed dev-libs/half-2.1.0 since it's not a dependency of miopen. Subslot `0/1` is added to half-1.12.0 so miopen can still depend on half-1.12.0 if half-2.x is added in the future.

PR #22832, opened 2021-11-05
sci-libs/opencascade-7.5.{2,3} - restrict dev-cpp/tbb dependency

waebbl:
sci-libs/opencascade-7.5.2-r2: revbump
Version tbb-2021.4.0 has some header files removed, so we restrict to
earlier versions.
Bug: https://bugs.gentoo.org/821655

sci-libs/opencascade-7.5.3-r3: revbump
Version tbb-2021.4.0 has some header files removed, so we need to
restrict to earlier versions.
Closes: https://bugs.gentoo.org/821655

sci-libs/opencascade-7.5.2-r1: drop revision
sci-libs/opencascade-7.5.3-r2: drop revision


thesamesam:
Not that it matters really but I’m happy for you to git mv for stuff like this (especially given it makes the diff easier to read). Thank you for doing this, will look shortly!

waebbl:
I try to think about it next time. Usually I prefer to use two atomic commits, even for such trivial things like revbumps. It makes me feel better, in case anything goes wrong and I have to go back in the history.
But yes, the diff is harder to catch, especially in the shell. Luckily GitHub does a good job in recognizing the actual move and shows a nice clean diff.

PR #22833, opened 2021-11-05
Update glibc to make libcrypt.so.1 an orphaned file

floppym:


ZeroChaos-:
This should help avoid failures with FEATURES="-preserve-libs protect-owned"
If this (or similar) isn't accepted, then a pkg_setup die should be added as protect-owned will prevent libxcrypt from installing after the -preserve-libs hack in the glibc ebuild

floppym:
There's a risk involved here: `/lib/libcrypt.so.1` will deleted for the short period between pkg_preinst and pkg_postinst.

ZeroChaos-:
Thanks!

PR #22847, opened 2021-11-06
dev-java/jaxb-api: hide tests for jdk-8

vaukai:
Closes: https://bugs.gentoo.org/796995


fordfrog:
@vaukai thanks! i did few changes to your commit. first, as this is a fix and there are no changes to the generated jar, no revbump is needed so i replaced the existing ebuild instead. and i also added reference to the related bug to your comment in the ebuild for future reference.

PR #22851, opened 2021-11-06
net-p2p/mldomkey: Add systemd service unit and remove obsolete settings from the OpenRC script

chusopr:
Sorry for the long PR message, but I want to unroll all the bits in this PR.

* It adds a systemd service unit.
  This service had a complex logic for its OpenRC script and a bunch of settings in a conf.d file which would be quite hard to port to systemd.
  Instead, I realized it would be much easier and convenient to simplify the OpenRC script and have a similar behaviour with OpenRC and systemd.
* The way the OpenRC script worked so far to stop the service was by sending the [`close_fds`](http://mldonkey.sourceforge.net/Close_fds), [`save`](http://mldonkey.sourceforge.net/Save) and [`kill`](http://mldonkey.sourceforge.net/Kill) commands to the process, first via telnet and then via HTTP, as an attempt to orderly shut it down before failing back to sending it a SIGTERM.
  However, this is unnecessary. Just sending a SIGTERM works because the process is able to catch it and orderly shutdown itself saving its configurations:
  ```
  2021/10/31 14:34:48 [dMain] Received SIGTERM, stopping MLDonkey...
  2021/10/31 14:34:49 [cCO] Options correctly saved
  2021/10/31 14:34:49 [cCO] Sources correctly saved
  2021/10/31 14:34:49 [cCO] Options backup as .tar.gz correctly saved
  2021/10/31 14:34:49 [dMain] Core stopped
  ```
  So, the whole telnet / HTTP complexity is now removed and now the service is stopped with a SIGTERM instead like any other regular service.
  This was made not only for the sake of simplicity but also because of the concerns raised (and discussed on #gentoo-dev-help) about calling wget and netcat from an OpenRC script and storing plaintext passwords in a conf.d file for the service to be able to use them to send those commands.
* By removing all netcat / HTTP stuff, the custom OpenRC commands `slow`, `fast` and `info` are also removed.
  Those commands were used to toggle MLDonkey between "slow" and "high" bandwidth configurations.
  However, this was implemented on OpenRC side conflicting with MLDonkey's own mechanism for this.
  MLDonkey supports two bandwidth configuration presets (`max_hard_upload_rate`, `max_hard_download_rate` and `max_opened_connections`, and `max_hard_upload_rate2`, `max_hard_download_rate2` and `max_opened_connections2`) and can switch between one or the other with the `bw_toggle` command.
  The OpenRC script ignored all that (or maybe it was not yet implemented on MLDonkey's side when the OpenRC script was made) and implemented its own bandwidth toggle mechanism conflicting with MLDonkey's own one by altering the values of `max_hard_upload_rate` and `max_hard_download_rate` to `LOW_UP` and `LOW_DOWN` values from the conf.d file respectively if the low preset was chosen or `HIGH_UP` and `HIGH_DOWN` for the high one, overriding MLDonkey's own ability to manage two bandwidth configuration presets.
* It has been changed how `MLDONKEY_DIR` conf.d setting is used.
  So far, the OpenRC script was first checking whether the directory existed. If it existed, it was starting MLDonkey there considering it an existing MLDonkey data directory. If it didn't exist, it created it and started MLDonkey there with the intention of using it as a new data directory for MLDonkey.
  This no longer works this way from MLDonkey version 2.6.2 (released 16 years ago!): https://github.com/ygrek/mldonkey/blob/master/docs/ChangeLog.old#L2459
  MLDonkey won't use an empty directory to store its data as the OpenRC script expects.
  Now, if MLDonkey doesn't find a `donwload.ini` file in the current directory, it will fail back to `~/.mldonkey` and won't use the directory where the process was started as it has been doing before.
  So, setting `MLDONKEY_DIR` to a non-existing directory with the hope that will be created and used to place MLDonkey files won't give the expected result as the OpenRC script was trying so far.
  The init script was modified to check if `MLDONKEY_DIR` was set and fail if it's set to a non-existing directory.
* For a similar reason, `acct-user/p2p` was modified to have a real home directory, otherwise new installations would fail trying to create ~/.mldonkey:
  ```
  [cO] MLDonkey is working in /dev/null/.mldonkey
  [cO] creating new MLDonkey base directory in /dev/null/.mldonkey
  [cO] Exception (Failure("/dev/null already exists but is not a directory")) trying to create dir /dev/null/.mldonkey
  ```
  I don't really like using `/home/p2p` as its home directory and this actually triggered a QA notice:
  ```
   * 
   *   /home
   * 
  ```
  But it was kept this way for consistency with the previous behaviour. Comments are welcome though.
* Lastly, since many options were removed from the conf.d file because they are no longer used (either because of the removed LOW/HIGH functionality or because they were required for wget / netcat) and, especially, since one of them was sensitive (the administrator password), both the ebuild and the OpenRC script will check for those retired settings and print a warning if they are still found in the conf.d file.

I know we don't like bundling more than one change per commit, but as you can see, all of them are interconnected.

### DIFF
--- a/net-p2p/mldonkey/mldonkey-3.1.7-r3.ebuild
+++ b/net-p2p/mldonkey/mldonkey-3.1.7-r3.ebuild
@@ -3,7 +3,7 @@
 
 EAPI="7"
 
-inherit autotools desktop flag-o-matic toolchain-funcs
+inherit autotools desktop flag-o-matic systemd toolchain-funcs
 
 DESCRIPTION="Multi-network P2P application written in Ocaml, with Gtk, web & telnet interface"
 HOMEPAGE="http://mldonkey.sourceforge.net/ https://github.com/ygrek/mldonkey"
@@ -17,7 +17,7 @@ IUSE="bittorrent doc fasttrack gd gnutella gtk guionly magic +ocamlopt upnp"
 
 REQUIRED_USE="guionly? ( gtk )"
 
-COMMON_DEPEND="dev-lang/perl
+RDEPEND="dev-lang/perl
 	dev-ml/camlp4:=
 	gd? ( media-libs/gd[truetype] )
 	gtk? (
@@ -35,12 +35,10 @@ COMMON_DEPEND="dev-lang/perl
 	)
 	!guionly? ( acct-user/p2p )
 "
-RDEPEND="${COMMON_DEPEND}
-	|| ( net-analyzer/netcat net-analyzer/openbsd-netcat )"
 # Can't yet use newer OCaml
 # -unsafe-string usage:
 # https://github.com/ygrek/mldonkey/issues/46
-DEPEND="${COMMON_DEPEND}
+DEPEND="${RDEPEND}
 	<dev-lang/ocaml-4.10:=[ocamlopt?]
 	bittorrent? (
 		|| (
@@ -141,8 +139,8 @@ src_install() {
 		done
 		use bittorrent && newbin "make_torrent${myext}" make_torrent
 
+		systemd_dounit "${FILESDIR}/${PN}.service"
 		newconfd "${FILESDIR}/mldonkey.confd" mldonkey
-		fperms 600 /etc/conf.d/mldonkey
 		newinitd "${FILESDIR}/mldonkey.initd" mldonkey
 	fi
 
@@ -171,16 +169,9 @@ src_install() {
 }
 
 pkg_postinst() {
-	if ! use guionly; then
-		echo
-		einfo "If you want to start MLDonkey as a system service, use"
-		einfo "the /etc/init.d/mldonkey script. To control bandwidth, use"
-		einfo "the 'slow' and 'fast' arguments. Be sure to have a look at"
-		einfo "/etc/conf.d/mldonkey also."
-		echo
-	else
-		echo
-		einfo "Simply run mlgui to start the chosen MLDonkey gui."
-		einfo "It puts its config files into ~/.mldonkey"
+	if [ -f /etc/conf.d/mldonkey ] && grep -qE "^(BASEDIR|SUBDIR|LOW_DOWN|LOW_UP|HIGH_DOWN|HIGH_UP|SERVER|PORT|TELNET_PORT|USERNAME|PASSWORD|MLDONKEY_TIMEOUT)=" /etc/conf.d/mldonkey; then
+		ewarn "The following settings are deprecated and will be ignored,"
+		ewarn "please remove them from /etc/conf.d/mldonkey:"
+		ewarn "LOW_DOWN LOW_UP HIGH_DOWN HIGH_UP SERVER PORT TELNET_PORT USERNAME PASSWORD MLDONKEY_TIMEOUT"
 	fi
 }
--- a/net-p2p/mldonkey/files/mldonkey.confd
+++ b/net-p2p/mldonkey/files/mldonkey.confd
@@ -9,7 +9,7 @@
 USER="p2p"
 
 # mldonkey's working dir (must be existing)
-MLDONKEY_DIR="/home/p2p/mldonkey"
+MLDONKEY_DIR="/var/lib/mldonkey"
 
 # logfile (/dev/null for nowhere)
 LOG="/var/log/mldonkey.log"
@@ -18,25 +18,5 @@ LOG="/var/log/mldonkey.log"
 #     set log_to_syslog true
 USE_LOGGER=false
 
-# bandwidth control, values in kb/s
-LOW_DOWN="6"
-LOW_UP="2"
-HIGH_DOWN="30"
-HIGH_UP="10"
-
 # nice level, 0<x<19, more nice -> less cpuspeed consumed
 NICE="19"
-
-# mldonkey server ip, usually localhost
-SERVER="localhost"
-
-# port for webinterface, usually 4080
-PORT="4080"
-
-# port for telnet interface, usually 4000
-TELNET_PORT="4000"
-
-# to enable password restricted access,
-# uncomment and set BOTH following vars:
-#USERNAME="admin"
-#PASSWORD=""
--- a/net-p2p/mldonkey/files/mldonkey.initd
+++ b/net-p2p/mldonkey/files/mldonkey.initd
@@ -3,7 +3,6 @@
 extra_started_commands="reload slow fast info"
 
 MLDONKEY_BINARY=${MLDONKEY_BINARY:-/usr/bin/mlnet}
-MLDONKEY_TIMEOUT=${TIMEOUT:-10}
 
 depend() {
 	need localmount net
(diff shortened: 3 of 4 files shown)


juippis (commented):
I was hoping someone with more knowledge about systemd took a look...

I don't know how I feel about the acct-user/p2p change. I feel like this directory should be controlled via init.d/service files 100 %. That's AFAIK how it should be done too. You can write bash, or well, POSIX tests in openrc files to check whether the directory exists, and create it with correct perms if not. I don't know how to do this with systemd, but doubt it's impossible.

chusopr:
Hi, @juippis, thanks for your review.

Yes, as I said in my original comment, I didn't like that very much either, but that's just the behaviour defined upstream: if you try to start MLDonkey process from an empty directory, it will fail back to `~/.mldonkey`. So we needed to make `~/.mldonkey` writable and that meant `p2p` user needed to have a real home directory.

But I just saw [this comment in the ChangeLog](https://github.com/ygrek/mldonkey/blob/master/docs/ChangeLog.old#L2468) when this change was introduced:

> * if variable MLDONKEY_DIR is used that directory has priority,
>   to simulate the behaviour without this patch use:
>     export MLDONKEY_DIR="." && ./mlnet

That means a workaround may be available for a behaviour that works better for us.

I will temporarily retire this PR as a draft to test that solution.

juippis:
Let us know when you've tested it, what you found, and it's ready to "go live".
Thanks for being thorough!

chusopr:
I just tested that solution and it works with simple modifications to the OpenRC and systemd scripts.

I just pushed a new version, these are the differences with the original:

* `MLDONKEY_DIR` reverted to the original value `/home/p2p/mldonkey` instead of `/home/p2p/.mldonkey` now that I found the way to make custom directories work even if they are empty.
* Made required changes for a custom working directory to work by using the `MLDONKEY_DIR` environment variable:
  * Added `Environment=MLDONKEY_DIR=/home/p2p/mldonkey` to the systemd service file.
  * Added `--env "MLDONKEY_DIR=${MLDONKEY_DIR}"` to `start-stop-daemon` when called from OpenRC.
* Removed commit making changes to `acct-user/p2p`.
* Recovered code in the OpenRC script that creates `${MLDONKEY_DIR}` if it doesn't exist.
  This cannot be done for the systemd service because it already runs as the `p2p` user, so I guess it will be left to the user to make sure they set `MLDONKEY_DIR` to a directory that exists and is writable.
* Removed code in the OpenRC script that made it exit with failure if `${MLDONKEY_DIR}` was not a valid pre-existing non-empty MLDonkey data directory.
* Added missing inherit for the `systemd` eclass

Both systemd and OpenRC were tested as working as intended with this change.

Now I'm wondering if it makes sense keeping `/home/p2p/mldonkey` as the default `MLDONKEY_DIR`, especially given that the home of the `p2p` user is `/dev/null` and not `/home/p2p`.

Maybe I could change `MLDONKEY_DIR` to something that made more sense like `/var/lib/mldonkey`.
Existing users shouldn't be affected because so far `MLDONKEY_DIR` was only set in `/etc/conf.d/mldonkey`. so it will be protected by `config-protect`.

juippis:
(I think you said in IRC you'll update this and remove the draft marking - let us know when it's ready)

chusopr:
Yes, I said so in IRC to change the default data dir from `/home/p2p/mldonkey` to something more suitable like `/var/lib/mldonkey` as discussed with sam_ there.

I just updated the PR now and removed the draft marking.

net-p2p/mldonkey/mldonkey-3.1.7-r3.ebuild
```diff
@@ -141,9 +139,12 @@ src_install() {
...
 
-		newconfd "${FILESDIR}/mldonkey.confd" mldonkey
-		fperms 600 /etc/conf.d/mldonkey
-		newinitd "${FILESDIR}/mldonkey.initd" mldonkey
+		if use systemd; then
+			systemd_dounit "${FILESDIR}/${PN}.service"
+		else
+			newconfd "${FILESDIR}/mldonkey.confd" mldonkey
+			newinitd "${FILESDIR}/mldonkey.initd" mldonkey
+		fi
```
juippis, line 147: We always install the service files regardless of init system that's being used.

juippis (commented):
I can't compile this :( 
```
signals_nat.c:185:13: error: variably modified 'sig_alt_stack' at file scope
  185 | static char sig_alt_stack[SIGSTKSZ];
      |             ^~~~~~~~~~~~~
make[3]: *** [Makefile:343: signals_nat_n.o] Error 1
make[3]: *** Waiting for unfinished jobs....
make[3]: Leaving directory '/var/tmp/portage/dev-lang/ocaml-4.09.0-r2/work/ocaml-4.09.0/runtime'
make[2]: *** [Makefile:1004: makeruntimeopt] Error 2
make[2]: Leaving directory '/var/tmp/portage/dev-lang/ocaml-4.09.0-r2/work/ocaml-4.09.0'
make[1]: *** [Makefile:417: opt.opt] Error 2
make[1]: Leaving directory '/var/tmp/portage/dev-lang/ocaml-4.09.0-r2/work/ocaml-4.09.0'
make: *** [Makefile:468: world.opt] Error 2
 * ERROR: dev-lang/ocaml-4.09.0-r2::gentoo failed (compile phase):
 *   emake failed
```

There seems to be some workaround in https://github.com/ygrek/mldonkey/issues/46 so maybe we can use later ocaml now? Also not that I understand anything about ocaml, but I see `-warn-error` in ocaml's build flags, is that the equivalent of `-Werror`?

ionenwks:
> signals_nat.c:185:13: error: variably modified 'sig_alt_stack' at file scope
>   185 | static char sig_alt_stack[SIGSTKSZ];
>       |             ^~~~~~~~~~~~~

For the record, these are a common glibc-2.34 issue and more particularly is https://bugs.gentoo.org/804498 (so stable users are okay for now, but glibc-2.34 is headed toward stabilization) -- if possible using a ocaml that's fixed would indeed be nicer, I'd imagine Gentoo would rather get rid of the old than fix it (maybe @thesamesam has input, I don't know much on that topic either)

net-p2p/mldonkey/mldonkey-3.1.7-r3.ebuild
chusopr: Oh, OK, I didn't know that. So the ebuild should install both OpenRC and systemd service files unconditionally?
In that case I don't need `systemd` in `IUSE` because it's only used here.

chusopr:
> There seems to be some workaround in [ygrek/mldonkey#46](https://github.com/ygrek/mldonkey/issues/46) so maybe we can use later ocaml now?

If you see [@turolomio's comment there](https://github.com/ygrek/mldonkey/issues/46#issuecomment-767618446), he was able to compile net-p2p/mldonkey with >=dev-lang/ocaml-4.10 by using a modified ebuild for dev-lang/ocaml that disables unsafe strings.

I doubt very much that we will be willing to disable safe strings in >=dev-lang/ocaml-4.10 just to make net-p2p/mldonkey compile.

> if possible using a ocaml that's fixed would indeed be nicer, I'd imagine Gentoo would rather get rid of the old than fix it (maybe @thesamesam has input, I don't know much on that topic either)

Yes, there is a bug tracking all packages that still don't work with safe strings (i.e., >=dev-lang/ocaml-4.10) that need to be fixed and I think net-p2p/mldonkey is the only one that is still pending: https://bugs.gentoo.org/749606

There is no much activity upstream other than small fixes and the safe string fix would require a big change to the code, so I see more likely that net-p2p/mldonkey is dropped from Gentoo than it is ported to safe strings upstream.

But I think all this is out of the scope of this PR, isn't it? And there is a [different ticket](https://bugs.gentoo.org/749609) for the >=dev-lang/ocaml-4.10 safe strings issue in net-p2p/mldonkey.

net-p2p/mldonkey/mldonkey-3.1.7-r3.ebuild
chusopr: Fixed.

juippis:
> But I think all this is out of the scope of this PR, isn't it? And there is a [different ticket](https://bugs.gentoo.org/749609) for the >=dev-lang/ocaml-4.10 safe strings issue in net-p2p/mldonkey.

I mean, I get you. We're fixing a systemd service file here, but I can't compile this on ~amd64. You revbump, rightfully, so everyone are going to get hit by the compilation error and no one can update (if it is about the glibc). We need to address that somehow before pushing this new version. I was hoping upstream would have some patch for the glibc error, but no luck?

thesamesam:
A bug should be filed for earlier OCamls needing the glibc fix, but I see the problem anyway. Looking at it...

thesamesam:
Fixed now, but do file bugs so I don't miss things if you need something fixed in another package.

chusopr:
So I guess this PR is now ready for review again? Or is it awaiting anything else from my side?

juippis (approved):
Sorry, was barely online last week. Thanks for your patience and @thesamesam for fixing the issue hastily!

PR #22853, opened 2021-11-07
dev-lang/python: add PGO

thesamesam:
TODO: use the full PGO target

Bug: https://bugs.gentoo.org/615412

### DIFF
--- a/dev-lang/python/python-3.10.0_p1.ebuild
+++ b/dev-lang/python/python-3.10.0_p1.ebuild
@@ -224,8 +224,7 @@ src_test() {
 	# bug 660358
 	local -x COLUMNS=80
 	local -x PYTHONDONTWRITEBYTECODE=
-	# workaround https://bugs.gentoo.org/775416
-	addwrite /usr/lib/python3.10/site-packages
+	addpredict /usr/lib/python3.10/site-packages
 
 	local jobs=$(makeopts_jobs "${MAKEOPTS}" "$(get_nproc)")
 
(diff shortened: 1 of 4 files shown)


dev-lang/python/python-3.10.0_p1.ebuild
```diff
@@ -178,6 +183,9 @@ src_configure() {
 		--without-ensurepip
 		--with-system-expat
 		--with-system-ffi
+
+		$(use_with lto)
```
mgorny, line 187: Wrong commit.

dev-lang/python/python-3.11.0_alpha2.ebuild
```diff
@@ -198,6 +206,18 @@ src_compile() {
 	# https://bugs.gentoo.org/823728
 	export SETUPTOOLS_USE_DISTUTILS=stdlib
 
+	if use pgo ; then
+		# bug 660358
+		local -x COLUMNS=80
+		local -x PYTHONDONTWRITEBYTECODE=
+
+		local jobs=$(makeopts_jobs "${MAKEOPTS}" "$(get_nproc)")
+		export PROFILE_TASK="-m test -j${jobs} --pgo-extended -x test_gdb"
```
mgorny, line 215: You seem to be doing this twice in this ebuild.

mgorny (changes requested):
Could you please limit the changes to just one ebuild for the purpose of review? I'm getting confused which version is the correct one.

Also, how about changing to `addpredict` first?

dev-lang/python/python-3.10.0_p1.ebuild
```diff
@@ -225,7 +225,7 @@ src_test() {
 	local -x COLUMNS=80
 	local -x PYTHONDONTWRITEBYTECODE=
 	# workaround https://bugs.gentoo.org/775416
```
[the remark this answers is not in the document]
thesamesam: You're right on the comment, but sadly, we really do need addpredict still.

mgorny (approved):
LGTM. I don't care about pgo/lto much but please backport the `addpredict` change to all versions.

PR #22857, opened 2021-11-07
app-emulation/firecracker-bin: Bump to 0.25.1, install seccompiler-bin

s-hamann:
Version bump `app-emulation/firecracker-bin` to 0.25.1.
This ebuild installs the optional `seccompiler-bin`, which was newly added in Firecracker 0.25.0.

I also removed a check on kernel VSOCK support for all versions in the tree. Firecracker does not actually use `AF_VSOCK` on the hypervisor.


app-emulation/firecracker-bin/firecracker-bin-0.25.1.ebuild
```diff
@@ -0,0 +1,80 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~arm64"
+
+RESTRICT="test strip"
+
+RDEPEND="!app-emulation/firecracker
+	acct-group/kvm"
+
+QA_PREBUILT="/usr/bin/firecracker
+	/usr/bin/jailer"
```
ionenwks, line 28: It's trivial so I'll add it on merge -- missing the new `usr/bin/seccompiler-bin`:
```
 * /usr/bin/seccompiler-bin
```
On a side-note, it's not /essential/ but QA_* variables should use relative paths (aka `="usr/..."` rather than `="/usr/..."` as (briefly and easy to miss) mentioned in the docs:
https://devmanual.gentoo.org/eclass-reference/ebuild/index.html#lbAG

Not that it's necessary to change them all, but since touching this I'll adjust for the latest ebuild.

app-emulation/firecracker-bin/firecracker-bin-0.25.1.ebuild
```diff
@@ -0,0 +1,80 @@
...
+pkg_pretend() {
+	if use kernel_linux && kernel_is lt 4 14; then
+		eerror "Firecracker requires a host kernel of 4.14 or higher."
+	elif use kernel_linux; then
+		if ! linux_config_exists; then
+			eerror "Unable to check your kernel for KVM support"
+		else
+			CONFIG_CHECK="~KVM ~TUN ~BRIDGE"
+			CONFIG_CHECK="~KVM ~TUN ~BRIDGE"
+			ERROR_KVM_AMD="If you have an AMD CPU, you must enable KVM_AMD in"
```
ionenwks, line 41: Believe accidentally deleted `ERROR_KVM=` instead of the duplicate here, I'll restore it to match other ebuilds.

ionenwks (approved):
Thanks.

app-emulation/firecracker-bin/firecracker-bin-0.25.1.ebuild
s-hamann: Thanks, I missed that.

app-emulation/firecracker-bin/firecracker-bin-0.25.1.ebuild
s-hamann: Thank you for fixing this.

PR #22869, opened 2021-11-08
media-sound/mpd: bump to 0.23.3

maxaudron:
Bump mpd to version 0.23.3 and a pipewire use flag


maxaudron:
I've added libnpupnp as RDEPEND behind use statements for "amd64?" and "x86?". 
From what I know this isn't really _prefered_ but I feel like it's the cleaner solution than adding the flag to all the archs "use.mask" files.
Else I'd remove the npupnp use flag completely.

mrksmrtn:
This upgraded ebuild builds nicely for me (with Pipewire) after installing one dep. I didn't test the npupnp USE flag. Sorry, I'm new-ish to Gentoo (switched back after over a decade hiatus). I needed the latest mpd to solve some Pipewire issues.

However, I did have to install `dev-libs/libfmt`. This is a new build dependency since mpd v0.23 (see [the NEWS file](https://github.com/MusicPlayerDaemon/MPD/blob/master/NEWS#L79) and [this commit](https://github.com/MusicPlayerDaemon/MPD/commit/a9c704b76e05571043ece260f517bfe7dc759195)). The build fails otherwise (full log available if you are really curious).

```
Run-time dependency fmt found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency fmt

meson.build:249:0: ERROR: Automatic wrap-based subproject downloading is disabled
```

maxaudron:
Thanks for catching that @mrksmrtn
I added the dep to the ebuild

thesamesam:
0.23.5 is now in tree, thank you for your PR

PR #22872, opened 2021-11-09
dev-util/perf-5.12: allow python 3.10

yellowhat:
Hi,
the current dev-util/perf ebuild allows python 3.7 to 3.9.
May I ask to allow 3.10?
Upstream I was not able to found any specific python version dependency/test.

Thanks


yellowhat:
Fixed upstream

PR #22878, opened 2021-11-10
sci-libs/vtk: bump to 9.1.0

waebbl:
Bug: https://bugs.gentoo.org/820593
Closes: https://bugs.gentoo.org/822780
Closes: https://bugs.gentoo.org/820521


waebbl:
I'd like to simplify the USE flags, remove some of them and enable those by default. This include boost, ffmpeg, gdal, theora. Also considering of summarizing the mysql, odbc and postgres USE flags into a single sql USE flag.

WDYT?

waebbl:
Is there a way for me to extensively test this package wrt to USE flag settings on one of our machines?

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+}
+
+src_prepare() {
+	# If we have system libraries available use these and delete
+	# the respecting files in ${S}/ThirdParty to save some space.
+	# Note: The following libraries are marked as internal by kitware
+	#	and can currently not unbundled:
+	#	diy2, exodusII, fides, h5part, kissfft, loguru, verdict, vpic,
+	#	vtkm, xdmf{2,3}, zfp
+	# Note: libharu is omitted: vtk needs an updated version (2.4.0)
```
waebbl, line 182: fmt is omitted too, needs >=libfmt-8.1.0

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+# TODO: check these and consider to use them
+#	VTK_LEGACY_REMOVE
+#	VTK_LEGACY_SILENT
+#	VTK_USE_MEMKIND
+#	VTK_BUILD_SCALED_SOA_ARRAYS
+#	VTK_DISPATCH_{AOS,SOA,TYPED}_ARRAYS
+
+	local mycmakeargs=(
+		-DVTK_ANDROID_BUILD=OFF
+		-DVTK_IOS_BUILD=OFF
```
waebbl, line 233: There's still a VTK_USE_COCOA option. IMO it's tied to USE=rendering

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+			-DVTK_MODULE_ENABLE_VTK_IOADIOS2="DONT_WANT"
+			-DVTK_MODULE_ENABLE_VTK_IOPDAL="DONT_WANT"
+			-DVTK_MODULE_ENABLE_VTK_MomentInvariants="DONT_WANT"
+			-DVTK_MODULE_ENABLE_VTK_PoissonReconstruction="DONT_WANT"
+			-DVTK_MODULE_ENABLE_VTK_Powercrust="DONT_WANT"
+			-DVTK_MODULE_ENABLE_VTK_RenderingOpenVR="DONT_WANT"
+			-DVTK_MODULE_ENABLE_VTK_SignedTensor="DONT_WANT"
+			-DVTK_MODULE_ENABLE_VTK_SplineDrivenImageSlicer="DONT_WANT"
+			-DVTK_MODULE_ENABLE_VTK_vtkDICOM="DONT_WANT"
+			-DVTK_MODULE_USE_EXTERNAL_vtkkissfft=ON
```
waebbl, line 308: don't needed. Kissfft is always internal as of now

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+
+REQUIRED_USE="
+	all-modules? ( boost ffmpeg gdal imaging mysql odbc postgres qt5 rendering theora views )
+	cuda? ( video_cards_nvidia )
+	java? ( rendering )
+	python? ( ${PYTHON_REQUIRED_USE} )
+	qt5? ( rendering )
+	tk? ( rendering python )
+	web? ( python )
+"
```
waebbl, line 57: add xor X offscreen here

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+		)
+	fi
+
+	# TODO: remote the USE flag and enable by default
+	if use theora; then
+		mycmakeargs+=( -DVTK_MODULE_ENABLE_VTK_IOOggTheora="WANT" )
+	fi
+
+	# FIXME: upstream provides 4 threading models, as of 9.1.0. These are
+	# sequential, stdpthread, openmp and tbb. AFAICS all of them can be
```
waebbl, line 495: s/stdpthread/stdthread

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+
+	if use openmp; then
+		mycmakeargs+=( -DVTK_SMP_ENABLE_OPENMP=ON )
+	fi
+
+	if use tbb; then
+		mycmakeargs+=( -DVTK_SMP_ENABLE_TBB=ON )
+	fi
+
+	# ON by default, hence the else clause is neede
```
waebbl, line 518: s/neede/needed/

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+		export FC=mpif90
+		export F90=mpif90
+		export F77=mpif77
+	fi
+
+	cmake_src_configure
+}
+
+src_test() {
+	nonfatal virtx cmake_src_test
```
waebbl, line 539: why does pkgcheck say inherit virtualx isn''t needed if we use virtx here?

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+		mv -v {E,e}xamples || die
+		dodoc -r examples
+		docompress -x /usr/share/doc/${PF}/examples
+
+		einfo "Installing datafiles"
+		insinto /usr/share/${PN}/data
+		doins -r "${S}/.ExternalData"
+	fi
+
+	# with MPI runpath's are not deleted properly
```
waebbl, line 566: still valid in 9.1.0?

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+	if use mpi; then
+		chrpath -d "${ED}"/usr/$(get_libdir)/*.so.${PV} || die
+	fi
+
+	use python && python_optimize
+
+	# environment
```
waebbl, line 573: do we really need this? With default USE flag settings, there's no `/usr/share/vtk/data` or `/usr/$(get_libdir)/vtk`

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+		mycmakeargs+=( -DVTK_SMP_ENABLE_OPENMP=ON )
+	fi
+
+	if use tbb; then
+		mycmakeargs+=( -DVTK_SMP_ENABLE_TBB=ON )
+	fi
+
+	# ON by default, hence the else clause is neede
+	if use threads; then
+		mycmakeargs+=( -DVTK_SMP_ENABLE_STDTHREAD=ON )
```
thesamesam, line 520: just =$(usex threads)? no need for the if statement?

sci-libs/vtk/vtk-9.1.0.ebuild
thesamesam, line 539: Why nonfatal? Won't this prevent us from knowing if it fails?

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+	else
+		mycmakeargs+=( -DVTK_SMP_IMPLEMENTATION_TYPE="Sequential" )
+	fi
+
+	if use openmp; then
+		mycmakeargs+=( -DVTK_SMP_ENABLE_OPENMP=ON )
+	fi
+
+	if use tbb; then
+		mycmakeargs+=( -DVTK_SMP_ENABLE_TBB=ON )
```
thesamesam, line 515: This might be another one we can just do =$(usex tbb) for. probably others too

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+)
+
+DOCS=( CONTRIBUTING.md README.md )
+
+CHECKREQS_DISK_BUILD="3G"
+# we want the EAPI 8 default
+CMAKE_BUILD_TYPE=RelWithDebInfo
+
+pkg_pretend() {
+	[[ MERGE_TYPE != binary ]] && has openmp && tc-check-openmp
```
thesamesam, line 143: ${}

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+		# as any of the cuda compile tasks can take up 7G!
+		# 10.2 GiB install directory, 6.4 GiB build directory with max. USE flags
+		CHECKREQS_MEMORY="7G"
+		CHECKREQS_DISK_BUILD="14G"
+	fi
+	check-reqs_pkg_setup
+}
+
+pkg_setup() {
+	[[ MERGE_TYPE != binary ]] && has openmp && tc-check-openmp
```
thesamesam, line 159: ${}

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+
+	if use examples; then
+		CHECKREQS_DISK_BUILD="4G"
+	fi
+	if use cuda; then
```
thesamesam, line 164: nit: new line between ifs

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+		# NOTE: This should actually equal to (number of build jobs)*7G,
+		# as any of the cuda compile tasks can take up 7G!
+		# 10.2 GiB install directory, 6.4 GiB build directory with max. USE flags
+		CHECKREQS_MEMORY="7G"
+		CHECKREQS_DISK_BUILD="14G"
+	fi
+	check-reqs_pkg_setup
```
thesamesam, line 155: nit: new line before this

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+# we want the EAPI 8 default
+CMAKE_BUILD_TYPE=RelWithDebInfo
+
+pkg_pretend() {
+	[[ MERGE_TYPE != binary ]] && has openmp && tc-check-openmp
+
+	if use examples; then
+		CHECKREQS_DISK_BUILD="4G"
+	fi
+	if use cuda; then
```
thesamesam, line 148: new: new line between ifs

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+		docompress -x /usr/share/doc/${PF}/examples
+
+		einfo "Installing datafiles"
+		insinto /usr/share/${PN}/data
+		doins -r "${S}/.ExternalData"
+	fi
+
+	# with MPI runpath's are not deleted properly
+	if use mpi; then
+		chrpath -d "${ED}"/usr/$(get_libdir)/*.so.${PV} || die
```
thesamesam, line 568: If we want to be _really_ portable, I think we need to use $(get_libname) or w/e from multilib.eclass instead of hardcoding ".so". But I'm mostly commenting this to tell you the problem/idea exists. I feel like it doesn't really matter for VTK.

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,601 @@
...
+	if use boost; then
+		mycmakeargs+=(
+			-DVTK_MODULE_ENABLE_VTK_InfovisBoost="WANT"
+			-DVTK_MODULE_ENABLE_VTK_InfovisBoostGraphAlgorithms="WANT"
+		)
+	fi
+
+	if use cuda; then
+		local cuda_arch=
+		case ${VTK_CUDA_ARCH:-native} in
```
thesamesam, line 322: I guess we may want to add a comment here explaining how to update this in future

sci-libs/vtk/vtk-9.1.0.ebuild
waebbl: Thought about this at one time as well. Should do it.

ionenwks: If wanted to, can even use it for multiple flags and potentially use a single'ish array, e.g.
```
array=(
	-DSTUFF
	$(usex flag '
		-DARG
		-DARG2' '')
	$(usex flag2 '
		-DARG3
		-DARG4' '')
)
```
But whether this is better is debatable (either way, note to replace by `$(usev flag -DFLAG)` whenever this goes EAPI-8 given that extra `''` is always ugly).

sci-libs/vtk/vtk-9.1.0.ebuild
waebbl: IIRC I added this in 9.0.1 in case virtx fails for some reason to avoid a false positive. According to the PMS nonfatal should return a non-zero exit status, in case the called function fails. So we could test for the return value to check if the tests failed.

sci-libs/vtk/vtk-9.1.0.ebuild
waebbl: Right, should work here too. Also for openmp.

sci-libs/vtk/vtk-9.1.0.ebuild
waebbl: The use examples if-clause will probably be removed anyway. With default USE flags, the package already needs more than 3G, so increased the default value to 4G in my current working copy.

sci-libs/vtk/vtk-9.1.0.ebuild
waebbl: You mean, like when NVidia adds new architectures and/or obsoletes old arches? Good idea!

waebbl:
Many thanks for your early review and feedback @thesamesam

sci-libs/vtk/vtk-9.1.0.ebuild
waebbl: Thanks for your feedback @ionenwks. There's only really a few which can be eliminated without causing cmake warnings about used but undefined variables. Most if statements are there, because the variables will only be defined once a certain scope is enabled.

ionenwks: That form of `$(usex flag 'add this or nothing' '')` or eapi8 `$(usev flag this-or-nothing)` doesn't set the variable at all if the flag isn't, it's essentially an embed if statement

sci-libs/vtk/vtk-9.1.0.ebuild
waebbl: X removed, so not needed actually

sci-libs/vtk/vtk-9.1.0.ebuild
waebbl: I mostly added a comment about this to test whether chrpath still needs to be pulled in. Actually trying to fix this, is atm probably beyond my knowledge / skills.

thesamesam:
> Is there a way for me to extensively test this package wrt to USE flag settings on one of our machines?

Sorry, thought I'd replied to this. Easiest way is to add it masked or unkeyworded and ask ago or toralf.

waebbl:
> > Is there a way for me to extensively test this package wrt to USE flag settings on one of our machines?
> 
> Sorry, thought I'd replied to this. Easiest way is to add it masked or unkeyworded and ask ago or toralf.

Ok thanks, sound like a good idea. Will do that.

waebbl:
> > Is there a way for me to extensively test this package wrt to USE flag settings on one of our machines?
> 
> Sorry, thought I'd replied to this. Easiest way is to add it masked or unkeyworded and ask ago or toralf.

Ok thanks, sound like a good idea. Will do that.

waebbl:
Some things I noticed during my last checks:

- Sometimes when jave use flag is set, this leads to ninja not finding build targets and the build breaks. This happened when only java use flag is set, but not when java was added together with other use flags. Need to test whether java use flag needs another use flag enabled in REQUIRED_USE
- Core vtk libraries build successfully with latest tbb. If the vtkm third party module is getting build, the build fails with latest tbb
- `USE=cuda` fails with gcc-11 and I also had failures in some cases with gcc-10. Probably best to mask cuda use flag for now, until we have figured out what's going on here
- Failing **test** rate is dependant on python version used. With an identical use flag setup, but differing python versions, I had ~25% failure rate with python3_{8,9} and ~56% failure rate with python3_10. Only tested with my usual setup, need to check this with only python use flag enabled.

waebbl:
> I'd like to simplify the USE flags, remove some of them and enable those by default. This include boost, ffmpeg, gdal, theora. Also considering of summarizing the mysql, odbc and postgres USE flags into a single sql USE flag.
> 
> WDYT?

Ping

I already removed theora USE flag, because theora is pulled in anyway and the use flag only enables an IO module. I'm gonna review the options passed for these USE flags (see comments in current ebuild) and probably enable those by default for now and the mask and remove them in a later revbump. IIRC most _only_ enable IO modules.

thesamesam:
> > I'd like to simplify the USE flags, remove some of them and enable those by default. This include boost, ffmpeg, gdal, theora. 

Up to you, I suppose. Depends on how likely consumers are to need these. Avoiding Boost is still useful for some people where possible but if the software is useless without it, no point bothering. Or if most consumers need it.

I'm afraid I don't thnk I know enough about VTK to know how useful it is to turn this stuff off.

>Also considering of summarizing the mysql, odbc and postgres USE flags into a single sql USE flag.
> > WDYT?

I'd leave mysql/postgres/odbc separate as this is how most packages work.

> 
> Ping
> 
> I already removed theora USE flag, because theora is pulled in anyway and the use flag only enables an IO module. I'm gonna review the options passed for these USE flags (see comments in current ebuild) and probably enable those by default for now and the mask and remove them in a later revbump. IIRC most _only_ enable IO modules.

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,661 @@
...
+	rendering? (
+		media-libs/freeglut
+		media-libs/glew:=
+		media-libs/libsdl2
+		sci-libs/proj:=
+		virtual/opengl
+		x11-libs/gl2ps
+		x11-libs/libXcursor
+	)
+	tbb? ( <dev-cpp/tbb-2021.4.0:= )
```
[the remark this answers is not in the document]
waebbl: > Actually, vtk-9.1.0 seems to be already fixed to support new tbb-2021.4.0 (well, at least, as reported by the user provided my patches for it and 9.0.3).

>     * Core vtk libraries build successfully with latest tbb. If the vtkm third party module is getting build, the build fails with latest tbb

I had it updated at first, but later needed to re-add the restriction due to failures in the vtkm module.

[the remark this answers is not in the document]
msva: Well, I have patches (that I take from upstream today, to make 9.0.3 work) fto fix that.

Can you please navigate me, if there is any other non-resolved issue here?
If there is no - wouldn't you mind if I'll commit your ebuild (with my patches and some additions) and add you as the author?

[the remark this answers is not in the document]
thesamesam: I'd much rather we wait until waebbl is happy and then commit it unkeyworded or masked for toralf or ago to do reverse dependency testing.

You can then make a PR with the changes you want to make in a new revision on top of that -- I'd prefer to review it separately.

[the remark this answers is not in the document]
msva: Actually, here is the whole difference:

```
--- vtk-9.1.0-r0.ebuild	2021-12-04 05:40:38.837523721 +0700
+++ vtk-9.1.0.ebuild	2021-12-04 05:41:27.377116532 +0700
@@ -107,7 +107,7 @@ RDEPEND="
 		x11-libs/gl2ps
 		x11-libs/libXcursor
 	)
-	tbb? ( <dev-cpp/tbb-2021.4.0:= )
+	tbb? ( dev-cpp/tbb:= )
 	tk? ( dev-lang/tk:= )
 	video_cards_nvidia? ( x11-drivers/nvidia-drivers[tools,static-libs] )
 	views? (
@@ -129,7 +129,7 @@ RDEPEND="
 DEPEND="
 	${RDEPEND}
 	dev-cpp/eigen
-	<dev-libs/pegtl-3
+	dev-libs/pegtl
 	dev-libs/utfcpp
 "
 BDEPEND="
@@ -145,6 +145,9 @@ BDEPEND="
 PATCHES=(
 	"${FILESDIR}"/${PN}-9.0.1-0001-fix-kepler-compute_arch-if-CUDA-toolkit-11-is-used.patch
 	"${FILESDIR}"/${PN}-8.2.0-freetype-2.10.3-provide-FT_CALLBACK_DEF.patch
+	"${FILESDIR}"/${PN}-9.0.3-cuda-11.5.0.patch
+	"${FILESDIR}"/${P}-pegtl-3.patch
+	"${FILESDIR}"/${P}-tbb-fix-for-bundled-vtkm.patch
 )
 
 DOCS=( CONTRIBUTING.md README.md )
```

And here is patches:
files/vtk-9.0.3-cuda-11.5.0.patch: https://gist.github.com/raw/be7ac35e2ff77c00896e9fb7885720c6
files/vtk-9.1.0-pegtl-3.patch: https://gist.github.com/raw/462381f875cca6fc96e3dd18aed18d93
files/vtk-9.1.0-tbb-fix-for-bundled-vtkm.patch: https://gist.github.com/raw/9f9cd7de552f5dcf779cc65416171658


And I will be okay even if you'll review them now and then @waebbl will take them to his PR and then agoand toralf will test it at whole (with support for pegtl-3 and tbb-2021).


Although, I still not finished building the 9.1.0, so maybe I'll add one or few patches a bit later.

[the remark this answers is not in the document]
msva: Also, maybe it would be wiser to specify dep atoms for `pegtl` and `tbb` as `<dev-libs/pegtl-4` and `<dev-cpp/tbb-2022:=` :man_shrugging:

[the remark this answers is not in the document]
thesamesam: Please include links to where each of these patches are from. Several of them appear quite invasive.

Which are upstream? Which aren't upstream? Are there open bugs for these?

I'd prefer to not patch <9.1 at all.

As mentioned on IRC, several of these are also too big to go in the repository, especially combined. Will need to put the ones we want to take into a tarball (but we can handle that at the point of merging, it's easier to keep them in files/ for the purposes of review).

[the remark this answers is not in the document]
msva: > Please include links to where each of these patches are from

files/vtk-9.0.3-cuda-11.5.0.patch: As mentioned at the top of the file, the source is user who provided it.
I tested that it really builds with mentioned CUDA toolkit version.

files/vtk-9.1.0-pegtl-3.patch: Also, patch was made by @perestoronin by taking https://github.com/USBGuard/usbguard/pull/453 as hints.

files/vtk-9.1.0-tbb-fix-for-bundled-vtkm.patch:
this commits from vtk-m upstream repo:
https://gitlab.kitware.com/vtk/vtk-m/-/commit/97a2408fc
https://gitlab.kitware.com/vtk/vtk-m/-/commit/1eea0bee1
https://gitlab.kitware.com/vtk/vtk-m/-/commit/46b026882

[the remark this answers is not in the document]
thesamesam: The CUDA patch and the pegtl one should both be submitted upstream or at least with a link to an upstream bug report. Gentoo is not maintaining a fork of VTK, so patches should go where they belong first.

I don't see the value in the TBB one right now given _very_ little works with newer TBB. Can you explain why it's necessary? And I'd still want waebbl's input there.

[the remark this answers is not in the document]
msva: > Can you explain why it's necessary

Not that it was "necessary" per se.

Just some hobby projects I'm using `vtk` in, already decided to move to `tbb-2021` and `pegtl-3` in the projects themselves (as both of them are already about a year old and have useful features).

And, as far as I get @perestoronin correctly, he is in the similar situation.

So, my intention was to contribute a bit of spare time after working day (well, actually, it took few days to test some combinations I thought of, and it anyway failed for Zorry 😿) to make it to work with  versions of deps that we're already using anyway (to avoid bundling on project side).


Anyway, I've got your point, and will try to make things right way when I'd be able to.

juippis:
> Is there a way for me to extensively test this package wrt to USE flag settings on one of our machines?

*pkg-testing-tools* and *tatt* can both do that.
https://wiki.gentoo.org/wiki/Package_testing#Tools

waebbl:
> > Is there a way for me to extensively test this package wrt to USE flag settings on one of our machines?
> 
> _pkg-testing-tools_ and _tatt_ can both do that. https://wiki.gentoo.org/wiki/Package_testing#Tools

Thanks @juippis, I know these tools :) My question was aimed at if it's possible to use one of the machines Gentoo posesses to do these jobs. I spent most of the last two weeks with testing different use flag combinations. But given the number of use flags for the package and my old machine, hence long build times, this can only cover a small amount of the possible combinations.

sci-libs/vtk/vtk-9.1.0.ebuild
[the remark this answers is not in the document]
waebbl: The tbb patches haven't found their way into the 9.1.0 package, but are in the current master, and thus will get into the next release of vtk. Given that, and what @thesamesam said, only very few packages actually support tbb-2021, I think there's no need to rush with these patches. But I'm not against adding them. If we decide to add them, I think, it's more advisable to split the patch to make it more manageable.

Haven't yet tested the pegtl-3 patch. It looks good to me, but I'd prefer if we open an issue upstream and provide a patch for it, before we add it to the tree. People who urgently need this patch for their own projects could be pointed to use `/etc/portage/patches` for this purpose.

Does the cuda patch also work for <cuda-11.5? I can't test with cuda-11.5, as it removed support for my sm_35 based GPU arch. Does it work with both gcc-10 and gcc-11? Given the lot of issues we lately had with cuda in vtk, I'm thinking about masking the cuda use flag for the time being, until we have figured out how to solve it in a stable manner. Building with cuda support has some meaningful hardware requirements and build times, so I'd rather be more conservative when it comes to patching these. Additionally this is another patch, where we should go upstream first and report it and try to help them figuring and fixing it.

juippis:
> > > Is there a way for me to extensively test this package wrt to USE flag settings on one of our machines?
> > 
> > 
> > _pkg-testing-tools_ and _tatt_ can both do that. https://wiki.gentoo.org/wiki/Package_testing#Tools
> 
> Thanks @juippis, I know these tools :) My question was aimed at if it's possible to use one of the machines Gentoo posesses to do these jobs. I spent most of the last two weeks with testing different use flag combinations. But given the number of use flags for the package and my old machine, hence long build times, this can only cover a small amount of the possible combinations.

I was gonna try and help with that, but 
```
>>> Emerging (84 of 84) sci-libs/vtk-9.1.0::gentoo
>>> Jobs: 83 of 84 complete, 1 running              Load avg: 76.7, 74.1, 51.1^C
```
I need my machine atm :P will try again later. If https://qa-reports.gentoo.org/output/genrdeps/dindex/sci-libs/vtk are the only rdeps it's pretty easy for me to test them too.

juippis:
```
Errors while running CTest
Output from these tests are in: /var/tmp/portage/sci-libs/vtk-9.1.0/work/vtk-9.1.0_build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
 * ERROR: sci-libs/vtk-9.1.0::gentoo failed (test phase):
```
https://dev.gentoo.org/~juippis/logs/vtk-9.1.0-tests.log

I've rand ~10 different USE flag combinations, all seem fine. Test phase fails. I'm gonna step out and do some rdeptesting most likely later tonight (overnight)

waebbl:
Thanks for test @juippis 
Yeah sadly tests are failing alot, I already noticed this. 

> Some things I noticed during my last checks:
> 
> Failing **test** rate is dependant on python version used. With an identical use flag setup, but differing python versions, I had ~25% failure rate with python3_{8,9} and ~56% failure rate with python3_10. Only tested with my usual setup, need to check this with only python use flag enabled.

waebbl:
I rebase and add a mask to package.mask after work today.

waebbl:
With the cuda-11.5 patch applied, it still fails with gcc-11 and works with gcc-10 on my test

waebbl:
@toralf @asarubbo would you mind giving the package an exhaustive USE flag testing once it's merged to test whether I missed anything?

toralf:
> @toralf @asarubbo would you mind giving the package an exhaustive USE flag testing once it's merged to test whether I missed anything?

The tinderbox automatically picks it up once it is in ::gentoo and then it is regularly scheduled forever.

waebbl:
> > @toralf @asarubbo would you mind giving the package an exhaustive USE flag testing once it's merged to test whether I missed anything?
> 
> The tinderbox automatically picks it up once it is in ::gentoo and then it is regularly scheduled forever.

@toralf It will get merged masked at first, so it can be tested.

juippis:
On it, will focus on this for today.

sci-libs/vtk/vtk-9.1.0.ebuild
```diff
@@ -0,0 +1,596 @@
...
+	dev-libs/icu:=
+	dev-libs/jsoncpp:=
+	dev-libs/libxml2:2
+	dev-libs/pugixml
+	media-libs/freetype
+	media-libs/libogg
+	media-libs/libpng:=
+	media-libs/libtheora
+	media-libs/tiff
+	<sci-libs/hdf5-1.12:=[mpi=]
```
amadio, line 74: I'm trying to install vtk-9.0.3 and I see that hdf5 needs to be installed with USE=hl for Vtk:
```
-- HDF5: Using hdf5 compiler wrapper to determine C configuration
-- Could NOT find HDF5 (missing: HDF5_HL_LIBRARIES) (found version "1.10.5")
CMake Error at CMake/vtkModule.cmake:4145 (message):
  Could not find the HDF5 external dependency.
Call Stack (most recent call first):
  CMake/vtkModule.cmake:4695 (vtk_module_find_package)
  CMake/vtkModule.cmake:4569 (vtk_module_third_party_external)
  ThirdParty/hdf5/CMakeLists.txt:8 (vtk_module_third_party)
```
It would also be nice if the restriction on the version could be lifted as well.

amadio: BTW, I can help with testing CUDA support, as I have a recent Nvidia card.

waebbl: Thanks for the hint. That's correct, see https://gitlab.kitware.com/vtk/vtk/-/blob/v9.0.3/ThirdParty/hdf5/CMakeLists.txt#L17. I've missed this so far, probably because it's set by default.

As for hdf5-1.12.5, this should be supported by 9.1.0, but I couldn't get it to work. Any insights into this, are welcome. I wouldn't bother with it for now, as there were more issue concerning hdf5-1.12 when I started working on 9.1.0. It will not work, however, for 9.0.3.

juippis (approved):
Here's my exhaustive test report. USE flag combination test was done with 20 rounds, randomized USE flags:

[INFO] >>> Following testing jobs will be executed:
=sci-libs/vtk-9.1.0    USE: -all-modules boost -cuda -doc examples ffmpeg gdal imaging java java mpi mysql -odbc -openmp -postgres -python -python_single_target_python3_10 -python_single_target_python3_8 -python_single_target_python3_9 qt5 rendering tbb -threads -tk -views -web
=sci-libs/vtk-9.1.0    USE: -all-modules boost -cuda doc -examples ffmpeg gdal imaging -java -java mpi -mysql odbc openmp postgres -python -python_single_target_python3_10 python_single_target_python3_8 python_single_target_python3_9 qt5 rendering tbb threads -tk views -web
=sci-libs/vtk-9.1.0    USE: -all-modules -boost -cuda doc examples ffmpeg gdal -imaging -java -java mpi -mysql -odbc openmp postgres python python_single_target_python3_10 -python_single_target_python3_8 -python_single_target_python3_9 qt5 rendering tbb -threads tk views web
=sci-libs/vtk-9.1.0    USE: -all-modules boost -cuda -doc examples -ffmpeg gdal -imaging -java -java -mpi mysql -odbc -openmp postgres -python python_single_target_python3_10 python_single_target_python3_8 -python_single_target_python3_9 -qt5 -rendering -tbb threads -tk -views -web
=sci-libs/vtk-9.1.0    USE: -all-modules -boost -cuda -doc -examples -ffmpeg -gdal -imaging -java -java mpi -mysql odbc openmp postgres python python_single_target_python3_10 -python_single_target_python3_8 -python_single_target_python3_9 -qt5 rendering tbb -threads tk -views -web
=sci-libs/vtk-9.1.0    USE: -all-modules -boost -cuda doc examples ffmpeg gdal imaging -java -java mpi mysql odbc openmp postgres -python python_single_target_python3_10 -python_single_target_python3_8 -python_single_target_python3_9 qt5 rendering tbb threads -tk -views -web
=sci-libs/vtk-9.1.0    USE: -all-modules -boost -cuda -doc examples -ffmpeg -gdal imaging -java java mpi mysql odbc openmp -postgres python -python_single_target_python3_10 python_single_target_python3_8 -python_single_target_python3_9 qt5 rendering -tbb threads tk -views web
=sci-libs/vtk-9.1.0    USE: -all-modules boost -cuda -doc -examples ffmpeg -gdal imaging -java -java -mpi mysql -odbc -openmp postgres python -python_single_target_python3_10 python_single_target_python3_8 -python_single_target_python3_9 -qt5 -rendering -tbb -threads -tk -views -web
=sci-libs/vtk-9.1.0    USE: -all-modules -boost -cuda doc examples -ffmpeg -gdal -imaging java -java mpi mysql -odbc openmp -postgres python python_single_target_python3_10 -python_single_target_python3_8 -python_single_target_python3_9 qt5 rendering -tbb threads -tk views -web
=sci-libs/vtk-9.1.0    USE: -all-modules -boost -cuda -doc -examples -ffmpeg -gdal imaging -java -java mpi mysql -odbc -openmp postgres -python -python_single_target_python3_10 -python_single_target_python3_8 -python_single_target_python3_9 -qt5 -rendering -tbb -threads -tk views -web
=sci-libs/vtk-9.1.0    USE: -all-modules -boost -cuda -doc examples -ffmpeg gdal -imaging -java -java mpi -mysql odbc openmp postgres -python python_single_target_python3_10 python_single_target_python3_8 python_single_target_python3_9 qt5 rendering tbb -threads -tk views -web
=sci-libs/vtk-9.1.0    USE: -all-modules boost -cuda doc examples ffmpeg -gdal -imaging java java -mpi mysql odbc openmp -postgres -python -python_single_target_python3_10 python_single_target_python3_8 python_single_target_python3_9 -qt5 rendering -tbb -threads -tk views -web
=sci-libs/vtk-9.1.0    USE: -all-modules boost -cuda doc -examples ffmpeg gdal -imaging -java -java mpi mysql -odbc -openmp -postgres python -python_single_target_python3_10 python_single_target_python3_8 -python_single_target_python3_9 qt5 rendering tbb threads tk -views -web
=sci-libs/vtk-9.1.0    USE: -all-modules -boost -cuda -doc -examples ffmpeg -gdal imaging -java java mpi mysql odbc -openmp -postgres -python -python_single_target_python3_10 -python_single_target_python3_8 -python_single_target_python3_9 qt5 rendering tbb threads -tk views -web
=sci-libs/vtk-9.1.0    USE: -all-modules -boost -cuda -doc -examples -ffmpeg -gdal -imaging -java java mpi mysql -odbc openmp -postgres python python_single_target_python3_10 -python_single_target_python3_8 -python_single_target_python3_9 qt5 rendering tbb threads tk -views -web
=sci-libs/vtk-9.1.0    USE: -all-modules boost -cuda doc -examples ffmpeg gdal -imaging -java java mpi mysql odbc -openmp -postgres python -python_single_target_python3_10 -python_single_target_python3_8 python_single_target_python3_9 qt5 rendering -tbb threads -tk -views web
=sci-libs/vtk-9.1.0    USE: -all-modules boost -cuda doc examples ffmpeg gdal -imaging -java -java -mpi mysql -odbc openmp postgres python -python_single_target_python3_10 python_single_target_python3_8 -python_single_target_python3_9 qt5 rendering tbb -threads tk -views -web
=sci-libs/vtk-9.1.0    USE: -all-modules boost -cuda doc -examples ffmpeg -gdal -imaging -java -java mpi mysql odbc openmp postgres -python -python_single_target_python3_10 -python_single_target_python3_8 python_single_target_python3_9 -qt5 rendering tbb threads -tk views -web
=sci-libs/vtk-9.1.0    USE: -all-modules boost -cuda -doc -examples ffmpeg gdal -imaging -java -java mpi -mysql -odbc -openmp -postgres -python -python_single_target_python3_10 python_single_target_python3_8 python_single_target_python3_9 -qt5 rendering -tbb -threads -tk -views -web
=sci-libs/vtk-9.1.0    USE: -all-modules -boost -cuda doc examples ffmpeg -gdal -imaging -java java mpi -mysql -odbc -openmp postgres python -python_single_target_python3_10 python_single_target_python3_8 -python_single_target_python3_9 qt5 rendering -tbb -threads -tk views web
=sci-libs/vtk-9.1.0    USE: <default flags>, FEATURES: test

Only one failed, the last with default flags and `FEATURES="test"`. 

Then onto rdeps test, based on: 
https://qa-reports.gentoo.org/output/genrdeps/rindex/sci-libs/vtk
(originally I just checked dindex, based on DEPEND, but comments about that below...)

**successfull** rdeps test:

----------
=sci-mathematics/maxima-5.44.0-r9 +X +vtk

success, although: 

```
my-test-container-snap-tmp ~ # grep -r libvtk /var/db/pkg/sci-mathematics/maxima-5.44.0-r9/
my-test-container-snap-tmp ~ # emerge -pv maxima

Calculating dependencies  ... done!
[ebuild     U  ] sci-mathematics/maxima-5.45.1-r2::gentoo [5.44.0-r9::gentoo] USE="sbcl unicode* -X* -clisp* (-clozurecl) -clozurecl64% (-cmucl) -ecls -emacs -gcl -gui -nls -test -vtk*" L10N="-de -es -pt -pt-BR" PYTHON_SINGLE_TARGET="python3_9%* -python3_8*" 0 KiB
```

which made me realize for some reason other versions of maxima don't DEPEND on vtk, just RDEPEND. 

----------

=sci-mathematics/maxima-5.45.1-r2 +vtk

success, but

```
 * VDB: detected possibly incorrect RDEPEND (sci-mathematics/maxima-5.45.1-r2)
 *              > dev-libs/ffcall
 *              > dev-libs/libsigsegv
 *              > dev-libs/libunistring:=
 *              > sys-libs/ncurses:=
 *              > sys-libs/readline:=
 * sci-libs/vtk < 
```

so at least it doesn't link to libvtk - maybe uses headers?

----------

=sci-libs/pcl-1.12.0

```
 * VDB: detected possibly incorrect RDEPEND (sci-libs/pcl-1.12.0)
 *                       > app-arch/lz4:=
 * dev-qt/qtconcurrent:5 < 
 * dev-qt/qtcore:5       < 
 * dev-qt/qtgui:5        < 
 * dev-qt/qtopengl:5     < 
 * media-libs/freeglut   < 
 *                       > media-libs/libpng:=
 * sci-libs/flann        < 
 * sci-libs/vtk:=        < 
 * virtual/opengl        < 
```

same here.

----------

=sci-libs/rtabmap-0.20.13-r1
success, but

```
 * VDB: detected possibly incorrect RDEPEND (sci-libs/rtabmap-0.20.13-r1)
 * dev-libs/boost:= < 
 *                  > dev-db/sqlite:3
 * sci-libs/vtk:=   < 
```

doesn't link to libvtk, only uses headers?

----------

=media-libs/opencv-4.5.4 
success with no notes.

----------


**failed** rdeps test:

----------

=sci-libs/avogadrolibs-1.93.0

/var/tmp/portage/sci-libs/avogadrolibs-1.93.0/work/avogadrolibs-1.93.0_build/avogadro/vtk/AvogadroVtk_autogen/EWIEGA46WW/../../../../../avogadrolibs-1.93.0/avogadro/vtk/vtkglwidget.h:22:10: fatal error: QVTKOpenGLWidget.h: No such file or directory
   22 | #include <QVTKOpenGLWidget.h>

compiles with older vtk...

----------

=sci-libs/opencascade-7.5.3-r3
VTK version (9.1.0)
CMake Error at CMakeLists.txt:746 (message):
  Could not find headers of used third-party products:
  3RDPARTY_VTK_INCLUDE_DIR

----------

=media-gfx/freecad-0.19.2-r5
opencascade failing with vtk-9.1.0.

----------


**unfinished** rdeps test:

----------

=sci-libs/oce-0.18.3-r3
requires older vtk, can't be tested

----------

Gonna merge since it's package.masked. Thanks a lot!

juippis:
c9edaa4fed9705cfa95690be0060b38c67d49e6a
6010f1d0f545b9ddf5f2ee3fd673332d8342821a

juippis:
Then in general hdf5 did cause some troubles, since something pulled the new one, vtk wanted the old one...

ionenwks:
> I don't fully understand the RDEPEND messages from the iwdevtools package. Does this mean, the ebuild has an RDEPEND string set for the package, but the resulting binaries don't link to the package?

Yes, with current USE for packageA in:
```
packageA < 
         > packageB
packageC | packageC:=
```
The superfluous dep check is just a guess though, package may be using something else from the package or using `dlopen()` but it doesn't hurt to look if it's really needed or not when see this. iwdevtools allows to set exclusion files if needed to silence known ones.

In case of packageB that means a missing dependency it's actually linking with. May not always be a "huge" deal if a dependency depend on it, but ignoring that tend to lead to problems later.

In C example, it's saying `:=` is probably needed given the dependency defines a subslot.

It's just meant to save some time instead of always check this manually.

https://github.com/ionenwks/iwdevtools#qa-vdb

PR #22880, opened 2021-11-10
dev-libs/snowball-stemmer: add 2.2.0

arkamar:
Version 2.2.0 has been released recently. It was necessary to rebase `shared-library` patch due to changes in
snowball-stemmer `GNUmakefile`.


dev-libs/snowball-stemmer/snowball-stemmer-2.2.0.ebuild
```diff
@@ -0,0 +1,54 @@
...
+
+S="${WORKDIR}/snowball-${PV}"
+
+PATCHES=(
+	"${FILESDIR}/${P}-shared-library.patch"
+)
+
+src_compile() {
+	export CC=$(tc-getCC)
+	export AR=$(tc-getAR)
```
thesamesam, line 34: tc-export AR CC instead (from toolchain-funcs.eclass)

dev-libs/snowball-stemmer/snowball-stemmer-2.2.0.ebuild
```diff
@@ -0,0 +1,54 @@
...
+	doheader include/libstemmer.h
+
+	dolib.so libstemmer.so.${PV}
+	dolib.so libstemmer.so.$(ver_cut 1)
+	dolib.so libstemmer.so
+
+	use static-libs && dolib.a libstemmer.a
+}
+
+src_test() {
```
thesamesam, line 52: Move this before src_install so it reflects the order of execution

arkamar:
@thesamesam thanks, I addressed both issues in FP.

dev-libs/snowball-stemmer/snowball-stemmer-2.2.0.ebuild
```diff
@@ -0,0 +1,53 @@
...
+
+MY_TESTDATA_COMMIT="0703f1d6a21802c3ff00c2c8b31bd255b74b2aec"
+
+DESCRIPTION="Snowball compiler and stemming algorithms"
+HOMEPAGE="https://snowballstem.org/ https://github.com/snowballstem/snowball/"
+SRC_URI="https://github.com/snowballstem/snowball/archive/v${PV}.tar.gz -> ${P}.tar.gz
+	test? ( https://github.com/snowballstem/snowball-data/archive/${MY_TESTDATA_COMMIT}.tar.gz -> snowball-data-${MY_TESTDATA_COMMIT}.tar.gz )"
+
+LICENSE="BSD"
+SLOT="0/$(ver_cut 1)"
```
thesamesam, line 16: I'm sorry to bother you one more time, but I wonder if you could run 'abidiff' (possibly using the iwdevtools bashrc wrapper) to see if it really has a stable ABI between e.g. 2.1.0 -> 2.2.0. I'm worried it doesn't, especially given upstream expect a static library, but it might be fine. They at least knew to increment the SONAME before, so..

thesamesam:
Thank you!

dev-libs/snowball-stemmer/snowball-stemmer-2.2.0.ebuild
arkamar: Sure, I'll check it.

arkamar: I am not sure if I did that correctly, but the ABI seems to be same
```sh
# qa-cmp -W /var/tmp/portage/dev-libs/snowball-stemmer-2.{1.0-r1,2.0}
CMP: dev-libs/snowball-stemmer-2.1.0-r1/image with dev-libs/snowball-stemmer-2.2.0/image
------> FILES(+0,-0) SONAME(+0,-0) ABI(nodebug) SIZE(+0.17%)
```
and
```sh
abidiff --verbose /var/tmp/portage/dev-libs/snowball-stemmer-2.1.0-r1/image/usr/lib64/libstemmer.so.2.1.0 /var/tmp/portage/dev-libs/snowball-stemmer-2.2.0/image/usr/lib64/libstemmer.so.2.2.0
```
does not return anything.

PR #22890, opened 2021-11-10
net-p2p/vuze: remove obsolete version | dev-java/miglayout: remove obsolete version

vaukai:


fordfrog:
so we're dropping the `x86` keyword, right?

vaukai:
> so we're dropping the `x86` keyword, right?

No.  The point is that `dev-java/swt:4.10` does not support `x86`. So we cannot lift its reverse dependencies to that branch.

fordfrog:
well, if we remove the older versions of these packages, we effectively remove `x86` keyword from those packages completely. i understand that newer swt supports only `amd64` and `ppc64`. do we really need to remove the old version(s) of `swt` as of now?

vaukai:
> well, if we remove the older versions of these packages, we effectively remove `x86` keyword from those packages completely. i understand that newer swt supports only `amd64` and `ppc64`. do we really need to remove the old version(s) of `swt` as of now?

No.  I am asking for removal of the newer revbumps, not the old versions.

fordfrog:
ah, my bad :-) thanks for explanation :-)

PR #22910, opened 2021-11-12
dev-java/*: various cleanups

ajakk:
Multiple cleanups necessary to cleanup commons-httpclient


fordfrog:
all those slots are still used so unless we move the depending apps to newer slots, we cannot drop these slots

ajakk:
Yeah, mostly just checking CI here to see what happens.

vaukai:
The security issue which was assigned to `dev-java/commons-httpclient:3` in fact was for a different package.
Slot 4 however should be [transferred to that other package](https://github.com/gentoo/gentoo/pull/23703).

PR #22924, opened 2021-11-13
Bump version of vim, vim-core and gvim to v8.2.3582

vaibhav-rustagi:


thesamesam:
Thanks! We'll need to file a bug for these CVEs and tag the commits with them (e.g. in the footer Bug: https://bugs.gentoo.org/123456). I'll do that now as you've already done the PR and there's no point in delaying this just for a formality.

Appreciate you doing this, will look shortly.

PR #22929, opened 2021-11-13
media-libs/svt-av1: remove no-force-avx patch

telans:
Aparently fixed by upstream.

Closes: https://bugs.gentoo.org/823140


juippis (approved):
I'll use the patch in the linked bug since it's identical and was pushed first. A simple "lgtm" comment in the bug would've been enough :P

PR #22938, opened 2021-11-13
sys-auth/elogind: add patch to enable clang-13 compilation

geraint0923:
The patch is extract from upstream:
https://github.com/elogind/elogind/commit/cd5390335303da11f3aeaa5431dd7a88c5a9b3e9

The related error is discussed at:
https://githubmemory.com/repo/elogind/elogind/issues/215

Signed-off-by: Yang Yang <geraint0923@gmail.com>

### DIFF
--- a/sys-auth/elogind/elogind-246.10-r3.ebuild
+++ b/sys-auth/elogind/elogind-246.10-r3.ebuild
@@ -52,6 +52,7 @@ PATCHES=(
 	"${FILESDIR}/${PN}-243.7-nodocs.patch"
 	"${FILESDIR}/${PN}-241.4-broken-test.patch" # bug 699116
 	"${FILESDIR}/${P}-revert-polkit-automagic.patch"
+	"${FILESDIR}/${P}-clang-undefined-symbol.patch"
 )
 
 pkg_setup() {
--- a/sys-auth/elogind/files/elogind-246.10-clang-undefined-symbol.patch
+++ b/sys-auth/elogind/files/elogind-246.10-clang-undefined-symbol.patch
@@ -1,0 +1,20 @@
+--- a/src/libelogind/sd-bus/bus-error.h
++++ b/src/libelogind/sd-bus/bus-error.h
+@@ -28,11 +28,17 @@ int bus_error_set_errnofv(sd_bus_error *e, int error, const char *format, va_lis
+  * the bus error table, and BUS_ERROR_MAP_ELF_USE has to be used at
+  * least once per compilation unit (i.e. per library), to ensure that
+  * the error map is really added to the final binary.
++ *
++ * In addition, set the retain attribute so that the section cannot be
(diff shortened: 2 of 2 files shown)


sys-auth/elogind/elogind-246.10-r2.ebuild
```diff
@@ -51,7 +51,8 @@ DOCS=( README.md src/libelogind/sd-bus/GVARIANT-SERIALIZATION )
 PATCHES=(
 	"${FILESDIR}/${PN}-243.7-nodocs.patch"
 	"${FILESDIR}/${PN}-241.4-broken-test.patch" # bug 699116
-	"${FILESDIR}/${P}-revert-polkit-automagic.patch"
+	"${FILESDIR}/${PN}-246.10-revert-polkit-automagic.patch"
```
telans, line 54: `${P}` still works here, `${PF}` wouldn't though

sys-auth/elogind/elogind-246.10-r2.ebuild
```diff
@@ -51,7 +51,8 @@ DOCS=( README.md src/libelogind/sd-bus/GVARIANT-SERIALIZATION )
...
+	"${FILESDIR}/${PN}-246.10-clang-undefined-symbol.patch"
```
telans, line 55:
```suggestion
	"${FILESDIR}/${P}-clang-undefined-symbol.patch"
```

sys-auth/elogind/elogind-246.10-r2.ebuild
telans, line 54:
```suggestion
	"${FILESDIR}/${P}-revert-polkit-automagic.patch"
```

geraint0923: Addressed the comment, changed back to `${P}`

geraint0923:
Thanks for the review, I just addressed the comments, @telans please take a look.

geraint0923:
Kindly ping:) 

Please let me know how I can help with this fix, given that clang-13 is already marked stable in Gentoo, this issue will affect anyone with the latest stable clang who wants to use sys-auth/elogind.

telans:
Generally it just takes a while for devs to review packages, I wouldn't give it too much thought

slashbeast:
No all commits are signed off, can you re-do the branch and put all your changes into single signed commit please? And confirm that the patch was sent to upstream as well.

yemouu:
The patch is from an upstream commit

geraint0923:
@slashbeast Thanks for the comment, I have squashed the commits into one single commit, and signed off the single commit. And this fix is actually extracted from the upstream: https://github.com/elogind/elogind/commit/cd5390335303da11f3aeaa5431dd7a88c5a9b3e9 .

Please help take a look, thanks!

thesamesam:
Thanks! I undid the revbump though as in most cases, this is a build-time only error (and it's a very uncommon thing to be setting, especially at this point).

PR #22950, opened 2021-11-14
Move some UIDs and GIDs out of 0..100 range, by QA policy

ulm:


ConiKost:
Shouldn't be acct-* packages be revbumped, so everyone having already installed them gets the new uid/gid?

ulm:
> Shouldn't be acct-* packages be revbumped, so everyone having already installed them gets the new uid/gid?

But they won't. Once the package is installed, the ID in /etc/group or /etc/passwd is fixed and won't be updated.

floppym:
The acct packages install files in /usr/lib/sysusers.d, and the content of those files will change. Please revbump.

floppym:
Also, a revbump will cause a (desired) error on systems where the sysadmin has set ACCT_{GROUP,USER}_ENFORCE_ID in make.conf.

floppym (approved):

ConiKost (approved):

ulm:
> ## Pull request CI report
> 
> _Report generated at_: 2021-11-14 21:15 UTC
> _Newest commit scanned_: [0192e2a](https://github.com/gentoo/gentoo/commit/0192e2ae57deab9e5e1e438da1ff05ecf84db74d)
> _Status_: x **broken**
> 
> New issues caused by PR:
> https://qa-reports.gentoo.org/output/gentoo-ci/0ecfeb43c2/output.html#acct-group/burrow_exporter
> [...]

These are spurious warnings, because pkgcheck still has 499 as maximum. I've filed https://github.com/pkgcore/pkgcheck/issues/356 for pkgcheck.

gyakovlev:
`acct-group/systemd-hostname` is only required for systemd-9999 for now, so it should be safe to move.
and sorry for using low id in the first place, I missed QA lead approval process.

PR #22954, opened 2021-11-15
package.mask: mask bitcoin <0.21.1

luke-jr:
While it remains not good to unmask later versions, it is now also unsafe to
continue to use these old ones.

### DIFF
--- a/profiles/package.mask
+++ b/profiles/package.mask
@@ -631,6 +631,22 @@ net-libs/libsoup:3.0
 # breaks if not all at least many revdeps. #805011 for tracker bug.
 >=net-libs/mbedtls-3.0.0
 
+# Luke Dashjr <luke-jr+gentoobugs@utopios.org> (2021-11-14)
+# Bitcoin has activated the Taproot protocol change, which is not enforced by
+# versions of Bitcoin Core/Knots prior to 0.21.1. This means these old
+# versions are no longer providing full node security, and are effectively
+# light wallets. However, protocol changes require user consent to be
+# effective, and inconsistent enforcement can lead to loss of funds on either
+# or both sides of the disagreement. As such, it is unethical for anyone to
+# make the decision but you, so you must opt-in to them explicitly (or opt to
+# reject Taproot - not supported by any current ebuild). If you do not know
+# what you are doing, learn more before receiving or sending further bitcoins.
+# (You must make a decision either way - simply not upgrading is insecure.)
+# To learn more, see https://bitcointaproot.cc
+<net-p2p/bitcoind-0.21.1
+<net-p2p/bitcoin-qt-0.21.1
+<net-libs/libbitcoinconsensus-0.21.1
+
 # Luke Dashjr <luke-jr+gentoobugs@utopios.org> (2021-11-04)
 # This release adds enforcement of the Taproot protocol change to the Bitcoin
 # rules, beginning in November. Protocol changes require user consent to be


profiles/package.mask
```diff
@@ -631,6 +631,22 @@ net-libs/libsoup:3.0
...
+# effective, and inconsistent enforcement can lead to loss of funds on either
+# or both sides of the disagreement. As such, it is unethical for anyone to
+# make the decision but you, so you must opt-in to them explicitly (or opt to
+# reject Taproot - not supported by any current ebuild). If you do not know
+# what you are doing, learn more before receiving or sending further bitcoins.
+# (You must make a decision either way - simply not upgrading is insecure.)
+# To learn more, see https://bitcointaproot.cc
+<net-p2p/bitcoind-0.21.1
+<net-p2p/bitcoin-qt-0.21.1
+<net-libs/libbitcoinconsensus-0.21.1
```
ajakk, line 648: If all bitcoin versions are going to be masked, why not merge the two masks? 

I don't think masking all versions is a good idea anyway, why not just clean up old versions and drop the mask?

luke-jr: Hopefully it will just be for a short time that both are masked. Right now, Taproot's activation is still somewhat ambiguous, so it remains important switching to the newer versions remains an explicit opt-in to Taproot.

candrews: > Taproot's activation is still somewhat ambiguous

FUD. Taproot was officially activated on Sunday on block 709,632.



> why not just clean up old versions and drop the mask?

Please, let's do that.

luke-jr: >FUD. Taproot was officially activated on Sunday on block 709,632.

You and I both subjectively think so, but are you willing to mine a block violating the Taproot rules with economic incentives for non-enforcing nodes to prove that objectively? Bitcoin is not a centralised network where this or that person dictates to others what the rules are.

As things stand right now, only about half the network appears to be enforcing Taproot, far from the necessary 85% to say anything certain.

ajakk: Gentoo doesn't really need to be opinionated about this. If you think both <0.21.1 and >=0.21.1 should be available, then let's keep them both available rather than masking all Bitcoin versions.

luke-jr: Both should be available, but users should not get >=0.21.1 unless they explicitly want to change their protocol rules (not merely upgrade software); There were better ways to do this than the masking approach, but that's what we ended up with months ago.

Since <0.21.1 is now insecure, it makes sense to mask it as well.

Flowdalic: I agree that an explicit opt-in for taproot is sensible. That said, masking all versions of bitcoind seems paradoxical. We should trust users to inform themselves about changes in newer versions of the software they use. This is especially true if the software is bitcoin. My idea would be to have the package display something like the following disclaimer (probably using the readme.gentoo-r1 eclass):

> WARNING: Bitcoin is a highly dynamic ecosystem. New software versions
> may change or introduce new consensus rules. Those updates may
> affect your financial risk. In addition, world updates of your system may
> pull in such new versions. Therefore, you may want to consider pinning the
> used version using a package mask to prevent unattended updates.
> 
> Always make sure to agree to the items found in ${PN}'s changelog
> before updating.

and remove the profile's package mask.

luke-jr: > That said, masking all versions of bitcoind seems paradoxical. 

It accomplishes the intended purpose: users are informed they have a decision they must make, and how to make it.

>My idea would be to have the package display something like the following disclaimer

A disclaimer doesn't fix the issue. The decision must be made by the end user, not merely disclaimed by some other decision-maker.

Flowdalic: > The decision must be made by the end user,

I keep asking myself if users do not make the decision by installing the software, Bitcoin in this case, and if my suggested disclaimer does not simply state the obvious. I am also curious why Bitcoin does not simply ask for user consent on the first invocation itself. Seems strange to view this as as something that the user has to opt-in, but defer the actual opt-in mechanism to the distributions, not not built it in the software directly.

luke-jr: >I keep asking myself if users do not make the decision by installing the software,

As long as they actually decide to install the software, which is not certain to be the case with bulk system updates.

>I am also curious why Bitcoin does not simply ask for user consent on the first invocation itself.

We (upstream) intentionally avoid supporting any automatic updates.

Where provided (eg, the Ubuntu PPA), similar mechanisms have been added to ensure explicit user opt-in for Taproot (eg, the .deb has an interactive prompt during update from a pre-Taproot version).

> Seems strange to view this as as something that the user has to opt-in, but defer the actual opt-in mechanism to the distributions, not not built it in the software directly.

I and other devs generally agree it would be a good idea to have something built-in, but that does not exist yet.

Flowdalic: > As long as they actually decide to install the software, which is not certain to be the case with bulk system updates.

Bulk system updates are always user triggered, even if they are automatically scheduled and unattended. Since Gentoo does not come with such a feature enabled per default, at one point the user had decided to perform a bulk update. Bulk updates do not happen out of thin air. I know that this can be seen as far fetched, but I am wondering if this isn't enough.

>> I am also curious why Bitcoin does not simply ask for user consent on the first invocation itself.
> We (upstream) intentionally avoid supporting any automatic updates.

I know, but is that really an answer why you should not be asking the user for consent if data from a previous version was found? After all, the user could very well compile a newer version of Bitcoin Core from source and run it. Right now, it appears to me that you don't have the manpower to build such a feature.

But then again, it is strange to ask the user if he really wants to run what he just tried to execute…

luke-jr: > Bulk system updates are always user triggered ...

Yes, but it's distinct from upgrading the single package or consensus rules specifically. There's a huge difference between "update my system" and "start imposing new rules on other Bitcoin actors I do business with".

>I know, but is that really an answer why you should not be asking the user for consent if data from a previous version was found? After all, the user could very well compile a newer version of Bitcoin Core from source and run it.

That's a very specific and intentional action, at least. And generally, to get a newer version, you have to go through a webpage that is very explicit about it including Taproot. But again, yes, it would make sense to add an additional safeguard.

PR #22960, opened 2021-11-15
sys-libs/glibc: replace systemd reexec with a warning

floppym:


mgorny:
How about warning early in `pkg_pretend`?

mattst88:
I saw some chatter on IRC, but I didn't fully understand. It sounds like there are some cases where it's possible that you're not able to restart systemd, and as a result we shouldn't try because doing so could leave the system in an unusable state?

floppym:
> It sounds like there are some cases where it's possible that you're not able to restart systemd, and as a result we shouldn't try because doing so could leave the system in an unusable state?

Yes, that's essentially what @Whissi is saying.

floppym:
> How about warning early in pkg_pretend?

That sounds reasonable to me, but I would like feedback from the toolchain project before I update the PR.

Whissi:
> How about warning early in `pkg_pretend`?

I would prefer `pkg_postinst`. `pkg_pretend` is happening too early and I believe it won't be visible anymore if user has verbose build enabled.

mgorny:
That's nonsense. The whole purpose of `pkg_pretend` is to make the messages visible while the user is still at the terminal and can make decisions *before* his system enters a broken state.

Also, nothing stops you from warning twice.

thesamesam:
While I like the idea of warning users on glibc major version upgrades, there's two issues here:
- Why isn't it okay to restart systemd in glibc, but it is okay to do it in the systemd ebuilds? When we know something breaks, I'm not yet convinced that inaction is helpful.
- Outside of this bug/PR, we should _explicitly_ formulate whatever policy there is/isn't on touching system daemons within ebuilds. And in any case, a news item is probably warranted for this cycle then?

floppym:
> Why isn't it okay to restart systemd in glibc, but it is okay to do it in the systemd ebuilds?

I have removed the reexec from the systemd ebuilds.

I also updated this PR to emit the warning in pkg_pretend and pkg_postinst.

PR #22961, opened 2021-11-15
net-misc/openssh: Upstream fix for CVE-2021-41617 for v8.5

hector-gao:
Patch source: https://src.fedoraproject.org/rpms/openssh/blob/f33/f/openssh-8.7p1-upstream-cve-2021-41617.patch

### DIFF
--- a/net-misc/openssh/openssh-8.5_p1-r3.ebuild
+++ b/net-misc/openssh/openssh-8.5_p1-r3.ebuild
@@ -132,6 +132,7 @@ src_prepare() {
 	eapply "${FILESDIR}"/${PN}-7.5_p1-disable-conch-interop-tests.patch
 	eapply "${FILESDIR}"/${PN}-8.0_p1-fix-putty-tests.patch
 	eapply "${FILESDIR}"/${PN}-8.0_p1-deny-shmget-shmat-shmdt-in-preauth-privsep-child.patch
+	eapply "${FILESDIR}"/${PN}-8.5_p1-upstream-cve-2021-41617.patch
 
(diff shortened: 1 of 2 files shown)


net-misc/openssh/openssh-8.5_p1-r3.ebuild
```diff
@@ -132,6 +132,7 @@ src_prepare() {
 	eapply "${FILESDIR}"/${PN}-7.5_p1-disable-conch-interop-tests.patch
 	eapply "${FILESDIR}"/${PN}-8.0_p1-fix-putty-tests.patch
 	eapply "${FILESDIR}"/${PN}-8.0_p1-deny-shmget-shmat-shmdt-in-preauth-privsep-child.patch
+	eapply "${FILESDIR}"/${PN}-8.7p1-upstream-cve-2021-41617.patch
```
thesamesam, line 135: There’s a typo in the actual patch file name so this won’t apply

hector-gao: Just fixed the patch name. Thank you!

hector-gao:
The fedora patch at https://src.fedoraproject.org/rpms/openssh/blob/f33/f/openssh-8.7p1-upstream-cve-2021-41617.patch applies to auth.c. However, the target function of the upstream fix is in misc.c. So I created a patch on branch V_8_5 with the same change.

mpagano:
Thanks for your work, hector-gao. Sorry this got missed in the mass.   At this time, no version in the tree is impacted by this.

PR #22964, opened 2021-11-15
app-editors/kakoune bump to 2021.11.08

mujo-hash:
Bump and drop old. Hopefully we can stabilize this version later. No patches required.


juippis (approved):
Thanks!

PR #22965, opened 2021-11-15
net-im/telegram-desktop: Bump to 3.2.4

mid-kid:
@gyakovlev


mid-kid:
@gyakovlev sidenote: webkitgtk support still requires the dbus use flag to be enabled. Should I make note of this somehow in the optfeature, or is that enough reason to keep the webkit use flag?

gyakovlev:
hey, testing.

no, we can simply mention it in the optfeature line
like this:
```
optfeature "shop payment support (requires USE=dbus enabled)" net-libs/webkit-gtk
```
I add it myself so you don't have to push/rebase/etc

gyakovlev:
pushed 1 minor improvements
https://github.com/gentoo/gentoo/commit/50b25e5efa6611d12e6030740db231eb77ce3025
https://github.com/gentoo/gentoo/commit/29f7dd1f90e4e04bc0f5a6cd8e21896b2679aad2

works fine for me on ppc64

mid-kid:
Thanks!

PR #22966, opened 2021-11-16
vulkan-sdk: Use the upstream sdk version

orbea:
This changes all of the applicable vulkan-sdk ebuilds to use the upstream vulkan-sdk versions and git tags.

Most of the projects associated with the vulkan-sdk all follow the sdk version and have corresponding git tags like `sdk-1.2.189.0`. These projects often have other release tags that may or may not correspond to stable vulkan-sdk version like `v1.2.189` or `v1.2.191` in the case of vulkan-tools (https://github.com/KhronosGroup/Vulkan-Tools/tags). In the past Gentoo has updated to versions that do not correspond to the upstream stable versions which can have unintended consequences (https://gitweb.gentoo.org/repo/gentoo.git/commit/dev-util/vulkan-tools?id=fc0c7ffef2e01b2599211b144d8fba270e1cad97). For example when I noticed this it was because I could not build gfxreconstruct (https://github.com/LunarG/gfxreconstruct/tags) which ships sdk tags as well as a second unrelated versioning scheme that doesn't necessarily correspond to the rest of the vulkan-sdk.

The best way to ensure Gentoo only ships the correct stable versions is use the sdk tags to have better consistency with upstream as recorded at vulkan.lunarg.com. Upstream has been consistent with this versioning scheme for a long time and there should be no worries of incompatibilities. Users should not notice any difference with this update.

https://vulkan.lunarg.com/sdk/home#linux

Some projects that follow a different versioning scheme are glslang, spirv-tools and spirv-headers, but they should not have an issue here. In the past it was required to look at the `known_good.json` files in the vulkan-sdk projects as well in glslang to find the compatible commit hashes, but this seems to have largely improved in the recent past where the latest stable tags work.

### DIFF
--- a/dev-util/vulkan-headers/vulkan-headers-9999.ebuild
+++ b/dev-util/vulkan-headers/vulkan-headers-9999.ebuild
@@ -10,9 +10,9 @@ if [[ ${PV} == *9999* ]]; then
 	EGIT_REPO_URI="https://github.com/KhronosGroup/${MY_PN}.git"
 	inherit git-r3
 else
-	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/sdk-${PV}.tar.gz -> ${P}.tar.gz"
 	KEYWORDS="~amd64 ~ppc64 ~x86"
-	S="${WORKDIR}"/${MY_PN}-${PV}
+	S="${WORKDIR}"/${MY_PN}-sdk-${PV}
 fi
 
 DESCRIPTION="Vulkan Header files and API registry"
--- a/dev-util/vulkan-tools/vulkan-tools-9999.ebuild
+++ b/dev-util/vulkan-tools/vulkan-tools-9999.ebuild
@@ -13,9 +13,9 @@ if [[ ${PV} == *9999* ]]; then
 	EGIT_SUBMODULES=()
 	inherit git-r3
 else
-	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/sdk-${PV}.tar.gz -> ${P}.tar.gz"
 	KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~riscv"
-	S="${WORKDIR}"/${MY_PN}-${PV}
+	S="${WORKDIR}"/${MY_PN}-sdk-${PV}
 fi
 
 DESCRIPTION="Official Vulkan Tools and Utilities for Windows, Linux, Android, and MacOS"
--- a/media-libs/vulkan-layers/vulkan-layers-9999.ebuild
+++ b/media-libs/vulkan-layers/vulkan-layers-9999.ebuild
@@ -13,9 +13,9 @@ if [[ ${PV} == *9999* ]]; then
 	EGIT_SUBMODULES=()
 	inherit git-r3
 else
-	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/sdk-${PV}.tar.gz -> ${P}.tar.gz"
 	KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~riscv ~x86"
-	S="${WORKDIR}"/${MY_PN}-${PV}
+	S="${WORKDIR}"/${MY_PN}-sdk-${PV}
 fi
 
 DESCRIPTION="Vulkan Validation Layers"
--- a/media-libs/vulkan-loader/vulkan-loader-9999.ebuild
+++ b/media-libs/vulkan-loader/vulkan-loader-9999.ebuild
@@ -12,9 +12,9 @@ if [[ ${PV} == *9999* ]]; then
 	EGIT_SUBMODULES=()
 	inherit git-r3
 else
-	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/sdk-${PV}.tar.gz -> ${P}.tar.gz"
 	KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~ppc64 ~riscv ~x86"
-	S="${WORKDIR}"/${MY_PN}-${PV}
+	S="${WORKDIR}"/${MY_PN}-sdk-${PV}
 fi
 
 DESCRIPTION="Vulkan Installable Client Driver (ICD) Loader"


sarnex:
Does this need a revbump? Is the distfile hash the exact same? If the distfiles are different we should probably revbump.

Also, this change has no real effect right, it's just different tags that we should follow in the future?

media-libs/vulkan-layers/vulkan-layers-1.2.189.0.ebuild
```diff
@@ -13,9 +13,9 @@ if [[ ${PV} == *9999* ]]; then
 	EGIT_SUBMODULES=()
 	inherit git-r3
 else
-	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/sdk-${PV}.tar.gz -> ${P}.tar.gz"
```
mattst88, line 16: There's a `sdk-1.2.189.2` tag in Vulkan-ValidationLayers. Should we go for that instead of .0?

I'd prefer a version bump to .2 rather than changing a stable ebuild's distfile.

media-libs/vulkan-loader/vulkan-loader-1.2.189.0.ebuild
```diff
@@ -12,9 +12,9 @@ if [[ ${PV} == *9999* ]]; then
 	EGIT_SUBMODULES=()
 	inherit git-r3
 else
-	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/sdk-${PV}.tar.gz -> ${P}.tar.gz"
```
mattst88, line 15: There's a `sdk-1.2.189.1` tag in Vulkan-Loader. Should we go for that instead of .0?

I'd prefer a version bump to .1 rather than changing a stable ebuild's distfile.

dev-util/vulkan-tools/vulkan-tools-1.2.189.0.ebuild
```diff
@@ -13,9 +13,9 @@ if [[ ${PV} == *9999* ]]; then
 	EGIT_SUBMODULES=()
 	inherit git-r3
 else
-	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/sdk-${PV}.tar.gz -> ${P}.tar.gz"
```
mattst88, line 16: There's a `sdk-1.2.189.1` tag in Vulkan-Tools. Should we go for that instead of .0?

I'd prefer a version bump to .1 rather than changing a stable ebuild's distfile.

dev-util/vulkan-headers/vulkan-headers-1.2.189.0.ebuild
```diff
@@ -10,9 +10,9 @@ if [[ ${PV} == *9999* ]]; then
 	EGIT_REPO_URI="https://github.com/KhronosGroup/${MY_PN}.git"
 	inherit git-r3
 else
-	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/sdk-${PV}.tar.gz -> ${P}.tar.gz"
```
mattst88, line 13: There's a `sdk-1.2.189.1` tag in Vulkan-Headers. Should we go for that instead of .0?

I'd prefer a version bump to .1 rather than changing a stable ebuild's distfile.

mattst88 (changes requested):
I'd rather not modify the stable ebuilds. It looks like in all cases there is a newer .1 or .2 sdk tag we can version bump to.

Also, please always update the -9999 ebuilds. Typically we bump the packages by copying from -9999.

orbea:
> Also, please always update the -9999 ebuilds. Typically we bump the packages by copying from -9999.

Good point, I will update this later today or tomorrow.

media-libs/vulkan-layers/vulkan-layers-1.2.189.0.ebuild
orbea: I am unsure, the lunarg site lists `1.2.189.0` as the latest version for both linux and osx and `1.2.189.2` for windows only. I presume there are windows only fixes in those newer tags. Regardless I suspect a new sdk release will be out in the near future, I see a `sdk-1.2.198.0` tag on the gfxreconstruct repo already, but the lunarg site has not been updated yet so maybe its not finalized yet.

orbea:
I updated this to only touch the `9999` ebuilds.

mattst88 (approved):
Looks good to me.

A couple questions:

1. do we know if the SDK-tags are any different than the non-SDK tag of the same version?
2. should we avoid ever shipping non-SDK versions, or is not stabilizing non-SDK versions sufficient?

sarnex:
> Looks good to me.
> 
> A couple questions:
> 
>     1. do we know if the SDK-tags are any different than the non-SDK tag of the same version?
> 
>     2. should we avoid ever shipping non-SDK versions, or is not stabilizing non-SDK versions sufficient?

Agree with Matt's questions, I'd like to see information stating only the sdk tagged ones are intended to be packed. Upstream is such a shitshow I don't really believe anything

orbea:
Sorry for the slow response.

> 1. do we know if the SDK-tags are any different than the non-SDK tag of the same version?

For years these are the only tags that all of the relevant repos consistently follow, I think historically it has been shown to be true. These are also the tags that get picked up for the upstream vulkan-sdk releases.

> 2. should we avoid ever shipping non-SDK versions, or is not stabilizing non-SDK versions sufficient?

I think they should be avoided in general, in the past using non sdk tags as mentioned in the op has led to serious compatibility conflicts with different parts of the sdk.

> I'd like to see information stating only the sdk tagged ones are intended to be packaged/used.

There might be more places discussing this, but its documented in the vulkan-headers README.md

> Note: Marked version releases have undergone thorough testing but do not imply the same quality level as SDK tags. SDK tags follow the sdk-<version>.<patch> format (e.g., sdk-1.1.92.0).

https://github.com/KhronosGroup/Vulkan-Headers#version-tagging-scheme

Also a explanation on the relevant issue. https://github.com/KhronosGroup/Vulkan-Headers/issues/23#issuecomment-445980357

orbea:
> Upstream is such a shitshow I don't really believe anything

Also if you think that then why not submit issue reports or pull requests upstream? In my experience they have been mostly responsive and reasonable. They merged all of my trivial build fixes.

sarnex:
> > Upstream is such a shitshow I don't really believe anything
> 
> Also if you think that then why not submit issue reports or pull requests upstream? In my experience they have been mostly responsive and reasonable. They merged all of my trivial build fixes.

Unfortunately, it's not that simple. Matt and I have been trying to get them to follow basic software release practices for years to little to no avail. One example is the still open https://github.com/KhronosGroup/SPIRV-Headers/issues/199

sarnex (approved):
But I'm okay with this since it does appear that they want us to use the sdk tags. I'll wait to see if Matt has any follow up comments on your recent answers

orbea:
Yea, good point with SPIRV-Headers, but for the sake of argument that repo, SPIRV-Tools and glslang all follow an unrelated versioning scheme not affected by this PR. Within the strict vulkan-sdk repos (Also vulkan-headers) my experience has been much better. If all else fails following the commits in the `known_good.json` files, for example from Vulkan-ValidationLayers repo should always end up with a working result even if its not ideal...

orbea:
Also there is also an issue tracker at vulkan.lunarg.com which sometimes might work better.

https://vulkan.lunarg.com/issue/home

PR #22981, opened 2021-11-17
net-misc/rsync: fix CVE-2020-14387

vteratipally:
Fix CVE-2020-14387 - the fix for the  CVE-2020-14387 is not released in 3.2.3. As per the https://download.samba.org/pub/rsync/NEWS, the last released version is over an year ago. Cherry-picking the fix to the current version.

Bug: https://bugs.gentoo.org/792576

Signed-off-by: Varsha Teratipally teratipally@google.com

### DIFF
--- a/net-misc/rsync/rsync-3.2.3-r5.ebuild
+++ b/net-misc/rsync/rsync-3.2.3-r5.ebuild
@@ -36,8 +36,10 @@ DEPEND="${RDEPEND}"
 
 src_prepare() {
 	local PATCHES=(
-		"${FILESDIR}/rsync-3.2.3-glibc-lchmod.patch"
-		"${FILESDIR}/rsync-3.2.3-cross.patch"
+		"${FILESDIR}/${P}-glibc-lchmod.patch"
+		"${FILESDIR}/${P}-cross.patch"
+		# Fix for (CVE-2020-14387) - net-misc/rsync: improper TLS validation in rsync-ssl script
+		"${FILESDIR}/${P}-verify-certificate.patch"
 	)
 	default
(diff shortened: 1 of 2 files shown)


scardracs:
Please change title in net-misc/rsync: fix CVE-2020-14387. This helps other on understand what you are actually doing

scardracs:
Also please add a commit with BUG: https://bugs.gentoo.org/792576. This helps bot adding a note for it on bgo

scardracs:
No, it's a CVE. You can't close it but only add BUG

vteratipally:
> No, it's a CVE. You can't close it but only add BUG, It will be closed by security team.

Thank you for the review. I updated the bug in the commit

net-misc/rsync/rsync-3.2.3-r5.ebuild
```diff
@@ -38,6 +38,7 @@ src_prepare() {
 	local PATCHES=(
 		"${FILESDIR}/rsync-3.2.3-glibc-lchmod.patch"
 		"${FILESDIR}/rsync-3.2.3-cross.patch"
+		"${FILESDIR}/rsync-3.2.3-CVE-2020-14387.patch"
```
scardracs, line 41: I'd suggest to use a more comprehensive for a patch, like ${P}-verify-ceritificate.patch instead of CVE number. (I think it's better to use ${P} instead of full rsync-3.2.3, same thing for other patches

thesamesam:
Thanks! I've adjusted the commit message to be prefixed with 'net-misc/rsync:" and changed it so we keep the old version but revbump to ~arch. We'll then stable it shortly in Gentoo.

PR #22983, opened 2021-11-17
Bitcoin 22.0

luke-jr:
Unsure if the USE=external-signer is the best way to handle that functionality (it depends on various other software not in the tree and configuration), but seems good enough for now.

Also, `system-leveldb` no longer has a performance impact, so maybe we want to drop the USE flag (and always use the system LevelDB)?


luke-jr:
Also, system-leveldb no longer has a performance impact, so maybe we want to drop the USE flag (and always use the system LevelDB)?

luke-jr:

Flowdalic:
> Unsure if the USE=external-signer is the best way to handle that functionality (it depends on various other software not in the tree and configuration), but seems good enough for now.

Am I missing something or does external-signer nothing atm?

> Also, `system-leveldb` no longer has a performance impact, so maybe we want to drop the USE flag (and always use the system LevelDB)?

I think you are the one who can judge this best. So if you believe that it can be dropped, then let's drop it. The more we un-bundel the better IMHO.

luke-jr:
1. Rebased to resolve conflicts in Manifest from older versions being dropped
2. Fixed `external-signer` USE flag to actually control build
3. Renamed `bpf` USE flag to the more common `systemtap`
4. Dropped the `system-leveldb` USE flag; it is now always enabled

Flowdalic:
Thanks for you contribution. I've fixed the EAPI fluke myself.

PR #22990, opened 2021-11-18
app-misc/task: cleanup and revbump for fish completions 

janko-j1:
I created a revbump to add fish completion files following the rule: "installing additional files that could be useful to the user". Please let me know if it is more a case of "installing an additional file of relatively little value".

### DIFF
--- a/app-misc/task/task-2.6.1-r1.ebuild
+++ b/app-misc/task/task-2.6.1-r1.ebuild
@@ -1,0 +1,70 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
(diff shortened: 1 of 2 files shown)


scardracs:
Please always add what are you doing in title (for eg: app-misc/task: revbump for fish completion). In that way other people/devs know what are you actually doing

thesamesam:
I think the commit message is OK as is but your suggestion is an improvement (just I wouldn’t block / wait on it).

janko-j1:
When pushing several commits I tend to let the global github title as is as it may be too long. As Sam said everything is in the commit message, but I'll try and summarize from now on.

I'll keep it as a revbump then.

PR #22994, opened 2021-11-19
selective package de-stabilization on ~ppc

gyakovlev:
CI torture checks for for https://bugs.gentoo.org/823768


thesamesam:
We probably can’t do ipython given it gets dragged in for tests a lot in Python land. If we can drop it now, chances are we would have to restore it again soon :(

gyakovlev:
the intention to de-stabilize some revdeps as well.

ipython/jupyter is the most painful stack to bump/stabilize afaik.

thesamesam:
The problem is that we AFAIK intentionally stabled it and the stack because of how common of a dep it is (like scipy, matplotlib)

gyakovlev:
ok that was bad idea indeed, closing

PR #22995, opened 2021-11-19
app-editors/vscod*: version bump to 1.62.3

AdelKS:
Hello,

The usual bump, nothing to report. It would be nice to have a tool that says if ever the package has new files when compared to the old one so it's easier to track binary packages like this one.

Adel


juippis (approved):
> The usual bump, nothing to report. It would be nice to have a tool that says if ever the package 
> has new files when compared to the old one so it's easier to track binary packages like this one.

You're in luck: Such a tool exists! And you know about it already ;)

https://github.com/ionenwks/iwdevtools#qa-cmp
https://github.com/ionenwks/iwdevtools#eoldnew

AdelKS:
I thought that I already asked this, don't know why I forgot about it. Thanks !

PR #23000, opened 2021-11-19
media-libs/mesa: Consolidate Intel VIDEO_CARDS settings

mattst88:
Two simple patches, removing IUSE="egl gbm" and fixing a bug in `REQUIRED_USE`.

The meat is in the third patch:

> VIDEO_CARDS has grown out of control in recent years, and the options don't really make sense outside of a single package. For example, VIDEO_CARDS=iris makes sense for Mesa, because it enables the Gen8+ Gallium OpenGL driver named Iris. VIDEO_CARDS=i965 makes sense as well because it enables the Gen4-Gen11 classic OpenGL driver named i965.
> 
> But how these VIDEO_CARDS settings should extend to other packages was never clearly defined. Until commit 564e4e3d1b14 ("x11-libs/libva: use optfeatures instead of VIDEO_CARDS") pulled in drivers depending on the VIDEO_CARD setting, but x11-libs/libva-intel-driver and x11-libs/libva-intel-media-driver support only partially overlapping sets of hardware, and neither correspond well to "i965" or "iris".
> 
> In truth, this level of flexibility (for example, the ability to build only i965 but not the older i915, or iris but not crocus) isn't worth the hassle and confusion it causes users. The insane marketing names of GPUs also doesn't make the situation any clearer.

A fourth patch does the same for Radeon, but I think I don't want to commit that now:

> I think I want to hold off on this until r300 is split off into the Mesa "Amber" LTS branch. Otherwise, the gallium/llvm requirements from r600/radeonsi would be required for r300 users.
>
> Additionally, radeonsi is masked on a number of architectures where r300 is possible. Combining the flags would make that situation difficult to manage.

### DIFF
--- a/media-libs/mesa/mesa-9999.ebuild
+++ b/media-libs/mesa/mesa-9999.ebuild
@@ -27,7 +27,7 @@ RESTRICT="
 "
 
 RADEON_CARDS="r300 r600 radeon radeonsi"
-VIDEO_CARDS="${RADEON_CARDS} crocus freedreno i915 intel iris lima nouveau panfrost v3d vc4 virgl vivante vmware"
+VIDEO_CARDS="${RADEON_CARDS} freedreno intel lima nouveau panfrost v3d vc4 virgl vivante vmware"
 for card in ${VIDEO_CARDS}; do
 	IUSE_VIDEO_CARDS+=" video_cards_${card}"
 done
@@ -38,7 +38,7 @@ IUSE="${IUSE_VIDEO_CARDS}
 	vulkan-overlay wayland +X xa xvmc zink +zstd"
 
 REQUIRED_USE="
-	d3d9?   ( || ( video_cards_iris video_cards_r300 video_cards_r600 video_cards_radeonsi video_cards_nouveau video_cards_vmware ) )
+	d3d9?   ( || ( video_cards_intel video_cards_r300 video_cards_r600 video_cards_radeonsi video_cards_nouveau video_cards_vmware ) )
 	vulkan? ( video_cards_radeonsi? ( llvm ) )
 	vulkan-overlay? ( vulkan )
 	video_cards_radeon? ( x86? ( llvm ) amd64? ( llvm ) )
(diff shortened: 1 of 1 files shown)


zx2c4:
The egl/gbm removal us a surprise. I think I can guess why, but I didn't see any text in the commit about it. Is it that you just think it's superfluous and sufficiently small that it doesn't cost anything to enable them both?

For the Intel patch, you forgot to remove iris from the VIDEO_CARDS string after removing its logic from the ebuild.

mattst88:
> The egl/gbm removal is a surprise. I think I can guess why, but I didn't see any text in the commit about it. Is it that you just think it's superfluous and sufficiently small that it doesn't cost anything to enable them both?

Right, exactly. I should mark it as closing https://github.com/gentoo/gentoo/pull/21742 in fact. I'll add some to the commit message.

> For the Intel patch, you forgot to remove iris from the VIDEO_CARDS string after removing its logic from the ebuild.

I left it to control the 'prefer-iris' flag, which selects which driver (i965 vs iris) to choose for Gen8-11. I feel fine about setting it based on `VIDEO_CARDS=intel && USE=gallium` since it's the default; I feel less certain about how to handle that with the crocus driver, which is not yet the default.

Maybe I'll just leave `video_cards_crocus` in place and remove `video_cards_iris`.

mattst88:
> Maybe I'll just leave `video_cards_crocus` in place and remove `video_cards_iris`.

Doesn't look like I can do that, actually. The `prefer-crocus` / `prefer-iris` options actually modify the list of PCI IDs that the `i965` driver will bind to. :(

Short of removing the user's ability to select between iris/crocus vs i965, I think I need to leave both `iris` and `crocus` in `VIDEO_CARDS` for now...

thesamesam:
> > Maybe I'll just leave `video_cards_crocus` in place and remove `video_cards_iris`.
> 
> Doesn't look like I can do that, actually. The `prefer-crocus` / `prefer-iris` options actually modify the list of PCI IDs that the `i965` driver will bind to. :(
> 
> Short of removing the user's ability to select between iris/crocus vs i965, I think I need to leave both `iris` and `crocus` in `VIDEO_CARDS` for now...

It's unfortunate but not your fault, sounds ok.

thesamesam (approved):
Thanks a lot!

Need to make sure we update the [wiki](https://wiki.gentoo.org/wiki/Intel#Feature_support) accordingly when merged.

mattst88:
Well that problem's now solved, as the classic drivers were removed from the main branch.

FireBurn:
I'm not sure I like the consolidation, I'm glad I'm not an Intel user any more

mattst88:
> I'm not sure I like the consolidation, I'm glad I'm not an Intel user any more

I'm sorry to have disappointed you.

The purpose of this change is to reduce user confusion and support costs.

My measurements indicate that e.g. crocus adds 30 seconds to the build on a 4-core/8-thread CPU. Even on slower systems, I think this is acceptable.

PR #23008, opened 2021-11-20
media-gfx/freecad - remove obsolete patch, simplify opencascade logic

waebbl:
**media-gfx/freecad-9999: remove obsolete patch**
- fix some dependencies
- simplify opencascade related code, which has been obsoleted due to removal
	of slot 7.4.0
- use default CMAKE_INSTALL_PREFIX (/usr) instead of /usr/$(get_libdir)/${PN}.
	This leads to some files being installed in /usr/{Ext,Mod}. A new patch
	fixes this. The lower case symlinks for the executables are no longer
	installed now, as the binaries are in the path and can be called directly.
- use relative patchs for CMAKE_INSTALL_* variables

Closes: https://bugs.gentoo.org/824550

**media-gfx/freecad-0.19.2-r4: simplify opencascade logic**
With sci-libs/opencascade:7.4.0 gone from the tree, this patch simplifies
the code needed to configure various opencascade related variables to cmake.


media-gfx/freecad/freecad-0.19.2-r4.ebuild
```diff
@@ -79,7 +78,7 @@ RDEPEND="
 		dev-libs/openssl:=
 		net-misc/curl
 	)
-	fem? ( sci-libs/vtk:=[boost,python,qt5,rendering,${PYTHON_SINGLE_USEDEP}] )
+	fem? ( sci-libs/vtk:=[boost(+),python,qt5,rendering,${PYTHON_SINGLE_USEDEP}] )
```
thesamesam, line 81: Requires a revbump usually, but we still have the flag and I guess you're just preparing for possible removal a while in the future? (not with the next version?)

waebbl: I can do the revbump, of course. Yes this is in preparation of a possible removal of the flag from vtk-9.1.0+.
I can do this in a separate commit later on, to avoid a revbump for now.

juippis (approved):

PR #23012, opened 2021-11-20
app-shells/ksh: update to 1.0.0_beta1

stha09:


gyakovlev:
let me test is before merging, I'll supply a list of  ~ keywords that can be added.

gyakovlev:
also after giving it another thought maybe let's mask 2020 version for like 2-3 weeks, and drop later instead of now? I was a bit too cocky on IRC =)
that way users will get a mask reason message and not just portage-initiated downgrade.

gyakovlev:
hmmm
failing to build on ppc64
```
/var/tmp/portage/app-shells/ksh-1.0.0_beta1/work/ksh-1.0.0-beta.1/src/lib/libast/stdio/vfwscanf.c:103:54: error: 'OPEN_MAX' undeclared (first use in this function); did you mean 'FOPEN_MAX'?
  103 |                 if (t = sfnew(NiL, buf, sizeof(buf), OPEN_MAX+1, SF_READ))
      |                                                      ^~~~~~~~
      |                                                      FOPEN_MAX
```

looks like not all systems define `POSIX.1` value `OPEN_MAX`


sysdeps/unix/sysv/linux/powerpc/bits/local_lim.h
```C
/* Minimum guaranteed maximum values for system limits.  Linux/PPC version.
   Copyright (C) 1993-2018 Free Software Foundation, Inc.
   This file is part of the GNU C Library.

   The GNU C Library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public License as
   published by the Free Software Foundation; either version 2.1 of the
   License, or (at your option) any later version.

   The GNU C Library is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
   Lesser General Public License for more details.

   You should have received a copy of the GNU Lesser General Public
   License along with the GNU C Library; see the file COPYING.LIB.  If
   not, see <http://www.gnu.org/licenses/>.  */

/* The kernel header pollutes the namespace with the NR_OPEN symbol
   and defines LINK_MAX although filesystems have different maxima.  A
   similar thing is true for OPEN_MAX: the limit can be changed at
   runtime and therefore the macro must not be defined.  Remove this
   after including the header if necessary.  */
#ifndef NR_OPEN
# define __undef_NR_OPEN
#endif
#ifndef LINK_MAX
# define __undef_LINK_MAX
#endif
#ifndef OPEN_MAX
# define __undef_OPEN_MAX
#endif
#ifndef ARG_MAX
# define __undef_ARG_MAX
#endif

/* The kernel sources contain a file with all the needed information.  */
#include <linux/limits.h>

/* Have to remove NR_OPEN?  */
#ifdef __undef_NR_OPEN
# undef NR_OPEN
# undef __undef_NR_OPEN
#endif
/* Have to remove LINK_MAX?  */
#ifdef __undef_LINK_MAX
# undef LINK_MAX
# undef __undef_LINK_MAX
#endif
/* Have to remove OPEN_MAX?  */
#ifdef __undef_OPEN_MAX
# undef OPEN_MAX
# undef __undef_OPEN_MAX
#endif
/* Have to remove ARG_MAX?  */
#ifdef __undef_ARG_MAX
# undef ARG_MAX
# undef __undef_ARG_MAX
#endif

/* The number of data keys per process.  */
#define _POSIX_THREAD_KEYS_MAX	128
/* This is the value this implementation supports.  */
#define PTHREAD_KEYS_MAX	1024
#define PTHREAD_SIGNAL_SAFE_KEYS_MAX	4

/* Controlling the iterations of destructors for thread-specific data.  */
#define _POSIX_THREAD_DESTRUCTOR_ITERATIONS	4
/* Number of iterations this implementation does.  */
#define PTHREAD_DESTRUCTOR_ITERATIONS	_POSIX_THREAD_DESTRUCTOR_ITERATIONS

/* The number of threads per process.  */
#define _POSIX_THREAD_THREADS_MAX	64
/* We have no predefined limit on the number of threads.  */
#undef PTHREAD_THREADS_MAX

/* Maximum amount by which a process can descrease its asynchronous I/O
   priority level.  */
#define AIO_PRIO_DELTA_MAX	20

/* Minimum size for a thread.  At least two pages for systems with 64k
   pages.  */
#define PTHREAD_STACK_MIN	131072

/* Maximum number of timer expiration overruns.  */
#define DELAYTIMER_MAX	2147483647

/* Maximum tty name length.  */
#define TTY_NAME_MAX		32

/* Maximum login name length.  This is arbitrary.  */
#define LOGIN_NAME_MAX		256

/* Maximum host name length.  */
#define HOST_NAME_MAX		64

/* Maximum message queue priority level.  */
#define MQ_PRIO_MAX		32768

/* Maximum value the semaphore can have.  */
#define SEM_VALUE_MAX   (2147483647)
```

gyakovlev:
opened https://github.com/ksh93/ksh/issues/341

gyakovlev:
it will also likely need this.

`elibc_musl? ( sys-libs/fts-standalone )`

gyakovlev:
hey, upstream people figured out build failures,

apparently `-fdiagnostics-color=always` was triggering failures ( have no idea why yet)

I've filtered that out and merged PR, also removed the commit dropping old versions for now.
let's take slow approach, test this ksh more and then mask and remove 2020.x versions, hope that's ok with you.

gyakovlev:
and I confirmed it build on musl without fts-standalone

PR #23013, opened 2021-11-20
sys-fs/genfstab: bump to EAPI 8

scardracs:


mattst88:
I don't think there's much point in bumping the EAPI for its own sake.

PR #23015, opened 2021-11-21
gui-apps/tuigreet: added git-live 9999 version

DjLogozzo:
Changes:
* Added git-live tuigreet-9999.ebuild

Signed-off-by: Dee-Jay Logozzo <gentoo@djl.id.au>

### DIFF
--- a/gui-apps/tuigreet/tuigreet-9999.ebuild
+++ b/gui-apps/tuigreet/tuigreet-9999.ebuild
@@ -1,0 +1,37 @@
+# Copyright 2017-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+CRATES="
+"
+
+inherit cargo
+
+DESCRIPTION="TUI greeter for greetd login manager"
+HOMEPAGE="https://github.com/apognu/tuigreet"
+
+if [ ${PV} == "9999" ] ; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/apognu/${PN}"
+else
+	SRC_URI="https://github.com/apognu/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz
+	$(cargo_crate_uris ${CRATES})"
+	KEYWORDS="~amd64 ~ppc64 ~riscv"
+fi
+
+src_unpack() {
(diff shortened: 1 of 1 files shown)


gui-apps/tuigreet/tuigreet-9999.ebuild
```diff
@@ -0,0 +1,37 @@
...
+"
+
+inherit cargo
+
+DESCRIPTION="TUI greeter for greetd login manager"
+HOMEPAGE="https://github.com/apognu/tuigreet"
+
+if [ ${PV} == "9999" ] ; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/apognu/${PN}"
```
juippis, line 16: In future, 
please.

gui-apps/tuigreet/tuigreet-9999.ebuild
```diff
@@ -0,0 +1,37 @@
...
+else
+	SRC_URI="https://github.com/apognu/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz
```
juippis, line 18: Same here for the PN part.

juippis (approved):
I am somewhat reluctant in adding live ebuilds for unmaintained ebuilds... but it's nice to see Gentoo can be used for upstream development, so I support it here. Thanks!

PS I've strip out your sign-off even previously because it doesn't seem like it's your real name. Please if you don't use your real name, don't add sign-off at all. It's better for us that way. 
If that is your real name, I apologize for my actions - just playing it safe.

DjLogozzo:
Hey @juippis
Thanks for approving my PR, I'm glad to be able to contribute back to the community.

> PS I've strip out your sign-off even previously because it doesn't seem like it's your real name. Please if you don't use your real name, don't add sign-off at all. It's better for us that way. 
> If that is your real name, I apologize for my actions - just playing it safe.
 
Dee-Jay is my real name though I completely understand your hesitancy in including it, it is a fairly rare name and quite common as a nickname, if you'd like I can provide proof that it is my legal name or if you have any suggestions on what else I can do to help the situation

juippis:
I trust you. Thank you very much for replying!

PR #23020, opened 2021-11-21
add missing ~arm64 keyword for sys-cluster/lmod

boegel:
We have been using Lmod for a while on `aarch64` systems, works fine there as it does on `ppc64le` (which is keyworded already).

cc @amadio


amadio:
@boegel Could you please confirm that `FEATURES=test emerge -1 sys-cluster/lmod` passes the tests?

boegel:
@amadio Sorry for not following up on this, thanks to you and @bedroge for pushing this through!

PR #23024, opened 2021-11-21
media-gfx/viewnior: Version bump 1.8.

tastytea:


juippis (approved):
```
Configuring config.h using configuration
WARNING: Gettext not found, all translation targets will be ignored.

data/meson.build:24:0: ERROR: Can not assign None to variable.

A full log can be found at /var/tmp/portage/media-gfx/viewnior-1.8/work/viewnior-1.8-build/meson-logs/meson-log.txt
 * ERROR: media-gfx/viewnior-1.8::gentoo failed (configure phase):
 *   (no error message)
 * 
```
So there seems to be a configure flag available for nls. I'll add the dep for now, but would appreciate if you investigated the possibility for making it a toggleable option.

tastytea:
Thanks! There is no configure flag for gettext.

PR #23028, opened 2021-11-22
app-crypt/hashcat: update to 6.2.5

Miezhiko:
Also drop 6.2.4 and merged patches


thesamesam:
Please cleanup in a separate commit but I'd also prefer to not get rid of the last version immediately. Thanks for the PR.

Miezhiko:
@thesamesam done,
6.2.5 is minor release, mostly with bug fixes, I personally think it's good to drop merged patches and update.

thesamesam:
> @thesamesam done, 6.2.5 is minor release, mostly with bug fixes, I personally think it's good to drop merged patches and update.

I understand that, but if we remove n-1 versions _immediately_, if there are regressions, people won't be able to easily get back to a version which works. We don't really gain anything at all from removing it instantly, it just makes it harder for people to downgrade if necessary.

Also, didn't they change some of the rocm requirements for this? (not checked yet, but saw the commits before)

Miezhiko:
@thesamesam alright, I've dropped clean up commit.

ZeroChaos-:
ACK

PR #23035, opened 2021-11-22
games-engines/scummvm: Bump to version 2.5.0

alanswanson:
Synced from the live ebuild adding a "gif" USE flag and updating the SRC_URI to the new downloads.scummvm.org address.


mgorny:
Thanks for the PR. I've seen the new release get tagged but the download wasn't available yet at the time and I've forgotten about it afterwards.

mgorny:
Good work. I've readded the `~riscv` keyword that was lost from 2.2.0, and copied your changes to 9999.

PR #23036, opened 2021-11-22
net-dns/unbound: Add libunbound-only USE flag to allow building library only

alanswanson:
Applications (such as getdns/stubby, opendkim, dane, asterisk and libreswan) using libunbound do not need the unbound resolver and dependencies.

For the maintainers consideration.

Closes: https://bugs.gentoo.org/803497


mgorny:
Why not `IUSE=+server`?

alanswanson:
> Why not `IUSE=+server`?

Most server USE flag usage seems to correspond to (user preferred) client only builds. Unbound doesn't have a client so server versus libunbound-only build usage might be 50:50 but any use flag the maintainer prefers would be fine.

alanswanson:
Superseded when net-dns/getdns allowed building _USE_ minimal with no libunbound dependency for stubby use only (and also mail-filter/opendkim added a _USE_ unbound flag).

PR #23038, opened 2021-11-22
riscv: keyword gnome-extra/cinnamon and mate-base/mate

liangyongxiang:


ConiKost (approved):
Ack for my packages dev-python/python-axolotl{,-curve25519}.

PR #23041, opened 2021-11-22
dev-scheme/racket: misc fixes

xgqt:


xgqt:
AD `has_version` (https://github.com/gentoo/gentoo/pull/20668#discussion_r686720997) will not work because the bytecode version counts, if we upgrade from 8.3_pre20211121 to  8.3_pre20211122 we need to copy the `PKGDB`.

xgqt:
CC @thesamesam @hattya

dev-scheme/racket/racket-8.2-r1.ebuild
```diff
@@ -19,12 +19,14 @@ LICENSE="
...
+# Bytecode generated by Racket is not compatible between versions.
+# In this case PV == bytecode version,
+# but for _patch, _pre, etc. it has to be set manually.
 SLOT="0/${PV}"
 KEYWORDS="amd64 ~arm ~ppc ~ppc64 x86"
-IUSE="X +chez +doc +futures +jit minimal +places +readline +threads"
+IUSE="X +chez +doc +futures +jit minimal +places +threads"
 # See bug #809785 re chez/threads
-REQUIRED_USE="futures? ( jit ) chez? ( threads )"
+REQUIRED_USE="futures? ( jit threads ) chez? ( threads ) places? ( threads )"
```
thesamesam, line 29: Link to any relevant upstream commits or bugs in the bug / here, if there are any.

dev-scheme/racket/racket-8.2-r1.ebuild
```diff
@@ -19,12 +19,14 @@ LICENSE="
 	chez? ( Apache-2.0 )
 	!chez? ( LGPL-3 )
 "
-# Bytecode generated by Racket is not compatible between versions
+# Bytecode generated by Racket is not compatible between versions.
+# In this case PV == bytecode version,
+# but for _patch, _pre, etc. it has to be set manually.
```
thesamesam, line 24: What does this mean? Why not just use ${PF} or similar?

dev-scheme/racket/racket-8.2-r1.ebuild
```diff
@@ -39,7 +41,6 @@ DEPEND="
 		x11-libs/pango[X]
 		x11-misc/shared-mime-info
 	)
-	readline? ( dev-libs/libedit )
```
thesamesam, line 42: Heh, yeah, that's.. confusing.

dev-scheme/racket/racket-8.2-r1.ebuild
xgqt: What I meant to say is that here `PV == SLOT`.
The SLOT should point to the bytecode version.
So in this case `PV == bytecode version` is also true.

dev-scheme/racket/racket-8.2-r1.ebuild
xgqt: afair it fails on configure if you do `USE="-threads places futures"`

xgqt: image

thesamesam: Ah, I was just saying that I thought you'd mentioned some work upstream, so we should mention it in the Gentoo bug if you have any links.

dev-scheme/racket/racket-8.2-r1.ebuild
thesamesam: Oh, I think I understand you now. Maybe clarify the comment a bit if you can?

thesamesam:
lgtm modulo the tiny nits I mentioned. Thanks!

dev-scheme/racket/racket-8.2-r1.ebuild
xgqt: the one we talked before was about `chez` + `-threads`, threads for futures/places are probably to stay

xgqt:
> lgtm modulo the tiny nits I mentioned. Thanks!
> 
> Not looked too closely at the replacing_version change.

thats the one change I most wanted to be verified
(no rush)

thesamesam:
heh, ok, remind me tomorrow :)

dev-scheme/racket/racket-8.2-r1.ebuild
```diff
@@ -135,30 +137,27 @@ src_install() {
...
+					die "failed to create a copy of ${rktd}"
+			fi
+		done
+	fi
 }
 
 pkg_postinst() {
 	post_X_update
 
+	optfeature "readline editing features in REPL" dev-libs/libedit
```
thesamesam, line 160: It really dlopens libedit or something? Doesn't link against it?

dev-scheme/racket/racket-8.2-r1.ebuild
```diff
@@ -135,30 +137,27 @@ src_install() {
...
-						die "failed to create a backup of ${rktd}"
-					cp "${EROOT}"/${rktd} "${ED}"/${rktd} ||
-						die "failed to create a copy of ${rktd}"
-				fi
-			done
-		fi
-	done
+	# If we are merging the same SLOT check if package
+	# database files exist and do not overwrite them
+	if has_version "${CATEGORY}/${PN}:${SLOT}"; then
```
thesamesam, line 142: I kind of wonder if this should be CONFIG_PROTECTed but I suppose users won't be touching it.

thesamesam (approved):
LGTM; just verify that the logic in pkg_*inst works as it should?

dev-scheme/racket/racket-8.2-r1.ebuild
xgqt: looking into `strace` - racket looks for libreadline in many palces

`openat(AT_FDCWD, "/lib64/libreadline.so.8", O_RDONLY|O_CLOEXEC) = 3`

no, racket does not link against them, it uses FFI to access them

dev-scheme/racket/racket-8.2-r1.ebuild
xgqt: they are not very friendly to edit, so I dont see a reason for CONFIG_PROTECT here

dev-scheme/racket/racket-8.2-r1.ebuild
thesamesam: wait, libedit or readline?

dev-scheme/racket/racket-8.2-r1.ebuild
thesamesam: ack

xgqt: yea, thats the strange think it looks for readline on my system

xgqt: https://github.com/racket/racket/blob/master/pkgs/racket-doc/scribblings/more/more.scrbl#L106

So it can used both, but I do not know by it does not look fro editline on my system (even when I unmerged readline)

thesamesam: Weird.

PR #23044, opened 2021-11-23
sys-apps/fwupd: 1.7.1-r1, 1.7.2

ixti:
* Add 1.7.1-r1
  * Add logitech HID++ plugin when `logitech` USE-flag is enabled
  * Apply `elogind` patch which is needed to enable `logind` plugin compilation (ensures that machine does not enter a low power mode when updates are being performed); Upstream has that fix as of 1.7.2 (https://github.com/fwupd/fwupd/pull/3944)
* Add 1.7.2 ebuild


Whissi:
I hope you don't mind that I am going to reject this PR: The changes will be in our 1.7.2 bump but I'll use my own bump since your 1.7.2 bump is missing almost every other 1.7.2 change and it is easier for me to do this on my own.

Thanks for your contribution and upstreaming the logind patch directly!

Whissi:
PS: There is no logitech hidpp plugin option like you added to ebuild. The plugin will get auto-enabled when gusb is enabled.

ixti:
> I hope you don't mind that I am going to reject this PR: The changes will be in our 1.7.2 bump but I'll use my own bump since your 1.7.2 bump is missing almost every other 1.7.2 change and it is easier for me to do this on my own.

Totally cool with that! :D
Thank you for a review and correct ebuild bump.

> Thanks for your contribution and upstreaming the logind patch directly!

My pleasure! :D Thank you for maintaining the ebuild.

> PS: There is no logitech hidpp plugin option like you added to ebuild. The plugin will get auto-enabled when gusb is enabled.

TIL. I thought `plugins/*` are the options. Now I know that I should have looked to `meson.options` instead :D

PR #23050, opened 2021-11-23
dev-lang/rakudo: update to 2021.10 version

Miezhiko:


dev-lang/rakudo/rakudo-2021.10.ebuild
```diff
@@ -0,0 +1,72 @@
...
+
+src_compile() {
+	emake NQP_JARS="${NQP}" BLD_NQP_JARS="${NQP}"
+}
+
+src_install() {
+	emake DESTDIR="${D}" NQP_JARS="${NQP}" BLD_NQP_JARS="${NQP}" install
+}
+
+src_test() {
```
thesamesam, line 70: After src_compile and before src_install to match the order in which it is executed.

dev-lang/moarvm/moarvm-2021.10.ebuild
```diff
@@ -0,0 +1,60 @@
...
+LICENSE="Artistic-2"
+SLOT="0"
+
+IUSE="asan clang debug doc +jit optimize static-libs ubsan"
+
+RDEPEND="dev-libs/libatomic_ops
+		>=dev-libs/libuv-1.26:=
+		dev-libs/libffi:="
+DEPEND="${RDEPEND}
+	clang? ( >=sys-devel/clang-3.1 )
```
thesamesam, line 30: Clang should be in BDEPEND unless it links against libclang (does it?)

dev-lang/moarvm/moarvm-2021.10.ebuild
```diff
@@ -0,0 +1,60 @@
...
+	dev-lang/perl"
```
thesamesam, line 31: Perl certainly seems like at least a BDEPEND as it calls it in the ebuild.

dev-lang/moarvm/moarvm-2021.10.ebuild
```diff
@@ -0,0 +1,60 @@
...
+	einfo "--prefix '${prefix}'"
+	einfo "--libdir '${libdir}'"
+	local myconfigargs=(
+		"--prefix" "${prefix}"
+		"--has-libuv"
+		"--has-libatomic_ops"
+		"--has-libffi"
+		"--libdir" "${libdir}"
+		"--compiler" "$(usex clang clang gcc)"
+		"$(usex asan        --asan)"
```
thesamesam, line 52: I'd drop the whitespace. The devmanual is clear that it's preferred to not do this.

dev-lang/nqp/nqp-2021.10.ebuild
```diff
@@ -0,0 +1,159 @@
...
+		dev-java/jline:0
+		dev-java/jna:4
+	)
+	moar? ( ~dev-lang/moarvm-${PV}[clang=] )"
+RDEPEND="${CDEPEND}
+	java? ( >=virtual/jre-1.8 )"
+DEPEND="${CDEPEND}"
+BDEPEND="${CDEPEND}
+	clang? ( sys-devel/clang )
+	java? ( >=virtual/jdk-1.8 )
```
thesamesam, line 36: I think we're supposed to use :* if it works with any Java version as far as we know.

(Also, note that nqp seems to get the BDEPEND part right.)

dev-lang/nqp/nqp-2021.10.ebuild
```diff
@@ -0,0 +1,159 @@
...
+REQUIRED_USE="|| ( java moar )"
+
+CDEPEND="java? (
+		dev-java/asm:4
+		dev-java/jline:0
+		dev-java/jna:4
+	)
+	moar? ( ~dev-lang/moarvm-${PV}[clang=] )"
+RDEPEND="${CDEPEND}
+	java? ( >=virtual/jre-1.8 )"
```
thesamesam, line 32: :* again, probably.

Miezhiko:
@thesamesam it seems like it was updated before review, we can maybe keep this open until the next update, I'm not sure

PR #23060, opened 2021-11-23
[DO NOT MERGE] sys-devel/gcc: add 11.2.1_p20211127

SoapGentoo:
upstream has indicated that 11.3 is unlikely to be
released soon, and 11.2 has accumulated enough bugs
to require an intermediate release for Gentoo.


thesamesam:
Let's tag:
- https://bugs.gentoo.org/811867 (s390 ICE)
- https://bugs.gentoo.org/823780 (glibc-2.34/AVX512 relocation)

thesamesam (approved):

sys-devel/gcc/gcc-11.2.1_p20211120.ebuild
```diff
@@ -0,0 +1,25 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="7"
+
+PATCH_GCC_VER="11.2.0"
```
SoapGentoo, line 6: - [ ] add back/rebase GCC patches

PR #21423, opened 2021-06-25
x11-libs/libva-intel-media-driver and media-libs/gmmlib version bump

Karlson2k:
Updated x11-libs/libva-intel-media-driver to version 21.2.2_pre
Updated media-libs/gmmlib to version 21.1.3

Updated patch for libva-intel-media-driver.

PR #23062, opened 2021-11-23
x11-libs/libva-intel-media-driver fixes and bump

mattst88:
* x11-libs/libva-intel-media-driver: Add USE=redistributable
 * x11-libs/libva-intel-media-driver: Remove IUSE=set-as-default
 * x11-libs/libva-intel-media-driver: Remove IUSE=custom-cflags
 * x11-libs/libva-intel-media-driver: Set CMAKE_BUILD_TYPE
 * x11-libs/libva-intel-media-driver: Version bump to 21.4.2


mattst88:
Cc: @Karlson2k

klondi (approved):
Everything here seems correct. Thanks for the good job Matt :)

klondi (commented):
You can change this:
```
local CMAKE_BUILD_TYPE="Release"
cmake_src_configure
```

Into this:
```
CMAKE_BUILD_TYPE="Release" cmake_src_configure
```

That limits the scope of the variable to that specific function call.

Karlson2k:
Thanks for the update, @mattst88 

I'd suggest some changes:
* `set-as-default` was added to close https://bugs.gentoo.org/737284. This is upstream mechanism to enforce driver selection, it makes sense to follow upstream here.
* `custom-cflags` still makes sense. First of all, this is the only configuration tested by upstream. If there is any problem with this driver, it makes sense to disable this flag to ensure that the problem is not related to user's `CFLAGS`.
* While `redistributable` flag is used by [sys-kernel/linux-firmware](https://packages.gentoo.org/packages/sys-kernel/linux-firmware) to enable features, it sounds very similar to [bindist](https://packages.gentoo.org/useflags/search?q=bindist) which is doing exactly the opposite and may confuse users. Maybe it's better to use [non-free](https://packages.gentoo.org/useflags/search?q=non-free) flag?

klondi:
Keep in mind that the kernels are redistributable (hence why bindist is not appropriate) and licensed under a free license (hence why non-free is not appropriate). The only issue is that they are provided in binary form (which you can modify and redistribute) instead of source code for, hence why I recommended redistributable since the licensing case is exactly the same as the one you will see for the firmware that linux-firmware prevents from installing when the flag is disabled.

Karlson2k:
Looks like binary kernels are not really "free".
They are enabled by `ENABLE_NONFREE_KERNELS` (see [README.md](https://github.com/intel/media-driver/blob/master/README.md)). Other GNU/Linux distributions call them ["non-free"](https://packages.debian.org/sid/intel-media-va-driver-non-free) as well.
I think if Intel names them so, it makes sense to do the same.

Karlson2k:
One more thing. This version hasn't been [released yet](https://github.com/intel/media-driver/tags). Currently it is just a tag in repo (so it is just a prerelease).
The @thesamesam [asked to not use any keywords for prerelease versions](https://github.com/gentoo/gentoo/pull/21423#discussion_r660274537). I agree with it as Intel doesn't mark this tags are a "release" for a propose. Some tags get "release" marks later (probably after a proper testing by upstream), some are not. I believe we can trust original author on this.

klondi:
No, they are not really free. Nor is most of the binary firmware provided by the linux-firmware package. It is not free because although you are free to transform and modify the kernels (or the firmware) you are not provided the source code to them.

For Ubuntu an Debian this is just called "Non free" because it is against the DFSG requirements (see https://www.debian.org/social_contract ). In particular against point number 2 regarding source code.

Gentoo is a lot more nuanced in this area (which is why we have the LICENSE variable to start with). Which is why simply calling the code "Non free".

bindist is used when the resulting binaries cannot be freely distributed (usually due to patents or incompatible licenses being mixed). That is clearly not the case here, the license explicitly allows redistribution.

non-free is a bit more complex and the packages using it all have different restrictions that do not match this one as well:
* media-fonts/intlfonts installs a font with non commercial usage requirements
* games-fps/gzdoom installs wad files with an array of fully proprietary licenses (check for example the Activision one which on the first paragraphs already forbids redistribution and commercial usage).
* dev-embedded/sdcc installs auxiliary files that require you to run the resulting binaries only on Microchip original microcontrollers.
* games-misc/ponysay installs auxiliary files which have licenses restricting commercial usage

These cases are way too far from the restrictions that not providing the source code implies. Intel is allowing you to use and distribute the binary kernels as you wish (as is the case with most of the linux-firmware). They are just not providing the source code to ease that task for you (nor the toolchain that generated this binaries). Otherwise you can modify the binaries, use them for commercial purposes, use them on any chips able to handle them and distribute them as much as you want. Obviously you can also make modifications and redistribute them if you so wish. There are also no patent licensing issues affecting them that we are aware of. The lack of source code would still make the resulting software non free, if we follow the FSF or Debian's guidelines, but it falls in line with what happens with the linux-firmwares instead of what happens with the other packages you mention.

Karlson2k:
@klondi Thanks for the detailed explanations. It was initially obvious that `bindist` is not applicable here.
I have been finally able to find licenses for "non-free" kernels. It is clear MIT license.
If we use term "free" not as FSF defines, then probably `redistributable` is OK for the name of this flag. Than I suggest to remove words "non-free" form `metadata.xml` as we are considering these files as a "free".
Something like:
`Install redistributable media kernel binaries (no source code) files (primarily needed for video encoding support)`

mattst88:
> Thanks for the update, @mattst88
> 
> I'd suggest some changes:
> 
>     * `set-as-default` was added to close https://bugs.gentoo.org/737284. This is upstream mechanism to enforce driver selection, it makes sense to follow upstream here and leave this flag.

Right. I don't think it's at all needed anymore. It seems like `libva-intel-media-driver` just didn't have PCI-IDs or something for Comet Lake. This shouldn't really be a common problem...

>     * `custom-cflags` still could be useful. First of all, this is the only configuration tested by the upstream. If there is any problem with this driver, it makes sense to disable this flag first to ensure that the problem is not related to user's `CFLAGS`. Let's not remove this flag.

I would agree if there were things like `-fno-strict-aliasing` that affected the semantics of the code, but actually just contains:

 * `-m32`/`-m64` - totally unnecessary
 * `-g` - should never be unconditionally set by the build system
 * `-O2` - don't want this set by the build system
 * `-D_FORTIFY_SOURCE=2` - default in Gentoo
 * `-fno-omit-frame-pointer` - probably set so profiling is possible; don't want this set unconditionally by the build system

I really don't think we're doing anything insane by removing these. The upstream project is very much not a community-oriented free software project, but instead a throw-it-over-the-wall corporate project with CFLAGS for their own needs polluting the build system.

> One more thing. This version hasn't been [released yet](https://github.com/intel/media-driver/tags). Currently it is just a tag in repo (so it is just a prerelease). The @thesamesam [asked to not use any keywords for prerelease versions](https://github.com/gentoo/gentoo/pull/21423#discussion_r660274537). I agree with it as Intel doesn't mark such tags as a "release" for a propose. Some tags get "release" marks later (probably after a proper testing by upstream), some are not. I believe we can trust original author on this.

I asked about this in https://github.com/intel/media-driver/issues/1290; They indicated that the tag itself is only made after internal qualification testing.

>  If we use term "free" not as FSF defines, then probably redistributable is OK for the name of this flag. Than I suggest to remove words "non-free" form metadata.xml as we are considering these files as a "free".
> Something like:
> `Install redistributable media kernel binaries (no source code) files (primarily needed for video encoding support)`

Sounds good.

thesamesam (approved):

Karlson2k:
> > I'd suggest some changes:
> > ```
> > * `set-as-default` was added to close https://bugs.gentoo.org/737284. This is upstream mechanism to enforce driver selection, it makes sense to follow upstream here and leave this flag.
> > ```
> 
> Right. I don't think it's at all needed anymore. It seems like `libva-intel-media-driver` just didn't have PCI-IDs or something for Comet Lake. This shouldn't really be a common problem...


Actually this is not the only purpose of this flag. If user has several drivers, like [`intel-vaapi-driver`](https://github.com/intel/intel-vaapi-driver) and this one, environment variable may override automatic selection of the driver as both drivers can be used for Gen8, Gen9, Gen9.5 (and never-released Gen10).

And looks like it is useful for users, as it was requested. https://bugs.gentoo.org/737284

> > ```
> > * `custom-cflags` still could be useful. First of all, this is the only configuration tested by the upstream. If there is any problem with this driver, it makes sense to disable this flag first to ensure that the problem is not related to user's `CFLAGS`. Let's not remove this flag.
> > ```
> 
> I would agree if there were things like `-fno-strict-aliasing` that affected the semantics of the code, but actually just contains:
> 
> * `-m32`/`-m64` - totally unnecessary
> * `-g` - should never be unconditionally set by the build system
> * `-O2` - don't want this set by the build system
> * `-D_FORTIFY_SOURCE=2` - default in Gentoo
> * `-fno-omit-frame-pointer` - probably set so profiling is possible; don't want this set unconditionally by the build system
> 
> I really don't think we're doing anything insane by removing these. The upstream project is very much not a community-oriented free software project, but instead a throw-it-over-the-wall corporate project with CFLAGS for their own needs polluting the build system.
> 

Thanks for proper analysis, @mattst88!

I waive my suggestion.
When I created the patch, this package was full of architecture-dependent flags, see https://github.com/intel/media-driver/blob/intel-media-20.4/media_driver/cmake/linux/media_compile_flags_linux.cmake#L41, but currently it is quite limited set of such flags. Remove this USE flag.

> > One more thing. This version hasn't been [released yet](https://github.com/intel/media-driver/tags). Currently it is just a tag in repo (so it is just a prerelease). The @thesamesam [asked to not use any keywords for prerelease versions](https://github.com/gentoo/gentoo/pull/21423#discussion_r660274537). I agree with it as Intel doesn't mark such tags as a "release" for a propose. Some tags get "release" marks later (probably after a proper testing by upstream), some are not. I believe we can trust original author on this.
> 
> I asked about this in [intel/media-driver#1290](https://github.com/intel/media-driver/issues/1290); They indicated that the tag itself is only made after internal qualification testing.

According to their answer these versions are fine (at least for unstable).
I've created [additional question](https://github.com/intel/media-driver/issues/1297) about this topic.  

Thanks for updating this PR and for proper analysis.

mattst88:
> Actually this is not the only purpose of this flag. If user has several drivers, like intel-vaapi-driver and this one, environment variable may override automatic selection of the driver as both drivers can be used for Gen8, Gen9, Gen9.5 (and never-released Gen10).
> 
> And looks like it is useful for users, as it was requested. https://bugs.gentoo.org/737284

Right, I saw this. It looks to me like the driver selection in libva is done by matching the KMS driver (e.g. `i915`) for DRM and Wayland, and by matching the Mesa 3D driver (e.g. `iris` or `i965`) for X11. 

I really suspect the problem this user had in bug #737284 was simply caused by libva not supporting multiple driver backends. The bug was filed in August 2020. I see this in git log

````
commit ca6f3bec5bc1ee98252bf2f8286db3856d1bb91c
Author:     Agostino Sarubbo <ago@gentoo.org>
AuthorDate: Thu Sep 24 08:43:49 2020 +0200
Commit:     Agostino Sarubbo <ago@gentoo.org>
CommitDate: Thu Sep 24 08:46:30 2020 +0200

    x11-libs/libva: amd64 stable wrt bug #728870
    
    RepoMan-Options: --include-arches="amd64"
    Signed-off-by: Agostino Sarubbo <ago@gentoo.org>

 x11-libs/libva/libva-2.7.1.ebuild | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
[...]
commit 4fc5208d96ccef987695110c351bc5d25cf7ed49
Author:     Agostino Sarubbo <ago@gentoo.org>
AuthorDate: Tue Aug 13 12:52:49 2019 +0200
Commit:     Agostino Sarubbo <ago@gentoo.org>
CommitDate: Tue Aug 13 12:52:49 2019 +0200

    x11-libs/libva: amd64 stable wrt bug #684108
    
    Package-Manager: Portage-2.3.69, Repoman-2.3.16
    RepoMan-Options: --include-arches="amd64"
    Signed-off-by: Agostino Sarubbo <ago@gentoo.org>

 x11-libs/libva/libva-2.5.0-r1.ebuild | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
````

So the user was likely using v2.5.0-r1. The commit that added multiple driver backend support to libva was https://github.com/intel/libva/commit/b4d0af7f43f0ef2d51f070c75429c0d6acca663b in v2.6.0 (see https://github.com/intel/libva/pull/351).

Support for loading the `iHD` driver on X11 was only added in https://github.com/intel/libva/commit/bc8a12af319feca180aa99301b7e930f5890c1dd in v2.8.0.

So I'm about 99% sure this is why the user had the problem. The environment variable was a hack to workaround these problems, but is no longer needed as far as I can tell.

mattst88:
Heh, also: sorry for the slow response. Both Gentoo Bugzilla and GitHub were down at the same time! :P

Karlson2k:
@mattst88 You did a huge job on source code analysis, this is cool. :)

However, I don't see why this USE flag must be removed.
The arguments to leave it are:
1. It is supported and used by the upstream.
2. It can be easily used for quick testing if something go wrong (now or in future). You never know what may change and where.
3. It is disabled by default and is a simple single flag, it does not pollute USE flags namespace.

mattst88:
> @mattst88 You did a huge job on source code analysis, this is cool. :)
> 
> However, I don't see why this USE flag must be removed. The arguments to leave it are:
> 
>     1. It is supported and used by the upstream.

I think it was documented upstream for the same reason I mentioned earlier: libva didn't support loading the iHD driver.

>     2. It can be easily used for quick testing if something go wrong (now or in future). You never know what may change and where.

For quick testing, you'd just create the file and run `env-update`, not rebuild a rather large package like `x11-libs/libva-intel-media-driver` just to add one small text file.

>     3. It is disabled by default and is a simple single flag, it does not pollute USE flags namespace.

It doesn't, but in my experience any time you give users something to harm themselves with, they will. This is inevitable but acceptable if there's an upside to the flexibility the USE flag offers, but there's essentially zero value in this situation.

thesamesam:
Yeah, I'd rather just restore it if it turns out to be needed. People can downgrade to stable if needed if there's an issue we haven't realised yet too. But I don't see why it should be (especially given it's essentially a configuration matter).

For me, it comes down to this: I think having it there suggests that it's a reasonable thing to do/use/toggle and I don't think that's the case.

Karlson2k:
Well, looks like I'm in the minority. :)

OK then, let's merge it.

mattst88:
Thanks!

Karlson2k:
@mattst88 Thank you.

PR #23064, opened 2021-11-24
app-emulation/xen: add upstream security patches

hydrapolic:
- Boot tested 4.14 on amd64.
- Compile tested 4.15 on ~amd64.


hydrapolic:
Thank you @Flowdalic

PR #23067, opened 2021-11-24
sci-electronics/kicad: fix cmake flags for occ

yehoshuapw:
Signed-off-by: Yehoshua Pesach Wallach <yehoshuapw@gmail.com>

No bug _yet_, but kicad failed to compile with use flag occ - since things that didn't need to ended up as part of the cmake flags.


yehoshuapw:
(Sorry about that, I missed the bug first when I encountered the problem myself)

thesamesam:
Thank you!

PR #23069, opened 2021-11-24
dev-libs/libtpms: Bump to 0.9.1, remove old 

salahcoronya:
New version fixes a bug that may cause timeouts in VM with qemu.

Remove old versions, as downgrading is problematic : https://github.com/stefanberger/libtpms/issues/284 . The sooner everyone is on 0.9, the better.


ionenwks (approved):
Sounds fair to me, thanks.

PR #23073, opened 2021-11-25
media-libs/netpbm: add v10.86.26

ceamac:
Add v10.86.26
Drop v10.76.00 because stabilization of v10.86.24 made it obsolete
Still keeping v10.70.00 because it's stable for ppc64


ionenwks (approved):
Thanks.

PR #23074, opened 2021-11-25
sci-mathematics/gimps: Version bump to 30.7 build 9

PPed72:
Closes: https://bugs.gentoo.org/827272


juippis (approved):
Would appreciate if you split adding and removing ebuilds to their own commits, so this should be 2 commits. But guess this won't cause harm. Thanks!

PR #23078, opened 2021-11-25
www-client/firefox: 91esr x86 build fix && 94 gcc+lto fix

juippis:


orlitzky:
I backported the lto fix to 91.3.0 and it does work, thanks.

juippis:
These should be included in the latest firefoxes.

PR #23079, opened 2021-11-25
net-wireless/aircrack-ng: move deps from PDEPEND to RDEPEND

floppym:
PDEPEND is meant for breaking circular dependencies. I don't see any
circular dependencies here.


ZeroChaos-:
PDEPEND is meant for things not needed until after merge.  NACK unless there is a rule only allowing it's use for "breaking circular deps"

ZeroChaos-:
It looks like the docs have been updated since I last read them.  While it's dumb to force portage to satisfy things before they are needed, ACK for following policy I guess.

PR #23091, opened 2021-11-26
kernel-2.eclass: Respect portage CC variable

10ne1:
kernel-2.eclass: Respect portage CC & AR variables

Starting with kernel>=v5.7 the build system can override the
tools vars by setting LLVM=1 [1], but older kernels still use
the default GNU tools, so to be able to use a full LLVM/Clang
build, CC should be set together with AR to the portage set
values.

Doing this avoids situations like building the kernel with
clang (using the set HOSTCC) but using gcc/gnu-ar for headers.

[1] a0d1c951ef08 kbuild: support LLVM=1 to switch the default tools to Clang/LLVM

Co-authored-by: Manoj Gupta <manojgupta@chromium.org>

### DIFF
--- a/eclass/kernel-2.eclass
+++ b/eclass/kernel-2.eclass
@@ -692,7 +692,7 @@ env_setup_xmakeopts() {
 	elif type -p ${CHOST}-ar >/dev/null; then
 		xmakeopts="${xmakeopts} CROSS_COMPILE=${CHOST}-"
 	fi
-	xmakeopts="${xmakeopts} HOSTCC=$(tc-getBUILD_CC)"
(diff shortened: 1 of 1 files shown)


mpagano:
I asked for comments from other developers and they had concerns with this change and potential unintended impacts.   Can this be posted to the mailing list (gentoo-dev) to get more input, please?

10ne1:
@mpagano Ok I'll also send to the mailing list, thanks!

10ne1:
@mpagano I sent [the patch to the ML](https://archives.gentoo.org/gentoo-dev/message/bf8121597a625fc4768f02c7d6ec3e7b) 2 weeks ago and nobody responded. Should I CC/send to someone explicitely or should I wait a bit more considering the holiday season is approaching?

PR #23095, opened 2021-11-27
app-emulation/virtualbox-modules: add basic clang+thinlto kernel support

gcarq:
To use this there's no need to do anything beside having a kernel built with clang+(thin)lto (normally through LLVM=1 LLVM_IAS=1).

Closes: https://bugs.gentoo.org/820536
Signed-off-by: Michael Egger <egger.m@protonmail.com>


telans:
Can confirm this works for me with no issues in VM management

PR #23099, opened 2021-11-28
dev-libs/gjs: fix compatibiliy with meson 0.60.2

Miezhiko:
Added patch from upstream to gjs 1.70.0, tested on ~amd64

related bug: https://bugs.gentoo.org/827538


thesamesam:
Fixed by https://gitweb.gentoo.org/repo/gentoo.git/commit/dev-libs/gjs?id=b707dc5311d1c6a1eefb9f6bf28cf4d217136192.

PR #23104, opened 2021-11-29
sys-devel/dwz: Version update to 0.14

annulen:
- sys-devel/dwz: Version update to 0.14
- sys-devel/dwz: Remove old version 0.13


sys-devel/dwz/dwz-0.14.ebuild
```diff
@@ -0,0 +1,31 @@
+# Copyright 1999-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
juippis: 8!

sys-devel/dwz/dwz-0.14.ebuild
```diff
@@ -0,0 +1,31 @@
+# Copyright 1999-2020 Gentoo Authors
```
juippis: You should use our automated tools to commit so it fixes the copyright header automatically. E.g. repoman or pkgdev.

juippis (approved):
Tests fail for me, 
```
child process exited abnormally
FAIL: /var/tmp/portage/sys-devel/dwz-0.14/work/dwz/testsuite/dwz.tests/odr-struct-multifile.sh

		=== dwz Summary ===

# of expected passes		56
# of unexpected failures	1
# of unsupported tests		4
make: *** [Makefile:145: check] Error 1
rm invalid-dw-at-stmt-list-encoding-dw.S no-multifile-prop-dw.S unavailable-dwarf-piece-dw.S
 * ERROR: sys-devel/dwz-0.14::gentoo failed (test phase):
 *   Make check failed. See above for details.
```
But you'll get a new bug report about that.

PR #23106, opened 2021-11-29
sci-libs/blis: bump version to 0.8.1

6-6-6:
Closes: https://bugs.gentoo.org/764098
Signed-off-by: Zhang Zongyu <zongyu@novazy.net>


sci-libs/blis/blis-0.8.1.ebuild
```diff
@@ -0,0 +1,114 @@
...
+
+RDEPEND="
+	eselect-ldso? (
+		!app-eselect/eselect-cblas
+		>=app-eselect/eselect-blas-0.2
+	)"
+
+BDEPEND="${PYTHON_DEPS}"
+DEPEND="${RDEPEND}
+	${PYTHON_DEPS}
```
heroxbd, line 29: Does blis have a Python linking or header dependency?

I think it only goes to `BDEPEND`.

heroxbd (changes requested):

sci-libs/blis/blis-0.8.1.ebuild
6-6-6: > Does blis have a Python linking or header dependency?

No. Blis uses Python for [flattening its headers](https://github.com/flame/blis/blob/b727645eb7a8df39dee74068f734da66322fe0b3/common.mk#L502).

> I think it only goes to `BDEPEND`.

I agree with it.

PR #23107, opened 2021-11-29
dev-java/jnr-ffi: restrict tests

vaukai:


dev-java/jnr-ffi/jnr-ffi-2.0.2-r1.ebuild
```diff
@@ -15,6 +15,8 @@ LICENSE="|| ( Apache-2.0 LGPL-3 )"
 SLOT="2"
 KEYWORDS="~ppc64"
 
+RESTRICT="test"
```
fordfrog, line 18: it would be nice to add an explanation to the ebuild why the tests are restricted so that anybody who wants to touch the ebuild knows without a need to investigate

PR #23109, opened 2021-11-29
media-video/mjpg-streamer: dropping problematic use flag

ChaosEngine:
media-video/mjpg-streamer: removing http-management USE because upstream is not accepting PR and there is a dependency build conflict.
Closes: https://bugs.gentoo.org/821253
Signed-off-by: Andrzej Pauli andrzej.pauli@gmail.com


thesamesam:
1. Could you fix the commit message? It doesn't have the 'cat/pkg: ...' prefix
2. Could you make it a bit more descriptive to say what you're doing & why?
3. This seems to add a new version which isn't obvious from the commit / PR title.

ChaosEngine:
> 1. Could you fix the commit message? It doesn't have the 'cat/pkg: ...' prefix
> 2. Could you make it a bit more descriptive to say what you're doing & why?
> 3. This seems to add a new version which isn't obvious from the commit / PR title.

Better?

juippis (commented):
I think the easiest solution is just to package.use.mask the problematic flag. We don't need to force rebuilds on users who haven't enabled that flag.

Another question: If the upstream is dead, does the program work overall? When's the time to get rid of it...?

ChaosEngine:
> I think the easiest solution is just to package.use.mask the problematic flag. We don't need to force rebuilds on users who haven't enabled that flag.

Bug was generated by automatic run on tinderbox (==build bot) so it had to be cleaned and fixed for every "USE-case"

> 
> Another question: If the upstream is dead, does the program work overall? When's the time to get rid of it...?

Yes, the program works pretty much alright. The upstream seldom accepts new PR and author is just rarely active in this area.

PR #23115, opened 2021-11-29
app-editors/vim: remove warning due to missing defaults.vim

dongsupark:
vim >= 8.2.3428 prints out a warning on missing file, when it is built with USE="minimal" and `app-editors/vim-core` is not installed.

```
$ vim
E1187: Failed to source defaults.vim
Press ENTER or type command to continue
```

To remove the warning, explicitly install /usr/share/vim/vim82/defaults.vim.

Closes: https://bugs.gentoo.org/820356
Signed-off-by: Dongsu Park <dpark@linux.microsoft.com>

### DIFF
--- a/app-editors/vim-core/vim-core-8.2.0814.ebuild
+++ b/app-editors/vim-core/vim-core-8.2.0814.ebuild
@@ -191,6 +191,9 @@ src_install() {
 		rm -rv "${ED}${vimfiles}"/{macros,print,tools,tutor} || die "rm failed"
 		rm -v "${ED}"/usr/bin/vimtutor || die "rm failed"
 
+		# Delete defaults.vim to avoid conflicts with one from vim[minimal]
+		rm -v "${ED}${vimfiles}"/defaults.vim || die "rm failed"
+
 		local keep_colors="default"
 		ignore=$(rm -fr "${ED}${vimfiles}"/colors/!(${keep_colors}).vim )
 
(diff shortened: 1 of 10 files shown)


thesamesam:
Won't this collide with vim-core?

thesamesam:
> @thesamesam It should not. I have added the check for `use minimal` to avoid colliding with vim-core. AFAICT vim-core is pulled in when `minimal` is not set.

I could have vim-core installed independently though, I think (even if this seems unlikely)? It seems like it would then collide.

dongsupark:
Thanks, you are right.
I updated the PR, so that vim and vim-core do not conflict with each other.

thesamesam:
This looks good now (we need || die on external commands like 'sed', see https://devmanual.gentoo.org/ebuild-writing/error-handling/) but could you verify gvim is okay now?

i.e. Does gvim use the file that vim-core used to install? Or does it not affect gvim?

dongsupark:
Thanks, updated the PR to add `|| die` to sed and rm commands.

Checked also gvim. That is not affected by vim-core at all.

thesamesam:
Thanks for being thorough here! Really appreciate the PR.

thesamesam:
Oh, note for future: we should've possibly revbumped here (it changes the installed files/runtime behaviour of these) but it seems like a trivial issue and we're going to end up bumping for the never-ending CVEs shortly enough I suspect, so not worth it.

PR #23118, opened 2021-11-29
sys-libs/glibc: allow disabling clone3 sysflag (breaks older Electron)

thesamesam:
Bug: https://bugs.gentoo.org/827386
Bug: https://bugs.gentoo.org/819045


ArsenArsen (approved):
works for electron programs that I use.

PR #23123, opened 2021-11-30
dev-python/argon2-cffi: Bump to v21.2.0

seifertm:
The Portage package dev-python/argon2-cffi was split up into two different packages by the upstream author: argon2-cffi and [argon2-cffi-bindings.](https://pypi.org/project/argon2-cffi-bindings/) Future releases of argon2-cffi will depend on argon2-cffi-bindings. [1]

It's expected that there will be an argon2-cffi release this year. [2] I think it makes sense to add dev-python/argon2-cffi-bindings a bit ahead of time, even if it does not have any dependents in Portage, yet.

I added myself, the Python project and Proxy Maint as maintainers of the new package. I hope this is okay.

[1] https://github.com/hynek/argon2-cffi/pull/74#issuecomment-981076807
[2] https://github.com/hynek/argon2-cffi/pull/106#issuecomment-975196328


hynek:
JFTR the next bindings release is imminent and will happen _before_ argon2-cffi goes out.  So I’d wait for it to save yourself double work.

seifertm:
Added ebuild for `dev-python/argon2-cffi-21.2.0`.

The new ebuild drops a bunch of keywords, due to new dependencies on `dev-python/argon2-cffi-bindings` and `dev-python/sphinx-notfound-page`. Once this is merged, I'm planning to file a keyword request to match the supported arches with the ebuild for v21.1.0.

`argon2-cffi-21.2.0.ebuild` adds two patches to accommodate changes in argon2-cffi's build system (see comments in the ebuild). This is pretty ugly. I'd appreciate any suggestions on how to solve this more cleanly.

dev-python/argon2-cffi-bindings/argon2-cffi-bindings-21.2.0.ebuild
```diff
@@ -0,0 +1,33 @@
...
+
+DOCS=( CHANGELOG.md README.md )
+
+distutils_enable_tests pytest
+
+src_prepare() {
+	export ARGON2_CFFI_USE_SYSTEM=1
+	# We cannot call usex in global scope, so we invoke it in src_prepare
+	export ARGON2_CFFI_USE_SSE2=$(usex cpu_flags_x86_sse2 1 0)
+	default
```
mgorny, line 32: I'm pretty sure the eclass will tell you that you didn't call `distutils-r1_src_prepare`. It's easier to do these exports in `src_configure()` since it doesn't do anything by default.

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+
+DESCRIPTION="CFFI bindings to the Argon2 password hashing library"
+HOMEPAGE="https://github.com/hynek/argon2-cffi"
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+DEPEND=">=dev-python/argon2-cffi-bindings-21.2.0"
```
mgorny, line 18: Does it actually need an explicit DEPEND though? I thought this was supposed to be a pure Python package now.

dev-python/argon2-cffi-bindings/argon2-cffi-bindings-21.2.0.ebuild
```diff
@@ -0,0 +1,33 @@
...
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="cpu_flags_x86_sse2"
+
+DEPEND="
+	app-crypt/argon2:=
+	virtual/python-cffi[${PYTHON_USEDEP}]
```
mgorny, line 20: This one's a BDEP+RDEP, I think.

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
```diff
@@ -0,0 +1,37 @@
...
+	)
+"
+
+DOCS=( AUTHORS.rst CHANGELOG.md FAQ.rst README.rst )
+# Patch the tool.flit.module.name entry in pyproject.toml to include the "src" subfolder
+# This allows pyproject2setuppy to retrieve the dynamic version and description correctly
+PATCHES+=( "${FILESDIR}/${P}-pyproject-module-path.patch" )
+# Sphinx's conf.py uses importlib.metadata to determine the package version. However, argon2 is
+# not installed when we try to build the docs. Therefore, we hardcode the release version in conf.py.
+PATCHES+=( "${FILESDIR}/${P}-sphinx-importlib.patch" )
```
mgorny, line 32: This is a no-go. It's just wrong to have a patch with hardcoded version number that you'll have to redo on every bump (and somebody else will probably miss that).

mgorny (changes requested):

dev-python/argon2-cffi-bindings/argon2-cffi-bindings-21.2.0.ebuild
seifertm: You're right, it makes more sense to make these calls in `src_configure`. I moved them over, and I included the call to `distutils-r1_src_configure` in case anything changes in the eclass.

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
seifertm: That's correct. I removed the bindings from DEPEND, leaving them only in RDEPEND. I also added the missing `PYTHON_USEDEP`.

dev-python/argon2-cffi-bindings/argon2-cffi-bindings-21.2.0.ebuild
seifertm: I first looked at some other Python packages. They seem to add the virtual to `DEPEND`. I then checked the python-cffi docs, but couldn't find anything specific. Lastly, I has a look at the resulting .so file and it only links against argon2, but not against libffi:
```
$ ldd /usr/lib/python3.9/site-packages/_argon2_cffi_bindings/_ffi.abi3.so
        linux-vdso.so.1 (0x00007ffd4abec000)
        libargon2.so.1 => /usr/lib64/libargon2.so.1 (0x00007efc22e8e000)
        libc.so.6 => /lib64/libc.so.6 (0x00007efc22ccf000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007efc22caf000)
        /lib64/ld-linux-x86-64.so.2 (0x00007efc22ec8000)
```
I suppose `BDEPEND` is the correct dependency for `virtual/python-cffi` then?

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
seifertm: Yes, fair point.

Both patches are single line changes, so I replaced them with sed commands. This makes them a bit more robust against changes in the patched files and avoids a hardcoded version by inserting `PV` as the project version.

It will break if upstream creates a release with a PEP440 version that is incompatible with Gentoo version numbers, e.g. a `.beta0` suffix. However, the code in `conf.py` doesn't address this case and such version numbers weren't used in the past.

What do you think of this solution?

dev-python/argon2-cffi-bindings/argon2-cffi-bindings-21.2.0.ebuild
mgorny: Yes.

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
mgorny: I suppose `sed` is fine. I don't think we really need to care that much, and if things break, we can address them then.

dev-python/argon2-cffi-bindings/argon2-cffi-bindings-21.2.0.ebuild
mgorny: I'm not going to insist on this but wouldn't it be simpler to declare `DEPEND` and `BDEPEND` first, then do `RDEPEND="${DEPEND} ${BDEPEND}"`?

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
```diff
@@ -0,0 +1,41 @@
...
+BDEPEND="
+	test? (
+		dev-python/hypothesis[${PYTHON_USEDEP}]
+	)
+"
+
+DOCS=( AUTHORS.rst CHANGELOG.md FAQ.rst README.rst )
+
+src_prepare() {
+	# Patch the tool.flit.module.name entry in pyproject.toml to include the "src" subfolder
```
mgorny, line 28: Please keep lines below 80 chars (or 72, if you prefer).

mgorny: (unless it's really impossible)

seifertm: Of course. I reformatted the code in `src_prepare` to stay below 80 chars. The exception is the second sed substitution which will always be exceed the character count with the proper indentation.

I also checked the second ebuild (argon2-cffi-bindings) and it looks fine with regards to line length.

dev-python/argon2-cffi-bindings/argon2-cffi-bindings-21.2.0.ebuild
seifertm: Sure, I don't have a strong opinion on this.

I adjusted the dependencies `argon2-cffi-bindings-21.2.0.ebuild` accordingly.

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
```diff
@@ -0,0 +1,44 @@
...
+"
+
+DOCS=( AUTHORS.rst CHANGELOG.md FAQ.rst README.rst )
+
+src_prepare() {
+	# Patch the tool.flit.module.name entry in pyproject.toml to include
+	# the "src" subfolder. This allows pyproject2setuppy to retrieve the
+	# dynamic version and description correctly.
+	sed -i -e 's:^name = "argon2":name = "src/argon2":' \
+		"${S}/pyproject.toml" || die "Unable to patch package name"
```
mgorny, line 32:
```suggestion
		pyproject.toml || die "Unable to patch package name"
```

We tend not to refer to the current directory via absolute paths.

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
```diff
@@ -0,0 +1,44 @@
...
+	# Sphinx's conf.py uses importlib.metadata to determine the package
+	# version. However, argon2-cffi is not installed when we try to build
+	# the docs. Therefore, we patch the release version in conf.py.
+	sed -i -e "s:^release = metadata.version(\"argon2-cffi\"):release = \"${PV}\":" \
+		"${S}/docs/conf.py" || die "Unable to patch version in docs"
```
mgorny, line 37:
```suggestion
		docs/conf.py || die "Unable to patch version in docs"
```

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
```diff
@@ -0,0 +1,44 @@
...
+	)
+"
+
+DOCS=( AUTHORS.rst CHANGELOG.md FAQ.rst README.rst )
+
+src_prepare() {
+	# Patch the tool.flit.module.name entry in pyproject.toml to include
+	# the "src" subfolder. This allows pyproject2setuppy to retrieve the
+	# dynamic version and description correctly.
+	sed -i -e 's:^name = "argon2":name = "src/argon2":' \
```
mgorny, line 31: That said, ideally I'd fix pp2sp for this but honestly, I'm so tired of having to repeat every stupid hack these people come up with…

```suggestion
	sed -i -e '/^name/s:":"src/:' \
```

(untested)

Generally you can try to make sed expressions shorter when it's unlikely for them to mismatch.

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
```diff
@@ -0,0 +1,44 @@
...
+		"${S}/pyproject.toml" || die "Unable to patch package name"
+	# Sphinx's conf.py uses importlib.metadata to determine the package
+	# version. However, argon2-cffi is not installed when we try to build
+	# the docs. Therefore, we patch the release version in conf.py.
+	sed -i -e "s:^release = metadata.version(\"argon2-cffi\"):release = \"${PV}\":" \
```
mgorny, line 36:
```suggestion
	sed -i -e "/^release/s:= .*:= \"${PV}\":" \
```

seifertm: Thanks, this is nice! Now the line width is upheld properly :)

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
seifertm: This could be problematic: There are two `name` attributes in the `pyproject.toml`:
The one in the `[project]` section refers to the package name on PyPI and the one in `[tool.flit.module]` refers to the  name of the Python package.

This sed expression would be prefix both names with `src/`, but we only want to change the second one. I left it as is for now.

seifertm:
Thanks for the suggestions! I wasn't sure how to properly integrate the GitHub commit suggestions, so I performed the changes manually and added you as a co-author to the commit.

dev-python/argon2-cffi/argon2-cffi-21.2.0.ebuild
seifertm: With regards to pp2sp: I don't think this is a hack particular to argon2-cffi. Flit seems to perform some kind of file name guessing: https://github.com/pypa/flit/blob/ec6e6b7c8994d5e79d0fd951d9e4c26a7c5d9977/tests/test_init.py#L62-L86

Do you know if other packages rely on Flit guessing the `src` prefix? If there are a bunch of them that would possibly warrant a change to pyproject2setuppy.

mgorny: Yeah, I'm talking about flit and poetry. pp2sp already tries name inside `src/` by default but I didn't assume they would still do the guessing if name was explicitly specified.

mgorny (approved):
Ok, I'm going to test and merge it shortly.

PR #23124, opened 2021-11-30
media-libs/libwmf: do not install .la files

PPed72:
Closes: https://bugs.gentoo.org/827894


thesamesam:
Sure nothing uses it?

Also, we should probably revbump for it.

PPed72:
> Sure nothing uses it?

I will check ASAP (I'm busy with a few other things).

> Also, we should probably revbump for it.

I forgot, my fault.

PR #23125, opened 2021-11-30
x11-misc/clipmenu: add USE-flags to select launcher

N-R-K:
while the default menu/frontend for clipmenu is dmenu, it can also work
with rofi, fzf or any other "menu" application and thus dmenu should not
be pulled in as a hard dep.

### DIFF
--- a/x11-misc/clipmenu/clipmenu-6.2.0-r1.ebuild
+++ b/x11-misc/clipmenu/clipmenu-6.2.0-r1.ebuild
@@ -13,6 +13,7 @@ LICENSE="Unlicense"
 SLOT="0"
 KEYWORDS="amd64 x86"
 IUSE="+dmenu rofi fzf"
+REQUIRED_USE="?? ( dmenu rofi fzf )"
 
 RDEPEND="
 	x11-misc/clipnotify
@@ -21,15 +22,14 @@ RDEPEND="
 	rofi? ( x11-misc/rofi )
 	fzf? ( app-shells/fzf )
 "
-REQUIRED_USE="?? ( dmenu rofi fzf )"
 
 src_prepare() {
 	default
 
 	if use rofi ; then
-		sed -i 's|CM_LAUNCHER=dmenu|CM_LAUNCHER=rofi' clipmenu || die "sed failed"
+		sed -i 's|CM_LAUNCHER=dmenu|CM_LAUNCHER=rofi|' clipmenu || die "sed failed"
 	elif use fzf ; then
-		sed -i 's|CM_LAUNCHER=dmenu|CM_LAUNCHER=fzf' clipmenu || die "sed failed"
+		sed -i 's|CM_LAUNCHER=dmenu|CM_LAUNCHER=fzf|' clipmenu || die "sed failed"
 	fi
 }
 
--- a/x11-misc/clipmenu/clipmenu-6.2.0-r1.ebuild
+++ b/x11-misc/clipmenu/clipmenu-6.2.0-r1.ebuild
@@ -12,13 +12,26 @@ SRC_URI="https://github.com/cdown/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
 LICENSE="Unlicense"
 SLOT="0"
 KEYWORDS="amd64 x86"
-IUSE="+launcher"
+IUSE="+dmenu rofi fzf"
 
 RDEPEND="
 	x11-misc/clipnotify
 	x11-misc/xsel
-	launcher? ( || ( x11-misc/dmenu x11-misc/rofi app-shells/fzf ) )
+	dmenu? ( x11-misc/dmenu )
+	rofi? ( x11-misc/rofi )
+	fzf? ( app-shells/fzf )
 "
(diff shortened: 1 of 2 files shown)


ferki (commented):
Thanks for your PR, @N-R-K! <3

I agree that we could improve this ebuild to accommodate for more choices. At the same time, I wonder what is the exact goal here, and whether the chosen implementation is the best fit for that goal?

My impression is that a [launcher](https://github.com/cdown/clipmenu/tree/6.2.0#supported-launchers) is not an optional dependency of clipmenu-6.2.0, but it's more like an [any-of-many](https://devmanual.gentoo.org/general-concepts/dependencies/#any-of-many-dependencies) type of dependency.

What do you think about using the following in `RDEPEND` instead of `optfeature` in the ebuild?

```
|| ( x11-misc/dmenu x11-misc/rofi app-shells/fzf )
```

edit: and perhaps adding a post-install note along the lines of "dmenu is the default launcher for clipmenu; if you use something else, don't forger to configure the `CM_LAUNCHER` environment variable accordingly".

N-R-K:
> What do you think about using the following in `RDEPEND` instead of `optfeature` in the ebuild?
> 
> ```
> || ( x11-misc/dmenu x11-misc/rofi app-shells/fzf )
> ```

This can work, yes. But there's lots of other "menu applications" out there as well, for example [fzy](https://github.com/jhawthorn/fzy) should also work. There are also lesser known projects which would also work here but are not in the gentoo repos ([choice](https://pedantic.software/projects/choice.html) would be one of these I can think of the top of my head).

ferki:
> This can work, yes. But there's lots of other "menu applications" out there as well, for example [fzy](https://github.com/jhawthorn/fzy) should also work. There are also lesser known projects which would also work here but are not in the gentoo repos ([choice](https://pedantic.software/projects/choice.html) would be one of these I can think of the top of my head).
> 
> Note: I haven't tried `fzy` or `choice` with clipmenu, but those apps also output their selection to standard out so they should be "dmenu compliant" in that sense.

Hmm, I guess that's right. It seems there are two groups of launchers:

- a few specified ones that are supported by clipmenu upstream explicitly in their docs (dmenu, rofi, fzf)
- an unknown number of "might work with clipmenu" (e.g. fzy, choice, etc.).

I am not sure how much sense it makes to install/use clipmenu without a launcher, but it feels wrong to offer clipmenu to Gentoo users without one by default. I'm also still unsure about the exact motivation here, but my guess is something like "please don't install a predefined launcher for me, I know what I'm doing".

In that case, I am thinking about the following idea:

1. introduce a "launcher" USE flag, or similar, which is enabled by default
2. when enabled, install any of the supported launchers, like: `launcher? ( || ( x11-misc/dmenu x11-misc/rofi app-shells/fzf ) )`
3. when disabled, don't install any of the launchers as a dependency, but show a post-install message like: "Clipmenu is installed without depending on a launcher. Make sure you install and configure one for your use case on your own."

@N-R-K: what do you think? Would that work for your needs?

N-R-K:
For me, the motivation is that I'm mostly using clipmenud as a "backend" and have my own script for the frontend (aka clipmenu itself). I'm also using a custom dmenu build and not gentoo/ebuild (I assume that's not uncommon for suckless software).

The launcher USE flag proposal works for me. Also I think we can install the README as docs and point users there for custom launcher if `launcher` useflag is not set.

N-R-K:
@ferki One problem with is current approach is that clipmenu will install fine if someone has `rofi` or `fzf` installed but _not_ `dmenu`. But clipmenu will default to `dmenu` as it's launcher and won't quite work.

N-R-K:
Check out https://github.com/gentoo/gentoo/pull/23125/commits/21ae304d3a9be376de0a64d77d6bcc17d5211972 : it uses `sed` to change the default launcher based on the USEflag. User can also disable all "frontends" to not install any of the deps.

juippis (approved):
Thank you both!

ferki:
It looks like I am late to the party, thanks for merging, @juippis! :pray: 

---

> For me, the motivation is that I'm mostly using clipmenud as a "backend" and have my own script for the frontend/clipmenu. I'm also using a custom dmenu build and not gentoo/ebuild (I assume that's not uncommon for suckless software). So I didn't quite like the fact that dmenu is being pulled in as a dep even though it wasn't strictly necessary.

Thanks for explaining the context, I have a better understanding of it now.

> @ferki One problem with is current approach is that clipmenu will install fine for someone who doesn't have `dmenu` installed but has `rofi` or `fzf`. But clipmenu will default to `dmenu` as it's launcher and won't quite work out of the box. Users will need to manually set `$CM_LAUNCHER` to fzf or rofi.

Right, that way we should have mentioned the need to set `CM_LAUNCHER` to the correct value in the post-install messages...

> Check out 21ae304 : it uses `sed` to change the default launcher based on the USEflag. User can also disable all "frontends" to not install any of the deps.

...but I like this approach even better :+1: I think it is an acceptable and manageable amount of patching with `sed` for now, and I also learned about `??` in `REQUIRED_USE`.

Thank you for your contribution, @N-R-K! :heart:

PR #23127, opened 2021-11-30
mail-filter/rspamd: empty Toolset.cmake file

arkamar:
The Toolset.cmake file seems to be unnecessary, it tries to control flags which in Gentoo should be controlled by user. This applies a change suggested in https://bugs.gentoo.org/827550#c1.

### DIFF
--- a/mail-filter/rspamd/rspamd-3.1.ebuild
+++ b/mail-filter/rspamd/rspamd-3.1.ebuild
@@ -68,6 +68,8 @@ src_prepare() {
 
 	rm -vrf contrib/{doctest,fmt,lua-bit,snowball,zstd} || die
 
+	> cmake/Toolset.cmake || die #827550
+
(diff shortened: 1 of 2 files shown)


thesamesam:
I'd use a more explanatory commit message (your commit is the method but not the actual outcome): "respect user linker choice" or similar. I'll change it to that if it's fine with you.

arkamar:
I have changed the commit message with more generic wording. While the change was originally motivated by `-fuse-ld` flag, the `Toolset.cmake` used to control also other `CMAKE_{C,CXX,{EXE,SHARED}_LINKER}_FLAGS_*` flags. Specifically, in my case, it used to prefer non-default linker and it enabled `-fdiagnostics-color=always` for ninja build system. Anyway, you can change it to something else if you want.

thesamesam:
Thanks a lot @arkamar as always!

PR #23129, opened 2021-12-01
riscv: unmask plasma-meta[accessibility,pulseaudio], akonadi[mysql] and octave[sundials]

alexfanqi:


a17r:
@dlan17 @alexfanqi please return to this PR and do the same for latest versions (including masked in tree).

alexfanqi:
sure, thanks for pointing it out. I will test them with the latest plasma version

alexfanqi:
tested with latest plasma, all pass tests and work fine.

PR #23130, opened 2021-12-01
app-shells/loksh: bump to 7.0

nikbott:
Signed-off-by: Vinícius Kury <viniciusreyeskury@protonmail.com>


app-shells/loksh/loksh-7.0.ebuild
```diff
@@ -0,0 +1,31 @@
...
+
+DESCRIPTION="Linux port of OpenBSD's ksh"
+HOMEPAGE="https://github.com/dimkr/loksh"
+SRC_URI="https://github.com/dimkr/loksh/releases/download/${PV}/${P}.tar.xz"
+
+LICENSE="public-domain"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
+
+DEPEND="sys-libs/ncurses:0="
```
thesamesam, line 16: If we're linking against it, we need it in RDEPEND too: RDEPEND="${DEPEND}" will suffice.

I'd also drop the "0" (it's implied).

app-shells/loksh/loksh-7.0.ebuild
```diff
@@ -0,0 +1,31 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
thesamesam, line 4: EAPI 8 should work here.

app-shells/loksh/loksh-7.0.ebuild
```diff
@@ -0,0 +1,31 @@
...
+
+DESCRIPTION="Linux port of OpenBSD's ksh"
+HOMEPAGE="https://github.com/dimkr/loksh"
+SRC_URI="https://github.com/dimkr/loksh/releases/download/${PV}/${P}.tar.xz"
+
+LICENSE="public-domain"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
+
+DEPEND="sys-libs/ncurses"
```
thesamesam, line 16: Need to keep the := if we're linking against it.

Also, avoid merge commits -- no need for them in pull requests, just rebase.

nikbott: Should I revert?
Sorry, my first time contributing to gentoo so things are a bit messy.

thesamesam: No worries! See [here](https://devmanual.gentoo.org/ebuild-maintenance/git/index.html#configuration-specific-to-the-gentoo-repository) to avoid merge commits (you can adjust your config to not do merges).

I don't mind trying to detangle/drop the merge commit this time around though. Or you can have a play and see if you can get it done. You might just be able to drop it with 'git rebase -i'.

nikbott:
I definetely did sign that commit.
Guess today is not a good one for me.
Would you prefer if I started over?

PR #23136, opened 2021-12-01
sci-mathematics/petsc: version bump and bugfix

efferre79:
The latest release is needed to backport the patch from upstream
to solve the issue of Gentoo bug #827185 (build issue with
USE="fftw -mpi").

Upstream fix is at https://gitlab.com/petsc/petsc/-/merge_requests/4604

Closes: https://bugs.gentoo.org/827185
Signed-off-by: Fabio Rossi <rossi.f@inwind.it>


thesamesam:
You forgot to add the ebuild file, I think

efferre79:
> You forgot to add the ebuild file, I think

yeah sorry, just pushed the missing files now

PR #23140, opened 2021-12-01
profiles/features/musl: mask net-misc/iputils[rarpd]

stintel:
Running rarpd requires a working ether_ntohost. In musl, this is
implemented as a stub that returns -1, so rarpd always fails. Prevent
users from having to debug this by masking the rarpd USE flag on musl.

### DIFF
--- a/profiles/features/musl/package.use.mask
+++ b/profiles/features/musl/package.use.mask
@@ -1,6 +1,10 @@
 # Copyright 1999-2021 Gentoo Authors.
 # Distributed under the terms of the GNU General Public License v2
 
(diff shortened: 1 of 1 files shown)


profiles/features/musl/package.use.mask
```diff
@@ -1,6 +1,9 @@
 # Copyright 1999-2021 Gentoo Authors.
 # Distributed under the terms of the GNU General Public License v2
 
+# rarpd relies on ether_ntohost which is a stub in musl
```
floppym, line 4: Add your name, email, and the date so we know who to blame.

floppym (changes requested):

floppym (approved):

PR #23143, opened 2021-12-02
*/*: mdds & friends keywording

telans:
This must be merged before libreoffice 7.3 has a proper release or is keyworded. The python compat updates are just syncs between the ebuilds.

### DIFF
--- a/app-text/libetonyek/libetonyek-0.1.10-r1.ebuild
+++ b/app-text/libetonyek/libetonyek-0.1.10-r1.ebuild
@@ -10,8 +10,7 @@ if [[ ${PV} == *9999 ]]; then
 else
 	MDDS_VER="2.0"
 	SRC_URI="https://dev-www.libreoffice.org/src/libetonyek/${P}.tar.xz"
-	# Unkeyworded while libreoffice has no release making use of this slot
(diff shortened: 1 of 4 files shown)


a17r:
This is way too early. It will present a conflict for every LO arch (going to be 7.2.3.2) user when we likely aren't going to keyword 7.3 before 7.3.1 release.

jospezial:
https://github.com/LibreOffice/core/commit/e31191d45c27965bb1a73577b9fcc99b5d88795a
"bump liborcus dependency in pkg-config check to >= 0.17.2"
master and libreoffice-7-3 branch.
https://bugs.gentoo.org/show_bug.cgi?id=830457

thesamesam:
> [LibreOffice/core@e31191d](https://github.com/LibreOffice/core/commit/e31191d45c27965bb1a73577b9fcc99b5d88795a) "bump liborcus dependency in pkg-config check to >= 0.17.2" master and libreoffice-7-3 branch. [bugs.gentoo.org/show_bug.cgi?id=830457](https://bugs.gentoo.org/show_bug.cgi?id=830457)

In future, please file bugs with precise wording on what you think needs doing, not just quoting links.

PR #23144, opened 2021-12-02
app-office/libreoffice{,-l10n}: add 7.3.0.0_beta1

telans:
Fixed installation of l10n files for a "Dev" release (alpha/beta)
Removed the ` && [[ ${PV} != *_* ]]` check for PDEPEND on the corresponding l10n version blocking alpha/beta releases. I'm not sure why that was there. It seems it should only apply for 9999 ebuilds which don't have a corresponding l10n release.

Ebuilds are unkeyworded for testing


a17r:
We don't even have a 7.3 stable branch ebuild yet. Rather introduce that, and drop the unnecessary changes to older ebuilds.

a17r:
Our workflow is as follows:
- apply fixes as we get aware of them to 9999 ebuild
- add stable branch live ebuild when upstream branches (no more upstream dep changes from then on, only downstream fixes)
- I don't consider beta ebuilds of much value, we have never packaged them in the last years, when we can have 7.3.9999 anyway
- start thinking about adding unkeyworded ebuilds once RCs appear

telans:
> Our workflow is as follows:
> 
>     * apply changes as we get aware of them to 9999 ebuild
> 
>     * add stable branch live ebuild when upstream branches (no more upstream dep changes from then on, only downstream fixes)
> 

Upstream has branched to `libreoffice-7-3` from what I can tell

>     * I don't consider beta ebuilds of much value, we have never packaged them in the last years, when we can have 7.3.9999 anyway
> 

Fair enough, I'll look at adding 7.3.9999, though I really don't like having too many live ebuilds on my system even if I want to try the greatest and latest (tm)

>     * start thinking about adding unkeyworded ebuilds once RCs appear

Thanks for the workflow explanation

PR #23149, opened 2021-12-02
media-libs/kimageannotator: bump to 0.5.3

dflogeras:
Bug: https://bugs.gentoo.org/828017
Signed-off-by: David Flogeras <dflogeras2@gmail.com>


media-libs/kimageannotator/kimageannotator-0.5.3.ebuild
```diff
@@ -0,0 +1,47 @@
...
+	x11-libs/libX11
+"
+DEPEND="${RDEPEND}
+	test? ( dev-qt/qttest:5 )
+"
+BDEPEND="
+	dev-qt/linguist-tools:5
+"
+
+S="${WORKDIR}/${MY_P}"
```
a17r, line 36: please move it below SRC_URI (per recent convention) as it is directly related

a17r (changes requested):

juippis (approved):

PR #23156, opened 2021-12-03
dev-qt/qtnetwork: drop USE=bindist for OpenSSL changes

thesamesam:
OpenSSL no longer has USE=bindist.

Bug: https://bugs.gentoo.org/762850


thesamesam:
Actually, I've gone for it elsewhere, so let's be consistent.

PR #23157, opened 2021-12-03
sys-boot/woeusb: version bump to 5.2.4, drop 5.2.2

PetrusZ:
New release mainly focuses on bug fix and refactoring.


scardracs:
Really happy to see someone takes it up :)

scardracs:
On line 27 please add '... || die "Failed to sed"'

juippis (approved):
Thanks!

@scardracs if you add a message, it should be informative. We can already see the command that died. Here it's not so necessary since there's just one sed command so you won't mistake it for another one.

On the other hand, I'd like a comment in the ebuild above the sed command describing what it does.

PR #23160, opened 2021-12-03
x11-base/xorg-server: restore setuid for non-systemd/logind meson build

alanswanson:
The meson build does not support the autotools build option install-setuid and when not using systemd or logind the suid_wrapper is not equivalent as it does not resolve tty or input device permissions unless /etc/X11/Xwrapper.config is created with;

allowed_users = anybody
needs_root_rights = yes

(Note _needs_root_rights=auto_ only checks video card permissions and nothing else.)

In which case you can skip the wrapper as this would be the default preference when using startx or xinit without systemd or logind from the console or init scripts. So revert to previous autotools build behaviour.

### DIFF
--- a/x11-base/xorg-server/xorg-server-21.1.1-r3.ebuild
+++ b/x11-base/xorg-server/xorg-server-21.1.1-r3.ebuild
@@ -1,0 +1,187 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+XORG_DOC=doc
+XORG_TARBALL_SUFFIX="xz"
+XORG_EAUTORECONF="no"
+inherit xorg-3 meson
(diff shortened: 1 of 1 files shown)


ionenwks:
Just to inform https://bugs.gentoo.org/828462 was opened which I believe is about the same issue.

droc12345:
The way the ebuild has been modified isn't optimal, you've completely cut out the wrapper and instead just went with the old suid way, which is fine (I prefer that myself) but others might just want the wrapper option.  You need to take that into account, and the easiest way is to have another flag, "suid" changes suid of binary, "wrap" allows wrapper including setting suid properly on the wrap program.

slashbeast:
I am fine with the way this pull request make changes. The old behavior  was pure suid and not wrapper unless logind interface was present. I am interested to hear other comments but unless any significant one will pop up, I will merge it.

PR #23161, opened 2021-12-03
[media-gfx/blender]: Bump versions and fix doc building.

DarkDefender:
Blender 3.0 will release in around 5 hours from when this pull request was created.

Everything seems to work fine on my end. However we need to create a new tarball for the 3.0 test files.
@thesamesam I've commented out the fetching of this on the live and new 3.0 ebuild.
We will need update the 3.0 ebuild when the test tarball is available from the gentoo mirrors.


thesamesam:
Thanks a lot, especially for getting it in so quickly!

DarkDefender:
Thanks for merging it so quickly!

I would have liked to submit it a bit earlier as well, but I don't know how that would work in practice as the source code tarball is usually only uploaded a few hours before release.

PR #23171, opened 2021-12-03
app-shells/mcfly: version bump (v0.5.10) and remove old

mm1ke:


app-shells/mcfly/mcfly-0.5.10.ebuild
```diff
@@ -0,0 +1,115 @@
...
+"To start using ${PN}, add the following to your shell:
+
+~/.bashrc
+eval \"$(mcfly init bash)\"
+
+~/.config/fish/config.fish
+mcfly init fish | source
+
+~/.zsh
+eval \"$(mcfly init zsh)\""
```
ionenwks, line 107: Was removing the `\` from `\$(mcfly init bash)` intended? I'll assume not and fix on merge.

This cause mcfly to be ran from the ebuild itself and create a giant readme printed to elog, i.e.:
```bash
To start using mcfly, add the following to your shell:

~/.bashrc
eval "#!/bin/bash

# Ensure stdin is a tty
[[ -t 0 ]] || return
```

ionenwks (approved):
Thanks

app-shells/mcfly/mcfly-0.5.10.ebuild
mm1ke: Thanks, yes that was not indended. Don't know how that happend, i though i simply copied the latest ebuild..

PR #23172, opened 2021-12-03
gnome-extra/gnome-commander: remove old and add version 1.12.3.1

turboscholz:


ionenwks (approved):
Thanks

I'll drop the remove old commit on merge, this is removing the only stable version.

PR #23177, opened 2021-12-04
media-libs/mesa-9999: Remove classic drivers

FireBurn:
The r100, r200, i915, i965 and classic nouveau have been removed upstream
Remove prefered iris and crocus checks
Make "intel" now select irus and crocus unless disabled

Left the gallium use flag so vulkan drivers can still be built by themselves


mattst88:
Thanks, I'd already written the same patch (but also removed the gallium USE flag) in https://github.com/gentoo/gentoo/pull/23000

I see some value in making OpenGL support optional, but I don't think I'm ready to take that on yet.

PR #23181, opened 2021-12-04
app-misc/broot: bump to 1.7.4 and remove old

creideiki:
Keep 1.6.6, since it's the last version that can be built with Rust < 1.56, but remove two older 1.7.x releases.


juippis (approved):
Thanks!

PR #23182, opened 2021-12-04
Add required workload definition file

kant2002:
I add /opt/dotnet-sdk-bin-6.0/metadata/workloads/6.0.100/userlocal file
which used to indicate which workloads are installed.
Without that file Nuget think this is broken installation and attempt to create that file
which become a problem for ebuilds which depends on dotnet.

Closes: https://bugs.gentoo.org/827712
Signed-off-by: Andrii Kurdiumov <kant2002@gmail.com>


zmedico (approved):
Looks good. Thanks!

PR #23184, opened 2021-12-04
dev-util/radare2: verbump to 5.5.4

stkw0:
Closes: https://bugs.gentoo.org/815046
Bug: https://bugs.gentoo.org/790284
Bug: https://bugs.gentoo.org/807061


juippis (commented):
```
* QA Notice: Symbolic link /usr/lib64/radare2/last points to /usr/lib64/radare2/5.5.0 which does not exist.
```
Huh?
```
# file /usr/lib64/radare2/last
/usr/lib64/radare2/last: broken symbolic link to 5.5.0
```

```
 * VDB: detected possibly incorrect RDEPEND (dev-util/radare2-5.5.0)
 * dev-libs/libzip | dev-libs/libzip:=
```

stkw0:
tbh I don't know why radare2 creates /usr/lib64/radare2 folder when it doesn't put anything there, but at least with the current ebuild I don't see that symlink issue.

ajakk:
Doesn't seem like the test phase is working properly?

```
>>> Test phase: dev-util/radare2-5.5.4
make -j24   test
make: 'test' is up to date.
>>> Completed testing dev-util/radare2-5.5.4
```

I'm not sure how much radare2 and rizin have diverged at this point, but some prefix mangling inside tests might be necessary, you might be able to adapt some patches from dev-util/rizin.

Note that I've also masked

stkw0:
> Note that I've also disabled tests in Rizin due to ambiguous licensing on the binaries they distribute, so would probably be necessary to do that here too.

If we recompile those binaries then it won't be a problem right? Although I guess we will have to search all different licenses of all the sources and add those to LICENSE

ajakk:
> > Note that I've also disabled tests in Rizin due to ambiguous licensing on the binaries they distribute, so would probably be necessary to do that here too.
> 
> If we recompile those binaries then it won't be a problem right? Although I guess we will have to search all different licenses of all the sources and add those to LICENSE

Correct, and there's so many that I've not done this yet

stkw0:
thanks. For now I will just do as you and disable tests, when I do the work of searching for the difference licenses and so I will enable it again.

dev-util/radare2/radare2-5.5.4.ebuild
```diff
@@ -0,0 +1,98 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
ajakk, line 4: EAPI=8?

dev-util/radare2/radare2-5.5.4.ebuild
```diff
@@ -0,0 +1,98 @@
...
+
+	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
+fi
+
+LICENSE="GPL-2"
+SLOT="0"
+IUSE="ssl"
+
+# Need to audit licenses of the binaries used for testing
+RESTRICT="test"
```
ajakk, line 31: Are you able to get tests working locally? This situation is obviously not ideal, but ensuring tests work for at least yourself would be great.

ajakk:
Other than these minor nits, lgtm

dev-util/radare2/radare2-5.5.4.ebuild
stkw0: Well, kind of, it works if I first emerge emerge radare2 and then I execute the tests. I don't know if it's possible to do an src_install before src_test is executed so all needed libs and binaries and in the correct paths, changing the path to include all stuff seems very ugly.

juippis: > I don't know if it's possible to do an src_install before src_test is executed so all needed libs and binaries and in the correct paths, changing the path to include all stuff seems very ugly.

Yes, you can use the "${BUILD_DIR}" most likely for that. It's somewhat commonly used, the python eclasses even has built-in helpers for that.

PR #23188, opened 2021-12-05
sys-libs/libgcc, libstdc++: new ebuilds, toolchain.eclass behavior

xandris:
This branch adds ebuilds that unpack and build just libgcc and libstdc++,
carefully avoiding building the entire compiler and borrowing the native
or cross compiler where relevant.

The main use case for me is building tiny binary packages containing
libgcc_s and libstdc++ for my cross-compile environments without having
to build the entire crossed-native compiler in PORTAGE_CONFIGROOT and
carefully trimming out the bits I don't want with INSTALL_MASK. The
src_unpack is still large and slow, but the compile for these two
ebuilds is still much faster than all of gcc.

This branch mostly keeps toolchain.eclass intact. The main changes are,
toggled by the value of PN:

- ?DEPEND is minimal for these two libraries
- IUSE only contains what is relevant (TODO)
- src_prepare edits out prerequisites to avoid building gcc itself
- src_configure skips irrelevant flags
- src_compile picks different targets and does a little extra work for
  libgcc
- src_install doesn't fail if the compiler drivers aren't present in
  ${D}
- is_cxx is always true for libstdc++
- /etc/env.d/gcc grows new files per ${PN}. sys-devel/gcc still owns the
  unadorned /etc/env.d/gcc/${CHOST}-${SLOT}

It keeps the same top-level configure process intact as much as possible
to benefit from the logic in the eclass and lean on gcc's own Makefiles
to configure subprojects correctly. For example, libstdc++ is configured
during a normal `make install-target-libstdc++-v3`, and gcc's Makefile
passes the correct arguments to libstdc++-v3's configure script,
effectively propagating forward the chosen USE flags.

Main tasks left/help wanted:

- Testing with libc other than glibc
- What USE flags are relevant to libgcc and libstdc++?
  - I think libstdc++ supports vtv and that gcc and libstdc++ need to
    keep that flag in sync. Is that true?
- How to handle SLOT and gcc-config?
- How can I share patches from FILESDIR in sys-devel/gcc? Are they
  relevant?
- I think there should be a BDEPEND on the compiler; I'm pretty sure
  these need gcc to build. How do I get the right BDEPEND for the cross
  compiler?
- Can we avoid unpacking the whole tarball? (Minor annoyance)

I'm sure there was more; I appreciate any feedback!


xandris:
sorry for the force pushes, I'll settle down now

juippis:
@gentoo/toolchain

thesamesam:
> @gentoo/toolchain

We know, fwiw, it's just that this is quite complicated. We're not sure which way we want to go with this yet (although this is possibly the right way).

Discussed it w/ @xandris already -- she knows I want to handle this somehow but it's a bit down the list. Bit like CET which I took a few months to get to. I hope to look at this when I get some time and feel creative enough to think if we want any alternative solutions. So, not looked at yet, but definitely not forgotten.

unhappy-ending:
Hello, just wanted to give a heads up that I'm currently testing building this with Clang on a LLVM based system with the intent of having systems built using either libc++ or libstdc++ without relying on GCC. Seems to build "fine" with GCC but I'm getting a `clang-14: error: unknown argument: '-fbuilding-libgcc'` at the moment. Kind of interesting since I've built GCC 12 with Clang which means it also built libgcc and libstdc++ as well. 

This is just initial testing and I can dive into this deeper in the future, simply wanted to report on my very brief initial test.

I'll report more when I get some more time to much around, I was eager to see what I could do initially.

Thank you for creating the ebuilds and toolchain.eclass.

PR #23198, opened 2021-12-06
dev-libs/boost: fix AR/RANLIB configuration

10ne1:
The ebuild jam configuration did not specify any values for the
archiver and ranlib so Boost.Build would end up using the default
GNU versions even when building with Clang++.

Setting the values ensures the proper llvm-ar / llvm-ranlib are
used in LLVM build configurations.


SoapGentoo:
Thanks! I skipped the 1.76 patch, seeing the ebuild is about to disappear from the tree.

PR #23202, opened 2021-12-06
app-admin/mkosi: add missing firmware deps for booting with qemu

257:
Signed-off-by: Paymon MARANDI <darwinskernel@gmail.com>

### DIFF
--- a/app-admin/mkosi/mkosi-12-r1.ebuild
+++ b/app-admin/mkosi/mkosi-12-r1.ebuild
@@ -18,7 +18,8 @@ RDEPEND="
 	dev-vcs/git
 	sys-apps/portage
 	sys-apps/systemd
-"
+	app-emulation/qemu
+	sys-firmware/edk2-ovmf"
 
 distutils_enable_tests pytest
 


app-admin/mkosi/mkosi-12.ebuild
```diff
@@ -1,7 +1,7 @@
 # Copyright 2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=8
+EAPI=7
```
thesamesam, line 4: why?

app-admin/mkosi/mkosi-12.ebuild
```diff
@@ -18,7 +18,8 @@ RDEPEND="
 	dev-vcs/git
 	sys-apps/portage
 	sys-apps/systemd
-"
+	app-emulation/qemu
+	sys-firmware/edk2-ovmf"
```
thesamesam, line 22: change in RDEPEND requires a revbump

app-admin/mkosi/mkosi-12.ebuild
257: typo

PR #23207, opened 2021-12-07
media-libs/netpbm: fix compilation with hardened clang

ceamac:
Add `-fPIC` to `CFLAGS_FOR_BUILD` if `CFLAGS` also has it.

`CC_FOR_BUILD` is only used for some internal tools and for testing.  I think leaking `CFLAGS` over `CFLAGS_FOR_BUILD` may break cross compilation.

### DIFF
--- a/media-libs/netpbm/netpbm-10.86.27.ebuild
+++ b/media-libs/netpbm/netpbm-10.86.27.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=8
 
-inherit multilib toolchain-funcs
+inherit flag-o-matic multilib toolchain-funcs
 
 DESCRIPTION="A set of utilities for converting to/from the netpbm (and related) formats"
 HOMEPAGE="http://netpbm.sourceforge.net/"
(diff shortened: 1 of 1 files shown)


media-libs/netpbm/netpbm-10.86.27.ebuild
```diff
@@ -73,6 +73,12 @@ netpbm_config() {
 	fi
 }
 
+netpbm_cflags_for_build() {
+	if is-flagq -fPIC; then
```
thesamesam, line 77: Nit: I'd add a bug reference above this so we know why we had to do it.

media-libs/netpbm/netpbm-10.86.27.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=8
 
-inherit multilib toolchain-funcs
```
thesamesam, line 6: Sorry if I asked this before: are we actually using multilib.eclass?

ceamac: I think we do, `pkgcheck scan` reports if we don't.  I just checked on another package:
```
UnusedInherits: version 1.3.5: unused eclass: multilib
```

ceamac: I removed it and got
```
IndirectInherits: version 10.86.27: multilib: indirect inherit usage: 'get_libname', line 58
```

Great tool :)

PR #23208, opened 2021-12-07
app-backup/dar: drop 2.7.1

ceamac:
Cleanup after stabilization


ionenwks (approved):
Thanks.

It's minor but I'll drop the `Bug:` as it's not necessary to link stabilization bugs when cleaning up, just result in more mail that's not really useful (makes more sense on bugs where the cleanup itself is fixing something, e.g. security).

ceamac:
Good to know.  Thanks.

PR #23213, opened 2021-12-07
net-dialup/rp-pppoe: 3.15-r2 - remove MAX_INTERFACES limit.

jkroonza:
The 64 MAX_INTERFACES is a problem for us.  To make matters worse the
number of interfaces we need to listen on are climbing at an alarming
rate.  Need a single pppoe-server instance to share IP pool.

### DIFF
--- a/net-dialup/rp-pppoe/files/rp-pppoe-3.15-no_max_interfaces.patch
+++ b/net-dialup/rp-pppoe/files/rp-pppoe-3.15-no_max_interfaces.patch
@@ -1,0 +1,91 @@
+pppoe-server: MAX_INTERFACES 64 is a problem for ULS.
+
+We currently require 77 interfaces, this code just lifts the limit entirely and
+will keep adding interfaces for as much RAM as you have to store an array as
+required.
+
+Signed-off-by: Jaco Kroon <jaco@uls.co.za>
+
+diff -rau rp-pppoe-3.15/src.o/pppoe-server.c rp-pppoe-3.15/src/pppoe-server.c
+--- rp-pppoe-3.15.o/src/pppoe-server.c	2021-05-07 15:18:00.000000000 +0200
++++ rp-pppoe-3.15/src/pppoe-server.c	2021-12-07 21:53:46.755693003 +0200
+@@ -115,8 +115,9 @@
+ ClientSession *BusySessions = NULL;
+ 
+ /* Interfaces we're listening on */
+-Interface interfaces[MAX_INTERFACES];
++Interface *interfaces = NULL;
+ int NumInterfaces = 0;
++int MaxInterfaces = 0;
+ 
+ /* The number of session slots */
+ size_t NumSessionSlots;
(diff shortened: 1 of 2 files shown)


net-dialup/rp-pppoe/files/rp-pppoe-3.15-no_max_interfaces.patch
```diff
@@ -0,0 +1,91 @@
...
++#define INIT_INTERFACES 8
+ 
+ /* Max. 64 sessions by default */
+ #define DEFAULT_MAX_SESSIONS 64
+@@ -107,7 +107,7 @@
+ extern ClientSession *Sessions;
+ 
+ /* Interfaces we're listening on */
+-extern Interface interfaces[MAX_INTERFACES];
++extern Interface *interfaces;
```
SoapGentoo, line 88: are there plans to submit this upstream?

SoapGentoo (changes requested):

net-dialup/rp-pppoe/files/rp-pppoe-3.15-no_max_interfaces.patch
jkroonza: The plan is to try yes.  But in that case I'll expand the patch to remove pointless extern statements as well (this one is as far as I can tell not needed).

SoapGentoo:
@Polynomial-C please have a look

net-dialup/rp-pppoe/files/rp-pppoe-3.15-no_max_interfaces.patch
floppym: For a feature enhancement, it generally makes more sense to submit to upstream first and backport from there.

SoapGentoo: The problem is, this upstream is a bit special, so I understand if trying to upstream it before is not a tractable option.

jkroonza: Well, the ML subscription doesn't seem to be working ... so if someone has another means of contacting Dianne ... please get me in contact.  Happy to try get all the gentoo (not only this) patches upstream.

jkroonza: https://dianne.skoll.ca/pipermail/rp-pppoe/2021q4/000626.html

jkroonza: Upstreamed.

floppym: Upstream: https://github.com/dfskoll/rp-pppoe/pull/2

a17r:
@Polynomial-C ping

PR #23214, opened 2021-12-07
x11-misc/xbindkeys: In-place ./configure fix.

jkroonza:
Closes: https://bugs.gentoo.org/828532

Thanks ionen for tracking this.


jkroonza:
Should the guile dep not be := as well?  Also, since all versions in tree is now >= the dependency here we can drop the version requirement?

Same for tk (but not libX11 since there are no slots or subslots).

PR #23216, opened 2021-12-08
eclass/tree-sitter-grammar: fix ABI autodetection

msva:
CC: @00-matt @sarnex ?
// PR as result of a conversation on IRC to make all interested people to discuss it here

### DIFF
--- a/eclass/tree-sitter-grammar.eclass
+++ b/eclass/tree-sitter-grammar.eclass
@@ -40,15 +40,13 @@ EXPORT_FUNCTIONS src_compile src_install
 # @INTERNAL
 # @DESCRIPTION:
 # This internal function determines the ABI version of a grammar library based
-# on the package version.
+# on a constant in the source file.
 _get_tsg_abi_ver() {
-	if ver_test -gt 0.21; then
-		die "Grammar too new; unknown ABI version"
-	elif ver_test -ge 0.19.0; then
-		echo 13
-	else
-		die "Grammar too old; unknown ABI version"
-	fi
+	# This sed script finds ABI definition string in parser source file,
+	# substitutes all the string until the ABI number, and prints remains
+	# (the ABI number itself)
+	sed -n '/#define LANGUAGE_VERSION/s/.* //p' "${S}"/parser.c ||
+		die "Unable to extract ABI version for this grammar"
 }
 
 # @FUNCTION: tree-sitter-grammar_src_compile
@@ -89,8 +87,10 @@ tree-sitter-grammar_src_compile() {
 tree-sitter-grammar_src_install() {
 	debug-print-function ${FUNCNAME} "${@}"
 
-	dolib.so "${WORKDIR}"/lib${PN}$(get_libname $(_get_tsg_abi_ver))
-	dosym lib${PN}$(get_libname $(_get_tsg_abi_ver)) \
+	local soname=lib${PN}$(get_libname $(_get_tsg_abi_ver))
+
+	dolib.so "${WORKDIR}/${soname}"
+	dosym "${soname}" \
 		  /usr/$(get_libdir)/lib${PN}$(get_libname)
 }
 fi


eclass/tree-sitter-grammar.eclass
```diff
@@ -42,7 +42,9 @@ EXPORT_FUNCTIONS src_compile src_install
 # This internal function determines the ABI version of a grammar library based
 # on the package version.
 _get_tsg_abi_ver() {
-	if ver_test -gt 0.21; then
+	if [[ -n ${TS_ABI} ]]; then
```
thesamesam, line 45: Please [document](https://devmanual.gentoo.org/eclass-writing/index.html#documenting-eclasses) this variable using eclassdoc tags.

thesamesam:
Thinking about it, please explain in the commit message why this is important & why it would be needed / when you should use it.

00-matt:
Currently the tree-sitter ABI stuff is a bit half-baked as all of the grammars currently in ::gentoo are compatible with the version of libtree-sitter.

Do you know which grammar doesn't line up with the current versioning scheme?

I was originally planning on reading the version as a symbol in the library, but it looks like we can coerce it out of the `parser.c(c)` file. Should have time to do that today, unless you wanted to have a go at doing it? :)

msva:
@00-matt @thesamesam 
take a look at new ABI-version detection way (that we discussed on IRC), please

eclass/tree-sitter-grammar.eclass
```diff
@@ -42,13 +42,7 @@ EXPORT_FUNCTIONS src_compile src_install
 # This internal function determines the ABI version of a grammar library based
 # on the package version.
```
00-matt, line 43:
```suggestion
# on a constant in the source file.
```

eclass/tree-sitter-grammar.eclass
```diff
@@ -42,13 +42,7 @@ EXPORT_FUNCTIONS src_compile src_install
...
 _get_tsg_abi_ver() {
-	if ver_test -gt 0.21; then
-		die "Grammar too new; unknown ABI version"
-	elif ver_test -ge 0.19.0; then
-		echo 13
-	else
-		die "Grammar too old; unknown ABI version"
-	fi
+	sed -r -n '/#define LANGUAGE_VERSION ([0-9]*)$/{s@@\1@;p}' "${S}"/parser.c
```
00-matt, line 45: Hmm this didn't work for me, also I think we need to die on failure.

I tried to simplify it a bit and ended up with this:

```suggestion
	sed -n '/#define LANGUAGE_VERSION/s/.* //p' "${S}"/parser.c ||
	  die 'Failed to extract ABI version'
```

00-matt (changes requested):
Thank you for working on this.

If it works for all of the tree-sitter-* packages I think it's worth sending it off to gentoo-dev ml for final review :)

eclass/tree-sitter-grammar.eclass
msva: > this didn't work for me

Uhm... Strange... I tested it on multiple grammars (like `-c`, `-haskell` before commiting. But ok, I'll apply your suggested changes in few minutes.

sarnex: +1 on the die

sarnex, line 45: Can we at least add a comment explaining the regex? It's obviously much less readable than ver_test

msva: @00-matt said that regex failed for him and suggested another way (which is also much more readable), I'll test and apply it in meantime.

msva:
I've applied suggestions from @00-matt.

Is it anything else?

thesamesam:
If sarnex and matt are happy, please send to the gentoo-dev ML with:
```git send-email --to="gentoo-dev@lists.gentoo.org" --cc="sarnex@gentoo.org" --cc="matt@offtopica.uk" HEAD~1```

I'd appreciate it if you could add a bit more information to the commit message about why it's needed / what purpose it serves (what was wrong with the old approach?). Things like this can help folks in future.

If you've not used ```git send-email``` before, you might need this in ~/.gitconfig:
```
[sendemail]
       	smtpEncryption = tls
       	smtpServer = smtp.gentoo.org
       	smtpUser = mva@gentoo.org
       	smtpServerPort = 587
       	annotate = yes
```

File: eclass/tree-sitter-grammar.eclass [hunk 4]
```diff
@@ -40,15 +40,10 @@ EXPORT_FUNCTIONS src_compile src_install
...
+# on a constant in the source file.
 _get_tsg_abi_ver() {
-	if ver_test -gt 0.21; then
-		die "Grammar too new; unknown ABI version"
-	elif ver_test -ge 0.19.0; then
-		echo 13
-	else
-		die "Grammar too old; unknown ABI version"
-	fi
+	sed -n '/#define LANGUAGE_VERSION/s/.* //p' "${S}"/parser.c ||
```
sarnex, line 45: I don't understand what this sed command is doing, I could look it up, but that kind of defeats the point. Can we add a comment?

sarnex (approved):
With that one thing fixed, lgtm

msva:
@sarnex I've added description comment
@thesamesam I've added verbose decription in commit message

please, approve

sarnex (approved):

thesamesam (approved):
Happy if sarnex & matt are, next step is posting to ML.

PR #23218, opened 2021-12-08
app-emulation/ruffle initial import

ionenwks:
@thesamesam tossing it here first in case any feedback (or anyone else), want to give it some time anyway to test a few more things.

* unsure on category, considered media-video but it does call itself an emulator and media-video isn't entirely right (e.g. games).
* wanted to use system libs for libcurl/libnghttp2 but I don't see a clean way (env vars or features) at the moment
* unsure if [patch] is the best way to handle these git crates, but it works.
* not overly fond of the essentially-optfeature IUSE=gui but the .desktop doesn't behave so well without anything for dialogs and would rather not force terminal. Hopefully get a normal gui in the future rather than tinyfiledialogs, that crate is also what uses this modified zlib license.


ionenwks:
done in 6c0020e7a1ff1672bfbff6ec80294ee2fe793262

PR #23220, opened 2021-12-08
sci-mathematics/arb: remove unused patch

mm1ke:


kiwifb:
Indeed it should have been removed along with arb-2.18.0. Thanks for offering to clean up left overs :)

PR #23232, opened 2021-12-09
dev-util/strace: Fix build on systems without ipx.h

arkamar:
I have encountered building issue of `strace` after recent stabilization which affects both stable versions.
```
x86_64-gentoo-linux-musl-gcc -DHAVE_CONFIG_H   -I./linux/x86_64 -I./linux/x86_64 -I./linux/generic -I./linux/generic -I. -I.  -DIN_STRACE=1       -Wall -Wextra-Wno-missing-field-initializers -Wno-unused-parameter -Wdate-time -Wformat-security -Wimplicit-fallthrough=5 -Winit-self -Wlogical-op -Wmissing-prototypes -Wnested-externs -Wold-style-definition -Wtrampolines -Wundef -Wwrite-strings  -march=native -O2 -pipe -frecord-gcc-switches -c -o libstrace_a-netlink.o `test -f 'netlink.c' || echo './'`netlink.c
net.c:34:11: fatal error: linux/ipx.h: No such file or directory
   34 | # include <linux/ipx.h>
      |           ^~~~~~~~~~~~~
compilation terminated.
```
The problem is related to combination of musl libc, which unlike glibc does not provide `ipx.h`, and kernel-5.15, where `ipx.h` was removed. Upstream already fixed this issue in `strace-5.15` and this PR backports appropriate patch to both `strace-5.12` and `strace-5.14`.

### DIFF
--- a/dev-util/strace/files/strace-5.12-detect-ipx-presence.patch
+++ b/dev-util/strace/files/strace-5.12-detect-ipx-presence.patch
@@ -1,0 +1,141 @@
+From: Eugene Syromyatnikov <evgsyr@gmail.com>
+Date: Wed, 3 Nov 2021 00:48:59 +0100
+Subject: [PATCH] Avoid relying on presence of ipx.h
+
+After Linux has broken UAPI in commit v5.15-rc1~157^2~207, it is well
+possible that neither kernel nor libc (such as musl, for example)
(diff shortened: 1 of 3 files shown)


thesamesam:
Thanks!

PR #23245, opened 2021-12-10
x11-libs/gtk+: Removed mesa from deps

xxkfqz:
Bug: https://bugs.gentoo.org/show_bug.cgi?id=733464

Fixes [Bug 733464](https://bugs.gentoo.org/show_bug.cgi?id=733464) and confusion with USE-flags on amd64 platform when emerging something depending on the gtk+:3:
```
The following USE changes are necessary to proceed:
 (see "package.use" in the portage(5) man page for more details)
# required by x11-libs/gtk+-3.24.29::gentoo
# required by x11-themes/adwaita-icon-theme-40.1.1::gentoo
>=media-libs/mesa-21.3.1 -abi_x86_32
```

### DIFF
--- a/x11-libs/gtk+/gtk+-3.24.29.ebuild
+++ b/x11-libs/gtk+/gtk+-3.24.29.ebuild
@@ -54,7 +54,6 @@ COMMON_DEPEND="
 	)
 	X? (
 		>=app-accessibility/at-spi2-atk-2.15.1[${MULTILIB_USEDEP}]
-		media-libs/mesa[X(+),${MULTILIB_USEDEP}]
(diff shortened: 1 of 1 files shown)


stefson:
what happened?

mattst88:
I don't know why the report was closed, but typically this is because the reporter gets tired of waiting for the patch to be merged (which I understand).

The problem, from my side, is that in order to feel safe merging this I'm going to have to do a bunch of testing. If things break, and they probably will, it'll be me left picking up the pieces and not the reporter. I just haven't had the time and motivation to do that testing on top of everything else I do just in the Gentoo/GNOME project.

PR #23246, opened 2021-12-10
dev-util/clazy: use llvm.eclass to declare compatibility

thesamesam:
We need to use llvm.eclass to ensure we don't use too new of
a version of LLVM and Clang.

Closes: https://bugs.gentoo.org/823726
Bug: https://bugs.gentoo.org/823726

### DIFF
--- a/dev-util/clazy/clazy-1.10-r1.ebuild
+++ b/dev-util/clazy/clazy-1.10-r1.ebuild
@@ -3,7 +3,8 @@
 
 EAPI=7
 
-inherit cmake
+LLVM_MAX_SLOT=12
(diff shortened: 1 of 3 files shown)


dev-util/clazy/clazy-1.10-r1.ebuild
```diff
@@ -3,7 +3,9 @@
 
 EAPI=7
 
-inherit cmake
+LLVM_MAX_SLOT=12
```
a17r, line 6: Are you sure? As far as I am aware 1.10 had a commit for LLVM 13 compatibility and I had no problems.

PR #23248, opened 2021-12-11
media-libs/amdgpu-pro-vulkan: Add new package

AdelKS:
Hello,

This the first PR in a series of PRs I will open to bring up Advanced Media Framwork (AMF) for AMD GPUs. For this the closed source AMD Vulkan driver needs to be present. I tested the both the x86 and amd64 drivers on my machine with the `Battle.net` launcher that is 32bit and uses Vulkan (through DXVK), and Overwatch, which is 64bit and uses Vulkan.

Of course, I will take care of maintaining this package (as I will be using it for streaming).

Feedback on the ebuild is welcome! Hopefully this will benefit everyone.

Thank you.

Adel

PR #23250, opened 2021-12-11
media-video/ffmpeg: enable using AMD AMF

AdelKS:
Hello,

This PR comes in parallel with  #23248 to enable using AMD's `h264_amf` encoder/decoder with their proprietary vulkan driver. I can successfully use `ffmpeg` with `obs-studio` with `h264_amf`. Although `obs-studio` needs to be run with `VK_ICD_FILENAMES=/usr/share/vulkan/icd.d/amd_pro_icd64` and with the StreamFX plugin (I will open a PR for that too)

I hope this will helpful for the community.

Feedback is welcome!


thesamesam:
You should instead package.use.mask the flag for all arches lacking the keyword, don't add it yourself

AdelKS:
Yea @thesamesam I wanted to ask about this actually. Since these are only C/C++ headers, they probably are arch agnostic. However, I did not check every single header file to see if they use arch specific intrinsics, but I doubt that.

AdelKS:
I will mask all the other keywords other than `~amd64` then ask for keywording in the future, maybe. You are right, because AMD's closed source driver that provides AMF only exists for amd64 anyway. I was thinking that people can have AMF support within FFMPEG in other arrches but won't be able to use it anyway, except when Vulkan integrates the needed extensions so Mesa can plug itself with AMF.

PR #23260, opened 2021-12-12
sys-cluster/pacemaker: bump to 2.1.2 and EAPI 8

scardracs:
Package-Manager: Portage-3.0.29, Repoman-3.0.3


sys-cluster/pacemaker/pacemaker-2.1.2.ebuild
```diff
@@ -0,0 +1,75 @@
...
+		$(use_with snmp)
+	)
+
+	econf "${myconf[@]}"
+}
+
+src_install() {
+	default
+	python_optimize
+	rm -rf "${D}/var/run" "${D}/etc/init.d"
```
scardracs, line 67: I don't really like that line but actually it was like that on older version.. If you have a better solution please tell me

ceamac: What's in `/etc/init.d`, what does it remove?

ceamac, line 67:
```suggestion
	rm -r "${ED}/var/run" "${ED}/etc/init.d" || die
```

sys-cluster/pacemaker/pacemaker-2.1.2.ebuild
```diff
@@ -0,0 +1,75 @@
...
+
+	newinitd "${FILESDIR}/${PN}.initd" "${PN}"
+
+	keepdir /var/lib/pacemaker/{blackbox,cib,cores,pengine}
+	keepdir /var/log/pacemaker/bundles
+
+	find "${D}" -name '*.la' -delete || die
```
ceamac, line 74:
```suggestion
	find "${ED}" -name '*.la' -delete || die
```

ceamac (changes requested):

scardracs:
@ceamac I have no idea on why it is removing /etc/init.d

scardracs:
Looking upstream I've found it's for supports openrc, which isn't supported by default

PR #23263, opened 2021-12-12
riscv: keyword dev-python/dask

alexfanqi:
tested on sifive unmatched


mgorny:
Have you tested with `ALLOW_TEST="network"`? If not, you may want to retest at least imageio, and look for other packages on the list that have `PROPERTIES="test_network"`.

mgorny:
CC @gentoo/risc-v

alexfanqi:
> PROPERTIES="test_network"

I didn't :-P Thanks for the tips. I will retest and come back later

alexfanqi:
`grep PROPERTIES $(git diff --name-only official/master..HEAD)` shows imageio is the only package requires `ALLOW_TEST="network"`. I removed it from this pr for now and will retest it in another pr.

PR #23269, opened 2021-12-12
app-text/zathura-pdf-mupdf: bump to 0.3.7

rndxelement:
I added an ebuild for version 0.3.7 of zathura-pdf-mupdf.
Also added support for EAPI 8, added dev-libs/gumbo dependecy (now required in 0.3.7, see: [https://github.com/pwmt/zathura-pdf-mupdf/commit/921da2d554c5153606f8a534e542524f644b25dd](https://github.com/pwmt/zathura-pdf-mupdf/commit/921da2d554c5153606f8a534e542524f644b25dd)) and improved a sed | die message.


Flowdalic:
Thank you for your contribution! Could you sync the live ebuild with your changes too?

rndxelement:
> Thank you for your contribution! Could you sync the live ebuild with your changes too?

Thanks! That's a good idea. I'll do that soon

rndxelement:
> Thank you for your contribution! Could you sync the live ebuild with your changes too?

Done :)

Flowdalic:
Thanks, iwdevtools report a possible incorrect RDEPEND:

```
 * VDB: detected possibly incorrect RDEPEND (app-text/zathura-pdf-mupdf-0.3.7)
 * dev-libs/gumbo         < 
 * media-libs/jbig2dec:=  < 
 * media-libs/openjpeg:2= < 
 * virtual/jpeg           < 
```

which made me look into zathura-pdf-mupdf's buildsystem. There is a `link-external` option, which [seems to be the only way to enable those dependencies](https://github.com/pwmt/zathura-pdf-mupdf/blob/921da2d554c5153606f8a534e542524f644b25dd/meson.build#L34). But even with `link-external=true` [1] they are not a dynamic dependency of the resulting libpdf-mupdf shared object (maybe because meson enables `as-needed` per default). I have no knowledge of the mupdf library or zathura, and this is behavior probably already exists in 0.3.6. But this is something we should definitely look into it.


1: I defined the following `src_configure` in the ebuild:
```
src_configure() {
	local emesonargs=(
		-Dlink-external=true
	)
	meson_src_configure
}
```

sebastinas:
`-Dlink-external=true` is only useful if Gentoo's `libmupdf` has undefined symbols. Historically (I don't know if that changed recently), upstream's `libmupdf.a` embedded all its dependencies. Some distributions stripped those dependencies and used the system dependencies. If that's the case for Gentoo, then you need `-Dlink-external=true`. If `libmupdf` is self-contained, you don't need that.

PR #23274, opened 2021-12-12
dev-python/wxpython: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829019


t0b3:
@leio any comments on this PR?

ps: would be nice to handle hereafter #23273

AndrewAmmerlaan:
Can we bump this to the latest version first? There is a problem with the tests so we can't be sure that it works with python3_10, maybe this test problem is fixed in the latest version?

t0b3:
https://github.com/gentoo/gentoo/commit/f6cb35e73272f2a84c2f67073de1f9193426e832

PR #23276, opened 2021-12-12
sci-physics/geant4_vmc: Bump to 5.4.

olifre:
Also, switch to EAPI 8.

Signed-off-by: Oliver Freyermuth <o.freyermuth@googlemail.com>

@amadio This is probably for you ;-). 
Note that upstream has not seen many changes, so this version is still not compatible with `sci-physics/geant-4.11.0'.


amadio:
Thanks, I have the ebuild for the new Geant4 ready, wondering if I should rename to sci-physics/geant4, but should take care of that and this all together sometime this week.

olifre:
Thanks! No hurry from my side :-). 

Concerning Geant4, checking [repology](https://repology.org/projects/?search=geant), it seems that indeed `geant4` and `geant4-data` are more common across distributions. Since the [GitLab repository](https://gitlab.cern.ch/geant4/geant4) by upstream is also named this way, I think a rename would better match upstream's nomenclature.

amadio:
Yes, the name `sci-physics/geant` is a historical thing in Gentoo, as Geant3 (the old one written in Fortran) was also packaged in Gentoo as `sci-physics/geant` with `SLOT=3` (while Geant4 uses `SLOT=4`). I've wondered about renaming for a long time, but it will be quite a hassle, so I think I will just bump versions now and deal with the renaming early next year.

olifre:
That sounds reasonable, my reason to rename `geant_vmc` to `geant4_vmc` was also due to historic packaging ;-). 
Indeed renaming will also affect all dependencies, users will have to rebuild etc., so it sounds like a good plan to go for the new name in the new year :+1: .

PR #23278, opened 2021-12-12
dev-python/pyalsa: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829021


t0b3:
in fact python 3.10 is already added by 67b4faede60c29c6f6b6e9522fa434c4749827ef

PR #23297, opened 2021-12-14
dev-cpp/highway, media-libs/libjxl: new packages, new use flag in gimp to use libjxl

novomesk:
gimp and kimageformats (and maybe also ImageMagick and gThumb) have support for a new image format called JPEG XL.

`media-libs/libjxl` from ::guru will be needed to enable support.

`dev-cpp/highway` is one of the dependencies needed by libjxl.


juippis (commented):
How are gimp and kimageformats using this support? Wouldn't they need some configure option to enable it?

I think there's a policy that we can't add libraries without consumers in to the tree, so you should modify the existing packages to have this support in the initial commits. So in here, add highway, add jpegxl, modify gimp and/or kimageformats in this PR.

novomesk:
gimp and kimageformats builds JXL plug-ins by default when `libjxl` is found, unless explicitly disabled.

kimageformats has `option(KIMAGEFORMATS_JXL "Enable plugin for JPEG XL format" ON)`

gimp can explicitly disable it: `AC_ARG_WITH(jpegxl, [  --without-jpegxl        build without JPEG XL support])`

I wanted to add those package dependencies one by one (to split big problem into smaller pieces).

If you want to add all-at-once, I can add other commits here too.

novomesk:
@band-a-prend Sergey, what use flag should we use in gimp for JPEG XL support?

Do you prefer **jxl** (it it also extension of JPEG XL files) or **jpegxl** ?

juippis:
> gimp and kimageformats builds JXL plug-ins by default when `libjxl` is found, unless explicitly disabled.

We hate this behaviour and it should be patched to become 100 % controllable configure option. Please also try to get the patch upstreamed. 

novomesk:
There is `-DKIMAGEFORMATS_JXL=OFF # TODO: libjxl not packaged yet` in gentoo's kimageformats-5.89.0.ebuild now.

Isn't this cmake option enough?

media-libs/libjxl/libjxl-9999.ebuild
```diff
@@ -0,0 +1,65 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="7"
```
juippis, line 4: 8 should work here (at least once you remove git-r3 eclass)

media-libs/libjxl/libjxl-9999.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	virtual/jpeg
+	virtual/opengl
+	media-libs/freeglut
+	media-libs/giflib
+	media-libs/openexr:=
+	dev-util/google-perftools
+	x11-misc/shared-mime-info
+	dev-cpp/gtest
+	dev-cpp/highway
+	media-libs/lcms
```
juippis, line 29: Sort alphabetically please.

media-libs/libjxl/libjxl-9999.ebuild
```diff
@@ -0,0 +1,65 @@
...
+		-DBUILD_SHARED_LIBS=ON
+		-DJPEGXL_ENABLE_BENCHMARK=OFF
+		-DJPEGXL_ENABLE_COVERAGE=OFF
+		-DJPEGXL_ENABLE_FUZZERS=OFF
+		-DJPEGXL_ENABLE_SJPEG=OFF
+		-DJPEGXL_WARNINGS_AS_ERRORS=OFF
+
+		-DJPEGXL_ENABLE_SKCMS=ON
+		-DJPEGXL_ENABLE_EXAMPLES=$(usex examples)
+		-DJPEGXL_ENABLE_VIEWERS=OFF
```
juippis, line 49: What's this? Do we want to toggle it?

media-libs/libjxl/libjxl-9999.ebuild
```diff
@@ -0,0 +1,65 @@
...
+		-DJPEGXL_ENABLE_PLUGINS=ON
```
juippis, line 50: Or this?

media-libs/libjxl/libjxl-9999.ebuild
```diff
@@ -0,0 +1,65 @@
...
+		-DJPEGXL_FORCE_SYSTEM_BROTLI=ON
+		-DJPEGXL_FORCE_SYSTEM_HWY=ON
+		-DJPEGXL_FORCE_SYSTEM_GTEST=ON
```
juippis, line 53: Is this needed with `-DBUILD_TESTING=OFF`?

media-libs/libjxl/libjxl-9999.ebuild
```diff
@@ -0,0 +1,65 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="7"
+
+inherit xdg cmake git-r3
+
+DESCRIPTION="JPEG XL image format reference implementation"
+HOMEPAGE="https://github.com/libjxl/libjxl"
+
+EGIT_REPO_URI="https://github.com/libjxl/libjxl.git"
```
juippis, line 11: Could you also add a release and not a pure-9999? You see -9999 can break *any* time by upstream while release is at least guaranteed to work on a state when it was merged. You can keep maintaining the -9999 if you wish in parallel, with:
 https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide/Style_Guide#Handling_multi_live.2Frelease_ebuilds

It seems upstream is alive with frequent releases too.

media-libs/libjxl/libjxl-9999.ebuild
thesamesam: I would probably keep it even if we can't build tests for now for future work.

media-libs/libjxl/libjxl-9999.ebuild
```diff
@@ -0,0 +1,65 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="7"
+
+inherit xdg cmake git-r3
```
thesamesam, line 6: sort

media-libs/libjxl/libjxl-9999.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	media-libs/lcms
+"
+
+BDEPEND=""
+
+RDEPEND="${DEPEND}"
+
+src_configure() {
+	local mycmakeargs=(
+		-DCMAKE_SKIP_RPATH=ON
```
thesamesam, line 38: Doesn't the eclass pass this for us?

media-libs/libjxl/libjxl-9999.ebuild
```diff
@@ -0,0 +1,65 @@
...
+		-DBUILD_TESTING=OFF
+		-DBUILD_SHARED_LIBS=ON
```
thesamesam, line 40: Eclass passes this for us too, right?

media-libs/libjxl/libjxl-9999.ebuild
```diff
@@ -0,0 +1,65 @@
...
+"
+
+BDEPEND=""
+
+RDEPEND="${DEPEND}"
+
+src_configure() {
+	local mycmakeargs=(
+		-DCMAKE_SKIP_RPATH=ON
+		-DBUILD_TESTING=OFF
```
thesamesam, line 39: ?

media-libs/libjxl/libjxl-9999.ebuild
novomesk: [SKCMS](https://skia.googlesource.com/skcms) is used internally by libjxl for color-conversion purposes. According libjxl developers it works better with libjxl than lcms2.

media-libs/libjxl/libjxl-9999.ebuild
novomesk: `JPEGXL_ENABLE_EXAMPLES` are used to build `jxlinfo` - it is useful utility to print information about JXL images.

`JPEGXL_ENABLE_PLUGINS` is needed to install mime-type and to build gdk-pixbuf plug-in.

thesamesam:
(Thanks for working on this by the way!)

media-libs/libjxl/libjxl-9999.ebuild
novomesk: Building those tests sometimes fail for weird reasons, takes time, that's why I disabled them.

media-libs/libjxl/libjxl-9999.ebuild
novomesk: `jxlinfo` doesn't have install target but it is very useful tool to have.

I install jxlinfo manually but without `CMAKE_SKIP_RPATH` there is a warning about invalid rpath.

media-libs/libjxl/libjxl-9999.ebuild
novomesk: So you suggest to remove the `-DBUILD_SHARED_LIBS=ON` line?

media-libs/libjxl/libjxl-9999.ebuild
novomesk: I know the `if [[ ${PV} == 9999 ]]; then` trick.

Unfortunately, the current and previous libjxl releases are not very maintainer friendly.

Ebuild for standard release will be much complicated. There is need for heavy patching and various workarounds to satisfy gentoo policies.

I was thinking to add it later.

media-libs/libjxl/libjxl-9999.ebuild
novomesk: @thesamesam Do you have some example how it should look like?

Those tests are minefield. I'll expect troubles even if they tests are behind use flag.

novomesk: BTW, the problem with the tests was mentioned in https://bugs.gentoo.org/822444

Maybe you can look at it as I was unable to reproduce the problem and I do not know why it happened.

novomesk:
@band-a-prend GIMP is switching from gtk-doc to gi-docgen, it will be necessary to adjust gimp-9999 package.

band-a-prend:
> @band-a-prend GIMP is switching from gtk-doc to gi-docgen, it will be necessary to adjust gimp-9999 package.

Thanks for notification. I hope look through and fix it on weekends (by separate pull request). Anyway I need to check and fix some other issues too for 2.99.x.

jospezial:
> @band-a-prend GIMP is switching from gtk-doc to gi-docgen, it will be necessary to adjust gimp-9999 package.

https://bugs.gentoo.org/show_bug.cgi?id=830429

band-a-prend:
Ready: https://github.com/gentoo/gentoo/pull/23714
Now waiting for merge.

Sorry for delay.
I really had to just quick fix of build instead of trying to investigate what should be the result of build of bleeding upstream branch.

novomesk:
`media-gfx/geeqie-9999` is another viewer which is using **libjxl** to display JPEG XL images (user needs only to enable .jxl extension).

Right now it works in _automagic_ way.

PR #23301, opened 2021-12-14
www-servers/uwsgi: fix python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829204


t0b3:
https://github.com/gentoo/gentoo/commit/ef2531d06ec0220933c13f4ddc7491de4845dc16

thesamesam:
The patches weren't applied, it seems, so should we do that?

t0b3:
> The patches weren't applied, it seems, so should we do that?

True, I think this would help, yes...

www-servers/uwsgi/files/uwsgi-2.0.19.1-py310-fix.patch
```diff
@@ -0,0 +1,34 @@
+From c8c4bd1b5439217f2cb2f146caf162de69638bc1 Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: Remember to include links to where patches come from at the top.

PR #23303, opened 2021-12-14
sys-auth/pambase: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/823158


thesamesam:
98d5bc78e5cec4e7bfc1d20199726d2d605efda9

PR #23305, opened 2021-12-14
sys-cluster/legion: bump to EAPI 8 and legion-21.09.0 

scardracs:
Summary: move legion-9999 and legion-21.09.0 to EAPI 8
Removed +hwloc (see https://github.com/StanfordLegion/legion/issues/575#issuecomment-992760257)
Removed older ebuilds for cleanup


thesamesam:
You didn't tag https://bugs.gentoo.org/796128 or the other bugs? Are they fixed?

Also, I probably wouldn't cleanup *everything* quite so quickly, even though it's not marked stable, just in case people need to downgrade quickly, but I suppose it's fine here.

PR #23309, opened 2021-12-15
Bump version of vim, vim-core and gvim to v8.2.3741

nobellium1997:
Bumping the version of vim, vim-core, and gvim to patch various CVE's affecting the current vim package.


thesamesam:
Thanks.

Some notes:
1. Please include the GCO sign-off in your commits.
2. Use repoman or pkgdev commit to make your commits. E.g. the Manifest misses a newline. It'd also help with 1.
3. File a bug for these CVEs if one doesn't already exist (it might). Tag the bug with: "Bug: https://bugs.gentoo.org/1234" in each commit above your signed-off-by line. repoman and pkgdev commit help you do this.

nobellium1997:
Ah yeah I read about `repoman` and `pkgdev` but I couldn't `emerge` those tools easily in my current dev environment. I'm using a `chroot` sdk which doesn't have a package definition for either `repoman` or `pkgdev` which is why I ended up using regular git commands.

thesamesam:
Out of interest, what environment are you working in that lacks the standard Gentoo ebuild repository? And what's the "SDK" here?

Feel free to email me (sam@g.o) if you're willing to describe it but not in public. Just interested in what your usecase is (especially as I'd like us to have better contacts with downstreams).

nobellium1997:
No problem, the sdk is OSS so I can explain! I work on a fork of Chromium OS which uses their own sdk to make package changes (see here https://chromium.googlesource.com/chromiumos/docs/+/main/developer_guide.md#create-a-chroot). To stay close to upstream and not maintain our own fork, we have to update version changes upstream so that we can patch vulnerabilities without having to create our own fork of the portage overlay.

thesamesam:
Excellent! I'm quite happy to be seeing more PRs from Chromium OS and anything related to it as a general trend recently. Reducing duplication of effort helps everybody, and it's also just nice to see more from downstreams so that we can communicate if coordination is needed.

We are interested also not just in version bumps but any sort of e.g. Clang or cross fixes too, FWIW.

nobellium1997:
Yeah totally agree, way better to contribute upstream and help everyone instead of staying on a tech island. 

Alright, I tagged the correct bug, have signed off on all commits, and fixed the linting issue with the manifest files. Is there anything else that needs to be done on this PR?

PR #23310, opened 2021-12-15
dev-qt/qtwebengine: add check-reqs for disk space and memory usage

thesamesam:
Not really married to any of the implementation details here. Just wanted to get a draft up so we can discuss it. Comment away!

---

For now, assumes:
- RAM/memory: 2GB per job for GCC, 1.5GB per job for Clang
- Disk (build): 7GB
- Disk (installed): none yet

(Also moves pkg_preinst to the end of the ebuild to reflect
phase execution order, as per devmanual.)

Closes: https://bugs.gentoo.org/570534

### DIFF
--- a/dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
+++ b/dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
@@ -5,7 +5,7 @@ EAPI=8
 
 PYTHON_COMPAT=( python2_7 )
 PYTHON_REQ_USE="xml(+)"
-inherit estack flag-o-matic multiprocessing python-any-r1 qt5-build
+inherit check-reqs estack flag-o-matic multiprocessing python-any-r1 qt5-build toolchain-funcs
 
 DESCRIPTION="Library for rendering dynamic web content in Qt5 C++ and QML applications"
 HOMEPAGE="https://www.qt.io/"
@@ -110,29 +110,47 @@ PATCHES=(
 	"${WORKDIR}/${PN}-5.15.2_p20211019-jumbo-build.patch" # bug 813957
 )
 
-pkg_preinst() {
-	elog "This version of Qt WebEngine is based on Chromium version 87.0.4280, with"
-	elog "additional security fixes from newer versions. Extensive as it is, the"
-	elog "list of backports is impossible to evaluate, but always bound to be behind"
-	elog "Chromium's release schedule."
-	elog "In addition, various online services may deny service based on an outdated"
-	elog "user agent version (and/or other checks). Google is already known to do so."
-	elog
-	elog "tldr: Your web browsing experience will be compromised."
-}
-
-src_unpack() {
-	# bug 307861
+qtwebengine_check-reqs() {
+	# bug #307861
 	eshopts_push -s extglob
 	if is-flagq '-g?(gdb)?([1-9])'; then
-		ewarn
 		ewarn "You have enabled debug info (probably have -g or -ggdb in your CFLAGS/CXXFLAGS)."
 		ewarn "You may experience really long compilation times and/or increased memory usage."
 		ewarn "If compilation fails, please try removing -g/-ggdb before reporting a bug."
-		ewarn
 	fi
 	eshopts_pop
 
+	[[ ${MERGE_TYPE} != binary ]] || return
+
+	# (check-reqs added for bug #570534)
+	#
+	# Estimate the amount of RAM required
+	# Multiplier is *10 because Bash doesn't do floating point maths.
+	# Let's crudely assume ~2GB per compiler job for GCC.
+	local multiplier=20
+
+	# And call it ~1.5GB for Clang.
+	if tc-is-clang ; then
+		multiplier=15
+	fi
+
+	local CHECKREQS_DISK_BUILD="7G"
+	local CHECKREQS_DISK_USR="7G"
+	local CHECKREQS_MEMORY=$(($(makeopts_jobs)*multiplier/10))G
+
+	check-reqs_${EBUILD_PHASE_FUNC}
+}
+
+pkg_pretend() {
+	qtwebengine_check-reqs
+}
+
+pkg_setup() {
+	qtwebengine_check-reqs
+	python-any-r1_pkg_setup
+}
+
+src_unpack() {
 	case ${QT5_BUILD_TYPE} in
 		live)    git-r3_src_unpack ;&
 		release) default ;;
@@ -225,4 +243,15 @@ src_install() {
 	if [[ ! -f ${D}${QT5_LIBDIR}/libQt5WebEngine.so ]]; then
 		die "${CATEGORY}/${PF} failed to build anything. Please report to https://bugs.gentoo.org/"
 	fi
+}
+
+pkg_preinst() {
+	elog "This version of Qt WebEngine is based on Chromium version 87.0.4280, with"
+	elog "additional security fixes from newer versions. Extensive as it is, the"
+	elog "list of backports is impossible to evaluate, but always bound to be behind"
+	elog "Chromium's release schedule."
+	elog "In addition, various online services may deny service based on an outdated"
+	elog "user agent version (and/or other checks). Google is already known to do so."
+	elog
+	elog "tldr: Your web browsing experience will be compromised."
 }


thesamesam:
[02:43:51]  <@sam_> have wondered about just adding something to check-reqs or multiprocessing.eclass to just clamp MAKEOPTS jobs for big packages

ionenwks (commented):
fwiw I build this with clang+thinlto+jumbo, my build for this snapshot used 5GB disk so 7GB sound reasonable for that too (no point in making these checks too tight imo).

Also used -j11 but it never came close to using 16GB (not that I paid attention to clang processes, but it's better to be on the safe side here too so 1.5 is probably fair)

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
```diff
@@ -110,15 +110,39 @@ PATCHES=(
...
-pkg_preinst() {
-	elog "This version of Qt WebEngine is based on Chromium version 87.0.4280, with"
-	elog "additional security fixes from newer versions. Extensive as it is, the"
-	elog "list of backports is impossible to evaluate, but always bound to be behind"
-	elog "Chromium's release schedule."
-	elog "In addition, various online services may deny service based on an outdated"
-	elog "user agent version (and/or other checks). Google is already known to do so."
-	elog
-	elog "tldr: Your web browsing experience will be compromised."
+CHECKREQS_DISK_BUILD="7G"
```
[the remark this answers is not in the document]
thesamesam: I'd seen that before (a while ago) and thought it was odd. Apparently totally forgot this time.

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
```diff
@@ -110,15 +110,39 @@ PATCHES=(
...
+
+	# Let's crudely assume ~2GB per compiler job for GCC.
+	local multiplier=2
+
+	# And call it ~1.5GB for Clang.
+	if tc-is-clang ; then
+		multiplier=1.5
+	fi
+
+	echo "$(($(makeopts_jobs)*${multiplier}))M"
```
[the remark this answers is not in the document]
thesamesam: Thanks, I should've realised the FP thing (you've probably said it before, thinking about it).

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
```diff
@@ -110,15 +110,39 @@ PATCHES=(
...
+}
+
+pkg_pretend() {
+	if [[ ${MERGE_TYPE} != binary ]] ; then
+		# check-reqs has a binary guard but let's wrap the whole thing
+		# for the memory-estimator so we don't try find e.g. CC on binpkg consumers.
+		export CHECKREQS_MEMORY=$(_required_ram_available)
```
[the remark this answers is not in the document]
thesamesam: Very snazzy!

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
```diff
@@ -110,29 +110,47 @@ PATCHES=(
...
 	eshopts_push -s extglob
 	if is-flagq '-g?(gdb)?([1-9])'; then
-		ewarn
-		ewarn "You have enabled debug info (probably have -g or -ggdb in your CFLAGS/CXXFLAGS)."
-		ewarn "You may experience really long compilation times and/or increased memory usage."
-		ewarn "If compilation fails, please try removing -g/-ggdb before reporting a bug."
-		ewarn
+		# Debug symbols require significantly more memory for the compiler
+		# to juggle.
+		multiplier=$((multiplier*2))
```
Pesa, line 133: doesn't it require more disk space as well?

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
```diff
@@ -110,29 +110,47 @@ PATCHES=(
...
 	fi
 	eshopts_pop
 
+	local CHECKREQS_DISK_BUILD="7G"
+	local CHECKREQS_DISK_USR="7G"
+	local CHECKREQS_MEMORY=$(($(makeopts_jobs)*${multiplier}/10))G
```
Pesa, line 139: I believe the `${}` is redundant inside `$((...))`

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
```diff
@@ -110,29 +110,47 @@ PATCHES=(
...
+		multiplier=15
+	fi
+
+	# bug #307861
 	eshopts_push -s extglob
 	if is-flagq '-g?(gdb)?([1-9])'; then
-		ewarn
-		ewarn "You have enabled debug info (probably have -g or -ggdb in your CFLAGS/CXXFLAGS)."
-		ewarn "You may experience really long compilation times and/or increased memory usage."
-		ewarn "If compilation fails, please try removing -g/-ggdb before reporting a bug."
```
Pesa, line 131: hmm I would keep the warning... maybe? Not a strong reason, but it has happened before that users "forgot" that they have debugging flags in their CXXFLAGS... On the other hand, who reads the warnings anyway? 😛

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
thesamesam: I think you're right. Ionen mentioned it [here](https://github.com/gentoo/gentoo/pull/23310#discussion_r769216237). It's not my ebuild so I'm happy to use no ${} if you prefer that.

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
thesamesam: Ah, yes!

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
thesamesam: Given the aim here is to hopefully reduce the # of people reporting OOM bugs, I think you're right here, I'll keep the warning?

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
ionenwks: No sure if debugging really need that much more memory (would need to check), when I mentioned x2 it was mostly for disk space.

ionenwks: It may be more than x2 too, I think chromium can get absurd with this.

thesamesam: Sorry, yes, that's obvious now you say it. I don't have a good estimate for how much more _memory_ usage it'll cause.

Might be worth not addressing it for memory and just handling it for disk where we can at least say >= 2x safely, I think.

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
thesamesam: OK, kept it. And dropped the logic for messing with check-reqs sizes based on debugging symbols for now for simplicity.

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
thesamesam: https://github.com/gentoo/gentoo/pull/23310#discussion_r771086219.

a17r:
sure why not.

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
```diff
@@ -110,29 +110,47 @@ PATCHES=(
...
+	# Let's crudely assume ~2GB per compiler job for GCC.
+	local multiplier=20
+
+	# And call it ~1.5GB for Clang.
+	if tc-is-clang ; then
+		multiplier=15
+	fi
+
+	local CHECKREQS_DISK_BUILD="7G"
+	local CHECKREQS_DISK_USR="7G"
```
Pesa, line 138: uhm really 7 GB of installed size? I don't currently have qtwebengine installed so cannot quickly check... can anyone confirm?

Pesa (commented):
Looks good to me, though you may want to double check that value for `CHECKREQS_DISK_USR`, seems too high to me.

dev-qt/qtwebengine/qtwebengine-5.15.2_p20211210.ebuild
a17r:
```
# equery s qtwebengine
 * dev-qt/qtwebengine-5.15.2_p20211216
         Total files : 263
         Total size  : 149.60 MiB
```

thesamesam: yeah, seems obviously wrong, will fix

thesamesam:
Done in 2b134e09a3ab94f3c33419cf0b8114d004cec477 thanks all!

PR #23311, opened 2021-12-15
check-reqs.eclass: clamp MAKEOPTS based on available memory

thesamesam:


mgorny:
To be honest, I'm not sure if this is conceptually sound.

The `njobs * 2 GiB` requirement is the absolute minimum we expect from users and we don't check for that. If a package uses check-reqs, then it means it has requirements above the minimum. In that case, clipping jobs to match our minimum requirement doesn't resolve the problem.

I mean, you're adding a workaround for misconfigured systems to packages where normally it isn't supposed to help anyway.

thesamesam:
> To be honest, I'm not sure if this is conceptually sound.
> 
> The `njobs * 2 GiB` requirement is the absolute minimum we expect from users and we don't check for that. If a package uses check-reqs, then it means it has requirements above the minimum. In that case, clipping jobs to match our minimum requirement doesn't resolve the problem.
> 
> I mean, you're adding a workaround for misconfigured systems to packages where normally it isn't supposed to help anyway.

I take your point (and agree it is the absolute minimum) but it doesn't change reality either. We get a _lot_ of users who don't realise that's the minimum. The handbook now documents reasonable standards but that doesn't help the thousands and thousands of users over the years who never updated it (and tbh, why should they know to update it?).

My expectation would be we use this for pkgs that:
1) has significant memory use and;
2) takes a while to build would use check-reqs for the memory component to save users some pain. 

Aside: I don't actually think we're even using check-reqs in its current tree state in enough packages anyway for stuff above the minimum requirements.

Not that many fit into both: qtwebengine, maybe llvm/clang, firefox, spidermonkey, rust, webkit-gtk, chromium. So, I don't expect to add this everywhere  I just want to have it to avoid having to continue marking qtwebengine dupes and avoid frustration when a long-running build fails. I don't think too much falls into that bin.

Alternatively, we warn in Portage? I don't think we could get away with clamping there though, it would have to be just a warning, which isn't very satisfactory.

(I also have sympathy for people who try to have a higher `njobs` but don't realise what the big packages are. It kind of sucks if you're on a box with a lot of cores that is slow at single-core perf. Most stuff will build fine with way higher than RAM/2GB. But a few won't.)

mgorny:
A warning in Portage would be a good first step. Then we can see how often the problem happens still.

mattst88:
I've never found the 2G*proc limit to be critical for anything except the largest packages (like firefox, chromium, etc). Those packages are precisely the ones that use the check-reqs eclass.

thesamesam:
Upon reflection (and after spending a lot of time looking at Portage earlier for something else),I don't really want to add even more complexity to Portage if I can help it for now.

As probably one of the main devs who handles user support, OOMs happen _all the time_ and part of the reason is that -jN where N = cores without regard for RAM actually works until you hit a big boy like chromium, hence they're confused when they find it doesn't work on a handful of packages.

eclass/check-reqs.eclass
```diff
@@ -43,16 +43,25 @@ case ${EAPI} in
...
 esac
 
-EXPORT_FUNCTIONS pkg_pretend pkg_setup
-
 if [[ ! ${_CHECK_REQS_ECLASS} ]]; then
 _CHECK_REQS_ECLASS=1
 
+inherit multiprocessing
+
+EXPORT_FUNCTIONS pkg_pretend pkg_setup
```
Arfrever, line 51: `EXPORT_FUNCTIONS` should be probably outside inclusion guard (`if [[ ! ${_CHECK_REQS_ECLASS} ]]; then`) (and still after any calls to `inherit`).

https://gitweb.gentoo.org/repo/gentoo.git/log/eclass?qt=grep&q=EXPORT_FUNCTIONS
https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=f24a4018a32e3cc7bf6017ccb944574acb7f4727

PR #23314, opened 2021-12-15
gui-apps/mako: Add sys-libs/basu sd-bus-provider

Flinkbaum:
Add sys-libs/basu as alternative sd-bus-provider besides elogind and
systemd

Bug: https://bugs.gentoo.org/783156
Signed-off-by: Michael Kupfer <mkupfer99@gmail.com>


thesamesam:
Thanks! Changes RDEPEND so needs a revbump, done that for you.

PR #23315, opened 2021-12-15
dev-util/clion: version bump to 2021.3 and drop older ebuilds

strahlc:
dev-util/clion: version bump to 2021.3
dev-util/clion: remove old versions


scardracs:
Please use a title like "dev-util/clion: version bump to 2021.3 and drop older ebuilds" for helping devs on understand faster what are you actually doing. Also consider signing your commits per GLEP 66: https://www.gentoo.org/glep/glep-0066.html#openpgp-signatures

PR #23318, opened 2021-12-15
dev-python/python-slugify-4.0.1: fix incorrect usage of DISTUTILS_USE_SETUPTOOLS

robert7k:
Closes: https://bugs.gentoo.org/829082


juippis (approved):
Thanks! I'm gonna strip the EAPI bump off, since we usually revbump on EAPI bumps. Especially since this has stable keywords, and existing revdeps. 

You could also bump the PYTHON_COMPAT to support 3.10, and remove support from 3.7.

PR #23322, opened 2021-12-15
dev-python/backports-{tempfile,weakref}: revbump to r2 and move to PYTHON3_{8..10}

scardracs:


mgorny:
Why would you be adding new implementations to a backport package?

scardracs:
Because bmaptools (on ::guru) depends on these for tests

mgorny:
Wouldn't it be better to fix bmaptools then?

scardracs:
> Wouldn't it be better to fix bmaptools then?

What you mean about "fix bmaptools"? Are these backports like "patches"?

PR #23323, opened 2021-12-15
acct-group/opendmarc: fix trailing space and bump to EAPI 8

scardracs:


grobian:
pushed, thanks

PR #23324, opened 2021-12-16
sys-block/hpssacli: add 4.21.7.0, 5.20.8.0

Flowdalic:
Closes: https://bugs.gentoo.org/674774


sys-block/hpssacli/hpssacli-4.21.7.0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+src_install() {
+	local -r ssacli_bindir="${WORKDIR}/opt/smartstorageadmin/ssacli/bin"
+
+	dosbin "${ssacli_bindir}"/ssacli
+	dosbin "${ssacli_bindir}"/ssascripting
+	dosbin "${ssacli_bindir}"/rmstr
+
+	dodoc "${ssacli_bindir}/ssacli-${MY_PV}.x86_64.txt"
+
+	gunzip usr/man/man8/ssacli.8.gz || die
```
Whissi, line 33: I think we are missing a `BDEPEND` on `app-arch/gzip` for this.

sys-block/hpssacli/hpssacli-4.21.7.0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+KEYWORDS="-* ~amd64"
+
+S="${WORKDIR}"
+
+src_unpack() {
+	rpm_src_unpack
+}
+
+src_install() {
+	local -r ssacli_bindir="${WORKDIR}/opt/smartstorageadmin/ssacli/bin"
```
Whissi, line 25: Is the read-only variable required?

Whissi (changes requested):

sys-block/hpssacli/hpssacli-4.21.7.0.ebuild
Flowdalic: [`app-arch/gzip` is part of the `@system` set](https://github.com/gentoo/gentoo/blob/a45623966ad5651c9e3beb9e09428dbc09f1da52/profiles/base/packages#L29), so I think we can omit it.

sys-block/hpssacli/hpssacli-4.21.7.0.ebuild
Flowdalic: Due it's nature, it can't be required. Personally I prefer marking things as read-only/immutable when they are. I am, however, happy to remove the `-r` if this is a blocker for you. :)

Flowdalic:
@Whissi friendly ping

sys-block/hpssacli/hpssacli-5.20.8.0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+SRC_URI="https://downloads.hpe.com/pub/softlib2/software1/pubsw-linux/p1736097809/v193233/ssacli-${MY_PV}.x86_64.rpm"
+
+LICENSE="hp-proliant-essentials"
+SLOT="0"
+KEYWORDS="-* ~amd64"
+
+S="${WORKDIR}"
+
+src_unpack() {
+	rpm_src_unpack
```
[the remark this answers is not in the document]
Flowdalic: Not sure if I understand, but if we drop this, nothing will unpack the `.rpm`. At least simply removing `src_unpack` breaks the ebuild for me.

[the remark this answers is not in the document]
thesamesam: It doesn't export it? Huh.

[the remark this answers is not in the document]
Flowdalic: I am sorry, I am being stupid. Yes the eclass does export it, S was just wrong after removing src_unpack.

PR #23326, opened 2021-12-16
dev-dotnet/dotnet-sdk-bin: version bumps + add .net core 3.1 as a new slot

ngg:
.NET Core 3.1 is an LTS version and officially supported by Microsoft until December 03, 2022, I think it would worth adding it. In my company, 3.1 is still very much in use as .NET 5 was not an LTS version and .NET 6 was only recently released.


ngg:
cc @zmedico

zmedico:
Thanks!

PR #23330, opened 2021-12-16
www-apache/mod_wsgi: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829359

### DIFF
--- a/www-apache/mod_wsgi/files/mod_wsgi-4.7.1-py310.patch
+++ b/www-apache/mod_wsgi/files/mod_wsgi-4.7.1-py310.patch
@@ -1,0 +1,126 @@
+From b439f1c411a9479ccc03c16465cdff50fede79d3 Mon Sep 17 00:00:00 2001
+From: Petr Viktorin <encukou@gmail.com>
+Date: Thu, 10 Jun 2021 15:45:03 +0200
+Subject: [PATCH] Use Py_CompileString rather than
+ PyParser_SimpleParseFile/PyNode_Compile
+From: https://github.com/GrahamDumpleton/mod_wsgi/commit/b439f1c411a9479ccc03c16465cdff50fede79d3
+
+---
+ src/server/mod_wsgi.c    | 68 +++++++++++++++++++++++++++++++---------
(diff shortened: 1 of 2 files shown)


www-apache/mod_wsgi/files/mod_wsgi-4.7.1-py310.patch
```diff
@@ -0,0 +1,125 @@
+From b439f1c411a9479ccc03c16465cdff50fede79d3 Mon Sep 17 00:00:00 2001
+From: Petr Viktorin <encukou@gmail.com>
+Date: Thu, 10 Jun 2021 15:45:03 +0200
+Subject: [PATCH] Use Py_CompileString rather than
```
djc, line 4: Where did this come from? Maybe put a link in there?

djc (commented):
Would like to verify the provenance of the patch here, otherwise seems fine.

www-apache/mod_wsgi/files/mod_wsgi-4.7.1-py310.patch
t0b3: It's simply backported from upstream https://github.com/GrahamDumpleton/mod_wsgi/commit/b439f1c411a9479ccc03c16465cdff50fede79d3
How would you like to mention it? Your suggestion is very welcome...

djc: Putting that link in the body of the patch email (that is, between the `Subject` header and the `---` preceding the diffstat), should work fine.

t0b3:
@djc thanks for your pragmatic proposition. Can you add the link on my
behalf? I'm only on emails these weeks. This would be perfect.

Dirkjan Ochtman ***@***.***> wrote on Sun., 19. Dez. 2021,
20:52:

djc (approved):

djc:
LGTM, thanks!

djc:
Rebased as e0ec91c1c0a09edb5c9d4b2e9b81d256e2be5d71, thanks!

PR #23332, opened 2021-12-16
sys-apps/policycoreutils: update the right store

t-lo:
The policycoreutils ebuild calls `semodule` in `postinst` to update SELinux stores. It does not, however, tells semodule the correct ROOT to use, so installing `policycoreutils` in a crossdev environment will actually update the **host's** store.

This patch adds `-S "${ROOT:-/}"` to the `semodule` call so the correct environment is updated.

First seen + fixed in Flatcar Container Linux: https://github.com/flatcar-linux/coreos-overlay/pull/1502


thesamesam:
Nice catch & thank you for submitting this up to us!

PR #23334, opened 2021-12-16
app-misc/broot: Bump to 1.7.5

creideiki:
Not sure if this is relevant, since the only changes between 1.7.4 and 1.7.5 are to the build defaults, and the ebuild overrides them anyway, so there are no changes to the ebuild and should be no actual effects. But at least this checks that the ebuild still works with the new defaults, and prevents anyone from complaining about a missing package version.


ionenwks (approved):
Well, the wasted cpu cycles can't be much worse than kernel 5.15.9 where its only change was a file permission :eyes:

PR #23353, opened 2021-12-17
sys-fs/snapraid: 12.1 version bump

AndCycle:


slashbeast:
We have 12.2 version in tree. I will close this one as obsolete.

PR #23356, opened 2021-12-17
dev-cpp/gtkmm: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829417


leio:
Please handle newer versions as well (SLOT=4.0), otherwise the newest won't have py3.10 support while older does

t0b3:
> Please handle newer versions as well (SLOT=4.0), otherwise the newest won't have py3.10 support while older does (probably same for pangomm and co, if you have other separate PRs for those)

Hmm, I thought gtkmm:4 has py3.10 support already, doesn't it?

PR #23357, opened 2021-12-17
dev-cpp/pangomm: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829416


t0b3:
https://github.com/gentoo/gentoo/commit/cbbc4230812e3d7484284a9fcd4ef9c01a833df3

PR #23363, opened 2021-12-17
dev-libs/sord: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829422


fordfrog:
@t0b3 merged, thanks!

PR #23365, opened 2021-12-17
gnome-base/gsettings-desktop-schemas: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829425


t0b3:
a5e2127f429cae431446b74bf8d764007236779b

PR #23366, opened 2021-12-17
gnome-base/gnome-keyring: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829424


t0b3:
@mattst88 any comments on this PR?

PR #23368, opened 2021-12-17
x11-libs/libva-intel-media-driver: fixed dependency on gmmlib version

Karlson2k:
media-libs/gmmlib-21.3.4 and later break compatibility with old API.

Bug: https://bugs.gentoo.org/829406

No rev-bumps.

### DIFF
--- a/x11-libs/libva-intel-media-driver/libva-intel-media-driver-20.4.5-r1.ebuild
+++ b/x11-libs/libva-intel-media-driver/libva-intel-media-driver-20.4.5-r1.ebuild
@@ -12,12 +12,9 @@ if [[ ${PV} == *9999 ]] ; then
 	fi
 	inherit git-r3
 else
-	MY_PV="${PV%_pre}"
-	SRC_URI="https://github.com/intel/media-driver/archive/intel-media-${MY_PV}.tar.gz"
-	S="${WORKDIR}/media-driver-intel-media-${MY_PV}"
-	if [[ ${PV} != *_pre* ]] ; then
-		KEYWORDS="amd64"
-	fi
+	SRC_URI="https://github.com/intel/media-driver/archive/intel-media-${PV}.tar.gz"
+	S="${WORKDIR}/media-driver-intel-media-${PV}"
+	KEYWORDS="amd64"
 fi
 
 DESCRIPTION="Intel Media Driver for VAAPI (iHD)"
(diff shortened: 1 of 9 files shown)


thesamesam:
This (<=) won't cover any revbumps for gmmlib.

Karlson2k:
Corrected to `<media-libs/gmmlib-21.3.4`

Karlson2k:
Related: as gmmlib [changed API and ABI](https://github.com/intel/gmmlib/commit/6143502f4e5275b9d285693704b9cfab04819f78), does it make sense to use slot or subslot for gmmlib?

thesamesam:
> Related: as gmmlib [changed API and ABI](https://github.com/intel/gmmlib/commit/6143502f4e5275b9d285693704b9cfab04819f78), does it make sense to use slot or subslot for gmmlib?

Yeah, existing binaries need to continue working, so we should be using a subslot. Consumers must be revbumped to depend on :=.

We can only use a slot if it's easy to install versions in parallel **and** make consumers use the right version easily (this is usually quite difficult).

Karlson2k:
Added commit with rebuild on libgmm subslot change with revbumps for libva-intel-media-driver.

thesamesam:
Thanks @Karlson2k. Which versions changed API and ABI? We need to add a subslot to the _newer_ one (no need to do it to the older one, as it has an implicit (sub)slot of 0/0).

Karlson2k:
The first version of [gmmlib](https://github.com/intel/gmmlib) with [the new API and ABI](https://github.com/intel/gmmlib/commit/6143502f4e5275b9d285693704b9cfab04819f78) is [21.3.4](https://github.com/intel/gmmlib/tree/intel-gmmlib-21.3.4). Version [22.0.0](https://github.com/intel/gmmlib/tree/intel-gmmlib-22.0.0) has [new API/ABI](https://github.com/intel/gmmlib/blob/intel-gmmlib-22.0.0/Source/GmmLib/CMakeLists.txt#L27) as well.

However, [version 21.3.5](https://github.com/intel/gmmlib/tree/intel-gmmlib-21.3.5) has [old API and ABI](https://github.com/intel/gmmlib/blob/intel-gmmlib-21.3.5/Source/GmmLib/CMakeLists.txt#L27).

Looks like they released 21.3.4 version with the new API but then decided to change API only for 22.0.0 versions.
Maybe add ebuild for gmmlib 21.3.5 and drop version 21.3.4 to simplify dependency relations?
Looking at the git history, it's clear that [version 21.3.5 is based on older](https://github.com/intel/gmmlib/commits/intel-gmmlib-21.3.5) commit than [version 21.3.4](https://github.com/intel/gmmlib/commits/intel-gmmlib-21.3.4)

Karlson2k:
I would add gmmlib version `21.3.5` as `21.3.3-21.3.5` or simply as `21.3.3-5`, because version 21.3.5 should go before version 21.3.4.

thesamesam:
> I would add gmmlib version `21.3.5` as `21.3.3-21.3.5` or simply as `21.3.3-5`, because version 21.3.5 should go before version 21.3.4.

21.3.3_p5? :/

Karlson2k:
@thesamesam I'm on it.

PR #23370, opened 2021-12-17
media-libs/gexiv2: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829432


t0b3:
@mattst88 any comments on this PR?

PR #23371, opened 2021-12-17
media-libs/graphene: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829434


t0b3:
@thesamesam do you think this PR is good? or you have any comments?

PR #23372, opened 2021-12-17
media-libs/jbig2dec: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829435


t0b3:
@tamiko any comments on this PR?

PR #23377, opened 2021-12-17
media-libs/lv2: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829440

### DIFF
--- a/media-libs/lv2/lv2-1.18.2-r1.ebuild
+++ b/media-libs/lv2/lv2-1.18.2-r1.ebuild
@@ -1,0 +1,62 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{8..10} )
+PYTHON_REQ_USE='threads(+)'
+
+inherit python-single-r1 waf-utils multilib-build multilib-minimal
+
+DESCRIPTION="A simple but extensible successor of LADSPA"
+HOMEPAGE="https://lv2plug.in/"
+SRC_URI="https://lv2plug.in/spec/${P}.tar.bz2"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~alpha amd64 arm arm64 ~hppa ~mips ppc ppc64 ~riscv ~sparc x86"
+IUSE="doc plugins"
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+BDEPEND="
+	plugins? ( virtual/pkgconfig )
+	doc? ( app-doc/doxygen dev-python/rdflib )
+"
+CDEPEND="
+	${PYTHON_DEPS}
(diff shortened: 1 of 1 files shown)


fordfrog:
this one does not pass the validation because of `dev-python/rdflib` dropping `ia64` keyword. not sure how to handle it, maybe we'd need to drop it too...

t0b3:
I'm open for dropping ia64 if it's no longer available on the dependency
rdflib.

Miroslav Šulc ***@***.***> wrote on Sat., 18. Dez. 2021,
12:02:

t0b3:
Hmm, dropping ~ia64 would impact all the depending ebuilds:
- lilv
- sratom
- ffmpeg
Maybe the best solution would be to add the ~ia64 KEYWORD to latest rdflib ebuils.

t0b3:
depends on PR https://github.com/gentoo/gentoo/pull/23494

arthurzam:
Well, adding ia64 to rdflib won't happen in near future (as it's dependency berkeleydb has ilegal instuction on it).
I propose a better solution which would be easiest and best in my opinion: mask doc use flag on ia64.

In `gentoo:/profiles/arch/ia64/package.use.mask` add a mask with comment and author for:
`media-libs/lv2 doc`
You can see an example made by me there for how it can look.

arthurzam:
Please split this into two commits:

1. mask the doc USE
2. enable py3.10

profiles/arch/ia64/package.use.mask
```diff
@@ -1,6 +1,10 @@
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Thomas Bettler <thomas.bettler@gmail.com> (2021-12-24)
+# dev-python/rdflib isn't keyworded here. Waiting for bug #814179
+media-libs/lv2 doc
```
thesamesam, line 6: Yep!

thesamesam: Hm, it's unconditionally RDEPENDing on it too... https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/lv2/lv2-1.18.2.ebuild#n38

Is that correct?

t0b3: Hmm, thank you for spotting this :) We need to take another solution...

thesamesam:
On the one hand, this is quite a heavy-handed approach (it'd be better to just add 3.10 to the latest version in a new revision, then block cleanup until we have a resolution). On the other hand, not convinced anyone is actually using this on ia64 anyway.

t0b3:
Result is now:
lv2-1.18.2 | PythonCompatUpdate | PYTHON_COMPAT update available: python3_10
lv2-1.18.2-r1 | DroppedKeywords | ia64

If you are fine with this, let's proceed ...

t0b3:
@thesamesam let's go then w/o ia64?

PR #23380, opened 2021-12-17
media-libs/suil: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829443


fordfrog:
@t0b3 merged, thanks!

PR #23382, opened 2021-12-17
dev-util/scons: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829445


band-a-prend:
Please see the related upstream issue: https://pairlist4.pair.net/pipermail/scons-users/2021-December/008697.html

The problem with this test could be ignored. The Scons python 3.10 is required to add python 3.10 to sci-libs/cantera

t0b3:
e58bb6829e489bb87bd028bbdad91c9aa5d3fd81

PR #23383, opened 2021-12-17
dev-util/spirv-tools: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829446


mattst88:
v1.2.198 is actually newer than v20210825, so I've updated the patch to modify that version instead.

PR #23392, opened 2021-12-17
app-crypt/openpgp-keys* → sec-keys/

mgorny:
CC @gentoo/qa

That said, I'm wondering if it shouldn't be `sec-key`. We don't seem being consistent there.


thesamesam (approved):
My weak pref is for keys.

gyakovlev:
ack for zfs and rust
I vote for `keys` if it counts.

gyakovlev (approved):

akhuettel (approved):

PR #23394, opened 2021-12-17
www-apps/gitea: bump to 1.15.7

hydrapolic:


thesamesam:
Not synced live?

hydrapolic:
> Not synced live?

Since they release every week I don't really know if there's a point of keeping live actually. But can do next time, good point.

And thanks @thesamesam !

PR #23395, opened 2021-12-17
net-analyzer/linkchecker: bump to 10.1.0

hydrapolic:

PR #23399, opened 2021-12-18
dev-libs/criterion: add snapshot

hydrapolic:


thesamesam:
See https://github.com/gentoo/gentoo/pull/23395#issuecomment-998402352 kind of, although I know the situation here.

hydrapolic:
> See [#23395 (comment)](https://github.com/gentoo/gentoo/pull/23395#issuecomment-998402352) kind of, although I know the situation here.

Last release/tag was 3 years ago. This is simply to bring us up to date because it's used by syslog-ng so we don't miss any updates and be earlier prepared when 2.4 comes out.

hydrapolic:
Thank you @thesamesam

PR #23402, opened 2021-12-18
app-shells/starship: Add 1.0.0

bowlofeggs:


ionenwks (approved):
Thanks.

On a side-note, looks like it started using edition 2021 in this version, meaning needs >=virtual/rust-1.56 (eclass depends on 1.53 for now). Not that it's a big deal considering 1.56.1 is stable, but do run into people using old rust at times, so I'll add in the BDEPEND to be pedantic.

https://github.com/starship/starship/commit/5ff08708c5e7d5b127d8583ff2be09a90d2e0e42

bowlofeggs:
Good catch @ionenwks, and thanks!

PR #23408, opened 2021-12-18
Bitcoin 21.2

luke-jr:
LTS branch


luke-jr:
cc @juippis @candrews @thesamesam @Flowdalic

dev-util/bitcoin-tx/bitcoin-tx-21.2.ebuild
```diff
@@ -0,0 +1,100 @@
...
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~mips ~ppc ~ppc64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="knots"
+
+DEPEND="
+	>=dev-libs/boost-1.68.0:=[threads(+)]
+	>dev-libs/libsecp256k1-0.1_pre20200911:=[recovery,schnorr]
+	>=dev-libs/univalue-1.0.4:=
```
Flowdalic, line 27: Should we add virtual/bitcoin-leveldb (see bug [#829412](https://bugs.gentoo.org/829412))?

luke-jr: No, I confirmed 21.2 isn't affected by that.

ajakk:
Please also drop the Taproot mask, not upgrading is insecure in all scenarios. We're well past its activation now.

luke-jr:
>Please also drop the Taproot mask,

It's still unethical to do so.

> not upgrading is insecure in all scenarios.

I opened #22954 weeks ago to address this.

> We're well past its activation now.

Activation is still ambiguous at this time.

thesamesam:
> > Please also drop the Taproot mask,
> 
> It's still unethical to do so.
> 

Masking everything is not an option. Sorry.

ajakk:
Yeah, I don't think it makes much sense for users to be forced to stay on non-Taproot Bitcoin when not upgrading is insecure for them, especially now that Taproot is activated. Do you expect Bitcoin to be masked in Gentoo forever? At that point, it might as well be removed, but no one wants that.

luke-jr:
>Masking everything is not an option. Sorry.

Then we should have gone with one of the other approaches.

>Yeah, I don't think it makes much sense for users to be forced to stay on non-Taproot Bitcoin

Nobody is forcing anyone.

> especially now that Taproot is activated.

Again, this is a subjective premise. It is only as active as the users choose to enforce it, and right now, half the network is not enforcing it.

> Do you expect Bitcoin to be masked in Gentoo forever?

That's not something I ever suggested.

Regardless, all this is off-topic for this PR.

Flowdalic:
If I am not mistaken the release is labeled 0.21.2 at https://bitcoincore.org/en/releases/0.21.2/. Is there any reason we would want to label it 21.2?

luke-jr:
Bitcoin Knots is officially 21.2, and is the preferred upstream, ideal for end users (the only reason USE=knots is disabled by default in the main tree packages, is that Gentoo has as an organisation adopted trolling-decided defaults over what is best for users).

Core is only still labelled 0.21.2 because the drop-the-zero versioning change isn't backported.

ajakk:
Does it really make sense to package two apparently different upstream packages in the same package?

luke-jr:
In this case, yes. (It's not unique to this case either. Eg, Psi and Psi+ share the same package)

Flowdalic:
> Bitcoin Knots is officially 21.2, and is the preferred upstream

I am not sure if this knots being the preferred upstream is generally true. Users surely would expect to get a vanilla Bitcoin Core if they emerge bitcoin in Gentoo. Hence I believe we should follow Bitcoin Core's versioning scheme here. Please adjust the PR accordingly.

luke-jr:
>I am not sure if this knots being the preferred upstream is generally true.

It is the case for these packages.

>Users surely would expect to get a vanilla Bitcoin Core if they emerge bitcoin in Gentoo.

Why would they? There is no reason for that assumption.

>Hence I believe we should follow Bitcoin Core's versioning scheme here. Please adjust the PR accordingly.

I don't agree.

luke-jr:
One more point: Core is abandoning 21.x in a few months, so after that point, it will necessarily be Knots-only anyway.

00-matt:
> > I am not sure if this knots being the preferred upstream is generally true.
>
> It is the case for these packages.
>
> > Users surely would expect to get a vanilla Bitcoin Core if they emerge bitcoin in Gentoo.
>
> Why would they? There is no reason for that assumption.

I would expect to get something as close to the upstream source as possible, especially for something as sensitive as Bitcoin.

luke-jr:
>I would expect to get something as close to the upstream source as possible, especially for something as sensitive as Bitcoin.

Bitcoin Knots is the upstream source in this case.

ajakk:
... or is it Bitcoin core? The difference here isn't clear, because it's the same Gentoo package.

juippis:
The commit you download source from, 
`https://github.com/bitcoin/bitcoin/archive/${BITCOINCORE_COMMITHASH}.tar.gz -> bitcoin-v0.${PV}.tar.gz` 
https://github.com/bitcoin/bitcoin/commit/af591f2068d0363c92d9756ca39c43db85e5804c
literally says "[0.21] Final changes for 0.21.2". 

This is really confusing for someone not involved in the bitcoin scene and I'm sorry but we're losing trust towards you. If we can't figure out what's happening, I believe we're going to either rewrite the bitcoin ebuilds honoring vanilla upstream and drop you from maintainers, or remove them altogether from the main repo and move to any overlay who wants to offer these packages. The situation currently is way too messy with what's upstream, and these package.masks.

luke-jr:
I don't know what you find so confusing.

Bitcoin Core is not any more vanilla than Bitcoin Knots. Bitcoin Knots is the primary upstream for these ebuilds. The only reason USE=knots is unset by default, is because Gentoo decided on a troll-dictated policy some years ago.

Bitcoin Core and Knots have both dropped the leading 0 in the version, but Core did not backport it for 21.2. It makes no sense to continue using the leading zero.

I did not recommend the package.mask approach, because it was clear in foresight that it would be a mess. That approach was insisted on by someone else. Please don't blame me for a situation I didn't create.

(As for trust, that is beyond my control. I have done nothing that should compromise your or anyone else's trust in me.)

juippis:
> I don't know what you find so confusing.

First of all everyone seems to disagree with you, and lately we've had a lot of noise over these packages. Lots of gentoo-dev ML traffic, package.masks, now simple version bumps are being questioned like this... To your credit I acknowledge you're probably the most knowledgeable about bitcoin in here, but that also increases the chance we're being controlled. So I'm just trying to understand
1) Why you say the upstream version is 21 when you're pulling sources from 0.21? 
2) Why is/was the package.mask needed and what we can do to remove it? 
3) Why you say knots is the upstream here if you're pulling source from the official bitcoin GH repo? 
4) Does any other distribution package "knots"? Actually, how are the other distros approaching bitcoin in their package repositories?

 
> Bitcoin Core is not any more vanilla than Bitcoin Knots. Bitcoin Knots is the primary upstream for these ebuilds. 

Can bitcoin core work without knots? What do you mean exactly by saying "no more vanilla"? I'd trust **that** as the upstream.


> Bitcoin Core and Knots have both dropped the leading 0 in the version, but Core did not backport it for 21.2. It makes no sense to continue using the leading zero. (Edit: Also note, Bitcoin Core will be abandoning maintenance of 21.x in a few months, whereas Bitcoin Knots will be the _only_ upstream for this branch long-term)

Is there a way for us to verify this? Why we follow knots and not upstream? Should knots be separated completely, and make -core and -knots block each other? 


> I did not recommend the package.mask approach, because it was clear in foresight that it would be a mess. That approach was insisted on by someone else. Please don't blame me for a situation I didn't create.

How can we solve it now that it was done?


> (As for trust, that is beyond my control. I have done nothing that should compromise your or anyone else's trust in me.)

I am sorry for the extra concern and probably work this has caused. But you can't deny there's been a lot of noise about bitcoin lately, and I'm just trying to get on top of the situation now since nothing seems to happen one way or another.

00-matt:
> Does any other distribution package "knots"? Actually, how are the other distros approaching bitcoin in their package repositories?

Only Gentoo and NixOS seem to package Bitcoin Knots. NixOS has it as a separate `bitcoind-knots` package.

> Can bitcoin core work without knots?

Yes, most people run the upstream Bitcoin Core without any modifications. According to <https://coin.dance/nodes> there are 14830 listening Bitcoin Core nodes and 76 listening Bitcoin Knots nodes.

Some of the Bitcoin Knots patches are quite useful, like using system libraries instead of vendored libraries (with the caveat the system version needs to be bug-for-bug compatible or possibly break consensus, perhaps we need a `system-libsecp256k1` USE flag? ;)).

> Should knots be separated completely, and make -core and -knots block each other?

I think that this is the best solution. Some people want Knots so it makes sense to keep that packaged, but as a user I would be very surprised if I went to install the Bitcoin daemon and got something else.

luke-jr:
>Why you say the upstream version is 21 when you're pulling sources from 0.21?

0.21 and 21 are both the same version, just we dropped the leading zero in our version schema. The cosmetic changes were not backported in Bitcoin Core (it's not strictly a bugfix, and Core will be dropping support for the branch entirely soon), so it still displays 0.21. Bitcoin Knots correctly displays 21.2 without the deprecated/dropped zero.

>Why is/was the package.mask needed and what we can do to remove it?

To ensure users are not upgraded from pre-Taproot to post-Taproot without explicitly consenting themselves. This is important, because it is unethical for anyone other than each user to make the decision to change the Bitcoin protocol for himself. Furthermore, the risks of doing so can include financial loss for not only the user, but also for others the user does business with.

>Why you say knots is the upstream here if you're pulling source from the official bitcoin GH repo?

There is no such thing as anything official in Bitcoin. The GitHub repo `bitcoin/bitcoin` is used by Bitcoin Core for legacy reasons (new repos are under the `bitcoin-core` GitHub group to avoid this confusion), but Bitcoin Core is just one of several Bitcoin clients. Bitcoin Knots is another client (generally better for end users).

Because Bitcoin Core actively tries to prevent using system libraries, the ebuilds are designed to support building Core with Knots's "syslibs" patch, or full Knots. This is most convenient by fetching Core's tarball and applying Knots's patches, but could just as well be done by fetching Knots's tarball separately, or even fetching just Knots's tarball + patches, and reverse-applying the patches.

>Does any other distribution package "knots"? Actually, how are the other distros approaching bitcoin in their package repositories?

We discourage other distros from packaging either Bitcoin Core or Knots because it takes extra care to avoid unintentional consensus breakage. Due to Portage's flexibility, I am able to maintain it safely for Gentoo.

I also maintain the PPAs used by Ubuntu users. There are separate PPAs for Core, Knots, and Knots LTS. Upon upgrading from a pre-Taproot version to a Taproot version, they require an interactive prompt wherein the user must choose to consent or not (in which case the upgrade is aborted).

>Can bitcoin core work without knots? What do you mean exactly by saying "no more vanilla"? I'd trust that as the upstream.

Bitcoin Core and Knots are two different clients, both based on (but distinct from) the original Bitcoin software released by Satoshi (Bitcoin's original creator) in 2009-2011. One is not vanilla and the other "not vanilla"; if an analogy is to be had, Core might be strawberry and Knots raspberry.

>Is there a way for us to verify this?

The PR for dropping the leading zero is https://github.com/bitcoin/bitcoin/pull/20223

The Bitcoin Core maintenance policy is documented at https://bitcoincore.org/en/lifecycle/

>Why we follow knots and not upstream?

Knots is one upstream. Core is a second, different upstream.

>Should knots be separated completely, and make -core and -knots block each other?

It doesn't really make sense to do so. We'd end up with exponentially more packages for no real benefit. (Note the long-term goal is to split each component up, so monolithic "Bitcoin Core" and "Bitcoin Knots" packages would also not be a good approach.)

>>I did not recommend the package.mask approach, because it was clear in foresight that it would be a mess. That approach was insisted on by someone else. Please don't blame me for a situation I didn't create.

>How can we solve it now that it was done?

We could mask or drop the pre-Taproot versions for some period of time (so Portage tells users their upgrade is masked and they can see more readily they need to make a decision), and then once Taproot has met a reasonable level of objective certainty (usually I'd say around 85% of nodes enforcing it, and no known public faction acting to oppose it), someone (ideally not me) can drop the mask and take on the responsibility for any problems that creates.

Or we could deploy one of the alternate approaches (PROPERTIES=interactive only triggering when upgrading from a pre-Taproot version, like the PPA, seems like a good way IMO) and just drop the masks entirely. That'd probably be smoother and faster, but I don't know if the others are willing to go that direction.

> (with the caveat the system version needs to be bug-for-bug compatible or possibly break consensus, perhaps we need a system-libsecp256k1 USE flag? ;)).

I don't think we do. libsecp256k1 is maintained specifically for Bitcoin, so care is taken to avoid consensus breaks. IF it were to ever happen, it would be a big deal, and we could add a stricter dependency at that time (and USE flag if needed). Note the only reason we had a system-leveldb USE flag in some versions, was because using the system LevelDB had a performance impact. (The current 22.0 ebuild drops that USE flag; if we want to keep it to decouple LevelDB, let me know.)

> as a user I would be very surprised if I went to install the Bitcoin daemon and got something else.

All the packages clearly identify whether they are coming from Bitcoin Core or Knots before building. Either one is equally "the Bitcoin daemon". Regardless, these ebuilds still default to Core, as much as I think that's not the ideal default. You would only get Knots if you explicitly set USE=knots.

juippis:
Thanks for your explanation!


> 
> The PR for dropping the leading zero is [bitcoin/bitcoin#20223](https://github.com/bitcoin/bitcoin/pull/20223)
> 
> The Bitcoin Core maintenance policy is documented at https://bitcoincore.org/en/lifecycle/
> 

Feels weird for us to drop it before "upstream" does. Also feels weird it's not complete yet when the process started 2020. But I see where it's supposed to go and agree it looks fine for us. 


> 
> > How can we solve it now that it was done?
> 
> We could mask or drop the pre-Taproot versions for some period of time (so Portage tells users their upgrade is masked and they can see more readily they need to make a decision), and then once Taproot has met a reasonable level of objective certainty (usually I'd say around 85% of nodes enforcing it, and no known public faction acting to oppose it), someone (ideally not me) can drop the mask and take on the responsibility for any problems that creates.
> 
> Or we could deploy one of the alternate approaches (PROPERTIES=interactive only triggering when upgrading from a pre-Taproot version, like the PPA, seems like a good way IMO) and just drop the masks entirely. That'd probably be smoother and faster, but I don't know if the others are willing to go that direction.
> 

What about sending a news item to bitcoin users **now**, then unmask the newer versions 2-4 weeks after the news item goes live? 1) I think it'd be rather painless, doesn't affect user experience as much as potentially masking **every** possible version packaged, and we should assume bitcoin users knows already it's happening?

dev-util/bitcoin-tx/bitcoin-tx-21.2.ebuild
juippis, line 27: This package has no subslot defined so `:=` doesn't do anything here.

dev-util/bitcoin-tx/bitcoin-tx-21.2.ebuild
```diff
@@ -0,0 +1,100 @@
...
+"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~mips ~ppc ~ppc64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="knots"
+
+DEPEND="
+	>=dev-libs/boost-1.68.0:=[threads(+)]
+	>dev-libs/libsecp256k1-0.1_pre20200911:=[recovery,schnorr]
```
juippis, line 26: Same here, no subslot defined. Etc these appear in pretty much every ebuild.

juippis (approved):

PR #23410, opened 2021-12-18
dev-util/bitcoin-tx-22.0: Fix build without LevelDB

luke-jr:


luke-jr:
cc @juippis @candrews @thesamesam @Flowdalic

PR #23412, opened 2021-12-18
media-video/ffmpeg: enable VMAF scoring

AdelKS:
Hello,

This PR bring an extra dependency, `media-libs/libvmaf`, to enable `ffmpeg` to compute VMAF scores that says how good the encoding `video.mp4` of a lossless video `video-lossless.avi` is. With this PR, one can enable the `vmaf` useflag in `ffmpeg` and do the following command:
```shell
ffmpeg -i video.mp4 -i video-lossless.avi -filter_complex libvmaf -f null -
```
which works on my machine. The "lossless" video, as far as I know, doesn't really need to be lossless, but it is indeed treated as the reference video. I made a first comparison between `vaapi` and `amf` on a reference lossless video I capture from my favorite game and these are the results
```
VAAPI VMAF score: 73.41, AMF VMAF score: 80.74
```
which shows that AMF, that is pending in #23250, is better than VAAPI for at least RDNA1 GPUs (I have RX 5700 XT). In this matter, I will PR a patch for mesa that just switches on [cabac](https://en.wikipedia.org/wiki/Context-adaptive_binary_arithmetic_coding) for RDNA and RDNA2 GPUs (I don't know why it's switched off by default but according to @CME42 it has been forgotten.)

Hope this is helpful for others,

A big thanks to @CME42 who made me discover all this !

Adel


media-libs/libvmaf/libvmaf-2.3.0.ebuild
```diff
@@ -0,0 +1,53 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+MULTILIB_COMPAT=( abi_x86_{32,64} )
+
+inherit meson multilib-minimal
```
thesamesam, line 8: You can use meson-multilib.

By the way, do we actually need multilib for this? Are there 32 bit applications which need it?

media-video/ffmpeg/ffmpeg-4.4.1-r2.ebuild
```diff
@@ -338,6 +339,11 @@ src_prepare() {
 	fi
 	default
 	echo 'include $(SRC_PATH)/ffbuild/libffmpeg.mak' >> Makefile || die
+
+	if usex vmaf; then
```
thesamesam, line 343: Conditional patching should be avoided if possible as it's too hard to e.g miss when it stops applying.

also, use, not usex here if we were keeping it

media-video/ffmpeg/metadata.xml
```diff
@@ -56,6 +56,7 @@
 	<flag name="svt-av1">Enables AV1 encoding support via <pkg>media-libs/svt-av1</pkg>.</flag>
 	<flag name="twolame">Enables MP2 encoding via <pkg>media-sound/twolame</pkg> as an alternative to the internal encoder.</flag>
 	<flag name="vidstab">Enables video stabilization filter using vid.stab library (<pkg>media-libs/vidstab</pkg>).</flag>
+	<flag name="vmaf">Enables Netflix' perceptual video quality assessment library (<pkg>media-libs/libvmaf</pkg>).</flag>
```
thesamesam, line 59: Typo

media-libs/libvmaf/libvmaf-2.3.0.ebuild
AdelKS: > You can use meson-multilib.

Noted!

> By the way, do we actually need multilib for this? Are there 32 bit applications which need it?

Now that you bring it up... This entirely unrelated to x86, x86 don't need this right ? That's what I had in mind initially, but I am still learning about how to handle other arches. If it's not needed for x86 I don't think it's useful.

thesamesam: Entirely unrelated, yep! And no worries. The different platforms we learn about are part of what makes Gentoo fun :)

multilib is specifically for (mainly) where we want to build something twice to get a 32 bit library and a 64 bit one for supporting 32 bit applications on a 64 bit machine (think of stuff like Steam)

AdelKS: > multilib is specifically for (mainly) where we want to build something twice to get a 32 bit library and a 64 bit one for supporting 32 bit applications on a 64 bit machine (think of stuff like Steam)

Got it, that's what I had in mind, I just assumed it was necessary for x86 without thinking any further.

>  If it's not needed for x86 I don't think it's useful.

On the other hand, a majority of deps in ffmpeg use `${MULTILIB_USEDEP}` so I wanted to make vmaf follow suite. What do you think ?

media-video/ffmpeg/ffmpeg-4.4.1-r2.ebuild
AdelKS: the patch is harmless so it can be applied without the if condition, will fix that !

media-libs/libvmaf/libvmaf-2.3.0.ebuild
thesamesam: It's only useful if something ends up needing it, so let's leave it out for now.

licenses/BSD-2-with-patent
```diff
@@ -0,0 +1,42 @@
+LICENSE - BSD+Patent
+SPDX short identifier: BSD-2-Clause-Patent
+
+Note: This license is designed to provide: a) a simple permissive license; b) that is compatible with the GNU General
+Public License (GPL), version 2; and c) which also has an express patent grant included.
+
+Copyright (c) 2020 Netflix, Inc.
```
zeromind, line 7: This should use placeholders for the year and copyright holders.

media-libs/libvmaf/libvmaf-2.3.0.ebuild
AdelKS: Okay so it `ffmpeg` fails to emerge if `libvmaf` isn't installed in a multilib scope :/ I think it's just simpler to keep the multilib ebuild than tyring to figure out how to make `ffmpeg` work without.

media-libs/libvmaf/libvmaf-2.3.0.ebuild
```diff
@@ -0,0 +1,45 @@
...
+	EMESON_SOURCE="${S}/libvmaf"
+	meson_src_configure
+}
+
+multilib_src_install_all() {
+	einstalldocs
+
+	insinto "/usr/share/vmaf"
+	doins -r "${S}/model"
+}
```
thesamesam, line 45: fix missing newline

media-libs/libvmaf/libvmaf-2.3.0.ebuild
```diff
@@ -0,0 +1,45 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+MULTILIB_COMPAT=( abi_x86_{32,64} )
```
thesamesam, line 6: ... is this actually needed?

media-libs/libvmaf/libvmaf-2.3.0.ebuild
```diff
@@ -0,0 +1,45 @@
...
+
+inherit meson-multilib
+
+DESCRIPTION="C libary for Netflix's Perceptual video quality assessment based on multi-method fusion."
+HOMEPAGE="https://github.com/Netflix/vmaf"
+
+if [[ ${PV} == "9999" ]]; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/Netflix/vmaf.git"
+	SRC_URI=""
```
thesamesam, line 16:
```suggestion
```

media-libs/libvmaf/libvmaf-2.3.0.ebuild
AdelKS: Okay, so I removed that and simply inherited `meson` eclass, basically [this version](https://github.com/gentoo/gentoo/blob/85931e308c9fa4ce7aaca166cf41c29863fda040/media-libs/libvmaf/libvmaf-2.3.0.ebuild). And my `ffmpeg` that is built with multilib support fails.

Or are you saying I can still inherit `meson-multilib` and not set `MULTILIB_COMPAT=( abi_x86_{32,64} )` ?

AdelKS: Yeah now that you mention it, it makes sense so other arches aren't just blocked because of this arbitrary restriction. done!

media-video/ffmpeg/files/vmaf-models-default-path.patch
```diff
@@ -0,0 +1,13 @@
+# Put default path of models to /usr/share/vmaf/model
```
laumann, line 1: I don't think the leading `#` is necessary here, but I don't know if it's convention to add comments to patches like this.

AdelKS: It's true that it's rather obvious what the patch does, do you suggest to remove the comment altogether ?

laumann: No, just the `#` prefix. I love comments :-)

AdelKS: Got it, done!

AdelKS:
Done, thanks!

jospezial:
```
>>> Preparing source in /var/tmp/portage/media-video/ffmpeg-9999/work/ffmpeg-9999 ...
 * Applying vmaf-models-default-path.patch ...
patching file libavfilter/vf_libvmaf.c
Hunk #1 FAILED at 72.
1 out of 1 hunk FAILED -- saving rejects to file libavfilter/vf_libvmaf.c.rej
 [ !! ]
 * ERROR: media-video/ffmpeg-9999::gentoo failed (prepare phase):
 *   patch -p1  failed with /var/tmp/portage/media-video/ffmpeg-9999/files/vmaf-models-default-path.patch
 * 
 * Call stack:
 *               ebuild.sh, line  127:  Called src_prepare
 *             environment, line 3113:  Called eapply '/var/tmp/portage/media-video/ffmpeg-9999/files/vmaf-models-default-path.patch'
 *             environment, line 1281:  Called _eapply_patch '/var/tmp/portage/media-video/ffmpeg-9999/files/vmaf-models-default-path.patch'
 *             environment, line 1219:  Called __helpers_die 'patch -p1  failed with /var/tmp/portage/media-video/ffmpeg-9999/files/vmaf-models-default-path.patch'
 *   isolated-functions.sh, line  112:  Called die
 *              die "$@"
```
https://github.com/FFmpeg/FFmpeg/commit/3d29724c008d8f27fecf85757152789b074e8ef9

thesamesam:
> ```
> >>> Preparing source in /var/tmp/portage/media-video/ffmpeg-9999/work/ffmpeg-9999 ...
>  * Applying vmaf-models-default-path.patch ...
> patching file libavfilter/vf_libvmaf.c
> Hunk #1 FAILED at 72.
> 1 out of 1 hunk FAILED -- saving rejects to file libavfilter/vf_libvmaf.c.rej
>  [ !! ]
>  * ERROR: media-video/ffmpeg-9999::gentoo failed (prepare phase):
>  *   patch -p1  failed with /var/tmp/portage/media-video/ffmpeg-9999/files/vmaf-models-default-path.patch
>  * 
>  * Call stack:
>  *               ebuild.sh, line  127:  Called src_prepare
>  *             environment, line 3113:  Called eapply '/var/tmp/portage/media-video/ffmpeg-9999/files/vmaf-models-default-path.patch'
>  *             environment, line 1281:  Called _eapply_patch '/var/tmp/portage/media-video/ffmpeg-9999/files/vmaf-models-default-path.patch'
>  *             environment, line 1219:  Called __helpers_die 'patch -p1  failed with /var/tmp/portage/media-video/ffmpeg-9999/files/vmaf-models-default-path.patch'
>  *   isolated-functions.sh, line  112:  Called die
>  * The specific snippet of code:
>  *              die "$@"
> ```
> 
> 
>     
>  
>   
> 
> [FFmpeg/FFmpeg@3d29724](https://github.com/FFmpeg/FFmpeg/commit/3d29724c008d8f27fecf85757152789b074e8ef9)

Please file a bug

AdelKS:
@jospezial thanks for the feedback. I disabled the application of the patch in the live ebuild, it seems that now the full path for the wanted model is mandatory, see the related commit here: https://github.com/gentoo/gentoo/pull/24621

PR #23415, opened 2021-12-19
www-client/google-chrome: add optional supports for emoji

scardracs:

### DIFF
--- a/www-client/google-chrome-beta/google-chrome-beta-97.0.4692.56-r1.ebuild
+++ b/www-client/google-chrome-beta/google-chrome-beta-97.0.4692.56-r1.ebuild
@@ -1,0 +1,118 @@
+# Copyright 2011-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+CHROMIUM_LANGS="am ar bg bn ca cs da de el en-GB es es-419 et fa fi fil fr gu he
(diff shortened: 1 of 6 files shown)


stha09:
Optional runtime dependencies are not allowed in ::gentoo, see https://projects.gentoo.org/qa/policy-guide/dependencies.html

Flowdalic:
If you add such an 'emoji' use flag to chrome, then, as consequence, you would need it to most GUI applications. This is clearly not viable. In general, I believe, Gentoo users are trusted to know which font packages to install to get the best emoji support. Even though, we could probably do a better job documenting it. That said, and keep in mind that it is the decision of the package maintainer, I wonder if `optfeature media-fonts/noto-emoji` in the chrome ebuild wouldn't be a bad idea.

scardracs:
> If you add such an 'emoji' use flag to chrome, then, as consequence, you would need it to most GUI applications. This is clearly not viable. In general, I believe, Gentoo users are trusted to know which font packages to install to get the best emoji support. Even though, we could probably do a better job documenting it. That said, and keep in mind that it is the decision of the package maintainer, I wonder if `optfeature media-fonts/noto-emoji` in the chrome ebuild wouldn't be a bad idea.

Yeah, that would be a better solution. I was adding it as I saw there are some packages that already have that IUSE

floppym:
We won't add a USE flag for this.

We used to output a message about installing various fonts, but I never liked it and removed it in 17a656b0a47a95521b49c2ed968e5a7cff508e93. I don't think it makes sense to output a message for a web browser when fonts affect every GUI app on the system.

PR #23423, opened 2021-12-19
dev-util/bear: bump to v3.0.17

dAgbeve:
Signed-off-by: Douglas Agbeve <douglas@agbeve.com>


gyakovlev:
thanks!
I was postponing bump because of test failure
https://github.com/rizsotto/Bear/issues/435

I've also added gcc-11 patch and libtool dependency in follow-up commit: https://github.com/gentoo/gentoo/commit/463aa8be2d7639642d95f825b52da4ca6badb6db

PR #23425, opened 2021-12-19
Cuda cleanup

SoapGentoo:


ionenwks (approved):
old tensorflow has `|| ( =toolkit-10* =toolkit-11.4* )` which is unsolvable, why is CI not picking this up, am I overlooking something?

net-misc/bfgminer has a nonsense dependency on nvidia-cuda-sdk[opencl], that one is solvable with its `|| ( )` but well...

these aside, doing this lgtm

scardracs:
> 

bfgminer requires both opencl and cuda to mine at max speed on nvidia cards iirc (https://github.com/luke-jr/bfgminer/blob/bfgminer/README.GPU#L229)

scardracs:
> old tensorflow has `|| ( =toolkit-10* =toolkit-11.4* )` which is unsolvable, why is CI not picking this up, am I overlooking something?
> 
> net-misc/bfgminer has a nonsense dependency on nvidia-cuda-sdk[opencl], that one is solvable with its `|| ( )` but well...
> 
> these aside, doing this lgtm

bfgminer requires both opencl and cuda to mine at max speed on nvidia cards iirc (https://github.com/luke-jr/bfgminer/blob/bfgminer/README.GPU#L229)

profiles/package.mask
```diff
@@ -33,6 +33,13 @@
 
 #--- END OF EXAMPLES ---
 
+# David Seifert <soap@gentoo.org> (2021-12-20)
+# Requires GCC 8, security vulnerabilities, switch to CUDA 11,
+# consider using clang+NVPTX backend for sm_30 cards.
+# Removal in 90 days, bug #721808, #760708.
+<dev-util/nvidia-cuda-sdk-11
+<dev-util/nvidia-cuda-toolkit-11
```
amadio, line 41: Please don't drop this yet, at least not until there's a fix for https://bugs.gentoo.org/823882. I needed to downgrade CUDA to 10.2 on a machine recently because of that.

SoapGentoo: maybe help fix it?

ionenwks:
> > old tensorflow has `|| ( =toolkit-10* =toolkit-11.4* )` which is unsolvable, why is CI not picking this up, am I overlooking something?
> > net-misc/bfgminer has a nonsense dependency on nvidia-cuda-sdk[opencl], that one is solvable with its `|| ( )` but well...
> > these aside, doing this lgtm
> 
> bfgminer requires both opencl and cuda to mine at max speed on nvidia cards iirc (https://github.com/luke-jr/bfgminer/blob/bfgminer/README.GPU#L229)

The dependency still doesn't make sense to me though. This is asking for one or the other, not both. And runtime-only sounds odd for cuda, libcuda is provided by nvidia-drivers and would be always available on an nvidia system either way. Not that I looked closely at this package.

profiles/package.mask
amadio: I did try to fix it, and didn't get anywhere. This problem with cuda-gdb is something that likely needs to be fixed by nvidia.

SoapGentoo: have you reported it like all of the several bugs I have reported in nvcc?

amadio: Didn't have time last year. I will try some debugging again now that I'm back from vacations, and file a bug if I can't fix it. Still, just keeping 10.2 as is for a little while longer at least gives people that need to debug CUDA code a way to do so.

amadio: Filed a bug and in 11.6.1, which I just pushed, cuda-gdb works again.

SoapGentoo: so we can finally sunset CUDA 10?

amadio: There's still a small issue with CUDA and some headers of GCC, but there's an easy workaround for that. I think we can drop CUDA 10 after a couple of weeks this new version is in the tree.

SoapGentoo: which GCC header issues? The `<functional>` stuff? That has been fixed in 11.6.1 (after I reported it)

amadio: No, `serializeintrin.h`.

PR #23430, opened 2021-12-20
net-misc/seafile{,-cilent}: version bump to 8.0.5

CaseOf:


ionenwks (commented):
Seems this is based on older ebuilds, missing the changes from f77c7b33723a0c934f0a7130d031935679d24f5e and 129e57fe7ac9e10d850ce62ce9e61b646ea441ac

May be good to cleanup some old versions too? Unless some reason to keep.

net-misc/seafile/seafile-8.0.5.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	eautoreconf
+	vala_src_prepare
+}
+
+src_configure() {
+	local myeconfargs=(
+		--disable-static
+	)
+	econf "${myeconfargs[@]}"
+}
```
ionenwks, line 50: Also, if update ebuild to EAPI-8, this can be dropped given --disable-static is passed by default then.

ionenwks: Or nevermind this bit, forgot vala.eclass isn't EAPI-8 yet.

CaseOf:
I dont know if it is allowed to add this or not, but I did work on a firejail profile https://gist.github.com/CaseOf/2f12ae0f19c35f6a5a048b49492d13f0

ionenwks:
Thanks, and sorry PR took so long to be looked at.

> I dont know if it is allowed to add this or not, but I did work on a firejail profile https://gist.github.com/CaseOf/2f12ae0f19c35f6a5a048b49492d13f0

Don't have a strong opinion here, although Gentoo isn't really pushing official firejail support. It'd make more sense if the file was coming from upstream rather than add things potential future maintainers won't know how to update/test.

PR #23431, opened 2021-12-20
net-fs/samba: Add AES acceleration is cpu_flags_x86_aes set

FireBurn:
This will enable AES acceleration if the cpu_flags_x86_aes USE flag is
set, otherwise "none" is passed

Bug: https://bugs.gentoo.org/821349

### DIFF
--- a/net-fs/samba/samba-4.15.3-r1.ebuild
+++ b/net-fs/samba/samba-4.15.3-r1.ebuild
@@ -22,8 +22,8 @@ S="${WORKDIR}/${MY_P}"
 
 LICENSE="GPL-3"
 SLOT="0"
-IUSE="acl addc ads ceph client cluster cups debug dmapi fam glusterfs
-gpg iprint json ldap pam profiling-data python quota +regedit selinux
+IUSE="acl addc ads ceph client cluster cpu_flags_x86_aes cups debug dmapi fam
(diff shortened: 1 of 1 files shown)


net-fs/samba/samba-4.15.3.ebuild
```diff
@@ -212,6 +212,7 @@ multilib_src_configure() {
 		$(multilib_native_use_with ads)
 		$(multilib_native_use_enable ceph cephfs)
 		$(multilib_native_use_with cluster cluster-support)
+		$(usex cpu_flags_x86_aes "--accel-aes=intelaesni" "--accel-aes=none")
```
thesamesam, line 215: Wonder if --accel-aces=$(usex cpu_flags_aes intelaesni none) would be nicer here, maybe.

Also, I'd use Closes: instead in the commit message.

thesamesam (commented):
Seems reasonable. @Polynomial-C

net-fs/samba/samba-4.15.3.ebuild
FireBurn: Where in myconf should that line go, to keep things tidy?

mattst88: No preference from me.

mattst88:
@Polynomial-C: I'm going to merge this if I don't hear from you.

PR #23435, opened 2021-12-20
net-misc/xmrig: Require newer hwloc

00-matt:
Since XMRig 6.16.0 with the GhostRider algo enabled, a newer version
of hwloc is required.

Closes: https://bugs.gentoo.org/828093
Signed-off-by: Matthew Smith <matt@offtopica.uk>


candrews:
Thank you!

PR #23437, opened 2021-12-20
www-apache/mod_security+crs: security bump

hydrapolic:
Tested on amd64, thanks!


hydrapolic:
Thank you @thesamesam

PR #23438, opened 2021-12-20
Add transitive dependencies for dev-java/log4j-core-2.17.0

Leo3418:
The commits in this pull request provide transitive dependencies of `dev-java/log4j-core-2.17.0`.  The dependency tree looks like:
```
dev-java/log4j-core-2.17.0
    dev-java/kafka-clients
        dev-java/lz4-java
            dev-java/mvel
        dev-java/snappy
            dev-libs/bitshuffle
```

Status of tests:
- `dev-java/mvel` tests can be run properly in `src_test`
- `dev-java/lz4-java` has only one missing direct test dependency [`com.carrotsearch.randomizedtesting:junit4-ant`](https://mvnrepository.com/artifact/com.carrotsearch.randomizedtesting/junit4-ant/2.7.8); `src_test` of this ebuild is runnable if the dependency is satisfied
- `dev-java/snappy` has only one missing direct test dependency [`org.apache.hadoop:hadoop-common:2.7.3`](https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-common/2.7.3); `src_test` of this ebuild is runnable if the dependency is satisfied
- `dev-java/kafka-clients` has multiple missing test dependencies; `src_test` is not implemented
- `dev-libs/bitshuffle` upstream does not provide any tests


vaukai:
@Leo3418 If you could start these 5 ebuilds directly with KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~86" it would save us some time for keywording.

Leo3418:
Converting to draft for keywording decision and potential EAPI 8 bump for all ebuilds using `java-ant-2.eclass` (https://archives.gentoo.org/gentoo-dev/message/388f20e7ac63330fa69f9325693f8e8e)

Leo3418:
@vaukai `dev-java/lz4-java` and `dev-java/kafka-clients` cannot be keyworded for `~arm`, `~arm64` or `~sparc` right now unless `dev-java/cpptasks:0` get those respective keywords. Except that `arm`, `sparc` and `x86` arches are `~arch` keyworded without proper testing, all ebuilds are keyworded to the maximum possible extent... and use EAPI 8 of course

thesamesam:
Why not do keyworidng the normal way? We want to ensure things are tested thoroughly.

Leo3418:
> Why not do keyworidng the normal way? We want to ensure things are tested thoroughly.

Tbh I'm inclined to do it the normal way too... @vaukai I'll take your words on this since my work in this PR is just the tip of the iceberg and will eventually be integrated with your work, which is the major part

dev-libs/bitshuffle/files/bitshuffle-0.3.5-Makefile
```diff
@@ -0,0 +1,33 @@
...
+LIBDIR = $(PREFIX)/lib
+INCLUDEDIR = $(PREFIX)/include
+
+install: libbitshuffle.so
+	$(INSTALL) -dm755 $(DESTDIR)$(LIBDIR)
+	$(INSTALL) -dm755 $(DESTDIR)$(INCLUDEDIR)
+	$(INSTALL) -m755 libbitshuffle.so $(DESTDIR)$(LIBDIR)
+	$(INSTALL) -m644 src/bitshuffle.h $(DESTDIR)$(INCLUDEDIR)
+	$(INSTALL) -m644 src/bitshuffle_core.h $(DESTDIR)$(INCLUDEDIR)
+	$(INSTALL) -m644 src/iochain.h $(DESTDIR)$(INCLUDEDIR)
```
fordfrog, line 33: i'm not really sure it's ok to install so file that does not contain version number

dev-java/mvel/files/mvel-2.3.2-fix-for-java-10+.patch
```diff
@@ -0,0 +1,24 @@
...
+--- a/src/main/java/org/mvel2/optimizers/impl/asm/ASMAccessorOptimizer.java
++++ b/src/main/java/org/mvel2/optimizers/impl/asm/ASMAccessorOptimizer.java
+@@ -121,7 +121,9 @@
+     } else if (javaVersion.startsWith("1.6")
+             || javaVersion.startsWith("1.7")
+             || javaVersion.startsWith("1.8")
+-            || javaVersion.startsWith("9")) {
++            || javaVersion.startsWith("9")
++            || javaVersion.startsWith("10")
++            || javaVersion.startsWith("11")) {
```
fordfrog, line 21: i'd do it the way [they have it in the latest code](https://github.com/mvel/mvel/blob/master/src/main/java/org/mvel2/optimizers/impl/asm/ASMAccessorOptimizer.java#L188-L194), that's more robust, this code ignores all future versions of java

dev-java/mvel/files/mvel-2.3.2-ignore-failing.tests.patch
```diff
@@ -0,0 +1,36 @@
...
+diff --git a/src/test/java/org/mvel2/tests/core/CoreConfidenceTests.java b/src/test/java/org/mvel2/tests/core/CoreConfidenceTests.java
+index ea7661bb..1c92c327 100644
+--- a/src/test/java/org/mvel2/tests/core/CoreConfidenceTests.java
++++ b/src/test/java/org/mvel2/tests/core/CoreConfidenceTests.java
+@@ -3431,7 +3431,7 @@ public class CoreConfidenceTests extends AbstractTest {
+     assertTrue(result);
+   }
+ 
+-  public void testMapAccessWithNestedMethodCall() {
++  public void noTestMapAccessWithNestedMethodCall() {
```
fordfrog, line 21: though it is just a matter of style, `@Ignore` annotation is the right way to ignore tests in `junit-4`.

Leo3418: That does not seem to work in this particular case:
```diff
diff --git a/src/test/java/org/mvel2/tests/core/CoreConfidenceTests.java b/src/test/java/org/mvel2/tests/core/CoreConfidenceTests.java
index ea7661bb..1c92c327 100644
--- a/src/test/java/org/mvel2/tests/core/CoreConfidenceTests.java
+++ b/src/test/java/org/mvel2/tests/core/CoreConfidenceTests.java
@@ -3431,7 +3431,8 @@ public class CoreConfidenceTests extends AbstractTest {
     assertTrue(result);
   }
 
+  @org.junit.Ignore
   public void testMapAccessWithNestedMethodCall() {
     String str = "map[aMethod(1)] == \"one\"";
 
     ParserConfiguration pconf = new ParserConfiguration();
@@ -3446,7 +3447,8 @@ public class CoreConfidenceTests extends AbstractTest {
     assertTrue(result);
   }
 
+  @org.junit.Ignore
   public void testMapAccessWithNestedProperty() {
     String str = "map[key] == \"one\"";
 
     ParserConfiguration pconf = new ParserConfiguration();
```

<details><summary>JUnit log</summary>

```
There were 2 failures:
1) testMapAccessWithNestedMethodCall(org.mvel2.tests.core.CoreConfidenceTests)
[Error: unable to resolve method: java.util.HashMap.aMethod() [arglength=1]]
[Near : {... aMethod(1) ....}]
                    ^
[Line: 1, Column: 8]
	at org.mvel2.optimizers.impl.asm.ASMAccessorOptimizer.getMethod(ASMAccessorOptimizer.java:1975)
	at org.mvel2.optimizers.impl.asm.ASMAccessorOptimizer.compileAccessor(ASMAccessorOptimizer.java:845)
	at org.mvel2.optimizers.impl.asm.ASMAccessorOptimizer.optimizeAccessor(ASMAccessorOptimizer.java:281)
	at org.mvel2.ast.ASTNode.optimize(ASTNode.java:159)
	at org.mvel2.ast.ASTNode.getReducedValueAccelerated(ASTNode.java:115)
	at org.mvel2.compiler.ExecutableAccessor.getValue(ExecutableAccessor.java:42)
	at org.mvel2.optimizers.impl.asm.ASMAccessorOptimizer.getCollectionPropertyAO(ASMAccessorOptimizer.java:1550)
	at org.mvel2.optimizers.impl.asm.ASMAccessorOptimizer.compileAccessor(ASMAccessorOptimizer.java:848)
	at org.mvel2.optimizers.impl.asm.ASMAccessorOptimizer.optimizeAccessor(ASMAccessorOptimizer.java:281)
	at org.mvel2.ast.ASTNode.optimize(ASTNode.java:159)
	at org.mvel2.ast.ASTNode.getReducedValueAccelerated(ASTNode.java:115)
	at org.mvel2.ast.BinaryOperation.getReducedValueAccelerated(BinaryOperation.java:117)
	at org.mvel2.compiler.ExecutableAccessor.getValue(ExecutableAccessor.java:42)
	at org.mvel2.MVEL.executeExpression(MVEL.java:953)
	at org.mvel2.tests.core.CoreConfidenceTests.testMapAccessWithNestedMethodCall(CoreConfidenceTests.java:3446)
2) testMapAccessWithNestedProperty(org.mvel2.tests.core.CoreConfidenceTests)
[Error: unable to resolve method: org.mvel2.tests.core.CoreConfidenceTests$POJO.key() [arglength=0]]
[Near : {... key ....}]
             ^
[Line: 1, Column: 1]
	at org.mvel2.optimizers.impl.refl.ReflectiveAccessorOptimizer.getMethod(ReflectiveAccessorOptimizer.java:1095)
	at org.mvel2.optimizers.impl.refl.ReflectiveAccessorOptimizer.getMethod(ReflectiveAccessorOptimizer.java:1086)
	at org.mvel2.optimizers.impl.refl.ReflectiveAccessorOptimizer.getMethod(ReflectiveAccessorOptimizer.java:1003)
	at org.mvel2.optimizers.impl.refl.ReflectiveAccessorOptimizer.getBeanProperty(ReflectiveAccessorOptimizer.java:693)
	at org.mvel2.optimizers.impl.refl.ReflectiveAccessorOptimizer.getBeanPropertyAO(ReflectiveAccessorOptimizer.java:488)
	at org.mvel2.optimizers.impl.refl.ReflectiveAccessorOptimizer.compileGetChain(ReflectiveAccessorOptimizer.java:393)
	at org.mvel2.optimizers.impl.refl.ReflectiveAccessorOptimizer.optimizeAccessor(ReflectiveAccessorOptimizer.java:163)
	at org.mvel2.ast.ASTNode.optimize(ASTNode.java:159)
	at org.mvel2.ast.ASTNode.getReducedValueAccelerated(ASTNode.java:115)
	at org.mvel2.compiler.ExecutableAccessor.getValue(ExecutableAccessor.java:38)
	at org.mvel2.optimizers.impl.refl.ReflectiveAccessorOptimizer.getCollectionPropertyAO(ReflectiveAccessorOptimizer.java:860)
	at org.mvel2.optimizers.impl.refl.ReflectiveAccessorOptimizer.compileGetChain(ReflectiveAccessorOptimizer.java:399)
	at org.mvel2.optimizers.impl.refl.ReflectiveAccessorOptimizer.optimizeAccessor(ReflectiveAccessorOptimizer.java:163)
	at org.mvel2.ast.ASTNode.optimize(ASTNode.java:159)
	at org.mvel2.ast.ASTNode.getReducedValueAccelerated(ASTNode.java:115)
	at org.mvel2.ast.BinaryOperation.getReducedValueAccelerated(BinaryOperation.java:117)
	at org.mvel2.compiler.ExecutableAccessor.getValue(ExecutableAccessor.java:42)
	at org.mvel2.MVEL.executeExpression(MVEL.java:953)
	at org.mvel2.tests.core.CoreConfidenceTests.testMapAccessWithNestedProperty(CoreConfidenceTests.java:3462)

FAILURES!!!
Tests run: 1102,  Failures: 2
```
</details>

dev-java/mvel/files/mvel-2.3.2-fix-for-java-10+.patch
Leo3418: Resolved as in 6b55375055d430d537ea183322e0d19d7b331b91

dev-libs/bitshuffle/files/bitshuffle-0.3.5-Makefile
Leo3418: There is not a versioned soname provided by Bitshuffle upstream as they have never provided a Makefile. What's the best thing to do for this situation?

dev-java/mvel/files/mvel-2.3.2-ignore-failing.tests.patch
Leo3418: Resolved as in 8d219d4ab1eea10a33c2ebe73de9653bac3b62bf

dev-libs/bitshuffle/files/bitshuffle-0.3.5-Makefile
fordfrog: i'd follow the standard versioning scheme. otherwise we'd get this returned back by the qa team i guess :-)

fordfrog:
could you please squash the `mvel` commits into one?

dev-libs/bitshuffle/files/bitshuffle-0.3.5-Makefile
Leo3418: Kind of got lost in my research in soname versioning, tried my best to make bfc1017aeacc7f08e090e32b442d454f7ef6181e. Suggestions welcomed

fordfrog: thank you! just please squash the commits again :-)

fordfrog:
i just gave a test to that `dev-java/mvel` and it looks fine, except i get a test failure with `openjdk-17` (does not occur with previous jdks):
```
There was 1 failure:
1) testMVEL308(org.mvel2.tests.core.PropertyAccessTests)
[Error: unknown exception in expression: field.key]
[Near : {... field.key ....}]
                      ^
[Line: 1, Column: 10]
	at org.mvel2.PropertyAccessor.get(PropertyAccessor.java:170)
	at org.mvel2.PropertyAccessor.get(PropertyAccessor.java:125)
	at org.mvel2.ast.ASTNode.getReducedValue(ASTNode.java:187)
	at org.mvel2.MVELInterpretedRuntime.parseAndExecuteInterpreted(MVELInterpretedRuntime.java:101)
	at org.mvel2.MVELInterpretedRuntime.parse(MVELInterpretedRuntime.java:47)
	at org.mvel2.MVEL.eval(MVEL.java:150)
	at org.mvel2.optimizers.impl.refl.ReflectiveAccessorOptimizer.optimizeSetAccessor(ReflectiveAccessorOptimizer.java:229)
	at org.mvel2.optimizers.dynamic.DynamicOptimizer.optimizeSetAccessor(DynamicOptimizer.java:96)
	at org.mvel2.compiler.CompiledAccExpression.setValue(CompiledAccExpression.java:55)
	at org.mvel2.ast.DeepAssignmentNode.getReducedValueAccelerated(DeepAssignmentNode.java:90)
	at org.mvel2.compiler.ExecutableAccessor.getValue(ExecutableAccessor.java:38)
	at org.mvel2.ast.ForEachNode.getReducedValueAccelerated(ForEachNode.java:111)
	at org.mvel2.compiler.ExecutableAccessor.getValue(ExecutableAccessor.java:42)
	at org.mvel2.MVEL.executeExpression(MVEL.java:968)
	at org.mvel2.tests.core.PropertyAccessTests.testMVEL308(PropertyAccessTests.java:489)
	... 25 trimmed
Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make public final java.lang.Object java.util.HashMap$Node.getKey() accessible: module java.base does not "opens java.util" to unnamed module @7358af75
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
	at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:199)
	at java.base/java.lang.reflect.Method.setAccessible(Method.java:193)
	at org.mvel2.PropertyAccessor.getBeanProperty(PropertyAccessor.java:601)
	at org.mvel2.PropertyAccessor.getNormal(PropertyAccessor.java:178)
	at org.mvel2.PropertyAccessor.get(PropertyAccessor.java:145)
	... 40 more

FAILURES!!!
Tests run: 1100,  Failures: 1
```
according to the backtrace ("Unable to...") it's really related just to `openjdk-17`. this might cause issues when using `openjdk-17` in the future. anyway, i suggest if you could check the upstream whether they already fixed that in their master and if so, try to backport it, or just simply disable this test for `openjdk-17` for now and in the meantime it might get fixed by the upstream.

Leo3418:
> according to the backtrace ("Unable to...") it's really related just to `openjdk-17`. this might cause issues when using `openjdk-17` in the future. anyway, i suggest if you could check the upstream whether they already fixed that in their master and if so, try to backport it, or just simply disable this test for `openjdk-17` for now and in the meantime it might get fixed by the upstream.

I tested the latest version (2.4.14) against `openjdk-bin-17`, ran `mvn test` directly within the Maven project, and got the same error, so it is still not fixed yet. Do you think df4e7bdcaec7e69ec438e08b709e6c75095538b8 is a feasible solution?

dev-libs/bitshuffle/files/bitshuffle-0.3.5-Makefile
Leo3418: Squashed!

dev-java/lz4-java/lz4-java-1.8.0.ebuild
```diff
@@ -0,0 +1,77 @@
...
+CDEPEND="
+	app-arch/lz4
+"
+
+# The version requirement on mvel is strict; mvel-2.4.* generates source files
+# incorrectly, which causes build failure of this package
+DEPEND="
+	>=virtual/jdk-1.8:*
+	dev-java/cpptasks:0
+	~dev-java/mvel-2.3.2:0
```
fordfrog, line 32: because of this i changed `dev-java/mvel-2.3.2` to slot `2.3`. otherwise this package would block usage of newer versions from other packages, once they use `mvel` as well.

fordfrog:
i just merged `dev-libs/bitshuffle` and `dev-java/mvel`. please note the change of slot in `dev-java/mvel`.

fordfrog:
wrt `dev-java/lz4-java`, it should respecf `CFLAGS`:
```
 * /usr/lib64/lz4-java/liblz4-java.so
```
also, again, it does not install versioned `.so` file so that has to be fixed too.

fordfrog:
also the same applies to `dev-java/snappy` (`CFLAGS` and `.so` version).

Leo3418:
@fordfrog In case you missed it in `#gentoo-java`:
```
12:14 <+Leo3418> fordfrog sam_: I continued to research on JNI sonaming a
little bit. The only case where a correct soname is needed for any JNI library
is on Android NDK, which isn't relevant for us, not even on Gentoo Android. I
brought this question to heroxbd too. Although he doesn't know a lot about JNI,
he thought that if the JNI library would never be probed by linkers then an
soname shouldn't be necessary

12:58 <+Leo3418> In addition, I haven't seen any Java package installing header
files yet. Would it be possible for one to use the interface provided by a
shared library without having the headers declaring the symbols in it?

13:12 <[Arfrever]> You can declare things locally and not include headers...

13:12 <+Leo3418> Certainly, but I would consider that as use of unexported
private API, which I wouldn't support if I were the package vendor
```
```
19:08 <+sam_> Leo3418: I think that sounds reasonable
```

My reasoning on `soname` for JNI shared libraries is:
1. If a package intends to provide its shared libraries to other programs, it should install header files that declare the symbols in the public API it exports
2. Any use of API not declared in the header files should be regarded as **unsupported** use of private/internal API
3. `dev-java/snappy` and `dev-java/lz4-java` do not install any header files, even if the package is built and installed by using upstream's build instructions and invoking Make/Ant/Maven/Gradle outside Portage
4. Based on 2 and 3, because Java packages involving JNI do not install header files, they should not be considered to export any public API
5. If no public API is exported by a shared library, it is not expected to be linked or loaded with any other programs other than the JVM
6. JVM does not use `soname` for loading JNI shared libraries; instead, it finds shared libraries that need to be loaded by searching for the file with the matched **file name** under paths set in the `java.library.path` property. References from the JDK source code:
   - [`Runtime.loadLibrary(String)` that is provided to Java applications as a means to load a library](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/lang/Runtime.java#L815-L830)
   - [JDK internal method that contains code for probing the library's object using its file name](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/lang/ClassLoader.java#L2607-L2661)
   - [implementation of the `System.mapLibraryName(String)` method that converts a library name to its object's file name](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/native/libjava/System.c#L445-L469) (e.g. [`System.mapLibraryName(foo)` returns `libfoo.so` on GNU/Linux](https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/hotspot/os/posix/include/jvm_md.h#L47-L55))
7. Based on 6, `soname` is not relevant in the context of JVM, so it is only useful for linking and loading outside JVM. Based on 5, such linking and loading are never intended to happen, so `soname` does not matter at all. I discussed this issue with @heroxbd and he also agreed with this

heroxbd:
> @Leo3418 If you could start these 5 ebuilds directly with KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86" it would save us some time for keywording.

@Leo3418 please don't.  You should not add the keywords that you have not tested.
Reference: https://devmanual.gentoo.org/ebuild-writing/common-mistakes/#untested-archs-in-keywords

heroxbd:
I am new to JNI.  AFAIK, @Leo3418's [argument](https://github.com/gentoo/gentoo/pull/23438#issuecomment-1024294572) on headers and so libraries make sense.  I think it is the best we can do to deal with missing soname's in JNI.

I suggest we take the QA warning for now.  If not, we should involve the QA team to figure out a rule for JNI so libraries.

fordfrog:
just tested `dev-java/snappy`. though it has java 1.8 set as the min version, i get this warning:
```
"$JAVA_HOME/bin/javac" -source 1.7 -target 1.7 -h src/main/java/org/xerial/snappy/ -d target/jni-classes -sourcepath src/main/java src/main/java/org/xerial/snappy/SnappyNative.java
"$JAVA_HOME/bin/javac" -source 1.7 -target 1.7 -h src/main/java/org/xerial/snappy/ -d target/jni-classes -sourcepath src/main/java src/main/java/org/xerial/snappy/BitShuffleNative.java
warning: [options] bootstrap class path not set in conjunction with -source 7
warning: [options] source value 7 is obsolete and will be removed in a future release
warning: [options] target value 7 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
warning: [options] bootstrap class path not set in conjunction with -source 7
warning: [options] source value 7 is obsolete and will be removed in a future release
warning: [options] target value 7 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
```
our ebuild should respect our min java version and not used the one hardcoded by upstream.

fordfrog:
@Leo3418 `dev-java/lz4-java` merged, thanks!

Leo3418:
> just tested `dev-java/snappy`. though it has java 1.8 set as the min version, i get this warning:
> 
> ```
> "$JAVA_HOME/bin/javac" -source 1.7 -target 1.7 -h src/main/java/org/xerial/snappy/ -d target/jni-classes -sourcepath src/main/java src/main/java/org/xerial/snappy/SnappyNative.java
> "$JAVA_HOME/bin/javac" -source 1.7 -target 1.7 -h src/main/java/org/xerial/snappy/ -d target/jni-classes -sourcepath src/main/java src/main/java/org/xerial/snappy/BitShuffleNative.java
> warning: [options] bootstrap class path not set in conjunction with -source 7
> warning: [options] source value 7 is obsolete and will be removed in a future release
> warning: [options] target value 7 is obsolete and will be removed in a future release
> warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
> warning: [options] bootstrap class path not set in conjunction with -source 7
> warning: [options] source value 7 is obsolete and will be removed in a future release
> warning: [options] target value 7 is obsolete and will be removed in a future release
> warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
> ```
> 
> our ebuild should respect our min java version and not used the one hardcoded by upstream.

Done:
https://github.com/gentoo/gentoo/pull/23438/files#diff-9a7d1daf0716f3ae2a6867fa9d8c4ee517d68a1ae0bb5a758d26fac5c5d46aebR34-R38
https://github.com/gentoo/gentoo/pull/23438/files#diff-7d94b6fb504fd60d1188fc319df21ebeab90175dc353b6fa623ef7abe3bd510c

fordfrog:
@Leo3418 merged, thanks!

Leo3418:
> @Leo3418 merged, thanks!

Thanks for your tireless review, patience and suggestions too!

PR #23440, opened 2021-12-20
vdr-plugin-2.eclass: fix content of @SUPPORTED_EAPIS

madmartin:
The eclass variable @SUPPORTED_EAPIS does not contain EAPI 8, while the
eclass itself supports EAPI 8. This leads to wrong warnings by pkgcheck.


thesamesam:
Thanks!

PR #23441, opened 2021-12-20
app-misc/ddcutil: add 1.2.1

reagentoo:
Closes: https://bugs.gentoo.org/829724
Signed-off-by: Dmitry Baranov <reagentoo@gmail.com>


app-misc/ddcutil/ddcutil-1.2.1.ebuild
```diff
@@ -0,0 +1,109 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
AndrewAmmerlaan, line 4: Could we bump this to `EAPI=8` please while we are touching the package anyway.

app-misc/ddcutil/ddcutil-1.2.1.ebuild
```diff
@@ -0,0 +1,109 @@
...
+
+src_prepare() {
+	default
+	eautoreconf
+	sed -e "s#usr/local/bin#usr/bin#" -i data/etc/udev/rules.d/45-ddcutil-usb.rules || die
+}
+
+src_configure() {
+	# Bug 607818.
+	replace-flags -O3 -O2
```
AndrewAmmerlaan, line 63: Is this still required?

reagentoo: Built successful without it. (app-misc/ddcutil-1.2.1  USE="-X -drm -introspection -usb-monitor -user-permissions" VIDEO_CARDS="-nvidia")
Also sed line has dropped from src_prepare.

app-misc/ddcutil/ddcutil-1.2.1.ebuild
reagentoo: Updated.

app-misc/ddcutil/ddcutil-1.2.1.ebuild
AndrewAmmerlaan: > Built successful without it. (app-misc/ddcutil-1.2.1 USE="-X -drm -introspection -usb-monitor -user-permissions" VIDEO_CARDS="-nvidia") Also sed line has dropped from src_prepare.

Does it build with `CFLAGS="... -O3...." CXXFLAGS="... -O3...."`? Cause that is what this line was working around.

reagentoo: > Does it build with `CFLAGS="... -O3...." CXXFLAGS="... -O3...."`? Cause that is what this line was working around.

Yes. My current config from make.conf:
```
COMMON_FLAGS="-march=x86-64-v3 -mtune=generic -O3 -pipe"
CFLAGS="${COMMON_FLAGS}"
CXXFLAGS="${COMMON_FLAGS}"
FCFLAGS="${COMMON_FLAGS}"
FFLAGS="${COMMON_FLAGS}"
```

AndrewAmmerlaan: Great, thanks for checking!

AndrewAmmerlaan:
Thanks :+1: 

I made some small changes in a follow-up commit: https://github.com/gentoo/gentoo/commit/9979db8b0e3ec4f1c760b70032626e7067030e74
I added subslot information to the ebuilds because version 1.2.1 installs libraries with a different SONAME than version 0.9.9. Dropped the `flag-o-matics` eclass since we are no longer using it, and fixed the `HOMEPAGE` to use `https`

I also added the latest version of ddcui, without keywords because it segfaults at launch: https://github.com/rockowitz/ddcui/issues/34

PR #23447, opened 2021-12-21
sci-libs/rocPRIM: remove unused cmake variable

littlewu2508:
__skip_rocmclang is used to avoid configuration error for
cmake-3.21.(1|2), which don't exist among ebuilds anymore, so this flag
is not recognized.

Closes: https://bugs.gentoo.org/829075


heroxbd:
Do you mean __skip_rocmclang is removed in 3.21.3?

Does the ebuild work with cmake 3.21.3 if there is no __skip_rocmclang handle?

littlewu2508:
> Do you mean __skip_rocmclang is removed in 3.21.3?
Exactly.
> Does the ebuild work with cmake 3.21.3 if there is no __skip_rocmclang handle?
Yes. 

Only cmake-3.21.(1|2) require this flag or it will fail; other version don't recognize this variable.

PR #23450, opened 2021-12-21
sci-libs/rocBLAS: various fixes

littlewu2508:
This PR contains three commits:

let TensileCreateLibrary respects MAKEOPTS
Closes: https://bugs.gentoo.org/822828

common fix for unuesd cmake variable:
Bugs: https://bugs.gentoo.org/829326

fix test failures:
Closes: https://github.com/ROCmSoftwarePlatform/rocBLAS/issues/1202

### DIFF
--- a/sci-libs/rocBLAS/rocBLAS-4.3.0.ebuild
+++ b/sci-libs/rocBLAS/rocBLAS-4.3.0.ebuild
@@ -98,7 +98,6 @@ src_configure() {
 		-DBUILD_CLIENTS_TESTS=$(usex test ON OFF)
 		-DBUILD_CLIENTS_BENCHMARKS=$(usex benchmark ON OFF)
 		${AMDGPU_TARGETS+-DAMDGPU_TARGETS="${AMDGPU_TARGETS}"}
-		-D__skip_rocmclang="ON" ## fix cmake-3.21 configuration issue caused by officialy support programming language "HIP"
 	)
 
 	CXX="hipcc" cmake_src_configure
--- a/sci-libs/rocBLAS/rocBLAS-4.3.0.ebuild
+++ b/sci-libs/rocBLAS/rocBLAS-4.3.0.ebuild
@@ -46,7 +46,8 @@ S="${WORKDIR}"/${PN}-rocm-${PV}
 
 PATCHES=("${FILESDIR}"/${PN}-4.3.0-fix-glibc-2.32-and-above.patch
 	"${FILESDIR}"/${PN}-4.3.0-change-default-Tensile-library-dir.patch
-	"${FILESDIR}"/${PN}-4.3.0-link-system-blas.patch )
+	"${FILESDIR}"/${PN}-4.3.0-link-system-blas.patch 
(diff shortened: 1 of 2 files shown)


sci-libs/rocBLAS/rocBLAS-4.3.0.ebuild
```diff
@@ -97,7 +99,6 @@ src_configure() {
 		-DBUILD_CLIENTS_TESTS=$(usex test ON OFF)
 		-DBUILD_CLIENTS_BENCHMARKS=$(usex benchmark ON OFF)
 		${AMDGPU_TARGETS+-DAMDGPU_TARGETS="${AMDGPU_TARGETS}"}
-		-D__skip_rocmclang="ON" ## fix cmake-3.21 configuration issue caused by officialy support programming language "HIP"
```
thesamesam, line 100: Depend on a newer CMake version?

littlewu2508: only cmake-3.21.1 and cmake-3.21.2 need this, version below this or higher than this don't need or recognize this flag, and will throw a warning (like https://bugs.gentoo.org/829326 described). The two version does not exists in portage tree now.

littlewu2508:
I found that the license is incorrect. rocBLAS is distributed under BSD.

littlewu2508:
@heroxbd Please have a look of this PR. I've got ROCm-5.0 ebuilds prepared, so it would be better if we apply existing fixes.

AndrewAmmerlaan:
> I found that the license is incorrect. rocBLAS is distributed under BSD.

We need both, rocBLAS is BSD but Tensile is MIT. 
By the way, is it at all possible to unbundle Tensile? There is an ebuild for stand-alone `dev-util/Tensile` in the `::science` repository.

littlewu2508:
> We need both, rocBLAS is BSD but Tensile is MIT. 

Tensile is only needed during compilation (python scripts that read configs and call compilers for GPU code). I'm not sure whether the MIT license of Tensile matters. 

> By the way, is it at all possible to unbundle Tensile? There is an ebuild for stand-alone `dev-util/Tensile` in the `::science` repository.

The Tensile in ::science is not up-to-date. However I can maintain the current version of Tensile, and it is possible to unbundle it if I do some hack.

AndrewAmmerlaan:
> The Tensile in ::science is not up-to-date. However I can maintain the current version of Tensile, and it is possible to unbundle it if I do some hack.

It is up to date now :smile:

AndrewAmmerlaan:
> I'm not sure whether the MIT license of Tensile matters.

Depends, are these files installed to the system? If so, the license matters, if not then we don't care.

AndrewAmmerlaan:
There's an `|| die` missing here:
```
sed -r -e "/pip install/s:([^ \"\(]*python) -m pip install ([^ \"\)]*):\1 setup.py install --single-version-externally-managed --root / WORKING_DIRECTORY \2:g" -i cmake/virtualenv.cmake
```

AndrewAmmerlaan:
Thanks :+1: 

I also moved Tensile into ::gentoo

PR #23452, opened 2021-12-21
dev-util/rt-tests: Avoid CC in install

eddy1021:
The ostype and machinetype are currently retrieved by `$(CC)
-dumpmachine`. The $(CC) dependency can be avoided by using `uname`.

### DIFF
--- a/dev-util/rt-tests/files/rt-tests-2.2-avoid_cc_in_install.patch
+++ b/dev-util/rt-tests/files/rt-tests-2.2-avoid_cc_in_install.patch
@@ -1,0 +1,25 @@
+diff --git a/Makefile b/Makefile
+index 569adc1..508d7a7 100644
+--- a/Makefile
++++ b/Makefile
+@@ -38,15 +38,12 @@ else
+ 	CFLAGS	+= -O0 -g
+ endif
+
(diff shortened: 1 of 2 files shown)


thesamesam:
It's a reasonable change although I wonder what the motivation is here. We shouldn't compile stuff in src_install but invoking CC there is legal, even if it is odd. Could you explain why you need this?

1. could you please submit this upstream (and put a link in the patch)? This may be a pain to rebase otherwise
2. include sign off in commit

eddy1021:
Otherwise, we need to also specify the name of the C compiler in `emake` in the `src_install`.
i.e. `emake CC=${tc-getCC} prefix=/usr DESTDIR="${ED}" install`.

Oops, I thought this repo is the upstream.

shifty91:
Just add `CC=$(tc-getCC)` to the `install` traget for now. In case the patch is accepted upstream, we can remove it. Thanks.

eddy1021:
Created another PR: https://github.com/gentoo/gentoo/pull/23462

juippis:
I guess with #23462 this one can be closed?

thesamesam:
Feel free to reopen once accepted upstream.

PR #23454, opened 2021-12-21
vdr-plugin-2.eclass: fix variable VDRPLUGIN

madmartin:
Variable VDRPLUGIN: remove declaration @INTERNAL, instead declare as
@OUTPUT_VARIABLE as it is used in many packages
This fixes errors thrown by pkgcheck

### DIFF
--- a/eclass/vdr-plugin-2.eclass
+++ b/eclass/vdr-plugin-2.eclass
@@ -15,10 +15,10 @@
 # Eclass for easing maintenance of vdr plugin ebuilds
 
 # @ECLASS-VARIABLE: VDRPLUGIN
-# @INTERNAL
+# @OUTPUT_VARIABLE
 # @DESCRIPTION:
 # The name of the vdr plugin, plain name without "vdr-" or "plugin" prefix or suffix.
-# This variable is derived from ${PN}
+# This variable is derived from ${PN} and is read-only for the ebuild.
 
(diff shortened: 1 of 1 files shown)


eclass/vdr-plugin-2.eclass
```diff
@@ -15,7 +15,6 @@
 # Eclass for easing maintenance of vdr plugin ebuilds
 
 # @ECLASS-VARIABLE: VDRPLUGIN
```
thesamesam, line 17: May want to add @OUTPUT_VARIABLE if ebuilds should read it

madmartin: Hi, ebuilds cant read it if they want... but not necessarily must do...

thesamesam: I think that still fits the criteria. I should've said "can" or "are allowed to"

madmartin: NP, you are citating the dev manual... "This tag indicates that the variable's value (which will be set by the eclass) should be read within an ebuild " 
IMHO the dev manual should be changed. Something like "variable is set by the eclass and read-only for the ebuild" ?

madmartin: will push in a minute

thesamesam:
Could you make a bug or file a PR for the devmanual? Thanks!

PR #23459, opened 2021-12-21
minor qa things for random gnome@ packages

gyakovlev:


mattst88:
I don't understand the QA-related bits. Is this because portage's flags-not-respected checks don't work for Rust?

gyakovlev:
> I don't understand the QA-related bits. Is this because portage's flags-not-respected checks don't work for Rust?

yeah, rust's analog for `-frecord-gcc-switches` is not there yet. we don't do `RUSTFLAGS` checking at all.
it will be done once available and is stable in rust.

so basically any pure rust project needs to specify ignored flags for now.

LDFLAGS (`-Wl,--defsym=__gentoo_check_ldflags__=0`) checker might work sometimes, but not always, depends on how final linkage is done.

mattst88:
Thanks!

PR #23461, opened 2021-12-22
app-emacs/lsp-mode: added lsp-mode-8.0.0-r2

Jturnerusa:
This PR will add some of the features from my lsp-mode ebuild into ::gentoo


app-emacs/lsp-mode/lsp-mode-8.0.0-r2.ebuild
```diff
@@ -0,0 +1,62 @@
...
+	>=app-emacs/spinner-1.7.3
+"
+DEPEND="${RDEPEND}"
+
+SITEFILE="50${PN}-gentoo.el"
+DOCS=( README.md CHANGELOG.org )
+RESTRICT="test" # test requires ert-runner which is not packaged
+
+src_compile() {
+	elisp-check-emacs-version
```
ulm, line 42: This belongs in `elisp-pkg-setup` (where it is by eclass default) but nowhere else.

app-emacs/lsp-mode/lsp-mode-8.0.0-r2.ebuild
```diff
@@ -0,0 +1,62 @@
...
+	ebegin "Compiling GNU Emacs Elisp files"
+	${EMACS} ${EMACSFLAGS} -L . -L "${S}"/clients -f batch-byte-compile *.el "${S}"/clients/*.el
+	eend $? "elisp-compile: batch-byte-compile failed" || die
+}
+
+src_install() {
+	elisp-make-autoload-file "${S}"/${PN}-autoload.el "${S}"/
```
ulm, line 49: `elisp-make-autoload-file` belongs in `src_compile`. (But see my comment about about excessive number of autoloads.)

app-emacs/lsp-mode/lsp-mode-8.0.0-r2.ebuild
```diff
@@ -0,0 +1,62 @@
...
+
+SITEFILE="50${PN}-gentoo.el"
+DOCS=( README.md CHANGELOG.org )
+RESTRICT="test" # test requires ert-runner which is not packaged
+
+src_compile() {
+	elisp-check-emacs-version
+	ebegin "Compiling GNU Emacs Elisp files"
+	${EMACS} ${EMACSFLAGS} -L . -L "${S}"/clients -f batch-byte-compile *.el "${S}"/clients/*.el
+	eend $? "elisp-compile: batch-byte-compile failed" || die
```
ulm, line 45: No need to do this by hand. Use `elisp-compile` instead (and set `BYTECOMPFLAGS` as necessary).

ulm (changes requested):

app-emacs/lsp-mode/lsp-mode-8.0.0-r2.ebuild
Jturnerusa: That's a much better idea! Thanks

PR #23470, opened 2021-12-22
app-editors/featherpad: Bump to 1.1.0

hansfn:
Also removed 0.18.0.


app-editors/featherpad/featherpad-1.1.0.ebuild
```diff
@@ -8,21 +8,30 @@ inherit xdg cmake
 DESCRIPTION="Lightweight Qt5 Plain-Text Editor for Linux"
 HOMEPAGE="https://github.com/tsujan/FeatherPad"
 SRC_URI="https://github.com/tsujan/FeatherPad/archive/V${PV}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/FeatherPad-${PV}"
 
 LICENSE="GPL-3+"
 SLOT="0"
-KEYWORDS="~amd64 ~x86"
+KEYWORDS="~amd64 ~riscv ~x86"
+IUSE="+X"
```
juippis, line 16: Out of curiosity, what does `-X` do here since it still pulls a lot of Qt stuff, including Qtgui. Is it supposed to be pure wayland-only option?

juippis (commented):
We usually separate adding and removing ebuilds to their individual commits, so this PR should have 2 commits: add a new ebuild, and remove old. But I don't mind it this time if you can answer me about the X flag.

juippis (approved):

app-editors/featherpad/featherpad-1.1.0.ebuild
hansfn: This is almost turning into a FAQ -  see metadata.xml.

hansfn:
I'll do adding and removing as separate commits next time. Thx for the feedback.

juippis:
Heh, indeed. The Github UI showed it as a new flag due to moving the ebuild into new one :P sorry for the troubles and thanks for the bump!

PR #23474, opened 2021-12-22
kicad version bump to 6.0.0

eeyrjmr:
6.0.0 was tagged and released earlier ( https://gitlab.com/kicad/code/kicad/-/tags/6.0.0 ).  The DOC's have not been released for 6.0.0 yet and thus these have not been bumped, likewise the meta has not been bumped. These packages have been locally built to confirm they work


thesamesam:
The pkgcheck / CI warnings here are kind of fragile and I think we can just drop the rename. I think we can fix that on merge if needed.

But can you:
1. git reset origin/master (no --hard, which will keep your changes, just drop all the commits, so they need to be re-staged and re-committed)
2. repoman commit in each directory (or use pkgdev commmit) to get the commit messages in the right format and such? (and we do one commit per package)

Then force push again. Thank you!

eeyrjmr:
done. 
The QA complaint could be resolved by a guard around rc as it is valid for how the rc are tagged (_ vs -)

PR #23483, opened 2021-12-23
games-util/springlobby: drop package

mfld-fr:
Mainstream is broken as outdated, and proxy maintainers are not active anymore.
Maintained in the registered 'spring' overlay by the Spring RTS game community.

Closes: https://bugs.gentoo.org/732582
Closes: https://bugs.gentoo.org/739398
Closes: https://bugs.gentoo.org/756715

Signed-off-by: mfld.fr gentoo@mfld.fr


Flowdalic:
It can't hurt to create a treeclean bug for this an CC treecleaners, see https://wiki.gentoo.org/wiki/Project:Treecleaner/Policy. We also have a procedure for removing packages: https://devmanual.gentoo.org/ebuild-maintenance/removal/index.html

mfld-fr:
@Flowdalic : CC'ed the treecleaners in the existing opened bugs. See also https://github.com/springlobby/overlay/issues/25 for the discussion & agreement from @abma as the Spring RTS community leader.

Flowdalic:
@mfld-fr before we start the process of treecleaning springlobby: would you be interested in taking over springlobby in ::gentoo? It appears you are active in maintaining games-util/springlobby in the spring overlay, so why not also maintain it in ::gentoo? :)

thesamesam:
> @mfld-fr before we start the process of treecleaning springlobby: would you be interested in taking over springlobby in ::gentoo? It appears you are active in maintaining games-util/springlobby in the spring overlay, so why not also maintain it in ::gentoo? :)

Yeah, had a similar question on Bugzilla too: https://bugs.gentoo.org/732582#c14.

mfld-fr:
@Flowdalic & @thesamesam : Hello and thanks for that good question, that I asked myself before choosing rather to work in the `spring` overlay.

From what I learnt recently with disappointment  from the Spring RTS community (forum & lobby chat), the original project is in a very bad shape, as the main workforce heavily moved to active forks like 'Beyond All Reasonable' and 'Zero-K', with their own in-game lobbies (mainly 'Chobby')  and infra... so I am afraid that after my current fixup of the related ebuilds and other Gentoo-related like the wikis, it is highly probable that they will be broken again in a near future because of the mainstream obsolescence...

This is the main reason that makes me thinking it does not worth to keep any related `ebuild` in Gentoo mainstream. If by chance the community that is currently sadly dying in the original ecosystem (stuck on the latest stable release of engine + official lobby + infra) revives again, it would be more easier for them to update in the overlay, and even change the packaging layout if needed.

Flowdalic:
@mfld-fr Thanks for the thorough explanation. It sure looks like treecleaning springlobby from ::gentoo is sensible. If the situation changes in the future, it can be re-added,

PR #23484, opened 2021-12-23
app-emulation/xen: add USE to disable boot symlinks

hydrapolic:
Closes: https://bugs.gentoo.org/829765


juippis (approved):
I get a feeling like this isn't the best possible solution for that bug but... guess we can change it afterwards if people are unhappy.

hydrapolic:
While I agree this is not the best solution, we don't have an auto-detection for /boot if it supports symlinks. You are right, we can improve it in the future.

hydrapolic:
Thanks @juippis

PR #23485, opened 2021-12-23
dev-util/spirv-llvm-translator: Version bump (13.0.0)

janhenke:
Besides bumping the matching LLVM version, this version requires the
upstream SPIRV header files, so an additonal CMAKE argument is required
to point to the installed headers (and they are a new build time
dependency).

LLVM eclass sadly does not support EAPI 8 yet, so this ebuild remains at
7.

Closes: https://bugs.gentoo.org/829845


dev-util/spirv-llvm-translator/spirv-llvm-translator-13.0.0.ebuild
```diff
@@ -0,0 +1,56 @@
...
+
+src_prepare() {
+	append-flags -fPIC
+	cmake_src_prepare
+}
+
+multilib_src_configure() {
+	local mycmakeargs=(
+		-DCMAKE_INSTALL_PREFIX="$(get_llvm_prefix ${LLVM_MAX_SLOT})"
+		-DLLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR="/usr/include/spirv"
```
thesamesam, line 44: We should ${BROOT} this

dev-util/spirv-llvm-translator/spirv-llvm-translator-13.0.0.ebuild
```diff
@@ -0,0 +1,56 @@
...
+
+RESTRICT="!test? ( test )"
+
+S="${WORKDIR}/${MY_P}"
+
+COMMON="sys-devel/clang:13=[${MULTILIB_USEDEP}]"
+DEPEND="${COMMON}
+		dev-util/spirv-headers"
+RDEPEND="${COMMON}"
+BDEPEND="test? ( dev-python/lit )"
```
thesamesam, line 30: We may need to use python-any-r1 to ensure we have Python for tests too, right? Or is it just needing some lit executable?

dev-util/spirv-llvm-translator/spirv-llvm-translator-13.0.0.ebuild
```diff
@@ -0,0 +1,56 @@
...
+LICENSE="UoI-NCSA"
+SLOT="13"
+KEYWORDS="~amd64"
+IUSE="test +tools"
+
+RESTRICT="!test? ( test )"
+
+S="${WORKDIR}/${MY_P}"
+
+COMMON="sys-devel/clang:13=[${MULTILIB_USEDEP}]"
```
thesamesam, line 26: I would use COMMON_DEPEND or you could just write these in a different order and not need a separate one at all.

dev-util/spirv-llvm-translator/spirv-llvm-translator-13.0.0.ebuild
```diff
@@ -0,0 +1,56 @@
...
+
+MY_PN="SPIRV-LLVM-Translator"
+MY_P="${MY_PN}-${PV}"
+
+DESCRIPTION="Bi-directional translator between SPIR-V and LLVM IR"
+HOMEPAGE="https://github.com/KhronosGroup/SPIRV-LLVM-Translator"
+SRC_URI="https://github.com/KhronosGroup/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="UoI-NCSA"
+SLOT="13"
```
thesamesam, line 18: Can I suggest you refactor this a bit so you only have to use '13' once ($(ver_cut) from the version, use it in LLVM_MAX_SLOT, SLOT, and the dependency)?

dev-util/spirv-llvm-translator/spirv-llvm-translator-13.0.0.ebuild
```diff
@@ -0,0 +1,56 @@
...
+	append-flags -fPIC
+	cmake_src_prepare
+}
+
+multilib_src_configure() {
+	local mycmakeargs=(
+		-DCMAKE_INSTALL_PREFIX="$(get_llvm_prefix ${LLVM_MAX_SLOT})"
+		-DLLVM_EXTERNAL_SPIRV_HEADERS_SOURCE_DIR="/usr/include/spirv"
+		-DLLVM_BUILD_TOOLS=$(usex tools "ON" "OFF")
+		$(usex test "-DLLVM_INCLUDE_TESTS=ON" "")
```
thesamesam, line 46: Can't do LLVM_INCLUDE_TESTS=$(usex test)? (OFF doesn't work?)

dev-util/spirv-llvm-translator/spirv-llvm-translator-13.0.0.ebuild
thesamesam: Oh, sorry, I see that now. Yeah, ignore this.

dev-util/spirv-llvm-translator/spirv-llvm-translator-13.0.0.ebuild
janhenke: Done

dev-util/spirv-llvm-translator/spirv-llvm-translator-13.0.0.ebuild
janhenke: Done

dev-util/spirv-llvm-translator/spirv-llvm-translator-13.0.0.ebuild
janhenke: Done

dev-util/spirv-llvm-translator/spirv-llvm-translator-13.0.0.ebuild
janhenke: Done

ConiKost:
Btw, do we need to keep mulilib? Based on p.g.o, I only see opencl-lang with multilib pkgs, but I may be wrong? Musn't dev-libs/opencl-clang bumped also to 13?

PR #23486, opened 2021-12-23
www-apps/gitea: bump to 1.15.8

hydrapolic:
Tested on ~amd64, thanks!


www-apps/gitea/gitea-9999.ebuild
```diff
@@ -1,123 +0,0 @@
-# Copyright 2016-2021 Gentoo Authors
-# Distributed under the terms of the GNU General Public License v2
-
-EAPI=7
```
thesamesam, line 4: Someone mentioned on IRC they're using this, but tbh, if it breaks often, I can't argue with dropping it.

hydrapolic: As mentioned in the commit message, the "live" version does not contain the frontend part compiled so we'd have to call npm install which would fail anyway. I got this working, but only the backend part so it's pretty useless.

A solution would be to compile the frontend (or pick it from some release) and put it to some archive and download along the git repo, but honestly I'll just wait for the release which they release very (maybe too) often.

thesamesam: I'm sorry Tomáš, I committed a sin by not checking! Yeah. Let's just dump it, they release very often anyway.

hydrapolic: We are all sinful after all, so no wories. But, we have hope :)

hydrapolic:
Thank you @thesamesam

PR #23487, opened 2021-12-23
Risc-V profile cleanup - highly experimental, DO NOT MERGE

akhuettel:


eclass/multilib.eclass
```diff
@@ -425,29 +428,32 @@ multilib_env() {
...
 			export CFLAGS_ilp32=${CFLAGS_ilp32--mabi=ilp32 -march=rv32imac}
 			export CHOST_ilp32=${CTARGET/riscv64/riscv32}
 			export CTARGET_ilp32=${CTARGET/riscv64/riscv32}
-			export LIBDIR_ilp32=${LIBDIR_ilp32-lib32/ilp32}
+			export LIBDIR_ilp32=${LIBDIR_ilp32-lib/ilp32}
 
-			: ${MULTILIB_ABIS=lp64d lp64 ilp32d ilp32}
-			: ${DEFAULT_ABI=lp64d}
+			local __libdir_riscvdefaultabi_variable="LIBDIR_${DEFAULT_ABI}"
+			export ${!__libdir_riscvdefaultabi_variable}="lib64"
```
Arfrever, line 439: This does not work with `!` here (too many dereferences):
`export ${!__libdir_riscvdefaultabi_variable}="lib64"` is expanded to `export lib64/lp64d=lib64`
`export ${__libdir_riscvdefaultabi_variable}="lib64"` is expanded to `export LIBDIR_lp64d=lib64`

However there is no need for that intermediate variable `__libdir_riscvdefaultabi_variable`. Single-statement solution is any of these:
```
eval "export LIBDIR_${DEFAULT_ABI}=\"lib64\""
eval "export LIBDIR_${DEFAULT_ABI}=lib64"
```

akhuettel: Surprising, but true. And fixed now.

That wasn't the only problem with the eclass though; the logic in case the libdir was already assigned elsewhere was wrong so far. (Never mind how likely the case of a *different* assignment elsewhere is...) Should also be fixed now.

eclass/multilib.eclass
```diff
@@ -415,6 +415,14 @@ multilib_env() {
 			: ${MULTILIB_ABIS=lp64d lp64 ilp32d ilp32}
 			: ${DEFAULT_ABI=lp64d}
 
+			# the default abi is set to the 1-level libdir default
+
+			local __libdir_riscvdefaultabi_variable="LIBDIR_${DEFAULT_ABI}"
+			local __libdir_riscvdefaultabi=${!__libdir_riscvdefaultabi_variable}
+			export ${__libdir_riscvdefaultabi_variable}=${__libdir_riscvdefaultabi:-lib64}
```
Arfrever, line 422: This 3-lines solution works, but it is alternatively possible to use 1-line solution:
```
eval "export LIBDIR_${DEFAULT_ABI}=\${LIBDIR_${DEFAULT_ABI}:-lib64}"
```

akhuettel: No thanks, I'm going for readability (and the hope that someone else will understand it in the future) there.

PR #23488, opened 2021-12-23
media-gfx/gimp: 2.10.30 version bump

band-a-prend:
Fix media-libs/gegl CVE: https://bugs.gentoo.org/829880


thesamesam:
Thank you!

PR #23489, opened 2021-12-23
net-vpn/protonvpn-cli: missing dep on dev-python/distro

ILMostro:
Closes: https://bugs.gentoo.org/823224
Signed-off-by: Amel Hodzic <ilmostro7@gmail.com>


ILMostro:
I didn't want to create needless revbumps, but since this is intended to fix runtime issues with the package, I'll probably have to force-push that here.  As I'm not quite 100% sure, and the activity in the lack of user activity in the assigned bug, I'll await a confirmation before I try to resolve this.

thesamesam:
> I didn't want to create needless revbumps, but since this is intended to fix runtime issues with the package, I'll probably have to force-push that here. As I'm not quite 100% sure, and the lack of user activity in the assigned bug, I'll await a confirmation before I try to resolve this.

Yep, need a revbump to ensure users pick up the fixed RDEPEND. Otherwise there's no way for the package manager to know.

thesamesam:
Merged as 6d3011325a6b4974ede57c5a5f48cc6f2a25ec49 (lost closes tag b/c squashed commits).

thesamesam:
Thanks!

PR #23493, opened 2021-12-24
media-gfx/hydrus: Add the lcms use flag to pillow, drop 461, add 467

vaartis:
Seems like this flag is required now. I had it enabled on my machine already, so didn't notice.

Closes: https://bugs.gentoo.org/829900


thesamesam:
signoff then will merge

vaartis:
Added the new version while I'm at it, with the same fix.

thesamesam:
Thank you!

PR #23495, opened 2021-12-24
sys-libs/libselinux: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/829922


t0b3:
This will help reduce complexity in QA report https://qa-reports.gentoo.org/output/gpyutils/39-to-310.svg around sys-libs/libselinux

thesamesam:
Nit: I'd use the commit description "sync live" here since you're not _really_ adding anything new (so no real need to worry about it). Nice spot, thank you!

thesamesam:
oh, btw, would you mind syncing your signed-off-by with your git author? you can put it in make.conf if using repoman.

PR #23494, opened 2021-12-24
dev-python/rdflib: add ~ia64

t0b3:
Closes: https://bugs.gentoo.org/829921

PR #23496, opened 2021-12-24
dev-python/berkeleydb: add ~ia64

t0b3:
Closes: https://bugs.gentoo.org/829925

Needed for PR https://github.com/gentoo/gentoo/pull/23494


t0b3:
WIP see https://bugs.gentoo.org/814179

PR #23503, opened 2021-12-24
Bump libuev 2.4

oz123:


thesamesam:
(See https://github.com/gentoo/gentoo/pull/23505#discussion_r775175447 too)

PR #23505, opened 2021-12-25
dev-libs/libite: add 2.5.1

oz123:
Closes: https://bugs.gentoo.org/828486


dev-libs/libite/libite-2.5.1.ebuild
```diff
@@ -0,0 +1,22 @@
...
+DESCRIPTION="A collection of useful BSD APIs"
+HOMEPAGE="https://github.com/troglobit/libite"
+SRC_URI="https://github.com/troglobit/libite/releases/download/v${PV}/${P}.tar.xz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+src_configure() {
+	econf --enable-static=no
```
thesamesam, line 15: Does --distable-static work? I guess not.

Also, EAPI 8? It should do that for us too ;)

PR #23515, opened 2021-12-25
app-backup/burp: bump for 2.4.0 (upstream stable)

antonfischl1980:
I have been using these ebuilds myself on ~60 machines for the last few months and thought about taking over maintainership of this package. Hopefully I don't make a fool out of myself.

I upgraded to EAPI=8. pkgcheck and repoman seem to be happy about these ebuild.

2.4.0 is latest stable upstream
2.5.4 is latest release upstream. Upstream defined 2.5.x as allways being "unstable/not meant for production"

Therefor 2.4.0 should be marked stable after the usual grace period and the existing versions 2.2.18-r2 and 2.3.36-r1 removed at that point.


app-backup/burp/burp-2.4.0.ebuild
```diff
@@ -0,0 +1,115 @@
...
+src_configure() {
+	local myeconfargs=(
+		--localstatedir=/var
+		--sysconfdir=/etc/burp
+		--enable-largefile
+		$(use_enable acl)
+		$(use_enable ipv6)
+		$(use_enable xattr)
+	)
+	# --runstatedir option will only work from autoconf-2.70 onwards
```
thesamesam, line 56: We can just BDEPEND on a newer autoconf instead of doing this hack with runstatedir.

(I don't think the eclass yet lets us do it, see bug [827852](https://bugs.gentoo.org/827852))

app-backup/burp/burp-2.4.0.ebuild
```diff
@@ -0,0 +1,115 @@
...
+}
+
+pkg_postinst() {
+	elog "Burp ebuilds now support the autoupgrade mechanism in both"
+	elog "client and server mode. In both cases it is disabled by"
+	elog "default. You almost certainly do NOT want to enable it in"
+	elog "client mode because upgrades obtained this way will not be"
+	elog "managed by Portage."
+
+	if [[ ! -e /etc/burp/CA/index.txt ]]; then
```
thesamesam, line 90:
```suggestion
	if [[ ! -e "${EROOT}"/etc/burp/CA/index.txt ]]; then
```

app-backup/burp/burp-2.4.0.ebuild
```diff
@@ -0,0 +1,115 @@
...
+		elog ""
+		elog "At first run burp server will generate DH parameters and SSL"
+		elog "certificates.  You should adjust configuration before."
+		elog "Server configuration is located at"
+		elog ""
+		elog "  /etc/burp/burp-server.conf"
```
thesamesam, line 96:
```suggestion
		elog "  ${EROOT}/etc/burp/burp-server.conf"
```

app-backup/burp/burp-2.4.0.ebuild
```diff
@@ -0,0 +1,115 @@
...
+	# According to PMS this can be a space-separated list of version
+	# numbers, even though in practice it is typically just one.
+	local oldver
+	for oldver in ${REPLACING_VERSIONS}; do
+		if [[ $(ver_cut 1 ${oldver}) -lt 2 ]]; then
+			ewarn "Starting with version 2.0.54 we no longer patch bedup to use"
+			ewarn "the server config file by default. If you use bedup, please"
+			ewarn "update your scripts to invoke it as"
+			ewarn ""
+			ewarn "  bedup -c /etc/burp/burp-server.conf"
```
thesamesam, line 109:
```suggestion
			ewarn "  bedup -c ${EROOT}/etc/burp/burp-server.conf"
```

thesamesam (commented):
Looks decent. Left a comment or two.

Other notes:
1. Could you also split the commit into two, for each new version you add?
2. If 2.5.x isn't for production, we should mask it for testing / drop keywords (I prefer masks for this, easier to not forget & explains why).

antonfischl1980:
I removed 2.5.4 and will submit a seperate PR once this here got in the way.
Do I understand this correctly? the 2.5.4 PR should contain the ebuild and in profiles/package.mask:
`#Anton Fischl ... (2021-xx-yy)
# Version 2.5.* is not meant for production according to upstream
=app-backup/burp-2.5*`

I also added BDEPEND on recent autoconf and CCed myself on https://bugs.gentoo.org/827852
Sould I change it to "WANT_AUTOCONF=latest" once this bug is closed?

last but not least I changed the ${EROOT} as you suggested. In hindsight its obvious, I just never had seen an ${EROOT} other than "/". Will remember this for future ebuilds.

thesamesam:
Nice! I think we're getting there.

1. It is okay to do 2.5.4 in this PR too if you wish. Just separate commits.
2. Please squash ("fixup") the "removed 2.5.4" commit into the first commit, unless you're going to act on 1., in which case it should just be dropped anyway ;)
3. Yeah, that sounds good to me re package.mask. Put a space before your name though: "# Anton ..."
4. Please sign off your commits (git commit -s does this). repoman commit will do this for you if [configured](https://wiki.gentoo.org/wiki/Gentoo_git_workflow#Signoff) to and pkgdev commit will do this for you too without configuration.

antonfischl1980:
Thank you for your support and patience.

1. I think seperating 2.4.0 (stable) and 2.5.x with profile/package.mask is cleaner (and I dont want my very first PR allready messing around in the really important stuff :) )
2. done
3. will do at later point
4. done manually for now. I will read up on doing this automatically right now.

thesamesam:
Excellent! And sorry, yes, in theory, WANT_AUTOCONF="latest" would work, although really we'd want some way to just say >=2.71 rather than "I only want 2.71"... that's an eclass deficiency (not actually checked if it has a mechanism for this, but I don't think so) and not your fault though

app-backup/burp/burp-2.4.0.ebuild
```diff
@@ -0,0 +1,115 @@
...
+SRC_URI="https://github.com/grke/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="AGPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~x86"
+IUSE="acl ipv6 test xattr"
+
+RESTRICT="!test? ( test )"
+
+CDEPEND="acct-group/burp
```
juippis, line 19: We prefer `COMMON_DEPEND` here, 
https://wiki.gentoo.org/wiki/Project:Proxy_Maintainers/User_Guide/Style_Guide#Avoid_using_single-char_.2ADEPEND_variable

app-backup/burp/burp-2.4.0.ebuild
```diff
@@ -0,0 +1,115 @@
...
+	dev-libs/openssl:0=
+	net-libs/librsync:=
+	sys-libs/ncurses:0=
+	sys-libs/zlib
+	virtual/libcrypt:=
+	acl? ( sys-apps/acl )
+	xattr? ( sys-apps/attr )"
+DEPEND="${CDEPEND}
+	test? ( dev-libs/check )"
+BDEPEND=">=sys-devel/autoconf-2.71
```
juippis, line 31: Yeah I don't think the eclass really handles "any-version-above-than" conditions that well...?

juippis (approved):

PR #23516, opened 2021-12-26
Bump pipenv 2021 11 23

oz123:
Continue work on removing vendor packages.

https://bugs.gentoo.org/717666


Flowdalic:
Thanks for your contribution. It seems that 7ca8ecf is actually a fixup of 5687c25, and ce6254f is a fixup of 49b573e. If so, then please squash the commits accordingly.

oz123:
@Flowdalic done.

Flowdalic:
@oz123 Thanks for the prompt work. :) My main concern is that the patches are rather large, they consume around 80 KiB in `files/` on disk, and I wonder if one couldn't sed the sources instead. Of course, another alternative would be to create a tarball with the patches and place it somewhere.

thesamesam:
> @oz123 Thanks for the prompt work. :) My main concern is that the patches are rather large, they consume around 80 KiB in `files/` on disk, and I wonder if one couldn't sed the sources instead. Of course, another alternative would be to create a tarball with the patches and place it somewhere.

Using sed to generate patches is a far better idea. They must be uploaded to devspace or similar (possibly by whoever merged this, I do that often) but blindly using sed in the ebuild is far likely to lead to silent runtime breakage. This exact problem happened with Calibre a few times.

oz123:
@thesamesam that is no problem. I created the patched using a set of sed commands.
I can simply add those in `src_prepare` I guess?
Here is how I prepared the patches:

```
PN=pipenv
PV=2021-11-23
packages=( colorama attr urllib3 requests dateutil idna )

for pkgName in "${packages[@]}"; do
	# pkgName=$1
	find ./ -type f -exec sed -i 's/from pipenv.vendor import '"${pkgName}"'/import '"${pkgName}"'/g' {} \;

	find ./ -type f -exec sed -i 's/from pipenv.vendor.'"${pkgName}"'\(.*\) import \(\w*\)/from '"${pkgName}"'\1 import \2/g' {} \;

	# import idna
	find ./ -type f -exec sed -i 's/import pipenv.vendor.'"${pkgName}"' as '"${pkgName}"'/import '"${pkgName}"'/g' {} \;
	
	git commit -m "remove vendored $pkgName" pipenv/
	git format-patch HEAD~1 
	mv -v 000* "$PN-$PV-remove-vendored-$pkgName.patch"
	mv -v "$PN-$PV-remove-vendored-$pkgName.patch" ../gentoo/dev-python/pipenv/files/
done
```
As for your concern about those breaking silently: I think it's possible to catch with the test phase missing imports in the actual CLI. Many of the vendored packaged are there only for bootstrapping pipenv (which  isn't a difficulty, if you install with a package manager like emerge). 

Let me know how do you prefer to proceed.

oz123:
Just adding another note here, it seems that removing certifi from vendor induced havoc. I still need to check, but we might still need to keep it, until I figure exactly what the problem is.

thesamesam:
> @thesamesam that is no problem. I created the patched using a set of sed commands. I can simply add those in `src_prepare` I guess? Here is how I prepared the patches:
> 
> ```
> PN=pipenv
> PV=2021-11-23
> packages=( colorama attr urllib3 requests dateutil idna )
> 
> for pkgName in "${packages[@]}"; do
> 	# pkgName=$1
> 	find ./ -type f -exec sed -i 's/from pipenv.vendor import '"${pkgName}"'/import '"${pkgName}"'/g' {} \;
> 
> 	find ./ -type f -exec sed -i 's/from pipenv.vendor.'"${pkgName}"'\(.*\) import \(\w*\)/from '"${pkgName}"'\1 import \2/g' {} \;
> 
> 	# import idna
> 	find ./ -type f -exec sed -i 's/import pipenv.vendor.'"${pkgName}"' as '"${pkgName}"'/import '"${pkgName}"'/g' {} \;
> 	
> 	git commit -m "remove vendored $pkgName" pipenv/
> 	git format-patch HEAD~1 
> 	mv -v 000* "$PN-$PV-remove-vendored-$pkgName.patch"
> 	mv -v "$PN-$PV-remove-vendored-$pkgName.patch" ../gentoo/dev-python/pipenv/files/
> done
> ```
> 
> As for your concern about those breaking silently: I think it's possible to catch with the test phase missing imports in the actual CLI. Many of the vendored packaged are there only for bootstrapping pipenv (which isn't a difficulty, if you install with a package manager like emerge).
> 
> Let me know how do you prefer to proceed.

Totally fine with me if you include a commented block in the ebulid on how you created it!

Yeah, I figured that test suites would catch most/all of it: my main concern was making sure there's a manual step to generate patches which makes it easier to avoid silent seds beocming outdated and so on. It's kind of a disincentive to try bumping without checking the changes upstream and so on, plus you can see the changes the patch actually made.

So, that sounds good to me!

oz123:
@thesamesam including the 'script to create the patches' in a comment still has the large patches in the tree. These are only going to get bigger (I intend to remove all the vendored packages). Are you sure you prefer this over execution of the sed commands in `src_prepare`?

thesamesam:
> @thesamesam including the 'script to create the patches' in a comment still has the large patches in the tree. These are only going to get bigger (I intend to remove all the vendored packages). Are you sure you prefer this over execution of the sed commands in `src_prepare`?

The person who merges the PR should tarball them and upload them to their devspace (or you could put them in a e.g. github repo) I think. But I won't insist as long as you're careful and make sure things work with seds in src_prepare (maybe add a comment emphasising the importance of running the test suite or something)?

Flowdalic:
> But I won't insist as long as you're careful and make sure things work with seds in src_prepare (maybe add a comment emphasising the importance of running the test suite or something)?

The sed arguments provided by @oz123 seem robust, in a sense that it seems feasible to use them in the ebuild. My remaining concerns are nullifie, given that the produced patches apparently did not have to be manually modified afterwards. Hence I would give the sed approach a chance. But please add the comment suggested by sam.

@oz123: You could speed up the ebuild by performing the patching in parallel, with the helpf of the `multiprocessing` eclass and coalesce the sed scripts with one sed call . Something like

```bash
inherit multiprocessing
…
src_prepare() {
    …
    local jobs
    jobs=$(makeopts_jobs)
    find ./ -type f -print0 |\
        xargs --max-procs="${jobs} --null \
        sed --in-place\
            -e 's/from pipenv.vendor import '"${pkgName}"'/import '"${pkgName}"'/g' \
            -e 's/from pipenv.vendor.'"${pkgName}"'\(.*\) import \(\w*\)/from '"${pkgName}"'\1 import \2/g' \
            -e 's/import pipenv.vendor.'"${pkgName}"' as '"${pkgName}"'/import '"${pkgName}"'/g'
    assert "Failed to sed sources"
}
```

(untested)

PR #23522, opened 2021-12-26
dev-ruby/rdoc: remove the hprefixify command on shebangs.

heroxbd:
If two ruby targets are specified, the scripts get `/usr/bin/env ruby`
shebangs.  "env ruby" is then replaced in the ebuilds to be
"ruby${ver}" (ruby27 or ruby30).  They needs to be prefixified.

Contrastingly, when there is one ruby target, the shebang is the
correct EPREFIX/usr/bin/ruby${ver}.  No prefixify should be applied.

To unify the two cases, the shebangs for the two-ruby-target case are
changed to be `EPREFIX/usr/bin/env ruby`.

Reference: https://github.com/gentoo/gentoo/pull/21046
Signed-off-by: Benda Xu <heroxbd@gentoo.org>

### DIFF
--- a/dev-ruby/rdoc/rdoc-6.3.2.ebuild
+++ b/dev-ruby/rdoc/rdoc-6.3.2.ebuild
@@ -13,7 +13,7 @@ RUBY_FAKEGEM_BINDIR="exe"
 
 RUBY_FAKEGEM_GEMSPEC="rdoc.gemspec"
 
-inherit prefix ruby-fakegem
+inherit ruby-fakegem
 
 DESCRIPTION="An extended version of the RDoc library from Ruby 1.8"
 HOMEPAGE="https://github.com/ruby/rdoc/"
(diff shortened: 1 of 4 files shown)


dev-ruby/rdoc/rdoc-6.3.2.ebuild
```diff
@@ -82,7 +82,6 @@ all_ruby_install() {
 				ruby_fakegem_binwrapper $bin /usr/bin/${bin}${version}
 				sed -i -e "1s/env ruby/ruby${version}/" \
 					"${ED}/usr/bin/${bin}${version}" || die
-				use prefix && hprefixify "${ED}/usr/bin/${bin}${version}"
 			fi
```
thesamesam, line 85: Can you drop the prefix inherit now?

heroxbd:
Sam James ***@***.***> writes:

> Can you drop the prefix inherit now?

Oh yes!

graaff:
I have now merged this and also applied the same change to the 6.4.0 ebuild.

heroxbd:
Thanks!

PR #23525, opened 2021-12-26
sci-libs/pdal: add pdal library

t0b3:
to work with point clouds (i.e. in qgis)


t0b3:
Closed in favor of https://github.com/gentoo/gentoo/pull/23526

PR #23526, opened 2021-12-26
sci-geosciences/qgis: add pdal for point clouds

t0b3:
QGIS > v3.18 supports point clouds through PDAL.
Gentoo may support QGIS built with PDAL support via USE flag.

QGIS built with PDAL can be used to work with point cloud data, i.e. to stream EPT point clouds from https://usgs.entwine.io/

depends on https://github.com/gentoo/gentoo/pull/23525


sci-libs/pdal/pdal-2.3.0.ebuild
```diff
@@ -0,0 +1,53 @@
...
+		${PYTHON_DEPS}
+		dev-python/numpy[${PYTHON_USEDEP}]
+	)
+"
+RDEPEND="${DEPEND}"
+
+S="${WORKDIR}/PDAL-${PV}-src"
+
+src_prepare() {
+	eapply_user
```
AndrewAmmerlaan, line 49: This is double, `cmake_src_prepare` will also apply the user patches

sci-libs/pdal/pdal-2.3.0.ebuild
```diff
@@ -0,0 +1,53 @@
...
+	sed -i -e 's:"${PDAL_LIB_INSTALL_DIR}/cmake/PDAL":"lib/cmake/PDAL":' CMakeLists.txt || die "sed failed!"
+
+	cmake_src_prepare
+}
```
AndrewAmmerlaan, line 53: I think we are missing a `src_configure` phase here that actually toggles building python/postgres/sqlite. Currently the flags appear to only pull in the dependency. We are probably going to need something like this:

```
src_configure() {
	local mycmakeargs=(
		-DBUILD_PYTHON="$(usex python)"
                .....etc
	)
	cmake_src_configure
}
```
Check the upstream `CMakeLists.txt` file for the names of the configure options.

sci-libs/pdal/pdal-2.3.0.ebuild
```diff
@@ -0,0 +1,54 @@
...
+	sys-libs/libunwind
+	sys-libs/zlib
+	postgres? ( dev-db/postgresql:*[xml] )
+"
+RDEPEND="${DEPEND}"
+
+S="${WORKDIR}/PDAL-${PV}-src"
+
+src_prepare() {
+	sed -i -e 's:"${PDAL_LIB_INSTALL_DIR}/cmake/PDAL":"lib/cmake/PDAL":' CMakeLists.txt || die "sed failed!"
```
thesamesam, line 36: Can you make this into a patch instead, please?

sci-libs/pdal/pdal-2.3.0.ebuild
```diff
@@ -0,0 +1,54 @@
...
+	sed -i -e 's:cmake_minimum_required(VERSION 2:cmake_minimum_required(VERSION 3:' tools/lasdump/CMakeLists.txt || die "sed failed!"
+	sed -i -e 's:cmake_minimum_required(VERSION 2:cmake_minimum_required(VERSION 3:' apps/CMakeLists.txt || die "sed failed!"
+
+	cmake_src_prepare
+}
+
+src_configure() {
+	local mycmakeargs=(
+		-DBUILD_PLUGIN_PGPOINTCLOUD="$(usex postgres)"
+		#-DWITH_LAZPERF=OFF
```
thesamesam, line 48: ?

sci-libs/pdal/pdal-2.3.0.ebuild
```diff
@@ -0,0 +1,54 @@
...
+
+EAPI=8
+
+inherit cmake
+
+DESCRIPTION="A C++ library for translating and manipulating point cloud data"
+HOMEPAGE="https://pdal.io/"
+SRC_URI="https://github.com/PDAL/PDAL/releases/download/${PV}/PDAL-${PV}-src.tar.gz"
+
+SLOT="0/13"
```
thesamesam, line 12: Swap order of SLOT and LICENSE please

t0b3:
@thesamesam thank you for your suggestions. should be fine now, else pls let me know...

sci-libs/pdal/metadata.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "https://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
```
thesamesam, line 3: Nit: could you add yourself too? Then I think we're ready.

t0b3:
@thesamesam ready to merge :+1:

PR #23527, opened 2021-12-26
sys-cluster/singularity: allow to build without cryptsetup.

heroxbd:
If there is no cryptsetup in the environment, singularity can be built
without encryption features.

Note that if USE=-crypt and cryptsetup is installed, singularity still
build with encryption features.  The upstream does not provide a switch
to the `mconfig` configuration script.

Reference: https://github.com/apptainer/singularity/issues/4506
Signed-off-by: Benda Xu <heroxbd@gentoo.org>

### DIFF
--- a/sys-cluster/singularity/singularity-3.8.5-r1.ebuild
+++ b/sys-cluster/singularity/singularity-3.8.5-r1.ebuild
@@ -1,0 +1,71 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit linux-info toolchain-funcs
+
+DESCRIPTION="Application containers for Linux"
+HOMEPAGE="https://sylabs.io"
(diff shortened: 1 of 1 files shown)


thesamesam:
>Note that if USE=-crypt and cryptsetup is installed, singularity still
build with encryption features. The upstream does not provide a switch
to the mconfig configuration script.

This sounds like an automagic dependency which is not allowed. In such cases, we would either always depend on cryptsetup, or fix the configure scirpt.

heroxbd:
> > Note that if USE=-crypt and cryptsetup is installed, singularity still
> > build with encryption features. The upstream does not provide a switch
> > to the mconfig configuration script.
> 
> This sounds like an automagic dependency which is not allowed (it causes problems for e.g. binary packages). In such cases, we would either always depend on cryptsetup, or fix the configure scirpt.

Interesting point.  That helps me make up my mind to patch the configure script.

mkszuba:
What sam has said: if it isn't possible to prevent singularity to use installed cryptsetup even for USE=-crypt, the latter has to remain an unconditional dependency. Feel free to reopen this PR if you do update the build scripts, or just open a new one!

PR #23529, opened 2021-12-26
media-gfx/inkscape: bump to v1.1.1

t0b3:


media-gfx/inkscape/inkscape-1.1.1.ebuild
```diff
@@ -0,0 +1,159 @@
...
+
+inherit cmake flag-o-matic xdg toolchain-funcs python-single-r1
+
+DESCRIPTION="SVG based generic vector-drawing program"
+HOMEPAGE="https://inkscape.org/"
+SRC_URI="https://media.inkscape.org/dl/resources/file/${P}.tar.xz"
+
+LICENSE="GPL-2 LGPL-2.1"
+SLOT="0"
+KEYWORDS="amd64 ~arm ~arm64 ~hppa ~ia64 ppc ppc64 ~riscv ~s390 ~sparc x86"
```
jonasstein, line 17: KEYWORDS need ~ for new ebuilds and after a version bump. Run ekeyword ~all YOUREBUILD

t0b3: Thanks for spotting this one

t0b3: resolved in new commit

dev-cpp/gtkmm/gtkmm-3.24.5.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{7..9} )
+PYTHON_COMPAT=( python3_{8..10} )
```
jonasstein, line 6: I think there should be no Py 3.10 changes in this commit, unless it is a strict dependency.

thesamesam: It's fine if it works and I think they needed to do it to allow Inkscape to get Python 3.10.

t0b3: Screenshot_20211230_083726
python 3.10 depgraph shows *r+bdep* for inkscape 
source: gentoo python 3.10 qa report https://qa-reports.gentoo.org/output/gpyutils/39-to-310.svg

t0b3: ... and yes Inkscape and many others work fine on my py310-only box since October.

PR #23534, opened 2021-12-27
sys-apps/shadow: check for invalid passwd entries

thesamesam:
- In pkg_postinst, check for invalid ${EROOT}/etc/passwd entries
  by running 'pwck' in read-only mode.

- Also fix another check for empty ${EROOT} (broken since EAPI 7 port).

Closes: https://bugs.gentoo.org/829872

### DIFF
--- a/sys-apps/shadow/shadow-4.10-r4.ebuild
+++ b/sys-apps/shadow/shadow-4.10-r4.ebuild
@@ -234,12 +234,18 @@ pkg_preinst() {
 }
 
 pkg_postinst() {
+	# Missing entries from /etc/passwd can cause odd system blips.
+	# See bug #829872.
+	if ! pwck -r -q -R "${EROOT:-/}" &>/dev/null ; then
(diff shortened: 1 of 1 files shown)


sys-apps/shadow/shadow-4.10-r4.ebuild
```diff
@@ -234,10 +234,17 @@ pkg_preinst() {
 }
 
 pkg_postinst() {
+	# Missing entries from /etc/passwd can cause odd system blips.
+	# See bug #829872.
+	if ! pwck -r -q -R "${EROOT:-/}" &>/dev/null ; then
+		ewarn "Running 'pwck' returned errors. Please run it by hand and then"
+		ewarn "run 'pwconv' afterwards!"
```
floppym, line 241: If the user corrects the shadow file using pwck, there should be no need to run pwconv.

thesamesam: I think I got excited about being able to use the same text as below and didn't think. Yes, of course.

floppym (approved):

PR #23538, opened 2021-12-27
media-gfx/graphite2: add python 3.10

t0b3:
Closes: https://bugs.gentoo.org/830038


t0b3:
b20f577be5280947b9cfcbf0021b4a4ec715e1b6

PR #23541, opened 2021-12-27
net-dns/knot: bump to 3.1.5

nemunaire:


Flowdalic:
Thanks, please also consider looking into https://bugs.gentoo.org/826346

PR #23545, opened 2021-12-28
net-proxy/squid: use system libltdl (avoid maintainer mode automake)

thesamesam:
This bug has quite a long story, apparently, but it seems to be
enough here to just drop the subdir & force usage of the system version
with configure arguments (previous arguments weren't quite right).

Closes: https://bugs.gentoo.org/830099


thesamesam:
@mikhailkoliada

mikhailkoliada (approved):

mikhailkoliada:
@thesamesam could we try upstream the patch? (No idea if it gonna fly)

thesamesam:
> @thesamesam could we try upstream the patch? (No idea if it gonna fly)

sadly I don't think so for now, but I can try again at some point

PR #23553, opened 2021-12-28
app-text/asciidoc-9.1.1 version bump

marcecj:
I wanted to bump to 10.x, but that's proven harder than expected, it seems like there's still some ironing out to do upstream (or I'm simply not knowledgeable enough in Python packaging).

However, I'm sorry to say that I'm not interested in maintaining this anymore, so I'm also going to mark this as `maintainer-needed` in an additional commit.


thesamesam:
No problem, and thank you for your long work on this!

thesamesam:
Please know you can come back if you are ever interested. Best wishes.

PR #23556, opened 2021-12-29
net-p2p/go-ipfs: verbump to 0.11.0

stkw0:


net-p2p/go-ipfs/go-ipfs-0.11.0.ebuild
```diff
@@ -1726,7 +1955,8 @@ src_test() {
 src_install() {
 	dobin ipfs
 	dobin ipfswatch
-	newbashcomp misc/completion/ipfs-completion.bash ipfs
+	./ipfs commands completion bash > ipfs-completion.bash
```
thesamesam, line 1958: || die on external commands which may fail

thesamesam:
Hm, you tagged the OpenRC bug but didn't change anything there?

net-p2p/go-ipfs/go-ipfs-0.11.0.ebuild
thesamesam, line 1958: || die on external commands which may fail

kakra: Does this even make sense? It just creates the file in the build directory but it never installs it.

I believe the `newbashcomp` should stay in place, and the generator should go into the build phase?

thesamesam: Doesn't it install it on the next line? But yeah, it could/should maybe go into src_compile.

kakra: Ah okay, on the full view I can see the next line - it was missing on the review comments...

Maybe this `newbashcomp` line shouldn't change at all, and instead the compile phase should just create the file at the proper location `misc/completion/...`.

stkw0:
> Hm, you tagged the OpenRC bug but didn't change anything there?

yeah, that was solved in the PR for 0.10.0 but I did not tagged that issue so it's still open when it's already solved. I guess it would be better to mention the PR that solved it and request a dev to close the issue, but the OpenRC bug feels more like "cosmetic" changes so I just added it in this PR

net-p2p/go-ipfs/go-ipfs-0.11.0.ebuild
stkw0: It installs it after see the next line:
```
    ./ipfs commands completion bash > ipfs-completion.bash
    newbashcomp ipfs-completion.bash ipfs
```

thesamesam:
> > Hm, you tagged the OpenRC bug but didn't change anything there?
> 
> yeah, that was solved in the PR for 0.10.0 but I did not tagged that issue so it's still open when it's already solved. I guess it would be better to mention the PR that solved it and request a dev to close the issue, but the OpenRC bug feels more like "cosmetic" changes so I just added it in this PR

Yeah, I'd mention that in the bug and just link to the commit for posterity.

net-p2p/go-ipfs/go-ipfs-0.11.0.ebuild
```diff
@@ -1726,7 +1955,8 @@ src_test() {
 src_install() {
 	dobin ipfs
 	dobin ipfswatch
-	newbashcomp misc/completion/ipfs-completion.bash ipfs
+	./ipfs commands completion bash > ipfs-completion.bash || die
```
kakra, line 1958: Maybe move this line to  `src_compile()` as
```bash
./ipfs commands completion bash > misc/completion/ipfs-completion.bash || die
```

and leave this one untouched?

Maybe some make target is just missing which would create this file?

net-p2p/go-ipfs/go-ipfs-0.11.0.ebuild
stkw0: > Doesn't it install it on the next line? But yeah, it could/should maybe go into src_compile.

The problem is that the generator requires ipfs binary to be compiled, so if it's in src_compile, as the binary is still not compiled it will use the generator for the previous version, of it's the first time it's emerged it will fail

thesamesam: I think it would work if you called default first then put this line after, right? Not that I'm super bothered about this, it is a valid point

stkw0: Oh yes, right, it was obvious sry

kakra: > as the binary is still not compiled it will use the generator for the previous version

That's the same situation with the current approach: `src_install()` doesn't actually install to the search path, it installs to some sort of "chroot" directory (the image path). You are probably using `./ipfs` for that very reason. And this means, the binary is available in the build path - just like it would be at the end of the `src_compile()` function - or did I miss something?

net-p2p/go-ipfs/go-ipfs-0.11.0.ebuild
stkw0: In 0.10.0 upstream removed misc/completion folder entirely. If upstream won't use that folder to distribute other potential completions I don't see much value in creating a directory just to put that file tbh.

I will move that to src_compile, thanks.

kakra: Ah, that's the reason. They probably outsourced that to package maintainers.

kakra (approved):

PR #23557, opened 2021-12-29
net-misc/networkmanager: add patch to fix VPN plugins

stintel:
Upstream-PR: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/1053/commits
Closes: https://bugs.gentoo.org/732708
Signed-off-by: Stijn Tintel <stijn@linux-ipv6.be>


stintel:
[Upstream](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/1053#note_1202366) doesn't want to add the extra capability.

Opinions on the suggested alternative of having the plugin install a systemd override file?

PR #23558, opened 2021-12-29
app-admin/profile-cleaner: version bump to 2.37 and eapi 8

kernlpanic:


mattst88:
Thanks, I've squashed the EAPI=8 patch into the version bump.

PR #23565, opened 2021-12-29
app-i18n/transifex-client: add new version 1.0.0

robert7k:
This is a complete rewrite of the previous version.

I'm aware that the [old version](https://github.com/transifex/transifex-client/) is technically a different project than the [new version](https://github.com/transifex/transifex-cli/), but it can be considered a direct successor. Both are called "Transifex Client". The old version will stop working once the old Transifex API is disabled. Both install the same executable `tx`.

I don't have the means to test on ppc64. If someone is able to test that, we could add it as a keyword to avoid a regression compared to the old version.


app-i18n/transifex-client/transifex-client-1.0.0.ebuild
```diff
@@ -0,0 +1,229 @@
...
+
+S=${WORKDIR}/cli-${PV}
+
+src_install() {
+	dobin bin/tx
+	dodoc README.md
+}
+
+src_test() {
+	go test ./...
```
thesamesam, line 228: Missing `|| die` (if it fails right now, the ebuild will carry on) and please put it before src_install to match execution order.

app-i18n/transifex-client/transifex-client-1.0.0.ebuild
```diff
@@ -0,0 +1,229 @@
...
+go-module_set_globals
+
+SRC_URI="https://github.com/transifex/cli/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz
+	${EGO_SUM_SRC_URI}"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+S=${WORKDIR}/cli-${PV}
```
thesamesam, line 220:
```suggestion
S="${WORKDIR}"/cli-${PV}
```

.. and I usually put it directly under S (with no blank new line between them).

thesamesam (commented):
As for ~ppc64, file a rekeywording bug once merged.

PR #23568, opened 2021-12-29
remove obsolete selinux modules

plsph:
Add cleanup actions to post installation step by identifying removed modules from OLD_MODS list, that is copy from previous ebuild and remove it from selinux module store if installed.

One must maintain OLD_MODS list on each new ebuild by copying it from previous MODS list.

### DIFF
--- a/sec-policy/selinux-base-policy/selinux-base-policy-2.20210203-r1.ebuild
+++ b/sec-policy/selinux-base-policy/selinux-base-policy-2.20210203-r1.ebuild
@@ -27,6 +27,7 @@ BDEPEND="
 	sys-apps/checkpolicy
 	sys-devel/m4"
 
+OLD_MODS="application authlogin bootloader clock consoletype cron dmesg fstools getty hostname hotplug init iptables libraries locallogin logging lvm miscfiles modutils mount mta netutils nscd portage raid rsync selinuxutil setrans ssh staff storage su sysadm sysnetwork systemd tmpfiles udev userdomain usermanage unprivuser xdg"
 MODS="application authlogin bootloader clock consoletype cron dmesg fstools getty hostname init iptables libraries locallogin logging lvm miscfiles modutils mount mta netutils nscd portage raid rsync selinuxutil setrans ssh staff storage su sysadm sysnetwork systemd tmpfiles udev userdomain usermanage unprivuser xdg"
 LICENSE="GPL-2"
 SLOT="0"
@@ -97,6 +98,7 @@ pkg_postinst() {
 
 	# Override the command from the eclass, we need to load in base as well here
 	local COMMAND="-i base.pp"
+	local DEL_MODS=""
 	if has_version "<sys-apps/policycoreutils-2.5"; then
 		COMMAND="-b base.pp"
 	fi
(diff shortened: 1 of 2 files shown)


sec-policy/selinux-base-policy/selinux-base-policy-2.20210203-r1.ebuild
```diff
@@ -27,6 +27,7 @@ BDEPEND="
 	sys-apps/checkpolicy
 	sys-devel/m4"
 
+OLD_MODS="application authlogin bootloader clock consoletype cron dmesg fstools getty hostname hotplug init iptables libraries locallogin logging lvm miscfiles modutils mount mta netutils nscd portage raid rsync selinuxutil setrans ssh staff storage su sysadm sysnetwork systemd tmpfiles udev userdomain usermanage unprivuser xdg"
```
perfinion, line 30: ooh good catch on the hotplug removal. This mostly looks good but i think i'd rather change OLD_MODS to just be the list of modules to remove instead of needing to be a copy of all of them. that'd be less of a maintenance burden i think,
The issue is I have bumps scripted so copying MODS to OLD_MODS is complicated. also I'll keep the modules to remove list for at least a years worth of packages just in case someone does not update regularly and skips some versions

can we change this to `OLD_MODS="hotplug"` (or maybe DEL_MODS is nicer?)

sec-policy/selinux-base-policy/selinux-base-policy-2.20210203-r1.ebuild
```diff
@@ -105,12 +107,26 @@ pkg_postinst() {
 		COMMAND="${COMMAND} -i ${i}.pp"
 	done
 
+	for i in ${OLD_MODS}; do
+		if [ -n "${MODS##*$i*}" ]; then
+			DEL_MODS="${DEL_MODS} ${i}"
```
perfinion, line 112: Once the line above is changed to just the list of modules to remove then this loop can be changed to die instead.

maybe something like:
```
for i in ${OLD_MODS}; do
    [[ ${MODS} != *${i}* ]] || die "Duplicate module in MODS and OLD_MODS: ${i}"
done
````

**Review by perfinion — CHANGES_REQUESTED** [review]
Good catch on the hotplug removal! this looks pretty good, just would prefer a small change which should be easier to maintain going forward :)

Thanks and happy new year!

File: sec-policy/selinux-base-policy/selinux-base-policy-2.20210203-r1.ebuild (hunk 1 above)
Comment (plsph) [reply sec-policy/selinux-base-policy/selinux-base-policy-2.20210203-r1.ebuild]: Sure, no problem, that just need to be aligned with upstream to keep DEL_MODS list up to date.

**Comment by plsph** [comment]
Sorry for force push, I keep forgeting about Certificate of Origin.
This should work as described.
Have a nice new year.

File: sec-policy/selinux-base-policy/selinux-base-policy-2.20210203-r1.ebuild [hunk 3]
```diff
@@ -27,6 +27,7 @@ BDEPEND="
 	sys-apps/checkpolicy
 	sys-devel/m4"
 
+DEL_MODS="hotplug"
```
jpds, line 30: Tiny nit: I think `DEL_MODS` should go underneath `MODS`

perfinion:
About the DCO, it looks like your git is misconfigured, i had to use `git commit --amend --author` to make the author field match up with your signed-off-by name. Look at `git config user.email` and `user.name` I think should do it.

I made some minor changes:
- Moved the die up into src_prepare, its not good to die in the postinst stage.
- removed the quotes in `[[ "${MODS}" != *${i}* ]]` otherwise it doesn't glob and wont match.
- The `if [[ -n ${DEL_MODS} ]]` was unnecessary around the for loop.
- Added to the -9999 package too cuz future bumps copy from the -9999 package

Happy New Year and Thanks for the PR!

PR #23569, opened 2021-12-29
Add sys-apps/systemd-utils

floppym:
This is intended to replace sys-boot/systemd-boot, sys-apps/systemd-tmpfiles, and sys-fs/udev. It should be easier to maintain one package instead of 3.


thesamesam (commented):
Thanks for working on this!

gyakovlev:
still no musl patchset yet.
I'll give it a test on some weird ppc64-be-musl box once it's there, otherwise seems you've sorted things out.

scope creep: we can add sysusers too =)
it has standalone build target and useful for adding users when ROOT!=/
our eclass even generates sysusers files.

gyakovlev:
https://cgit.openembedded.org/openembedded-core/tree/meta/recipes-core/systemd/systemd_250.3.bb#n32

musl patchset is there.
I'll try to do a new tarball before EOW if you don't beat me to it, bit busy this week.

PR #23570, opened 2021-12-29
Opencl clang

dzejej:
Simple patch for llvm 13


dzejej:
Bad PR. I try do a better one in a while

PR #23571, opened 2021-12-29
Intel gc

dzejej:
Fix a compilation bug https://bugs.gentoo.org/827651. Some references and comments are within ebuild file


dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
```diff
@@ -0,0 +1,112 @@
...
+EAPI=7
+
+CMAKE_BUILD_TYPE="Release"
+LLVM_MAX_SLOT="13"
+MY_PN="igc"
+MY_P="${MY_PN}-${PV}"
+#PYTHON_COMPAT=( python3_{8..10} )
+PYTHON_COMPAT=( python3_10 )
+
+inherit cmake flag-o-matic llvm python-any-r1 git-r3
```
ConiKost, line 13: This is wrong. Using vcs in a non 999 ebuild is not correct.

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
```diff
@@ -0,0 +1,112 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
ConiKost, line 4: How about EAPI8?

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
```diff
@@ -0,0 +1,112 @@
...
+
+CMAKE_BUILD_TYPE="Release"
```
ConiKost, line 6: Not needed with EAPI8 then.

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
```diff
@@ -0,0 +1,112 @@
...
+LLVM_MAX_SLOT="13"
+MY_PN="igc"
+MY_P="${MY_PN}-${PV}"
+#PYTHON_COMPAT=( python3_{8..10} )
+PYTHON_COMPAT=( python3_10 )
```
ConiKost, line 11: Why Python 3.10 only?

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
```diff
@@ -0,0 +1,112 @@
...
+
+inherit cmake flag-o-matic llvm python-any-r1 git-r3
+
+DESCRIPTION="LLVM-based OpenCL compiler for OpenCL targetting Intel Gen graphics hardware"
+HOMEPAGE="https://github.com/intel/intel-graphics-compiler"
+SRC_URI="https://github.com/intel/${PN}/archive/${MY_P}.tar.gz -> ${P}.tar.gz"
+EGIT_REPO_URI="https://github.com/KhronosGroup/SPIRV-Tools.git"
```
ConiKost, line 18: This is wrong. Using vcs in a non 999 ebuild is not correct. Please point to snapshot archives. Could you explain further why those are needed?

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
```diff
@@ -0,0 +1,112 @@
...
+EGIT_REPO_URI1="https://github.com/KhronosGroup/SPIRV-Headers.git"
+EGIT_REPO_URI2="https://github.com/intel/vc-intrinsics.git"  EGIT_COMMIT2="8ee879314584e6630688b0a3b290d065dcabb383"
+# protobuf version 3.19.1 - but it is unnecesary 
+#EGIT_REPO_URI3="https://github.com/protocolbuffers/protobuf" EGIT_COMMIT3="7c40b2df1fdf6f414c1c18c789715a9c948a0725"
```
ConiKost, line 22: Well, if not needed, than remove it?

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
```diff
@@ -0,0 +1,112 @@
...
+BDEPEND="
+	${PYTHON_DEPS}
+	>=sys-devel/lld-${LLVM_MAX_SLOT}
+"
+
+## Fixes compilation 
+## https://bugs.gentoo.org/827651
+src_unpack() {
+	default
+	git-r3_fetch ${EGIT_REPO_URI}
```
ConiKost, line 46: Update to snapshot archives.

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
```diff
@@ -0,0 +1,112 @@
...
+	#git-r3_fetch ${EGIT_REPO_URI3} ${EGIT_COMMIT3}
+	#git-r3_checkout ${EGIT_REPO_URI3} "${WORKDIR}/protobuf"
+}
+
+PATCHES=(
+	"${FILESDIR}/${PN}-1.0.9-no_Werror.patch"
+	"${FILESDIR}/${PN}-1.0.8173-opencl-clang_version.patch"
+	"${FILESDIR}/${PN}-1.0.8365-disable-git.patch"
+	"${FILESDIR}/${PN}-1.0.8365-cmake-project.patch"
+	"${FILESDIR}/${PN}-fix-lld-prebuilt.patch"
```
ConiKost, line 61: Please add version number to patch and use syntax as in other patches.

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
```diff
@@ -0,0 +1,112 @@
...
+		-DIGC_BUILD__VC_ENABLED="ON"
+
+		# See comments on https://github.com/intel/intel-graphics-compiler/issues/148#issuecomment-916228192
+		# There is on place so:
+		-DIGC_OPTION__LINK_KHRONOS_SPIRV_TRANSLATOR=ON
+		-DIGC_OPTION__USE_KHRONOS_SPIRV_TRANSLATOR_IN_VC=ON
+		-DIGC_OPTION__SPIRV_TRANSLATOR_MODE=Prebuilds
+		-DVC_INTRINSICS_SRC="${WORKDIR}/vc-intrinsics"
+		-DPROTOBUF_SRC_ROOT_FOLDER="${WORKDIR}/protobuf"
+		-DSPIRVLLVMTranslator_ROOT="/usr/lib/llvm/13/include/LLVMSPIRVLib/"
```
ConiKost, line 102: Please update path for correct version number from llvm class and prefix.

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
```diff
@@ -0,0 +1,112 @@
...
+		-DCMAKE_LIBRARY_PATH="$(get_llvm_prefix ${LLVM_MAX_SLOT})/$(get_libdir)"
+		-DIGC_OPTION__ARCHITECTURE_TARGET="Linux64"
+		-DIGC_OPTION__CLANG_MODE="Prebuilds"
+		-DIGC_OPTION__LLD_MODE="Prebuilds"
+		-DIGC_OPTION__LLDELF_H_DIR="${EPREFIX}/usr/include/lld/Common"
+		-DIGC_OPTION__LLVM_MODE="Prebuilds"
+		-DIGC_OPTION__LLVM_PREFERRED_VERSION="${llvm_version##*-}"
+
+		# VectorCompiler needs work, as at the moment upstream
+		# only supports building vc-intrinsics in place.
```
ConiKost, line 92: As you enabled it, you should change the comment. Does VC works for you?

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
```diff
@@ -0,0 +1,112 @@
...
+		# See comments on https://github.com/intel/intel-graphics-compiler/issues/148#issuecomment-916228192
+		# There is on place so:
+		-DIGC_OPTION__LINK_KHRONOS_SPIRV_TRANSLATOR=ON
+		-DIGC_OPTION__USE_KHRONOS_SPIRV_TRANSLATOR_IN_VC=ON
+		-DIGC_OPTION__SPIRV_TRANSLATOR_MODE=Prebuilds
+		-DVC_INTRINSICS_SRC="${WORKDIR}/vc-intrinsics"
+		-DPROTOBUF_SRC_ROOT_FOLDER="${WORKDIR}/protobuf"
+		-DSPIRVLLVMTranslator_ROOT="/usr/lib/llvm/13/include/LLVMSPIRVLib/"
+		# see https://issueexplorer.com/issue/intel/intel-graphics-compiler/212
+		-DBUILD_SHARED_LIBS:BOOL=OFF
```
ConiKost, line 104: I don't think, this is a good idea, I assume this will enable static libs?

ConiKost (changes requested):
Please also remove all unrelated commits from this PR.

janhenke:
Sorry, this cannot proceed as it currently is. Please remove all the commits not belonging to dev-util/intel-graphics-compiler from this branch and open seperate PRs for these unreleated changes.

Also I just saw that to opened #23572 with the same changes, but a different description. You can edit your PRs, no need to open new ones.

ConiKost:
@janhenke Could you have a look here? Thanks!

janhenke:
I would prefer to first have all the unrelated changes removed, as I find it difficult to find the real diff here. But then I surely will take a look.

ConiKost:
> I would prefer to first have all the unrelated changes removed, as I find it difficult to find the real diff here. But then I surely will take a look.

Yeah agree, but you can already look into the single commit for intel stuff, rest seems completly unrelated.

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
dzejej: I do not know if 3.10 is really necessary, but cmake checks for 3.10 and find it (python 3.10 is stable so I have installed). 
I put it rather for remark of more investigations about if we really need 3.10.

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
dzejej: My knowledge about ebuild system is rather poor, so my first goal is to make it compilable. Probably it could be better done, but sorry I do not how.

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
dzejej: Only for documentation. In final realise it should be remove.

dzejej:
Sorry guys, I think I mess something up with my git (and I do not know what :) ). All 3 PR commits should be separate and it was my intention, so please just  look only at the ebuild (and patchees associated with this ebuild)  witch are market by subject. 
But all three commits are necessary to clinfo works somehow.

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
dzejej: Because is based on intel-graphics-compiler-1.0.9441.ebuild  ??

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
AndrewAmmerlaan: cmake will always find (and compile with) only the latest enabled python target unless you put the cmake stuff inside a call to `python_foreach_impl cmake-stuff.....`. 

If python is only required during build time (which is implied by the usage of `python-any-1`) then this behaviour is correct and you can (and should) just expand PYTHON_COMPAT.

If python is also required during runtime the ebuild should use `python-single-r1` or `python-r1` in combination with `python_foreach_impl`.

dev-util/intel-graphics-compiler/intel-graphics-compiler-1.0.9636.ebuild
AndrewAmmerlaan: Github has the ability to download any commit as a tarball, so you can replace the git stuff with `SRC_URI="https://github.com/example/example/archive/${COMMIT}.tar.gz -> name-${P}.tar.gz`

dzejej:
The new PR will take into account  these suggestions.

PR #23572, opened 2021-12-29
intel-compute-runtime

dzejej:
intel-compute-runtime-21.49.21786 with some fixes. But due to llvm-13.0.0 bug - see
 https://issueexplorer.com/issue/intel/intel-graphics-compiler/204
the clinfo utility only partially works


dzejej:
I try to resubmit with only the correct files.

PR #23573, opened 2021-12-29
net-vpn/strongswan: bump to 5.9.4

rndxelement:
I added an ebuild for strongswan-5.9.4 with support for EAPI 8.
From the changelog at https://github.com/strongswan/strongswan/releases/tag/5.9.4 I couldn't identify anything impacting the ebuild, so the new ebuild is quite the same as the prior ebuild.
The security bug https://bugs.gentoo.org/818841 got a fix in the new upstream release, so I tagged it in this commit.


hydrapolic:
https://github.com/gentoo/gentoo/pull/23574

PR #23578, opened 2021-12-30
app-crypt/sequoia-sq: initial import

Flowdalic:
Closes: https://bugs.gentoo.org/724538


app-crypt/sequoia-sq/sequoia-sq-0.25.0.ebuild
```diff
@@ -0,0 +1,307 @@
...
+
+src_install() {
+	cargo_src_install --path sq
+
+	doman sq/man-sq-net-autocrypt/*
+
+	newbashcomp target/sq.bash sq
+
+	insinto /usr/share/zsh/site-functions
+	doins target/_sq
```
gyakovlev, line 301: install `fish` too?

```
	insinto /usr/share/fish/vendor_completions.d
	doins target/sq.fish
```

gyakovlev (approved):
lgtm! (but install fish completions too please, just 2 extra lines with no eclass/or dep changes)

I confirm trick with `CARGO_TARGET_DIR` works and is ok to use.

I had some changes to `cargo.eclass` planned that may affect this, but it'll happen later and I'll adjust ebuild as needed.

PR #23580, opened 2021-12-30
Bump dev-util/clion to bugfixed version

MarcusGrass:
See this issue https://youtrack.jetbrains.com/issue/JBR-3896 for the nature of the bug (perfomance related with X)
Only bumped the minor version tag, works locally but this is my first ebuild-edit so if I've forgotten something please let me know


MarcusGrass:
Doesn't seem to be re-labeled closing in favor of: https://github.com/gentoo/gentoo/pull/23603

PR #23584, opened 2021-12-30
Patches for kio, kio-extras, kinit, konsole to make X optional via cmake option() Bug: https://bugs.gentoo.org/813450

44616e46:


a17r:
Not enough packages. Did you try to upstream them?

44616e46:
After discussing on IRC I wanted to add context here as well, I haven't looked into upstreaming the changes, the diff's for CMakeLists.txt were done against the extracted sources. @thesamesam suggested to have one commit per package and so I rebased the changes to reflect that and added signoff.

a17r:
kde-frameworks/* patches will be picked together with the dozen or so other affected patches as soon as I get to it.

a17r:
Thanks for your work, however all listed packages were fixed in consultation with upstream and backports already reached Gentoo ebuild repository. You can track progress here: https://bugs.gentoo.org/813450

PR #23586, opened 2021-12-30
net-im/telegram-desktop-bin: bump version to 3.4.3

henning-schild:


henning-schild:
yet another release before the merge ...

henning-schild:
I will delay the dropping for another MR, doing it in the same MR just causes too much work given the frequent fixes from upstream

PR #23589, opened 2021-12-30
Opencl 13

dzejej:
I hope that PR is now OK ? (i.e. but means of number of changed files)


thesamesam:
I think you need to have a clean master branch then make a new branch for each chance

PR #23592, opened 2021-12-30
sci-geosciences/grass: ebuild update

hamarituc:
The PR bumps the packet version, removes old ebuilds an does some maintenance on the live ebuild. The changes of #23338 (adding Python 3.10) are already taken into account for the new ebuild, but not for the old ones to avoid merge conflicts.


sci-geosciences/grass/grass-8.0.0_rc1.ebuild
```diff
@@ -0,0 +1,271 @@
...
+MY_PM=${MY_PM/.}
+MY_P=${P/_rc/RC}
+
+DESCRIPTION="A free GIS with raster and vector functionality, as well as 3D vizualization"
+HOMEPAGE="https://grass.osgeo.org/"
+SRC_URI="https://grass.osgeo.org/${MY_PM}/source/${MY_P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0/8.0"
+KEYWORDS="~amd64 ~ppc ~x86"
```
thesamesam, line 22: I prefer to keep RCs unkeyworded usually so that people aren't subjected to beta software unless they want it, so I've made this one unkeyworded, but if you disagree, we can talk about it, of course. Maybe grass is OK with it and their RCs are more stable, but I dunno.

hamarituc: It's OK. Thank you for the hint. I will bear it in mind in the future.

tmszi: I use/test dev version 8.1 (9999) and 8.0 RC1 every day, and it's fairly stable.

thesamesam: @tmszi If as upstream, you think it's pretty reasonable for us to be shipping it in `~arch` (testing), then I'm OK with us doing that (but we won't mark RC versions stable).

tmszi: I agree with you, we will leave it without `~arch` (testing) for now.

PR #23595, opened 2021-12-30
dev-libs/intel-compute-runtime: 21.49.21786

dzejej:
intel-compute-runtime-21.49.21786 with some fixes. But due to
llvm-13.0.0 bug(?) - see
https://issueexplorer.com/issue/intel/intel-graphics-compiler/204
or bug(?) in intel graphics compiler the clinfo utility only
partially works.


juippis (commented):
I can't test this due to https://bugs.gentoo.org/827651

I think this package needs someone more knowledgeable about the current situation since it looks like newer versions that depend on llvm:12 are masked. I don't think programs that only support llvm:11 will live long in Gentoo anymore.

Ping @ConiKost

dzejej:
See the PR #23620 about intel-graphics-compiler: version 1.0.9636.

PR #23597, opened 2021-12-31
x11-wm/xmonad: update xmonad-0.15 -> xmonad-0.17.0

bartenbach:


x11-wm/xmonad-contrib/xmonad-contrib-9999.ebuild
```diff
@@ -0,0 +1,41 @@
+# Copyright 1999-2019 Gentoo Authors
```
thesamesam, line 1: Wrong header. Use pkgcheck and pkgdev / repoman to commit.

dev-haskell/x11/metadata.xml
```diff
@@ -1,14 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE pkgmetadata SYSTEM "https://www.gentoo.org/dtd/metadata.dtd">
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
```
thesamesam, line 2: ?

dev-haskell/x11/metadata.xml
```diff
@@ -1,14 +1,11 @@
...
 		<name>Gentoo Haskell</name>
 	</maintainer>
-	<longdescription>
-		A Haskell binding to the X11 graphics library. The binding is a direct
-		translation of the C binding; for documentation of these calls, refer
-		to \"The Xlib Programming Manual\", available online at
-		&lt;http://tronche.com/gui/x/xlib/&gt;.
-	</longdescription>
+	<use>
+		<flag name="pedantic">Be pedantic (-Werror and the like)</flag>
```
thesamesam, line 9: Do we actually want this as a flag..? Just turn it off?

thesamesam (commented):
Thanks! Could you also split the xmonad and xmonad-contrib changes into separate commits?

(One commit per package)

Flowdalic:
@bartenbach friendly ping :)

wincent925:
Would it be possible to merge the ebuilds from the gentoo-haskell ebuild repo into the main gentoo repo?
The ebuild repo already has the latest versions for xmonad.
https://github.com/gentoo-haskell/gentoo-haskell/blob/master/x11-wm/xmonad/xmonad-0.17.0.ebuild

n1rdeks:
and where Xmonad 0.17 and GHC latest version? Holy...

thesamesam:
You could just be polite. It would have been (and is still) perfectly acceptable to say you won't have time to address review comments and to ask someone else to take it over.

n1rdeks:
I apologize too, but u did not give me a chance for testing. Half year I did not see new ver. If I hadn't visited the site, I wouldn't have known about new ver and beautiful changelog. Ohoh... Larry (the cow? or bull?)- come back to people! Or Gentoo die...

Axarva:
@n1rdeks I believe the devs are working to the best of their capabilities. If you aren’t satisfied, with all due respect, you should probably try becoming a dev yourself rather than complaining how Gentoo is “dying” because xmonad isn’t getting an update. Also, using stack to install xmonad will give you the latest version, although it isn’t (officially) supported by portage (it works just fine). 

Jumping to other distros is also an option, but if you know your way around the system, you should be fine.

thesamesam:
>     Compiled? Test passed? Create ~arch version. With all due respect, if u not writing on JS - no need update NodeJS? Or if u not understand syntax Go? Or Rust? Zig? Need update this languages?

Because if I'm doing, in this scenario, nodejs, Go, Rust, and Zig, I may not have time to handle all the random Haskell packages too if I don't have help.

Anyway, if you're fine with bumping stuffl ocally, you can easily do the same with Gentoo, y'know.

PR #23608, opened 2021-12-31
Uclibc cleanup

SoapGentoo:


dev-lang/perl/perl-5.34.0-r5.ebuild
```diff
@@ -518,7 +518,7 @@ src_configure() {
 	[[ ${COLUMNS:-1} -ge 1 ]] || unset COLUMNS # bug #394091
 
 	# Perl has problems compiling with -Os in your flags with glibc
-	use elibc_uclibc || replace-flags "-Os" "-O2"
+	replace-flags "-Os" "-O2"
```
thesamesam, line 521: I don't expect you to necessarily try this but if it's "with glibc", it really should've been ``use elibc_glibc && ...`` in the first place..

@akhuettel, any ideas?

dev-libs/newt/newt-0.52.21-r1.ebuild
```diff
@@ -23,7 +23,6 @@ RDEPEND="
 	${PYTHON_DEPS}
 	>=dev-libs/popt-1.6
 	=sys-libs/slang-2*
-	elibc_uclibc? ( sys-libs/ncurses:0= )
```
thesamesam, line 26: It _really_ only needs ncurses on uclibc?

dev-perl/Ace/Ace-1.920.0-r6.ebuild
```diff
@@ -55,7 +54,7 @@ src_compile() {
 		# Parallel compile breaks :(
 		"-j1"
 	)
-	if use elibc_glibc || use elibc_musl || use elibc_uclibc ; then
+	if use elibc_glibc || use elibc_musl ; then
 		mymake+=( "LIBS=-ltirpc -lm" )
 		mymake+=( "USEROPTS=-I/usr/include/tirpc -fPIC" )
```
thesamesam, line 59: Nit: ${ESYSROOT}/usr/include/tirpc which should be fine.

dev-perl/Ace/Ace-1.920.0-r6.ebuild
```diff
@@ -55,7 +54,7 @@ src_compile() {
 		# Parallel compile breaks :(
 		"-j1"
 	)
-	if use elibc_glibc || use elibc_musl || use elibc_uclibc ; then
+	if use elibc_glibc || use elibc_musl ; then
 		mymake+=( "LIBS=-ltirpc -lm" )
```
thesamesam, line 58: Not sure that -lm belongs there and would ideally use pkg-config but let's not mess with it for now given how Pelr stuff is.

media-libs/libpulse/libpulse-15.0.ebuild
```diff
@@ -47,7 +47,6 @@ RDEPEND="
 	)
 	|| (
 		elibc_glibc? ( virtual/libc )
-		elibc_uclibc? ( virtual/libc )
```
thesamesam, line 50: no idea what this construct is even on about but yeah fine

profiles/prefix/windows/winnt/use.mask
```diff
@@ -46,7 +46,6 @@ selinux
 scanner
 smartcard
 svga
-uclibc
 # unicode is totally b0rked on interix
```
thesamesam, line 49: This comment can go: we don't have Interix anymore.

sys-apps/shadow/shadow-4.10-r4.ebuild
```diff
@@ -86,8 +86,6 @@ src_configure() {
 	)
 	econf "${myeconfargs[@]}"
 
-	has_version 'sys-libs/uclibc[-rpc]' && sed -i '/RLOGIN/d' config.h #425052
```
thesamesam, line 89: Wow, very stale.

sys-fs/xfsdump/xfsdump-3.1.9-r1.ebuild
```diff
@@ -24,7 +24,6 @@ RDEPEND="
 DEPEND="${RDEPEND}
 	nls? (
 		sys-devel/gettext
-		elibc_uclibc? ( dev-libs/libintl )
```
thesamesam, line 27: needs to be kept with virtual/libintl instead, although it should really be in RDEPEND too

sys-libs/musl-nscd/musl-nscd-1.0.3.ebuild
```diff
@@ -22,8 +22,7 @@ SLOT="0"
 IUSE="minimal"
 
 DEPEND="
```
thesamesam, line 24: This is broken as-is and we need to revbump for adding the blockers to RDEPEND too.

x11-terms/mrxvt/mrxvt-0.5.4.ebuild
```diff
@@ -26,7 +26,6 @@ RDEPEND="
 		media-libs/fontconfig
 		media-libs/freetype
 		x11-libs/libXft
-		elibc_uclibc? ( dev-libs/libiconv )
```
thesamesam, line 29: Replace it with a dependency on the virtual for e.g. darwin?

x11-terms/mrxvt/mrxvt-0.5.4.ebuild
```diff
@@ -43,11 +42,6 @@ PATCHES=(
 src_prepare() {
 	default
 	eautoreconf
-
-	if use elibc_uclibc && use truetype; then
-		# It is stated in the README "Multichar support under XFT requires GNU iconv"
```
thesamesam, line 48: seems odd this actually needs to go. you can probably get away with a cheesy has_version here libiconv as glibc will always (probably!) provide it and people don't change libcs.

thesamesam (commented):
some minor bits I noticed, but looks good once sorted.

re toolchain.eclass/any other eclass mentions, let's do that separately. plan is to remove the patchset stuff but keep any CHOST related hacks as portability stuff like that is sometimes hard to get back, if it's not too ugly.

dev-libs/newt/newt-0.52.21-r1.ebuild
SoapGentoo: I see no sign of curses being used in that package *anywhere*, I haven't checked whether it calls one of the executables, but it definitely doesn't link to curses

thesamesam (approved):

PR #23617, opened 2022-01-01
dev-java/resin-servlet-api: Drop 4.0.27-r1 (wrong slot)

vaukai:


fordfrog:
@vaukai merged, thanks!

PR #23627, opened 2022-01-02
dev-java/netty-buffer: Java heap space

vaukai:
"OutOfMemoryError: Java heap space" could be reproduced
with CHECKREQS_MEMORY="512M". Setting now to "2048M".
Closes: https://bugs.gentoo.org/829822

### DIFF
--- a/dev-java/netty-buffer/netty-buffer-4.0.36-r1.ebuild
+++ b/dev-java/netty-buffer/netty-buffer-4.0.36-r1.ebuild
@@ -1,0 +1,85 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# Skeleton command:
+# java-ebuilder --generate-ebuild --workdir . --pom pom.xml --download-uri https://github.com/netty/netty/archive/refs/tags/netty-4.0.36.Final.tar.gz --slot 0 --keywords "~amd64 ~arm64 ~ppc64 ~x86" --ebuild netty-buffer-4.0.36.ebuild
+
+EAPI=8
+
+JAVA_PKG_IUSE="doc source test"
+MAVEN_ID="io.netty:netty-buffer:4.0.36.Final"
+JAVA_TESTING_FRAMEWORKS="junit-4"
+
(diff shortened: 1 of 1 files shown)


fordfrog:
if it's only fix of failing tests then no need to revbump. i did not check all the changes and github does not show that to me so not sure whether this is the only change.

vaukai:
> if it's only fix of failing tests then no need to revbump. i did not check all the changes and github does not show that to me so not sure whether this is the only change.

It was also for some minor style updates but not really worth the revbump.  Thanks for the review.

fordfrog:
now when i see the changes it really needs a revbump. there's the change in eapi version, looks like deps have been updated, min java version...

vaukai:
> now when i see the changes it really needs a revbump. there's the change in eapi version, looks like deps have been updated, min java version...

Revbumped.  Should help getting rid of older dom4j versions ...

Flowdalic:
I would appreciate a better commit subject. if I am reading the message right, then the subject could be "dev-java/netty-buffer: increase heap to 2048M for tests". I am also not sure why check_env() isn't simply inlined in src_test()? Then CHECKREQS_MEMORY could become a local variable, not cluttering the globals.

vaukai:
> I would appreciate a better commit subject. if I am reading the message right, then the subject could be "dev-java/netty-buffer: increase heap to 2048M for tests". 

Commit message adjusted to proposal.

> I am also not sure why check_env() isn't simply inlined in src_test()? Then CHECKREQS_MEMORY could become a local variable, not cluttering the globals.

This coding is simply taken over from what's used in [dev-java/bcprov](https://gitweb.gentoo.org/repo/gentoo.git/commit/dev-java/bcprov/bcprov-1.69.ebuild?id=dcc9a94d7c88352328c9c238c352c0680b88467d).

Flowdalic:
> This coding is simply taken over from what's used in [dev-java/bcprov](https://gitweb.gentoo.org/repo/gentoo.git/commit/dev-java/bcprov/bcprov-1.69.ebuild?id=dcc9a94d7c88352328c9c238c352c0680b88467d).

Ahh, now I also see that this is due check-reqs eclass.

PR #23629, opened 2022-01-02
www-apps/tt-rss: Add PHP 8.0 support and drop PHP 7.2

creideiki:
[According to upstream](https://community.tt-rss.org/t/support-for-php-8-1/5089/13), the live Git version has supported PHP 8.0 for a while, but not 8.1 yet. I have been running my server on PHP 8.0 for a few months, without issue. Update the `-99999999` live Ebuild to include PHP 8.0 support.

Also remove the PHP 7.2 support from all versions, since PHP 7.2 is no longer in the tree, so supporting it makes Repoman unhappy.

Neither of these should require revbumps, IMHO: users of the live ebuild will update frequently anyway, and nobody should be running PHP 7.2 since it's no longer in the tree.

### DIFF
--- a/www-apps/tt-rss/tt-rss-20200922.ebuild
+++ b/www-apps/tt-rss/tt-rss-20200922.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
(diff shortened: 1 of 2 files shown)


chewi:
Thanks. I'll process this shortly. If you think the current live code is good then we can take a new snapshot around now too.

creideiki:
> If you think the current live code is good then we can take a new snapshot around now too.

I just made a new pull of the live ebuild on my server (to test this pull request), so I only have a couple of hours of runtime, but everything seems to work so far (this is on commit [97baf3e8b9be699d972b91a159ccbe0891efe8ae](https://git.tt-rss.org/fox/tt-rss/src/commit/97baf3e8b9be699d972b91a159ccbe0891efe8ae), which has been HEAD since December 23). In general, I usually pull once a month or so, and can't remember ever having any problems, so it's probably fine to just make regular time-based snapshots.

Do note that the configuration format has changed since Gentoo's last snapshot, so it might be a good idea to have an `if [[ -n "${REPLACING_VERSIONS}" ]]` block in `pkg_postinst()` which prints [the migration instructions](https://tt-rss.org/wiki/GlobalConfig#migrating-from-old-style-config.php).

PR #23631, opened 2022-01-02
www-apps/radicale: bump to 3.1.4

henning-schild:


www-apps/radicale/radicale-3.1.0.ebuild
```diff
@@ -0,0 +1,93 @@
+# Copyright 1999-2021 Gentoo Authors
```
henning-schild, line 1: its a new year

bedwardly-down:
Are you looking into taking charge of the acct-user/radicale and acct-group/radicale packages as well? I found for my particular usage, I had to modify the radicale user a bit.

henning-schild:
> Are you looking into taking charge of the acct-user/radicale and acct-group/radicale packages as well? I found for my particular usage, I had to modify the radicale user a bit.

Yes, those two come with it. Please share your problem and your current solution to it. I tried it for my use-case and did not find any issues, maybe you are on systemd?

bedwardly-down:
I'm on Openrc and use SELinux (this being one of the critical elements). The core issue I ran into was getting it to play nicely with using githooks to push to my Gitea repo and using a custom ssh key that root had no access to but could be added to the radicale home directory along with storing the git-credentials so I wouldn't have to always put in my credentials. Without SELinux, it was fine, but using the strict profile for it meant that the default permissions and policies used weren't going to cut it.

Currently, the radicale user's shell and home directory environment variable seem to default to /dev/null; this caused running any commands as the radicale user to fail with a "this account can't execute shell commands" error. If you're going to save your credentials so you don't have to manually enter them every time without saving them or your ssh keys directly to root, the shell needed to be set to /sbin/nologin instead. Setting the user's homedir environment variable to /var/lib/radicale was simple enough.

Commands used to fix this issue:
```
sudo usermod -s /sbin/nologin radicale
sudo usermod -m -d /var/lib/radicale
```

For SELinux, the labels needed to be slightly tweaked. I currently have /var/lib/radicale labeled as user_home_dir_t and most of its contents as user_home_t with the .ssh directory being labeled as ssh_home_t so that the sshd daemon can read its contents on start like it would for any normal user home directory, but there's probably better ways to handle it. If you're not running SELinux, it seems to work perfectly fine outside of the above commands.

henning-schild:
@bedwardly-down ok that is a little involved but valid

But it has nothing to do with the version bump proposed here. Only that i should add myself as maintainer there as well.

In fact "acct-user/radicale-0-r1" has the shell fixed to "/sbin/nologin" whether we can also set a home should probably be discussed in an issue separate to this discussion here. And in any case the setting of the home can stay your local hack for the stuff you added.

How did you configure the git pushing? Is that a "regular" configuration option like a hook or something ... in which case we very likely need a home. If it is cron ... it could potentially be discarded as your personal fun ... but i do not see why not set a valid home, others might like that as well.

bedwardly-down:
I definitely agree that this is a separate issue altogether. 

The git hooks used are radicale’s built in ones from the docs: https://radicale.org/v3.html#versioning-with-git. I added a `&& git push origin master` at the end of the commands from that section. I’d love to see this get maintained. 🤟

henning-schild:
> I definitely agree that this is a separate issue altogether.
> 
> The git hooks used are radicale’s built in ones from the docs: https://radicale.org/v3.html#versioning-with-git. I added a `&& git push origin master` at the end of the commands from that section. I’d love to see this get maintained. love_you_gesture

I will take the maint if people agree. Might even look into that backup feature myself, while i would probably want hg instead of git. That is another story but might enable me to see similar issues you seem to have with git.

> Also, now that I think about it, I’m using a hardened profile. That’s probably why /sbin/nologin wasn’t used here. In fact, that’s probably the reason for all of this. I haven’t tried on a normal profile yet but for my personal server, i’m implementing every security option I can and then dialing back where needed (like in this case).

Did not check hardened, maybe that is the reason. But let us go through with the MR adding a version bump and myself as maintainer and see about the other things in other threads.

henning-schild:
@bedwardly-down maintaining things is for sure more rewarding with active users who explain their use-cases like you did here

henning-schild:
@bedwardly-down in fact i had to set a HOME after all

henning-schild:
@juippis i found you involved in earlier merges on that package, any way to actually get any feedback on this

juippis (approved):

PR #23637, opened 2022-01-02
dev-python/graph-tool: Bump to 2.44

count0:


thesamesam:
Thanks! Could you:
1. run `pkgdev manifest` or `repoman manifest` or `ebuild ... manifest`
2. Include your signoff in your commit message? (`git commit -s` will do this)

PR #23642, opened 2022-01-03
media-gfx/iscan: Migrating to media-gfx/imagescan

gringus:
Migrating media-gfx/iscan to media-gfx/imagescan along with network plugin. Reverting 5bdebec86ddfa74cea155326b7c7863b8db84c36, so iscan-2 users can make use of iscan-2 compartible network plugin (download does not fail anymore).


thesamesam:
Why close?

gringus:
Decided to replace it with https://github.com/gentoo/gentoo/pull/24015 where I only add new packages. Masking iscan-3* and iscan-plugin-network-nt-1.1.3 and adding news can be done as separate task.

PR #23643, opened 2022-01-04
net-p2p/rtorrent: Remove dependency on screen for daemonising.

Razorfang:
Since rtorrent version 0.9.7, there is a built-in option for daemonising the program. This means that the old method of using _screen_ is outdated. This also means that we can remove the _screen_ dependency by using the new daemonising method.

I have tested the new openrc-related changes on my own PC and it works fine, but I have **not** tested the systemd-related changes.


ss23:
Any reason for the rename of `rtorrentd` to `rtorrent` in some places (e.g. init file)?

When it comes to a change like this, I think it's generally better to just do a bump than to change the old versions people may already be using, so just creating a 0.9.8-r1 for example.

Razorfang:
> Any reason for the rename of `rtorrentd` to `rtorrent` in some places (e.g. init file)?

No reason, I just preferred the name. You're right though, don't change what isn't broken.

> When it comes to a change like this, I think it's generally better to just do a bump than to change the old versions people may already be using, so just creating a 0.9.8-r1 for example.

Thanks for the feedback. This is my first time creating a pull request here. I'll make some changes later.

Razorfang:
I'm closing this because I want to re-implement the change in a different way. I'll open a new pull request soon.

PR #23644, opened 2022-01-04
media-video/wireplumber: add fix to stop stream move on HDMI suspend

pinkflames:
The patch looks like it might fix other things too but the best known
issue it should be resolving is that without it having a display
suspend or enter power off state would make audio streams be moved
to the next highest priority device. Such behavior might be
PulseAudio default but for PipeWire it's considered undesirable and
a bug to fix.

Bug: https://gitlab.freedesktop.org/pipewire/wireplumber/-/issues/94


thesamesam:
CI failure is fine obviously, I'll just merge this after PW

PR #23645, opened 2022-01-04
net-p2p/rtorrent: Remove dependency on screen for daemonising.

Razorfang:
Since rtorrent version 0.9.7, there is a built-in option for daemonising the program. This means that the old method of using screen is outdated. This also means that we can remove the screen dependency by using the new daemonising method.

I have tested the new openrc-related changes on my own PC and it works fine, but I have not tested the systemd-related changes.

### DIFF
--- a/net-p2p/rtorrent/files/rtorrent-r1.init
+++ b/net-p2p/rtorrent/files/rtorrent-r1.init
@@ -6,6 +6,7 @@ description="rTorrent BitTorrent client"
 command="/usr/bin/rtorrent"
 command_args="-o system.daemon.set=true"
 command_background=true
+command_user="${USER}"
 pidfile="/run/rtorrent.pid"
 
 depend()
--- a/net-p2p/rtorrent/files/rtorrent-r1.init
+++ b/net-p2p/rtorrent/files/rtorrent-r1.init
@@ -1,0 +1,15 @@
+#!/sbin/openrc-run
+# Distributed under the terms of the GNU General Public License v2
+
+description="rTorrent BitTorrent client"
+
+command="/usr/bin/rtorrent"
+command_args="-o system.daemon.set=true"
+command_background=true
+pidfile="/run/rtorrent.pid"
+
+depend()
+{
+   use net ypbind nis
+   after slapd mysqld postgresql
+}
--- a/net-p2p/rtorrent/files/rtorrentd_at-r1.service
+++ b/net-p2p/rtorrent/files/rtorrentd_at-r1.service
@@ -1,0 +1,17 @@
+# This configuration file is taken from the Arch wiki.
+# https://wiki.archlinux.org/title/RTorrent#Systemd_service_as_a_daemon_for_a_user
+
+[Unit]
+Description=rTorrent for %i
+After=network.target
+
+[Service]
+Type=simple
+User=%i
+Group=%i
+WorkingDirectory=/home/%i
+# Modify the next line to the absolute path for rtorrent.lock, for example
+ExecStartPre=-/bin/rm -f /home/%i/.session/rtorrent.lock
+ExecStart=/usr/bin/rtorrent -o system.daemon.set=true
+Restart=on-failure
+RestartSec=3
--- a/net-p2p/rtorrent/rtorrent-0.9.8-r1.ebuild
+++ b/net-p2p/rtorrent/rtorrent-0.9.8-r1.ebuild
@@ -59,6 +59,7 @@ src_install() {
 	default
 	doman doc/rtorrent.1
 
-	newinitd ${FILESDIR}/rtorrent-r1.init rtorrent
+	newinitd "${FILESDIR}/rtorrent-r1.init" rtorrent
+	newconfd "${FILESDIR}/rtorrentd.conf" rtorrent
 	systemd_newunit "${FILESDIR}/rtorrentd_at-r1.service" "rtorrentd@.service"
 }
--- a/net-p2p/rtorrent/rtorrent-0.9.8-r1.ebuild
+++ b/net-p2p/rtorrent/rtorrent-0.9.8-r1.ebuild
@@ -1,0 +1,64 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+inherit autotools linux-info systemd
+
+DESCRIPTION="BitTorrent Client using libtorrent"
+HOMEPAGE="https://rakshasa.github.io/rtorrent/"
+SRC_URI="http://rtorrent.net/downloads/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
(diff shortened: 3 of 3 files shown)


Razorfang:
If this does get merged, I'll also update the wiki instructions here: [here](https://wiki.gentoo.org/wiki/RTorrent#Customizing_the_daemonizer)

ss23:
Looks good to me @Razorfang. Before it gets merged, you'll probably need to complete the "Missing GCO sign-off" step that the bot posted here. The GURU guide has some information on how to automate the signing process - https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors#Repository.2Fgit_setup

perfinion:
oh nice! yeah if you can add the Signed-off-by then I'll merge it. The author field is already good so just `git commit --amend -s --no-edit` is probably enough to fix it.

Razorfang:
That doesn't seem to work for me. I just finished following all the instruction on the wiki (at least I think so). When typing

```
git commit --amend --no-edit -s
```

, I get the following error:
```
error: gpg failed to sign the data
fatal: failed to write commit object
```

Razorfang:
@ss23  @perfinion  Alright, I added my sign-off message.

ss23:
It looks like this new `rtorrent-r1.init` doesn't support configuring the user rtorrent runs as through `/etc/conf.d/rtorrentd` anymore. I am not sure how common this is, but it's something I use. With this change, I think it can only run as root?

Razorfang:
> It looks like this new `rtorrent-r1.init` doesn't support configuring the user rtorrent runs as through `/etc/conf.d/rtorrentd` anymore. I am not sure how common this is, but it's something I use. With this change, I think it can only run as root?

I haven't tried what you've done. I'll have a look at it this when I can find time, hopefully this week.

Razorfang:
@ss23  Alright, I see that the config file where you used to set the user isn't being copied in my ebuild. It's because between `rtorrent-0.9.8.ebuild` and `rtorrent-0.9.8-r1.ebuild`, I didn't move over the `newconfd` call.

```
$ diff rtorrent-0.9.8.ebuild rtorrent-0.9.8-r1.ebuild
15c15
< IUSE="daemon debug selinux test xmlrpc"
---
> IUSE="debug selinux test xmlrpc"
23d22
< 	daemon? ( app-misc/screen )
63,67c62,63
< 	if use daemon; then
< 		newinitd "${FILESDIR}/rtorrentd.init" rtorrentd
< 		newconfd "${FILESDIR}/rtorrentd.conf" rtorrentd
< 		systemd_newunit "${FILESDIR}/rtorrentd_at.service" "rtorrentd@.service"
< 	fi
---
> 	newinitd ${FILESDIR}/rtorrent-r1.init rtorrent
> 	systemd_newunit "${FILESDIR}/rtorrentd_at-r1.service" "rtorrentd@.service"
```

I don't think that alone will fix the issue though, because in the old `rtorrentd.init` file, the user was set via a call to `start-stop-daemon`, which my ebuild does not call at all. I instead use the `command` and `command_args` keywords.

Would adding `command_user` to the conf.d file and the previously mentioned fix be sufficient to solve your issue?

Razorfang:
@ss23  I have a partial fix for you.

Here's a diff that would make the change work:
```
diff --git a/net-p2p/rtorrent/files/rtorrent-r1.init b/net-p2p/rtorrent/files/rtorrent-r1.init
index df75e1a0d05..e16f262bfbe 100644
--- a/net-p2p/rtorrent/files/rtorrent-r1.init
+++ b/net-p2p/rtorrent/files/rtorrent-r1.init
@@ -6,6 +6,7 @@ description="rTorrent BitTorrent client"
 command="/usr/bin/rtorrent"
 command_args="-o system.daemon.set=true"
 command_background=true
+command_user="${USER}"
 pidfile="/run/rtorrent.pid"

 depend()
diff --git a/net-p2p/rtorrent/rtorrent-0.9.8-r1.ebuild b/net-p2p/rtorrent/rtorrent-0.9.8-r1.ebuild
index 7c4e641a1c2..0624ec56d31 100644
--- a/net-p2p/rtorrent/rtorrent-0.9.8-r1.ebuild
+++ b/net-p2p/rtorrent/rtorrent-0.9.8-r1.ebuild
@@ -59,6 +59,7 @@ src_install() {
        default
        doman doc/rtorrent.1

-       newinitd ${FILESDIR}/rtorrent-r1.init rtorrent
+       newinitd "${FILESDIR}/rtorrent-r1.init" rtorrent
+       newconfd "${FILESDIR}/rtorrentd.conf" rtorrent
        systemd_newunit "${FILESDIR}/rtorrentd_at-r1.service" "rtorrentd@.service"
 }
```

The only problem though is the contents of `/etc/conf.d/rtorrent`. If you set, for example, `USER=Alice`, everything works fine. However, if you try using the existing default of `USER=$USER`, then you get this error:
```
$ sudo rc-service rtorrent restart
 * Caching service dependencies ...                                                                                                                                                                            [ ok ]
 * Starting rtorrent ...
 * start-stop-daemon: user `--background' not found
 * Failed to start rtorrent                                                                                                                                                                                    [ !! ]
 * ERROR: rtorrent failed to start
```

It looks like the `$USER` value passed to `command_user` is empty for some reason, meaning that the `$USER` expansion in `/etc/conf.d/rtorrent` doesn't expand out properly with openrc. I don't know if this is a bug with my script or a bug with openrc itself.

Razorfang:
Yup, I just checked by printing the value of the args generated by `rc-service`.

Here's what the args are with the name "Alice" specified in `/etc/conf.d/rtorrent`:
```
start-stop-daemon --start --exec /usr/bin/rtorrent --pidfile /run/rtorrent.pid --user Alice --background --make-pidfile -- -o system.daemon.set=true 
```

And here's what the args are with the name "$USER" in `/etc/conf.d/rtorrent`, which should expand out but doesn't.
```
start-stop-daemon --start --exec /usr/bin/rtorrent --pidfile /run/rtorrent.pid --user --background --make-pidfile -- -o system.daemon.set=true '
```

Razorfang:
Alright, so one of the Gentoo developers told me [here](https://forums.gentoo.org/viewtopic-p-8690709.html#8690709) that we can't expand environment variables in the openrc config file with my changes.

Razorfang:
The long and short of it is:

- With the latest commit I added, we can now use the config file to set users again.
- This config file must have a defined value, and does not support expanding environment variables such as $USER. This is different from the old rtorrent daemon implementation which had $USER as its default value, which presumably worked.
- The wiki currently suggest replacing the default value value of $USER with a defined user [https://wiki.gentoo.org/wiki/RTorrent#Customizing_the_daemonizer](here)

thesamesam:
@perfinion

perfinion:
Thanks! EAPI6 is deprecated so I updated to EAPI8 while merging. the new EAPI passes sysroot which caused some issues so I also had to add a sed to change -syslibroot= to --sysroot.

PR #23646, opened 2022-01-04
dev-lang/julia: 1.7.1-r1

dzejej:
Add required version for libutf8proc and patchelf.
Add explicit option for  USE_SYSTEM_LIBBLASTRAMPOLINE
Realy use system libraries for sparse matrix


thesamesam:
Minor nit (you're doing great!): could you make the commit summary (first line) more descriptive?

e.g. "build fixes; use system libraries"

thesamesam:
FWIW, as per the QA / CI warnings, we prefer to not pin deps like this. Better to do >= if possible rather than =...* or similar.

PR #23652, opened 2022-01-04
dev-lang/julia: more LLVM 13 patches

dzejej:
I also reduce julia-1.7.{0->1}-do_not_set_rpath.patch
 - it means more binary files are patchelf

Bug: https://bugs.gentoo.org/819381


tamiko:
@dzejej I have also renamed all patches to better reflect the ebuild verions where they are used: https://github.com/gentoo/gentoo/commit/e13b594c70d5d85e0180f248de12224cd0139015

PR #23662, opened 2022-01-05
net-wireless/gnuradio: backported 3.8.4.0 qt compat

l29ah:
Closes: https://bugs.gentoo.org/829272


ZeroChaos-:
I will bump to 3.8.5.0 which includes this fix.  thanks for the poke

PR #23668, opened 2022-01-05
x11-misc/x2x: version bump to 1.30

rndxelement:
I Added an ebuild for x2x-1.30-10.
It uses the current upstream source, which now resides on GitHub (the old source is 404 forever).
I moved the patches into the files folder and checked which ones are actually needed and not already fixed upstream.
Only one patch is left, which I made compatible with the new version.

I refactored the ebuild, added the correct licenses and simplified.
I'm open for improvements of course.

Furthermore, I added myself as a proxied maintainer.

Closes: https://bugs.gentoo.org/754132


x11-misc/x2x/x2x-1.30.ebuild
```diff
@@ -0,0 +1,53 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit toolchain-funcs
+
+DESCRIPTION="A utility to connect the mouse and keyboard to another X"
+HOMEPAGE="https://github.com/dottedmag/x2x"
+SRC_URI="https://github.com/dottedmag/x2x/archive/refs/tags/debian/1.30-10.tar.gz"
```
Flowdalic, line 10: You should rename the tarball. Having a file named `1.30-10.tar.gz` is meaningless and also prone to name conflicts. Also note that you may want to use $PV here.

x11-misc/x2x/x2x-1.30.ebuild
```diff
@@ -0,0 +1,53 @@
...
+
+src_prepare() {
+	default
+
+	# Call the bootstrap script
+	./bootstrap.sh
+}
+
+src_configure() {
+	default
```
Flowdalic, line 42: A phase function whose body only consists of `default` does usually not need to be explicitly declared.

x11-misc/x2x/x2x-1.30.ebuild
```diff
@@ -0,0 +1,49 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit toolchain-funcs
+
+DESCRIPTION="A utility to connect the mouse and keyboard to another X"
+HOMEPAGE="https://github.com/dottedmag/x2x"
+SRC_URI="https://github.com/dottedmag/x2x/archive/refs/tags/debian/1.30-10.tar.gz -> ${P}.tar.gz"
```
Flowdalic, line 10:
```suggestion
SRC_URI="https://github.com/dottedmag/x2x/archive/refs/tags/debian/${PV}-10.tar.gz -> ${P}.tar.gz"
```

x11-misc/x2x/x2x-1.30.ebuild
```diff
@@ -0,0 +1,49 @@
...
+}
+
+src_compile() {
+	emake CC="$(tc-getCC)" CDEBUGFLAGS="${CFLAGS}" \
+		EXTRA_LDOPTIONS="${LDFLAGS}"
+}
+
+src_install() {
+	default
+	dodoc README AUTHORS INSTALL ChangeLog ChangeLog.old
```
Flowdalic, line 48: This should use the DOCS array (see `man 5 ebuild`). Then you do not need to explicitly declare the `src_install` phase function.

x11-misc/x2x/x2x-1.30.ebuild
```diff
@@ -0,0 +1,49 @@
...
+	# Patch to fix bug #126939
+	# AltGr does not work in x2x with different keymaps:
+	"${FILESDIR}"/${PN}_1.30-10-keymap.patch
+)
+
+src_prepare() {
+	default
+
+	# Call the bootstrap script
+	./bootstrap.sh
```
Flowdalic, line 38: This just calls `autoreconf -isv`, and hence I think you want to inherit the `autotools` class and simply call eautoreconf.

```
src_preapare() {
   default
   eautoreconf
}
```

Note that if you really need to invoke bootstrap.sh, then this invocation is missing a `|| die`.

Flowdalic (changes requested):

x11-misc/x2x/x2x-1.30.ebuild
```diff
@@ -0,0 +1,43 @@
...
+	>=x11-misc/imake-1.0.8-r1"
+
+S="${WORKDIR}"/${PN}-debian-1.30-10
+
+PATCHES=(
+	# Patch to fix bug #126939
+	# AltGr does not work in x2x with different keymaps:
+	"${FILESDIR}"/${PN}_1.30-10-keymap.patch
+)
+DOCS=( README AUTHORS INSTALL ChangeLog ChangeLog.old "$S"/docs/*.txt )
```
Flowdalic, line 33: Using `$S` in `DOCS` seems fragile. In fact, I get

```

Calculating dependencies -/home/flo/data/repos/gentoo/gentoo/x11-misc/x2x/x2x-1.30.ebuild: line 33: no match: /data-scratch/var-tmp/portage/x11-misc/x2x-1.30/work/x2x-debian-1.30-10/docs/*.txt
... done!
```
when emerging this. Why not just

```
DOCS=( README AUTHORS … docs/*.txt )
```

x11-misc/x2x/x2x-1.30.ebuild
rndxelement: Yeah, I was thinking in a wrong direction. This is way simpler.

x11-misc/x2x/x2x-1.30.ebuild
rndxelement: Yeah, I was thinking in a wrong direction. This is way simpler.

rndxelement:
The CI is failing with "failed sourcing ebuild: /var/lib/repo-mirror-ci/pull/tmp/x11-misc/x2x/x2x-1.30.ebuild: line 33: no match: docs/*.txt". Why is that?

Flowdalic:
Turns out that glob in `DOCS` is a bad idea. I've slightly modified the ebuild. Thanks for your contribution.

rndxelement:
Thanks a lot for your help.

PR #23669, opened 2022-01-06
games simulation/simutrans: Version bump 123.0.1

tastytea:
- Switched to upstream versioning (no idea why we prefixed it with 0)
- All PakSets but the default one are now in
  games-simulation/simutrans-paksets to make updating them without
  recompiling simutrans possible


tastytea:
Please don't merge until <https://bugs.gentoo.org/830664> is resolved.

juippis (commented):
Has this PR been updated according to the bug? Looks like it is. 

I'm not a fan of these USE flags here but they seem reasonable enough to keep. 

Let me know is this ready?

tastytea:
I removed the pak64.japan pakset with the unusual license because the legal situation is still not clear.

@juippis It is ready now.

juippis (approved):
Looks good but before merging, can you say if this is worrysome or not:
```
# simutrans
dr_fatal_notify: ERROR: Failed to initialize backend.
```
Also getting
```
./configure: line 5531: svn: command not found
```
for simutrans-123.0.1

tastytea:
> ```
> # simutrans
> dr_fatal_notify: ERROR: Failed to initialize backend.
> ```

Is that in a test environment? That's likely because no X server / wayland compositor is installed. I figured that everyone wanting to play this game has one of the two installed, and most other SDL games seem to handle it the same.

> Also getting
> 
> ```
> ./configure: line 5531: svn: command not found
> ```

That's expected. `svn` is used to get the revision (for online gaming), which wouldn't work of course. I [set that manually](https://github.com/gentoo/gentoo/pull/23669/files#diff-0ac147257ffb2711dfd0856447bb0e9cf6de8a716dbcfc3bad23d14ebce5c1abR11) instead

juippis:
> Is that in a test environment? That's likely because no X server / wayland compositor is installed. I figured that everyone wanting to play this game has one of the two installed, and most other SDL games seem to handle it the same.

Figures.

> 
> > Also getting
> > ```
> > ./configure: line 5531: svn: command not found
> > ```
> 
> That's expected. `svn` is used to get the revision (for online gaming), which wouldn't work of course. I [set that manually](https://github.com/gentoo/gentoo/pull/23669/files#diff-0ac147257ffb2711dfd0856447bb0e9cf6de8a716dbcfc3bad23d14ebce5c1abR11) instead

The main worry is just that will the outcome change if you do/don't have svn installed? Maybe something to be patched out, if found to be problematic. 

Thanks!

tastytea:
> The main worry is just that will the outcome change if you do/don't have svn installed? Maybe something to be patched out, if found to be problematic.

As far as I can see, the build system first tests if it is in a svn repo before it tries to do anything with it. Should be fine.

PR #23671, opened 2022-01-06
profiles/features/musl: mask dev-db/oracle-instantclient

fedeliallalinea:
Closes: https://bugs.gentoo.org/830659


thesamesam:
We should mask any use flags which drag this in too

thesamesam:
Thanks! Note that mask entries should go to the top of the files, not the bottom, but I'll do that on merge.

PR #23674, opened 2022-01-06
net-libs/libmicrohttpd: fixed missing multilib dep

Karlson2k:
Closes: https://bugs.gentoo.org/830555


blueness:
Thanks for catching this.

PR #23677, opened 2022-01-06
net-im/telegram-desktop: Bump to 3.4.3

mid-kid:
@gyakovlev

Feel free to keyword PPC64 whenever you can. RISCV keywording would be nice too, but I'm not sure you're able to test.


gyakovlev:
LGTM, re-keyworded ppc64, builds fine.

I've updated copyright years for you, if you touch ebuild in 2022 - update it =)
pkgcheck will scream at you and pkgdev can  do it for you (used as `pkgdev commit` instead of `repoman commit`)

gyakovlev:
as for riscv - I do have hardware, but it runs headless, I don't really want to build all the deps on it =) will take a week probably.
so let's just open keywording bug if you want it keyworded.

gyakovlev:
https://bugs.gentoo.org/830712

PR #23687, opened 2022-01-07
Remove oracle-instantclient-{basic,jdbc,odbc,sqlplus}

fedeliallalinea:
Commits that allow you to delete old oracle-instantclient-* packages.
These packages are now only found as dependencies in the 
dev-db/oracle-instantclient unified package so that it gives conflict if someone has still installed them (I don't know if this is the correct way to proceed).


thesamesam:
Thanks for working on this.

Note that each of these needs to be revbumped to ensure that users are migrated to the newer packages (especially, but not at all limited to, those who do partial upgrades and so on).

fedeliallalinea:
> Note that each of these needs to be revbumped to ensure that users are migrated to the newer packages (especially, but not at all limited to, those who do partial upgrades and so on).

Even the live ebuilds or just the ones with a version?

> Not sure if I fully get your comment about conflicts. Could you explain a bit more?

I see now that old ebuilds are only as metapackages, so I don't see any problems (I was thinking of file collisions).

Dessa:
the net-analyzer/zabbix change seems odd. the ewarn either needs to be reworked with proper useflags or removed entirely (honestly probably the latter). i don't use oracle-instantclient so i can't help there, @fordfrog ?

fedeliallalinea:
> the net-analyzer/zabbix change seems odd. the ewarn either needs to be reworked with proper useflags or removed entirely (honestly probably the latter). i don't use oracle-instantclient so i can't help there, @fordfrog ?

I saw that but I don't understand what meant by installing all oracle packages. I doubt sqlplus is needed for any purpose, probably you should to put dev-db/oracle-instantclient[jdbc,sdk] or dev-db/oracle-instantclient[odbc,sdk] as dependency.

thesamesam:
> > Note that each of these needs to be revbumped to ensure that users are migrated to the newer packages (especially, but not at all limited to, those who do partial upgrades and so on).
> 
> Even the live ebuilds or just the ones with a version?
> 


Just versioned ones

> > Not sure if I fully get your comment about conflicts. Could you explain a bit more?
> 
> I see now that old ebuilds are only as metapackages, so I don't see any problems (I was thinking of file collisions).

thesamesam:
We need to last-rite (package.mask) the old package(s) for 30 days and send last-rites on gentoo-dev* (the developer who merges this will handle that part for you)

fedeliallalinea:
> We need to last-rite (package.mask) the old package(s) for 30 days and send last-rites on gentoo-dev* (the developer who merges this will handle that part for you)

Can/should I send the message on the mailing list?

PR #23694, opened 2022-01-08
sys-boot/os-probe fix for kernel 5.15 with ntfs3

sergeygalkin:
Bug: https://bugs.gentoo.org/830753
Closes: https://bugs.gentoo.org/830753
Signed-off-by: Sergei Galkin sergey.gals@gmail.com

### DIFF
--- a/sys-boot/os-prober/os-prober-1.79-r2.ebuild
+++ b/sys-boot/os-prober/os-prober-1.79-r2.ebuild
@@ -1,0 +1,105 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+inherit readme.gentoo-r1 toolchain-funcs
+
+DESCRIPTION="Utility to detect other OSs on a set of drives"
+HOMEPAGE="https://salsa.debian.org/installer-team/os-prober"
+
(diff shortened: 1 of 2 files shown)


sys-boot/os-prober/os-prober-1.79-r2.ebuild
```diff
@@ -0,0 +1,105 @@
+# Copyright 1999-2021 Gentoo Authors
```
Peter-Levine, line 1: repoman complains about the header.  Just change the date from `1999-2021` to `1999-2022`.

Peter-Levine (changes requested):
Very good. Thanks.  Try to use [repoman](https://wiki.gentoo.org/wiki/Repoman) to check and apply commits to check any QA issues.  And no need to use `Bug:` and `Closes:` tags for the same commit.  If it's intended to close the bug, and not just cite it, opt for `Closes:` only.

sys-boot/os-prober/os-prober-1.79-r2.ebuild
sergeygalkin: Thanks Peter, all fixed

Peter-Levine (approved):

PR #23697, opened 2022-01-08
x11-themes/arc-theme: add 20220105 & co-maintain

gyakovlev:
Hey

good news.
I actually use it so I can co-maintain if you don't mind.
new version no longer requires inkscape ( for release tarballs)
and fixes crashes of some apps with gtk-4.6 

```
 Gdk-ERROR **: 00:55:43.947: Resource path /org/gnome/arc-theme/assets/titlebutton-close.png s not a valid image: Unrecognized image file format
 ```

here's the diff

```diff
--- arc-theme-20211018.ebuild	2021-10-23 11:04:16.585302702 -0700
+++ arc-theme-20220105.ebuild	2022-01-09 14:18:54.590080008 -0800
@@ -1,63 +1,46 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
 
-# USE="-* gtk2 gtk3 gtk4 xfce" ebuild ${P}.ebuild clean compile
-# cd $(portageq envvar PORTAGE_TMPDIR)/portage/x11-themes/${P}/work
-# find ${P}-build/ -name "*.png" | xargs tar Jcvf /usr/portage/distfiles/${P}-pngs.tar.xz --owner=root --group=root
+PYTHON_COMPAT=( python3_{8..10} )
 
-inherit meson toolchain-funcs
+inherit meson python-any-r1
 
 DESCRIPTION="A flat theme with transparent elements for GTK 2/3/4 and GNOME Shell"
 HOMEPAGE="https://github.com/jnsh/arc-theme"
-SRC_URI="https://github.com/jnsh/${PN}/releases/download/${PV}/arc-theme-${PV}.tar.xz
-	pre-rendered? ( https://dev.gentoo.org/~chewi/distfiles/${P}-pngs.tar.xz )"
+SRC_URI="https://github.com/jnsh/${PN}/releases/download/${PV}/arc-theme-${PV}.tar.xz"
+
 LICENSE="GPL-3"
 SLOT="0"
 KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
-IUSE="cinnamon gnome-shell +gtk2 +gtk3 +gtk4 mate +pre-rendered xfce"
-
-SASSC_DEPEND="
-	dev-lang/sassc
-"
+IUSE="cinnamon gnome-shell +gtk2 +gtk3 +gtk4 mate +transparency xfce"
 
-SVG_DEPEND="
-	!pre-rendered? (
-		media-gfx/inkscape
-		media-gfx/optipng
-	)
-"
+GLIB_DEPEND="dev-libs/glib"
+SASSC_DEPEND="dev-lang/sassc"
 
-# Supports various GTK+3, GNOME Shell, and Cinnamon versions and uses
-# pkg-config to determine which set of files to build. Updates will
-# therefore break existing installs but there's no way around this. At
-# least GTK+3 is unlikely to see a release beyond 3.24.
+# Supports various GTK, GNOME Shell, and Cinnamon versions and uses
+# --version option for gnome-shell and cinnamon to determine which set of files to build.
+# Updates will therefore break existing installs but there's no way around this.
 BDEPEND="
+	${PYTHON_DEPS}
 	>=dev-util/meson-0.56.0
 	cinnamon? (
 		${SASSC_DEPEND}
 		gnome-extra/cinnamon
 	)
 	gnome-shell? (
+		${GLIB_DEPEND}
 		${SASSC_DEPEND}
-		>=gnome-base/gnome-shell-3.18
-	)
-	gtk2? (
-		${SVG_DEPEND}
+		>=gnome-base/gnome-shell-3.28
 	)
 	gtk3? (
+		${GLIB_DEPEND}
 		${SASSC_DEPEND}
-		${SVG_DEPEND}
-		virtual/pkgconfig
-		=x11-libs/gtk+-3.24*:3
 	)
 	gtk4? (
+		${GLIB_DEPEND}
 		${SASSC_DEPEND}
-		${SVG_DEPEND}
-	)
-	xfce? (
-		${SVG_DEPEND}
 	)
 "
 
@@ -71,42 +54,28 @@
 "
 
 src_configure() {
+	# Cinnamon still uses metacity themes for its window manager.
+	# so we enable metacity theme too if USE=cinnamon
+	# but only enable metacity if USE=mate
 	local themes=$(
 		printf "%s," \
-		$(usev cinnamon) \
+		$(usev cinnamon "cinnamon metacity") \
 		$(usev gnome-shell) \
 		$(usev gtk2) \
 		$(usev gtk3) \
 		$(usev gtk4) \
-		$(usex mate metacity "") \
-		$(usex xfce xfwm "")
+		$(! use cinnamon && usev mate metacity) \
+		$(usev xfce xfwm)
 	)
 
 	local emesonargs=(
+		# 4.0 dropped in Gentoo, 4.2 works with 4.6
+		# if we change that to allow auto-detection, a pkg-config dep will be required.
+		-Dgtk4_version=4.2
 		-Dthemes="${themes%,}"
-		-Dgtk3_version=3.24
-		-Dgtk4_version=4.2 # 4.0 dropped in Gentoo, works with 4.4.
+		$(meson_use gnome-shell gnome_shell_gresource)
+		$(meson_use transparency)
 	)
 
-	if use pre-rendered; then
-		emesonargs+=(
-			$(if tc-is-cross-compiler; then
-				  echo --cross-file
-			  else
-				  echo --native-file
-			  fi)
-			"${FILESDIR}"/pre-rendered.ini
-		)
-	fi
-
 	meson_src_configure
 }
-
-src_compile() {
-	# fontconfig issue?
-	# https://bugs.gentoo.org/666418#c28
-	use pre-rendered ||
-		addpredict "${BROOT}"/usr/share/inkscape/fonts/.uuid.TMP-XXXXXX
-
-	meson_src_compile
-}
```


gyakovlev:
I think it may need a gtk4 dep as it uses it to query version, specifying version is no longer supported.

jnsh:
> I think it may need a gtk4 dep as it uses it to query version, specifying version is no longer supported.

Specifying version for gtk4 is still supported. It has only been dropped from gtk3 since the theme only ships with theme for version 3.24, and with that, you don't any more need `pkgconfig` and `gtk+-3.24` for building the gtk3 theme.

If you want to auto-detect version for the gtk4 theme, it requires `gtk4` and `pkgconfig`. However, with the current version there's only one gtk4 theme version available, so you could just specify the version with `-Dgtk4_version=` to avoid these dependencies.

Also, `glib` should only be required for building themes that compile the theme in a .gresource file, i.e. gtk3, gtk4 and gnome-shell.

gyakovlev:
thanks for chiming in @jnsh =)
always nice to see upstream maintainers here.

I think I've sorted out deps correctly now.
it builds fine without gtk3 or gtk4 installed at all.
and I added glib dep only if gnome-shell is enabled and gresource is requested.

jnsh:
`glib` is _always_ required for gtk3 and gtk4 themes as well, since they are compiled into a gresource.

Apart from that, things look correct to me now. Although it's been awhile since I've used gentoo, so I'm not too familiar with the current EAPI and best practices :)

gyakovlev:
> `glib` is _always_ required for gtk3 and gtk4 themes as well, since they are compiled into a gresource.
> 

```
# glib-compile-resources dependency
foreach theme : ['gnome-shell', 'gtk3', 'gtk4']
  if theme in get_option('themes')
    if theme == 'gnome-shell' and not get_option('gnome_shell_gresource')
      continue
    endif
    glib_compile_resources = find_program('glib-compile-resources')
    break
  endif
endforeach
```

ah I see I misread this `continue` here. it skips `gresource` for gnome-shell if not requested, but always wants it for gtk3/4.
not that's it's possible to get  gtk without glib, but let's fix it while at it =)

> Apart from that, things look correct to me now. Although it's been awhile since I've used gentoo, so I'm not too familiar with the current EAPI and best practices :)
> 
> EDIT: One more thing, the metacity theme should probably be compiled in with USE="cinnamon" as well, since Cinnamon still uses metacity themes for its window manager. Currently it's only built for MATE, if I've not mistaken..
I'll see what I can do with that.
we can require USE=mate if USE=cinnamon is set, but that's not very user-friendly tbh.

chewi:
Thanks for doing this. I had started looking at it myself but didn't get that far. I was in two minds about whether to still support building the PNGs, but it just doesn't seem worth it. I doubt anybody cares.

I trust you'll get this right between the two of you so feel free to merge when you're done.

chewi (approved):

jnsh:
> I was in two minds about whether to still support building the PNGs, but it just doesn't seem worth it. I doubt anybody cares.

Do you have some specific use case in mind where this could be useful? I've been wondering whether I should add an option to force re-rendering the assets to the build system, but I'm not sure if that'd be worth it. Only case I could think of is if anyone wants to edit the assets before building (e.g. change colors), but that doesn't sound relevant for the gentoo package.

Either way, if you ever want to do this, removing the pre-rendered PNG files should cause them to get re-rendered during build. You can see the locations for the rendered PNGs in the [render-asset.py](https://github.com/jnsh/arc-theme/blob/master/meson/render-assets.py#L13) script (the third item in each list is the output directory for rendered assets, and can simply `rm *.png` each dir).

The latest ebuild revision seems ok as far as I can tell btw :)

gyakovlev:
thanks everyone for participation and feedback.
merged it.

chewi:
> Do you have some specific use case in mind where this could be useful?

I occasionally custom patch the odd package that goes into my system, and it's nice that Gentoo makes this easy. I'd like to think you could do that with a package like this too, but like I said, I doubt anybody cares.

> Either way, if you ever want to do this, removing the pre-rendered PNG files should cause them to get re-rendered during build. You can see the locations for the rendered PNGs in the [render-asset.py](https://github.com/jnsh/arc-theme/blob/master/meson/render-assets.py#L13) script.

Yeah, I was considering that, but ideally we wouldn't need to hard code the paths in the ebuild. Doesn't matter. :slightly_smiling_face:

PR #23708, opened 2022-01-09
dev-java/jna-5.10.0: DEPEND on x11-libs/libXt

Leo3418:
Resolves <https://bugs.gentoo.org/830812>


fordfrog:
@Leo3418 merged, thanks!

PR #23714, opened 2022-01-09
media-gfx/gimp: 9999 fix fail due to migration from gtk-doc to gi-docgen

band-a-prend:
Fixes `media-gfx/gimp-9999` with `USE="doc"` due to changes of appropriate dependencies.
Closes: https://bugs.gentoo.org/830429

Additional DEPEND/RDEPEND small fixes:
Closes: https://bugs.gentoo.org/826578
Closes: https://bugs.gentoo.org/824218


novomesk:
I tried this `gimp-9999.ebuild` and it works well for me.

PR #23716, opened 2022-01-09
profiles: last-rite most of sys-fabric/

SoapGentoo:

### DIFF
--- a/profiles/package.mask
+++ b/profiles/package.mask
@@ -33,6 +33,34 @@
 
 #--- END OF EXAMPLES ---
 
+# David Seifert <soap@gentoo.org> (2022-01-16)
+# Most of these packages have been merged into sys-cluster/rdma-core and
+# have long been dead and unmaintained.
+# Removal in 30 days, bug #830999.
+sys-fabric/ibacm
+sys-fabric/infiniband-diags
+sys-fabric/infinipath-psm
+sys-fabric/libcxgb3
+sys-fabric/libcxgb4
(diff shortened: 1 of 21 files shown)


profiles/package.mask
```diff
@@ -33,6 +33,29 @@
 
 #--- END OF EXAMPLES ---
 
+# David Seifert <soap@gentoo.org> (2022-01-09)
+# No other distro packages these, long dead and unmaintained.
```
Flowdalic, line 37: You may want to point out that at least some of those are now part of sys-cluster/rdma-core (sys-fabric/infiniband-diags for example).

SoapGentoo: thanks, I will improve the message once the PR is green

profiles/package.mask
```diff
@@ -33,6 +33,34 @@
 
 #--- END OF EXAMPLES ---
 
+# David Seifert <soap@gentoo.org> (2022-01-16)
+# Most of these packages have been merged into sys-cluster/rdma-core and
+# have long been dead and unmaintained.
+# Removal in 30 days, bug #601692.
```
jsmolic, line 39: Could we also tag https://bugs.gentoo.org/830999 here

jsmolic (approved):
Thanks, let's ship this finally :tada:

jsmolic:
Can you close https://bugs.gentoo.org/831003 with the 2nd commit while at it

PR #23717, opened 2022-01-09
dev-python/pipenv: fix missing sed command

oz123:
Unfortunately, shells.py has one path not covered
by unit tests. Hence, this was not caught earlier.

The last sed added handels such cases:

```
from .vendor import pexpect
```


oz123:
This is obsoleted by #23720

PR #23722, opened 2022-01-10
media-gfx/hydrus: Fix LCMS error for 468 too, copied the wrong file

vaartis:
Copied the wrong file when bumping

Closes: https://bugs.gentoo.org/830827


vaartis:
Also added the latest release.

PR #23735, opened 2022-01-11
dev-util/bats: Stabilize 1.5.0 ALLARCHES, #828355

henning-schild:
Closes: https://bugs.gentoo.org/828355

### DIFF
--- a/dev-util/bats/bats-1.5.0.ebuild
+++ b/dev-util/bats/bats-1.5.0.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=8
@@ -12,7 +12,7 @@ SRC_URI="https://github.com/${MY_PN}/${MY_PN}/archive/v${PV}.tar.gz -> ${P}.tar.
 
 LICENSE="MIT"
 SLOT="0"
-KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~riscv ~x86"
+KEYWORDS="amd64 arm arm64 ppc64 ~riscv x86"
 
 DEPEND="app-shells/bash:*"
 RDEPEND="${DEPEND}"


henning-schild:
Not sure i am doing this the right way, let me know if i am missing any steps. Just copying a pattern found in 5d93369522f07418b8fed2c902ada19653feb314

henning-schild:
@thesamesam

henning-schild:
Not doing ppc64 because there is no stable "sys-process/parallel" and not doing riscv because there is not even a keyword on "sys-process/parallel"

41479a8f50a379e7c23c60f470ba10708c8b9570 should probably have added `~riscv` to "sys-process/parallel" @mkszuba

thesamesam:
> Not doing ppc64 because there is no stable "sys-process/parallel" and not doing riscv because there is not even a keyword on "sys-process/parallel"
> 
> [41479a8](https://github.com/gentoo/gentoo/commit/41479a8f50a379e7c23c60f470ba10708c8b9570) should probably have added `~riscv` to "sys-process/parallel" @mkszuba

But it doesn't strictly depend on parallel?

The depgraph wouldn't allow it and CI would fail if a dependency wasn't satisfied.

Note that we usually do stabilisations through arch testers in bugs although I guess it bothers me less for ALLARCHES packages.

henning-schild:
> Well, no, if we don't stable it now, we can't cleanup / ppc64 users are stuck on N old version. We can't just drop stable keywords for an arch for no reason, especially given a package might need it (or a user). It's stable there for a reason.

I know, nobody was talking about dropping anything yet.

> Not sure if you'd reluctance to do ppc64 here.. I've just added CC-ARCHES to the bug so it's tested normally.

Not reluctant, just have no clue how that works and when things are allowed to be stable and who decides that. The other bugs of that sort have been resolved by other people.

I will add all but "riscv" ... meaning there will be "arm" which was not in 1.3.0 ... and we will effectively skip 1.4.1, after this is done we can drop 1.3.0 and 1.4.1 because the greatest will be stable ... but i will do that in another MR i guess

thesamesam:
> > Well, no, if we don't stable it now, we can't cleanup / ppc64 users are stuck on N old version. We can't just drop stable keywords for an arch for no reason, especially given a package might need it (or a user). It's stable there for a reason.
> 
> I know, nobody was talking about dropping anything yet.
> 
> > Not sure if you'd reluctance to do ppc64 here.. I've just added CC-ARCHES to the bug so it's tested normally.
> 
> Not reluctant, just have no clue how that works and when things are allowed to be stable and who decides that. The other bugs of that sort have been resolved by other people.
> 

it's generally the same arches as the previous version unless someone requested a new one in a bug for some reason.

> I will add all but "riscv" ... meaning there will be "arm" which was not in 1.3.0 ... and we will effectively skip 1.4.1, after this is done we can drop 1.3.0 and 1.4.1 because the greatest will be stable ... but i will do that in another MR i guess

All that needs to be done is the precious stable arches are re-tested and individually marked stable for the new version. ALLARCHES is for packages with architecture independent code where the old stable keywords can be copied to the newer version if it is tested on just one architecture.

I have CCed the needed arches in the bug by adding the CC-ARCHES keyword to it.

thesamesam:
(and jsmolic has just tested it.)

henning-schild:
> Maintainers are not required to self stabilise.

Mhh i was assigned the bug, that is why i started this MR.

thesamesam:
> > Maintainers are not required to self stabilise.
> 
> Mhh i was assigned the bug, that is why i started this MR.

Maintainers are assigned stabilisation bugs and need to acknowledge them / add "CC-ARCHES" (I mentioned this on the bug) to the bug KEYWORDS. Usually don't need to self stabilise unless it's awkward to test.

henning-schild:
> We can try make the dev manual clearer if you have suggested phrasing?

Let us wait until next time. I hope i will remember the process or actual read that dev manual. At the moment i am just a proxy maint and simply was not aware of the steps and did not read the manual. Meaning that manual might be fine, i just did not read it.

PR #23738, opened 2022-01-11
sci-mathematics/gimps: add 32 bit binary

PPed72:
Closes: https://bugs.gentoo.org/830996


thesamesam:
Nice, thanks!

PR #23741, opened 2022-01-11
pecl-mongodb: add 1.11

ullrichp:
Add ebuild for pecl-mongodb-1.11.0 and update necessary deps.


dev-libs/libbson/libbson-1.19.1.ebuild
```diff
@@ -0,0 +1,48 @@
...
+SRC_URI="https://github.com/mongodb/mongo-c-driver/releases/download/${PV}/mongo-c-driver-${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="amd64 ~hppa ~ppc ~riscv ~sparc x86"
+IUSE="examples static-libs"
+
+DEPEND="dev-python/sphinx
+	>=dev-python/docutils-0.14
+	<dev-python/docutils-0.18"
```
ullrichp, line 19: I am absolutely not sure about that but the build breaks with this error:
```
pkg_resources.DistributionNotFound: The 'docutils<0.18,>=0.14' distribution was not found and is required by Sphinx
```
when installing with docutils-0.18.1.

thesamesam:
Hi!

Could you do one commit per package? You can do this by doing `git reset origin/master` (no `--hard`), then git add in each directory individually, then run `pkgdev commit` or `repoman commit`.

This will also ensure the commit messages are prefixe correctly with `cat/pkg: foo` (e.g. `dev-php/pecl-mongodb: add 1.11`).

The commit messages will also need a signoff like `Signed-off-by: real name <email>`.

PR #23742, opened 2022-01-11
sci-libs/silo: readd test supports (not ready yet)

scardracs:


sci-libs/silo/silo-4.11.ebuild
```diff
@@ -13,13 +13,14 @@ S="${WORKDIR}/${P}-bsd"
 LICENSE="BSD"
 SLOT="0"
 KEYWORDS="~amd64 ~x86"
-IUSE="browser +hdf5 +silex"
-# Waiting for fix/answer upstream
-# See https://github.com/LLNL/Silo/issues/234
-RESTRICT="test"
+IUSE="browser +hdf5 +silex test"
```
jsmolic, line 16: I don't think we need to add `test` flag here as you don't have any test-specific deps or are using `test` use flag anywhere in the ebuild

sci-libs/silo/silo-4.11.ebuild
```diff
@@ -13,13 +13,14 @@ S="${WORKDIR}/${P}-bsd"
...
+RESTRICT="!test? ( test )"
 
 RDEPEND="
+	dev-qt/qtcore:5
```
jsmolic, line 20: How come you're adding new deps? This needs a revbump if we're changing RDEPEND

scardracs:
@jsmolic see bugs linked for added deps

sci-libs/silo/silo-4.11.ebuild
scardracs: test is required or repoman throws that error  

```
RepoMan scours the neighborhood...
>>> Creating Manifest for /home/scardracs/git/gentoo/sci-libs/silo
  IUSE.missing [fatal]          1
   sci-libs/silo/silo-4.11-r1.ebuild: RESTRICT: USE flag 'test' referenced in conditional '!test?' is not in IUSE

Note: use --include-dev (-d) to check dependencies for 'dev' profiles

Please fix these important QA issues first.
RepoMan sez: "Make your QA payment on time and you'll never see the likes of me."

```

jsmolic: Just drop the RESTRICT clause too, we need it only if we're using the `test` USE flag

jsmolic (approved):

PR #23749, opened 2022-01-12
net-im/slack: version bump (4.23.0)

fldu:
Bump slack version

Signed-off-by: Florian Dubourg <florian@dubourg.cz>


fldu:
Removed the trailing blank line

juippis (approved):
I've probably asked before but... why does this need to be multilibbed?

PR #23750, opened 2022-01-12
distutils-r1.eclass: PEP517 mode

mgorny:
CC @gentoo/python 

Quick summary:
1. PEP517 mode is enabled via `DISTUTILS_USE_PEP517`. This variable is also used to specify the build system, so the eclass adds the correct deps.
2. Setuptools used "hybrid" mode, that is we call `setup.py build` first (which takes advantage of jobs), then use PEP517 build.
3. We do complete wheel build+install in `python_compile()`.
4. The staging directory gets venv features. `distutils_install_for_testing` is no longer necessary.
5. `DISTUTILS_USE_SETUPTOOLS` is gone, as nowadays it only worked to distinguish between pure bdep and bdep+rdep, when packages explicitly specified it in install_requires. Nowadays you need to do rdep yourself if you need it.
6. PEP517 mode also enables vendored distutils in setuptools.


mgorny:
I'm thinking `DISTUTILS_PEP517` should probably take a build system name, and add appropriate deps. This will also mean we verify it against `pyproject.toml` (like we did for setuptools). `DISTUTILS_USE_SETUPTOOLS` should probably be banned when pep517 is used, and people need to deal with runtime deps explicitly.

There's also a question of how it relates to vendored distutils. I suppose we could skip the intermediate step and assume non-PEP517 = stdlib distutils, PEP517 = vendored distutils. This would avoid the egg-info mess (since PEP517 switches to dist-info).

eclass/distutils-r1.eclass
```diff
@@ -78,7 +78,28 @@ esac
...
+# in python_compile(), venv-style install tree is provided
+# to python_test() and python_install() just merges the temporary
+# install tree into real fs.
+#
+# The variable specifies the build system used.  Currently,
+# the following values are supported:
+# - flit - flit_core backend
+# - setuptools - distutils or setuptools (incl. legacy mode)
+# - standalone - standalone build systems without external deps
+#                (used for bootstrapping).
```
arthurzam, line 96: Could you run doc generation for eclass into man pages. I have a feeling we need extra newline between items in the list.

arthurzam (approved):
Thank you very much

eclass/distutils-r1.eclass
```diff
@@ -751,16 +875,90 @@ distutils-r1_python_compile() {
...
+
+		# python likes to compile any module it sees, which triggers sandbox
+		# failures if some packages haven't compiled their modules yet.
+		addpredict "${EPREFIX}/usr/lib/${EPYTHON}"
+		addpredict /usr/lib/pypy3.8
+		addpredict /usr/lib/portage/pym
+		addpredict /usr/local # bug 498232
+
+		local -x WHEEL_BUILD_DIR=${BUILD_DIR}/wheel
+		mkdir -p "${WHEEL_BUILD_DIR}"
```
SoapGentoo, line 906:
```suggestion
		mkdir -p "${WHEEL_BUILD_DIR}" || die
```

SoapGentoo (approved):

PR #23763, opened 2022-01-12
app-cdr/cdrdao: new EAPI, new maintainer

madmartin:
add dependency gnome-base/gconf
drop keyword hppa due to gnome-base/gconf
drop support for _rc ebuild

Closes: https://bugs.gentoo.org/745351

### DIFF
--- a/app-cdr/cdrdao/cdrdao-1.2.4-r1.ebuild
+++ b/app-cdr/cdrdao/cdrdao-1.2.4-r1.ebuild
@@ -1,0 +1,51 @@
+# Copyright 2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit autotools flag-o-matic
+
+DESCRIPTION="Burn CDs in disk-at-once mode -- with optional GUI frontend"
+HOMEPAGE="http://cdrdao.sourceforge.net/"
+SRC_URI="mirror://sourceforge/${PN}/${P}.tar.bz2"
+
+LICENSE="GPL-2+"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~ia64 ~ppc ~ppc64 ~riscv ~sparc ~x86"
+IUSE="encode mad vorbis"
+
+DEPEND="app-cdr/cdrtools
+	encode? ( >=media-sound/lame-3.99 )
+	mad? (
+		media-libs/libao
+		media-libs/libmad
+	)
+	vorbis? (
+		media-libs/libao
+		media-libs/libvorbis
+	)"
+RDEPEND="${DEPEND}
+	!app-cdr/cue2toc"
+BDEPEND="gnome-base/gconf
+	virtual/pkgconfig"
+
+PATCHES=(
+	"${FILESDIR}/${P}-ax_pthread.patch"
+	"${FILESDIR}/${P}-wformat-security.patch"
+)
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	append-cxxflags -std=c++11
+	local myeconfargs=(
+		--without-gcdmaster
+		$(use_with vorbis ogg-support)
+		$(use_with mad mp3-support)
+		$(use_with encode lame)
+	)
+	econf "${myeconfargs[@]}"
+}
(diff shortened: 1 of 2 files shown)


app-cdr/cdrdao/cdrdao-1.2.4-r1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+	mad? (
+		media-libs/libao
+		media-libs/libmad
+	)
+	vorbis? (
+		media-libs/libao
+		media-libs/libvorbis
+	)"
+DEPEND="${COMMON_DEPEND}
+	virtual/pkgconfig"
```
thesamesam, line 29: BDEPEND

app-cdr/cdrdao/cdrdao-1.2.4-r1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+DESCRIPTION="Burn CDs in disk-at-once mode -- with optional GUI frontend"
+HOMEPAGE="http://cdrdao.sourceforge.net/"
+SRC_URI="mirror://sourceforge/${PN}/${P}.tar.bz2"
+
+LICENSE="GPL-2+"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~ia64 ~ppc ~ppc64 ~riscv ~sparc ~x86"
+IUSE="encode mad vorbis"
+
+COMMON_DEPEND="
```
thesamesam, line 17: We can just drop COMMON_DEPEND and fold it in to DEPEND/RDEPEND now we have BDEPEND, in most cases (like here).

app-cdr/cdrdao/cdrdao-1.2.4-r1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	# Fix building with latest libsigc++
+	append-cxxflags -std=c++11
+	find -name '*.h' -exec sed -i '/sigc++\/object.h/d' {} + || die
```
thesamesam, line 47: Things like this would be better off with a patch tbh but fine to do that on next bump if it ever happens.

app-cdr/cdrdao/cdrdao-1.2.4-r1.ebuild
```diff
@@ -0,0 +1,56 @@
...
+	"${FILESDIR}/${P}-wformat-security.patch"
+)
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	# Fix building with latest libsigc++
```
thesamesam, line 45: Would be worth digging up a bug reference for this and adding it to the comment so we can easily check if it's relevant and drop it in future, if possible.

app-cdr/cdrdao/metadata.xml
```diff
@@ -1,10 +1,16 @@
...
+		<name>Martin Dummer</name>
+	</maintainer>
+	<maintainer type="project" proxied="proxy">
+		<email>proxy-maint@gentoo.org</email>
+		<name>Proxy Maintainers</name>
+	</maintainer>
 	<longdescription>
-	CD command line recording, ripping and copying tool.	Especially ISOs- and
-	bin/cue-files are handled very well.
+		CD command line recording, ripping and copying tool. Especially ISO- and bin/cue-files are handled very well.
```
thesamesam, line 13:
```suggestion
		CD command line recording, ripping and copying tool. Especially ISOs and bin/cue-files are handled very well.
```

app-cdr/cdrdao/cdrdao-1.2.4-r1.ebuild
madmartin: done

app-cdr/cdrdao/cdrdao-1.2.4-r1.ebuild
madmartin: yeah... examining this showed it can be removed at all!

app-cdr/cdrdao/metadata.xml
madmartin: done

juippis (approved):

PR #23764, opened 2022-01-12
dev-python/kombu: fix amqp min version

tgbugs:
a runtime error is produced because kombu requirements lists amqp-5.0.9
as the min version instead of amqp-5.0.6

Signed-off-by: Tom Gillespie <tgbugs@gmail.com>


sbraz:
Can you please revbump and fix the copyright year?

sbraz:
Thanks, did you make sure tests pass with 3.10?

If so, can you please edit the commit message to mention the COMPAT addition? It's a nontrivial change and should be highlighted.

> edit: If PYTHON_COMPAT was updated do we technically need the revbump?
It's a controversial question but I think most of us do not revbump when we change it.

sbraz:
Looks like one dependency only lacks Python 3.10. Feel free to either test it and add it to this PR or just drop the COMPAT change.

tgbugs:
I went with dropping the COMPAT change and keeping the revbump because there are a number of `test` dependencies that need to have 3_10 enabled. Running the tests on 3_10 with just Pyro4 and a few other components enabled passed, but there were a bunch of tests that were skipped. So for now keeping it at 3_9 until the rest of the deps catch up is best.

sbraz:
Weird, I thought https://qa-reports.gentoo.org/output/gpyutils/39-to-310.svg also showed test dependencies. In any case, this is merged, thanks!

tgbugs:
You are correct. kazoo is the only one missing, but good we got this fix in since there was a runtime error. I'll update the kazoo compat and run the tests and create a separate pr.

PR #23768, opened 2022-01-13
media-sound/flacon: bump version to 8.2.0

kernlpanic:
Closes: https://bugs.gentoo.org/753164
Signed-off-by: Dennis Eisele kernlpanic@dennis-eisele.de


media-sound/flacon/flacon-8.2.0.ebuild
```diff
@@ -0,0 +1,83 @@
...
+
+src_configure() {
+	local mycmakeargs=(
+		-DBUILD_TESTS="$(usex test)"
+	)
+	cmake_src_configure
+}
+
+src_test() {
+	virtx "${BUILD_DIR}/tests/${PN}_test"
```
thesamesam, line 61: Careful!

```
# @FUNCTION: virtx
# @USAGE: <command> [command arguments]
# @DESCRIPTION:
# Start new Xvfb session and run commands in it.
#
# IMPORTANT: The command is run nonfatal !!!
#
# This means we are checking for the return code and raise an exception if it
# isn't 0. So you need to make sure that all commands return a proper
# code and not just die. All eclass function used should support nonfatal
# calls properly.
```

kernlpanic: To be honest.  The tests fail anyway on the new and the old version. Don't know why but sometimes they pass and sometimes they don't.

thesamesam: Still, we should fix the usage, but you can restrict the tests if they're inconsistent.

kernlpanic: Yes you're right just wanted to inform you. 
Thanks for the feedback btw.

PR #23769, opened 2022-01-13
sci-geosciences/grass: bump dev version to 8.1

tmszi:
Bump [GRASS GIS dev version to 8.1](https://github.com/OSGeo/grass/commit/029cd8ea8e9c98b575f57f148b1552c8821a0494) and fix metadata long description.

### DIFF
--- a/sci-geosciences/grass/grass-9999.ebuild
+++ b/sci-geosciences/grass/grass-9999.ebuild
@@ -9,15 +9,15 @@ WX_GTK_VER="3.0-gtk3"
 
 inherit autotools desktop git-r3 python-single-r1 toolchain-funcs wxwidgets xdg
 
-MY_P="${PN}8.1"
-MY_PM="${MY_P/.}"
-
 DESCRIPTION="A free GIS with raster and vector functionality, as well as 3D vizualization"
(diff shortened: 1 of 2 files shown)


sci-geosciences/grass/grass-9999.ebuild
```diff
@@ -236,7 +236,7 @@ os.environ\[\"GRASS_PYTHON\"\] = \"${EPYTHON}\":" \
 
 	# set proper GISDBASE directory path in the demolocation .grassrc80 file
 	sed -e "s:GISDBASE\:.*$:GISDBASE\: ${gisbase}:" \
-		-i "${ED}"${gisbase}/demolocation/.grassrc80 || die
+		-i "${ED}"${gisbase}/demolocation/.grassrc81 || die
```
thesamesam, line 239: Not something you _necessarily_ need to fix right now, but it'd be nice to use some manipulation of `${SLOT}` for this (e.g. `.${SLOT/./}`) instead, so we don't forget to update it.

Also, should probably update the comment above.

tmszi: Yes you are right, thanks.

PR #23770, opened 2022-01-13
media-video/ffmpeg: fix mixed strip usage

10ne1:
Even though stripping is disabled for compiled binaries
(--disable-stripping), some configure phase test binaries
get stripped and this causes Clang builds to use the GNU
strip instead of LLLVM strip.

Fix this by specifying to configure the proper version of
strip to be used via --strip="$(tc-getSTRIP)"

Bug: https://bugs.gentoo.org/831128

### DIFF
--- a/media-video/ffmpeg/ffmpeg-4.2.4-r1.ebuild
+++ b/media-video/ffmpeg/ffmpeg-4.2.4-r1.ebuild
@@ -463,6 +463,7 @@ multilib_src_configure() {
 		--cc="$(tc-getCC)" \
 		--cxx="$(tc-getCXX)" \
 		--ar="$(tc-getAR)" \
+		--strip="$(tc-getSTRIP)" \
 		--optflags="${CFLAGS}" \
 		$(use_enable static-libs static) \
 		"${myconf[@]}"
--- a/media-video/ffmpeg/ffmpeg-4.3.1-r1.ebuild
+++ b/media-video/ffmpeg/ffmpeg-4.3.1-r1.ebuild
@@ -464,6 +464,7 @@ multilib_src_configure() {
 		--cxx="$(tc-getCXX)" \
 		--ar="$(tc-getAR)" \
 		--nm="$(tc-getNM)" \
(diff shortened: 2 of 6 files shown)


media-video/ffmpeg/ffmpeg-4.2.4-r2.ebuild
```diff
@@ -463,6 +463,7 @@ multilib_src_configure() {
 		--cc="$(tc-getCC)" \
 		--cxx="$(tc-getCXX)" \
 		--ar="$(tc-getAR)" \
+		--strip="$(tc-getSTRIP)" \
```
thesamesam, line 466: Oh, this is probably something you've not needed to worry about on your end (maybe because you want a rebuild to be attempted or something), but we generally reserve revision bumps for:
- installed files / image changes
- RDEPEND changes
- [(some other bits)](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/)

so I wouldn't expect to need to bother here, because:
- we shouldn't be installing stripped binaries most of the time anyway (`FEATURES="nostrip"` is on by default)
- usually, the `strip` implementation shouldn't make a difference to the image, but I think it _can_ if you were e.g. doing cross and it picked the wrong one which lacked support for your architecture, but then.. Portage would pick the right one anyway

I've rambled there but I think the gist is that we don't need to revbump (and force a recompile) for people but if I've missed something, let me know.

10ne1: Thanks for the info. That is correct, yes, we don't need a rev-bump and it doesn't make sense to force a recompile for everyone just because we want to fix this minor GNU vs LLVM build inconsistency.

I will update the PR.

thesamesam: No problem!

I would do it if it was *FLAGS, or compiler choice. It's just that here it makes no difference to the end result, so it's a build-time failure, hence no need. So kind of a symptom of the weirdness of this one :)

PR #23777, opened 2022-01-13
openjdk:17 bootstrap using our snapshot

arthurzam:
@gyakovlev Here is my PR for review
Some notes:

1. please recheck all my changes to `profiles/` so it is done correctly. Also maybe we should update my comments (better phrasing)
2. All snapshots across arches will use same base directory for the snapshot, for now decided on `openjdk-bootstrap-${PV}`
3. Script to create tarball from xpak - [here](https://dev.gentoo.org/~arthurzam/prepare-openjdk-bootstrap.sh).
4. Use `I_KNOW_WHAT_I_AM_DOING_BOOTSTRAP` for our internal bundle everything.
5. After we clean the result here, I will copy the same code to `openjdk:11`, and then for `x86`.


dev-java/openjdk/openjdk-17.0.1_p12.ebuild
```diff
@@ -10,12 +10,18 @@ SLOT="$(ver_cut 1)"
 
 DESCRIPTION="Open source implementation of the Java programming language"
 HOMEPAGE="https://openjdk.java.net"
-SRC_URI="https://github.com/openjdk/jdk${SLOT}u/archive/refs/tags/jdk-${MY_PV}.tar.gz -> ${P}.tar.gz"
+SRC_URI="
+	https://github.com/openjdk/jdk${SLOT}u/archive/refs/tags/jdk-${MY_PV}.tar.gz
+		-> ${P}.tar.gz
+	!system-bootstrap? (
+		ppc64? ( big-endian? ( https://dev.gentoo.org/~arthurzam/distfiles/dev-java/${PN}/${PN}-bootstrap-17.0.1_p12-ppc64.tar.xz ) )
```
gyakovlev, line 17: let's make version a variable somewhere after inherit line.
maybe even 2 variables.
like
```bash
PPC64BE_BOOTSTRAP_TAR="17.0.1_p12"
X86_BOOTSTRAP_TAR="17.0.1_p12"
```
and reuse that in `SRC_URI` and in `pkg_setup`

profiles/arch/powerpc/ppc64/64ul/package.use.force
```diff
@@ -0,0 +1,7 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# Arthur Zamarin <arthurzam@gentoo.org> (2022-01-13)
+# Bootstrapping is possible using our build snapshot, but force
+# the user to rebuild  
+dev-java/openjdk:17 -system-bootstrap jbootstrap
```
gyakovlev, line 7: this will ALWAYS force user to to internal bootstrap.
we want users to use system version after initial one happened.
need to re-think all that a bit.

gyakovlev:
I was thinking, and looks like it'll be better if we do separate package after all.

`openjdk-bootstrap:17`, which simply untars tarball to `/opt/openjdk-boostrap/17`

that way we can avoid all those ugly masks, required use and hacks.
we'll still need ability to disable system libs, so that one stays.

It might be easier if I do that I'm talking about myself, I'm having hard time finding words to explain it without actually doing.

gyakovlev:
I did some work on your ebuild too

https://gist.github.com/gyakovlev/ffdc817b2cafaadc89aa28faf1a14df6

profiles/arch/powerpc/ppc64/64ul/package.use.force
arthurzam: Actually after rechecking on multiple arches, it looks like the logic for system-bootstrap works:
on amd64 & arm64, when running `equery u openjdk:17`, I don't see system-bootstrap
on ppc64be, when running `equery u openjdk:17`, I see system-bootstrap and I can enable or disable it.

arthurzam:
> I did some work on your ebuild too
> 
> https://gist.github.com/gyakovlev/ffdc817b2cafaadc89aa28faf1a14df6

I have imported those changes into PR - they are very good.

I'm not sure which variant goes better for my eyes. I prefer to not install our bootstrap version at all (someone somewhere will try to use it for java something outside it's purpose), but your variant is cleaner ebuild wise.

gyakovlev:
> > I did some work on your ebuild too
> > https://gist.github.com/gyakovlev/ffdc817b2cafaadc89aa28faf1a14df6
> 
> I have imported those changes into PR - they are very good.
> 
> I'm not sure which variant goes better for my eyes. I prefer to not install our bootstrap version at all (someone somewhere will try to use it for java something outside it's purpose), but your variant is cleaner ebuild wise.

I'm in the same boat.
-bootstrap pkg is cleaner.
but this version is nicer and more robust.

I have an a bit of crazy idea of not requiring openjdk at all on arches/profiles with bootstrap tarballs.
but it's very hard to specify such dependency without useflag and it will not scale if we introduce more bootstrap things.

I was thinking it could open up a way to bootstrap on musl, without maintaining musl specific binpkg.

dev-java/openjdk/openjdk-17.0.1_p12.ebuild
```diff
@@ -155,12 +188,12 @@ src_configure() {
...
-		--with-harfbuzz=system
-		--with-lcms=system
-		--with-libjpeg=system
-		--with-libpng=system
+		--with-freetype="${I_KNOW_WHAT_I_AM_DOING_BOOTSTRAP:-system}"
+		--with-giflib="${I_KNOW_WHAT_I_AM_DOING_BOOTSTRAP:-system}"
+		--with-harfbuzz="${I_KNOW_WHAT_I_AM_DOING_BOOTSTRAP:-system}"
+		--with-lcms="${I_KNOW_WHAT_I_AM_DOING_BOOTSTRAP:-system}"
+		--with-libjpeg="${I_KNOW_WHAT_I_AM_DOING_BOOTSTRAP:-system}"
+		--with-libpng="${I_KNOW_WHAT_I_AM_DOING_BOOTSTRAP:-system}"
```
gyakovlev, line 196: maybe change it so something shorter?
if's ok name as 1 time use, but this looks UUUGLY.

maybe `${XPAK_BOOTSTRAP:-system}` ?

profiles/arch/powerpc/ppc64/64ul/package.use.force
gyakovlev: yeah I might got confused. should be ok.

gyakovlev (approved):
ok this is looking good.
think on shorter variable name though.
I don't mind long one, but it just triggers me a bit =)

let's not merge it yet though, as it will get stable ppc64 keyword, which is not I'd like to do =)
in about 5 days java bump will happen, I have some other changes queued, I want to remove pch flag completely and do some cosmetics.

gyakovlev:
or if you want to get it merged (I don't mind) - revbump resetting to ~ppc64.
I completely forgot that we already got openjdk:17 stable.

or leave until x86 is done. up to you.

arthurzam:
> or if you want to get it merged (I don't mind) - revbump resetting to ~ppc64. I completely forgot that we already got openjdk:17 stable.
> 
> or leave until x86 is done. up to you.

This PR is a proof of concept and base for our next step. I think we will wait until next openjdk version bump, and then publish good "surprise" for ppc64be and x86.
I wanted to prepare everything so we aren't stuck with bump next week.

arthurzam:
I have added x86 bootstrap for 11 and 17.
In new java bump, I will do latest bootstrap version.

@gyakovlev are there any missing places where we should "unmask" or "keyword" for jdks to work? Like some virtuals?

profiles/arch/x86/package.use.force
```diff
@@ -1,6 +1,12 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
+# Arthur Zamarin <arthurzam@gentoo.org> (2022-01-16)
+# Bootstrapping is possible using our build snapshot, but force
+# the user to rebuild
+dev-java/openjdk:11 -system-bootstrap jbootstrap
+dev-java/openjdk:17 -system-bootstrap jbootstrap
+
```
gyakovlev, line 9: let's add
`REQUIRED_USE="!system-bootstrap? ( jbootstrap )"`  in the ebuild.

and maybe don't use.force `jbootstrap` here, it's already enabled with `+` in ebuild.

gyakovlev:
merged slightly rebased variant of this for new versions.

gyakovlev:
and yeah, huge thanks for working on this.

PR #23784, opened 2022-01-14
openjdk-bootstrap as separate package

gyakovlev:
alternative variant, just for testing


arthurzam (commented):
Looks good for me. Less noise in complicated openjdk ebuild, but we do get the our bootstrapped version installed

gyakovlev:
closing this for now, let's go other way after all.
in case this will be needed we'll revive it.

PR #23785, opened 2022-01-14
dev-java/qdox: MAVEN_ID | dev-java/xalan: MAVEN_ID | dev-java/rhino: MAVEN_ID

vaukai:


Flowdalic:
What is `MAVEN_ID`? Either my grep foo is to weak or there is no eclass that processes this variable. But if this variable is not processed, then what's the motivation for adding it?

vaukai:
> What is `MAVEN_ID`? Either my grep foo is to weak or there is no eclass that processes this variable. But if this variable is not processed, then what's the motivation for adding it?

[java-ebuilder](https://wiki.gentoo.org/wiki/Java-ebuilder) uses them to recognize dependencies.

fordfrog:
@vaukai merged, thanks!

PR #23790, opened 2022-01-14
app-misc/hivex: Do not depend hardly on ruby

nertpinx:
The ruby-ng eclass exports src_prepare function and it should be only called
conditionally just like all the other functions, so add a better src_prepare
function as otherwise the build can fail if no compatible ruby target is
selected in RUBY_TARGETS even if hivex is emerging with -ruby.

Signed-off-by: Martin Kletzander <nert.pinx@gmail.com>


juippis (approved):
Makes sense, thanks!

PR #23793, opened 2022-01-14
dev-java/c3p0: Update for EAPI 8

Leo3418:
Despite the fact that there's a test suite in this package's source tree, I found it unfeasible to run it from Portage because it requires an existing and running external SQL database. The 0.9.5.1 ebuild attempts to run it via the `junit-tests` Ant target, but no tests will actually be run because the target will not be executed unless the `junit.jar.file` property is set, which is not the case in that ebuild.

Anyway, I have already done some manual testing with its reverse dependencies in both `::gentoo` and `::spark-overlay`. More testing is welcome!


fordfrog:
@Leo3418 merged, thanks!

PR #23796, opened 2022-01-14
sys-devel/gdb: fix GCC/Clang build mixture

10ne1:
GDB configure will use a mix of GCC and Clang due to
$GCC_FOR_TARGET defaulting to GCC in Clang-configured
builds, so set the var to ensure the proper compilers
are detected and to avoid mixing them.

Before setting the variable (example from ChromiumOS):
checking for x86_64-cros-linux-gnu-gcc... x86_64-cros-linux-gnu-gcc

After:
checking for gcc... (cached) x86_64-cros-linux-gnu-clang

Bug: https://bugs.gentoo.org/831202

### DIFF
--- a/sys-devel/gdb/gdb-11.1.ebuild
+++ b/sys-devel/gdb/gdb-11.1.ebuild
@@ -193,6 +193,9 @@ src_configure() {
 	# source-highlight is detected with pkg-config: bug #716558
 	export ac_cv_path_pkg_config_prog_path="$(tc-getPKG_CONFIG)"
 
+	# ensure proper compiler is detected for Clang builds: bug #831202
+	export GCC_FOR_TARGET="${CC_FOR_TARGET:-$(tc-getCC)}"
+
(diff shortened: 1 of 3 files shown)


sys-devel/gdb/gdb-11.1.ebuild
```diff
@@ -193,6 +193,9 @@ src_configure() {
 	# source-highlight is detected with pkg-config: bug #716558
 	export ac_cv_path_pkg_config_prog_path="$(tc-getPKG_CONFIG)"
 
+	# ensure proper compiler is detected for Clang bulids: bug #831202
```
thesamesam, line 196:
```suggestion
	# ensure proper compiler is detected for Clang builds: bug #831202
```

10ne1: @thesamesam Thanks for catching this! I force pushed the fix so GH won't add an additional commit to keep the log clean.

PR #23799, opened 2022-01-14
media-video/wireplumber: bump to 0.4.7 r1 for Pro Audio routing fix

pinkflames:
I've tempered hopes for this to be the end of fixes for 0.4.7 but for now this is the best we have.


media-video/wireplumber/wireplumber-0.4.7-r1.ebuild
```diff
@@ -54,6 +54,10 @@ RDEPEND="${DEPEND}"
 
 DOCS=( {NEWS,README}.rst )
 
+PATCHES=(
+	"$FILESDIR"/${P}-default-nodes-handle-nodes-without-Routes.patch
```
thesamesam, line 58: ${FILESDIR}

PR #23802, opened 2022-01-15
sci-mathematics/why3: fix race condition in parallel make doc

fx-carton:
"make doc" uses sphinx to build both latex and html documentation. Both
sphinx rules in the makefile include "-d doc/.doctrees", ie. the same
path for "the cached environment and doctree files". In a parallel make
build, the rules are called simulateously, which means the cached files
could be read by one process while they are still being written by the
other.

I believe this is what causes bug #831168, with the following error:

    Extension error (sphinx.environment.collectors.toctree):
    Handler <bound method TocTreeCollector.get_updated_docs of [...]>
    for event 'env-get-updated' threw an exception
    (exception: pickle data was truncated)

@xgqt I'm not familiar with sphinx, but I think this what causes the reported bug. What do you think?

I've not revbumped, because it wouldn't change installed files/dependencies, only fix the build in case this bug happens; but I can revbump if it's needed.


fx-carton:
Allowing html and latex docs to be built in parallel seemed cleaner, but forcing -j1 would work as well. No strong opinion on this.

PR #23811, opened 2022-01-15
sci-geosciences/grass: version bump

hamarituc:
Bump to 8.0.0RC2 and removal of old files.


thesamesam:
Hm, there's also #23769, but they're compatible it looks like :)

PR #23325, opened 2021-12-16
more dependencies for dev-java/log4j-core-2.17.1

vaukai:
This PR now depends on https://github.com/gentoo/gentoo/pull/23438 which should be merged first.
https://github.com/gentoo/gentoo/pull/23438 contains `dev-java/kafka-clients` as a direct dependency of 'dev-java/log4j-core-2.17.1'.

There was a recent change wrt [java9](https://issues.apache.org/jira/browse/LOG4J2-3241) so that 'log4j-api-java9' and 'log4j-core-java9' get removed from the dependencies.

PR #23812, opened 2022-01-15
dev-java/cpptasks: Use EAPI 8, min Java 1.8, enable tests

Leo3418:
Update a dependency package needed by #23438 and #23325.


fordfrog:
@Leo3418 merged, thanks!

PR #23814, opened 2022-01-15
sci-libs/opencascade: fixes for building revdeps

waebbl:
- fix installation paths for 7.5
Use the same path as for the 7.6 series /usr/$(get_libdir)/${PN}
instead of /usr/$(get_libdir)/${P}, so the changed environment
files works for all versions.
Bugs: https://bugs.gentoo.org/831054
Bugs: https://bugs.gentoo.org/831069

- don't pass USE_OPENGL option (7.6.0)
The option defaults to on and was erroneously passed as with the
invert value of the gles2 USE flag, which lead to only libTKOpenGles
library being built, but no libTKOpenGL library, resulting in build
failures in revdeps.
Closes: https://bugs.gentoo.rog/831069

- rename gles2 USE flag to gles2-only (7.5)
Bugs: https://bugs.gentoo.org/831069
Bugs: https://bugs.gentoo.org/831126

### DIFF
--- a/sci-libs/opencascade/opencascade-7.5.2-r5.ebuild
+++ b/sci-libs/opencascade/opencascade-7.5.2-r5.ebuild
@@ -16,7 +16,7 @@ S="${WORKDIR}/occt-V${MY_PV}"
 LICENSE="|| ( Open-CASCADE-LGPL-2.1-Exception-1.0 LGPL-2.1 )"
 SLOT="0/${PV_MAJ}"
 KEYWORDS="~amd64 ~arm64 ~x86"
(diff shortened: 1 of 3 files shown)


thesamesam:
Great work, thank you for tackling this!

thesamesam:
(Typos in Bug tag btw ;))

waebbl:
> (Typos in Bug tag btw ;))

I was uncertain if it's Bug or Bugs. Sadly I chose the wrong ;)

thesamesam:
Both! ;)

But no worries. Thank you for the work!

PR #23823, opened 2022-01-16
net-ftp/ncftp: bump to EAPI 8 and fix bugs

ceamac:
Rename `configure.in` to `configure.ac` and disable `strip`

### DIFF
--- a/net-ftp/ncftp/ncftp-3.2.6-r4.ebuild
+++ b/net-ftp/ncftp/ncftp-3.2.6-r4.ebuild
@@ -1,0 +1,55 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+inherit autotools toolchain-funcs
+
+DESCRIPTION="An extremely configurable ftp client"
+HOMEPAGE="https://www.ncftp.com/"
+SRC_URI="
+	https://ftp.mirrorservice.org/sites/ftp.${PN}.com/${PN}/${P}-src.tar.xz
+"
+
+LICENSE="Clarified-Artistic"
+SLOT="0"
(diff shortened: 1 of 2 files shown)


net-ftp/ncftp/ncftp-3.2.6-r4.ebuild
```diff
@@ -0,0 +1,56 @@
...
+		econf \
+		$(use_enable pch precomp) \
+		--disable-ccdv \
+		--disable-universal
+}
+
+src_install() {
+	default
+
+	DOCS=( README.txt doc/*.txt )
```
Flowdalic, line 53: Sometimes  a glob in `DOCS` could cause issues,  see also https://github.com/gentoo/gentoo/pull/23668#issuecomment-1010143030. 

I recommend to use `dodoc` and `dohtml`, potentially with `-r`, instead.

ceamac: `DOCS` and `HTML_DOCS` with globbing cause problems if declared at global scope, but used inside a function they work fine.

But passing `-r` to `einstalldocs` does nothing, so I should change this anyway.

`dohtml` is banned in EAPI 7, and the suggestion [here](https://devmanual.gentoo.org/function-reference/install-functions/) is to switch to `einstalldocs`, that's why I tried this.  I also tried declaring `DOCS` and `HTML_DOCS` at global level without globbing, but I didn't like the result.

So I can either remove `-r` or simply revert to `docinto`/`dodoc` as in `ncftp-3.2.6-r3.ebuild`.  What do you think is better?

Flowdalic: I think

```
dodoc README.txt doc/*.txt
docinto html
dodoc doc/html/*.html
```

is just fine. One could maybe consider replacing the final `dodoc` with `dodoc -r doc/html`.

ceamac: Done.  Thank you

Flowdalic:
Thanks for your contribution.

PR #23826, opened 2022-01-16
dev-java/jsch: bump to 0.1.55

vaukai:


fordfrog:
@vaukai merged, thanks! i noticed you dropped the osgi stuff and examples from the ebuild but i see no problem in that.

PR #23828, opened 2022-01-16
dev-java/cpptasks: Fix Javadoc generation on JDK 11+

Leo3418:
On JDK 11+, dependencies should be added to the classpath for `javadoc` to ensure successful doc generation.  This is the root cause of the bug ticket fixed by this PR and might have something to do with the bugs tracked by https://bugs.gentoo.org/820869 too.


fordfrog:
@Leo3418 nice catch, thanks! could you please also add your findings to the referenced bug?

Leo3418:
> @Leo3418 nice catch, thanks! could you please also add your findings to the referenced bug?

Done: https://bugs.gentoo.org/831294#c5
Please let me know if you'd like more details in the bug!

fordfrog:
@Leo3418 thanks! :-)

PR #23838, opened 2022-01-17
distutils-r1 PEP517 followup fixes

mgorny:
CC @gentoo/python

### DIFF
--- a/eclass/distutils-r1.eclass
+++ b/eclass/distutils-r1.eclass
@@ -971,6 +971,12 @@ distutils-r1_python_compile() {
 			--no-compile-bytecode ||
 			die "installer failed"
 
+		# clean the build tree; otherwise we may end up with PyPy3
+		# extensions duplicated into CPython dists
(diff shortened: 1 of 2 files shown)


eclass/distutils-r1.eclass
```diff
@@ -164,6 +166,10 @@ _distutils_set_globals() {
 				bdep+='
 					dev-python/flit_core[${PYTHON_USEDEP}]'
 				;;
+			maturin)
+				bdep+='
+					dev-python/maturin[${PYTHON_USEDEP}]'
```
ionenwks, line 171:
```suggestion
					dev-util/maturin[${PYTHON_USEDEP}]'
```
Formerly it didn't even install any python stuff, so guess dev-util made sense at the time. I made it install the pep517 backend after picking it up.

mgorny: Thanks for catching my thinko.

thesamesam (approved):

PR #23840, opened 2022-01-17
dev-python/pipenv: retain tomlkit in v2022.1.8

oz123:
Bug: https://bugs.gentoo.org/831101
Bug: https://bugs.gentoo.org/717666


juippis (commented):
I may have forgotten what's the right way to handle this, but can't you just depend on *dev-python/tomlkit*? And since pipenv is not stable either, you don't have to wait for the deps to be stabilized either.

oz123:
@juippis, when I added dependencies that were not stable, there were QA errors.
In any case, I already opened a request to stabilize tomlkit.

PR #23841, opened 2022-01-17
dev-java/antlr-runtime: new package

vaukai:
Bug: https://bugs.gentoo.org/742554

This is only the first part.
Next step should be stringtemplate:4
either from https://github.com/antlr/stringtemplate4/blob/master/build.xml
or https://github.com/antlr/stringtemplate4/blob/master/src/BUILD.bazel


Leo3418:
Thanks! This will be helpful in fixing the bug. I've just created a new version of `dev-java/stringtemplate:4` that uses it, which should ultimately resolve the issue. Once this is merged, I'll open a pull request. https://github.com/Leo3418/gentoo/blob/bug-742554/dev-java/stringtemplate/stringtemplate-4.3.1.ebuild

vaukai:
> Thanks! This will be helpful in fixing the bug. I've just created a new version of `dev-java/stringtemplate:4` that uses it, which should ultimately resolve the issue. Once this is merged, I'll open a pull request. https://github.com/Leo3418/gentoo/blob/bug-742554/dev-java/stringtemplate/stringtemplate-4.3.1.ebuild

Thanks, but L14 is exactly what should be avoided: https://github.com/Leo3418/gentoo/blob/bug-742554/dev-java/stringtemplate/stringtemplate-4.3.1.ebuild#L14

Leo3418:
> > Thanks! This will be helpful in fixing the bug. I've just created a new version of `dev-java/stringtemplate:4` that uses it, which should ultimately resolve the issue. Once this is merged, I'll open a pull request. https://github.com/Leo3418/gentoo/blob/bug-742554/dev-java/stringtemplate/stringtemplate-4.3.1.ebuild
> 
> Thanks, but L14 is exactly what should be avoided: https://github.com/Leo3418/gentoo/blob/bug-742554/dev-java/stringtemplate/stringtemplate-4.3.1.ebuild#L14

I haven't found a way to build stringtemplate-4 without it yet. The build process requires class `org.antlr.Tool`, which is not provided by `antlr-runtime`. https://github.com/antlr/stringtemplate4/blob/4.3.1/build.xml#L81

Nevertheless, I was still able to patch `build.xml` so that pre-built JAR will not be put into the classpath during compilation.

Does `antlr-tool` mentioned by fordfrog at https://bugs.gentoo.org/742554#c9 provide `org.antlr.Tool`?

vaukai:
> Does `antlr-tool` mentioned by fordfrog at https://bugs.gentoo.org/742554#c9 provide `org.antlr.Tool`?

That's true. and it needs strringtemplate as a dependency.  So there is a circular dep.  My assumption was that "upstream" had some way to bootstrap stringtemplate without using a prebuilt antlr3-tool

Leo3418:
> That's true. and it needs strringtemplate as a dependency. So there is a circular dep. My assumption was that "upstream" had some way to bootstrap stringtemplate without using a prebuilt antlr3-tool

Unfortunately I can't find evidence suggesting that, even though the upstream supports 3 different build systems:

- Ant: https://github.com/antlr/stringtemplate4/blob/4.3.1/build.xml#L16 Downloads a pre-built JAR
- Maven: https://github.com/antlr/stringtemplate4/blob/4.3.1/pom.xml#L80-L96 Uses a Maven plugin, which we cannot incorporate with any Java eclass yet
- Bazel: https://github.com/antlr/stringtemplate4/blob/139f34243e516fc9d2cc4db8eeaa825014d631cc/src/BUILD.bazel#L35-L37 Says `@antlr3_bootstrap//jar`, looks like a pre-built

Leo3418:
Once this PR is merged, I'll submit another PR containing https://github.com/Leo3418/gentoo/commit/259b26dc50ae12578690c4f7b824a6e50cf8b5ed, which should finally resolve the bug.

PR #23822, opened 2022-01-16
dev-java/glassfish-xmlrpc-api: EAPI 8, restrict to java 1.8

vaukai:
Do not merge.  There might be a way to make it work with java-11

Bug: https://bugs.gentoo.org/831295
Package-Manager: Portage-3.0.28, Repoman-3.0.3
Signed-off-by: Volkmar W. Pogatzki <gentoo@pogatzki.net>

PR #23848, opened 2022-01-17
dev-java/jakarta-xml-soap-api: New package

Leo3418:
This package provides the `javax.xml.soap` Java package required by some Java ebuilds since Java 9, such as `dev-java/glassfish-xmlrpc-api`.  It is needed to resolve bugs like https://bugs.gentoo.org/831295 if we don't want to otherwise restrict those Java ebuilds to Java 1.8.

Related pull request: #23822


Leo3418:
Switched upstream from https://github.com/javaee/javax.xml.soap/tree/1.4.0 to https://github.com/eclipse-ee4j/saaj-api/tree/1.4.2/api as introduced by @fordfrog in https://bugs.gentoo.org/831295#c7 .  The latter source [uses module `jakarta-activation`](https://github.com/eclipse-ee4j/saaj-api/blob/1.4.2/api/src/main/java/module-info.java#L19), for which we already have `dev-java/jakarta-activation-api:1`.  The former source [uses `java-activation` instead](https://github.com/javaee/javax.xml.soap/blob/1.4.0/src/main/jdk9/module-info.java#L46) and would require [an extra patch](https://github.com/gentoo/gentoo/commit/55638f495d11e023f01db04a2afef3ae9b1bef2a#diff-0606d2c84c4a7c56e9569656cb9555e390e810f1535c40ca9df267ea3a67545f).

fordfrog:
@Leo3418 merged, thanks!

PR #23849, opened 2022-01-18
dev-java/ant-core: Fix Javadoc generation on JDK 11+; EAPI 8

Leo3418:
This is yet another instance of package that fails Javadoc build on JDK 11+, and the root cause is the same: packages needed by the source files are not present in the classpath passed into the `javadoc` program.  Unfortunately, for `dev-java/ant-core`, which inherits `java-ant-2.eclass`, a fix like a5cafff5294e38d8661b1c2468c7479f5aef4229 does not work.  After all, how are you supposed to call `ant` to build Ant when it is being bootstrapped upon the first installation?  Much of `java-ant-2.eclass` logic is skipped in this ebuild (which seemed to include classpath rewriting); the `./build.sh` script in Ant source tree is invoked to build Ant without Ant.

The `./build.sh javadocs` command used to build the Javadoc will build documentation for every Ant module (namely every `dev-java/ant-*` package), which means that all dependencies for all Ant modules are needed for `ant-core`'s `USE="doc"` even if `ant-core` itself does not depend on them.  And because the logic to rewrite classpath in `java-ant-2.eclass` is not applied, the dependencies need to be symlinked into the `lib/optional` directory, where Ant itself's `build.xml` will search for the dependencies.

### DIFF
--- a/dev-java/ant-core/ant-core-1.10.9-r1.ebuild
+++ b/dev-java/ant-core/ant-core-1.10.9-r1.ebuild
@@ -1,0 +1,161 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+# Don't depend on itself.
+JAVA_ANT_DISABLE_ANT_CORE_DEP="true"
+
+# Rewriting build.xml files for the testcases has no use at the moment.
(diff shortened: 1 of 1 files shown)


fordfrog:
@Leo3418 merged, thanks!

PR #23853, opened 2022-01-18
dev-python/eventlet: python 3.10 support

Miezhiko:
needed to drop 3.9 from my install

### DIFF
--- a/dev-python/eventlet/eventlet-0.33.0.ebuild
+++ b/dev-python/eventlet/eventlet-0.33.0.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=8
 
-PYTHON_COMPAT=( python3_{8..9} )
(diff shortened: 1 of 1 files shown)


mgorny:
Did you actually test it?

Miezhiko:
not fully, they've added support in last version https://github.com/eventlet/eventlet/pull/715

but "rudimentary", it doesn't affect my tasks but possibly warning about py10 being experimental here should be added

thesamesam:
> not fully, they've added support in last version [eventlet/eventlet#715](https://github.com/eventlet/eventlet/pull/715)
> 
> but "rudimentary", it doesn't affect my tasks but possibly warning about py10 being experimental here should be added

We don't want to add it until they properly support it and tests pass.

PR #23855, opened 2022-01-18
Bump kicad to 6.0.1

zpuskas:
Bump KiCad to 6.0.1, drop 6.0.0.

Additional fixes:
- update ebuilds to EAPI8
- fix some minimal ebuild dependencies from 5.99 to 6.0.0
- add "nls" USE flag to kicad which enables building and installing translations
- fix -DKICAD_DOCS to point to Gentoo's versioned documentation path, which will make documentation menu work in the GUI if documentation is installed
- make kicad-doc use ${P} again, so we can have a predictable path for kicad itself
- clean up left over deprecated variables


sci-electronics/kicad/kicad-6.0.1.ebuild
```diff
@@ -91,13 +93,13 @@ src_configure() {
 	xdg_environment_reset
 
 	local mycmakeargs=(
-		-DKICAD_DOCS="${EPREFIX}/usr/share/doc/${PF}"
+		-DKICAD_DOCS="${EPREFIX}/usr/share/doc/${PN}-doc-${PV}"
```
thesamesam, line 96: This doesn't match the other change which turns it back to ${P}?

zpuskas: The current -DKICAD_DOCS path is incorrect and the GUI will not work properly regardless of whether the local documentation is installed or not. -DKICAD_DOCS needs to point to the directory where the documentation files are installed, which is determined by the app-doc/kicad-doc ebuild. When this path is set correctly (as above) then when using the Help->Help menu in the KiCad application GUI, it will actually open the locally installed documentation. Thus this is the correct change to make, as this will resolve to `/usr/share/doc/kicad-doc-6.0.1`.

Note that this is a build time hardcoded path for KiCad, and it's safe to assume that sci-electronics/kicad will be merged before app-doc/kicad-doc (latter might not be merged at all). Any revisions done to app-doc/kicad would result in an invalid path in an already included in the KiCad binary, if kicad-doc were to continue using ${PF} as revisions for kicad and kicad-doc might be different, thus the revert to ${P} there. Also note it's not reasonable to ask the user to rebuild KiCad in case the documentation ebuild receives a revision, so this also goes against using ${PF} in kicad-doc.

If kicad-doc is not installed, KiCad won't find the docs in the predefined path and the binary will revert to current behaviour, where it will notify the user of missing local docs and will offer to go to the online documentation instead.

thesamesam: Got it. But I would put some of this in a comment.

thesamesam:
Thank you!

PR #23857, opened 2022-01-18
www-servers/lighttpd: Bump to version 1.4.63

herbmillerjr:
Migrate to meson build system.

Not sure how we want to handle html documentation. This isn't part of upstream's meson migration yet and it doesn't seem like a great idea to run the old configure just for docs. Removed block that builds html docs for time being.


herbmillerjr:
FYI - Due to https://bugs.gentoo.org/830691, we may want to delay merging this pull request.

gstrauss:
FYI: lighttpd 1.4.64 has been released, including the fix for CVE-2022-22707, which may affect rare configs.

lighttpd 1.4.64 defaults to using pcre2 (and pcre2 has been available as an option since lighttpd 1.4.62).  Please migrate gentoo to prefer pcre2 over pcre.  If pcre is required, then the build must explicitly specify `-Dwith_pcre=true`

lighttpd 1.4.64 removes long-deprecated modules.  The dependencies on gdbm and memcached have been removed.  It is also recommended that the dependency on bzip2 for mod_deflate be disabled by default.

@herbmillerjr wrote:
> Not sure how we want to handle html documentation. This isn't part of upstream's meson migration yet and it doesn't seem like a great idea to run the old configure just for docs.

patches welcome upstream.  What would you like to see?

Similarly, if you find your patch to meson.build for lua detection useful, please submit it upstream.  I am considering changing the search order to look for 'lua' before looking for specific versions.

gstrauss:
lighttpd meson build improvements (on lighttpd master branch; not yet released)
https://git.lighttpd.net/lighttpd/lighttpd1.4/commit/876a1e1b2cc21282a0df5245004197cdffe1d380
https://git.lighttpd.net/lighttpd/lighttpd1.4/commit/7792bc72305eda3a7bac52c4461fe11119595bb3

gstrauss:
@herbmillerjr this PR is back on your plate.  I think that lighttpd upstream (git master branch) incorporates most of the changes proposed here.

You correctly point out that lighttpd upstream does not include a `doc/meson.build`
If you would like to propose a `doc/meson.build` to lighttpd upstream, I will gladly review it.

herbmillerjr:
Going to close this for now since 1.4.64 was completed in another PR. I'll continue to tackle the meson migration and open a PR with the next version bump.

PR #23859, opened 2022-01-18
dev-java/glassfish-xmlrpc-api: Fix build error on Java 11+; EAPI 8

Leo3418:
Successor of #23822.  `dev-java/glassfish-xmlrpc-api` requires Java package `javax.xml.soap`, which is no longer in the JDK proper as of Java 11.  This PR declares the provider of `javax.xml.soap` as a dependency so it can be built on JDK 11.


fordfrog:
@Leo3418 merged, thanks! just please note that we cannot close the referenced bug until the failing ebuild is gone so i changed that in the commit message. for your new version of the ebuild, we should wait for at least two days and then we can file stabilization request i guess.

Leo3418:
> @Leo3418 merged, thanks! just please note that we cannot close the referenced bug until the failing ebuild is gone so i changed that in the commit message. for your new version of the ebuild, we should wait for at least two days and then we can file stabilization request i guess.

Sure, thanks for the correction and patience! I've still got things to learn...

PR #23860, opened 2022-01-18
dev-java/xml-commons-external: Fix Javadoc on JDK 11+ for 1.4.01-r2

Leo3418:
By passing `-source 8` to `javadoc` from JDK 11, there is no need to patch the original source files to eliminate uses of `_` as an identifier.  This will be good enough until `javadoc` no longer supports the `-source 8` option.


fordfrog:
@Leo3418 merged, thanks!

PR #23863, opened 2022-01-18
dev-libs/spdlog: force use of external fmt lib

stkw0:
Closes: https://bugs.gentoo.org/827889


ionenwks (commented):
Any reason this switch to cmake-multilib at same time?

Typically multilib is only added if something needs it. Furthermore, it'd need to depend on multilib libfmt if kept.

stkw0:
The idea was to support possible uses of this library in a multilib environment, although I don't know any package that needs it currently.

thesamesam:
> The idea was to support possible uses of this library in a multilib environment, although I don't know any package that needs it currently.

I'd leave it until something needs it.

thesamesam:
ping

PR #23864, opened 2022-01-18
media-sound/mumble: bump to 1.4.230

returntoreality:
Adds new version and new use flag for the new pipewire support.


thesamesam:
@Polynomial-C

Flowdalic:
I am not sure why, but `/usr/bin/mumble` does not link against `libopus.so`:

```
$ scanelf -n /usr/bin/mumble
 TYPE   NEEDED FILE 
ET_DYN libPocoZip.so.71,libsndfile.so.1,libQt5Sql.so.5,libQt5Svg.so.5,libQt5Widgets.so.5,libXi.so.6,librt.so.1,libspeex.so.1,libspeexdsp.so.1,libQt5DBus.so.5,libdns_sd.so.1,libasound.so.2,libPocoXML.so.71,libPocoFoundation.so.71,libpthread.so.0,libprotobuf.so.28,libssl.so.1.1,libcrypto.so.1.1,libQt5Network.so.5,libQt5Xml.so.5,libQt5Gui.so.5,libX11.so.6,libQt5Core.so.5,libstdc++.so.6,libm.so.6,libgcc_s.so.1,libc.so.6,ld-linux-x86-64.so.2 /usr/bin/mumble 
```

I would have expected `libopus.so` here. But maybe I am missing something?

returntoreality:
It seems it is loaded at runtime using QLibrary:
https://github.com/mumble-voip/mumble/blob/v1.4.230/src/mumble/OpusCodec.h#L17
and only the includes are used when using a non-bundled opus:
https://github.com/mumble-voip/mumble/blob/v1.4.230/src/mumble/CMakeLists.txt#L650

benaryorg:
I'm currently building with these flags:

```text
[ebuild     U  ] media-sound/mumble-1.4.230::gentoo [1.3.4::gentoo] USE="nls%* pulseaudio -alsa -dbus -debug -g15 -jack -pipewire% -portaudio -rnnoise -speech -test% -zeroconf (-opus%*) (-oss%) (-pch%)" 9221 KiB
```

And it fails with this here:

```text
-- Looking for sys/cdefs.h
-- Looking for sys/cdefs.h - not found
CMake Error at overlay_gl/CMakeLists.txt:39 (message):
  Can't find the 32bit version of sys/cdefs.h - did you install g++-multilib?


-- Configuring incomplete, errors occurred!
```

I'm running *nomultilib* on *amd64*, so I'm pretty sure it shouldn't *need* those 32bit versions.
Does this belong here or upstream?

(GitHub won't let me upload the *emerge-info.txt* so it lives [in ipfs now](https://pedosinspeedos.net/ipfs/QmWUzuspPSpu7K3tmh7K5eKoZG4jqmPwfZdAgusLnDVunU), if you're on ipv4-only try [this link](https://ipfs.io/ipfs/QmWUzuspPSpu7K3tmh7K5eKoZG4jqmPwfZdAgusLnDVunU) though it *might* load forever, same for the [build.log](https://pedosinspeedos.net/ipfs/QmTZ6XsmkB9XitRBsyvoYa8zg8MYjE19HrNevT2SHFAnPN) ([ipv4](https://ipfs.io/ipfs/QmTZ6XsmkB9XitRBsyvoYa8zg8MYjE19HrNevT2SHFAnPN)))

returntoreality:
Indeed the overlay is built for both 32bit and 64bit. To make it work without a patch the build of the overlay would need to be disabled on non-multilib systems. I could add a use flag for the overlay and mask it in the no-multilib profile. At least this is what they suggest upstream: https://github.com/mumble-voip/mumble/blob/v1.4.230/docs/dev/build-instructions/build_linux.md

returntoreality:
@benaryorg can you test if the new commit fixes your issue?

adeliktas:
can confirm this ebuild works, but in this release version i noticed a minor sound crackling in the audio assistant.

ppfeufer:
Add `BUILD_NUMBER` variable to cmake arguments to get the right version string

See: https://github.com/mumble-voip/mumble/issues/5538#issuecomment-1024972060

PR #23869, opened 2022-01-19
dev-java/gson: exclude 1 test if vm_version="1.8"

vaukai:
Closes: https://bugs.gentoo.org/820338


fordfrog:
@vaukai merged, thanks! i also noticed that this package fails some tests with `openjdk-17`. it seems this package is strongly bound to a specific version, or at least the tests. i did not have time to investigate it deeper though.

PR #23874, opened 2022-01-19
games-arcade/lbreakout: use https

mm1ke:
Simple HOMEPAGE update to use https instead of http.

### DIFF
--- a/games-arcade/lbreakout/lbreakout-010315-r1.ebuild
+++ b/games-arcade/lbreakout/lbreakout-010315-r1.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=8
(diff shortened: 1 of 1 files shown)


mm1ke:
Hmm, interesting, it doesn't work for me now too. 
However it did work when i made the PR. Maybe sourceforge has a problem?

ionenwks:
> Hmm, interesting, it doesn't work for me now too. However it did work when i made the PR. Maybe sourceforge has a problem?

I checked last time I looked at lbreakout's ebuild and others too (months ago), and always end up wondering why I bothered because these never work with sourceforge.net links.

I don't think it's a temporary issue, maybe accidentally checked the wrong link?

mm1ke:
> > Hmm, interesting, it doesn't work for me now too. However it did work when i made the PR. Maybe sourceforge has a problem?
> 
> I checked last time I looked at lbreakout's ebuild and others too (months ago), and always end up wondering why I bothered because these never work with sourceforge.net links.
> 
> I don't think it's a temporary issue, maybe accidentally checked the wrong link?

Sorry, my bad. Actually had a mistake in the new link. Should work now :)

ionenwks (approved):
> Sorry, my bad. Actually had a mistake in the new link. Should work now :)

Oh, good to know wrt that link, guess I'll (at least) update the other lgames links for consistency.

Thanks.

PR #23878, opened 2022-01-20
dev-cpp/tbb: fix undefined __TBB_machine_fetchadd4

alexfanqi:
didn't create a new revision because  the behavior of arches unaffected by the bug shouldn't be changed.

Closes: https://bugs.gentoo.org/827883
Signed-off-by: Alex Fan <alex.fan.q@gmail.com>


thesamesam:
Thank you!

PR #23880, opened 2022-01-20
net-voip/captagent: revision bump to fix clang compile

jkroonza:
Upstream: https://github.com/sipcapture/captagent/pull/245

Not expecting a new release soon.

Closes: https://bugs.gentoo.org/831391


thesamesam:
Note that not necessary per-se to revbump for compile fixes if truly compile-time-only but e.g. implicit function declarations can have runtime effects, so let's do it here!

PR #23882, opened 2022-01-20
app-emulation/libvirt: Rebase libvirt-6.7.0-do-not-use-sysconfig.patch

zippy2:
The live ebuild applies libvirt-8.1.0-do-not-use-sysconfig.patch
which does no longer apply cleanly because of libvirt's commit
8eb4461645. Rebase the patch onto current master.


zippy2:
CC @tamiko

tamiko:
@zippy2 Merci! Applied.

PR #23883, opened 2022-01-20
dev-util/cscope: Fix direct call of pkg-config in configure script

semihalf-czapiga-jakub:
Signed-off-by: Jakub Czapiga <jacz@semihalf.com>

### DIFF
--- a/dev-util/cscope/cscope-15.8a-r2.ebuild
+++ b/dev-util/cscope/cscope-15.8a-r2.ebuild
@@ -26,6 +26,7 @@ SITEFILE="50${PN}-gentoo.el"
 src_prepare() {
 	eapply "${FILESDIR}/${PN}-15.7a-ocs-sysdir.patch" #269305
 	eapply "${FILESDIR}/${PN}-15.6-darwin.patch"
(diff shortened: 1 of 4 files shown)


thesamesam:
Would you mind making a patch to `configure.ac` instead, especially given we already call `eautoreconf`?

But huge thanks for the change!

semihalf-czapiga-jakub:
@thesamesam Is the new version close, to what You anticipated?
Regarding pushing changes to upstream: I do not think, that cscope repository is even alive at this point. It's a great project, but their website is old, and mentions only version 15.8a (release around 2012).

semihalf-czapiga-jakub:
@thesamesam Can you take a look? Is it my fault that bot closed this PR?

thesamesam:
Nono, it's all OK, it's actually merged -- it's just because of how we merge commits (github is a mirror, so I apply patches with `git am`).

Your commit is safe & sound here: https://github.com/gentoo/gentoo/commit/19edcd377f662c55aae15bc8fc15ae0fe7e07c3a. Thank you!

semihalf-czapiga-jakub:
Oh. Thank you very much :)

PR #23884, opened 2022-01-20
app-crypt/gss-ntlmssp: new package for gss ntl support

Miezhiko:
gss-ntlmssp is important for communication with Windows services on windows, without this set up for krb5 user is getting error `Error: GSSAPI operation failed with error - An unsupported mechanism was requested. NTLM authentication requires the GSSAPI plugin 'gss-ntlmssp'`

It's ported to most of other distributions https://github.com/PowerShell/PowerShell-Docker/issues/124
(there is no ntlm use flag in app-crypt/mit-krb5)

I've tested myself that current ebuild solves problems and communication protocols are working with using it.


xgqt:
@gentoo/kerberos ?

PR #23886, opened 2022-01-20
sci-libs/vtk-8.2.0 - drop version

waebbl:
Only one package, sci-libs/oce, which is masked for last-riting, depends
on this version.
Won't build with ffmpeg-5.

Bug: https://bugs.gentoo.org/831595
Bug: https://bugs.gentoo.org/832625


a17r:
ffmpeg-5 is masked anyway. there is time left to investigate before we introduce upper-bound restrictions everywhere.

waebbl:
Upstream will likely backport a patch for 9.1, but probably not for 9.0.3 and earlier. See https://gitlab.kitware.com/vtk/vtk/-/issues/18445 for status.

a17r:
Will it be so complex that we can't backport to 9.0 at least?

waebbl:
Nope, it doesn't look that complex. I'll give it a try.

waebbl:
It can not be cleaned-up yet, since the sci-libs/oce package depends on v8*.

According to the upstream oce maintainer[1] we might think about last riting sci-libs/oce, depending on revdeps, as it's currently unmaintained and they suggest using opencascade instead.

[1] https://github.com/tpaviot/oce/issues/745#issuecomment-992285943

a17r:
@zpuskas ^

zpuskas:
> It can not be cleaned-up yet, since the sci-libs/oce package depends on v8*.

Historically opencascade was slow to release and was less community friendly, thus KiCAD relied on the OCE fork, but this has changed in the 6 series where they only support opencascade now. I've kept the 5 series of KiCAD in the tree as I know that some Gentoo users are using the program for work and I wanted to provide some time for them to migrate to the new version, since it had quite a few non backward compatible changes.

That being said as far as I can see the only user of sci-libs/oce at this point is the sci-electronics/kicad-5.1.12 ebuild. Since we have the KiCAD 6 series in the tree already, I think fist we need to start by dropping the last version of KiCAD 5 series and then we can last rite sci-libs/oce.

This is tracked in https://bugs.gentoo.org/832625.

waebbl:
> That being said as far as I can see the only user of sci-libs/oce at this point is the sci-electronics/kicad-5.1.12 ebuild. Since we have the KiCAD 6 series in the tree already, I think fist we need to start by dropping the last version of KiCAD 5 series and then we can last rite sci-libs/oce.
> 
> This is tracked in https://bugs.gentoo.org/832625.

Thanks for your effort. Sounds like a good way to go to me.
I will then drop 8.2.0, once sci-libs/oce has been last rited.

a17r:
This PR can now be turned into vtk-8.2.0 cleanup, IINM.

a17r:
ping

PR #23903, opened 2022-01-21
net-im/telegram-desktop-bin: bump version to 3.4.8

henning-schild:


ionenwks (approved):
Got QA notice due to its .desktop file changes, however had a look at it and it's due to desktop-file-utils not knowing about xdg spec 1.5 yet (including its addition of SingleMainWindow key)

So I'll add a QA_DESKTOP_FILE to skip that noise, should be removable in the future.

Was:
```
 * QA Notice: This package installs one or more .desktop files that do not
 * pass validation.
 * 
 * 	/usr/share/applications/telegramdesktop.desktop: error: file contains key "SingleMainWindow" in group "Desktop Entry", but keys extending the format should start with "X-"
 * 	/usr/share/applications/telegramdesktop.desktop: error: value "1.5" for key "Version" in group "Desktop Entry" is not a known version
 * 
```

PR #23908, opened 2022-01-22
sys-fs/cryptsetup: support for using detached LUKS headers

shaohme:
I thought it would be a nice feature to support detached LUKS headers when running OpenRC.

### DIFF
--- a/sys-fs/cryptsetup/cryptsetup-2.3.6-r3.ebuild
+++ b/sys-fs/cryptsetup/cryptsetup-2.3.6-r3.ebuild
@@ -1,0 +1,127 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit autotools linux-info tmpfiles
+
+DESCRIPTION="Tool to setup encrypted devices with dm-crypt"
+HOMEPAGE="https://gitlab.com/cryptsetup/cryptsetup/blob/master/README.md"
+SRC_URI="https://www.kernel.org/pub/linux/utils/${PN}/v$(ver_cut 1-2)/${P/_/-}.tar.xz"
+
+LICENSE="GPL-2+"
+SLOT="0/12" # libcryptsetup.so version
+[[ ${PV} != *_rc* ]] && \
+KEYWORDS="~alpha amd64 arm arm64 hppa ~ia64 ~mips ppc ppc64 ~riscv ~s390 sparc x86"
+CRYPTO_BACKENDS="gcrypt kernel nettle +openssl"
+# we don't support nss since it doesn't allow cryptsetup to be built statically
+# and it's missing ripemd160 support so it can't provide full backward compatibility
+IUSE="${CRYPTO_BACKENDS} +argon2 nls pwquality reencrypt static static-libs +udev urandom"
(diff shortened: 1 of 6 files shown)


sys-fs/cryptsetup/files/1.6.7-dmcrypt.confd
```diff
@@ -44,6 +44,7 @@ dmcrypt_retries=5
 #                                    for blkid (see -t option).  This is safer than using
 #                                    the full path to the device.
 # key='</path/to/keyfile>[:<mode>]'  == Fullpath from / or from inside removable media.
+# use_header='</path/to/header>'     == Full path to detached LUKS header file instead of .
```
floppym, line 47: I think this should just be called `header` to match the cryptsetup option name and the setting from crypttab.

floppym (changes requested):

sys-fs/cryptsetup/files/1.6.7-dmcrypt.confd
shaohme: I thought about that too, but named it such avoiding collide with the existing ${header} variable in the script.
If we are sure it won't collide I'll happily rename it.

floppym: Ah, I did not realize that we were already using that variable name.

I would probably rename the existing `header` variable since it is used internally and not exposed to users.

It looks like the existing `header` variable is used to control some `einfo` and `ebegin` statements. Maybe change the variable name to `print_header`.

shaohme: renamed 'header' to 'print_header'

floppym (approved):

floppym (changes requested):
A few things before I can merge this:
- Do not modify 1.6.7-dmcrypt.confd and 1.6.7-dmcrypt.rc. These are old versions that should not receive feature enhancements.
- Do not modify 2.4.0-dmcrypt.confd and 2.4.0-dmcrypt.rc. Instead, copy them to a new version and apply your changes.
- Revbump the latest cryptsetup ebuild and update the src_install function to install the new confd and rc files.
- Squash all changes into 1 commit.
- Ensure your commit message compiles with [GLEP 66](https://www.gentoo.org/glep/glep-0066.html#commit-messages).

shaohme:
> A few things before I can merge this:
> 
>     * Do not modify 1.6.7-dmcrypt.confd and 1.6.7-dmcrypt.rc. These are old versions that should not receive feature enhancements.
> 
>     * Do not modify 2.4.0-dmcrypt.confd and 2.4.0-dmcrypt.rc. Instead, copy them to a new version and apply your changes.
> 
>     * Revbump the latest cryptsetup ebuild and update the src_install function to install the new confd and rc files.
> 
>     * Squash all changes into 1 commit.
> 
>     * Ensure your commit message compiles with [GLEP 66](https://www.gentoo.org/glep/glep-0066.html#commit-messages).

Should be done now.

floppym:
We will stabilize 2.4.3 in [bug 831982](https://bugs.gentoo.org/831982).

We can stabilize 2.4.3-r1 once the init script changes have been in ~arch for a bit. If there are any regressions, hopefully people will report them.

PR #23909, opened 2022-01-22
dev-util/astyle: keyword -3.1-r2 for ~riscv

JerSjjj:


tamiko:
@JerSjjj Would you mind opening a keywording bug on bugs.gentoo.org? We will have to ask the risc arch team to do the actual keywording.

PR #23911, opened 2022-01-22
distutils-r1 pep517: don't install licenses and prefix fixes

mgorny:
CC @gentoo/python @ionenwks

### DIFF
--- a/eclass/distutils-r1.eclass
+++ b/eclass/distutils-r1.eclass
@@ -973,8 +973,8 @@ distutils-r1_python_compile() {
 
 		# TODO: workaround for a bug in installer; remove once we depend
 		# on a properly fixed version
-		if [[ -d ${root}/usr/bin ]]; then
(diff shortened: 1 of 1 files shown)


SoapGentoo (approved):

arthurzam (approved):

eclass/distutils-r1.eclass
```diff
@@ -973,22 +973,27 @@ distutils-r1_python_compile() {
 
 		# TODO: workaround for a bug in installer; remove once we depend
 		# on a properly fixed version
```
thesamesam, line 975: While here, please add a bug reference to this so we know when to remove it in future.

thesamesam (approved):

eclass/distutils-r1.eclass
mgorny: Done.

ionenwks (approved):

PR #23913, opened 2022-01-22
eclass/vala.eclass: Support EAPI 8

parona-source:
Support EAPI 8 by dropping unnecessary inherit of eutils.eclass, the `in_iuse` function used from eutils.eclass is included in supported EAPIs already. See table [12.21](https://projects.gentoo.org/pms/8/pms.html#x1-132003r21) from PMS for EAPI 8.


parona-source:
Closing due to changes being merged which make this redundant.

PR #23915, opened 2022-01-22
dev-python/jupyter_server: enable doc generation

arthurzam:


dev-python/black/black-21.12_beta0-r1.ebuild
```diff
@@ -0,0 +1,67 @@
+# Copyright 2020-2022 Gentoo Authors
```
mgorny, line 1: Plz do PEP517 here while you're revbumping it.

dev-python/sphinxemoji/sphinxemoji-0.2.0.ebuild
```diff
@@ -0,0 +1,18 @@
+# Copyright 2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+PYTHON_COMPAT=( python3_{8..10} )
```
mgorny, line 6: Any reason not to use PEP517 here?

mgorny (changes requested):
Plz PEP517 all the new packages.

arthurzam:
Thanks, done and merged!

PR #23919, opened 2022-01-22
dev-java/commons-digester: Drop 1.8.1-r2

vaukai:


fordfrog:
i guess `games-puzzle/pauker` isn't compatible with the newer version of `dev-java/lucene`, right? as if it is, i'd rather see the old version of `dev-java/lucene` gone than the newer one.

vaukai:
> i guess `games-puzzle/pauker` isn't compatible with the newer version of `dev-java/lucene`, right? as if it is, i'd rather see the old version of `dev-java/lucene` gone than the newer one.

The old version of `dev-java/lucene` doesn't bother us. It's `lucene:3.6` pulling in the oldest slot of `commons-digester`.  But maybe we could borrow a much newer version from `spark overlay`?

PR #23920, opened 2022-01-22
media-libs/opencv: Drop 4.5.0-r3 and 4.5.1-r3

vaukai:
Versions depending on >=virtual/jdk-1.6 will never compile with jdk-17


Amynka:
Looks good.. Thanks !

PR #23921, opened 2022-01-22
sys-cluster/openmpi: Drop versions depending on >=virtual/jre-1.6

vaukai:


Arfrever:
If these versions fail with newer Java, then you might just drop `java` USE flag in them.

scardracs:
Not a good idea to drop openmpi so lightly, see https://bugs.gentoo.org/828032#c9

vaukai:
> Not a good idea to drop openmpi so lightly, see https://bugs.gentoo.org/828032#c9

I see.  Could you then switch them to `java 1.8:*` please?

scardracs:
> > Not a good idea to drop openmpi so lightly, see https://bugs.gentoo.org/828032#c9
> 
> I see.  Could you then switch them to `java 1.8:*` please?

I can't actually, I need cluster team to look at it

PR #23923, opened 2022-01-23
app-shells/starship: Add 1.2.1 and drop 0.58.0

bowlofeggs:


ionenwks:
Thanks.

I removed the openssl dep given all the openssl crates are gone and it wasn't linking with it anymore.

PR #23927, opened 2022-01-23
python-utils-r1.eclass: Make epytest disable plugins we don't want + distutils-r1 installer bump

mgorny:
CC @gentoo/python


arthurzam (approved):

eclass/python-utils-r1.eclass
```diff
@@ -1321,6 +1321,14 @@ epytest() {
 		-Wdefault
 		# override color output
 		"--color=${color}"
+		# disable the undesirable-dependency plugins by default to
+		# trigger missing argument strips.  strip options that require
```
sbraz, line 1325: Why 2 spaces here and on the next line?

eclass/python-utils-r1.eclass
```diff
@@ -1321,6 +1321,14 @@ epytest() {
 		-Wdefault
 		# override color output
 		"--color=${color}"
+		# disable the undesirable-dependency plugins by default to
```
sbraz, line 1324: The dash in that word seems weird to me.

eclass/python-utils-r1.eclass
thesamesam: After the full stop? It's a an (older) style thing. I've never done it but some people learned it that way.

sbraz: Ah OK, thanks, it seems weird to me (just like [this](https://www.python.org/dev/peps/pep-0008/#inline-comments)).

mgorny: It makes spotting sentence boundaries easier.

SoapGentoo: the yanks call is "french spacing" (eventhough IIRC France never did this). I also don't like it, but w/e

SoapGentoo (approved):

arthurzam (approved):

thesamesam (approved):

thesamesam:
Merged.

PR #23928, opened 2022-01-23
profiles/base: mask sys-libs/db[java]

thesamesam:
Broken with OpenJDK 11+ and we're not aware of anybody actually
needing these anyway.

Bug: https://bugs.gentoo.org/713866


thesamesam:
merged

PR #23933, opened 2022-01-23
net-p2p/freenet: Fix build and runtime errors with JNA 5.x

Leo3418:
`dev-java/jna` has been updated to 5.10.0, which requires updates to this package:

- This package uses JNA's `Pointer.SIZE` API, which has been removed since JNA 5.0.0.  A replacement of the API is `Native.POINTER_SIZE`, which is present in both JNA 4.x and 5.x. (https://bugs.gentoo.org/830847, https://github.com/kaitoy/pcap4j/issues/191)

- Since JNA 5.0.0, the default JNI library loading mechanism has been changed, which would cause this package to crash upon launch, unless system property `jna.nosys` is set to `false`, which restores the 4.x library loading behavior compatible with this package. (https://github.com/java-native-access/jna/issues/384)

### DIFF
--- a/net-p2p/freenet/files/0.7.5_p1491-update-for-jna-5.x.patch
+++ b/net-p2p/freenet/files/0.7.5_p1491-update-for-jna-5.x.patch
@@ -1,0 +1,31 @@
+From 830b4bb3948bd69fbd9b10001940f1775051114a Mon Sep 17 00:00:00 2001
+From: Yuan Liao <liaoyuan@gmail.com>
+Date: Sun, 23 Jan 2022 10:09:17 -0800
+Subject: [PATCH] Replace Pointer.SIZE with Native.POINTER_SIZE for JNA 5.x
+
+This patch is backward compatible with JNA 4.x because
+Native.POINTER_SIZE is present in both 4.x and 5.x.
+
+Bug: https://github.com/kaitoy/pcap4j/issues/191
+Bug: https://bugs.gentoo.org/830847
+Signed-off-by: Yuan Liao <liaoyuan@gmail.com>
+---
+ src/freenet/io/comm/UdpSocketHandler.java | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/src/freenet/io/comm/UdpSocketHandler.java b/src/freenet/io/comm/UdpSocketHandler.java
+index 0cd975f16..c004613a9 100644
+--- a/src/freenet/io/comm/UdpSocketHandler.java
++++ b/src/freenet/io/comm/UdpSocketHandler.java
+@@ -120,7 +120,7 @@ public class UdpSocketHandler implements PrioRunnable, PacketSocketHandler, Port
+ 			    return false;
+ 			int ret = -1;
+ 			try {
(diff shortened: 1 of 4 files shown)


net-p2p/freenet/files/0.7.5_p1491-update-for-jna-5.x.patch
```diff
@@ -0,0 +1,31 @@
+From 830b4bb3948bd69fbd9b10001940f1775051114a Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: Does this need upstreaming to https://github.com/freenet/fred?

Leo3418: The upstream is still using JNA 4.5.2 in the Gradle project. If they don't want to support using different versions of dependencies, then this patch is not quite useful to them. It is upstream-able though

thesamesam: Ah, I see. I would still say it's worth sending to them though for future proofing.

net-p2p/freenet/freenet-0.7.5_p1488-r2.ebuild
```diff
@@ -0,0 +1,169 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+JAVA_PKG_IUSE="doc source"
+
+inherit epatch java-pkg-2 java-ant-2 systemd
```
vaukai, line 8: DeprecatedEclass: version 0.7.5_p1488-r2: uses deprecated eclass: epatch (migrate to eapply from EAPI 7)

net-p2p/freenet/freenet-0.7.5_p1491-r1.ebuild
```diff
@@ -0,0 +1,165 @@
...
+inherit java-pkg-2 java-ant-2 systemd
+
+DESCRIPTION="An encrypted network without censorship"
+HOMEPAGE="https://freenetproject.org/"
+#	https://github.com/${PN}/seedrefs/archive/build0${PV#*p}.zip -> seednodes-${PV}.zip
+SRC_URI="
+	https://github.com/${PN}/fred/archive/build0${PV#*p}.zip -> ${P}.zip
+	https://github.com/${PN}/seedrefs/archive/build01480.zip -> seednodes-0.7.5_p1480.zip
+	mirror://gentoo/freenet-ant-1.7.1.jar"
+
```
vaukai, line 17:
```
DeadUrl: version 0.7.5_p1491-r1: SRC_URI: 403 Client Error: Forbidden for url: http://gentoo-distfiles.mirrors.tds.net/distfiles/freenet-ant-1.7.1.jar
TarballAvailable: version 0.7.5_p1491-r1: zip archives used when tarball available: [ https://github.com/freenet/fred/archive/build01491.zip https://github.com/freenet/seedrefs/archive/build01480.zip ]
```

net-p2p/freenet/freenet-0.7.5_p1488-r2.ebuild
Leo3418: Let's see whether this package's maintainer is interested in fixing these QA issues :)

PR #23934, opened 2022-01-23
app-i18n/transifex-client: skip tests depending on network connection.

robert7k:
Closes: https://bugs.gentoo.org/831772


app-i18n/transifex-client/transifex-client-1.0.0.ebuild
```diff
@@ -219,6 +219,8 @@ SLOT="0"
 KEYWORDS="~amd64 ~x86"
 
 src_test() {
+	# Skip tests depending on a network connection. Bug #831772
+	rm internal/txlib/update_test.go
```
thesamesam, line 223: || die on external commands which may fail

PR #23936, opened 2022-01-23
net-mail/grepmail: version bump to 5.3111

rndxelement:
I added an ebuild for grepmail-5.3111 with support for EAPI 8.
The new ebuild doesn't need patches, it uses a source from CTAN and the
tests are working now.  

Sources for why the patches aren't needed anymore:
- [5.30.33-fix_nonexistent_mailbox_test.patch](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=549782)  
- [5.30.33-midnight.patch](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=559588)  

The sed call in src_prepare() isn't needed anymore because of: [https://github.com/coppit/grepmail/blob/master/CHANGES](https://github.com/coppit/grepmail/blob/master/CHANGES) in the section "Version 5.3105"  

Closes: https://bugs.gentoo.org/494350
Signed-off-by: Philipp Roesner <rndxelement@protonmail.com>


net-mail/grepmail/grepmail-5.3111.ebuild
```diff
@@ -0,0 +1,36 @@
...
+DEPEND="${RDEPEND}
+	test? (
+		dev-perl/Test-Compile
+		dev-perl/Test-Pod
+		dev-perl/Test-Pod-Coverage
+	)
+"
+
+src_prepare() {
+	perl-module_src_prepare
```
thesamesam, line 35: We should be able to drop this as the Perl eclass should export the phase.

thesamesam:
Good work & thanks for doing the analysis of the changes. It's really nice to see!

rndxelement:
> Good work & thanks for doing the analysis of the changes. It's really nice to see!

Thanks!

PR #23937, opened 2022-01-24
dev-lang/python: fix hardcoding C/LDFLAGS from build time

mgorny:
So apparently we've been building CPython wrong and patching distutils to fix our mistake. Since py3.11 / setuptools vendored distutils no longer carry our patch, let's fix building Python so that CFLAGS and LDFLAGS from build time aren't embedded in sysconfig.

TODO: check pypy3


mgorny:
PyPy3 doesn't suffer from this issue.

floppym:
Interesting. That's a nice improvement over the patch.

arthurzam (approved):

thesamesam (approved):
Great!

PR #23939, opened 2022-01-24
x11-misc/qps: loosen version constraint on lxqt-base/liblxqt

AdelKS:
So one can use lxqt-base/liblxqt-9999

Closes: https://bugs.gentoo.org/831886

Signed-off-by: Adel KARA SLIMANE <adel.ks@zegrapher.com>

### DIFF
--- a/x11-misc/qps/qps-2.4.0-r1.ebuild
+++ b/x11-misc/qps/qps-2.4.0-r1.ebuild
@@ -18,13 +18,12 @@ BDEPEND="
 	>=dev-util/lxqt-build-tools-0.10.0
 "
 DEPEND="
-	=lxqt-base/liblxqt-1*:=
 	>=dev-qt/qtcore-5.15:5
 	>=dev-qt/qtdbus-5.15:5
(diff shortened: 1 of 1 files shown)


Chiitoo (approved):
Usually (always?) for new revisions, we create a new file with a commit, and then in a later commit remove the old version (instead of renaming).

Or at least that's how I've been doing it.  I couldn't find this documented anywhere at this time though...

Also, this change might not even require a new revision since users would not have been able to install the conflicting ebuilds, but I do wonder if it's possible to run into issues if one already has lxqt-1.* and qps-2.4.* installed, and then goes for the 9999s... which I believe should trigger a rebuild for qps and as such, I think it would be okay with this change in?

Will need to ask around for a confirmation.

Other than that, looks good to me.

Thanks!

thesamesam:
I'm fine with `git mv` where things were definitely wrong before, but otherwise, yeah, `cp`.

While we can rely on slot rebuilds here I think, it's not a big deal to revbump here and it means we're not having to do any mental gymnastics to be sure we're right.

Cheers both!

PR #23942, opened 2022-01-24
Update sys-apps/usbredir to 0.12.0

inode64:
Signed-off-by: INODE64 <ffelix@inode64.com>


juippis:
Thanks! I wanted to fix some style "issues" so it so was faster to do the bump by myself.

PR #23943, opened 2022-01-24
app-pda/pilot-link: Fix autoconf issue with popt.

gryf:
Nested calling autoconf macros for detecting popt is a probable reason
for weirdness in generated configure script. I've followed similar
approach as with png patch.

Closes: https://bugs.gentoo.org/831985


jonasstein:
Do you mean `Closes:` instead of `Bug:` ?

gryf:
Done.

PR #23946, opened 2022-01-24
dev-python/argon2-cffi-bindings: Added dev-python/setuptools_scm to BDEPEND

seifertm:
dev-python/setuptools_scm is a build-time dependency for argon2-cffi-bindings-21.2.0 [1] and it's missing.

[1] https://github.com/hynek/argon2-cffi-bindings/blob/21.2.0/pyproject.toml#L2


mgorny (changes requested):
Don't revbump for build-time fixes.

seifertm:
Got it, thanks! I removed the -r1 ebuild and added the dependency to the existing ebuild.

mgorny (approved):
Thanks, I'm adding it to my commit queue.

PR #23950, opened 2022-01-25
media-libs/nv-codec-headers: add new version 11.1.5.1

Tatsh:
Add new version by copying 11.1.5.0 ebuild to 11.1.5.1

New version has a somewhat significant change: [Add a subset of CUDA EGLStreams interop API](https://git.videolan.org/?p=ffmpeg/nv-codec-headers.git;a=commit;h=50d5add953977a2050a7b7e569b0ff4df1a1ce3e) which is used by [nvidia-vaapi-driver](https://github.com/elFarto/nvidia-vaapi-driver)


ionenwks:
@Tatsh see you've added nvidia-vaapi-driver to your overlay, adding this to ::gentoo been on my TODO since seen it on nvidia's forums but been waiting for it to mature before testing it. Not that I'd be against adding it now if you wanted to proxy it.

Tatsh:
> @Tatsh see you've added nvidia-vaapi-driver to your overlay, adding this to ::gentoo been on my TODO since seen it on nvidia's forums but been waiting for it to mature before testing it. Not that I'd be against adding it now if you wanted to proxy it.

If you use Firefox and are willing to set up the environment variables and configs, then it is working minimally. But for Chrome even with latest updates it's not ready yet. It actually makes Chrome assume the GPU is broken and it reverts everything to software.

I also am interested in seeing this work for Kodi since the `vdpau` USE pulls in the ancient and not working (for me) `x11-libs/libva-vdpau-driver`.

Tatsh:
If you are going to add nvidia-vaapi-driver to ::gentoo, you should also make a virtual/ package for that so file, and allow Kodi to use nvidia-vaapi-driver instead of libva-vdpau-driver. Similar thing could be done with `USE=vaapi VIDEO_CARDS=nvidia` for mpv.

Tatsh:
I tested Kodi with nvidia-vaapi-driver and it works much better than the old one ever did. It's nice to have some maintained code there. Too bad it still does bad things to Chrome. That can be worked around with an environment variable `LIBVA_DRIVER_NAME=intel` or similar (even if the Intel file is not present), which I don't mind for now.

PR #23955, opened 2022-01-25
x11-terms/sakura: remove unused patch

mm1ke:


ionenwks:
Oh right, remembered deleting the opacity patch but forgot I rebased the gentoo.patch and left the old one there.

Thanks.

PR #23956, opened 2022-01-25
x11-misc/qt5ct: add patch to allow xdg desktop portal dialog

fedeliallalinea:
Closes: https://bugs.gentoo.org/832065


mgorny:
Please ping me if the patch is merged upstream or they don't reply within 2 weeks or so.

fedeliallalinea:
> Please ping me if the patch is merged upstream or they don't reply within 2 weeks or so.

Patch [applied upstream](https://sourceforge.net/p/qt5ct/tickets/84/), but a new release is not yet available.

thesamesam:
> > Please ping me if the patch is merged upstream or they don't reply within 2 weeks or so.
> 
> Patch [applied upstream](https://sourceforge.net/p/qt5ct/tickets/84/), but a new release is not yet available.

@mgorny

PR #23959, opened 2022-01-25
sys-devel/flang: new subpackage from LLVM-13

dzejej:
This is llvm based fortran language compiler.
Important: To compile this you need sys-devel/mlir package
which I earlier PR
The patches comes from Fedora 35


sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -0,0 +1,100 @@
...
+		use doc &&  mycmakeargs+=(
+			-DLLVM_ENABLE_SPHINX:BOOL=ON
+			-DLLVM_ENABLE_DOXYGEN=ON
+			-DFLANG_INCLUDE_DOCS=ON
+			-DSPHINX_WARNINGS_AS_ERRORS=OFF
+			-DSPHINX_EXECUTABLE=/usr/bin/sphinx-build
+		)
+		use test && mycmakeargs+=(
+				-DLLVM_BUILD_TESTS=ON
+				-DLLVM_EXTERNAL_LIT=/usr/bin//lit
```
thesamesam, line 90:
```suggestion
				-DLLVM_EXTERNAL_LIT="${BROOT}"/usr/bin/lit
```

sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -0,0 +1,100 @@
...
+}
+
+PATCHES=(
+	"${FILESDIR}"/flang-Link-against-libclang-cpp.so.patch
+	"${FILESDIR}"/flang-PATCH-flang-Disable-use-of-sphinx_markdown_tables.patch
+)
+
+pkg_setup() {
+	use system-llvm && llvm_pkg_setup
+	use system-llvm &&  export CC=clang
```
thesamesam, line 55: Don't align:
```suggestion
	use system-llvm && export CC=clang
```

but this isn't ideal anyway, it doesn't respect the user's choice for these variables. Also, if we _were_ going to  do this, we should do:
`if use system-llvm ; then
	export ...
fi`

I think you should be able to just get rid of this, right? We always want to use the system LLVM copy.

Or must Flang be built with Clang specifically or something (which will require some machinery but not quite like this)?

thesamesam:
On second thought, please do cherry-pick the MLIR commit into this branch and then force push so they're in the same PR.

sys-devel/flang/flang-13.0.0.ebuild
dzejej: The problem is as follows. 
flang could be compiled either by g++ or by clang+.  Because gentoo is about choices, I would like to leave the possibility of choosing g++/clang+ - maybe some other USE flag could be chosen ? 
Is there any USE flag which determines CC/CXX compiler for particular ebuild despite of system wide CC/CXX compiler ?
And one more think, it is enough to set up CC and CXX for flang. 

Personally, on my computer I stick with GNU toolchain (particularly CXX=g++), but when I choose clang+ for flang   all warning disappear.

dzejej:
The new ebulds for mlir and flang is updated and PR

PR #23960, opened 2022-01-25
media-plugins/vdr-streamdev: version bump to 0.6.3

madmartin:
new EAPI 8
make USE=server the default
adopt proxy maintainership


media-plugins/vdr-streamdev/vdr-streamdev-0.6.3.ebuild
```diff
@@ -0,0 +1,78 @@
...
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~x86"
+IUSE="client +server"
+REQUIRED_USE="|| ( client server )"
+
+BDEPEND="acct-user/vdr"
+DEPEND=">=media-video/vdr-2.3"
+RDEPEND="${DEPEND}"
+
+QA_FLAGS_IGNORED="
```
thesamesam, line 23: Is this stuff prebuilt or not? Why do we need to set QA_FLAGS_IGNORED?

madmartin: All packages media-plugins/vdr-* are build with the CFLAGS taken from the BDEPENDed media-video/vdr. This may result into a QA message that this package is ignoring the current CFLAGS.
Copying the CFLAGS from main vdr package is necessary to asssure that the plugins (libraries) match exactly.

media-plugins/vdr-streamdev/vdr-streamdev-0.6.3.ebuild
```diff
@@ -0,0 +1,78 @@
...
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~x86"
+IUSE="client +server"
+REQUIRED_USE="|| ( client server )"
+
+BDEPEND="acct-user/vdr"
+DEPEND=">=media-video/vdr-2.3"
+RDEPEND="${DEPEND}"
```
juippis, line 21: I would put the acct-* deps in DEPEND that also gets inherited by RDEPEND. Both of them.

juippis (commented):
```
 * QA Notice: Plugin isn't converted to gettext handling!
 * QA Notice: Fixing include of libsi-headers
```
is this a problem? Looks good otherwise, but see the comment about acct* deps.

madmartin:
@juippis now changed to 

DEPEND="acct-user/vdr
    >=media-video/vdr-2.3"
BDEPEND="${DEPEND}"
RDEPEND="${DEPEND}"

juippis (approved):

PR #23967, opened 2022-01-26
sys-devel: mlir new subbpackage from LLVM-13

dzejej:
flang is llvm based fortran language compiler.
mlir is llvm subpackage needed for flang


sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -0,0 +1,111 @@
...
+"
+
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+RESTRICT="!test? ( test )"
+
+LLVM_MAX_SLOT=13
+
+pkg_setup() {
+	use system-llvm && llvm_pkg_setup
+	if use clang ; then
```
stefson, line 49: for use clang, you need to pull in sys-devel/lld. Otherwise this happens: 
```
CMake Error at /usr/lib/llvm/13/lib64/cmake/llvm/HandleLLVMOptions.cmake:298 (message):
  Host compiler does not support '-fuse-ld=lld'
Call Stack (most recent call first):
```

stefson:
for sys-devel/flang: the package, as is now, needs an excessive amount of 4gb of ram per thread. Can you try to reduce that, please? llvm turns off assertions to do that.

stefson:
it seems as if sys-devel/flang can't find a few libs. They are present though in work/flang-13.0.0_build-abi_x86_64.amd64/lib/

```
[189/237] cd /var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0_build-abi_x86_64.amd64/tools/f18 && /usr/bin/cmake -E make_directory /var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0_build-abi_x86_64.amd64/include/flang && /var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0_build-abi_x86_64.amd64/bin/f18 -fsyntax-only -module-dir /var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0_build-abi_x86_64.amd64/include/flang /var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0/module/__fortran_type_info.f90
FAILED: include/flang/__fortran_type_info.mod /var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0_build-abi_x86_64.amd64/include/flang/__fortran_type_info.mod 
cd /var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0_build-abi_x86_64.amd64/tools/f18 && /usr/bin/cmake -E make_directory /var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0_build-abi_x86_64.amd64/include/flang && /var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0_build-abi_x86_64.amd64/bin/f18 -fsyntax-only -module-dir /var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0_build-abi_x86_64.amd64/include/flang /var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0/module/__fortran_type_info.f90
/var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0_build-abi_x86_64.amd64/bin/f18: error while loading shared libraries: libFortranLower.so.13: cannot open shared object file: No such file or directory
```

full build log: [build.log.zip](https://github.com/gentoo/gentoo/files/7943660/build.log.zip)

dzejej:
In ebuld file there is 
export LD_LIBRARY_PATH="${S}_build-abi_${ABI}.${ARCH}/lib:${LD_LIBRARY_PATH}"
- for me it works.  Just add this line after export:
echo ${LD_LIBRARY_PATH) 
and watch if your path to build have something like
/var/tmp/portage/sys-devel/flang-13.0.0/work/flang-13.0.0_build-abi_x86_64.amd64/lib

sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -0,0 +1,111 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+CMAKE_ECLASS=cmake
+
+PYTHON_COMPAT=( python3_{8..10} )
+inherit cmake-multilib llvm python-single-r1
```
thesamesam, line 9: Please address my other comments on the old PRs, e.g llvm.org.eclass and fixing SRC_URI.

sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -0,0 +1,111 @@
...
+pkg_setup() {
+	use system-llvm && llvm_pkg_setup
+	if use clang ; then
+		export CC=clang
+		export CXX=clang++
+		export AR=llvm-ar
+		export LD=lld
+		export RANLIB=llvm-ranlib
+		export NM=llvm-nm
+	fi
```
thesamesam, line 56: Like I said on the other PR, we let users do stuff like this via package.env, not USE flags.

we only do e.g USE=clang if hacks are needed to make things build and we avoid it if possible.

sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -0,0 +1,111 @@
...
+	eapply -p2 "${DISTDIR}/flang-PATCH-flang-Disable-use-of-sphinx_markdown_tables.patch"
+	eapply_user
+	cmake_src_prepare
+}
+
+multilib_src_configure() {
+	local mycmakeargs=(
+		-GNinja
+		-DCMAKE_INSTALL_PREFIX="${EPREFIX}/usr/lib/llvm/${SLOT}"
+		-DMLIR_TABLEGEN_EXE="${EPREFIX}/usr/lib/llvm/${SLOT}/bin/mlir-tblgen"
```
thesamesam, line 76: BROOT instrad of EPREFIX as it's in BDEPEND only?

sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -0,0 +1,111 @@
...
+		-DCMAKE_BUILD_TYPE=Release
+		-DCLANG_DIR="${EPREFIX}/usr/lib/llvm/${SLOT}/$(get_libdir)/cmake/clang/"
+		-DMLIR_DIR="${EPREFIX}/usr/lib/llvm/${SLOT}/$(get_libdir)/cmake/mlir/"
```
thesamesam, line 79: If it's really BDEPEND only then BROOT

sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -0,0 +1,111 @@
...
+		-DBUILD_SHARED_LIBS:BOOL=ON
```
thesamesam, line 80: See comment from other PR.

sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -0,0 +1,111 @@
...
+		-DLLVM_LINK_LLVM_DYLIB:BOOL=ON
+		-DLLVM_ENABLE_THREADS=ON
+
+		-DCMAKE_PREFIX_PATH="${EPREFIX}/usr/lib/llvm/${SLOT}/$(get_libdir)/cmake/llvm/"
+		-Wno-dev
+		)
+		use clang &&  mycmakeargs+=(
+			-DLLVM_ENABLE_LLD=ON
+		)
+		use doc &&  mycmakeargs+=(
```
thesamesam, line 90: Please fix the remaining comments from other PR

thesamesam (changes requested):
Thanks! You didn't need to make a new PR though. You could've cherry-picked the commit and force pushed to the old branch. It makes easier review then to ensure comments were addressed and so on.

sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -0,0 +1,111 @@
...
+		use test && mycmakeargs+=(
+				-DLLVM_BUILD_TESTS=ON
+				-DLLVM_EXTERNAL_LIT="${BROOT}"/usr/bin//lit
+				-DPython3_EXECUTABLE="${PYTHON}"
+		)
+	cmake_src_configure
+}
+
+multilib_src_compile() {
+	export LD_LIBRARY_PATH="${S}_build-abi_${ABI}.${ARCH}/lib:${LD_LIBRARY_PATH}"
```
dzejej, line 106: The ${ABI} returns amd64, but I need x86_64.
export LD_LIBRARY_PATH="${S}-${PV}_build-abi_x86_${ABI_X86}.${ARCH}/lib
 - is this correct ? (${ABI} -> x86_${ABI_X86} )

thesamesam: No, because we might not be running on x86 :)

We can maybe look in the eclass to figure out how it generates that string.

But I now realise what you might want is ${BUILD_DIR} instead which solves our problems.

mgorny:
How about starting with 14.0.0.9999, and make sure all the necessary fixes are found on the main branch?

dzejej:
I am afraid that the problem is rather in ebuild files.

sys-devel/flang/flang-13.0.0.ebuild
dzejej: I'm afraid that I couldn't quite understand what the problem is. Sorry. 
It is about CC/CXX compiler - it is not recommended to change system wide compiler
or it is about the clang flag - it should be used in different way.

thesamesam: If users want to use Clang, they should be using `package.env` e.g. https://wiki.gentoo.org/wiki/Clang#Clang_environments.

We only provide `USE=clang` if certain hacks are needed to make it work (like Firefox).

So you should be able to just remove all of this.

dzejej:
Sorry - I thinking about LD_LIBRARY_PATH problem.
I do not know if the patches are mandatory, if flang could be compiled without the patches or even flang compiler works correctly (with or without patches). First, I want to make flang avialable to system.

thesamesam:
> Sorry - I thinking about LD_LIBRARY_PATH problem. I do not know if the patches are mandatory, if flang could be compiled without the patches or even flang compiler works correctly (with or without patches). First, I want to make flang avialable to system.

I believe the LD_LIBRARY_PATH issue can be solved by using ${BUILD_DIR} instead.

sys-devel/flang/flang-13.0.0.ebuild
dzejej: Thank you very much. Now I get it what was about. 
Now I understand that my question was quite stupid :)

thesamesam: No such thing as a stupid question and you know I'm happy to help you learn! 👍

sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -0,0 +1,118 @@
...
+	cmake_src_compile
+	if use doc ; then
+		cmake_src_compile doxygen-flang
+	fi
+}
+
+pkg_postinst() {
+	elog "You should consider switching to the clamg/clamg++ "
+	elog " for flang if you didn't that. See: "
+	elog " https://wiki.gentoo.org/wiki/Clang#Clang_environments "
```
dzejej, line 117: if !tc-is-clang; then
                elog "You should consider switching to the clamg/clamg++ "
                elog "for building flang. See:"
                elog "   https://wiki.gentoo.org/wiki/Clang#Clang_environments"
        fi
This is better :)

sys-devel/flang/flang-13.0.0.ebuild
```diff
@@ -6,19 +6,19 @@ EAPI=7
 CMAKE_ECLASS=cmake
 
 PYTHON_COMPAT=( python3_{8..10} )
-inherit cmake-multilib llvm python-single-r1
+inherit cmake-multilib python-single-r1 llvm llvm.org
 
 DESCRIPTION="Flang is a ground-up implementation of a Fortran front end written in modern C++"
 HOMEPAGE=" https://github.com/llvm/llvm-project/tree/main/flang
 		http://flang.llvm.org/docs/"
-SRC_URI="https://github.com/llvm/llvm-project/releases/download/llvmorg-${PV}/llvm-project-${PV}.src.tar.xz
```
stefson: hold on, you just removed the src_uri for the llvmorg tarball :)

dzejej:
as mgorny saddest the transition to 14.0.0.9999, I have the situation:
[     U *] sys-libs/libomp-14.0.0.9999
[  NS   *] sys-devel/llvm-14.0.0.9999
 - it is save tu upgrade libomp (I have one system which I would like not to break up). The same situation is with sys-devel/llvmgold and sys-devel/lld. In summary - is it possible to install these packages alongside stable ones ?

dzejej:
update to 14.0.0.9999 :)

band-a-prend:
@stefson 
> for sys-devel/flang: the package, as is now, needs an excessive amount of 4gb of ram per thread. Can you try to reduce that, please? llvm turns off assertions to do that.

I early tried to prelare flang 13.0 ebuild to and encountered the same problem so I have to test ebuild with MAKEOOTS="-j1". Finally it failed after several hours due to absence of clang-run and smth. else. The build time was too long for me.

@dzejej ,
do you stop attempts to prepare flang ebuild? The termux project also provides this package now.

PR #23968, opened 2022-01-26
app-shells/starship: add "notify-rust" USE flag

pasha132:
Hello!
According to [discussion](https://github.com/starship/starship/discussions/2731) "notify-rust" support isn't enabled by default, even if crate is installed.
This can be checked with such command:
```bash
STARSHIP_LOG=debug starship module cmd_duration -d 60000 2>&1 | tail -n2
[DEBUG] - (starship::modules::cmd_duration): This version of starship was built without notification support.
```
And after adding USE-flag:
```bash
STARSHIP_LOG=debug starship module cmd_duration -d 60000 2>&1 | tail -n2
[DEBUG] - (starship::config): Config found for "cmd_duration": Some(Table({"show_notifications": Boolean(true)}))
```

### DIFF
--- a/app-shells/starship/starship-1.0.0-r1.ebuild
+++ b/app-shells/starship/starship-1.0.0-r1.ebuild
@@ -1,0 +1,293 @@
+# Copyright 2020-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+CRATES="
+	ahash-0.4.7
+	aho-corasick-0.7.18
+	ansi_term-0.11.0
+	ansi_term-0.12.1
+	arrayref-0.3.6
+	arrayvec-0.5.2
+	async-io-1.6.0
+	attohttpc-0.18.0
+	atty-0.2.14
+	autocfg-1.0.1
+	base64-0.13.0
+	bitflags-1.3.2
+	blake2b_simd-0.5.11
+	block-0.1.6
+	block-buffer-0.7.3
+	block-buffer-0.9.0
+	block-padding-0.1.5
+	byte-tools-0.3.1
+	byte-unit-4.0.13
+	byteorder-1.4.3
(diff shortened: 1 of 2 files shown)


ionenwks:
Not overly familiar with starship, but until the maintainer has a look.. but is there any reason to have a USE to control it? Or can it be enabled unconditionally?

Generally do the latter when it doesn't add new dependencies (and a crate hardly count, I thought it may need libnotify at first but looks standalone), unless behavior changes that users may want to be able to opt-out (asking given wondering why upstream is making it a opt-in).

I'd say no reason to change 1.0.0 too, 1.2.0 will eventually replace it.

pasha132:
It seem that `notify-rust` dependency is [optional](https://github.com/starship/starship/blob/f47fdc59120850c9a461a70696850a83c10150fd/Cargo.toml#L40) for now. And `notify-rust` [depends](https://github.com/hoodie/notify-rust/blob/main/README.md) only from dbus (but I'am not sure).

I made it as a new USE just for not breaking user experience: for now it is just don't work event if crate installed.

I can remove 1.0.0 for now? And for 1.2.0 just wait for feedback from maintainer?

bowlofeggs:
Hi @pasha132 and @ionenwks !

It seems that notify-rust is not an optional dependency in starship 1.3.0, which was recently released. This commit seems to have moved it to a required dependency:

https://github.com/starship/starship/commit/dcf66659d0e87e619a1a881b2745eb68dbcd41a1

I'm planning to bump to 1.3.0 soon, so should we close this PR in anticipation of that effort?

bowlofeggs:
I opened https://github.com/gentoo/gentoo/pull/24141 to update to 1.3.0, which should have the feature that this PR was adding. I suggest we close this PR in favor of the other one.

ionenwks:
Makes sense, I'll close it then. Thanks.

PR #23974, opened 2022-01-26
dev-java/jsr305: Drop 3.0.2 | dev-java/typesafe-config: Drop 1.2.1 and 1.3.0-r2 | dev-java/gson: Drop 2.8.7

vaukai:


fordfrog:
ah, sorry, i first did the cleanup and then went here.

PR #23975, opened 2022-01-26
net-vpn/strongswan: version bump to 5.9.5

rndxelement:
Added an ebuild for strongswan-5.9.5.  
The [new version](https://github.com/strongswan/strongswan/releases) doesn't seem to require changes in the ebuild, so the new ebuild is the same as the current ebuild for strongswan-5.9.4.

Signed-off-by: Philipp Roesner <rndxelement@protonmail.com>


net-vpn/strongswan/strongswan-5.9.5.ebuild
```diff
@@ -0,0 +1,311 @@
...
+		ewarn "You should re-emerge ${PN} with at least the 'caps' USE flag enabled."
+		ewarn
+	fi
+	if use non-root; then
+		elog
+		elog "${PN} has been installed without superuser privileges (USE=non-root)."
+		elog "This imposes several limitations mainly to the IKEv1 daemon 'pluto'"
+		elog "but also a few to the IKEv2 daemon 'charon'."
+		elog
+		elog "Please carefully read: http://wiki.strongswan.org/wiki/nonRoot"
```
Flowdalic, line 285: This is now different from the existing ebuilds now that 0dccf910fa8936f6d02499f785f105fe0ec3c3fa is merged. I suggest to sync the message.

Flowdalic (changes requested):

net-vpn/strongswan/strongswan-5.9.5.ebuild
rndxelement: Thanks, I edited my commit with the modified elogs.

PR #23989, opened 2022-01-27
PCSX2 wont build without enabling DSDL2_API first

jusefn:
`CMake Error at cmake/ApiValidation.cmake:113 (message):
  wxWidgets is linked to SDL2.  Please use -DSDL2_API=TRUE Call Stack (most recent call first):
  cmake/SearchForStuff.cmake:188 (WX_vs_SDL)
  CMakeLists.txt:27 (include)`

this change in the cmake flags seems to fix it.


ionenwks:
Ah, happens only with wxGTK[sdl], didn't have it set so never ran into this.

That aside, the option is kind of silly, it doesn't do anything but silence the error (aka no build changes I can see). Seems to be something upstream forgot to cleanup, but guess no harm in adding for now.

ionenwks:
Thanks.

PR #23991, opened 2022-01-28
www-apps/gitea: security bump to 1.15.10

hydrapolic:
Tested on ~amd64, thanks!


hydrapolic:
Thank you @thesamesam

PR #24000, opened 2022-01-28
media-libs/openjpeg: Fix mandir and docdir for generated pkgconfig

laumann:
Update the patch to remove the ${prefix} variable from mandir and docdir
as it results in incorrect paths.

Bug: https://bugs.gentoo.org/694776


laumann:
@thesamesam Can I ask for your feedback on this? For some reason, this PR wasn't labeled with maintainer-needed.

laumann:
Changes:
 - Created a revbump (not sure if that needs explicit mention in the commit message)
 - Dropped all `arch` to `~arch` at repoman's request

I've made a copy of the patch for the revbump, made the changes I wanted to make in the new patch, copied the ebuild, modified the PATCHES array (and set the copyright year to 2022).

PR #24001, opened 2022-01-28
net-dns/{bind,bind-tools} 9.18.0 bump

erayaslan:
see https://bugs.gentoo.org/832218


mikhailkoliada:
@erayaslan could you provide a diff?

erayaslan:
[bind-9.18.0.diff.txt](https://github.com/gentoo/gentoo/files/7963406/bind-9.18.0.diff.txt)

[bind-tools-9.18.0.diff.txt](https://github.com/gentoo/gentoo/files/7963407/bind-tools-9.18.0.diff.txt)

diffs attached

mikhailkoliada:
@erayaslan Thanks!

I would not like to drop chroot as this is widely-used, rather add it with a systemd unit too

mikhailkoliada:
@erayaslan I would also probably drop static-libs entirely, I can not think of any use cases for this in bind

net-dns/bind/bind-9.18.0.ebuild
```diff
@@ -0,0 +1,151 @@
...
+	dnstap? ( dev-libs/fstrm dev-libs/protobuf-c )
+	doh? ( net-libs/nghttp2 )
+	geoip? ( dev-libs/libmaxminddb )
+	gssapi? ( virtual/krb5 )
+	idn? ( net-dns/libidn2 )
+	lmdb? ( dev-db/lmdb )
+	xml? ( dev-libs/libxml2 )
+"
+BDEPEND="
+	test-extra? (
```
mikhailkoliada, line 42: I'd leave it under the test USE flag, to generalize the installations and not to make additional checks what to enable/disable

mikhailkoliada (changes requested):

net-dns/bind/bind-9.18.0.ebuild
```diff
@@ -0,0 +1,151 @@
...
+	selinux? ( sec-policy/selinux-bind )
+	sys-process/psmisc
+	!net-dns/bind-tools
+"
+
+S="${WORKDIR}/${MY_P}"
+
+src_configure() {
+	local myeconfargs=(
+		--prefix="${EPREFIX}"/usr
```
mikhailkoliada, line 59: Eprefix is needed for /etc and /var dirs too I believe

mikhailkoliada (changes requested):

net-dns/bind-tools/bind-tools-9.18.0.ebuild
```diff
@@ -0,0 +1,117 @@
...
+	for tool in cds dsfromkey importkey keyfromlabel keygen \
+		revoke settime signzone verify; do
+		doman ${man_dir}/dnssec-"${tool}".1
+	done
+
+	emake DESTDIR="${D}" -C bin/tools/ install
+	doman ${man_dir}/{arpaname,mdig,named-journalprint,named-rrchecker,nsec3hash}.1
+
+	# just leave the tools to be installed
+	rm -rf "${D}"/usr/include/
```
mikhailkoliada, line 114: `|| die`

mikhailkoliada (changes requested):

erayaslan:
> I would not like to drop chroot as this is widely-used, rather add it with a systemd unit too

sure, your call. but I'm not going to do it as I do not think it is worth the code complexity and maintenance burden.

hopefully, the above notes will help with the bump.

closing

net-dns/bind-tools/bind-tools-9.18.0.ebuild
erayaslan: with -f I do not think it will ever fail

net-dns/bind/bind-9.18.0.ebuild
erayaslan: the reason was you can not run

`# USE=test ebuild bind-9.18.0.ebuild test`

as that results in circular dependencies with pytest.  so this was a dirty hack but yep maybe not the best idea

PR #23904, opened 2022-01-21
dev-java/{antlr*:3.5,stringtemplate:4}: Break circular dependency

Leo3418:
The commits in this PR implement the solution envisioned by @fordfrog at https://bugs.gentoo.org/742554#c9 for breaking the `stringtemplate-4` and `antlr-3.5` circular dependency, which should eventually resolve the issue.

Thanks @vaukai for the `dev-java/antlr-runtime` ebuild!

PR #24006, opened 2022-01-29
dev-java/antlr: Version bump to 4.9.3

Leo3418:
The commits in this pull request aim to update ANTLR to 4.9.3 to serve the following purposes:
- To update the package from a release that is nearly 6 years old,
- To aid documentation generation of `sci-mathematics/coq`, which requires ANTLR 4.7 or higher,
- To split `antlr-runtime` and `antlr-tool` just like in https://github.com/gentoo/gentoo/pull/23904, so this package can get rid of pre-built binaries entirely during the build process


fordfrog:
`dev-java/antr-runtime`, `dev-java/json-api` and `dev-java/jol-core` merged.

fordfrog:
for `dev-java/antlr-tool`, you should not call `java` directly but imo should instead retrieve it using `java-config -J`. it did not work for me, i got test failures. my system vm is `openjdk-17`. i did not have time to dig into this deeper but no matter what i set to `JAVA_PKG_FORCE_VM`, it always complained that source 6 is not supported, so it obviously always used `openjdk-17`.

Leo3418:
> for `dev-java/antlr-tool`, you should not call `java` directly but imo should instead retrieve it using `java-config -J`. it did not work for me, i got test failures. my system vm is `openjdk-17`. i did not have time to dig into this deeper but no matter what i set to `JAVA_PKG_FORCE_VM`, it always complained that source 6 is not supported, so it obviously always used `openjdk-17`.

Fixed, although directly calling `java` should not be the root cause of the test failure on Java 17.  It's because one of the test sources would invoke `javac` with hard-coded `-source 1.6 -target 1.6` in the arguments.  It has been [patched](https://github.com/gentoo/gentoo/pull/24006/files#diff-cf2a8cac9ba8022f2de2fb9bd07809e34f9b790d62264d31d435c8f3e3ab1318) and tests should now pass on every Java version we currently have.

fordfrog:
@Leo3418 merged, thanks!

PR #23702, opened 2022-01-08
app-admin/graylog: bump to 4.1.11 / drop to m-n

hydrapolic:
Thanks

PR #24010, opened 2022-01-29
acct-{user,group}/graylog: drop to m-n

hydrapolic:
Forgot to do this in https://github.com/gentoo/gentoo/pull/23702, thanks


hydrapolic:
Thank you @ionenwks

PR #24012, opened 2022-01-29
dev-libs/criterion: bump to 2.4.0_rc

hydrapolic:


dev-libs/criterion/criterion-2.4.0_rc.ebuild
```diff
@@ -0,0 +1,65 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI="8"
+
+PYTHON_COMPAT=( python3_{7..10} )
```
thesamesam, line 6: On bumps, please drop 3.7 as it's a noop (eclass ignores it).

hydrapolic: Will do, thanks @thesamesam

PR #24017, opened 2022-01-29
net-news/snownews: bump to 1.9

capezotte:


msharov:
The CI report is referring to the unused `autotools` inherit. snownews `configure` script is not generated by autotools.

There are a few other problems as well, such as the perl libs in `RDEPEND`. v1.9 does not use perl for anything.

Next, `src_prepare` calls `pkg_config` to find ncurses and manually substitutes `-lncursesw` with the output. `configure` already calls `pkg_config` and does the necessary substitutions, so this are not needed. There is also the substitution to remove the `-s` flag on `$(INSTALL)` call. That is also no longer correct. 

Installation calls `${INSTALL_PROGRAM}`, which is set to `install -m 755` by default. The executables are not stripped during installation; they are built stripped using the `-s` flag added to `ldflags` on `Config.mk.in` line 34. If you want symbols in the non-debug build, replace `-s` with `-g` and replace `-g0` on the line above with `-g -rdynamic`.

`src_prepare` should now be empty and could be removed.

`src_configure` incorrectly calls `./configure --prefix="${EPREFIX}/usr"`. It should call `./configure --prefix=/usr`. Installation prefix should instead be passed as `make DESTDIR="${EPREFIX}" install`. The comment above configure call also incorrectly states that configure is written in perl.

`src_compile` does not pass the custom flags correctly. The passed variables should be `CFLAGS` and `LDFLAGS`, not `EXTRA_CFLAGS/LDFLAGS`:
```sh
make CC="$(tc-getCC)" CFLAGS="${CFLAGS}" LDFLAGS="${LDFLAGS}"
```

msharov:
So what's the reason for rejection now?

thesamesam:
They closed the PR by themselves..I don't know why?

PR #24019, opened 2022-01-30
net-analyzer/termshark: new ebuild

hamarituc:
@thesamesam requested to move Volatility from my personal gentoo overlay into `::gentoo`. See https://github.com/hamarituc/gentoo-overlay/commit/1586c288c467e45d553564b96ac02ca3861c99fc#r64788747


net-analyzer/termshark/termshark-2.3.0.ebuild
```diff
@@ -0,0 +1,98 @@
...
+	"gopkg.in/ini.v1 v1.62.0 github.com/go-ini/ini"
+	"gopkg.in/tomb.v1 dd632973f1e7 github.com/go-tomb/tomb"
+	"gopkg.in/yaml.v2 v2.4.0 github.com/go-yaml/yaml"
+	"gopkg.in/yaml.v3 496545a6307b github.com/go-yaml/yaml"
+	"nhooyr.io/websocket v1.8.7 github.com/nhooyr/websocket"
+	"salsa.debian.org/vasudev/gospake2 adcc69dd31d5 salsa.debian.org/vasudev/gospake2/-/"
+	"salsa.debian.org/vasudev/ed25519group 3aff6cc605d469a7d618240e4a5cee3e1a861b43 salsa.debian.org/vasudev/ed25519group/-/"
+)
+
+inherit golang-vcs-snapshot
```
thesamesam, line 66: I _think_ this eclass is considered obsolete and we should be using go-module.eclass instead.

net-analyzer/termshark/termshark-2.3.0.ebuild
```diff
@@ -0,0 +1,98 @@
...
+
+DESCRIPTION="A terminal UI for tshark, inspired by Wireshark"
+HOMEPAGE="https://termshark.io/"
+
+SRC_URI="https://github.com/gcla/termshark/archive/v${PV}.tar.gz -> ${P}.tar.gz
+	${EGO_VENDOR_URI}"
+
+KEYWORDS="~amd64 ~arm ~x86"
```
thesamesam, line 74: Don't put ~arm here (or even ~x86 unless you tested it there): see [here](https://devmanual.gentoo.org/keywording/index.html#keywording-new-packages).

net-analyzer/termshark/termshark-2.3.0.ebuild
```diff
@@ -0,0 +1,98 @@
...
+RESTRICT="mirror"
+LICENSE="MIT"
```
thesamesam, line 76: Resort these to match skel.ebuild, so:
LICENSE/SLOT/KEYWORDS/RESTRICT.

net-analyzer/termshark/termshark-2.3.0.ebuild
```diff
@@ -0,0 +1,98 @@
...
+SLOT="0"
+
+RDEPEND="
+	net-analyzer/wireshark[dumpcap,pcap,tshark]
+"
+
+BDEPEND="dev-lang/go"
```
thesamesam, line 83: The eclass should handle this for us.

net-analyzer/termshark/termshark-2.3.0.ebuild
```diff
@@ -0,0 +1,98 @@
...
+SRC_URI="https://github.com/gcla/termshark/archive/v${PV}.tar.gz -> ${P}.tar.gz
+	${EGO_VENDOR_URI}"
+
+KEYWORDS="~amd64 ~arm ~x86"
+RESTRICT="mirror"
+LICENSE="MIT"
+SLOT="0"
+
+RDEPEND="
+	net-analyzer/wireshark[dumpcap,pcap,tshark]
```
thesamesam, line 80: I haven't chcked, but does it link against Wireshark/use its libs? If so, it needs to be in DEPEND too.

We should use := if linking against it too.

thesamesam (changes requested):

net-analyzer/termshark/termshark-2.3.0.ebuild
hamarituc: `termshark` executes `dumpcap` and `tshark` at runtime and reads data via IPC. I didn't find any indication of linking against Wireshark. I will put a comment inside the ebuild to clarify this.

net-analyzer/termshark/termshark-2.3.0.ebuild
hamarituc: not necessary as it is just a runtime dependency, see above.

net-analyzer/termshark/termshark-2.3.0.ebuild
thesamesam: Ping ;)

net-analyzer/termshark/termshark-2.3.0.ebuild
thesamesam: I don't think you force-pushed after fixing this?

net-analyzer/termshark/termshark-2.3.0.ebuild
hamarituc: I took a while to figure out how to migrate from `golang-vcs-snapshot` to `go-module`. `ebuildtester` is still running at the moment. I will push as soon as it succeeds ;) I just resolved the items I fixed locally.

juippis (commented):
I don't know if you follow our gentoo-dev mailing list, but we're currently reworking the way we're packaging go software in Gentoo. I believe you should aim to package this via the "new way", and unfortunately I'm not very familiar with it right now.

thesamesam:
Switching to dep tarball & merging. Thanks for your huge patience. It'll be quicker from now on.

My cheesy script will be useful, maybe, hopefully: https://github.com/thesamesam/sam-gentoo-scripts/blob/main/maint/bump-go. It needs some more love but it works well enough for me to do Go dep tarballs quickly.

PR #24022, opened 2022-01-30
More PEP517 & other distutils-r1 fixes

mgorny:
CC @gentoo/python 

I've tested it with `dev-python/python-systemd` so far. I'd like to test more packages before merging it.

### DIFF
--- a/eclass/distutils-r1.eclass
+++ b/eclass/distutils-r1.eclass
@@ -544,6 +544,9 @@ esetup.py() {
 	if [[ ${DISTUTILS_USE_SETUPTOOLS} == pyproject.toml ]]; then
 		setup_py=( -m pyproject2setuppy )
 	elif [[ ! -f setup.py ]]; then
+		if [[ ! -f setup.cfg ]]; then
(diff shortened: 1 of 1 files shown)


SoapGentoo (approved):

mgorny:
Ok, this isn't going to work. While it fixes a few packages, it causes import conflicts on more where we currently rely on modules being loaded in cwd. Maybe we can make it with some conditional to reduce the breakage, e.g. limiting it to PEP517 mode but not sure if it's worth it.

arthurzam (approved):

arthurzam:
I must say the `epytest || die` looked like a mistake to me (after a lot of usage of `epytest` without `|| die`), but wanted to check to be sure, and behold:
`epytest` returns the final return value (after doing `|| die -n`), and it is `distutils-r1_src_test` that manages the check and die.

mgorny:
No, explicit `||die` is only needed under `nonfatal`.

arthurzam (approved):

thesamesam (approved):

PR #24028, opened 2022-01-31
media-tv/plex-media-server: Plex Media Server 1.25.4.5487 is now available to everyone.

comio:
Add me as proxy maintainer and refresh Plex ebuilds, removing olds and adding the new version from Plex (no plex pass required).

Bug: https://bugs.gentoo.org/832412
Closes: https://bugs.gentoo.org/832412


thesamesam:
Hi!

Thanks for doing this.

Some minor comments:
1. No need for Bug _and_ Closes (just do Closes, as it's the same as Bug, but it closes the bug too)
2. Ideally remove old versions in a separate commit (and don't modify the init files and such for existing versions, copy them to a new version instead, and ref that in the new ebuild). This makes it easier to revert if necessary and for users to try older versions if there's a problem.
3. Could you include your signoff in commits? (`git commit -s` / `Signed-off-by: name <email>`)? Thank you!
4. https://bugs.gentoo.org/600696 may need reflecting in the ebuild.

comio:
@thesamesam 

I just upgraded the gentoo's ebuild to new release.
I hope that it's fine now.

ciao

juippis:
@thesamesam what's the situation here? Seems to work on openrc.

thesamesam:
I'm sorry, I genuinely thought we'd merged this!

PR #24041, opened 2022-02-01
profiles/features/musl: mask net-im/mattermost-desktop-bin

ceamac:
Mask binary package linked against glibc on musl profile


juippis (approved):
I guess the binary should also RDEPEND on glibc.

PR #24045, opened 2022-02-01
net-misc/turbovnc: fix java-vm dependency

ceamac:
The linked bug was triggered by setting system java-vm to openjdk-11.  turbovnc would still use openjdk-8 and fail to link with media-libs/libjpeg-turbo (which used openjdk-11).


net-misc/turbovnc/turbovnc-2.2.7-r1.ebuild
```diff
@@ -40,6 +40,10 @@ RDEPEND="
 	x11-apps/xkbcomp
 "
 
+# net-misc/turbovnc does not build this file, it merely copies it
+# from media-libs/libjpeg-turbo
+QA_PREBUILT="usr/share/turbovnc/classes/libturbojpeg.so"
```
thesamesam, line 45: It's not ideal and it'd ideally be a symlink but Java is hard to wrangle. let's kepe this in mind ofr future though.

PR #24046, opened 2022-02-01
Update zen-sources to version 5.16.5, current in repo is out of date

barkabar999:
Hello, this is a new contributor here. I have found a discrepancy in versioning. You see, the Zen Kernel team has made a new release, v5.16.5. The release happened 2 hours ago, hence the repo is out of date. Zen kernel users need to be considered here. The kernel they are using had a new release. So the time has come for a version bump.


anarchpenguin:
As the maintainer I do not need a PR, this is not a package that is bumped every time a new version is dropped.

PR #24054, opened 2022-02-02
dev-java/xerces: Bump to 2.12.2 (CVE-2022-23437) 

vaukai:


dev-java/xerces/xerces-2.12.2.ebuild
```diff
@@ -0,0 +1,55 @@
...
+	default
+	mkdir "${JAVA_RESOURCE_DIRS}" || die
+	cp -r "${JAVA_SRC_DIR}/org" "${JAVA_RESOURCE_DIRS}" || die
+	find "${JAVA_RESOURCE_DIRS}" -type f -name '*.java' -exec rm -rf {} + || die
+	rm "${JAVA_RESOURCE_DIRS}"/org/apache/xerces/{dom/org.apache.xerces.dom.DOMImplementationSourceImpl,xs/datatypes/package.html,parsers/org*} || die
+
+	local vm_version="$(java-config -g PROVIDES_VERSION)"
+	if [[ "${vm_version}" != "1.8" ]] ; then
+		rm -rv "src/org/w3c" || die
+	fi
```
fordfrog, line 49: i'm afraid this block might cause some issues during runtime. as if one compiles the package with jdk > 1.8 and then attempts to use the package with jdk 1.8, `org.w3c` won't be on the classpath and it will end up with an exception.

Flowdalic: Would a solution be to raise the (R)DEPEND requirement to jre-1.11?

vaukai: > Would a solution be to raise the (R)DEPEND requirement to j(d|r)e-1.11?

That way it'd be useless for `jdk-1.8` and would urge us to keep the older vulnerable version ...

fordfrog:
@vaukai merged, thanks!

PR #24060, opened 2022-02-03
Cinnamon 5.2 updates (22/2/3)

SparkyBluefang:
* Bump some release versions.
* Fix incorrect docs path for `cinnamon-session`. Not sure if this is the most ideal solution, but it seems to work.
* Update `nemo` to use `gtk-doc` instead of `doc`, since all of the docs are gtk-docs.
* Fix missing dep in `cinnamon` and replace optional EDS support patch with one from upstream.


gnome-extra/cinnamon/files/cinnamon-5.2.7-eds-detection.patch
```diff
@@ -0,0 +1,112 @@
+From ef463cc0aaedd714f2956daab227aeda1d87897e Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: Include links to where patches come from: https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html#clean-patch-howto

thesamesam: let me know when this bit is fixed

thesamesam:
Thank you as ever for your hard work!

PR #24065, opened 2022-02-03
python-utils-r1: python_has_version helper

mgorny:
The rough idea is to collect as much as possible in one batch, to avoid more cache regens in the neat time.

So far, besides various cleanups and refactorings:
1. New `distutils_pep517_installer` helper, for special cases where we need to install additional packages.
2. Much more verbosity in any-r1-style interpreter choice, to help users understand what's happening.
3. `python_has_version` wrapper with verbose output suitable for the above.
4. Regression fix for `build_sphinx`.
5. stdlib version matching for `python_gen*` and `python_setup`.

Will explain in greater detail when I send the patches to the ml.

CC @gentoo/python

### DIFF
--- a/dev-libs/libwacom/libwacom-1.12.ebuild
+++ b/dev-libs/libwacom/libwacom-1.12.ebuild
@@ -35,9 +35,10 @@ BDEPEND="
 "
 
 python_check_deps() {
-	has_version -b "dev-python/python-libevdev[${PYTHON_USEDEP}]" &&
-	has_version -b "dev-python/pyudev[${PYTHON_USEDEP}]" &&
-	has_version -b "dev-python/pytest[${PYTHON_USEDEP}]"
+	python_has_version \
+		"dev-python/python-libevdev[${PYTHON_USEDEP}]" \
+		"dev-python/pyudev[${PYTHON_USEDEP}]" \
+		"dev-python/pytest[${PYTHON_USEDEP}]"
 }
 
 pkg_setup() {
--- a/eclass/distutils-r1.eclass
+++ b/eclass/distutils-r1.eclass
@@ -424,11 +424,9 @@ distutils_enable_sphinx() {
 		python_check_deps() {
 			use doc || return 0
 
-			local hasv_args=( -b )
-			[[ ${EAPI} == 6 ]] && hasv_args=( --host-root )
 			local p
 			for p in dev-python/sphinx "${_DISTUTILS_SPHINX_PLUGINS[@]}"; do
-				has_version "${hasv_args[@]}" "${p}[${PYTHON_USEDEP}]" ||
+				python_has_version "${p}[${PYTHON_USEDEP}]" ||
 					return 1
 			done
 		}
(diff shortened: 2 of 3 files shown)


mgorny:
New fancy output:
Screenshot_2022-02-04_15-46-02

arthurzam (approved):
Looks good.
Maybe we should kill and port last stragglers of EAPI=6 for the python eclasses... (outside of scope for this PR).

By the fact we are finally removing various long deprecated things, should we give the full time for review over gentoo-dev ML?
I think we have given enough time to migrate (I mean overlays, ::gentoo is clear), so maybe a useless wait, but this wait might remove possible conflicts with other developers.

Also good idea might be to mention in the [00/*] message of the patchset a list of newly added functions, so people know they exist (and read the docs for how to use) and we recommend to start using them.
I see those ML mails as a review process, but also a way of micro changelog for other bystanders developers. So mentioning the changes might be worth it.

arthurzam (approved):
Changes looks nice.
I quite like the resulting move to `${EPYTHON} - <<-EOF`. It is longer, but looks more organized, and less error prone (If for example we forget the `\`).

The changes for stdlib are also useful, and it looks quite nice.

eclass/distutils-r1.eclass
```diff
@@ -936,9 +936,13 @@ _distutils-r1_get_backend() {
 	if [[ -f pyproject.toml ]]; then
 		# if pyproject.toml exists, try getting the backend from it
 		# NB: this could fail if pyproject.toml doesn't list one
-		build_backend=$("${EPYTHON}" -c 'import tomli; \
-			print(tomli.load(open("pyproject.toml", "rb")) \
-				["build-system"]["build-backend"])' 2>/dev/null)
+		build_backend=$(
```
thesamesam, line 938: Nice!

eclass/python-utils-r1.eclass
```diff
@@ -123,14 +123,19 @@ _python_set_impls() {
 	if [[ $(declare -p PYTHON_COMPAT) != "declare -a"* ]]; then
 		die 'PYTHON_COMPAT must be an array.'
 	fi
+
```
thesamesam, line 126: Thank you!

thesamesam (approved):

thesamesam (approved):

dev-libs/libwacom/libwacom-1.12.ebuild
```diff
@@ -35,9 +35,10 @@ BDEPEND="
 "
 
 python_check_deps() {
-	has_version -b "dev-python/python-libevdev[${PYTHON_USEDEP}]" &&
-	has_version -b "dev-python/pyudev[${PYTHON_USEDEP}]" &&
-	has_version -b "dev-python/pytest[${PYTHON_USEDEP}]"
+	python_has_version \
+		"dev-python/python-libevdev[${PYTHON_USEDEP}]" \
```
mattst88, line 39: Should these have `-b` like they did before?

mattst88: Oh, perfect!

mattst88 (approved):

PR #24070, opened 2022-02-03
media-gfx/hydrus: Drop 471, 467; add 472, 473

vaartis:
Closes: https://bugs.gentoo.org/832605

Seems like 471 has a problem with tests, so I replaced it with 472.


vaartis:
Added the next version too.

PR #24072, opened 2022-02-03
dev-util/its4: Avoid calling g++ directly + fix build for clang/LLVM toolchain

laumann:
This revbump accomplishes two things:

 * Bump EAPI=8 from 5. The biggest change is adding `eapply_user` in
   `src_prepare()`

 * Use `--with-cpp` argument for call to `./configure`, so the script
   doesn't try to find a compiler

Closes: https://bugs.gentoo.org/724268

---

The config.C file has five instances of string literals being
immediately followed by NEWLINE (a macro that expands to another string
literal). C++11 requires a space between string literals.

Closes: https://bugs.gentoo.org/738936

### DIFF
--- a/dev-util/its4/its4-1.1.1-r2.ebuild
+++ b/dev-util/its4/its4-1.1.1-r2.ebuild
@@ -22,6 +22,11 @@ src_prepare() {
 	sed -i \
 		-e 's/$(CC) -o/$(CC) $(OPTIMIZATION) $(EXTRA_FLAGS) -o/' \
 		"${S}"/Makefile.in || die
+
+	# Bug 738936 fails to compile with clang/LLVM toolchain
+	sed -i \
+		-e 's/"NEWLINE/" NEWLINE/g'\
+		"${S}"/config.C || die
 	eapply_user
 }
 
--- a/dev-util/its4/its4-1.1.1-r2.ebuild
+++ b/dev-util/its4/its4-1.1.1-r2.ebuild
@@ -15,20 +15,10 @@ KEYWORDS="~amd64 ~ppc ~x86"
 
 S="${WORKDIR}/${PN}"
 
-src_prepare() {
-	sed -i \
-		-e 's,iostream.h,iostream,g'\
-		"${S}"/configure || die
-	sed -i \
-		-e 's/$(CC) -o/$(CC) $(OPTIMIZATION) $(EXTRA_FLAGS) -o/' \
-		"${S}"/Makefile.in || die
-
-	# Bug 738936 fails to compile with clang/LLVM toolchain
-	sed -i \
-		-e 's/"NEWLINE/" NEWLINE/g'\
-		"${S}"/config.C || die
-	eapply_user
-}
+PATCHES=(
+	"${FILESDIR}/${PN}-1.1.1-r2-cpp-headers-and-opt-flags.patch"
+	"${FILESDIR}/${PN}-1.1.1-r2-ensure-spaces-in-string-literals.patch" # bug 738936
+)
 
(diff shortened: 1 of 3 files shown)


laumann:
Difference between -r1 and -r2:

```diff
--- its4-1.1.1-r1.ebuild	2022-01-26 10:27:28.181650326 +0100
+++ its4-1.1.1-r2.ebuild	2022-02-03 21:36:01.495569771 +0100
@@ -1,7 +1,7 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI="5"
+EAPI=8
 
 inherit toolchain-funcs
 
@@ -11,7 +11,7 @@
 
 LICENSE="ITS4"
 SLOT="0"
-KEYWORDS="~amd64 ppc x86"
+KEYWORDS="~amd64 ~ppc ~x86"
 
 S="${WORKDIR}/${PN}"
 
@@ -22,13 +22,14 @@
 	sed -i \
 		-e 's/$(CC) -o/$(CC) $(OPTIMIZATION) $(EXTRA_FLAGS) -o/' \
 		"${S}"/Makefile.in || die
+	eapply_user
 }
 
 src_configure() {
 	# WARNING
 	# non-standard configure
 	# do NOT use econf
-	./configure --prefix=/usr --mandir=/usr/share/man --datadir=/usr/share/its4 || die "configure failed"
+	./configure --with-cpp=$(tc-getCXX) --prefix=/usr --mandir=/usr/share/man --datadir=/usr/share/its4 || die "configure failed"
 }
 
 src_compile() {
```

laumann:
I asked about these fixes in #gentoo-proxy-maint and was told that neither change necessarily require a revbump, but EAPI updates generally do.

So another way to add these fixes is to forego the EAPI change and just apply the fixes to the -r1 ebuild.

dev-util/its4/its4-1.1.1-r2.ebuild
```diff
@@ -0,0 +1,48 @@
...
+src_prepare() {
+	sed -i \
+		-e 's,iostream.h,iostream,g'\
+		"${S}"/configure || die
+	sed -i \
+		-e 's/$(CC) -o/$(CC) $(OPTIMIZATION) $(EXTRA_FLAGS) -o/' \
+		"${S}"/Makefile.in || die
+
+	# Bug 738936 fails to compile with clang/LLVM toolchain
+	sed -i \
```
thesamesam, line 27: Would you mind using a patch for this instead? It feels a bit fragile.

thesamesam: (Feel free to fold the other 2 seds into their own single patch too - i.e. it's okay to combine those two into one, but I'd do config.C in its own.)

laumann: Sure thing.

laumann: Done. I downloaded the sources, made a temporary git repo out of it, and applied the sed lines turning them into patches in the process. The patches are created with git format-patch.

laumann:
Should also note that I tested the Clang/LLVM with the following:

```
CXX="x86_64-pc-linux-gnu-clang++" CC="x86_64-pc-linux-gnu-clang" ebuild ./its4-1.1.1-r2.ebuild clean install
```
confirmed that the current stable its4-1.1.1-r1.ebuild fails to build, and that its4-1.1.1-r2.ebuild builds successfully.

thesamesam:
Thank you!

PR #24073, opened 2022-02-03
acct-user/vdr: fix creation of user

madmartin:


acct-user/vdr/vdr-0-r1.ebuild
```diff
@@ -1,4 +1,4 @@
-# Copyright 2021 Gentoo Authors
+# Copyright 2022 Gentoo Authors
```
juippis, line 1: I think you're manually updating this? Our tools (repoman, pkgdev) would've fixed it with `# Copyright 2021-2022 Gentoo Authors` which is the correct form.

juippis (approved):

PR #24078, opened 2022-02-04
add new package : sys-kernel/xanmod-sources

hedmo:
hi . for more info : https://gitlab.com/src_prepare/src_prepare-overlay/-/merge_requests/258

i have changed inherit , metadata.xml , eapi and clean out .

regards


SoapGentoo:
sorry, we can't add new kernel sources that are proxy maintained. Historically, everything but `vanilla-sources` and `gentoo-sources` had proxy maintainers that disappeared after half a year again. This ontop of our policy not to add new packages via proxy-maint unless the submitter has extensive history documenting their reliability. Maybe consider putting it in `::guru` though?

hedmo:
@SoapGentoo . thanks for the fast reply . as you can see in the link i added i was told to make  a PR from @alicef_  .

regards

SoapGentoo:
That is unfortunate and regrettable for the following reasons
1. Alice doesn't speak for proxy-maint and its policies
2. is unlikely to actually help maintain it

hedmo:
okej . thank you for the info and your time . i will close it then.

regards

hedmo:
hi again . i reopen it again ... i have done some research and if i understand it the main problem is the metadata.xml .. as i am using repoman i like to have a clean output and have to have a metadata.xml .

Flowdalic:
> i have done some research and if i understand it the main problem is the metadata.xml

The main problem is that there where bad experiences with third-party kernels in Gentoo. Largely due to the maintainers of those kernel packages disappearing. I concur with @SoapGentoo that this should be put in ::guru, at least for now. If this kernel proves to be a major success and it shows a good maintenance track record in ::guru, then we could consider adding it to ::gentoo.

hedmo:
@Flowdalic . thank you for a clear answer . i have been running and maintain it for quite some time now . i will keep it at src_prepare-overlay an my local for now .

regards

juippis:
I was mostly worried about the proxy-maint team not having a reliable way of testing kernel sources, unless one member actually uses them. But gkernelci might help us to provide such support.

PR #24082, opened 2022-02-04
app-text/mupdf: fixed unconditional BDEPEND

rndxelement:
The x11-base/xorg-proto is only needed if the X use flag is enabled.

Closes: https://bugs.gentoo.org/832628


juippis (approved):
Thanks, makes sense.

PR #24088, opened 2022-02-05
dev-python/virtualenv: PoC on unbundling seed wheels

mgorny:
virtualenv bundles a number of wheels to bootstrap packages inside the new venv. Let's try to install them as part of Gentoo packages and replace the bundled wheels with our ones (automatically bumping to the newest versions as a result).

We'll still have to keep the old versions for old Pythons though.

CC @gentoo/python


arthurzam (approved):
Looks good and should work, but quite complicated logic (I don't mean it as a bad thing).

mgorny:
We've done this already, and much simpler.

PR #20619, opened 2021-04-30
gnome-extra/gnome-commander: version bump to 1.12.1

turboscholz:
Signed-off-by: Uwe Scholz <u.scholz83@gmx.de>

PR #24092, opened 2022-02-05
gnome-extra/gnome-commander: add version 1.14.0

turboscholz:
This version removes the dependency from gnome-vfs.

Also, the changelog URL is corrected in the metadata.xml file.


gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
a17r, line 4: What about EAPI-8?

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	chm? ( dev-libs/chmlib )
+	exif? ( >=media-gfx/exiv2-0.14 )
+	gsf? ( gnome-extra/libgsf:= )
+	pdf? ( >=app-text/poppler-0.18 )
+	samba? ( gnome-base/gvfs[samba] )
+	taglib? ( >=media-libs/taglib-1.4 )
+	unique? ( >=dev-libs/libunique-0.9.3:1 )
+"
+BDEPEND="
+	${RDEPEND}
```
a17r, line 40: `RDEPEND` make no sense in `BDEPEND`.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	dev-util/gtk-doc-am
+	sys-devel/gettext
+	virtual/pkgconfig
+"
+DEPEND="
```
a17r, line 45: This is likely where you wanted to put `RDEPEND` (instead of above).

a17r (changes requested):

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
turboscholz: I will check. Why does `repoman -dx full` does not complain?

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
turboscholz: Hm.... It was like that also in previous versions. I can remove it and check again.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
a17r: It will start complaining about it once EAPI-7 becomes deprecated - we are quite some time away from that, but no reason to wait for porting either. ;)

turboscholz: Aha :+1:

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
a17r: I mentioned the problem here: https://github.com/gentoo/gentoo/pull/20619#discussion_r624756558

For whatever reason this slipped through on pushing the PR unfortunately. Most likely the code was changed and the original change request was no longer visible to me as a consequence.

turboscholz: Oha. Got it, thanks. Will change accordingly.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+HOMEPAGE="https://gcmd.github.io/"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="chm exif gsf pdf samba taglib test +unique"
+RESTRICT="!test? ( test )"
+
+RDEPEND="
+	app-text/yelp-tools
```
mattst88, line 20: yelp-tools should likely be in BDEPEND.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	dev-libs/atk
```
mattst88, line 21: I see no evidence that this dependency is correct.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	dev-libs/glib:2
```
mattst88, line 22: Should be `>=dev-libs/glib-2.70.0:2`. The dependency in configure.ac is 2.66, but there's an additional check for 2.70, so just require that since it's already stable.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	gnome-base/gconf:2
```
mattst88, line 23: I see no evidence this is required.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	media-libs/fontconfig
```
mattst88, line 24: I see no evidence this is required.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	media-libs/freetype
```
mattst88, line 25: I see no evidence this is required.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	media-libs/harfbuzz:=
```
mattst88, line 26: I see no evidence this is required.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	x11-libs/cairo
```
mattst88, line 27: I see no evidence this is required.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	x11-libs/gdk-pixbuf:2
```
mattst88, line 28: This is required but it's a dependency of gtk+, and configure.ac doesn't check for it. I think we should probably remove it?

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	x11-libs/gtk+:2
```
mattst88, line 29: Should be `>=x11-libs/gtk+-2.24:2`

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	x11-libs/pango
```
mattst88, line 30: I see calls to `pango_` functions, but no `#include`s of pango headers nor checks in configure.ac for pango.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	chm? ( dev-libs/chmlib )
```
mattst88, line 31: This does nothing. I've made a PR to remove the configure.ac logic upstream: https://gitlab.gnome.org/GNOME/gnome-commander/-/merge_requests/27

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	exif? ( >=media-gfx/exiv2-0.14 )
+	gsf? ( gnome-extra/libgsf:= )
```
mattst88, line 33: Should be `gsf? ( >=gnome-extra/libgsf-1.12:= )`

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
turboscholz: Updated. Well, I only can admit that it slipped through my fingers too... °_°

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+	pdf? ( >=app-text/poppler-0.18 )
+	samba? ( gnome-base/gvfs[samba] )
+	taglib? ( >=media-libs/taglib-1.4 )
+	unique? ( >=dev-libs/libunique-0.9.3:1 )
+"
+BDEPEND="
+	dev-util/gtk-doc-am
+	sys-devel/gettext
```
mattst88, line 41: Should be `>=sys-devel/gettext-0.19.7`

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
```diff
@@ -0,0 +1,65 @@
...
+"
+DEPEND="
+	${RDEPEND}
+	test? ( >=dev-cpp/gtest-1.7.0 )
+"
+
+src_configure() {
+	gnome2_src_configure \
+		--disable-static \
+		$(use_with chm libchm) \
```
mattst88, line 52: Please just pass `--without-libchm` since the config option does nothing.

mattst88 (changes requested):

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
turboscholz: there was this commit c32dfd87cbb from @juippis where this dependency was added. I did not find the time to check why this was done, maybe he can comment. But I agree that it can be removed.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
turboscholz: only available in portage is 0.21, do you really think the explicit version is needed?

a17r: In general we keep minimum version reqs until after some time (2 years) they have left the tree. gnome proj may have different standards.

mattst88: Yeah, I don't really have a preference. I guess it just seems easier to put the version in so you don't have to think about it. :shrug:

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
juippis: grep -i atk /var/db/pkg/gnome-extra/gnome-commander-1.14.0/NEEDED.ELF.2
X86_64;/usr/lib64/gnome-commander/plugins/libfileroller.so;libfileroller.so;;libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libpangocairo-1.0.so.0,**libatk-1.0.so.0**,libcairo.so.2,libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,libpangoft2-1.0.so.0,libpango-1.0.so.0,libgobject-2.0.so.0,libglib-2.0.so.0,libharfbuzz.so.0,libfontconfig.so.1,libfreetype.so.6,libstdc++.so.6,libc.so.6,libgcc_s.so.1;x86_64

scanelf -n /usr/lib64/gnome-commander/plugins/libfileroller.so
 TYPE   NEEDED FILE
ET_DYN libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libpangocairo-1.0.so.0,**libatk-1.0.so.0**,libcairo.so.2,libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,libpangoft2-1.0.so.0,libpango-1.0.so.0,libgobject-2.0.so.0,libglib-2.0.so.0,libharfbuzz.so.0,libfontconfig.so.1,libfreetype.so.6,libstdc++.so.6,libc.so.6,libgcc_s.so.1 /usr/lib64/gnome-commander/plugins/libfileroller.so

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
juippis: X86_64;/usr/lib64/gnome-commander/plugins/libfileroller.so;libfileroller.so;;libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libpangocairo-1.0.so.0,libatk-1.0.so.0,libcairo.so.2,libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,libpangoft2-1.0.so.0,libpango-1.0.so.0,libgobject-2.0.so.0,libglib-2.0.so.0,libharfbuzz.so.0,**libfontconfig.so.1**,libfreetype.so.6,libstdc++.so.6,libc.so.6,libgcc_s.so.1;x86_64

 TYPE   NEEDED FILE
ET_DYN libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libpangocairo-1.0.so.0,libatk-1.0.so.0,libcairo.so.2,libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,libpangoft2-1.0.so.0,libpango-1.0.so.0,libgobject-2.0.so.0,libglib-2.0.so.0,libharfbuzz.so.0,**libfontconfig.so.1**,libfreetype.so.6,libstdc++.so.6,libc.so.6,libgcc_s.so.1 /usr/lib64/gnome-commander/plugins/libfileroller.so

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
juippis: X86_64;/usr/lib64/gnome-commander/plugins/libfileroller.so;libfileroller.so;;libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libpangocairo-1.0.so.0,libatk-1.0.so.0,libcairo.so.2,libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,libpangoft2-1.0.so.0,libpango-1.0.so.0,libgobject-2.0.so.0,libglib-2.0.so.0,libharfbuzz.so.0,libfontconfig.so.1,**libfreetype.so.6**,libstdc++.so.6,libc.so.6,libgcc_s.so.1;x86_64

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
juippis: X86_64;/usr/lib64/gnome-commander/plugins/libfileroller.so;libfileroller.so;;libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libpangocairo-1.0.so.0,libatk-1.0.so.0,libcairo.so.2,libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,libpangoft2-1.0.so.0,libpango-1.0.so.0,libgobject-2.0.so.0,libglib-2.0.so.0,**libharfbuzz.so.0**,libfontconfig.so.1,libfreetype.so.6,libstdc++.so.6,libc.so.6,libgcc_s.so.1;x86_64

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
juippis: X86_64;/usr/lib64/gnome-commander/plugins/libfileroller.so;libfileroller.so;;libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libpangocairo-1.0.so.0,libatk-1.0.so.0,**libcairo.so.2**,libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,libpangoft2-1.0.so.0,libpango-1.0.so.0,libgobject-2.0.so.0,libglib-2.0.so.0,libharfbuzz.so.0,libfontconfig.so.1,libfreetype.so.6,libstdc++.so.6,libc.so.6,libgcc_s.so.1;x86_64

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
juippis: We don't trust indirect deps like that. If it's required by your package, you should list it as a dependency. I know gtk+ will always depend on gdk-pixbuf, but why change habits per special packages.

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
juippis: I see pango being linked to final libraries and binaries.

X86_64;/usr/lib64/gnome-commander/plugins/libfileroller.so;libfileroller.so;;libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,**libpangocairo-1.0.so.0**,libatk-1.0.so.0,libcairo.so.2,libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,**libpangoft2-1.0.so.0**,**libpango-1.0.so.0**,libgobject-2.0.so.0,libglib-2.0.so.0,libharfbuzz.so.0,libfontconfig.so.1,libfreetype.so.6,libstdc++.so.6,libc.so.6,libgcc_s.so.1;x86_64
X86_64;/usr/bin/gnome-commander;;/usr/lib64/gnome-commander;libgcmd.so.0,libgmodule-2.0.so.0,libunique-1.0.so.0,libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,**libpango-1.0.so.0**,libexiv2.so.27,libgsf-1.so.114,libgobject-2.0.so.0,libglib-2.0.so.0,libstdc++.so.6,libm.so.6,libgcc_s.so.1,libc.so.6;x86_64

juippis:
Also in addition:
```
configure: error: You need the 'flex' lexer generator to compile Gnome Commander

!!! Please attach the following file when seeking support:
!!! /var/tmp/portage/gnome-extra/gnome-commander-1.14.0/work/gnome-commander-1.14.0/config.log
 * ERROR: gnome-extra/gnome-commander-1.14.0::gentoo failed (configure phase):
 *   econf failed

```

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
turboscholz: Updated, along with the addition of `sys-devel/flex`

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
turboscholz: To be honest it is new to me that we have to list all indirect dependencies in the ebuild, even if they are linked, but unused, libraries. Is this a standard requirement? My impression is that most of the libraries you have added in commit c32dfd87 are linked into gnome commander through `libgcmd_la_LIBADD = $(GTK_LIBS)` (see [here](https://gitlab.gnome.org/GNOME/gnome-commander/-/blob/master/libgcmd/Makefile.am#L27) for example). Do I as a package maintainer always have to make sure that indirect dependencies have to be list in RDEPEND even if they are not used? In the current case it would have included all through `>=x11-libs/gtk+-2.24.0:2` in RDEPEND. Maybe I am missing something here. But at the moment I don't feel that we have to add all those packages.

I agree, though, that there should be an additional check for gdk-pixbuf, as it is indeed used in the code. Probably it was missed to check for it in the configure.ac, as @mattst88 mentioned. I think we can add it again, but only in another PR when I also add the check in the configure.ac in the upstream code.

juippis: > To be honest it is new to me that we have to list all indirect dependencies in the ebuild, even if they are linked, but unused, libraries. Is this a standard requirement?

How/why are they linked but unused? Is this some general GTK-thing?

> My impression is that most of the libraries you have added in commit [c32dfd8](https://github.com/gentoo/gentoo/commit/c32dfd87cbb848085cf9bb53d29888815a6a83d7) are linked into gnome commander through `libgcmd_la_LIBADD = $(GTK_LIBS)` (see [here](https://gitlab.gnome.org/GNOME/gnome-commander/-/blob/master/libgcmd/Makefile.am#L27) for example). 

I'm a simple guy. I don't know what `libgcmd_la_LIBADD = $(GTK_LIBS)` does or why it's there. But if I **scanelf -n** whatever gets compiled and see a list of **needed** libraries in result, then I consider they're needed by the package. 

> Do I as a package maintainer always have to make sure that indirect dependencies have to be list in RDEPEND even if they are not used? 

First of all why are they not used, but linked? Anyway yes, that's how we generally write ebuilds. 

> In the current case it would have included all through `>=x11-libs/gtk+-2.24.0:2` in RDEPEND. Maybe I am missing something here. But at the moment I don't feel that we have to add all those packages.

How would the end-result here change if gtk+ for example dropped one of these dependencies, or put them behind conditions? That's the scenario we fear with indirect deps.

All in all I hope you *understand* where the extra deps came from, and why in *future* they might be added again, as some of the tools we use flags this ebuild incomplete now. I perhaps see this more as an upstream issue...?

turboscholz: > How/why are they linked but unused? Is this some general GTK-thing?

If I get it correctly, (I am not entirely sure) the mentioned line of code in `configure.ac` command is responsible for adding linker statements after `configure` was run. After this, all created Makefiles contain the following:

```
GTK_CFLAGS = -I/usr/include/gtk-2.0 -I/usr/lib64/gtk-2.0/include -I/usr/include/pango-1.0 -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/lib64/libffi/include -I/usr/include/harfbuzz -I/usr/include/freetype2 -I/usr/include/libmount -I/usr/include/blkid -I/usr/include/fribidi -I/usr/include/uuid -I/usr/include/cairo -I/usr/include/pixman-1 -I/usr/include/libpng16 -I/usr/include/gdk-pixbuf-2.0 -I/usr/include/atk-1.0 -pthread
GTK_LIBS = -lgtk-x11-2.0 -lgdk-x11-2.0 -lpangocairo-1.0 -latk-1.0 -lcairo -lgdk_pixbuf-2.0 -lgio-2.0 -lpangoft2-1.0 -lpango-1.0 -lgobject-2.0 -lglib-2.0 -lharfbuzz -lfontconfig -lfreetype
```

But in the source code, no explicit calls of those libraries need to exist. Which would make the program work even if one of those libraries are missing on the system. It is responsibility of upstream code to check via configure.ac if a library (for example `cairo`) is really existing on the system via PKG_CHECK_MODULES _if_ it is used in the code. And only if that is the case, I as the gentoo package maintainer need to make sure that the library is also existing in RDEPEND. That is how I understand it.

You can, for example check `media-sound/potamus`, I think this tool has the same issue (it is a C program using gtk+ libs):
```
$ scanelf -n /usr/bin/potamus
 TYPE   NEEDED FILE 
ET_DYN libglade-2.0.so.0,libgtk-x11-2.0.so.0,libgdk_pixbuf-2.0.so.0 ...
```
but the gentoo ebuild does not explicitly list `libgdk_pixbuf`, instead, it just forces `x11-libs/gtk+:2` in RDEPEND.

That being said, if gtk+ would drop one of those libraries, I would not care because Gnome Commander is not using them (because in the `configure.ac` there is no hint that it is a expected library of the system and also there are no source code calls to those libraries).

I would suggest in the end not to believe in tools claiming that a library is needed if it was just linked in by the build process. But this is just my suggestion. I am a proxy-maintainer and do not know the internals of Gentoo ebuild creation process completely. I don't like to add more complexity as needed, so I am okay with how the current gnome-commander-1.14.0 ebuild looks like. Hope you understand what I mean. (And yes, I am also the upstream maintainer, so I know that many of those libraries added in commit [c32dfd8](https://github.com/gentoo/gentoo/commit/c32dfd87cbb848085cf9bb53d29888815a6a83d7) are not explicitly used.

a17r: Are our standard LDFLAGS not passed to this package? Unused libraries should not remain linked. https://bugs.gentoo.org/234710

turboscholz: Thanks for this hint, @a17r. I was not aware of `--as-needed`, but now I understand that this flag is added in the build through `emerge`. Documentation can be found [here](https://wiki.gentoo.org/wiki/Project:Quality_Assurance/As-needed). So I quickly compared the output of the `scanelf` command of my local installation of gnome commander (trough `./autogen.sh && make && make install`)

```
scanelf -n /usr/local/bin/gnome-commander 
 TYPE   NEEDED FILE 
ET_DYN libgcmd.so.0,libgmodule-2.0.so.0,libpthread.so.0,libunique-1.0.so.0,
libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libpangocairo-1.0.so.0,libatk-1.0.so.0,
libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,libpangoft2-1.0.so.0,libpango-1.0.so.0,
libharfbuzz.so.0,libfontconfig.so.1,libfreetype.so.6,libexiv2.so.27,libtag.so.1,
libz.so.1,libgsf-1.so.114,libxml2.so.2,libpoppler-glib.so.8,libgobject-2.0.so.0,
libglib-2.0.so.0,libcairo.so.2,libstdc++.so.6,libm.so.6,libgcc_s.so.1,
libc.so.6 /usr/local/bin/gnome-commander
```
... with the output after I did a `emerge gnome-commander` of the new version 1.14.0:
```
scanelf -n /usr/bin/gnome-commander 
 TYPE   NEEDED FILE 
ET_DYN libgcmd.so.0,libgmodule-2.0.so.0,libpthread.so.0,libunique-1.0.so.0,
libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,
libpango-1.0.so.0,libexiv2.so.27,libpoppler-glib.so.8,libgobject-2.0.so.0,
libglib-2.0.so.0,libstdc++.so.6,libm.so.6,libgcc_s.so.1,libc.so.6 /usr/bin/gnome-commander
```
As one can see, the unnecessary libraries are not linked into the emerged binary. So we are good here, at least from my perspective and the PR can be merged.

@juippis Did you build gnome commander manually like I did in the first step above?

turboscholz: Oh, okay, I see now that he used `/usr/lib64/gnome-commander/plugins/libfileroller.so` for scanning ...

turboscholz: Well, and there I can confirm what @juippis saw:
```
scanelf -n /usr/lib64/gnome-commander/plugins/libfileroller.so
 TYPE   NEEDED FILE 
ET_DYN libgtk-x11-2.0.so.0,libgdk-x11-2.0.so.0,libpangocairo-1.0.so.0,
libatk-1.0.so.0,libcairo.so.2,libgdk_pixbuf-2.0.so.0,libgio-2.0.so.0,
libpangoft2-1.0.so.0,libpango-1.0.so.0,libgobject-2.0.so.0,libglib-2.0.so.0,
libharfbuzz.so.0,libfontconfig.so.1,libfreetype.so.6,libstdc++.so.6,
libc.so.6,libgcc_s.so.1 /usr/lib64/gnome-commander/plugins/libfileroller.so
```
although I see a `--as-needed` flag in the linking step in the emerge output. I honestly don't know why those libraries are still linked. But I also can assure that the fileroller plugin does not use those libraries in the code at all.

libtool: link: x86_64-pc-linux-gnu-g++  -fPIC -DPIC -shared 
-nostdlib /usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../lib64/crti.o 
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/crtbeginS.o  .libs/file-roller-plugin.o   
-lgtk-x11-2.0 -lgdk-x11-2.0 -lpangocairo-1.0 -latk-1.0 -lcairo -lgdk_pixbuf-2.0 
-lgio-2.0 -lpangoft2-1.0 -lpango-1.0 -lgobject-2.0 -lglib-2.0 
-lharfbuzz -lfontconfig -lfreetype -Wl,**--as-needed** -L/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0 
-L/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../lib64 
-L/lib/../lib64 -L/usr/lib/../lib64 -L/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../x86_64-pc-linux-gnu/lib 
-L/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../.. 
-lstdc++ -lm -lc -lgcc_s /usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/crtendS.o 
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../lib64/crtn.o  
-march=ivybridge -O2 -Wl,-O1   -Wl,-soname -Wl,libfileroller.so -o .libs/libfileroller.so

juippis: Thanks for digging into it. But deep down, is this an upstream or gentoo issue?
My CPU is kind of occupied for today with Firefox bumps but I should be able to test and merge this package tomorrow, unless of course, someone gets there first.

turboscholz: From what the blog post above is saying this seems to be an upstream issue of libtool when using autotools and linking a shared library. Not specifically Gentoo related. I am actually working on migrating over from autotools and make to meson and ninja, which are the modern replacements. Hopefully this issue will go away automatically, then.

juippis (approved):
Had this, 
```
/bin/sh: line 1: glib-mkenums: command not found
make[2]: *** [Makefile:695: org.gnome.gnome-commander.enums.xml] Error 127
```
so I'm going to add glib-utils to BDEPEND. But with that, thanks for your patience! And I hope the efforts of porting to meson will pay off.

turboscholz:
Am Wed, 09 Feb 2022 09:44:23 -0800 schrieb juippis:
>
> @juippis approved this pull request.
>
> Had this,
> ```
> /bin/sh: line 1: glib-mkenums: command not found
> make[2]: *** [Makefile:695: org.gnome.gnome-commander.enums.xml] Error 127
> ```
> so I'm going to add glib-utils to BDEPEND. But with that, thanks for your patience! And I hope the efforts of porting to meson will pay off.

Thank's a lot for the findings!

gnome-extra/gnome-commander/gnome-commander-1.14.0.ebuild
turboscholz: Hi! It took a while but since yesterday, [gnome-commander-1.16.0](https://github.com/gentoo/gentoo/blob/master/gnome-extra/gnome-commander/gnome-commander-1.16.0.ebuild) is in portage which does not have this issue with that many linked libraries anymore - as the program uses Meson now. Example:
```
$ scanelf -n /usr/lib64/gnome-commander/plugins/libfilerollerplugin.so
 TYPE   NEEDED FILE 
ET_DYN libglib-2.0.so.0,libgio-2.0.so.0,libgobject-2.0.so.0,libgtk-x11-2.0.so.0,libgdk_pixbuf-2.0.so.0,libstdc++.so.6,libgcc_s.so.1,libc.so.6 /usr/lib64/gnome-commander/plugins/libfilerollerplugin.so
```

I'm happy :smiley:

PR #24097, opened 2022-02-06
www-apps/radicale: add missing RDEPENDS on setuptools

henning-schild:
Closes: https://bugs.gentoo.org/811117


henning-schild:
I am not sure the first commit is a proper solution for that bug, or how important that bug is after all.

I decided to put the version bump in the same PR to not cause merge conflicts, but i would not want the bump be stalled by tinderbox bug for too long.

juippis (approved):
```
 * QA Notice: This package seems to contain tests but they are not enabled.
 * Please either run tests (via distutils_enable_tests or declaring
 * python_test yourself), or add RESTRICT="test" along with an explanatory
 * comment if tests cannot be run.
```
You might receive a bug report about this, but let's hope not. Time to drop some older versions? :)

henning-schild:
> ```
>  * QA Notice: This package seems to contain tests but they are not enabled.
>  * Please either run tests (via distutils_enable_tests or declaring
>  * python_test yourself), or add RESTRICT="test" along with an explanatory
>  * comment if tests cannot be run.
> ```
> 
> You might receive a bug report about this, but let's hope not. Time to drop some older versions? :)

I am already on both. And there are bugs on both.

PR #24122, opened 2022-02-07
profiles/base: mask USE=default-libcxx on sys-devel/clang

Jannik2099:
As discussed in #gentoo-llvm, this flag is too much of a footgun especially with the rising popularity of system-clang. Any system utilizing this is already likely broken without further special tweaks


profiles/base/package.use.mask
```diff
@@ -6,6 +6,15 @@
 # This file is only for generic masks. For arch-specific masks (i.e.
 # mask everywhere, unmask on arch/*) use arch/base.
 
+# Michał Górny <mgorny@gentoo.org> (2022-02-07)
+# (on behalf of Jannik Glückert <jannik.glueckert@gmail.com>)
+# libstdc++ and libc++ have ABI-incompatible containers,
+# and any system using libc++ without further advanced tweaks
+# is most likely already broken. Still keep the flag around for adventurers.
+# With the rising popularity of system-clang this flag is becoming
+# a bigger and bigger footgun. Other default-* flags have no ABI impact.
```
mgorny, line 15: How about:

```suggestion
# libstdc++ and libc++ have different public ABI that can cause fatal
# incompatibilities unless all C++ packages are built with the same
# stdlib.  Mask the flag by default to prevent accidental damage.
```

Jannik2099:
I think a libcxx subprofile would be better, as we have no data on how common usage of GNU C++ containers (which would be exclusive to libstdc++) is, and because you still may occasionally encounter a package down the line that needs gcc, which default-libcxx could make impossible to install. Additionally, packages may unconsciously rely on implementation or undefined behavior which libstdc++ generally aims to implement in the safest / least breaking way

Arfrever:
Currently `features/clang` profile enables all LLVM-related things, not just using Clang as C/C++ compiler.
E.g. it sets `LD="ld.lld"`, `LDFLAGS="${LDFLAGS} -fuse-ld=lld -rtlib=compiler-rt -unwindlib=libunwind -Wl,--as-needed"` and variables for `llvm-*` tools (`AR="llvm-ar"`, `AS="llvm-as"`, ...).

Jannik2099:
I'm aware. but none of those have any ABI impact, unlike libc++ - lld has been tested in tinderboxes and we have eliminated most related incompatibilities, compiler-rt is literally just replacing startup files which are... not particularly exciting, and the various binutils replacements are mostly dropins (except as). More importantly, all of those options can be selectively switched back for problematic packages.

Using a different STL does not bring this luxury and you're essentially trapped after building a sufficient amount of packages with it. There has been no widespread testing with libc++ and most system-clang users I've spoken to do not use it. Additionally there may be behaviour differences that are permitted by the standard but sometimes overlooked by developers.

It's definitely possible to run a system with libc++, but we do not have remotely enough data as to how much churn users would expect and how many roadblocks they could hit, and there is no clear recovery path if such a roadblock is encountered - hence I'd prefer to see it in a seperate subprofile for more experienced users

PR #24124, opened 2022-02-08
dev-util/ostree: Removed deletion of 15_ostree script

jpds:
This is essential for Gentoo-based ostree deployments and does nothing on non-ostree installs.


zmedico:
Thanks!

PR #24126, opened 2022-02-08
media-gfx/blender: Bump to v3.0.1

xxkfqz:


DarkDefender:
As this is just a copy of the existing 3.0, ebuild. LGTM!

Sadly I have a lot of IRL stuff going on right now so I haven't had time to spend on bumping ebuilds and general maintenance :disappointed: 

Perhaps you could also bump the 2.93 ebuild as well? We are currently two point releases behind there as well.

(I don't have commit rights to the gentoo repo, so you will have to wait for the gentoo repo maintainers to merge this)

DarkDefender:
Note that you will also have to update the commit messages as they do not have any sign off section.
Here is a reference for the things you can include in the commit messages: https://www.gentoo.org/glep/glep-0066.html#commit-messages

xxkfqz:
@DarkDefender Is sign off section is strongly required? I just copied someone else's ebuild without any edits. Plus, I don't want to reveal my real name

DarkDefender:
I think that it is required, yes. But don't take my word for it.
Well, see when some people with actual authority drops by :)

Worst case, I can do it. But I currently don't have any stable internet connection or computer.
So it might take a while.

thesamesam:
Ideally yes, but it doesn't matter so much if you didn't change the ebuild at all.

thesamesam:
I ended up doing a substantial amount of work to sort out openexr 3 which involved a lot of ebuild changes, but thank you!

(Nothing wrong with you ebuild but it needed doing anyway)

PR #24127, opened 2022-02-08
www-apps/gitea: security bump

hydrapolic:
https://github.com/go-gitea/gitea/releases/tag/v1.16.1
SECURITY - Update JS dependencies, fix lint

Tested on amd64, thanks!


hydrapolic:
Thank you @thesamesam

PR #24129, opened 2022-02-08
app-containers/crun: bump to 1.4.2 and cleanup

robertgzr:


MatthewGentoo:
Do you think it's worth waiting until a newer version is stabilised before dropping version 0.19.1-r1?

PR #24130, opened 2022-02-08
dev-java/reflections: Bump to 0.9.12

vaukai:
Get rid of dependency on guava.


fordfrog:
the latest version i can see at github is 0.10.2. is that incompatible with `dev-java/jna`?

vaukai:
> the latest version i can see at github is 0.10.2. is that incompatible with `dev-java/jna`?

`0.10.2` depends on [jboss-vfs](https://github.com/ronmamo/reflections/blob/5ff7a1d366e8656dfa0f2f476991b558b037e919/pom.xml#L104-L110) which itself has even more dependencies that all are not yet in ::gentoo.

fordfrog:
ok, and why using jar sources instead of sources from github? https://github.com/ronmamo/reflections/tree/0.9.12
also, those sources also contain tests which would make verification that the ebuild is ok much easier...

vaukai:
> ok, and why using jar sources instead of sources from github? https://github.com/ronmamo/reflections/tree/0.9.12 also, those sources also contain tests which would make verification that the ebuild is ok much easier...

I saw that only after sending the PR.  Please leave it there and I'll rebase it on Github sources later today or tomorrow.

vaukai:
Tests enabled with 3 test failures

PR #24133, opened 2022-02-08
Vim update

vaibhav-rustagi:
Updating app-editors/gvim, app-editors/vim-core and app-editors-vim to v8.2.4328.


thesamesam:
Thanks! There's a security bug filed for these CVEs, so I'm going to add a 'Bug: https://bugs.gentoo.org/830994' tag to each commit. In future, if possible, search for a CVE on Bugzilla, and if it is not there, could you file a bug real quick (you don't need to supply details, just the CVE & package) so we don't lose it for tracking purposes?

vaibhav-rustagi:
I will make sure from next time to have a bug associated with the CL's. Thanks

PR #24139, opened 2022-02-09
net-news/snownews: add 1.9

capezotte:
There was an old PR with the a "bump to 1.9" message, but that's no longer accurate as the ebuild had to rewritten to account for big changes to the build system.


msharov:
gentoo-bot complains about there being "no bugs assigned". There are four bugs currently open for gentoo snownews:

- https://bugs.gentoo.org/725224
- https://bugs.gentoo.org/725250
- https://bugs.gentoo.org/731690
- https://bugs.gentoo.org/716716

All of them will be corrected by this PR.

net-news/snownews/snownews-1.9.ebuild
```diff
@@ -0,0 +1,46 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+inherit toolchain-funcs
```
ConiKost, line 5: New line before please.

net-news/snownews/snownews-1.9.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+DESCRIPTION="Command-line RSS feed reader"
+HOMEPAGE="https://github.com/msharov/snownews"
+SRC_URI="https://github.com/msharov/snownews/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc ~x86 ~amd64-linux ~x86-linux"
+IUSE=""
```
ConiKost, line 14: If empty, drop it.

net-news/snownews/snownews-1.9.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+COMMON_DEPEND="
+	>=dev-libs/libxml2-2.5.6
+	sys-libs/ncurses:=[unicode(+)]
```
ConiKost, line 18: Is the sorting somehow intentional? Otherwise alphabetical please.

net-news/snownews/snownews-1.9.ebuild
```diff
@@ -0,0 +1,46 @@
...
+	dev-libs/openssl
```
ConiKost, line 19: Package is slotted.

net-news/snownews/snownews-1.9.ebuild
```diff
@@ -0,0 +1,46 @@
...
+	net-misc/curl
+	sys-devel/gettext
```
ConiKost, line 21: I think, this should go into ```BDEPEND```.

net-news/snownews/snownews-1.9.ebuild
```diff
@@ -0,0 +1,46 @@
...
+"
+RDEPEND="
+	${COMMON_DEPEND}
+"
+
+DEPEND="
+	${COMMON_DEPEND}
+"
+
+IUSE="debug"
```
ConiKost, line 31: Wrong place ;-) See ```skel.ebuild```. It goes after ```IUSE```.

net-news/snownews/snownews-1.9.ebuild
```diff
@@ -0,0 +1,46 @@
...
+BDEPEND="virtual/pkgconfig"
+
+src_prepare() {
+	default
+	# Disable stripping in the build system - leave it to the package manager
+	sed -i -e '/ldflags/s/-s/-g -rdynamic/' -e '/cflags/s/-g0/-g/' Config.mk.in || die
+}
+
+src_configure() {
+	econf "$(use_with debug)"
```
ConiKost, line 41: Does the use flag ```debug``` make any sense? Based on https://github.com/msharov/snownews/blob/master/Config.mk.in, it looks like, that's it's only setting some different (debug) use flags?

net-news/snownews/snownews-1.9.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+COMMON_DEPEND="
+	>=dev-libs/libxml2-2.5.6
+	sys-libs/ncurses:=[unicode(+)]
+	dev-libs/openssl
+	net-misc/curl
+	sys-devel/gettext
+"
+RDEPEND="
+	${COMMON_DEPEND}
```
ConiKost, line 24: I don't see the logic here. Instead adding ```${COMMON_DEPEND}```, you could also add them do ```DEPEND``` and include in ```RDEPEND```?

ConiKost (changes requested):

ConiKost:
You could also add ```Closes: https://bugs.gentoo.org/796779``` into your commit message, so the corresponding bug ticket will be closed.

net-news/snownews/snownews-1.9.ebuild
msharov: `configure --with-debug` will set `debug` variable in `Config.mk`, enabling debug build flags, such as `-g`. Without that option the executable is built optimized, with `NDEBUG` define turning off asserts for example. The debug build is only needed for debugging, obviously; most people should use the optimized and stripped executable.

msharov:
Another thing is the explicit version requirement on libxml2:
```
>=dev-libs/libxml2-2.5.6
```
I am not sure where that came from; I wrote the code for the current libxml2 version, I don't know if I used any API specific to 2.5.6, since the documentation does not list minimum version numbers for every function. It would probably be better to just depend on libxml2 without any specific version.

net-news/snownews/snownews-1.9.ebuild
capezotte: That's just how the 1.6 ebuild did things, but yeah, now that you've pointed it out it makes much more sense.

capezotte:
The libxml2 mininum version requirement was added by the old version 1.6 ebuild mantainer. I suppose it was derived from their own testing.

capezotte:
This PR will now drop 1.6.10-r1. If there's interest in keeping it, I'll add it back (probably with a patch for the musl build).

net-news/snownews/snownews-1.9.ebuild
ConiKost: That's why I asked. We don't use use flags for setting various build flags, since anyone can via portage set per package own *flags and add things like ```-g```, if he wishes. Only, if debug does also enable more specific parts in the program itself.

ConiKost (changes requested):
Please don't drop directly 1.6, as it's currently marked stable.

msharov:
1.6 is "stable" only in the sense that it will stay broken forever. I certainly have no intention whatsoever to backport fixes to it. Just please let it go already...

capezotte:
> Please don't drop directly 1.6, as it's currently marked stable.

Should I remove the `Closes:` of 1.6.10-related bugs from the commit messages, then?

ConiKost:
Well, if 1.6 is really completly unusable as for now, fine to drop it, but please in a separate commit.

ConiKost (approved):

msharov:
The bot cloned the commit and didn't copy the bug links, so the bugs remain open...

PR #24141, opened 2022-02-10
app-shells/starship: Add 1.3.0

bowlofeggs:


bowlofeggs:
Two notes for this update:
* This release adds the Undistract Me feature, as requested by @pasha132 in https://github.com/gentoo/gentoo/pull/23968.
* There is already a 1.2.1 in testing. I am assuming that it's OK to have two different versions in testing at the same time, but in case that's undesirable I wanted to call attention to that here.

ionenwks:
> * There is already a 1.2.1 in testing. I am assuming that it's OK to have two different versions in testing at the same time, but in case that's undesirable I wanted to call attention to that here.

That's no issue, as the maintainer add / keep / remove / stabilize the versions you feel make sense. Common to keep previous in case there's issues or as the next stable candidate (not that it necessarily has to be the stable pick), and it's good to have the latest one around either way.

PR #24167, opened 2022-02-11
gui-apps/foot*: fix bad dependency

Sevz17:
ncurses[minimal] does not install foot terminfo, so we need to change to ncurses[-minimal]

### DIFF
--- a/gui-apps/foot-terminfo/foot-terminfo-1.11.0-r1.ebuild
+++ b/gui-apps/foot-terminfo/foot-terminfo-1.11.0-r1.ebuild
@@ -12,7 +12,7 @@ LICENSE="MIT"
 SLOT="0"
 KEYWORDS="~amd64"
 
-RDEPEND="!>=sys-libs/ncurses-6.3"
+RDEPEND="!>=sys-libs/ncurses-6.3[-minimal]"
 BDEPEND="sys-libs/ncurses"
 
 src_prepare() {
--- a/gui-apps/foot/foot-1.11.0-r1.ebuild
+++ b/gui-apps/foot/foot-1.11.0-r1.ebuild
@@ -34,7 +34,7 @@ DEPEND="
 RDEPEND="
 	${COMMON_DEPEND}
 	|| (
-		>=sys-libs/ncurses-6.3
+		>=sys-libs/ncurses-6.3[-minimal]
 		~gui-apps/foot-terminfo-${PV}
 	)
 "


ArsenArsen:
lgtm, should we keep foot-terminfo past ncurses-6.3 being lowest due to minimal? or is installing w/o terminfo a supported use-case

Sevz17:
> should we keep foot-terminfo past ncurses-6.3 being lowest due to minimal?

I think not

> Hence the recommendation to simply use ncurses’ terminfo definitions if available.

The author of foot recommends using ncurses terminfo

Sevz17:
Since the the use description says "Install a very minimal build ..." and is disabled by default I think is not a bug

ArsenArsen:
there's no need to require -minimal in that case, I think

rxvt-unicode doesn't seem to

ionenwks:
> there's no need to require -minimal in that case, I think
> 
> rxvt-unicode doesn't seem to

That's because it installs rxvt terminfo even with USE=minimal
```
        # We need the basic terminfo files in /etc for embedded/recovery. #37026
        einfo "Installing basic terminfo files in /etc..."
        local terms=(
            # Dumb/simple values that show up when using the in-kernel VT.
            ansi console dumb linux
            vt{52,100,102,200,220}
            # [u]rxvt users used to be pretty common.  Probably should drop this
            # since upstream is dead and people are moving away from it.
            rxvt{,-unicode}{,-256color}
            # xterm users are common, as is terminals re-using/spoofing it.
            xterm xterm-{,256}color
            # screen is common (and reused by tmux).
            screen{,-256color}
            screen.xterm-256color
```
Lol at the comment though, and I get the feeling this move is very ancient.

That aside if foot can't reasonably function without its terminfo (haven't tried but I assume it's the case given how bad kitty is without kitty-terminfo), I do think it wouldn't hurt to depend on [-minimal] even if it'd be the first ebuild to do so for ncurses.

ionenwks:
One bit of a downside with foot-terminfo is that users that have [-minimal] right now may end up kind of confused when we remove the package, portage will refuse to depclean it and won't suggest to set [-minimal] (the mask message will help while it's still up though), I guess it'll help if we can get rid of it as soon as possible unless do want to keep+maintain it forever.

The [-minimal] check is better than wondering why you have malfunctioning terminal either way.

ArsenArsen:
hmm, interesting. foot's terminfo is based off of xterms, so one could reasonably get away with `term=xterm-256color` in their configs (which I appear to have set at some point and forgot about, whoops)

not installing terminfo seems to clash with the policy of not having flags for small files, though? terminfo on my system with more than just base ncurses is ~7mb, and could be reasonably controlled with install masks; is `minimal` a relic of the past?

at any rate, the PR is fine, and foot is usable with xterm if it comes down to it; if that tweaking being required is not acceptable, I'm fine with merging this

PR #24168, opened 2022-02-12
net-misc/iputils, sys-apps/iproute2, net-firewall/iptables: remove obsolete USE=ipv6

mgorny:
At this point, USE=ipv6 is only used to control creating "ping6" and "tracepath6" symlinks.  Just remove the flag and install the links unconditionally.

### DIFF
--- a/sys-apps/iproute2/iproute2-5.16.0.ebuild
+++ b/sys-apps/iproute2/iproute2-5.16.0.ebuild
@@ -18,7 +18,7 @@ HOMEPAGE="https://wiki.linuxfoundation.org/networking/iproute2"
 
 LICENSE="GPL-2"
 SLOT="0"
-IUSE="atm berkdb bpf caps elf +iptables ipv6 libbsd minimal selinux"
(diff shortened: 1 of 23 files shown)


mgorny:
Let's include iproute2 too.

I've decided to add `(+)` to all ipv6 flags in revdeps, to save on future revbumps. It's extremely unlikely that we'll be removing USE=ipv6 to remove IPv6 support — we are far more likely to just enable that functionality unconditionally.

SoapGentoo (approved):

sys-apps/iproute2/iproute2-5.16.0.ebuild
```diff
@@ -59,12 +59,6 @@ doecho() {
 }
 
 src_prepare() {
-	if ! use ipv6 ; then
```
thesamesam, line 62: I think it's left over from uclibc.

https://bugs.gentoo.org/326849

mgorny:
Added iptables.

PR #24172, opened 2022-02-12
Unslot OpenEXR

thesamesam:
TODO: write explanation in commit message(s)


thesamesam:
cc @waebbl 

I've come to the conclusion that slotting can't work. It's a huge amount of work to try keep consumers working and the work actually increases as more and more software starts to support 3 as we have to patch build system paths as well as e.g includes. Then we also hit ABI issues as much of these packages must all be using the same version together (not some using 2 and some using 3).

media-gfx/openvdb/openvdb-8.2.0-r3.ebuild
```diff
@@ -46,7 +44,7 @@ RDEPEND="
 	)
 	utils? (
 		media-libs/ilmbase:=
-		media-libs/openexr:0=
```
waebbl, line 49: same here: dep on ilmbase?

media-gfx/povray/povray-3.7.0.0-r2.ebuild
```diff
@@ -0,0 +1,143 @@
...
+			for version_dir in $(echo "${EROOT}"etc/"${PN}"/* | grep "^[0-9]" | sort -rn); do
+				if [ -e "${EROOT}etc/${PN}/${version_dir}/${conf_file}" ]; then
+					mv "${EROOT}etc/${PN}/${version_dir}/${conf_file}" "${EROOT}etc/${PN}" || die
+					elog "Note: ${conf_file} moved from ${EROOT}etc/povray/${version_dir}/ to ${EROOT}etc/povray/"
+					break
+				fi
+			done
+		fi
+	done
+}
```
waebbl, line 143: Would bumping to 3.8.0 work with OpenEXR-3?

waebbl:
Trying to test this PR after work.

media-gfx/mandelbulber/mandelbulber-2.26-r1.ebuild
```diff
@@ -0,0 +1,97 @@
...
+	media-libs/libpng:0=
+	sci-libs/gsl:=
+	joystick? ( dev-qt/qtgamepad:5 )
+	opencl? (
+		dev-libs/clhpp
+		virtual/opencl
+	)
+	openexr? (
+		media-libs/ilmbase:=
+		media-libs/openexr:=
```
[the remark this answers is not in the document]
thesamesam: It needs to be imath, good spot!

media-gfx/povray/povray-3.7.0.0-r2.ebuild
thesamesam: @SoapGentoo

media-gfx/openvdb/openvdb-8.2.0-r3.ebuild
thesamesam: Ah, -r4 is fixed. So this is fine.

media-gfx/povray/povray-3.7.0.0-r2.ebuild
waebbl: Yes, I was probably thinking of that version. Saw this a few months ago, and thought it's already released.

waebbl:
Seems to work fine. Could build all of the packages in a chroot.

However, some revdeps are still missing, e.g. kde-frameworks/kimageformats. Is this intentional?

thesamesam:
> Seems to work fine. Could build all of the packages in a chroot.
> 
> However, some revdeps are still missing, e.g. kde-frameworks/kimageformats. Is this intentional?

Updated that one and blender-9999, leaving the rest as they should just be stable versions and I don't want to hit stable users just yet. Thank you for the feedback, testing, and comments!

Let's do it.

custom82:
when i try to emerge ilmbase-2.5.7-r1 

 * This package will overwrite one or more files that may belong to other
 * packages (see list below). You can use a command such as `portageq
 * owners / <filename>` to identify the installed package that owns a
 * file. If portageq reports that only one package owns a file then do
 * NOT file a bug report. A bug report is only useful if it identifies at
 * least two or more packages that are known to install the same file(s).
 * If a collision occurs and you can not explain where the file came from
 * then you should simply ignore the collision since there is not enough
 * information to determine if a real problem exists. Please do NOT file
 * a bug report at https://bugs.gentoo.org/ unless you report exactly
 * which two packages install the same file(s). See
 * https://wiki.gentoo.org/wiki/Knowledge_Base:Blockers for tips on how
 * to solve the problem. And once again, please do NOT file a bug report
 * unless you have completely understood the above message.
 * 
 * Detected file collision(s):
 * 
 *      /usr/include/OpenEXR/IlmThreadForward.h
 *      /usr/include/OpenEXR/IlmThreadExport.h
 *      /usr/include/OpenEXR/IlmThreadNamespace.h
 *      /usr/include/OpenEXR/IlmThreadMutex.h
 *      /usr/include/OpenEXR/IlmThreadSemaphore.h
 *      /usr/include/OpenEXR/IlmThread.h
 *      /usr/include/OpenEXR/IlmThreadPool.h
 *      /usr/include/OpenEXR/IexMathIeeeExc.h
 *      /usr/include/OpenEXR/IexMathFloatExc.h
 *      /usr/include/OpenEXR/IexForward.h
 *      /usr/include/OpenEXR/IexExport.h
 *      /usr/include/OpenEXR/IexNamespace.h
 *      /usr/include/OpenEXR/Iex.h
 *      /usr/include/OpenEXR/IexMacros.h
 *      /usr/include/OpenEXR/IexErrnoExc.h
 *      /usr/include/OpenEXR/IexThrowErrnoExc.h
 *      /usr/include/OpenEXR/IexMathExc.h
 *      /usr/include/OpenEXR/IexBaseExc.h
 *      /usr/lib64/libIlmThread.so
 *      /usr/lib64/libImath.so
 *      /usr/lib64/libIex.so
 * 
 * Searching all installed packages for file collisions...
 * 
 * Press Ctrl-C to Stop
 * 
 * dev-libs/imath-3.1.4-r2:3::local
 *      /usr/lib64/libImath.so
 * 
 * media-libs/openexr-3.1.4-r1:0::local
 *      /usr/include/OpenEXR/Iex.h
 *      /usr/include/OpenEXR/IexBaseExc.h
 *      /usr/include/OpenEXR/IexErrnoExc.h
 *      /usr/include/OpenEXR/IexExport.h
 *      /usr/include/OpenEXR/IexForward.h
 *      /usr/include/OpenEXR/IexMacros.h
 *      /usr/include/OpenEXR/IexMathExc.h
 *      /usr/include/OpenEXR/IexMathFloatExc.h
 *      /usr/include/OpenEXR/IexMathIeeeExc.h
 *      /usr/include/OpenEXR/IexNamespace.h
 *      /usr/include/OpenEXR/IexThrowErrnoExc.h
 *      /usr/include/OpenEXR/IlmThread.h
 *      /usr/include/OpenEXR/IlmThreadExport.h
 *      /usr/include/OpenEXR/IlmThreadForward.h
 *      /usr/include/OpenEXR/IlmThreadMutex.h
 *      /usr/include/OpenEXR/IlmThreadNamespace.h
 *      /usr/include/OpenEXR/IlmThreadPool.h
 *      /usr/include/OpenEXR/IlmThreadSemaphore.h
 *      /usr/lib64/libIlmThread.so
 * 
 * Package 'media-libs/ilmbase-2.5.7-r1' NOT merged due to file
 * collisions. If necessary, refer to your elog messages for the whole
 * content of the above message.

thesamesam:
1. Done, merged, yay!
2. As for:
> when i try to emerge ilmbase-2.5.7-r1

Please file a bug (I think you already did at https://bugs.gentoo.org/833428), but the ::local seems suspicious given I know the blocker is there. Can you suggest (ideally on the bug so I don't miss it) where the blocker is missing if you think it really is?

waebbl:
Thank you very much @thesamesam for all your effort and time spending on Gentoo projects and packages.

PR #24180, opened 2022-02-13
net-vpn/tor: checksum verification fix + verify-sig.eclass: Support for detach-signed checksums

mgorny:
@blueness, this is half-tested because you need to bump Tor keys first.


blueness:
This works.  I was about to push it, but I'm not sure if you want to do something else first since it does touch an eclass.

mgorny:
Well, technically I did only send it for review today but feel free to push anytime.

blueness:
okay pushed

PR #24185, opened 2022-02-13
 eclass/postgres.eclass: migrate to GLEP 81

ConiKost:
eclass/postgres.eclass: migrate to GLEP 81
Note: I haven't yet reserved UID, will do upon approval.

pgpool is the only package, which is calling ```postgres_new_user```.


ConiKost:
@titanofold @patricklauer Could you review? Thanks!

PR #24186, opened 2022-02-14
dev-util/jenkins-bin:lts: correct issue with service and openjdk configuration

samuelbernardo:
Correct issue with jenkins service type and change configuration to use recommended default openjdk:11 version
Bugs:
https://bugs.gentoo.org/828650
https://bugs.gentoo.org/828651


dev-util/jenkins-bin/files/jenkins-bin.confd
```diff
@@ -60,3 +60,15 @@ JENKINS_PIDFILE="/var/run/jenkins.pid"
 # War File
 #
 JENKINS_WAR="/opt/jenkins/jenkins.war"
+
+#
+# Java customization
+#
+JAVA_HOME=/usr/lib64/openjdk-11
```
graaff, line 67: What is the reason for hardcoding a specific java version here? Currently jenkins uses the system-vm version and that seems to work fine? I'd also rather not hardcode a specific version if multiple versions work (e.g. 11 and 17).

graaff (commented):
Apologies for the late review, it looks like this PR never surfaced on anyone's list to review.

dev-util/jenkins-bin/files/jenkins-bin-r2.service
```diff
@@ -3,9 +3,18 @@ Description=Jenkins Daemon
 
 [Service]
 EnvironmentFile=/etc/conf.d/jenkins
-Type=forking
-ExecStart=/usr/bin/java ${JENKINS_JAVA_OPTIONS} -DJENKINS_HOME=${JENKINS_HOME} -jar ${JENKINS_WAR} --daemon --logfile=/var/log/jenkins/jenkins.log --httpPort=${JENKINS_PORT} --debug=${JENKINS_DEBUG_LEVEL} --handlerCountMax=100 --handlerCountMaxIdle=20 --accessLoggerClassName=winstone.accesslog.SimpleAccessLogger --simpleAccessLogger.format=combined --simpleAccessLogger.file=/var/log/jenkins/access_log ${JENKINS_ARGS}
 User=jenkins
+Type=simple
```
graaff, line 7: This change has already been handled via the associated bug. If the other changes are still needed then they will need to be redone based on the r4 version of the service file.

dev-util/jenkins-bin/files/jenkins-bin.confd
samuelbernardo: > What is the reason for hardcoding a specific java version here? Currently jenkins uses the system-vm version and that seems to work fine? I'd also rather not hardcode a specific version if multiple versions work (e.g. 11 and 17).

Hello @graaff 
That time it doesn't work with the icedtea-8 release that was the default one for system-vm.
I need to install openjdk-11 that was required for multiple Jenkins plugins and the only way to do that was with this approach.
I also found this kind of approach in other Java related packages. The only missing point is that I could improve it as part of a use flag related to the Java requirement, instead of being hardcoded as you pointed out.

graaff: The ebuild now requires either jre:17 or jre:11 (and both versions work with Jenkins). This assumes that the system java is one of those, which is not necessarily the case.

dev-util/jenkins-bin/Manifest
```diff
@@ -1,2 +1,7 @@
+AUX jenkins-bin-r2.init 1782 BLAKE2B 25c78b64308be71d605f3c4f2d33c5cf1074ab431e23a170f7d36002d668def7dbd56b0b45590692047d7d88c6884ec1cc40cb4f077f7edcb853300301805002 SHA512 e0fd85f653bbf5ac80dc929cb265ac86a08047f9102d57f2ac6194cd31794310a3c062824423552a3ee65e298506332436fca037f96963358dc511982f900be8
```
thesamesam, line 1: These entries shouldn't be here, as we use thin manifests in ::gentoo.

dev-util/jenkins-bin/files/jenkins-bin.confd
Flowdalic: I am currently in a hurry, so I can only comment briefly. But I hope you get the idea.

Hardcoding a specific java version is usually not required.

```
$ depend-java-query --get-vm=virtual/jre-19
openjdk-bin-19
```

queries the system for a suitable JVM. Details about the returned JVM can be resolved via java-config. For example

```bash
$ java-config -P openjdk-bin-19
# Autogenerated by java-config
# Java Virtual Machine: Eclipse Temurin JDK 19.0.2_p7
JAVA_HOME=/opt/openjdk-bin-19.0.2_p7
JDK_HOME=/opt/openjdk-bin-19.0.2_p7
JAVAC=/opt/openjdk-bin-19.0.2_p7/bin/javac
PATH=/opt/openjdk-bin-19.0.2_p7/bin
ROOTPATH=/opt/openjdk-bin-19.0.2_p7/bin
LDPATH=/opt/openjdk-bin-19.0.2_p7/lib/:/opt/openjdk-bin-19.0.2_p7/lib/server/
MANPATH=
```

samuelbernardo: @Flowdalic So, for example, if someone wants to use openjdk-17 with Jenkins, but his default system-vm is IcedTea JDK, that means that the initialization script for Jenkins service should use java-config command to query for a supported Java VM and load it to the environment?
Don't know if there is any eclass that should help with this kind of task, something that I'm looking forward for...

Flowdalic: I've finally had the time to look more into our java tooling. Ideally jenkins-bin ebuild would use `java-pkg_dolauncher` from `java-utils-2.eclass`. This would install a launcher which would perform basically the steps I've mentioned above. I think this may requires also calling `jav_pkg_dowar` or maybe `java_pkg_dojar`, or maybe in your case `java-pkg_addcp /opt/jenkins`.

samuelbernardo: Thank you @Flowdalic for the advise. I'm going to do that.

PR #24187, opened 2022-02-14
dev-libs/libdispatch: Support musl

mid-kid:


dev-libs/libdispatch/libdispatch-5.5.ebuild
```diff
@@ -46,5 +49,6 @@ src_configure () {
 	export HOST_CXX="$(tc-getBUILD_CXX)"
 	tc-export CC CXX LD AR NM OBJDUMP RANLIB PKG_CONFIG
 
+	local mycmakeargs=( -DBUILD_TESTING=OFF )
```
thesamesam, line 52: Why?

dev-libs/libdispatch/files/libdispatch-5.3.3-musl.patch
```diff
@@ -0,0 +1,39 @@
+--- swift-corelibs-libdispatch-swift-5.3.3-RELEASE.orig/dispatch/source.h
```
thesamesam, line 1: 1. Please use 'git-like patches', i.e. a/ and b/ instead of looong.orig/ and looong/
2. Include a link to the relevant Gentoo bug(s) at the top of the patch
3. Include a link to where it's been forwarded/reported upstream if applicable (please do)

thesamesam (commented):
Thank you! Could you also update the Bug which your commit refers to (b/c dupe)?

dev-libs/libdispatch/libdispatch-5.5.ebuild
mid-kid: Avoids a warning under musl (easily fixable but eh), and the tests aren't used so I thought it'd make sense to just disable them.

thesamesam: Put a comment indicating such then (but maybe they should be enabled?)

mid-kid: I'll add a comment indicating that this disables building the testsuite. But why should it be enabled without an src_test?

thesamesam: Because we always want to enable running test suites?

and yes, comment, but explain why you're disabling it

mid-kid: Sure, I've removed the line and fixed the testing code to not produce a warning. I don't really see the point of building the tests while the package doesn't support testing, but whatever, it builds fast anyway.

thesamesam: I don't understand. How does it not support testing? I'm saying it should if tests exist?

mid-kid: Oh, like that. Yeah I have no idea why this package has no src_test, I just noticed the option and realized the tests weren't being used anyway.
I guess this is something that should be handled in a separate PR.

PR #24194, opened 2022-02-14
app-admin/pass-otp: Unconditionally depend on qrencode

lanodan:
As pointed out by mgorny on IRC,
app-admin/pass already hard-depends on it so the USE flag is useless.


mgorny:
Don't revbump over such a minor change.

PR #24196, opened 2022-02-15
net-fs/cifs-utils: introduce python USE flag

dongsupark:
To prevent cifs-utils from pulling in unnecessary python libs, we need to introduce a new USE flag `python`.
Doing that, `--enable-pythontools` will be enabled only if the python flag is turned on.
So we can disable pythontools when the flag is turned off.


SoapGentoo (approved):

whitslack:
This change introduced a bug: https://bugs.gentoo.org/856763

thesamesam:
Fixed, thanks!

whitslack:
@thesamesam: I think `python_fix_shebang` will die too if `EPYTHON` isn't set (which it won't be).

thesamesam:
absolutely, on it

PR #24203, opened 2022-02-15
app-misc/broot: bump to 1.9.3 and remove old

creideiki:


Flowdalic:
Thanks. Note that I'd like to stable broot 1.9 at some point. It is ok if we remove 1.9.1 now, but please keep 1.9.2 as a potential broot 1.9 stable candidate.

PR #24208, opened 2022-02-16
www-client/w3m: fix build with imlib2

N-R-K:
Closes: https://bugs.gentoo.org/833171


N-R-K:
The fix was merged upstream https://github.com/tats/w3m/commit/9874c87ad13934651b4faf8b3aa192b7db3894c0 and the current ebuild is a bit lagging behind anyways. Should I instead create a snapshot ebuild?

juippis (commented):
If a revbump is required for this, then yeah maybe the snapshot would be more worthwhile. But you can just fix the build by not revbumping, that's also fine.

N-R-K:
@juippis I'll create a snapshot in that case. Do I leave the current ebuild as is then or do I backport the patch? Also in the case of the snapshot, afaik new ebuilds need to marked unstable `~` on all arch, is that correct?

N-R-K:
The PR was updated to leave the current ebuild as is, and introduce a snapshot ebuild instead. Would appreciate if someone can take a look :)

www-client/w3m/w3m-0.5.3_p20220216.ebuild
```diff
@@ -0,0 +1,136 @@
...
+	fi
+	if use l10n_ja ; then
+		docinto ja
+		dodoc doc-jp/{HISTORY,README,keymap,menu}*
+		newdoc Bonus/README README.Bonus
+		docinto html/ja
+		dodoc doc-jp/*.html
+	else
+		rm -rf "${ED}"/usr/share/man/ja || die
+	fi
```
juippis, line 135: We should just always install these files unconditionally and not depend on USE flags. Do you want to look into updating that or should I?

Rest looks good, although there are some things I personally dislike, not worth going into right now.

N-R-K: > We should just always install these files unconditionally and not depend on USE flags.

It's done this way on the older ebuild as well, I haven't touched anything not related to version bump.

> Do you want to look into updating that or should I?

Simply removing the if check and leaving the block as is should work right? If so then I can make the changes myself.

N-R-K: > Simply removing the if check and leaving the block as is should work right?

Seems to have worked :) @juippis

juippis: > It's done this way on the older ebuild as well, I haven't touched anything not related to version bump.

I know. Generally when doing version bumps we try to look that things are "top notch". The `rm -f` here alerted me, since that's usually a command that's kept for years even though it hasn't been relevant for a long time. I.e., the files haven't existed for many versions.


> Simply removing the if check and leaving the block as is should work right? If so then I can make the changes myself.

You should also remove the USE flags altogether from the ebuild. I fear there might be some reverse dependency breakage, but haven't looked at that yet.

N-R-K: > You should also remove the USE flags altogether from the ebuild.

`ja` is doing something else as well, at line 78. `de` should be safe to remove I guess.

juippis (approved):

PR #24212, opened 2022-02-16
app-shells/mcfly: version bump, remove old

mm1ke:
Hi,

This is a simple version bump for `app-shells/mcfly`, however since i switched to `fzf` some time ago i'm not using mcfly anymore, which is why i'm removing myself from the maintainers.


ionenwks (approved):
> This is a simple version bump for `app-shells/mcfly`, however since i switched to `fzf` some time ago i'm not using mcfly anymore, which is why i'm removing myself from the maintainers.

I see, I use fzf a bit as well so it's understandable.

I'll send the up for grabs to gentoo-dev, thanks for the bump and leaving the package in a good state.

PR #24214, opened 2022-02-16
vala.eclass: Support EAPI 8

ulm:
Function vala_src_prepare did not call eapply_user, so it could not be
used as a stand-alone phase function but must be called explicitly.
Rename it to vala_setup, which can be called either from pkg_setup or
from src_prepare. Add a trivial vala_src_prepare wrapper in existing
EAPIs, so that functionality there does not change.

Apparently, eutils and multilib eclasses are not used, therefore no
longer inherit them in EAPI 8.

Reviewed-by: Mart Raudsepp <leio@gentoo.org>
Signed-off-by: Ulrich Müller <ulm@gentoo.org>

### DIFF
--- a/eclass/vala.eclass
+++ b/eclass/vala.eclass
@@ -1,4 +1,4 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 # @ECLASS: vala.eclass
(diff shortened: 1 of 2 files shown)


eclass/vala.eclass
```diff
@@ -175,4 +182,11 @@ vala_src_prepare() {
 	export PKG_CONFIG_PATH="${T}/pkgconfig:${PKG_CONFIG_PATH}"
 }
 
+# @FUNCTION: vala_src_prepare
+# @DESCRIPTION:
+# Exported phase function in EAPIs 6 and 7. Calls vala_setup.
+if [[ ${EAPI} == [67] ]]; then
```
mgorny, line 188: Thank you for doing this. I will finally be able to remove inappropriate comments from my ebuilds!

eclass/vala.eclass
```diff
@@ -103,14 +105,17 @@ vala_depend() {
 # VALA_MAX_API_VERSION, VALA_MIN_API_VERSION, and VALA_USE_DEPEND.
 vala_best_api_version() {
 	local u v
+	local hv_opt="-b"
+	[[ ${EAPI} == 6 ]] && hv_opt=""
```
mgorny, line 109: How about making it an array, and avoiding word splitting (or whatever the relevant magic is called)?

```suggestion
	local hv_opt=( -b )
	[[ ${EAPI} == 6 ]] && hv_opt=()
```

eclass/vala.eclass
```diff
@@ -103,14 +105,17 @@ vala_depend() {
...
+
 	u=$(_vala_use_depend)
 
 	for v in $(vala_api_versions); do
-		has_version $([[ $EAPI == [1-6] ]] || echo -b) "dev-lang/vala:${v}${u}" && echo "${v}" && return
+		has_version ${hv_opt} "dev-lang/vala:${v}${u}" && echo "${v}" && return
```
mgorny, line 114: …and then:

```suggestion
		has_version "${hv_opt[@]}" "dev-lang/vala:${v}${u}" && echo "${v}" && return
```

eclass/vala.eclass
ulm: Seems rather over-engineered to me, given that this will be dropped once EAPI 6 is gone.

PR #24217, opened 2022-02-16
media-sound/pulseaudio-daemon: Update 15.0 split build to patches merged upstream

i-garrison:
Added patch moving systemd socket activation to server implementation.
No functional changes. This patch is only required to apply upstream
daemon/client split patch with minimal adjustment.
https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/640

Updated daemon/client split patch to version merged upstream, rebased to 15.0
https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/659


media-sound/pulseaudio-daemon/pulseaudio-daemon-15.0.ebuild
```diff
@@ -148,9 +148,11 @@ DOCS=( NEWS README )
 
 S="${WORKDIR}/${MY_P}"
```
thesamesam, line 149: My suggestion would be to revbump (`git mv` should be fine, I guess) just to ease debugging so we know everyone is in the same state, even if in theory it should all be the same.

i-garrison: Since both media-libs/libpulse and media-sound/pulseaudio-daemon have the same change, updated both to 15.0-r1

PR #24221, opened 2022-02-16
sci-libs/flann: fix a couple of bugs

Fat-Zer:


thesamesam:
Looks good. Only nit would be descriptive commit summaries for future if you can, thank you!

PR #24224, opened 2022-02-17
dev-java/log4j-core: Generate plugin listing file for consumers

Leo3418:
The new ebuild generates a serialized plugin listing file for the Log4j 2 built-in plugins, so these plugins can be loaded during the runtime, and Log4j 2 consumer programs can use them.

More information about Log4j 2 plugins: https://logging.apache.org/log4j/2.x/manual/plugins.html

### DIFF
--- a/dev-java/log4j-core/log4j-core-2.17.1-r1.ebuild
+++ b/dev-java/log4j-core/log4j-core-2.17.1-r1.ebuild
@@ -1,0 +1,171 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# Skeleton command:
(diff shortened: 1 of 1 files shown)


fordfrog:
@Leo3418 merged, thanks! in the future, when fixing a bug of an ebuild, you can do a revbump as you did and drop the old one (simply rename the ebuild to a revbump). otherwise we should not close the original bug, as there is still the affected ebuild in the tree. i dropped the old one so it's fine now.

Leo3418:
> @Leo3418 merged, thanks! in the future, when fixing a bug of an ebuild, you can do a revbump as you did and drop the old one (simply rename the ebuild to a revbump). otherwise we should not close the original bug, as there is still the affected ebuild in the tree. i dropped the old one so it's fine now.

Gotcha! I thought the older revision would still have been kept in the tree for some time. I'll clean up the old ebuild from now on for small improvements. Sorry for the hassle, and thanks for the advice!

fordfrog:
np at all. but it makes no sense to keep an old (buggy) version when we introduce a bugfix. the only exception would be if the old would be stable and the new one won't be stable. but bufixes usually go straight to the stable ebuild, as they fix a bug in the stable ebuild.

PR #24235, opened 2022-02-17
net-fs/nfs-utils: pull sqlite if nfsv4

tormath1:
from `./configure.ac`, it seems that this dependency is only required if
`nfsv4` compilation option is enabled.

Bug: https://bugs.gentoo.org/696250

<hr>

See also: http://git.linux-nfs.org/?p=steved/nfs-utils.git;a=blob;f=configure.ac;h=50e9b321dcf3d144e998b7d2318af9197c307a34;hb=HEAD#l359


mattst88:
Looks right to me.

mattst88 (approved):

PR #24236, opened 2022-02-17
dev-java/rome: EAPI 8, min java 1.8:*

vaukai:
Bug: https://bugs.gentoo.org/831010
Bug: https://bugs.gentoo.org/833410

### DIFF
--- a/dev-java/rome/rome-0.9-r4.ebuild
+++ b/dev-java/rome/rome-0.9-r4.ebuild
@@ -1,0 +1,53 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+JAVA_PKG_IUSE="doc source test"
+JAVA_TESTING_FRAMEWORKS="junit"
+
+inherit java-pkg-2 java-pkg-simple
+
+DESCRIPTION="Java framework for RSS and Atom feeds"
+HOMEPAGE="https://rometools.github.io/rome/"
+SRC_URI="mirror://gentoo/${P}.zip"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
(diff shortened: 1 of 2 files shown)


dev-java/rome/rome-0.9-r4.ebuild
```diff
@@ -0,0 +1,49 @@
...
+CP_DEPEND="dev-java/jdom:0"
+
+DEPEND="
+	${CP_DEPEND}
+	virtual/jdk:1.8
+	test? ( dev-java/ant-junit:0 )"
+
+RDEPEND="
+	${CP_DEPEND}
+	virtual/jre:1.8"
```
fordfrog, line 28: please comment in the ebuild why we need restriction to `:1.8`

fordfrog: thank you. this can be fixed by specifying which `Module` class to use. i did not check the code but my guess is that the correct one is `com.sun.syndication.feed.module.Module`. so simply replacing each `Module` in the source code with `com.sun.syndication.feed.module.Module` should fix it. so it would be `com.sun.syndication.feed.module.Module` and `com.sun.syndication.feed.module.Module.class` instead of just `Module` or `Module.class`.

Leo3418: @vaukai Some patches for packages with a similar situation for your reference:
https://github.com/6-6-6/spark-overlay/blob/master/dev-java/jackson-mapper-asl/files/jackson-mapper-asl-1.9.13-fix-for-java-9%2B-Module-class.patch
https://github.com/6-6-6/spark-overlay/blob/master/dev-java/yecht/files/yecht-1.1-fix-for-java-9%2B-Module-class.patch

vaukai: Thanks for advice and examples :-)

fordfrog:
@vaukai merged, thanks!

PR #24238, opened 2022-02-17
media-gfx/blender Fix live ebuild and add support for nanovdb.

DarkDefender:
@waebbl Do you have time to look into CUDA support for nanovdb?
I don't have and CUDA card to test and implement it myself.

I talked to Brecht (the main Cycles dev), the told me the only way to actually check if nanovdb support is working is to render a scene with volumes. If it works, the smoke should render, otherwise it should just be empty/corrupted.

You could be able to test this with the `tests/render/openvdb/openvdb_smoke.blend` in the automated test bundle.
The reason Blender uses nanovdb for rendering in cycles is that it should reduce memory usage by a bit.
So that is something you could look out for too.

### DIFF
--- a/media-gfx/blender/blender-9999.ebuild
+++ b/media-gfx/blender/blender-9999.ebuild
@@ -27,7 +27,7 @@ LICENSE="|| ( GPL-3 BL )"
 IUSE="+bullet +dds +fluid +openexr +system-python +system-numpy +tbb \
 	alembic collada +color-management cuda +cycles \
 	debug doc +embree +ffmpeg +fftw +gmp headless jack jemalloc jpeg2k \
-	man ndof nls openal +oidn +openimageio +openmp +opensubdiv \
+	man +nanovdb ndof nls openal +oidn +openimageio +openmp +opensubdiv \
 	+openvdb +osl +pdf +potrace +pugixml pulseaudio sdl +sndfile standalone test +tiff valgrind"
 RESTRICT="!test? ( test )"
 
@@ -90,7 +90,7 @@ RDEPEND="${PYTHON_DEPS}
 	)
 	opensubdiv? ( >=media-libs/opensubdiv-3.4.0[cuda=] )
 	openvdb? (
-		>=media-gfx/openvdb-8.2.0-r2:=
+		>=media-gfx/openvdb-9.0.0:=[nanovdb?]
 		dev-libs/c-blosc:=
 	)
 	osl? ( >=media-libs/osl-1.11.16.0-r3:= )
@@ -239,7 +239,7 @@ src_configure() {
 		-DWITH_MEM_VALGRIND=$(usex valgrind)
 		-DWITH_MOD_FLUID=$(usex fluid)
 		-DWITH_MOD_OCEANSIM=$(usex fftw)
-		-DWITH_NANOVDB=OFF
+		-DWITH_NANOVDB=$(usex nanovdb)
 		-DWITH_OPENAL=$(usex openal)
 		-DWITH_OPENCOLLADA=$(usex collada)
 		-DWITH_OPENCOLORIO=$(usex color-management)
--- a/media-gfx/blender/blender-9999.ebuild
+++ b/media-gfx/blender/blender-9999.ebuild
@@ -123,11 +123,6 @@ BDEPEND="
 	nls? ( sys-devel/gettext )
 "
 
-PATCHES=(
-	"${FILESDIR}"/${PN}-3.0.1-openexr.patch
-	"${FILESDIR}"/${PN}-3.0.1-openimageio-2.3.patch
-)
-
 blender_check_requirements() {
 	[[ ${MERGE_TYPE} != binary ]] && use openmp && tc-check-openmp
 
@@ -234,7 +229,6 @@ src_configure() {
 		-DWITH_HEADLESS=$(usex headless)
 		-DWITH_INSTALL_PORTABLE=OFF
 		-DWITH_IMAGE_DDS=$(usex dds)
-		-DOPENEXR_ROOT_DIR="${ESYSROOT}/usr/$(get_libdir)/OpenEXR-3"
 		-DWITH_IMAGE_OPENEXR=$(usex openexr)
 		-DWITH_IMAGE_OPENJPEG=$(usex jpeg2k)
 		-DWITH_IMAGE_TIFF=$(usex tiff)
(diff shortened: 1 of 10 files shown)


media-gfx/openvdb/openvdb-9.0.0-r2.ebuild
```diff
@@ -14,23 +14,23 @@ SRC_URI="https://github.com/AcademySoftwareFoundation/${PN}/archive/v${PV}.tar.g
...
-	^^ ( abi6-compat abi7-compat abi8-compat )
+	^^ ( abi6-compat abi7-compat abi8-compat abi9-compat )
 	python? ( ${PYTHON_REQUIRED_USE} )
 "
 RDEPEND="
 	>=dev-cpp/tbb-2021.4.0:=
 	dev-libs/boost:=
 	dev-libs/jemalloc:=
 	dev-libs/log4cplus:=
-	>=dev-libs/imath-3.1.4-r2:=
```
thesamesam, line 30: Keep this to ensure we're using an imath without the bad headers (Imath-3 vs Imath)

thesamesam (commented):
Nice to see you back! I hope things are better now!

media-gfx/openvdb/openvdb-9.0.0-r2.ebuild
DarkDefender: I must have changed and included these two by mistake. I will revert these two changes.

DarkDefender:
> Nice to see you back! I hope things are better now!

It is starting to get a bit better. But I think I might be a bit flaky for a quite some time still.
But I'll try to get in some overdue changes. 

Thanks again for taking care of so much of the maintenance!

media-gfx/blender/blender-9999.ebuild
```diff
@@ -90,7 +90,7 @@ RDEPEND="${PYTHON_DEPS}
 	)
 	opensubdiv? ( >=media-libs/opensubdiv-3.4.0[cuda=] )
 	openvdb? (
-		>=media-gfx/openvdb-8.2.0-r2:=
+		>=media-gfx/openvdb-8.2.0-r2:=[nanovdb?]
```
waebbl, line 93: Seeing that you only added the nanovdb USE flag to v9.0.0, shouldn't we restrict this to >=openvdb-9.x?

waebbl:
@DarkDefender I can test the nanovdb feature and if it's working with cuda.

DarkDefender:
> @DarkDefender I can test the nanovdb feature and if it's working with cuda.

Thanks!
Note that if we add CUDA support, we should probably add a `cuda` useflag as well as they have a switch for this in nanovdb:
https://academysoftwarefoundation.github.io/openvdb/NanoVDB_HowToBuild.html

Though it should now autodetect cuda if you have the libraries installed.
So you should be able to test without editing the ebuild.

waebbl:
> > @DarkDefender I can test the nanovdb feature and if it's working with cuda.
> 
> Thanks! Note that if we add CUDA support, we should probably add a `cuda` useflag as well as they have a switch for this in nanovdb: https://academysoftwarefoundation.github.io/openvdb/NanoVDB_HowToBuild.html
> 
> Though it should now autodetect cuda if you have the libraries installed. So you should be able to test without editing the ebuild.

Thanks for the hint. I'll take a look at the build system and see about cuda support. Noticed the nanovdb option already during a earlier patch of the package, but lacked the time to implement it.

media-gfx/blender/blender-9999.ebuild
```diff
@@ -263,6 +257,7 @@ src_configure() {
 		-DWITH_SDL=$(usex sdl)
 		-DWITH_STATIC_LIBS=OFF
 		-DWITH_SYSTEM_EIGEN3=ON
+		-DWITH_SYSTEM_FREETYPE=ON
```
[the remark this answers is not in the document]
DarkDefender: The variable doesn't say what it does -_-

There is no bundled freetype in blender.
The flag currently only makes the `brotli` library optional.
I have poked the upstream developer who introduced this change to clarify and hopefully change this.

media-gfx/blender/blender-9999.ebuild
DarkDefender: Yeah, I think I'll just be lazy with this and bump support to 9.x as you suggested.
Cleaner that way.

waebbl:
This patch to your PR builds fine and seems to work. At least the image looks identical to one created with openvdb-8 / blender-3.0 and the smoke is visible, so I assume the cuda functions from nanovdb are working.

[0001-media-gfx-openvdb-enable-CUDA-support-for-NanoVDB.patch.txt](https://github.com/gentoo/gentoo/files/8098770/0001-media-gfx-openvdb-enable-CUDA-support-for-NanoVDB.patch.txt)

DarkDefender:
> so I assume the cuda functions from nanovdb are working.
I assume you rendered the cycles scene with CUDA and not the CPU render?
Just double checking, I've made that mistake in the past when looking for GPU render issues.

Thanks for the patch! I've integrated it into this PR.

waebbl:
> > so I assume the cuda functions from nanovdb are working.
> 
> I assume you rendered the cycles scene with CUDA and not the CPU render? Just double checking, I've made that mistake in the past when looking for GPU render issues.
> 
> Thanks for the patch! I've integrated it into this PR.

Yes I did :) At first CPU was active, but then I wondered about the low memory footprint and checked the preferences.

BTW, a comparison on memory usage leads to only a minor reduction with nanovdb for this scene. openvdb-9 with nanovdb needs ~360M and openvdb-8 without nanovdb needs ~370M.

media-gfx/openvdb/openvdb-9.0.0-r3.ebuild
```diff
@@ -14,16 +14,18 @@ SRC_URI="https://github.com/AcademySoftwareFoundation/${PN}/archive/v${PV}.tar.g
...
-	numpy? ( python )
 	^^ ( abi6-compat abi7-compat abi8-compat abi9-compat )
+	cuda? ( nanovdb )
+	numpy? ( python )
 	python? ( ${PYTHON_REQUIRED_USE} )
 "
+
 RDEPEND="
-	>=dev-cpp/tbb-2021.4.0:=
+	>=dev-cpp/tbb-2020.3
```
thesamesam, line 28: dropped :=

waebbl: 2020.3 doesn't have subslots

thesamesam: It has an implicit 0/0 which then, when upgrading to a newer version, is a change (e.g. 0/2021 or whatever), so it matters if we know a newer version has them, as we want to get rebuilt when the ABI changes

waebbl: i see

PR #24241, opened 2022-02-17
app-i18n/transifex-client: add new version 1.0.1

robert7k:


app-i18n/transifex-client/transifex-client-1.0.1.ebuild
```diff
@@ -0,0 +1,230 @@
...
+	"gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b"
+	"gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b/go.mod"
+	)
+go-module_set_globals
+
+SRC_URI="https://github.com/transifex/cli/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz
+	${EGO_SUM_SRC_URI}"
+S="${WORKDIR}"/cli-${PV}
+
+LICENSE="Apache-2.0"
```
MatthewGentoo, line 217: While Transifex itself is Apache-2.0 licensed, the LICENSE variable needs to cover the licenses of the dependencies that we download too.

I used https://github.com/google/go-licenses to grab them all and update this. It might be worth using it after every update just in case any dependencies have been added or removed.

PR #24242, opened 2022-02-17
net-libs/webkit-gtk: add eme-free flag [please reassign]

bqv:


mattst88:
EME is basically DRM, like as required by Netflix?

bqv:
> EME is basically DRM, like as required by Netflix?

Correct :)

telans:
Perhaps the use flag should be called `eme-free`? It would match that of www-client/firefox

bqv:
> Perhaps the use flag should be called `eme-free`? It would match that of www-client/firefox

@telans done :)

leio:
So your goal is to enable experimental support for it, but it got changed to be named `eme-free` from something else (that I can't get a compare out of github from the force push anymore) based on earlier feedback?
As-is, this ends up enabling experimental support by default, and I don't think we want that. And `IUSE=+eme-free` would give a weird impression, I think

bqv:
> So your goal is to enable experimental support for it, but it got changed to be named `eme-free` from something else (that I can't get a compare out of github from the force push anymore) based on earlier feedback? As-is, this ends up enabling experimental support by default, and I don't think we want that. And `IUSE=+eme-free` would give a weird impression, I think.
> 
> Edit: Based on PR summary change I see it was named IUSE=eme before the force push

Yes, originally I had the `eme` USE, not enabled by default, which enabled EME support. Then changed to `eme-free` to match firefox, which I suppose implicitly made it enable EME by default. What would you suggest as a resolution?

leio:
To my knowledge, this doesn't bring any value to Gentoo, as it won't get you DRM working, sorry. There is no EME support in open source webkitgtk. More information on the closed bug.

PR #24250, opened 2022-02-18
net-misc/r8125: new package for Realtek r8125 driver

Karlson2k:
This package is vendor drivers for Realtek RTL8125 network cards.
While Linux kernel has support for these card, vendor driver provides more features, which are not available in in-tree driver.

Additional features includes:
* RSS and Multi TX Queue (makes sense for 2.5Gbps speeds)
* PTP
* Probably better stability under heavy load

Bug: https://bugs.gentoo.org/736400

Similar package for Arch: https://aur.archlinux.org/packages/r8125-dkms

Gentoo has similar package for older Realtek NICs: [net-misc/r8168](https://packages.gentoo.org/packages/net-misc/r8168)


net-misc/r8125/r8125-9.007.01.ebuild
```diff
@@ -0,0 +1,38 @@
+# Copyright 1999-2021 Gentoo Authors
```
hydrapolic, line 1: 2022

Karlson2k: Thanks, fixed.

PR #24258, opened 2022-02-18
sys-process/iotop: delayacct issue

samuelbernardo:
Add post install instructions to add delayacct into kernel parameters.

Closes: #833617

### DIFF
--- a/sys-process/iotop/iotop-0.6.ebuild
+++ b/sys-process/iotop/iotop-0.6.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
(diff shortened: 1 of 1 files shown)


samuelbernardo:
> The format is: `Closes: https://bugs.gentoo.org/num`.

Just corrected commit message. Hope everything ok now.

thesamesam:
Almost -- no need for both Bug and Closes. Closes is stronger than Bug, so just do that :)

sys-process/iotop/iotop-0.6.ebuild
```diff
@@ -33,3 +33,15 @@ PATCHES=(
 pkg_setup() {
 	linux-info_pkg_setup
 }
+
+pkg_postinst() {
+	cat << EOF | elog
+To enable the SWAPIN/IO columns Edit /etc/default/grub,
+adding "delayacct" as an option to the GRUB_CMDLINE_LINUX_DEFAULT
```
Flowdalic, line 40: grub is only one way to boot the Linux kernel. I think the first message should be "To enable the SWAPIN/IO columns, set the 'delayacct' kernel command-line parameter". Then you can go on with: "For example if using grub, set or append GRUB_CMDLINE_LINUX_DEFAULT with 'delayacct' in /etc/default/grub". Maybe a reference to the kernel's documentation on delay accounting can't hurt too: https://www.kernel.org/doc/html/latest/accounting/delay-accounting.html

samuelbernardo: @Flowdalic Thanks for the review... I'm going update the PR with your suggestions.

samuelbernardo: Done.
I think thjis should be ready.

Flowdalic (changes requested):
Looks good, but please fix the QA issues (trailing whitespace).

PR #24261, opened 2022-02-18
Migrate more Java packages to dev-java/log4j-12-api

Leo3418:
A patch for `net-p2p/vuze` was added to make it compatible with Log4j 2, which could save it from a last rite otherwise (https://github.com/gentoo/gentoo/pull/24245).  As a side effect, more memory is required by the JVM running `javac` during build time.  An extra amount of 192 MB memory was needed on my `amd64` test machines, and I assumed the same amount for other architectures.

`app-text/jabref` and `net-p2p/vuze` ebuilds were also reworked for EAPI 8, compliance with XDG conventions, and so on.

Please let me know if the change to `dev-java/ant-core` is trivial enough to warrant an immediate clean-up of `-r2`.

### DIFF
--- a/app-text/jabref/jabref-2.10-r7.ebuild
+++ b/app-text/jabref/jabref-2.10-r7.ebuild
@@ -1,0 +1,132 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+JAVA_PKG_IUSE="doc test"
+
+inherit desktop java-pkg-2 java-ant-2 virtualx xdg-utils
+
+MY_PV="${PV/_beta/b}"
+
+DESCRIPTION="Java GUI for managing BibTeX and other bibliographies"
+HOMEPAGE="https://www.jabref.org/"
+SRC_URI="mirror://sourceforge/${PN}/JabRef-${MY_PV}-src.tar.bz2"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+
+CP_DEPEND="
+	dev-java/antlr:0
+	dev-java/antlr:3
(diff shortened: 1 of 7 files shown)


app-text/jabref/jabref-2.10-r7.ebuild
```diff
@@ -0,0 +1,86 @@
...
+	dev-java/spin:0
+	dev-java/microba:0
+	>=dev-java/glazedlists-1.8.0:0"
+
+DEPEND="
+	virtual/jdk:1.8
+	${CP_DEPEND}"
+
+RDEPEND="
+	virtual/jre:1.8
```
fordfrog, line 40: please add a comment why we need to restrict to `:1.8`

fordfrog:
please squash the commits for each ebuild into one.

fordfrog:
`dev-java/ant-core` merged, thanks! i dropped `-r2`, but just because this is a security revbump and we want to get stabilized this one and get rid of the old ones. otherwise i'd keep the old one as it would be soon eligible for stabilization.

app-text/jabref/jabref-2.10-r7.ebuild
Leo3418: done!

fordfrog:
`app-text/jabref` merged, thanks! i just dropped that >1.8 note from the commit message as we stick to `:1.8`.

fordfrog:
`net-p2p/vuze` merged too, thanks!

Leo3418:
> `app-text/jabref` merged, thanks! i just dropped that >1.8 note from the commit message as we stick to `:1.8`.

That note actually just refers to JDK in `DEPEND`, as the JDK atom in `DEPEND` is the `>=virtual/jdk-1.8:*` wildcard. Only the JRE version in `RDEPEND` is limited to `1.8`. But since the commit is already merged, let's keep it as-is.

PR #24264, opened 2022-02-18
profiles: mark sys-libs/e2fsprogs-libs deprecated

thesamesam:
Bug: https://bugs.gentoo.org/806875


thesamesam:
Done since.

PR #24266, opened 2022-02-19
dev-python/pyxattr: keyword ~riscv. 

ahmedcharles:


dlan17:
please submit this as two separated PR
1) for pyxattr, please make sure you have tested with "test" enabled
   emerge -v1o --with-test-deps dev-python/pyxattr
   FEATURES=test emerge -v1 dev-python/pyxattr
2) for musl, we'd accept once we have musl setup

ahmedcharles:
I tested this, as above.

dlan17:
we'd like to keep consistency..
so, no need to add "full-stop" at end of the commit message, I will fix this
but please keep mind this next time

PR #24269, opened 2022-02-19
app-crypt/tpm2-tss: Bump to 3.2.0, remove old

salahcoronya:
The distribution tarball app-crypt/tpm2-tss-3.2.0 was generated incorrect by upstream, see https://github.com/tpm2-software/tpm2-tss/issues/231, so the libtpms TCTI  is disabled for this version to work around it.

This is the first release that should actually work with openssl-3.0 (3.1.0 compiles, but TPM2_CreatePrimary doesn't work), which will be needed for tpm2-openssl (not in portage). 

Test require at least curl-7.80.0 (currently not marked stable). 

Integration tests are now enabled. This requires app-crypt/swtpm, which is not keyworded on riscv, so i dropped it for now, A keywording bug can file afterwards.

Also clean up an old version tpm2-tss.


app-crypt/tpm2-tss/tpm2-tss-3.2.0.ebuild
```diff
@@ -11,30 +11,32 @@ SRC_URI="https://github.com/tpm2-software/${PN}/releases/download/${PV}/${P}.tar
...
 
 RDEPEND="acct-group/tss
 	acct-user/tss
-	fapi? (
-		dev-libs/json-c
-		net-misc/curl
-	)
-	gcrypt? ( dev-libs/libgcrypt:0= )
-	openssl? ( dev-libs/openssl:0= )"
+	fapi? ( dev-libs/json-c
```
thesamesam, line 24: Not subscribing to json-c subslot? Do we not link against it?

app-crypt/tpm2-tss/tpm2-tss-3.2.0.ebuild
```diff
@@ -11,30 +11,32 @@ SRC_URI="https://github.com/tpm2-software/${PN}/releases/download/${PV}/${P}.tar
 
 LICENSE="BSD-2"
 SLOT="0"
-KEYWORDS="amd64 arm arm64 ppc64 x86"
-IUSE="doc +fapi gcrypt +openssl static-libs test"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
+IUSE="doc +fapi  +openssl mbedtls static-libs test"
```
thesamesam, line 15: whitespace

app-crypt/tpm2-tss/tpm2-tss-3.2.0.ebuild
salahcoronya: I did not know it had a subslot. Fixed.

app-crypt/tpm2-tss/tpm2-tss-3.2.0.ebuild
salahcoronya: fixed

PR #24273, opened 2022-02-19
gui-apps/grim-1.4.0: install bash completions with eclass

parona-source:
* Disable bash-completions switch and instead install bash completions
  with bash-completion-r1 eclass.
* This serves to avoid the build system requiring
  app-shells/bash-completion or having to go out of your way to remove
  this requirement from build system itself.

The bash-completions switch was enabled yesterday here https://github.com/gentoo/gentoo/pull/24182 which then introduced this issue https://bugs.gentoo.org/833595.

### DIFF
--- a/gui-apps/grim/grim-1.4.0-r1.ebuild
+++ b/gui-apps/grim/grim-1.4.0-r1.ebuild
@@ -1,9 +1,9 @@
 # Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
 
-inherit meson
+inherit bash-completion-r1 meson
 
 DESCRIPTION="Grab images from a Wayland compositor"
 HOMEPAGE="https://github.com/emersion/grim"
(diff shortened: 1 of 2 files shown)


MatthewGentoo (commented):
Since the installed files are changing, you should perform a rev bump so that users rebuild the package.

(also, does the -9999 ebuild need touching too?)

ArsenArsen:
besides the above, check whether the fish completions behave well

thanks for getting to this before I did btw <3

parona-source:
@MatthewGentoo Modified commit to so that the changes are in 1.4.0-r1 instead also updated 9999 with these changes. Updated both of them eapi 8 as well as while I was at it.

parona-source:
@ArsenArsen the fish completions dont require fish like the bash completions required bash-completion. You can check it here https://github.com/emersion/grim/blob/v1.4.0/contrib/completions/meson.build

ArsenArsen:
it still tries to query it. it may be best to force it to use the fallback path to ensure consistency

parona-source:
@ArsenArsen changed it so that it installs the fish completions with doins. Also double checked the dependencies and updated them.

thesamesam:
Thanks both!

PR #24274, opened 2022-02-19
net-proxy/squid: add sys-libs/tdb, remove sys-libs/db

hydrapolic:
http://www.squid-cache.org/Versions/v5/squid-5.4.1-RELEASENOTES.html#ss2.4

"This release deprecates use of BerkleyDB in favour of TrivialDB."

### DIFF
--- a/net-proxy/squid/squid-5.4.1-r1.ebuild
+++ b/net-proxy/squid/squid-5.4.1-r1.ebuild
@@ -52,8 +52,8 @@ COMMON_DEPEND="acct-group/squid
 	esi? ( dev-libs/expat dev-libs/libxml2 )
 	gnutls? ( >=net-libs/gnutls-3.1.5:= )
 	logrotate? ( app-admin/logrotate )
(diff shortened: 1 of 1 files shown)


net-proxy/squid/squid-5.4.1.ebuild
```diff
@@ -52,8 +52,8 @@ COMMON_DEPEND="acct-group/squid
 	esi? ( dev-libs/expat dev-libs/libxml2 )
 	gnutls? ( >=net-libs/gnutls-3.1.5:= )
 	logrotate? ( app-admin/logrotate )
-	>=sys-libs/db-4:*
```
thesamesam, line 55: Revbump as it'll be automagic / will have worked if people had the right stuff installed accidentally.

(just git mv)

hydrapolic: Ok, pushed.

hydrapolic:
Thank you @thesamesam

PR #24276, opened 2022-02-19
app-emulation/xen: add 4.15.2

hydrapolic:
Compile tested on ~amd64.


hydrapolic:
Thank you @thesamesam

PR #24280, opened 2022-02-20
net-fs/samba: Make libunwind optional, it's only for debug and backtraces

inode64:
Make libunwind optional for net-fs/samba, it's only for debug and backtraces, You see source code in lib/util/fault.c and source3/utils/smbcontrol.c

I have analyzed other distributions such as they compile samba to verify that it is not necessary, I attach links:

* https://www.linuxfromscratch.org/blfs/view/10.0-systemd/basicnet/samba.html
* https://packages.debian.org/bullseye/samba
* https://archlinux.org/packages/extra/x86_64/samba/

I compile without libunwind, it compiles and run without problems.

Package-Manager: Portage-3.0.30-r1, Repoman-3.0.3-r1
Signed-off-by: Fco Javier Felix <web@inode64.com>


thesamesam:
This is automagic with your change: unwind will be used if installed even if the flag is off, which is problematic for e.g binary packages. Please pass an option to configure or similar to respect the flag.

inode64:
Yes, I already saw it, but there is no flag in the configuration process, one will have to be made for the lib/util/wscript_configure file

thesamesam:
Could you try to add one and/or report an issue upstream requesting one, given e.g. the portability issues with requiring unwind? Thanks!

inode64:
I have tried to create a request to send it to the samba team and I am not clear :-(

thesamesam:
> I have tried to create a request to send it to the samba team and I am not clear :-(

Thanks for working on it! I think you need an account on their gitlab.

inode64:
I make the request to samba team 
https://gitlab.com/samba-team/samba/-/merge_requests/2401

thesamesam:
Could you rebase this with the version sent upstream, and include links to the upstream PR/commits? It should be good to go then. Thanks a lot!

SoapGentoo (changes requested):

SoapGentoo:
@inode64 one last rebase, then I will merge it, and only do it for the latest ebuild

a17r:
ping

inode64:
pong
I'm on a trip

inode64:
I'm sorry I had to create another commit, I don't know why I destroyed this one, (I don't master the world of git) 
https://github.com/gentoo/gentoo/pull/27422

PR #24281, opened 2022-02-20
net-analyzer/nagios-check_rbl: bump 1.6.3

antonfischl1980:
also bumped EAPI=8
updated missing dep according to Makefile.PL ( virtual/perl-Socket )

Signed-off-by: Anton Fischl <github@fischl-online.de>


orlitzky:
Thanks! I had to edit your commit message a bit to conform to out length limit (~75 chars), but otherwise LGTM out of the box. The perl/Socket dependency is new so thanks for catching that.

antonfischl1980:
I keep the commit message length in mind for the future.

thank you for the comment in the ebuild, otherwise I wouldn't have "git blam"ed upstreams Makefile.PL

PR #24292, opened 2022-02-20
dev-java/swingx-ws: Use EAPI 8, set minimum Java version to 1.8

Leo3418:
Fix build failure with JDK 11+, which has not been reported yet.


fordfrog:
@Leo3418 merged, thanks!

PR #24294, opened 2022-02-21
net-misc/unison: bump to version 2.51.5 

kernlpanic:
Closes: https://bugs.gentoo.org/818247
Signed-off-by: Dennis Eisele <kernlpanic@dennis-eisele.de>


net-misc/unison/unison-2.51.5.ebuild
```diff
@@ -0,0 +1,105 @@
...
+
+LICENSE="GPL-2"
+SLOT="$(ver_cut 1-2)"
+KEYWORDS="~amd64 ~arm ~ppc ~x86 ~amd64-linux ~x86-linux ~ppc-macos ~sparc-solaris"
+IUSE="debug doc gtk +ocamlopt threads"
+
+# ocaml version so we are sure it has ocamlopt use flag
+BDEPEND="dev-lang/ocaml:=[ocamlopt?]
+	doc? ( app-text/dvipsk
+			app-text/ghostscript-gpl
```
Flowdalic, line 20: The indentation here is changed, compared with the previous version's ebuild. Please rectify that.

Flowdalic (changes requested):
IMHO you don't need to put the EAPI bump in an extra commit.

net-misc/unison/unison-2.51.5.ebuild
kernlpanic: Thanks for the feedback. I fixed that.

PR #24299, opened 2022-02-21
sys-apps/firejail: remove unused patch

mm1ke:


Flowdalic:
This looks valid. But since the history shows that this patch file was never used, I wonder if it was simply forgotten to be added to the ebuild. @hlein @thesamesam, since you did f83326db36c6215b3fb69cf9630c5e3b53d32c43 where this file was added: can it be removed or should it be added tot he ebuild?

hlein:
> This looks valid. But since the history shows that this patch file was never used, I wonder if it was simply forgotten to be added to the ebuild.

Aha, yup, I'm promoting that patch from my local test overlay to the main tree... except I forgot to include the bit that applies it in the ebuild. Will fix now...

PR #24300, opened 2022-02-21
app-emulation/libvirt: Fix build with >=dev-libs/libxslt-1.1.35

zippy2:
Libvirt failed to build with new libxslt because of a bug in one
of libvirt's xsl files. Backport the fix from upstream repo.

Closes: https://bugs.gentoo.org/833586


juippis:
Are there any runtime issues or is this purely a build-time issue? We probably shouldn't revision bump the ebuilds for build-time fix.

zippy2:
> Are there any runtime issues or is this purely a build-time issue? We probably shouldn't revision bump the ebuilds for build-time fix.

Yeah, it's build-time only. So in that case the revision is not bumped?

PR #24303, opened 2022-02-21
sci-libs/pdal: fix tests

t0b3:
Closes: https://bugs.gentoo.org/833820


sci-libs/pdal/files/pdal-2.3.0-fix_tests_for_proj811.patch
```diff
@@ -0,0 +1,109 @@
+From 1a7415f67cfe13f816345ce4b7fc1c68f0a45c83 Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: Nit: include links to relevant Gentoo bugs & upstream commits for patches

PR #24307, opened 2022-02-22
dev-util/poke: bump to version 2.1

kernlpanic:
Closes: https://bugs.gentoo.org/828835
Signed-off-by: Dennis Eisele <kernlpanic@dennis-eisele.de>


dev-util/poke/poke-1.4.ebuild
```diff
@@ -0,0 +1,51 @@
...
+	virtual/pkgconfig
+"
+
+src_configure() {
+	econf \
+		--disable-gui \
+		--disable-libnbd \
+		$(use_enable machine-interface mi) \
+		$(use_enable nls) \
+		$(use_enable static-libs static)
```
thesamesam, line 42: I suspect we don't actually want/need static-libs for this at all.

thesamesam: (If you drop the line and simplify src_install, it should be fine w/ EAPI 8 which passes --disable-static)

mid-kid:
The current version of poke is 2.1. I don't think adding an ebuild for 1.4 isn't very useful at this point.

kernlpanic:
> The current version of poke is 2.1. I don't think adding an ebuild for 1.4 is very useful at this point.

Thanks for the feedback you are right. I changed the new version to 2.1

dev-util/poke/poke-1.4.ebuild
kernlpanic: Thanks for the feedback. I fixed that

thesamesam:
Thanks! See 4a590d047fee354e2ca2b7e5a521ee8c9bb170e4 for some style changes.

kernlpanic:
Thank you for merging and for applying the style changes :) I'll take a better look next time before submitting

PR #24309, opened 2022-02-22
dev-libs/cereal: disable tests requiring -m32 for riscv and arm{,64}

alexfanqi:
the portability tests require -m32 flag, disable them on arch without this option.

Closes: https://bugs.gentoo.org/833866
Signed-off-by: Alex Fan <alex.fan.q@gmail.com>


thesamesam:
The option only exists for Apple? 
https://github.com/USCiLab/cereal/blob/master/CMakeLists.txt#L10

alexfanqi:
> The option only exists for Apple? https://github.com/USCiLab/cereal/blob/master/CMakeLists.txt#L10

I was also unsure prevoously. But after some research,  option is equivalent to a boolean cache variable in cmake and variable defined in commandline is by default cached. I guess should be the same🤔

I think the condition on mac is meant to turn it on by default.

thesamesam:
> > The option only exists for Apple? https://github.com/USCiLab/cereal/blob/master/CMakeLists.txt#L10
> 
> I was also unsure previously. But after some research, option is equivalent to a boolean cache variable in cmake and variable defined in commandline is by default cached. I guess should be the same🤔
> 
> I think the condition on mac is intended to turn it on by default.

Are you sure? At the very least, we get an unused CMake QA warning if using it. I had to remove it yesterday for that reason

alexfanqi:
> > > The option only exists for Apple? https://github.com/USCiLab/cereal/blob/master/CMakeLists.txt#L10
> > 
> > 
> > I was also unsure previously. But after some research, option is equivalent to a boolean cache variable in cmake and variable defined in commandline is by default cached. I guess should be the same🤔
> > I think the condition on mac is intended to turn it on by default.
> 
> Are you sure? At the very least, we get an unused CMake QA warning if using it. I had to remove it yesterday for that reason (can't easily link to commit on mobile but you can find it)
> 
> I've never dug into whether conditional options are really always defined but I kind of doubt it (it'd be weird)

unused variable is perhaps because test is not enabled? That variable is only used when tests are built in cmake. Does it make sense to add a condition on USE=test for appending this flag.

dev-libs/cereal/cereal-1.3.1-r1.ebuild
```diff
@@ -40,6 +40,11 @@ src_configure() {
 
 		-DWITH_WERROR=OFF
 	)
+	# bug 833866, disable tests requiring -m32
+	# condition on test to avoid unused variable warning
```
xen0n, line 44:
```suggestion
	# conditional on USE=test to avoid unused variable warning
```

alexfanqi:
> As discussed on IRC, I'm a bit concerned about what happens if they change the actual usage to be if(APPLE) guarded and I think I'd rather an upstreamable patch than relying on this as a hack.

That makes sense. I am asking upstream to change it

thesamesam:
any news?

PR #24311, opened 2022-02-22
subversion.eclass: Drop support for EAPIs 4 and 5, add EAPI 8

ulm:


hattya (approved):
I think that `ESVN_BOOTSTRAP` and `ESVN_PATCHES` can also be removed.

ulm:
> I think that `ESVN_BOOTSTRAP` and `ESVN_PATCHES` can also be removed.

Right, thanks. I had missed this.

PR #24312, opened 2022-02-22
dev-java/ant-eclipse-ecj: Drop unneeded slots 4.10 and 4.13 | dev-java/eclipse-ecj: Drop unneeded slots 4.10 and 4.13

vaukai:


fordfrog:
i think we should first check whether the depending apps can be updated to the newer versions of ecj.

vaukai:
> i think we should first check whether the depending apps can be updated to the newer versions of ecj.

According to my analysis there was nothing dependent on those 2 slots.  Did I miss something?

fordfrog:
no, i just meant we would still keep that old 4.5 and remove newer versions though it might be that the apps depending on 4.5 might be updated to use newer ecj instead.

vaukai:
> no, i just meant we would still keep that old 4.5 and remove newer versions though it might be that the apps depending on 4.5 might be updated to use newer ecj instead.

Got it.  But as long things work I see no reason to touch them.  So we better leave it as is atm and try upgrading once reverse deps have issues.  Closing this PR

PR #24324, opened 2022-02-22
app-crypt/tpm2-tss-3.2.0: Fix version information for system without git, remove broken old

salahcoronya:
Upstream changed the way it produced version numbers in the file - it now uses a git command instead of a hardcoded number in configure.ac. Thus systems without git got pkgconfig files with broken versions, and nothing would compile against it. Many of the other tpm2 projects are doing the same thing, but the fix is simple: substitute the git command with ${PV}.

Verified tpm2-abrmd and app-crypt/tpm2-tools compile with the patch.


app-crypt/tpm2-tss/tpm2-tss-3.2.0-r1.ebuild
```diff
@@ -49,6 +49,9 @@ pkg_setup() {
 
 src_prepare() {
 	default
+	sed -i \
```
thesamesam, line 52: Is this because we're not using release tarballs, or is it because we're calling eautoreconf?

Upstream should consider using the gnulib version of this which is the only method I'm aware of that does this correctly.

thesamesam: In any case, put a comment in explaining it & refer to the bug(s).

salahcoronya: its because of eautoreconf

thesamesam: can you add the comment?

thesamesam: also, can I suggest putting a check in src_install to read the .pc file, and compare it to PV, and if it doesn't match, die, to force manual investigation?

thesamesam: I'll do this so we get the fix in ASAP.

salahcoronya: I'm working on it now, i created the patch during my break. The comment is no problem:
Calling eautoreconf causes configure.ac to get the version information
via git. Without git, pkgconfig file will be broken. So substitute
${PV} for [git describe --tags --always --dirty] so git isn't required`

Checking the libtool file to make sure their valid is harder. I'm working working on some construction like this in src_install:
       $(tc-getPKG_CONFIG) --with-path="${D}/usr/$(get_libdir)/pkgconfig" \
        --exists "tss2-sys == ${PV}" \
        || die "Invalid pkg-config created. Check configure.ac"
But it doesn't quite work

thesamesam: I've done this (not committed), what do you think?

```
diff --git a/app-crypt/tpm2-tss/tpm2-tss-3.2.0-r1.ebuild b/app-crypt/tpm2-tss/tpm2-tss-3.2.0-r1.ebuild
index 4f8e544d9a79..2282dd36c271 100644
--- a/app-crypt/tpm2-tss/tpm2-tss-3.2.0-r1.ebuild
+++ b/app-crypt/tpm2-tss/tpm2-tss-3.2.0-r1.ebuild
@@ -49,9 +49,12 @@ pkg_setup() {

 src_prepare() {
 	default
+
+	# See bug #833887 (and similar); eautoreconf means .pc file gets wrong version.
 	sed -i \
 	"s/m4_esyscmd_s(\[git describe --tags --always --dirty\])/${PV}/" \
 		"configure.ac" || die
+
 	eautoreconf
 }

@@ -77,6 +80,12 @@ src_configure() {

 src_install() {
 	default
+
+	if [[ ${PV} != $(sed -n -e 's/^Version: //p' "${ED}/usr/$(get_libdir)/pkgconfig/tss2-sys.pc") ]] ; then
+		# Safeguard for bug #833887
+		die "pkg-config file version doesn't match ${PV}! Please report a bug!"
+	fi
+
 	find "${D}" -name '*.la' -delete || die
 }
```

salahcoronya: Looks about right except the ebuild doesn't inherit multilib , I think that's needed for get_libdir .

thesamesam: Not needed in [EAPI 6+](https://blogs.gentoo.org/mgorny/2015/11/13/the-ultimate-guide-to-eapi-6/). I'll commit my version to get it in but we can adjust it later if you want at all, of course.

I'm going to see if we can do this more generally as a QA check as we had another example of this just today (liburing)!

ionenwks: `get_libdir` is always available and part of [PMS](https://projects.gentoo.org/pms/8/pms.html), it's rather rare you need to inherit multilib directly (most cases are for building a x86-only package on amd64 with abi_x86_32 libs).

salahcoronya: Ok great, lets do it. A general QA check for this would be wonderful, since this isn't the first time this has bit me, and the all the tpm2 have migrated to this.

thesamesam:
Good work, thanks for responding so quickly to the bugs -- I'll see if I can slap something together and add it to https://github.com/gentoo/portage/pull/764 and we'll go from there.

PR #24325, opened 2022-02-22
net-misc/balance: add net_bind_service capability

glorg:
This changes binary capabilities and allows binding to privileged ports for non-root

### DIFF
--- a/net-misc/balance/balance-3.57-r1.ebuild
+++ b/net-misc/balance/balance-3.57-r1.ebuild
@@ -1,0 +1,43 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit fcaps toolchain-funcs
+
+DESCRIPTION="TCP Load Balancing Port Forwarder"
+HOMEPAGE="http://www.inlab.de/balance.html"
+SRC_URI="http://www.inlab.de/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc ~x86"
+IUSE=""
+
+DEPEND=""
+RDEPEND=""
+
+PATCHES=( "${FILESDIR}"/${P}-Makefile.patch )
+
+FILECAPS=(
+	'CAP_NET_BIND_SERVICE' '/usr/sbin/balance'
+)
+
+src_prepare() {
+	default
+
+	tc-export CC
+}
+
+src_install() {
+	default
+
+	#autocreated on program start, if missing
+	rm -rv "${ED}/var/run" || die
+}
+
+pkg_postinst() {
+	fcaps_pkg_postinst
+	elog "To run as non-root, be sure to have rendezvous directory created"
+	elog "with either 'mkdir -m 01777 /var/run/balance' or using tmpfiles."
+}


net-misc/balance/balance-3.57-r1.ebuild
```diff
@@ -0,0 +1,40 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
```
arkamar, line 4: Use newer `EAPI` (like 8), version 6 is deprecated https://wiki.gentoo.org/wiki/Project:Package_Manager_Specification#Council_approval_and_use_in_Gentoo_repository

net-misc/balance/balance-3.57-r1.ebuild
```diff
@@ -0,0 +1,40 @@
...
+
+inherit toolchain-funcs fcaps
```
arkamar, line 6: Sort inherits alphabetically

net-misc/balance/balance-3.57-r1.ebuild
```diff
@@ -0,0 +1,40 @@
...
+DESCRIPTION="TCP Load Balancing Port Forwarder"
+HOMEPAGE="http://www.inlab.de/balance.html"
+SRC_URI="http://www.inlab.de/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc ~x86"
+IUSE="caps"
+
+DEPEND="caps? ( sys-libs/libcap )"
```
arkamar, line 17: `sys-libs/libcap` is not needed here, `balance` does not use anything from this library and `fcaps` inherit sets its dependencies automatically.

net-misc/balance/balance-3.57-r1.ebuild
```diff
@@ -0,0 +1,40 @@
...
+	default
+
+	#autocreated on program start, if missing
+	rmdir "${D}"/var/run/${PN}
+}
+
+pkg_postinst() {
+	if use caps ; then
+		fcaps 'CAP_NET_BIND_SERVICE=+eip' "/usr/sbin/balance"
+		/bin/mkdir -m 01777 /var/run/balance/
```
arkamar, line 38: This line should not be here. `/var/run` is usually symlink to `/run` which is mounted as `tmpfs`, therefore `/var/run/balance` will not survive restart anyway. Btw, `balance` will probably create it on start, according to the comment in `src_install`.

net-misc/balance/balance-3.57-r1.ebuild
```diff
@@ -0,0 +1,40 @@
...
+inherit toolchain-funcs fcaps
+
+DESCRIPTION="TCP Load Balancing Port Forwarder"
+HOMEPAGE="http://www.inlab.de/balance.html"
+SRC_URI="http://www.inlab.de/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc ~x86"
+IUSE="caps"
```
arkamar, line 15: `caps` USE is meant for binaries that can conditionally use `sys-libs/libcap`. Here, `filecaps` would be more appropriate. However, I personally think it is not necessary to apply them conditional in this case.

arkamar (changes requested):

net-misc/balance/balance-3.57-r1.ebuild
glorg: balance creates in on start. Permissions might be resolbed by start-up script that is not provided for this tool, so it's up to end user.

net-misc/balance/balance-3.57-r1.ebuild
glorg: Indeed, and even extra flag is nor really required, removing

net-misc/balance/balance-3.57-r1.ebuild
```diff
@@ -0,0 +1,43 @@
...
+	default
+
+	#autocreated on program start, if missing
+	rm -rv "${ED}/var/run" || die
+}
+
+pkg_postinst() {
+	fcaps_pkg_postinst
+	elog "To run as non-root, be sure to have rendezvous directory created"
+	elog "with either 'mkdir -m 01777 /var/run/balance' or using tmpfiles."
```
arkamar, line 42: I personally don't think this message is necessary here, because `balance` itself prints following message on start:
```
ERROR: rendezvous directory not available and/or creatable
       please create /var/run/balance/ with mode 01777 like this: 
       # mkdir -m 01777 /var/run/balance/
```
But it is probably ok to go with it.

arkamar:
I think it would be better to change title of the commit message to somehting like `net-misc/balance: add capability`. The message is meant for devs and they see it is related to 3.57-r1 revision in diff. Also append `[please reassign]` to the pull request title in order to remove `no signoff` label.

net-misc/balance/balance-3.57-r1.ebuild
glorg: In case `/var/run` is on tmpfs and cleaned with reboots, this will point the user on what tool (tmpfiles) to use for solving this problem correctly and what's required for the server to run.

arkamar (approved):

PR #24334, opened 2022-02-24
app-vim/gentoo-syntax: remove MY_P from v1

arkamar:
Recent "downgrade" from `20211208` to `0.20211208` made me look to the git history why it happened and I noticed that `MY_P` is not needed anymore in version `1`.


mgorny:
Good catch, thanks!

PR #24340, opened 2022-02-24
x11-wm/pekwm: version bump 0.2.1

krejvl:
A new ebuild for pekwm version 0.2.1. Pekwm moved to Cmake since version
0.2.0, so the older ebuilds based on autotools can't be used and this
one was created from scratch.

Changes:

- Themes are handled by pekwm_theme app (requires git) and not need to be
  installed by ebuild.
- Significant change to pekwm menu config (Exec vs. ShellExec and '&'
  usage) that causes some apps doesn't start (e.g. thunderbird-bin).

Signed-off-by: Vlastimil Krejčíř <vlastimil.krejcir@gmail.com>


x11-wm/pekwm/pekwm-0.2.1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+
+multilib_src_configure() {
+	local mycmakeargs=(
+		-DENABLE_XSHAPE="yes"
+		-DENABLE_XRANDR="yes"
+		-DENABLE_JPEG=$(usex jpeg)
+		-DENABLE_PNG=$(usex png)
+		-DENABLE_XINERAMA=$(usex xinerama)
+		-DENABLE_XFT=$(usex truetype)
+		-DENABLE_XPM=$(usex xpm)
```
juippis, line 44: Some of these seem to be undefined, 
```
* One or more CMake variables were not used by the project:
 *   ENABLE_JPEG
 *   ENABLE_PNG
 *   ENABLE_XPM
 *   ENABLE_XRANDR
 *   ENABLE_XSHAPE
```

x11-wm/pekwm/pekwm-0.2.1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+IUSE="contrib debug +jpeg +png themes truetype +xft xinerama +xpm"
+
+RDEPEND="
+	png? ( media-libs/libpng:0 )
+	jpeg? ( virtual/jpeg:0 )
+	xpm? ( x11-libs/libXpm )
+	x11-libs/libXrender
+	truetype? ( x11-libs/libXft )
+	xinerama? ( x11-libs/libXinerama )
+	themes? ( dev-vcs/git )"
```
juippis, line 31: This seems a bit weird. Also this USE flag has really no other function in the ebuild except a postinst message. I see there's "pekwm-themes" in the Manifest file, so maybe you can always or conditionally download the theme package and always or conditionally install the themes via this ebuild? 
https://gitweb.gentoo.org/repo/gentoo.git/tree/games-arcade/lbreakout2/lbreakout2-2.6.5-r2.ebuild I guess here's one example of how to do it, but by default src_unpack should also deal with it.

x11-wm/pekwm/pekwm-0.2.1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+
+DESCRIPTION="A lightweight window manager initially based on aewm++"
+HOMEPAGE="
+	https://www.pekwm.se/
+	https://github.com/pekdon/pekwm
+"
+SRC_URI="
+	https://github.com/pekdon/${PN}/releases/download/release-${PV}/pekwm-${PV}.tar.gz
+"
+S="${WORKDIR}"/${PN}\-${PV}
```
juippis, line 17:
```suggestion
```

x11-wm/pekwm/pekwm-0.2.1.ebuild
```diff
@@ -0,0 +1,87 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+CMAKE_ECLASS=cmake
```
juippis, line 6: This is the default with >=EAPI-8.
```suggestion
```

x11-wm/pekwm/pekwm-0.2.1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+
+multilib_src_install() {
+
+	cmake_src_install
+
+	# Install contributor scripts into doc folder
+	if use contrib; then
+		docinto contrib
+		dodoc "${S}"/contrib/lobo/*.{pl,vars,png} "${S}"/contrib/lobo/README
+	fi
```
juippis, line 58: Since the `contrib` USE flag is not used anywhere else, i.e. it has no real function in this ebuild, I'd just install these unconditionally and post the postinst message unconditionally too.

x11-wm/pekwm/pekwm-0.2.1.ebuild
```diff
@@ -0,0 +1,87 @@
...
+
+	# Insert an Xsession
+	echo -e "#!/bin/sh\n\n/usr/bin/${PN}" > "${T}"/"${PN}" || die
+	exeinto /etc/X11/Sessions
+	doexe "${T}"/"${PN}"
```
juippis, line 63:
```suggestion
exeinto /etc/X11/Sessions
newexe - pekwm <<- _EOF_
	#!/bin/sh
	/usr/bin/pekwm
_EOF_
```
should also work. Note I didn't test.

x11-wm/pekwm/pekwm-0.2.1.ebuild
juippis, line 31:
```
 * VDB: detected possibly incorrect RDEPEND (x11-wm/pekwm-0.2.1)
 * media-libs/libpng   | media-libs/libpng:=
 *                     > x11-libs/libX11
 *                     > x11-libs/libXext
 * x11-libs/libXrender <
```
So add libX11 and libXext as deps, and verify whether libXrender is needed or not (e.g. the header can be used)

juippis (commented):
Hey again. 

Small things here and there but looking good overall. Let me know if you struggle implementing some of my comments.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: I'm not expert on CMakeLists.txt, but those are defined in the pekwm CMakeLists.txt (and all of them are "ON" by default). I didn't notice this before, but yes this warning is present. My first guess is that they are defined in CMakeLists.txt and not used then, because the definition in the CMakeLists.txt is the same for ENABLE_XINERAMA, but it doesn't complain about this variable. I'll try to figure it out.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: The themes doesn't need to be downloaded anymore, there's a new tool called pekwm_theme, which can handle (select, download and install) this task. pekwm_theme needs git to run. I wasn't sure how to cope with this. I don't and didn't use "theme" feature at all, so I'm not sure what users expect from ebuild...

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: OK, we can make this.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: Yes, not needed, it compiles correctly with suggested patch.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: I tested this, it works.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: Change commited 4bcdd47.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: Change commited 771609d.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: My fault, they should be ENABLE_IMAGE_[JPEG,PNG,XPM] instead of ENABLE_[JPEG,PNG,XPM]. And ENABLE_XSHAPE should be ENABLE_SHAPE, the same for ENABLE_XRANDR -> ENABLE_RANDR. And ENABLE_SHAPE and ENABLE_RANDR is "ON" in CMakeLists.txt, so don't have to be listed. Corrected fd21711.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: I've been thinking of it during weekend and I can't find optimal solution. If one likes to use pekwm_theme then git is required. However, themes can be installed without using pekwm_theme, so you don't need git. If you don't have git and try to use pekwm_theme, pekwm_theme prints warning that git is required and exits. What's recommended? Put git as mandatory requirement or not? Or something else?

juippis: Sounds like a good candidate for *optfeature*.
https://devmanual.gentoo.org/eclass-reference/optfeature.eclass/index.html

krejvl: Good point, it's exactly what we need. Commited (including a slight change to elog messages) eb4f7ad.

x11-wm/pekwm/pekwm-0.2.1.ebuild
```diff
@@ -0,0 +1,86 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit cmake-multilib desktop optfeature
```
juippis, line 6: One thing I didn't spot earlier: Why enable multilib? I can't think of a reason for that.

All in all this looks good apart for few stylistic issues but I can fix those. Let me know if you consider this finished (+ the multilib issue).

x11-wm/pekwm/pekwm-0.2.1.ebuild
```diff
@@ -0,0 +1,86 @@
...
+	https://github.com/pekdon/pekwm
+"
+SRC_URI="
+	https://github.com/pekdon/${PN}/releases/download/release-${PV}/pekwm-${PV}.tar.gz
+"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~mips ~ppc ~x86"
+IUSE="debug +jpeg +png truetype +xft xinerama +xpm"
```
juippis, line 20: `debug` and `xft` are not being used in this ebuild. Please re-check especially the `xft` flag.

x11-wm/pekwm/pekwm-0.2.1.ebuild
```diff
@@ -0,0 +1,86 @@
...
+
+RDEPEND="
+	png? ( media-libs/libpng:0 )
+	jpeg? ( virtual/jpeg:0 )
+	xpm? ( x11-libs/libXpm )
+	x11-libs/libX11
+	x11-libs/libXext
+	truetype? ( x11-libs/libXft )
+	xinerama? ( x11-libs/libXinerama )"
```
juippis, line 29:
```suggestion
	x11-libs/libX11
	x11-libs/libXext
	jpeg? ( virtual/jpeg:0 )
	png? ( media-libs/libpng:0 )
	truetype? ( x11-libs/libXft )
	xinerama? ( x11-libs/libXinerama )
	xpm? ( x11-libs/libXpm )"
```

We like to sort it this way.

x11-wm/pekwm/pekwm-0.2.1.ebuild
```diff
@@ -0,0 +1,86 @@
...
+	# Install contributor scripts into doc folder
+	docinto contrib
+	dodoc "${S}"/contrib/lobo/*.{pl,vars,png} "${S}"/contrib/lobo/README
+
+	# Insert an Xsession
+exeinto /etc/X11/Sessions
+newexe - pekwm <<- _EOF_
+	#!/bin/sh
+	/usr/bin/pekwm
+_EOF_
```
juippis, line 59:
```suggestion
	# Insert an Xsession
	exeinto /etc/X11/Sessions
	newexe - pekwm <<- _EOF_
		#!/bin/sh
		/usr/bin/pekwm
	_EOF_
```
Don't forget to indent.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: Yes, you're right. This is a result of my experiments. To be honest, when I was playing with it, the pure cmake didn't work for me, while cmake-multilib did. It's obvious that it was for a different reason(s) then, but I assumed that cmake-multilib is better supported, even if multilib doesn't make sense for a user app. However, it doesn't make any harm, so I let it in a multilib mode. My fault, I was wrong. It's corrected now via b6be44e.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: Debug is supported by pekwm, so I add this via CMAKE_BUILD_TYPE=$(usex debug Debug). xft removed. Commit b75cdc3.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: OK, but I only commited your previous suggestion, I assumed that this is OK for code inserts.

x11-wm/pekwm/pekwm-0.2.1.ebuild
krejvl: And I would appreciate if you could correct any stylistic issues.

juippis (approved):
Thank you very much!

PR #24347, opened 2022-02-25
app-i18n/transifex-client: add new version 1.0.3

robert7k:


thesamesam:
Nice catch re slugify too.

PR #24349, opened 2022-02-25
Add build time dependency for 'introspection'

ckaktus:
Created new revisions for sys-auth/libfprint 1.94.3, 1.94.2, 1.94.1, 1.92.0 and 1.90.7.
Which add build time dependency to dev-libs/libgusb[introspection], when USE-Flag 'introspection' is active.

### DIFF
--- a/sys-auth/libfprint/libfprint-1.90.7-r1.ebuild
+++ b/sys-auth/libfprint/libfprint-1.90.7-r1.ebuild
@@ -1,0 +1,53 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit meson udev
+
+DESCRIPTION="Library to add support for consumer fingerprint readers"
+HOMEPAGE="https://cgit.freedesktop.org/libfprint/libfprint/ https://github.com/freedesktop/libfprint https://gitlab.freedesktop.org/libfprint/libfprint"
+SRC_URI="https://github.com/freedesktop/libfprint/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="LGPL-2.1+"
+SLOT="2"
+KEYWORDS="~alpha ~amd64 ~arm ~arm64 ~ia64 ~ppc ~ppc64 ~riscv ~sparc ~x86"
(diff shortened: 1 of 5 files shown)


thesamesam:
Thanks!

Would you mind using 'repoman commit' or 'pkgdev commit' to make your commit, and/or manually add the `Signed-off-by` tag and make the commit message adhere to [GLEP 66](https://www.gentoo.org/glep/glep-0066.html#commit-messages)?

Let me know if you need help.

ckaktus:
Hi @thesamesam, I emerged repoman and got an local clone of the repo. After setting up my git for gpg signing I tried `LANG=C repoman -a commit`.
But it fails with:
```
error: gpg failed to sign the data
fatal: failed to write commit object
!!! Exiting on git (shell) error code: 128
```
I tried signing with git alone, this works fine. What am I missing?

thesamesam:
Hi! This is a total PITA. It's because `repoman` tries to GPG sign all commits because `metadata/layout.conf` in the repository says it should. But we don't need contributors to do this, only developers people pushing to the repo.

[There's more information on the wiki](https://wiki.gentoo.org/wiki/Gentoo_git_workflow#GPG_Configuration) but we have two choices:
1. You edit `metadata/layout.conf` and disable commit signing locally (don't include that change in your commits)
or
2. We run `GIT_TRACE=1 repoman commit` (don't worry about getting the commit message right for a moment as it'll die again) and then we try running the `gpg` command which fails directly, and debug from there.

We can always do 2. another time if you want.

ckaktus:
I did number 1. Is it now right?

thesamesam:
> I did number 1. Is it now right?

Looking great! We'll let CI pass then I'll merge. Congratulations on your first contribution. To the first of many? 🥳

ckaktus:
Thanks for the help. I'm a little proud that I could give something back :smile:
I guess there will be some ebuild I could contribute.

thesamesam:
> Thanks for the help. I'm a little proud that I could give something back 😄 I guess there will be some ebuild I could contribute.

You should be! Merged as https://github.com/gentoo/gentoo/commit/fa89888397f205b8de031417d3b011ee3e64bb38. I've unrevbumped as it's a [BDEPEND only change](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html).

PR #24351, opened 2022-02-26
media-gfx/gimp: 2.99.10 dev version bump

band-a-prend:


thesamesam:
Thanks! I've also synced the live ebuilds.

PR #24352, opened 2022-02-26
dev-java/commons-logging: Let build.xml accept log4j-12-api as Log4j 1.2

Leo3418:
https://bugs.gentoo.org/834036 and the `EANT_GENTOO_CLASSPATH` workaround both stemmed from the fact that `build.xml` does not accept the `log4j12.jar` system property to contain multiple file paths separated by `:`.  This also means that only one JAR can be passed into the value of `log4j12.jar`.  Although use of Log4j 2 and the log4j-12-api bridge means that multiple JARs have to be used, they can be combined into a single temporary JAR to satisfy the requirement of `build.xml`.


dev-java/commons-logging/commons-logging-1.2-r5.ebuild
```diff
@@ -55,20 +58,39 @@ src_prepare() {
...
+		# test to return a negative result.  However, classes from multiple
+		# Log4j 2 JARs are needed to compile the sources.  So, we combine
+		# them into a single JAR on the go.
+		# https://bugs.gentoo.org/834036
+		mkdir -p "${T}/log4j-2" ||
+			die "Failed to create temporary directory for Log4j 2 classes"
+		pushd "${T}/log4j-2" > /dev/null ||
+			die "Failed to enter temporary directory for Log4j 2 classes"
+		local jar="$(java-config -j)"
+		"${jar}" -xf "$(java-pkg_getjar log4j-12-api-2 log4j-12-api.jar)" ||
```
thesamesam, line 79: There's nothing wrong with this at all but I wonder if for patterns like this, we should use a loop instead, e.g.

```
local target
# Could even do 12-api, api, core?
for target in log4j-12-api log4j-api log4j-core ; do
	einfo "Running ${jar} blah blah"
	"${jar}" -xf "$(java-pkg_getjar ${target}-2 ${target}.jar)" || die "Failed to extract files from ..."
done
```

Dunno, just an idea, something to keep in mind for stuff like this even if you think it's not a good idea here.

Leo3418: Looks good to me, I'll try that

Leo3418: Done, thanks for the suggestion!

fordfrog:
@Leo3418 merged, thanks!

PR #24361, opened 2022-02-26
media-fonts/fonts-meta: bump to v2

juippis:
- add more latin fonts, remove some personally preferred options,
 - add l10n_ru for a minimal cyrillic support,
 - combine japan, korean and chinese fonts under 'cjk' for tree
   consistency,
 - drop "free" USE flag and provide the "core" fonts under 'latin'
   USE. May be subject to change.

I thought about unconditionally installing the "latin" stuff.

### DIFF
--- a/media-fonts/fonts-meta/fonts-meta-2.ebuild
+++ b/media-fonts/fonts-meta/fonts-meta-2.ebuild
@@ -1,0 +1,126 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+DESCRIPTION="Meta package for fonts to cover most needs"
+HOMEPAGE="https://wiki.gentoo.org/wiki/No_homepage"
+
+LICENSE="metapackage"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="cjk emoji +latin ms"
+
+LANGS="as bn bo brx doi dz gu hi km kn kok ks mai ml mr ne or pa ru sa sat sd si syc ta te th"
+for lang in ${LANGS}; do
+	IUSE+=" l10n_${lang}"
+done
+unset lang LANGS
+
+RDEPEND="media-fonts/cantarell
+	media-fonts/droid
+	media-fonts/noto
+	media-fonts/oldstandard
+	media-fonts/powerline-symbols
+	media-fonts/ubuntu-font-family
+	media-libs/fontconfig
+	|| (
+		media-fonts/source-pro
+		media-fonts/anonymous-pro
+		media-fonts/cascadia-code
+		media-fonts/fantasque-sans-mono
+		media-fonts/fira-code
+		media-fonts/fira-mono
+		media-fonts/hack
+		media-fonts/hermit
+		media-fonts/ibm-plex
+		media-fonts/inconsolata
+		media-fonts/iosevka
+		media-fonts/jetbrains-mono
+		media-fonts/montecarlo
+		media-fonts/roboto
+		media-fonts/terminus-font
+	)
+	cjk? (
+		media-fonts/ipaex
+		media-fonts/ja-ipafonts
+		media-fonts/koruri
+		media-fonts/monafont
+		media-fonts/nanum
+		media-fonts/noto-cjk
+		media-fonts/opendesktop-fonts
+		media-fonts/shinonome
+		media-fonts/source-han-sans
+		media-fonts/unfonts
+		media-fonts/wqy-microhei
+		media-fonts/wqy-zenhei
+		|| (
+			media-fonts/mikachan-font-otf
+			media-fonts/mikachan-font-ttf
+		)
+	)
+	emoji? (
+		|| (
+			media-fonts/noto-emoji
+			media-fonts/joypixels
+		)
+	)
+	l10n_as? ( media-fonts/lohit-assamese )
+	l10n_bn? ( media-fonts/lohit-bengali )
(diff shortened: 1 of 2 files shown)


media-fonts/fonts-meta/fonts-meta-2.ebuild
```diff
@@ -0,0 +1,108 @@
...
+
+EAPI=8
+
+DESCRIPTION="Meta package for fonts to cover most needs"
+HOMEPAGE="https://wiki.gentoo.org/wiki/No_homepage"
+
+LICENSE="metapackage"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="cjk emoji latin ms"
```
juippis, line 12:
```suggestion
IUSE="cjk emoji +latin ms"
```

media-fonts/fonts-meta/fonts-meta-2.ebuild
```diff
@@ -0,0 +1,108 @@
...
+		media-fonts/liberation-fonts
+		media-fonts/noto
+		media-fonts/oldstandard
+		media-fonts/open-sans
+		media-fonts/quivira
+		media-fonts/signika
+		media-fonts/tex-gyre
+		media-fonts/ubuntu-font-family
+		media-fonts/urw-fonts
+		media-libs/fontconfig
```
DarthGandalf, line 101: this doesn't belong to "latin?", and is already dependent unconditionally at the top

media-fonts/fonts-meta/fonts-meta-2.ebuild
```diff
@@ -0,0 +1,108 @@
...
+	l10n_th? ( media-fonts/thaifonts-scalable )
+	latin? (
+		media-fonts/cantarell
+		media-fonts/courier-prime
+		media-fonts/dejavu
+		media-fonts/droid
+		media-fonts/font-bh-ttf
+		media-fonts/font-cursor-misc
+		media-fonts/liberation-fonts
+		media-fonts/noto
```
DarthGandalf, line 93: How did you choose which fonts to use? I see you replaced hack and source-pro with other fonts

DarthGandalf: My wish for this ebuild is to have a good selection of reasonably popular fonts, and to render all the glyphs correctly, (provided that the correct language USE-flag is set).

I don't know how to measure popularity of fonts. I personally use hack in terminal, and can readd it back to my world file when this PR is merged. Tbh, I thought hack is pretty popular (in my social bubble), so I probably wouldn't remove it yet. But that goes to my preference, ignoring preference of other users.

Maybe install all the popular fonts, and let users choose which font to use among the installed ones? After all, users always have a choice to install the wanted fonts manually, and ignore this meta package if they want to minimize the list of fonts installed

juippis: > My wish for this ebuild is to have a good selection of reasonably popular fonts, and to render all the glyphs correctly, (provided that the correct language USE-flag is set).

Yep, agree there.

> 
> I don't know how to measure popularity of fonts. I personally use hack in terminal, and can readd it back to my world file when this PR is merged. Tbh, I thought hack is pretty popular (in my social bubble), so I probably wouldn't remove it yet. But that goes to my preference, ignoring preference of other users.
> 

I just googled "popular (monospace) fonts" and hack was rarely represented in those lists. 

> Maybe install all the popular fonts, and let users choose which font to use among the installed ones? 

If I understand you correctly here, that crossed my mind, but I'd rather leave the choice for users to begin with. This way you're not installing 100 fonts where you use 1. This is especially true with monospace fonts. 
What do you think about this dirty solution:
```
|| (
	monospace-font-1
	monospace-font-2
	monospace-font-3
	monospace-font-4
)
```

We'd add this to the very core, or under "latin".

> After all, users always have a choice to install the wanted fonts manually, and ignore this meta package if they want to minimize the list of fonts installed

Indeed, but I believe we should attempt to provide reasonable defaults here. I thought about including pretty much ~every font under media-fonts/* in this ebuild one way or another (`extra` use flag for example), but it quickly became a stupid idea, since users can choose to install whatever extra they want.

DarthGandalf: >  What do you think about this dirty solution:

That works. In that case the first font in the list should have distinguishable 0Oo, lI1|. Not all fonts you selected for the "new" list satisfy that.

> We'd add this to the very core, or under "latin".

I'd say core. Having *a* monospace font (which is not necessary latin-only) is useful even without fonts specific to latin.

> ~every font under media-fonts/*

I actually had tried this for a while, the results were bad. Some wild font (don't know which one) was causing rendering issues by declaring that it has all the glyphs, but actually having only some. And there was a case where some programs couldn't decide which font to use to render certain glyphs, and randomly were switching between fonts

media-fonts/fonts-meta/fonts-meta-2.ebuild
```diff
@@ -0,0 +1,127 @@
...
+	media-fonts/ubuntu-font-family
+	media-libs/fontconfig
+	|| (
+		media-fonts/source-pro
+		media-fonts/anonymous-pro
+		media-fonts/cascadia-code
+		media-fonts/fantasque-sans-mono
+		media-fonts/fira-code
+		media-fonts/fira-mono
+		media-fonts/fira-sans
```
juippis, line 34: The package description says this is a monospace font, but it doesn't look like it...
https://mozilla.github.io/Fira/
```suggestion
```

DarthGandalf (approved):

PR #24368, opened 2022-02-27
dev-java/stringtemplate: Fix 4.3.1 test failure when upgrading slot 4

Leo3418:
A test failure (as reported in the linked bug) might occur when an older
version of ST 4 is already installed on the system.  The failure is
caused by multiple factors: the way java-pkg-simple.eclass generates the
test classpath, the fact that the tests launch new JVM instances under a
different working directory, and the behavior of the JVM upon an invalid
path in the classpath.

As of the time when this commit is created, the test classpath computed
by java-pkg-simple.eclass will contain the following elements, in order:
1. The directory containing the test classes compiled by the eclass
2. The path to the JAR built by the eclass **relative to ${S}**
3. Absolute paths to dependency JARs, for both compile and test
   dependencies

ST 4 has an implicit test dependency on itself via ANTLR 3.5 (which is
yet another issue pertaining to the troublesome, malformed ANTLR 3.5 and
ST 4 circular dependency).  This means a version of slot 4 that is
possibly older than 4.3.1 might be installed on the system and added to
the test classpath within element No. 3 when the tests are being run.

Some of the tests will call the 'java' command to execute ST 4 in a new
JVM instance; the test classpath generated by java-pkg-simple.eclass
will be reused in the classpath of the new JVM.  The unfortunate factor
that triggers the test failure is that the new JVM's working directory
can become different from the one of the original JVM running the JUnit
tests, which is ${S} when this package is being built by Portage.

Note that element No. 2 in the test classpath is a relative path: after
the working directory is changed, it will be invalid.  However, JVM is
lax in invalid path elements in the classpath: it will just ignore them
and emit a "class not found" error or alike only after it has tried all
other paths in the classpath to locate a class.

With this behavior, JVM will pick up the copy of ST 4 already installed
on the system from element No. 3 in the classpath after it detects that
element No. 2 is an invalid path.  Therefore, the tests will be run
against ST 4 that is *installed on the system* instead of the copy that
has just been built.

This explains why some tests would fail when an older version of ST 4 is
already installed; effectively, that old version was being tested by the
test suite for the new version, and there is no guarantee that all tests
would pass in this case.  A corollary conclusion is that if the same
version of ST 4 is being built and installed twice, and the second build
has tests enabled, then the tests would pass, although effectively it
would be the artifact produced by the first build being tested against.

Closes: https://bugs.gentoo.org/834138

### DIFF
--- a/dev-java/stringtemplate/stringtemplate-4.3.1.ebuild
+++ b/dev-java/stringtemplate/stringtemplate-4.3.1.ebuild
@@ -68,6 +68,29 @@ src_prepare() {
 	rm -v "${JAVA_TEST_SRC_DIR}/org/stringtemplate/v4/test/TestEarlyEvaluation.java" || die
 }
 
+src_test() {
+	# Make sure no older versions of this slot are present in the classpath
+	# https://bugs.gentoo.org/834138#c4
+	local old_ver_cp="$(nonfatal java-pkg_getjars "${PN}-${SLOT}")"
+	local new_test_cp="$(\
+		java-pkg_getjars --with-dependencies "${JAVA_TEST_GENTOO_CLASSPATH}")"
+	new_test_cp="${new_test_cp//"${old_ver_cp}"/}"
+
+	# Some of the test cases require an absolute path to the JAR being tested
+	# against to be in the classpath, due to the fact that they call the 'java'
+	# command outside ${S} and reuse the classpath for the tests:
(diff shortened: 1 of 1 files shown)


fordfrog:
@Leo3418 merged, thanks! quite an investigation... :-) i got a warning `java-pkg_ensure-dep: The ebuild is attempting to use stringtemplate-4, which is not declared with a SLOT in DEPEND.` in the test phase but that's understandable. tests do not fail anymore fore me. well done! :-)

PR #24374, opened 2022-02-28
sys-process/below: Add 0.5.0 ebuild

hodgesds:
Tested with `ebuild below-0.5.0.ebuild install`. 

Signed-off-by: Daniel Hodges <hodges.daniel.scott@gmail.com>


sys-process/below/below-0.5.0.ebuild
```diff
@@ -0,0 +1,218 @@
...
+DEPEND="
+	${RDEPEND}
+	sys-libs/ncurses
+"
+
+QA_FLAGS_IGNORED="usr/bin/below"
+src_install() {
+	cargo_src_install --path below
+
+	keepdir /var/log/below
```
Flowdalic, line 217: This is missing the systemd service when compared to the previous versions ebuild. Is this intentional?

sys-process/below/below-0.5.0.ebuild
```diff
@@ -0,0 +1,218 @@
...
+	$(cargo_crate_uris)
+"
+LICENSE="Apache-2.0 Apache-2.0-with-LLVM-exceptions BSD-2 BSD Boost-1.0 ISC LGPL-2.1 MIT MPL-2.0 Unlicense"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc64"
+
+BDEPEND="
+	sys-devel/clang
+	virtual/pkgconfig
+	>=virtual/rust-1.56.0[rustfmt]
```
Flowdalic, line 202: nit: the previous version ebuild states here 'virtual/rust-1.56', i.e., the version identifier is without the trailing '.0'. For easier comparison, this would stay the same way as the previous version.

sys-process/below/below-0.5.0.ebuild
```diff
@@ -0,0 +1,218 @@
+# Copyright 2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+CRATES="
+ahash-0.6.3
```
Flowdalic, line 7: The previous version ebuild indents the crate's names by one tab. This should be done here too.

sys-process/below/below-0.5.0.ebuild
```diff
@@ -0,0 +1,218 @@
+# Copyright 2021 Gentoo Authors
```
Flowdalic, line 1: The copyright should state '2022'.

Flowdalic (changes requested):
Thanks for your contribution, please find my comments in line.

sys-process/below/below-0.5.0.ebuild
hodgesds: Bad copy/paste, updated.

PR #24380, opened 2022-02-28
dev-cpp/ms-gsl: Version bump 4.0.0

janhenke:


dev-cpp/ms-gsl/ms-gsl-4.0.0.ebuild
```diff
@@ -0,0 +1,33 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
juippis, line 4: Let's start using 8 where possible.

dev-cpp/ms-gsl/ms-gsl-9999.ebuild
```diff
@@ -1,4 +1,4 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
```
juippis, line 4: Here too.

juippis (approved):

PR #24388, opened 2022-03-02
app-emulation/libvirt: version bump to 8.1.0

zippy2:
Ideally, this would be way simpler, just introduce new ebuild and
append hashes to the Manifest file. Unfortunately, a nasty
crasher was found in the freshly released 8.1.0 so we need to
backport the fix. And while at it, rebase two patches that don't
apply cleanly anymore (libvirt-6.7.0-do-not-use-sysconfig.patch
and libvirt-6.7.0-fix-paths-for-apparmor.patch).

Closes: https://bugs.gentoo.org/834483


app-emulation/libvirt/libvirt-8.1.0.ebuild
```diff
@@ -0,0 +1,337 @@
...
+
+	# Depending on configuration option, libvirt will create some bogus
+	# directoreis. They are either not used, or libvirtd is able to create
+	# them on demand, so let's remove them.
+	#
+	# Note, we are using -f here so that rm does not fail or warn if the
+	# directory is nonexistent.
+	rm -rf "${D}"/etc/sysconfig
+	rm -rf "${D}"/var
+	rm -rf "${D}"/run
```
juippis, line 308: This is exactly the reason we usually don't use `-f`, we end up carrying these commands forever when they stop being relevant.

juippis (approved):
Thanks! Do you have a clue what causes:
```
src/security/apparmor/meson.build:17:2: ERROR: File usr.libexec.libvirt.virt-aa-helper.in does not exist.

A full log can be found at /var/tmp/portage/app-emulation/libvirt-8.1.0/work/libvirt-8.1.0-build/meson-logs/meson-log.txt
 * ERROR: app-emulation/libvirt-8.1.0::gentoo failed (configure phase):
 *   (no error message)
```
```
atom: =app-emulation/libvirt-8.1.0, USE flags: 'apparmor audit -bash-completion -caps -dtrace firewalld -fuse glusterfs iscsi iscsi-direct libssh libvirtd -lvm -lxc -nfs nls numa -openvz -parted -pcap -policykit qemu rbd -sasl udev -verify-sig virt-network -virtualbox wireshark-plugins -xen -zfs'
```

PR #24391, opened 2022-03-03
dev-util/packer: add 1.7.10

mrgz:
Signed-off-by: Marcos Rodrigues Gonzalez <mrgonzalez1275@gmail.com>


juippis (commented):
We're currently updating our policy regarding go packaging. You'll have to vendor the deps by yourself, if upstream doesn't provide a pre-vendored tarball. Please see the gentoo-dev mailing list about it, and monitor https://gitweb.gentoo.org/repo/gentoo.git/log/eclass/go-module.eclass when it gets the vendor support. Then we can process this PR.

mrgz:
@juippis, I've been following some of it, but are we blocking all EGO_SUM-based packages by now, even without being finished (eclass and docs)?

Thanks for reviewing it.

juippis:
You _can_ already use the manually vendored tarball, I just don't know how to package them by the "right way" before the eclass docs got finished. I don't do go...

but if you can get it to work, go ahead! I guess you'll want to git checkout to the tag you're releasing and call some **go vendor** or something.

thesamesam:
ping

mrgz:
Hey, @thesamesam. Sorry, been very busy last weeks.

Where is a safe place to put the go mod download result? Looking at other solutions to address EGO_SUM migration, most of them are using dev areas at dev.gentoo.org.

The compressed tar file is ~730M (almost half of it, azure deps)

Cheers,

thesamesam:
No problem! I was just going through some open PRs :)

Could you put it in a repo on GitHub or similar for now? But oh man that's huge. IIRC it _can_ be smaller with 'go mod vendor' and see if it works?

AFAIK that's usually smaller but isn't guaranteed to always work.

mrgz:
The "go mod vendor" approach doesn't work: https://dpaste.com/DJWHCEZBX. But "go mod download" works well. 

I shrank the "mod download" result to 600MB with -J, though it is still pretty big. I'll upload it somewhere and rebase the commit.

juippis (approved):
This still doesn't feel right to me. But we'll go with how we are instructed. Thanks! 

I'm gonna cleanup the old versions to get rid of that massive Manifest file.

PR #24393, opened 2022-03-03
app-misc/fdutils: New version

laumann:
* Bumps to EAPI=8

 * Most patching from previous version is no longer necessary (except
   for one)

 * Keeps all the Debian patches minus one

Closes: https://bugs.gentoo.org/826430


laumann:
@thesamesam This builds (also with USE="doc") but I have a few doubts:

 - The only debian patch that's being dropped appears to be necessary for Linux 2.6, but it's not necessary for my current kernel version (5.16)
 - There are two new variables in doc/Makefile.in - `@CC_FOR_BUILD@` and @EXEEXT_FOR_BUILD@` that aren't replaced when configure is called. I figured they should be replaced with `$(CC)` and the empty string respectively, but I don't know if that's the right approach

laumann:
Changes:

 - Made it version 5.6_p2 (upstream version + debian patch version)
 - Shuffled the dependencies around a bit
 - The new 5.6 patches were also renamed

app-misc/fdutils/fdutils-5.6_p2.ebuild
```diff
@@ -0,0 +1,64 @@
...
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="doc"
+
+RDEPEND=">=sys-fs/mtools-4"
+DEPEND="${RDEPEND}
+	virtual/os-headers
+"
+BDEPEND="
+	sys-apps/texinfo
```
laumann, line 25: I wonder if this should be
```
doc? ( sys-apps/texinfo )
```

PR #24398, opened 2022-03-04
sys-apps/bat: add v0.20.0

ixti:
Signed-off-by: Alexey Zapparov <alexey@zapparov.com>


ixti:
@toshokan another version bump :D

sys-apps/bat/bat-0.20.0.ebuild
```diff
@@ -0,0 +1,202 @@
...
+
+BDEPEND="virtual/pkgconfig"
+DEPEND="
+	>=dev-libs/libgit2-1.1.0:=[threads]
+	dev-libs/oniguruma:=
+"
+# >app-backup/bacula-9.2[qt5] has file collisions, #686118
+RDEPEND="${DEPEND}
+	!>app-backup/bacula-9.2[qt5]
+"
```
00-matt, line 173: Is this blocker still required?

ixti: Unfortunately, yes. `app-backup/bacula[qt5]` builds and installs `bat` (bacula administration tool) – a GUI form of `bconsole`.

ixti: Oh, actually it might be not. Seems like bacula is installing bat to `/usr/sbin`

ixti:
```
 * Messages for package app-backup/bacula-11.0.5:

 * This package will overwrite one or more files that may belong to other
 * packages (see list below). You can use a command such as `portageq
 * owners / <filename>` to identify the installed package that owns a
 * file. If portageq reports that only one package owns a file then do
 * NOT file a bug report. A bug report is only useful if it identifies at
 * least two or more packages that are known to install the same file(s).
 * If a collision occurs and you can not explain where the file came from
 * then you should simply ignore the collision since there is not enough
 * information to determine if a real problem exists. Please do NOT file
 * a bug report at https://bugs.gentoo.org/ unless you report exactly
 * which two packages install the same file(s). See
 * https://wiki.gentoo.org/wiki/Knowledge_Base:Blockers for tips on how
 * to solve the problem. And once again, please do NOT file a bug report
 * unless you have completely understood the above message.
 * 
 * Detected file collision(s):
 * 
 * 	/usr/share/man/man1/bat.1.bz2
 * 
 * Searching all installed packages for file collisions...
 * 
 * Press Ctrl-C to Stop
 * 
 * sys-apps/bat-0.20.0:0::ixti-overlay
 * 	/usr/share/man/man1/bat.1.bz2
 * 
 * Package 'app-backup/bacula-11.0.5' NOT merged due to file collisions.
 * If necessary, refer to your elog messages for the whole content of the
 * above message.
```

ixti: Still needed.

00-matt: Thank you for checking!

toshokan (approved):

toshokan:
Looks good to me, thank you @ixti!

MatthewGentoo (approved):

PR #24399, opened 2022-03-04
sys-fs/loop-aes: Version bump, minor Gentoo compat fixes

hlein:
Upstream version bump added 5.16 support. Also, updates to the bundled
build-initrd.sh script to be compatible with Gentoo's install ISO kernel
(see bug).

Closes: https://bugs.gentoo.org/834546


sys-fs/loop-aes/loop-aes-3.7w.ebuild
```diff
@@ -0,0 +1,69 @@
...
+SRC_URI="http://loop-aes.sourceforge.net/loop-AES/${MY_P}.tar.bz2"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~hppa ~ppc ~sparc ~x86"
+IUSE="cpu_flags_x86_aes extra-ciphers keyscrub cpu_flags_x86_padlock"
+
+DEPEND="app-crypt/loop-aes-losetup"
+
+PATCHES=( "${FILESDIR}"/build-initrd_explicit-losetup.patch )
```
juippis, line 21: We like to prepend these patch names with `${P}`, 
https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html

juippis (approved):

PR #24401, opened 2022-03-04
sys-cluster/glusterfs: systemd support.

jkroonza:
Closes: https://bugs.gentoo.org/834502


jkroonza:
Three main branches, just went with straight to stable for the ebuilds that was already stable.  really should get onto stabling 9.5, and it looks like 10.1 may just as well go too, will get rid of 8.X next month.  Really wish we had installation stats of sorts.

MatthewGentoo (approved):

MatthewGentoo:
>  Really wish we had installation stats of sorts

We have [Goose](https://github.com/mgorny/goose/) and [Gander](https://github.com/mgorny/gander/) but they are incomplete and not used (yet :))

PR #24402, opened 2022-03-04
www-client/seamonkey: update to version 2.53.11

BioMike:
PR to update www-client/seamonkey to version 2.53.11.

Important changes to the ebuild:
* Patch files are sourced from [my gentoo-seamonkey-patches repo](https://github.com/BioMike/gentoo-seamonkey-patches)
* Patches are applied on an individual basis (to allow more flexibility in the future).
* Applied suggested improvements and fixes from previous PR: [23983](https://github.com/gentoo/gentoo/pull/23983)
* Removed the --enable-extensions config flag, which also needed removal of this flag from mozcoreconf-v6.eclass. Didn't encounter any problems building current stable seamonkey against this changed eclass. Seamonkey is only user of that eclass.
* Included a GTK+2 removal patch and removal of the GTK+2 dependency. I've tried this on my system and didn't encounter any problems with it. Upstream said there were a few issues with it, but didn't specify which (most likely other system targets, e.g. MS Windows).
*  Included myself in metadata.xml.


www-client/seamonkey/seamonkey-2.53.11.ebuild
```diff
@@ -0,0 +1,585 @@
...
+SLOT="0"
+SYSTEM_IUSE=( +system-{av1,harfbuzz,icu,jpeg,libevent,libvpx,png,sqlite} )
+IUSE="+chatzilla cpu_flags_arm_neon +crypt dbus debug +gmp-autoupdate +ipc jack
+lto pulseaudio selinux startup-notification test wifi"
+IUSE+=" ${SYSTEM_IUSE[@]}"
+KEYWORDS="~amd64 ~ppc64 ~x86"
+
+RESTRICT="!test? ( test )"
+
+ASM_DEPEND=">=dev-lang/yasm-1.1"
```
juippis, line 55: This seems redundant. Did this maybe sometime plan to support nasm too?

www-client/seamonkey/seamonkey-2.53.11.ebuild
```diff
@@ -0,0 +1,585 @@
...
+	eapply "${SM_PATCHDIR}/2018_audio_device_backends.patch"
+	eapply "${SM_PATCHDIR}/6001_add_missing_header_for_basename.patch"
+	eapply "${SM_PATCHDIR}/6002_add_alternate_name_for_private_siginfo_struct_member.patch"
+	eapply "${SM_PATCHDIR}/6003_fix_syscall_wrappers_on_musl.patch"
+	eapply "${SM_PATCHDIR}/6004_musl_drop_alloc_hooks.patch"
+	eapply "${SM_PATCHDIR}/6005_musl_memory_report.patch"
+	eapply "${SM_PATCHDIR}/6006_musl_pthread_setname.patch"
+	eapply "${SM_PATCHDIR}/6007_musl_fix_tools.patch"
+	eapply "${SM_PATCHDIR}/6008_musl_fix_toolkit.patch"
+	eapply "${SM_PATCHDIR}/6018_audio_backend_update_generated_files.patch"
```
juippis, line 224: I think `eapply ${WORKDIR}/gentoo-${PN}-patches-${PV}/${PN}/*` would work here. 
Although I'd very much prefer to switch to something similar that firefox, thunderbird and spidermonkey uses at the moment.

eclass/mozcoreconf-v6.eclass
```diff
@@ -76,18 +76,6 @@ mozconfig_use_with() {
...
-# enable or disable an extension based on a USE-flag
-#
-# Example:
-# mozconfig_use_extension gnome gnomevfs
-# => ac_add_options --enable-extensions=gnomevfs
-mozconfig_use_extension() {
-	declare minus=$(use $1 || echo -)
-	mozconfig_annotate "${minus:-+}$1" --enable-extensions=${minus}${2}
-}
-
```
juippis, line 90: Do we *need* to remove this option now?

eclass/mozcoreconf-v6.eclass
```diff
@@ -289,12 +277,6 @@ mozconfig_final() {
 	done
 	echo "=========================================================="
 	echo
-
-	# Resolve multiple --enable-extensions down to one
-	declare exts=$(sed -n 's/^ac_add_options --enable-extensions=\([^ ]*\).*/\1/p' \
-		.mozconfig | xargs)
-	sed -i '/^ac_add_options --enable-extensions/d' .mozconfig
-	echo "ac_add_options --enable-extensions=${exts// /,}" >> .mozconfig
```
juippis, line 297: This too.

juippis (commented):
You should also squash these commits to 1-2 at max, and describe your changes in the commit message box.

Doing some build tests now.

www-client/seamonkey/seamonkey-2.53.11.ebuild
```diff
@@ -0,0 +1,585 @@
...
+	x11-libs/gdk-pixbuf
+	>=x11-libs/gtk+-3.4.0:3
+	x11-libs/libX11
+	x11-libs/libXcomposite
+	x11-libs/libXdamage
+	x11-libs/libXext
+	x11-libs/libXfixes
+	x11-libs/libXrender
+	x11-libs/libXt
+	>=x11-libs/pango-1.22.0
```
juippis, line 91:
```suggestion
	x11-libs/libXt
	x11-libs/libxcb:=
	>=x11-libs/pango-1.22.0
```

www-client/seamonkey/seamonkey-2.53.11.ebuild
BioMike: I agree, this could have some improvement. This is a part that I haven't touched yet after taking over the ebuild from the previous maintainer.

www-client/seamonkey/seamonkey-2.53.11.ebuild
BioMike: It's mostly done like this to keep the lines short and ebuild file size reduced.

The way patches were applied in previous ebuilds is the same way firefox, thunderbird and spidermonkey do it, but there is one patch that's USE flag dependent and was patched completely separate from the other patches. I changed to this to keep all the patches together. Some of the other patches in there could also become USE flag dependent, mostly the system-* patches.

eclass/mozcoreconf-v6.eclass
BioMike: Not sure about this part, but the part below passes the option flag, even if empty, to the seamonkey configure script. This causes the seamonkey configure script to fail due to the flag being present. Do note that this configure option is obsolete for quite some time (has only been in there to support the flash plugin, which Gentoo doesn't have any more).

eclass/mozcoreconf-v6.eclass
BioMike: See above.

www-client/seamonkey/seamonkey-2.53.11.ebuild
BioMike: Removed redundancy in latest commit.

www-client/seamonkey/seamonkey-2.53.11.ebuild
BioMike: Added in latest commit.

eclass/mozcoreconf-v6.eclass
juippis: Yeah I was looking into just disabling the option, but I guess this is fine too.

www-client/seamonkey/seamonkey-2.53.11.ebuild
juippis: I don't think you understood what I meant here. Indeed I want to squash these 30 lines into one, and reverse the logic for libvpx patch, same way as is done in firefox+lto. But I can test the syntax and fix-on-merge if that's fine for you.

www-client/seamonkey/seamonkey-2.53.11.ebuild
```diff
@@ -0,0 +1,584 @@
...
+	done
+
+	use system-libvpx \
+		&& eapply -p2 "${SM_PATCHDIR}/1009_seamonkey-2.53.3-system_libvpx-1.8.patch"
+
+	# Allow user to apply any additional patches without modifing ebuild
+	eapply_user
+
+	# Don't error for format with gcc-9
+	grep -rl -- '-Werror=format' | xargs sed -i 's/error=format/no-&/' || die
```
juippis, line 239: I'd prefer the `find -exec sed` syntax here but guess this piping is fine too.

BioMike: This is from the previous maintainer, I only added the "|| die" . I agree that a "find/sed"-way of doing this is better readable. I'm not sure if this line is still really needed as all build targets have moved on to gcc-11 by now and I don't encounter errors when building the development version of seamonkey manually (outside of portage). I'll have a closer look at this with the upcoming release.

BioMike:
Latest commit reverts the way of applying patches.

www-client/seamonkey/seamonkey-2.53.11.ebuild
```diff
@@ -0,0 +1,554 @@
...
+
+	# Shell scripts sometimes contain DOS line endings; bug 391889
+	grep -rlZ --include="*.sh" $'\r$' . |
+	while read -r -d $'\0' file ; do
+		einfo edos2unix "${file}"
+		edos2unix "${file}"
+	done
+
+	use system-libvpx \
+		&& eapply -p2 ""${WORKDIR}"/gentoo-${PN}-patches-${PV}-r1/USE_flag/1009_seamonkey-2.53.3-system_libvpx-1.8.patch"
```
thesamesam, line 203: Nit (I can fix this on merge):
```suggestion
		&& eapply -p2 "${WORKDIR}"/gentoo-${PN}-patches-${PV}-r1/USE_flag/1009_seamonkey-2.53.3-system_libvpx-1.8.patch
```

www-client/seamonkey/seamonkey-2.53.11.ebuild
```diff
@@ -0,0 +1,554 @@
...
+	cd "${S}/${P}-l10n" || die
+	unpack ${l10n_sources}
+}
+
+src_prepare() {
+	# Apply our patches
+	eapply "${WORKDIR}"/gentoo-${PN}-patches-${PV}-r1/${PN}
+
+	# Shell scripts sometimes contain DOS line endings; bug 391889
+	grep -rlZ --include="*.sh" $'\r$' . |
```
thesamesam, line 196: I think it's probably better to pipe this into the loop but it's not a big deal for now. We can do it next time.

juippis (approved):

www-client/seamonkey/seamonkey-2.53.11.ebuild
BioMike: If there are no other objections, please do. While you're at it, please do the suggestion by juippis below (regarding moving the && to the line above) as well.

www-client/seamonkey/seamonkey-2.53.11.ebuild
BioMike: This was suggested by floppym in the referenced bug in 2012, not sure how often this happens these days (upstream having QA checks against this?).

www-client/seamonkey/seamonkey-2.53.11.ebuild
thesamesam: ack, thanks juippis!

thesamesam:
Merged as:
- https://github.com/gentoo/gentoo/commit/0dd5dd3c8e28899d1be993cb5ddac78b232e4c9a
- https://github.com/gentoo/gentoo/commit/0dbb107e45d99a2ae60e26dfa9c0c40cdcfe85b2
- https://github.com/gentoo/gentoo/commit/2ed4096c377a7223fc27633dda6a98e0fcac9d87

My minor changes after:
- https://github.com/gentoo/gentoo/commit/83c3a7b6b08406cded7765d70107a74670b0b72d
- https://github.com/gentoo/gentoo/commit/8b91b86f9b0536130e467ee5f52cc41c27dddc75

Thanks a lot for your significant work here!

juippis:
Thank you **both**!

PR #24404, opened 2022-03-04
x11-misc/xscreensaver: fix configure with USE=-opengl

hlein:
Deliberately disabling opengl became a fatal error in configure.

Closes: https://bugs.gentoo.org/834582


hlein:
Note, upstream replied with essentially WONTFIX, don't care about non-GL builds (although then I wonder, why even have a `--without-gl` that does not work).

So I think Gentoo's choices are either merge this and own the patch, or, drop `USE=-opengl` support in our ebuild.

juippis (commented):
I think we keep using the patch as long as it works.
A note: I couldn't build this **at all** without `+gdk-pixbux` and *x11-libs/gdk-pixbuf-xlib* being installed. There may be some conditions required there.

And a question: why revbump, this is fixing a configure fault only right? Nothing in runtime is affected so far?

hlein:
> And a question: why revbump, this is fixing a configure fault only right? Nothing in runtime is affected so far?

Oh, good call, it does not really need a revbump, since on an affected system it's impossible to build the current version, and on an unaffected system, this change should not matter.

hlein:
> > And a question: why revbump, this is fixing a configure fault only right? Nothing in runtime is affected so far?
> 
> Oh, good call, it does not really need a revbump, since on an affected system it's impossible to build the current version, and on an unaffected system, this change should not matter.

OK, cleaned up the unneeded revbump @juippis

PR #24413, opened 2022-03-05
media-video/wireplumber: add a fix for PW 0.3.48

pinkflames:
PW 0.3.48 contains a change that may break PulseAudio support without
a fix on WP side. This commit adds that fix as wireplumber-0.4.8-r3
ebuild. Because ~testing only needs to work with ~testing, it has not
been verified that the r3 works as intended with PW versions older than
the current ~testing version. However there's no reason to believe it
would not, so the version minimum is not being bumped up.

Bug: https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/2189

### DIFF
--- a/media-video/pipewire/pipewire-0.3.48.ebuild
+++ b/media-video/pipewire/pipewire-0.3.48.ebuild
@@ -108,7 +108,7 @@ DEPEND="${RDEPEND}"
 
 # TODO: Consider use cases where pipewire is not used for driving audio
 # Doing so with WirePlumber currently involves editing Lua scripts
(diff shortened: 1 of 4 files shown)


thesamesam:
We should make sure PW depends on at least this version to avoid duplicate bug reports and confusion for partial upgrades.

also, thanks!

pinkflames:
I was under impression that ~testing is meant to be used with other [latest] ~testing packages, therefore any bugs from mixed ~testing and stable or not latest ~testing Gentoo ebuilds was on the user. But, of course, this is a cheap thing to do, so I'll make another commit to do that.

PR #24415, opened 2022-03-05
net-analyzer/smokeping: drop init script 'restore' function

ajakk:
Bug: https://bugs.gentoo.org/602652
Signed-off-by: John Helmert III <ajak@gentoo.org>


ajakk:
Smokeping is last rited now.

PR #24420, opened 2022-03-06
net-news/sfeed: add v1.3 and some improvements

N-R-K:
removes unneeded depandencies
adds theme support via USE flag and savedconfig
downgrade to EAPI 7 (8 doesn't support savedconfig)

Closes: https://bugs.gentoo.org/834240


net-news/sfeed/sfeed-1.2.ebuild
```diff
@@ -45,4 +49,6 @@ src_install() {
 		DOCPREFIX="${EPREFIX}/usr/share/doc/${P}" \
```
AndrewAmmerlaan, line 49:
```suggestion
		DOCPREFIX="${EPREFIX}/usr/share/doc/${PF}" \
```
Correct location for the documentation is `PF` not `P` (PF also includes the revision)

net-news/sfeed/sfeed-1.2.ebuild
```diff
@@ -12,29 +12,33 @@ SRC_URI="https://codemadness.org/releases/${PN}/${P}.tar.gz"
...
+
+	restore_config $(printf "themes/%s.h " ${THEMES[@]//-/_})
+}
 
 src_compile() {
 	emake \
 		CC=$(tc-getCC) \
 		RANLIB=$(tc-getRANLIB) \
 		AR=$(tc-getAR) \
-		SFEED_CURSES=$(usex ncurses "sfeed_curses" "")
```
ionenwks, line 37: Unrelated to the changes here, but the CC/RANLIB/AR above this line are lacking quotes

if for example had CC="gcc -m32" (or anything else, but -m32 is a common case) it'll fail to build.

net-news/sfeed/sfeed-1.2.ebuild
```diff
@@ -12,29 +12,33 @@ SRC_URI="https://codemadness.org/releases/${PN}/${P}.tar.gz"
 LICENSE="ISC"
 SLOT="0"
 KEYWORDS="~amd64"
-IUSE="ncurses"
-
-DEPEND="ncurses? ( sys-libs/ncurses )"
-RDEPEND="${DEPEND}
-	net-misc/curl
```
ionenwks, line 19: curl is used by sfeed_update and shouldn't be considered an implicit system dep (unlike coreutils, glibc, and virtual/awk to a rough extend).

net-news/sfeed/sfeed-1.2.ebuild
```diff
@@ -12,29 +12,33 @@ SRC_URI="https://codemadness.org/releases/${PN}/${P}.tar.gz"
...
-	sys-apps/coreutils
-	sys-libs/glibc
-	virtual/awk
-	www-client/lynx
-	x11-misc/xclip
-	x11-misc/xdg-utils"
```
ionenwks, line 25: fwiw these are used by sfeed_curses/_content, lynx notably to dump html to text (although could be semi-optfeatures, not that I've tried sfeed I don't have more to say).

net-news/sfeed/sfeed-1.2.ebuild
```diff
@@ -12,29 +12,33 @@ SRC_URI="https://codemadness.org/releases/${PN}/${P}.tar.gz"
...
+
+DEPEND="ncurses? ( sys-libs/ncurses:= )"
+RDEPEND="${DEPEND}"
+BDEPEND="virtual/pkgconfig"
+
+src_configure() {
+	for name in "${THEMES[@]}"; do
+		if use theme-${name}; then
+			SFEED_THEME="${name//-/_}"
+		fi
```
ionenwks, line 28: just style but I'd `use theme-${name} &&`

ionenwks (commented):
Just a quick overview given I remembered the pkg-config bothering me when I seen the guru ebuild.

net-news/sfeed/sfeed-1.2.ebuild
```diff
@@ -12,37 +12,43 @@ SRC_URI="https://codemadness.org/releases/${PN}/${P}.tar.gz"
...
+THEMES=( mono{,-highlight} newsboat templeos )
+IUSE="+ncurses +$(printf "theme-%s " ${THEMES[@]})"
+REQUIRED_USE="ncurses? ( ^^ ( $(printf "theme-%s " ${THEMES[@]}) ) )"
+
+DEPEND="ncurses? ( sys-libs/ncurses:= )"
+RDEPEND="${DEPEND}"
+BDEPEND="virtual/pkgconfig"
+
+src_configure() {
+	for name in "${THEMES[@]}"; do
```
ionenwks, line 25: One more thing, missing `local name`

net-news/sfeed/sfeed-1.2.ebuild
N-R-K: curl is an optional runtime dep. should be opt-feature i think.

net-news/sfeed/sfeed-1.2.ebuild
N-R-K: same as curl, these are all optional runtime deps. I'll put them into opt-feature.

N-R-K:
> Just a quick overview given I remembered the pkg-config bothering me when I seen the guru ebuild.

Thanks for the review. I'm off for today though, will fix the issues tomorrow.

net-news/sfeed/sfeed-1.2.ebuild
```diff
@@ -12,29 +12,33 @@ SRC_URI="https://codemadness.org/releases/${PN}/${P}.tar.gz"
...
 
 src_compile() {
 	emake \
 		CC=$(tc-getCC) \
 		RANLIB=$(tc-getRANLIB) \
 		AR=$(tc-getAR) \
-		SFEED_CURSES=$(usex ncurses "sfeed_curses" "")
+		SFEED_CURSES=$(usex ncurses "sfeed_curses" "") \
+		SFEED_THEME="${SFEED_THEME}" \
+		SFEED_CURSES_LDFLAGS="${LDFLAGS} $(pkg-config --libs ncurses)"
```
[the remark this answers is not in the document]
N-R-K: i'll fix it tomorrow, thanks.

_NOTE to self: sfeed_curses offers a bundled minicurses library as alternative to ncurses. should probably add a useflag for that..._

N-R-K:
New version of sfeed was released just today. Mostly was documentation related change.

Any suggestion on what to do in this situation? I'm thinking about leaving the current ebuild as is and adding the new version (with the improvements) instead.

PR #24422, opened 2022-03-06
dev-libs/nss: remove prelink files

ConiKost:
Since prelink has been last-rited, removing all prelink related stuff,
which is being installed.

@gentoo/mozilla ok for merge?


juippis (approved):
Thanks!

PR #24423, opened 2022-03-06
mail-client/evolution: remove prelink files

ConiKost:
Since prelink has been last-rited, removing all prelink related stuff,
which is being installed.

@gentoo/gnome: ok for merge?


mattst88:
I wouldn't bother revbumping. The files aren't going to hurt anything if they're installed, and it's not worth causing people rebuilds.

ConiKost:
Allright, will commit without revbump.

PR #24429, opened 2022-03-07
app-crypt/tpm2-abrmd: Bump to 2.4.1 remove old

salahcoronya:
Has minor bugfixes compared to 2.4.0. 

Although tpm2-abrmd does not yet contain the problematic construction in configure.ac used in tpm2-tss which produced broken *.pc files, many of the tpm2-* projects have started using it, it could in the future, so add a similar safeguard in src_install like tpm2-tss has.


thesamesam:
Good work, thanks!

PR #24436, opened 2022-03-07
www-servers/nginx: Link with libpcre for lua module

gringus:
Signed-off-by: Marcin Deranek <marcin.deranek@slonko.net>
Closes: https://bugs.gentoo.org/832134


gringus:
Yes, I saw it. Thank you.

PR #24442, opened 2022-03-07
app-antivirus/fangfrisch: Add systemd unit files

rseichter:
Ebuild now includes a systemd service and timer.

Please note that I currently don't have access to a systemd-based Gentoo for testing the changes included in this pull request.

### DIFF
--- a/app-antivirus/fangfrisch/fangfrisch-1.5.0-r1.ebuild
+++ b/app-antivirus/fangfrisch/fangfrisch-1.5.0-r1.ebuild
@@ -1,0 +1,71 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{8,9,10} )
+
+inherit distutils-r1 readme.gentoo-r1 systemd
+
+DESCRIPTION="Update and verify unofficial Clam Anti-Virus signatures"
+HOMEPAGE="https://github.com/rseichter/fangfrisch https://pypi.org/project/fangfrisch/"
+SRC_URI="https://github.com/rseichter/fangfrisch/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+MY_CONF="/etc/${PN}.conf"
+MY_DBDIR="/var/lib/${PN}"
+DISABLE_AUTOFORMATTING=1
+DOC_CONTENTS="See https://rseichter.github.io/fangfrisch/ for the official
+documentation.
+
+### Fresh installations:
+
+Modify ${MY_CONF} according to your preferences.
+Assuming you place the database into ${MY_DBDIR}
+(recommended), execute the following commands in a root shell:
+
+mkdir -m 0770 ${MY_DBDIR}
+chgrp clamav ${MY_DBDIR}
+sudo -u clamav -- fangfrisch -c ${MY_CONF} initdb
+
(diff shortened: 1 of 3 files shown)


app-antivirus/fangfrisch/fangfrisch-1.5.0-r1.ebuild
```diff
@@ -0,0 +1,73 @@
...
+KEYWORDS="~amd64 ~x86"
+
+DEPEND=">=dev-python/requests-2.22.0[${PYTHON_USEDEP}]
+	>=dev-python/sqlalchemy-1.3.11[${PYTHON_USEDEP}]"
+RDEPEND="${DEPEND}"
+
+python_prepare_all() {
+	sed -i -e '/SQLAlchemy/d' setup.py || die
+	# Due to the nature of Fangfrisch, most tests require network
+	# connectivity. Upstream CI reports show that the tests are
```
Alessandro-Barbieri, line 48: there is `FEATURES="test_network"` for that purpose

app-antivirus/fangfrisch/fangfrisch-1.5.0-r1.ebuild
```diff
@@ -0,0 +1,73 @@
...
+	if [ -d tests ]; then
+		rm -r tests || die
+	fi
+	distutils-r1_python_prepare_all
+}
+
+python_install_all() {
+	insinto /etc
+	doins "${FILESDIR}/${PN}.conf"
+	if use systemd; then
```
Alessandro-Barbieri, line 60: no need to add an useflag, just install them unconditionally. systemd-cron is smart enough to not run the cronjob if a timer with the same name is present

app-antivirus/fangfrisch/fangfrisch-1.5.0-r1.ebuild
rseichter: Given that I am the author of Fangfrisch, I know for a fact that the tests succeed. 😉 Besides, various ClamAV signature providers require some form of license key which I won't add to the Gentoo Git repo.

app-antivirus/fangfrisch/fangfrisch-1.5.0-r1.ebuild
rseichter: I looked at existing ebuilds, and it seems to me that "if use systemd" is quite common?

ConiKost: It more common to install unconditionally init scripts. We don't want to trigger rebuilds, which only install 2 files more at the end. Every use can always filter stuff via ```INSTALL_MASK```.

rseichter: Fixed.

app-antivirus/fangfrisch/fangfrisch-1.5.0-r1.ebuild
thesamesam: Right, that's what my suggestion was going to be :)

app-antivirus/fangfrisch/fangfrisch-1.5.0-r1.ebuild
thesamesam: For reference (whole site is worth reading too): https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0301

PR #24444, opened 2022-03-07
kde-apps/konsole: Add support for plugin-ssh-manager use flag

JustArchi:
`konsole` package supports plugin called "ssh manager", which can be enabled through optional define flag.

I've made use of that define flag through newly added `plugin-ssh-manager` local use flag. I've verified through local build that the package compiles and works as expected with and without the flag - properly bundling and omitting mentioned functionality.

This is my first contribution to the gentoo project, so I apologize for eventual mistakes, I tried my best.

Thank you in advance for considering this enhancement.


JustArchi:
Added `Signed-off-by` sign according to the bot feedback :sunglasses:.

JustArchi:
I think my work here is finished in regards to all gentoo guidelines I managed to find on my own, @gentoo/kde please review and let me know if there is anything further that requires correction, thanks :sunglasses:.

JustArchi:
Linked bug report :sunglasses:

a17r:
Thanks for your contribution, however in `21.08.3` the plugin was experimental, and we will not enable it in stable for that reason.
21.12.3 is pending stabilisation, and since the plugin needs no extra dependencies, and it is default enabled in the build system, we do not have to do anyhting with the ebuild.

From 21.12.3 CMakeLists.txt:
`option(ENABLE_PLUGIN_SSHMANAGER "Build the SSHManager plugin" ON)`

JustArchi:
@a17r wouldn't it make sense to accept this MR as a way for user to disable the SSH plugin if it's deemed as unwanted? We could make it default USE flag if you'd like to, but somebody could still be interested in `-plugin-ssh-manager` USE flag to skip it.

I could change the PR to default-enable and include only in the `21.12.3` instead. What do you think?

a17r:
It does not add any dependencies. What does disabling it achieve then, save 128K disk space?

JustArchi:
:man_shrugging: alright then, thanks for reply.

PR #24453, opened 2022-03-08
media-libs/gmmlib: added new subslot

Karlson2k:
While gmmlib version 2.0.3 has not introduced new so-name, it is required to rebuild dependent packages as otherwise they are crashing.

Closes: https://bugs.gentoo.org/834634

### DIFF
--- a/media-libs/gmmlib/gmmlib-22.0.3-r1.ebuild
+++ b/media-libs/gmmlib/gmmlib-22.0.3-r1.ebuild
@@ -12,7 +12,10 @@ S="${WORKDIR}/${PN}-intel-${P}"
 
 KEYWORDS="~amd64"
 LICENSE="MIT"
-SLOT="0/12"
+# gmmlib version 22.0.3 made breaking ABI changes without changing the
+# soname, so the "_1" subslot suffix was added. There is no need to keep
+# this suffix after the next soname version bump.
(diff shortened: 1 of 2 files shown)


media-libs/gmmlib/gmmlib-22.0.3-r1.ebuild
```diff
@@ -12,7 +12,7 @@ S="${WORKDIR}/${PN}-intel-${P}"
 
 KEYWORDS="~amd64"
 LICENSE="MIT"
-SLOT="0/12"
```
thesamesam, line 15: I think we may want to do PV if they can't be trusted to bump SONAME.

Karlson2k: It's the first time when ABI changed without soname change for gmmlib. Most probably it was accidental.
It the moment I would say that this change is enough.

thesamesam: OK. Please do setup iwdevtools and enable debugging symbols and its verbose abidiff output then to make sure such issues are noticed in future where possible. Thanks!

thesamesam:
Thanks! Please do setup iwdevtools for yourself and ask for help if needed. It's very useful.

PR #24454, opened 2022-03-08
net-fs/s3fs: version bump

arkamar:
Just a simple version bump.


MatthewGentoo (approved):
Thank you for keeping s3fs up to date :)

It looks like the proxied-maintainer has been away for a few years, would you like to make yourself the maintainer?

arkamar:
I was considering that, but I don't have a personal interest/need of this package anymore. I mean, I do version bumps just because it is straight forward and I still have notifications for new releases set, but it would be better to advertise it as `maintainer-needed` for anyone who really uses this package.

PR #24456, opened 2022-03-08
sys-devel/llvm: warn when LLVM_TARGETS changes

MatthewGentoo:


MatthewGentoo:
We could have compared the targets using a loop, but I think it is shorter and more obvious to sort and then compare with `!=`.

thesamesam:
> NAK. This looks like returning back to EAPI 0, with all the horror. I'm not going to approve removing use.force unless there is a real working way to prevent bricking the toolchain.
> 
> If you want to address this, it needs to be addressed upstream. For a start, you need to fix automagic use of LLVM targets in various packages (yes, there is automagic involved).

This doesn't really help the significant number of confused users on the forums, bugs, and IRC.

1. I don't think this is that bad, given [gtk+ should do this for Wayland](https://bugs.gentoo.org/624960) (but currently does nothing, and has silent breakage)
2. If you're really against it, a compromise is to have a warn in pkg_postinst or something that it's use.force'd on, as a significant number of people don't seem to get it

But ultimately, while this isn't ideal, it's also not really within our gift to fix. I asked matthew to work on this because I'd been quite supportive of (and pushed for it) the use.force idea and I hadn't realised just how much confusion it would cause. Sometimes stuff like that happens.

MatthewGentoo:
Agreed that it's still all a bit gross, and a bit upsetting that we can't (easily) add LLVM_TARGETS to lld.

For users that unforce the USE flags and mess with them, we could have a tool that compares the symbols provided by libllvm, and the needed symbols in consumers, and rebuild to keep consistency.

```sh
$ # check LLVM_TARGETS
$ nm -D --without-symbol-versions /usr/lib/llvm/13/lib64/libLLVM-13.so |
>     awk '/LLVMInitialize.+Target$/ { print substr($3, 15) }'
AMDGPUTarget
BPFTarget
NVPTXTarget
RISCVTarget
WebAssemblyTarget
X86Target
XCoreTarget
$ # check all libraries and executables for consumers
$ nm -D --without-symbol-versions /usr/lib64/liblldELF.so.13 |
>     awk '/LLVMInitialize.+Target$/ { print substr($2, 15) }'
AMDGPUTarget
BPFTarget
NVPTXTarget
X86Target
XCoreTarget
$ # not matching -> add lld to the rebuild list
```

mgorny:
Portage clearly distinguishes use-forced flags. If people don't get it, then they should either RTFM or the docs should be improved for Portage. It's not for individual packages to teach people how to use their PM.

PR #24459, opened 2022-03-09
app-editors/{vscode,vscodium}: version bump to 1.65.1

AdelKS:
Hello @arthurzam,

A small version bump along with masking the packages for musl systems.

Adel


arthurzam (approved):
Thanks

AdelKS:
I fixed the commit message of the last commit so it contains `Bug: https://bugs.gentoo.org/834082` instead of just ` https://bugs.gentoo.org/834082` so the bug gets updated.

PR #24460, opened 2022-03-09
media-libs/amdgpu-pro-vulkan and media-video/amdgpu-pro-amf version bump

AdelKS:
Minor version bump,

Works on my machine :tm: 

Adel


juippis (approved):
```
 * 
 * 	/usr/lib64/libamfrt64.so.1.4.24: libdrm.so.2
 * 
```
so I'll add libdrm to dependency list. I wanted to have a word or two about these ebuilds but in the end they work and my differences would just be personal preferences, so not gonna bother :)

Thanks!

AdelKS:
> so I'll add libdrm to dependency list. 

I do not know how I missed that, I added that the the current and new versions.

> I wanted to have a word or two about these ebuilds but in the end they work and my differences would just be personal preferences, so not gonna bother :)

I would be happy to have your feedback and opinion on things :D

PR #24461, opened 2022-03-09
dev-util/sysdig updates

hhoffstaette:
Two fixes for sysdig:

https://bugs.gentoo.org/731268
https://bugs.gentoo.org/834664

Signed-off-by: Holger Hoffstätte <holger@applied-asynchrony.com>

### DIFF
--- a/dev-util/sysdig/files/sysdig-0.27.1-luajit-2.1.patch
+++ b/dev-util/sysdig/files/sysdig-0.27.1-luajit-2.1.patch
@@ -1,0 +1,99 @@
+
+https://bugs.gentoo.org/834664
+https://github.com/draios/sysdig/pull/921
+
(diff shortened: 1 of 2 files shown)


dev-util/sysdig/files/sysdig-0.27.1-luajit-2.1.patch
```diff
@@ -0,0 +1,93 @@
+From 1147b9a4f6f1f9e97b43735e7980da387837de4a Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: Could you put a link to the relevant Gentoo bug(s) at the top of the patch + where the commit is from?

hhoffstaette: Makes sense - done.

thesamesam:
Thank you!

PR #24464, opened 2022-03-09
sys-boot/tboot: Bump to 1.10.5

salahcoronya:
Contains 2 fixes to 1.10.4, one to fix compilation under gcc 12, one to fix inaccessible log. Also contains a patch by me to fix the PCONF element (my first fix, accepted upstream, was incomplete). The patch has been sent upstream but not yet applied.


sys-boot/tboot/files/tboot-1.10.5-fix-pconf-element.patch
```diff
@@ -0,0 +1,21 @@
+# HG changeset patch
```
thesamesam, line 1: No link to upstream submission? Can you add it in a follow-up?

salahcoronya: Upstream doesn't have a pull requestor or a bug reporter. Everything is done through e-mail. I can link to the e-mail in the archive though: https://sourceforge.net/p/tboot/mailman/message/37616841/

PR #24468, opened 2022-03-09
dev-java/slf4j* version bump

vaukai:


dev-java/slf4j-ext/slf4j-ext-1.7.36.ebuild
```diff
@@ -0,0 +1,74 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+# Skeleton command:
+# java-ebuilder --generate-ebuild --workdir . --pom pom.xml --download-uri https://github.com/qos-ch/slf4j/archive/v_1.7.36.tar.gz --slot 0 --keywords "~amd64 ~arm64 ~ppc64 ~x86" --ebuild slf4j-ext-1.7.36.ebuild
+
+EAPI=8
+
+JAVA_PKG_IUSE="doc source"
+MAVEN_ID="org.slf4j:slf4j-ext:1.7.36"
+# JAVA_TESTING_FRAMEWORKS="junit-4"
```
Leo3418, line 11: I've happened to work on fixing test errors for `slf4j-ext` with `log4j-12-api` before.  Due to obsolescence of Appenders in `log4j-12-api`, all test verifications that would rely on an Appender must be removed, leading to lower test coverage, but the tests could pass.  Feel free to grab my patch verbatim: https://github.com/6-6-6/spark-overlay/blob/master/dev-java/slf4j-ext/files/slf4j-ext-1.7.32-test-fixes-for-log4j-12-api.patch

Sadly, the upstream might never accept this patch because this is exactly *the* stubborn upstream that decided to [stick with Log4j 1.2](https://github.com/qos-ch/reload4j).  They don't seem open to moving forward with Log4j 2.

fordfrog:
merged everything except `dev-java/slf4j-ext` as that has an unresolved comment.

dev-java/slf4j-ext/slf4j-ext-1.7.36.ebuild
vaukai: > I've happened to work on fixing test errors for `slf4j-ext` with `log4j-12-api` before. Due to obsolescence of Appenders in `log4j-12-api`, all test verifications that would rely on an Appender must be removed, leading to lower test coverage, but the tests could pass. Feel free to grab my patch verbatim: https://github.com/6-6-6/spark-overlay/blob/master/dev-java/slf4j-ext/files/slf4j-ext-1.7.32-test-fixes-for-log4j-12-api.patch
> 
> Sadly, the upstream might never accept this patch because this is exactly _the_ stubborn upstream that decided to [stick with Log4j 1.2](https://github.com/qos-ch/reload4j). They don't seem open to moving forward with Log4j 2.

Sorry for replying so late.  I took me a while to make up my mind.  Now I see that `slf4j-1.7.32` uses `slf4j-log4j12` for testing while `1.7.36` switched from `slf4j-log4j12` to `slf4j-reload4j`.  On the one side I think we should follow upstream as long as possible, on the other I think it'd be very interesting following your approach switching to `log4j-12-api`.  Not now, but later after having those [appender related changes](https://logging.apache.org/log4j/2.x/changes-report.html#a2.17.2) that came with `log4j-2.17.2`.

At the moment, I think it should be ok using it (there are no consumers yet) as is.

fordfrog:
@vaukai merged, thanks!

PR #24474, opened 2022-03-10
dev-cpp/highway, media-libs/libjxl: new packages, new use flag in gimp to use libjxl

novomesk:
This PR is instead the old PR https://github.com/gentoo/gentoo/pull/23297 which I closed.


juippis (commented):
Can we get a snapshot release for libjxl? I think firefox-98 was supposed / is supporting jpeg xl.

novomesk:
Firefox nightly has support behind a flag, normal Firefox has the flag but no support.

Google Chrome has support behind a flag.

Next digiKam has JPEG XL support (try appimage from https://files.kde.org/digikam/ JXL files are opened via plug-in from kimageformats)

What exactly is the snapshot release? Is it something like unreleased libjxl0.7 version in debian?
https://www.mail-archive.com/debian-devel-changes@lists.debian.org/msg735919.html

novomesk:
@juippis I added `libjxl-0.7.0_p20220311.ebuild`

It is slightly different to live ebuild. 0.7.0_p20220311 use system's lcms2 for color management.

Live ebuild use bundled skcms. skcms provides extra features but it is not packager friendly.

media-libs/libjxl/Manifest
```diff
@@ -0,0 +1 @@
+DIST libjxl-0.7.0_p20220311.tar.gz 17323185 BLAKE2B 53b139f5774bdb335ea32911b4bb388bdf012ce0e787b1a6b57e8c4a47b19a81bfa85dccebccb034e8241057b2d98e1b2784dfa5555abba0ffbd558a2765d8fc SHA512 ffaf8c505035991c2ca2340109597c870b1599fd884f426763d00f5e6033230e8b889d6985b7054338cc1d3e13a77a92078c655833152ff0fc1ed03d30486e68
```
juippis, line 1: I think the more correct version would be `0.6.1_p...`

juippis (commented):
Can we deal with this on libjxl:
```
# Documentation generated by Doxygen
find_package(Doxygen)
if(DOXYGEN_FOUND)
set(DOXYGEN_GENERATE_HTML "YES")
set(DOXYGEN_GENERATE_XML "YES")
set(DOXYGEN_STRIP_FROM_PATH "${CMAKE_CURRENT_SOURCE_DIR}/lib/include")
set(DOXYGEN_USE_MDFILE_AS_MAINPAGE "README.md")
if(JPEGXL_WARNINGS_AS_ERRORS)
set(DOXYGEN_WARN_AS_ERROR "YES")
endif()
set(DOXYGEN_QUIET "YES")
```
seems unconditional and somewhat nasty.

media-libs/libjxl/Manifest
novomesk: When you look in `/usr/lib64/pkgconfig/libjxl.pc` there is `Version: 0.7.0`
The API is already different than in 0.6.1

juippis: What about 0.7.0_preDate. If they release 0.7.0 then the _p would make it downgrade, but _pre is lower than .0.

novomesk: OK, `_pre` is good. I'll change that. BTW according the info I have, the next released version will be 1.0. But who knows...

novomesk:
@juippis The Doxygen thing is resolved.

Hello71:
this package should have much more use flags, in particular openexr and gtest are not useful for 99% of users. also:

- excessive submodules are downloaded in -9999
- lcms is not used if skcms is enabled
- system allocator should normally be preferred over custom allocator, or at least behind use flag

please reference my overlay: https://cgit.alxu.ca/gentoo-overlay.git/tree/media-libs/libjxl/libjxl-9999.ebuild.

novomesk:
@Hello71 Thank you for your example. I am glad I am not alone interested in libjxl.

We tried similar skcms download several months ago but downloaded tarball changed checksum every day. Did you have the same problem? We tried to host the archive elsewhere so the content is unchanged.

One important thing I am missing in your ebuild:
The jxlinfo utility from examples. I use it often during development.

Those tools called viewers I keep disabled because they are not installed.

I agree with the openexr use flag. BTW I used it to create premultiplied JXL in the past.

I would leave java interface disabled, do you know someone using that?

I prefer limited but proved to work features in libjxl rather than wide spectrum of untested possibilities.

For example I like that you have multilib but I am afraid of new issues.

novomesk:
I implemented part of the feedback. You can review again.

Hello71:
> We tried similar skcms download several months ago but downloaded tarball changed checksum every day. Did you have the same problem? We tried to host the archive elsewhere so the content is unchanged.

hm, I think I misremembered that gitiles has stable archives. i think since skcms is quite small (14M skcms/.git), it should be ok to download it using git-r3? not sure.

> One important thing I am missing in your ebuild: The jxlinfo utility from examples. I use it often during development.

i flip-flopped on this a couple times. i think if jxlinfo is intended to be a standard part of libjxl installation, then it should be installed upstream. failing that, it would at least be better to patch libjxl/examples.cmake to call install function instead of using -DCMAKE_SKIP_RPATH.

> Those tools called viewers I keep disabled because they are not installed.

you are right, i don't use viewer so i forgot to check it is actually useful

> I agree with the openexr use flag. BTW I used it to create premultiplied JXL in the past.

yes, certainly there are some uses, but i think most users will not use it.

> I would leave java interface disabled, do you know someone using that?

no, but i think nobody is silly enough to USE=java globally, so i think it doesn't hurt to include it. i don't think it is a big issue either way though.

> I prefer limited but proved to work features in libjxl rather than wide spectrum of untested possibilities.
> 
> For example I like that you have multilib but I am afraid of new issues.

multilib is required for ffmpeg, which is not suitable for guru overlaying anyways. i think there is minimal risk to add it though: since cmake already uses out-of-tree builds, any breakage from adding multilib (e.g. conflicting installed files) should only occur when multilib is actually turned on.

novomesk:
I added **multilib** support.

juippis (approved):
Let's try it out and see what happens. Thanks a lot!

novomesk:
@a17r Andreas, when you update `kde-frameworks/kimageformats` can you also add use flag **jpegxl** ? So it will be the same as in gimp?

Hello71:
something I just noticed, why did you only add jpegxl to gimp-9999 and not 2.99.10?

novomesk:
> something I just noticed, why did you only add jpegxl to gimp-9999 and not 2.99.10?

You are right, it should be added to `gimp-2.99.10.ebuild` as well, otherwise it works in automagic mode (which Gentoo developers don't like).

band-a-prend:
@Hello71 
gimp-9999 depends now on >=gexiv2-0.14.0 so it couldn't be build currently in gentoo.

USE=jpegxl will be added to gimp-2.99.10 explicitly on revision bump.

novomesk:
Off topic: I don't understand what is the problem with gexiv2 0.14. I compiled/installed it manually and gimp-9999 works with that.

thesamesam:
> Off topic: I don't understand what is the problem with gexiv2 0.14. I compiled/installed it manually and gimp-9999 works with that.

https://bugs.gentoo.org/835781#c1

novomesk:
I saw the comments in the bug report. I don't understand those python issues.
I use gexiv2 in GIMP as C library, that's the core/essential functionality which works.

thesamesam:
> I saw the comments in the bug report. I don't understand those python issues.
> I use gexiv2 in GIMP as C library, that's the core/essential functionality which works.

Yes, but other packages need the Python support.

juippis:
> (which Gentoo developers don't like).

It's a user experience issue :) you might depclean the lib when nothing depends on it, but it's been automagically and undetectably enabled on some program now most likely breaking it.

band-a-prend:
Updated revision of media-gfx/gimp-2.99.10-r1 with USE="jpegxl": https://github.com/gentoo/gentoo/pull/24765
Checked that some jpegxl samples are opened now by gimp-2.99.10 compiled with this option enabled.

novomesk:
@band-a-prend Sergey, there is some new problem with the live GIMP ebuild:
```
gimpunit_pdb.c:25:10: fatal error: stamp-pdbgen.h: No such file or directory
   25 | #include "stamp-pdbgen.h"
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [Makefile:2090: gimpunit_pdb.lo] Error 1
make[3]: Leaving directory '/var/tmp/portage/media-gfx/gimp-9999/work/gimp-9999/libgimp'
make[2]: *** [Makefile:1670: all] Error 2
make[2]: Leaving directory '/var/tmp/portage/media-gfx/gimp-9999/work/gimp-9999/libgimp'
make[1]: *** [Makefile:876: all-recursive] Error 1
make[1]: Leaving directory '/var/tmp/portage/media-gfx/gimp-9999/work/gimp-9999'
make: *** [Makefile:777: all] Error 2
 * ERROR: media-gfx/gimp-9999::gentoo failed (compile phase):
 *   emake failed
```
but if I try to build it myself, both via autotools or meson works with a little patching.

PR #24478, opened 2022-03-10
dev-php/pecl-mongodb: Bump to 1.12.1 (with dev-libs/libbson and dev-libs/mongo-c-driver)

ej-agas:
Also with
`dev-libs/libbson` to `1.21.1`
`dev-libs/mongo-c-driver` to `1.21.1`


thesamesam:
Could you rebase?

thesamesam:
@hydrapolic @jkroonza could one of you rebase it please? (also if one of you could handle dev-php/pecl-stomp asap, i'd very much appreciate it)

jkroonza:
@thesamesam I'm off the opinion that libbson should be part of the mongo-c-driver package as a USE flag.

They both download from the same source archive, and mongo-c-driver depend on dev-libs/libbson (which comes from the same sources).

Looks liks bson may be outright required, so just drop libbson and -DENABLE_BSON in mongo-c-driver instead?

The only dependencies on libbson is mongo-c-driver and pecl-mongodb.

So *in my opinion* we should last-rite dev-libs/libbson indicating to depend on mongo-c-driver intstead, and then axe it.

And yea, I'm making progress on this, but need to pause here waiting for feedback :).

thesamesam:
> @thesamesam I'm off the opinion that libbson should be part of the mongo-c-driver package as a USE flag.
> 

This sounds reasonable _but_ I'm in a hurry to stable fixed versions of these because OpenSSL 1.1.x is EOL in a few days. Let's maybe a file a bug laying this out but keep the status quo for now?

> [...]
> Looks liks bson may be outright required, so just drop libbson and -DENABLE_BSON in mongo-c-driver instead?
>  [...]

If you do this, does mongo-c-driver install libbson, or does it just statically link to it?

jkroonza:
OK, will keep status quo for now.  Expect an updated PR by close-of-business.

jkroonza:
CMake Error at CMakeLists.txt:28 (message):
  ENABLE_BSON=SYSTEM is no longer supported.  Use USE_SYSTEM_LIBBSON=ON for
  the same effect.

Right, so we have no choice.  Can we get exemption from the normal last-rite process?  Or do I push a libbson ebuild that just depends on mongo-c-driver with same version?  And mongo-c-driver depends on !<libbson-${PV} ?

PR #24482, opened 2022-03-11
media-gfx/hydrus: Fix building without the doc flag

vaartis:
A problem with the fact that this directory only exists when building documentation, so check for that and only remove it and symlink the documentation where needed if it's been built.


juippis (approved):
Eww, looking at this hackery, at what point we're making the docs mandatory instead? :)

vaartis:
> Eww, looking at this hackery, at what point we're making the docs mandatory instead? :)

I mean, it doesn't look good, but the making them mandatory is hardly necessary, they're all published online. Maybe I should try to only _include_ things when installing, instead of _exluding_ those that are unneccessary. That should clean things up.

PR #24483, opened 2022-03-11
dev-java/cortado: Restrict to jdk:1.8

vaukai:
Bug: https://bugs.gentoo.org/833397

### DIFF
--- a/dev-java/cortado/cortado-0.6.0-r2.ebuild
+++ b/dev-java/cortado/cortado-0.6.0-r2.ebuild
@@ -1,0 +1,37 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit java-pkg-2 java-ant-2
+
+DESCRIPTION="Multimedia framework for Java written by Fluendo"
+HOMEPAGE="https://www.theora.org/cortado/"
+SRC_URI="https://downloads.xiph.org/releases/cortado/${P}.tar.gz"
+
+LICENSE="GPL-2 LGPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc64 ~x86"
+
+# Package 'sun.audio' seems to have moved to 'javax.sound'.
+# More work would be needed.  Presently we restrict to jdk:1.8
+DEPEND="virtual/jdk:1.8"
+RDEPEND=">=virtual/jre-1.8:*"
+
+EANT_BUILD_TARGET="stripped"
+
+src_prepare() {
+	default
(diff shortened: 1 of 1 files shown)


vaukai:
@Flowdalic

dev-java/cortado/cortado-0.6.0-r2.ebuild
```diff
@@ -0,0 +1,60 @@
...
+
+DESCRIPTION="Multimedia framework for Java written by Fluendo"
+HOMEPAGE="https://www.theora.org/cortado/"
+SRC_URI="https://downloads.xiph.org/releases/cortado/${P}.tar.gz"
+
+LICENSE="GPL-2 LGPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc64 ~x86"
+
+# To prevent following errors we restrict to jdk:1.8
```
Flowdalic, line 16: What follows seems overly long. The gist is that the packages uses `sun.audio` which is not available in newer Java versions.

dev-java/cortado/cortado-0.6.0-r2.ebuild
```diff
@@ -0,0 +1,60 @@
...
+#     [javac]       AudioPlayer.player.stop(stream);
+#     [javac]                  ^
+DEPEND="virtual/jdk:1.8"
+RDEPEND=">=virtual/jre-1.8:*"
+
+EANT_BUILD_TARGET="stripped"
+
+src_prepare() {
+	default
+	echo "#!/bin/sh" > scripts/get-revision || die
```
Flowdalic, line 49: This should use a "here document", instead of invoking `echo` twice. See https://devmanual.gentoo.org/tools-reference/cat/#here-documents

dev-java/cortado/cortado-0.6.0-r2.ebuild
```diff
@@ -0,0 +1,60 @@
...
+	echo "echo ${PV}" >> scripts/get-revision || die
+
+#	Package 'sun.audio' seems to have moved to 'javax.sound'. But more word is needed.
```
Flowdalic, line 52: typo: s/word/work/

Flowdalic (changes requested):
As always, thanks for your contribution. I have only minor remarks to improve the ebuild.

dev-java/cortado/cortado-0.6.0-r2.ebuild
vaukai: Done

dev-java/cortado/cortado-0.6.0-r2.ebuild
vaukai: Done.

PR #24486, opened 2022-03-11
bump to 0.17.4

heeplr:
* update default config
* in cron script:
  * replace tempfile with mktemp (since debianutils 5.x dropped it) 
  * fix shellcheck issues
  * minor cleanup


thesamesam:
- Can you prefix each commit with `cat/pkg: ...` (app-forensics/aide)?
- We may want to squash all commits into one or into logical groupings ("cleanup init script"), etc
- Please include the GCO signoff in commits (`Signed-off-by: real name <email>`)
- We should do these changes in a new revision to avoid affecting existing installs, so copy each of the files in `files/` you want to modify first, then modify them, then we will copy the ebuild too and reference the new files in there

heeplr:
@thesamesam I've applied the changes according to your suggestions

thesamesam (approved):
Great work! I revbumped aide to make sure these changes are propagated.

PR #24489, opened 2022-03-11
dev-db/mysql-connector-c++: drop 1.1.12-r1

thesamesam:


thesamesam:
Missed https://bugs.gentoo.org/834439.

PR #24496, opened 2022-03-11
www-apps/gitea: security bump to 1.16.3

hydrapolic:
Tested on amd64, thanks!


MatthewGentoo (approved):
Thanks!

I have amended the commit message to tag https://bugs.gentoo.org/834982

hydrapolic:
Thanks @MatthewGentoo, however 1.16.3 fixes a different security bug as https://bugs.gentoo.org/834982. A fix for that hasn't been released yet.

PR #24501, opened 2022-03-12
Mask proton-ge packages in multilib profiles

CHN-beta:


thesamesam:
We don't normally mask stuff in ::gentoo for overlays, but I appreciate there's some precedent of doing this for Steam and stuff.

@gentoo/games?

chewi:
I'm pretty sure package.mask works in overlays. Why do we need this here?

ionenwks:
> I'm pretty sure package.mask works in overlays. Why do we need this here?

Never experimented but pretty sure can't for profile-specific masks -- not that I'm fond at the idea of adding them because of it.

If this is mostly due to pkgcheck noise, you can enable/disable checking only specific profiles.

chewi:
> If this is mostly due to pkgcheck noise, you can enable/disable checking only specific profiles.

It must be. Running Steam on non-multilib is a total non-starter. Even if Steam does go 64-bit, there are still plenty of games that aren't.

PR #24502, opened 2022-03-12
profiles/base: Remove mate-base/mate[bluetooth] stable USE mask

Leo3418:
The only dependency of this USE flag (net-wireless/blueman) has been stabilized on every architecture where mate-base/mate has a stable version, so this USE stable mask is obsolete.


jsmolic:
Good catch, thank you

PR #24504, opened 2022-03-12
media-gfx/freecad: depend on boost-1.77.0

Alessandro-Barbieri:
Closes: https://bugs.gentoo.org/834991


waebbl:
I don't think it's a good solution to restrict to 1.77.0 on a global scale. There seems to be no general issue with 1.78.0. Others and myself can successfully build with this version of boost.

a17r:
NACK.

thesamesam:
As others have said, no, we can't do this. If nothing else, huge rebuilds to downgrade. We need to investigate the bug properly.

PR #24506, opened 2022-03-12
media-video/mplayer: add add riscv support & ~riscv keywording

Rabenda:
media-sound/toolame app-i18n/enca media-libs/libggiwmh  ~riscv keywording 
these make repoman happy in media-video/mplayer


media-video/mplayer/mplayer-1.4-r1.ebuild
```diff
@@ -234,6 +234,8 @@ src_prepare() {
 				"${FILESDIR}"/ffmpeg44.patch
 		has_version '>=media-video/ffmpeg-5' && eapply \
 				"${WORKDIR}"/${P}-ffmpeg5.patch
+		# Patch the configure script for the RISC-V architecture
+		eapply "${FILESDIR}/add-configure-riscv-support.patch" # bug 812855
```
dlan17, line 238: I'd suggest puttng the comments into the patch file, same with bug id, so make the ebuild simple and clean

also, I'd think it should also work with : PATCHES=(  "${FILESDIR}/add-configure-riscv-support.patch" ) ? but it's kind of different taste of style, so I'm fine with this

media-video/mplayer/files/add-configure-riscv-support.patch
```diff
@@ -0,0 +1,10 @@
+--- MPlayer-1.4/configure       2018-12-15 16:38:24.000000000 +0100
++++ MPlayer-1.4-mod/configure   2021-06-19 20:41:00.735402324 +0200
+@@ -1761,6 +1761,7 @@
+       nios2) host_arch=nios2 ;;
+       vax) host_arch=vax ;;
+       xtensa*) host_arch=xtensa ;;
++      riscv*) host_arch=generic ;;
+       *) host_arch=UNKNOWN ;;
+   esac
+   }
```
dlan17, line 10: I'd suggest split this commit into two
1) add patch to support riscv (including change to ebuild)
2) keyword the package
usually we'd suggest to open a bug for adding patches in order to gain maintainer's review or ACK, but in this case, it's perfect to re-use the same bug id

also we'd suggest to use git format patch (adding commit message info if necessary), but since this patch is rather straightforward, so I'm fine with no commit message.

Rabenda: Done. I split it.

media-video/mplayer/mplayer-1.4-r1.ebuild
Rabenda: use PATCHES= in new

PR #24507, opened 2022-03-12
media-fonts/lxgw-wenkai: v1.200 -> v1.222

rtgiskard:
Signed-off-by: Zheng Zhu <rtgiskard@gmail.com>


MatthewGentoo (changes requested):
According to [the commit policy](https://wiki.gentoo.org/wiki/Gentoo_git_workflow#Commit_policy) you should make atomic commits, that is, each commit makes one logical change.

In this case, you should add the new version (1.222), then commit again to drop the old version (1.200).

You can force push and re-use this pull request, please comment or join [#gentoo-proxy-maint](https://www.gentoo.org/get-involved/irc-channels/) on IRC if you need any help with git :)

rtgiskard:
Wrel

> According to [the commit policy](https://wiki.gentoo.org/wiki/Gentoo_git_workflow#Commit_policy) you should make atomic commits, that is, each commit makes one logical change.
> 
> In this case, you should add the new version (1.222), then commit again to drop the old version (1.200).
> 
> You can force push and re-use this pull request, please comment or join [#gentoo-proxy-maint](https://www.gentoo.org/get-involved/irc-channels/) on IRC if you need any help with git :)

I see ..  👀

MatthewGentoo (approved):
Great! Thank you for fixing that.

PR #24508, opened 2022-03-12
dev-qt/qt3d: ~riscv keyworing

Rabenda:
dev-qt/qtgamepad: ~riscv keywording
Make repoman happy in dev-qt/qt3d

### DIFF
--- a/dev-qt/qt3d/qt3d-5.15.3.ebuild
+++ b/dev-qt/qt3d/qt3d-5.15.3.ebuild
@@ -9,7 +9,7 @@ inherit qt5-build
 DESCRIPTION="3D rendering module for the Qt5 framework"
 
 if [[ ${QT5_BUILD_TYPE} == release ]]; then
-	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
+	KEYWORDS="~amd64 ~arm ~arm64 ~riscv ~x86"
 fi
 
 # TODO: tools
--- a/dev-qt/qtgamepad/qtgamepad-5.15.3.ebuild
+++ b/dev-qt/qtgamepad/qtgamepad-5.15.3.ebuild
@@ -8,7 +8,7 @@ inherit qt5-build
 DESCRIPTION="Qt module to support gamepad hardware"
 
 if [[ ${QT5_BUILD_TYPE} == release ]]; then
-	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
+	KEYWORDS="~amd64 ~arm ~arm64 ~riscv ~x86"
 fi
 
 IUSE="evdev qml sdl"


dev-qt/qt3d/qt3d-5.15.2-r11.ebuild
```diff
@@ -28,6 +28,10 @@ DEPEND="${RDEPEND}
 	vulkan? ( dev-util/vulkan-headers )
 "
 
+PATCHES=(
+	"${FILESDIR}/${P}-riscv-atomic.patch" #bug 803470
```
arthurzam, line 32: It looks like it would be a "long time" downstream patch (which is thankfully quite easy to maintain).
In that case, I think it would be better to use the `${PN}-5.15.2` instead of `${P}`, as it would make bumping easier in future.

arthurzam (commented):
I'm not sure why I was called, as I'm not maintainer from Qt project, my here are my two cents from this simple PR.
But I do think more input from someone from Qt project is a good idea.

dev-qt/qt3d/qt3d-5.15.2-r11.ebuild
dlan17: oops, @arthurzam sorry for highlighting you.. I probably should at pesa
basically, I think it's a quite trivial patch, should be easy to maintain

anyway, thanks for the review/comments

Pesa: agree with @arthurzam's comment

Pesa (changes requested):
small typo in the 2nd commit message

dev-qt/qt3d/files/qt3d-5.15.2-riscv-atomic.patch
```diff
@@ -0,0 +1,10 @@
+diff --git a/src/plugins/renderers/opengl/opengl.pri b/src/plugins/renderers/opengl/opengl.pri
+index 1682ab7..e0ca986 100644
+--- a/src/plugins/renderers/opengl/opengl.pri
++++ b/src/plugins/renderers/opengl/opengl.pri
+@@ -39,3 +39,5 @@ qtConfig(qt3d-simd-sse2):!qtConfig(qt3d-simd-avx2) {
+     CONFIG += simd
+     QMAKE_CXXFLAGS += $$QMAKE_CFLAGS_SSE2
+ }
++
++gcc:isEqual(QT_ARCH, "riscv64"):qtConfig(std-atomic64): LIBS += -latomic
```
Rabenda, line 10: It seems wrong @dlan17 
qtConfig(std-atomic64) has 64bit atomic. But it should determine !qtConfig(std-atomic64) to use latomic.
Original error is `__atomic_exchange_1` missing. ATOMIC_NON64BIT is needed

Rabenda: https://doc.qt.io/qt-5/qmake-test-function-reference.html#qtcompiletest-test

qtcompiletest can resolve it

Rabenda: In use qtcompiletest case, it needs two test-qt-projects in patch. It looks not graceful

dev-qt/qt3d/qt3d-5.15.2-r11.ebuild
Rabenda: Now use ${PN}-5.15.2

dev-qt/qt3d/files/qt3d-5.15.2-riscv-atomic.patch
Rabenda: New patch solve it.

dev-qt/qt3d/files/qt3d-5.15.2-riscv-atomic.patch
```diff
@@ -0,0 +1,65 @@
...
+diff --git a/src/plugins/renderers/opengl/opengl.pri b/src/plugins/renderers/opengl/opengl.pri
+index 1682ab7..a92220c 100644
+--- a/src/plugins/renderers/opengl/opengl.pri
++++ b/src/plugins/renderers/opengl/opengl.pri
+@@ -39,3 +39,5 @@ qtConfig(qt3d-simd-sse2):!qtConfig(qt3d-simd-avx2) {
+     CONFIG += simd
+     QMAKE_CXXFLAGS += $$QMAKE_CFLAGS_SSE2
+ }
++
++gcc:isEqual(QT_ARCH, "riscv64"):qtConfig(need-libatomic): LIBS += -latomic
```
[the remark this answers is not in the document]
Rabenda: Ok, I will remove it.

Pesa:
> Please test https://invent.kde.org/qt/qt/qtbase/-/merge_requests/138 applied against qtcore alternatively.

And please @Rabenda test the above patch first. If we can solve this properly in qtcore, we will no longer have to patch a bunch of other packages.
That test is for function pointers so I'm not convinced the patch will work for 1- and 2-byte atomics (which is the problem on riscv), but it's worth trying it, and maybe we can adapt the test to check for more variations, like you did in your latest patch?

Rabenda:
https://github.com/riscv-collab/riscv-gcc/issues/12 use Option 4
https://stackoverflow.com/questions/23250863/difference-between-pthread-and-lpthread-while-compiling
In riscv, "-pthread" equivalent to add link option "--push-state --as-needed -latomic --pop-state"

Rabenda:
https://github.com/felixonmars/archriscv-packages/blob/master/qt5-base/riscv64.patch

Archlinux riscv fix method

a17r:
It would be preferable to make that a patch instead, so it could theoretically at least be upstreamed.

dev-qt/qtcore/files/qtcore-5.15.2-atomic.patch
```diff
@@ -0,0 +1,41 @@
...
+         "cxx11_random": {
+@@ -466,7 +466,7 @@
+                     "shm_open(\"test\", O_RDWR | O_CREAT | O_EXCL, 0666);",
+                     "shm_unlink(\"test\");"
+                 ],
+-                "qmake": "linux: LIBS += -lpthread -lrt"
++                "qmake": "linux: LIBS += -pthread -lrt"
+             }
+         },
+         "linkat": {
```
dlan17, line 41: hi @Pesa @a17r , it occur to me that current '-pthread' version is much better (less invasive), what do you think?

@Rabenda can you also report this to qt upstream?

dlan17:
hi all, please hold this for a while, we will try to rebase the patch to qt-5.15.3 (as @a17r suggested) and re-test

PR #24509, opened 2022-03-12
app-crypt/libsecret: Create an emulated TPM2 for tests

salahcoronya:
Libsecret runs tests against the TPM2 already in the machine and require tpm2-abrmd. Hence the tests will fail if the user does not have tpm2-abrmd installed or does not have a TPM2 in the machine. It shouldn't do this - it should provision a virtual TPM2 spawn an emulator (swtpm) and the use the swtpm TCTI with tpm2-abrmd. This isn't the first time I've encountered a package that assumed the TPM2 simulator environment was setup prior. 

However its not too difficult to setup the TPM2 simulator for the test, which is what this patch does. Note that neither swtpm, tpm2-abrmd, nor >=tpm2-tss-3.2.0 is currently stable however, and tpm2-abrmd doesn't have riscv keywords. Its possible to remove the dependency on >=tpm2-tss-3.2.0 by using TCP ports rather than domain sockets, if desired.  The problem with that the swtpm might fail if the port is already in use.

### DIFF
--- a/app-crypt/libsecret/libsecret-0.20.5-r1.ebuild
+++ b/app-crypt/libsecret/libsecret-0.20.5-r1.ebuild
@@ -1,0 +1,141 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+PYTHON_COMPAT=( python3_{8..10} )
+VALA_USE_DEPEND=vapigen
+
+inherit bash-completion-r1 gnome2 meson-multilib python-any-r1 vala virtualx
+
+DESCRIPTION="GObject library for accessing the freedesktop.org Secret Service API"
+HOMEPAGE="https://wiki.gnome.org/Projects/Libsecret"
+
+LICENSE="LGPL-2.1+ Apache-2.0" # Apache-2.0 license is used for tests only
+SLOT="0"
+
+IUSE="+crypt gtk-doc +introspection test tpm +vala"
+RESTRICT="!test? ( test )"
+REQUIRED_USE="
+	vala? ( introspection )
+	gtk-doc? ( crypt )
+"
+
+KEYWORDS="~alpha amd64 arm arm64 ~ia64 ~mips ppc ppc64 ~riscv sparc x86"
+
+DEPEND="
+	>=dev-libs/glib-2.44:2[${MULTILIB_USEDEP}]
+	crypt? ( >=dev-libs/libgcrypt-1.2.2:0=[${MULTILIB_USEDEP}] )
+	tpm? ( >=app-crypt/tpm2-tss-3.0.3 )
+	introspection? ( >=dev-libs/gobject-introspection-1.54:= )
+"
+RDEPEND="${DEPEND}
+	virtual/secret-service"
+BDEPEND="
+	app-text/docbook-xml-dtd:4.2
+	dev-libs/libxslt
+	dev-util/gdbus-codegen
+	dev-util/glib-utils
+	>=sys-devel/gettext-0.19.8
+	virtual/pkgconfig
+	gtk-doc? (
+		app-text/docbook-xml-dtd:4.1.2
+		>=dev-util/gi-docgen-2021.7
+	)
+	test? (
+		$(python_gen_any_dep '
+			dev-python/mock[${PYTHON_USEDEP}]
+			dev-python/dbus-python[${PYTHON_USEDEP}]
+			introspection? ( dev-python/pygobject:3[${PYTHON_USEDEP}] )')
+		introspection? ( >=dev-libs/gjs-1.32 )
+		tpm? (	app-crypt/swtpm
+			app-crypt/tpm2-abrmd
+			>=app-crypt/tpm2-tss-3.2.0
+		)
+	)
+	vala? ( $(vala_depend) )
+"
+
+tpm2_run_with_emulator() {
+#	eval $(dbus-launch --sh-syntax) || die
+	export XDG_CONFIG_HOME=${T}/.config/swtpm
+	${BROOT}/usr/share/swtpm/swtpm-create-user-config-files || die
+
+	mkdir -p ${XDG_CONFIG_HOME}/mytpm1 || die
+	swtpm_setup_args=(
+		--tpm2
+		--tpmstate ${XDG_CONFIG_HOME}/mytpm1
+		--createek
+		--allow-signing
+		--decryption
+		--create-ek-cert
+		--create-platform-cert
+		--lock-nvram
+		--overwrite
+		--display
+	)
+	swtpm_setup "${swtpm_setup_args[@]}" || die
+
+	swtpm_socket_args=(
+		--tpm2
+		--tpmstate dir=${XDG_CONFIG_HOME}/mytpm1
+		--flags startup-clear
+		--ctrl type=unixio,path=${XDG_CONFIG_HOME}/mytpm1/swtpm.socket.ctrl
+		--server type=unixio,path=${XDG_CONFIG_HOME}/mytpm1/swtpm.socket
+		--pid file=${XDG_CONFIG_HOME}/mytpm1/swtpm.pid
+		--daemon
+	)
+	swtpm socket "${swtpm_socket_args[@]}" || die
+
+	tpm2_abrmd_args=(
+		--logger=stdout
+		--tcti=swtpm:path=${XDG_CONFIG_HOME}/mytpm1/swtpm.socket
+		--session
+		--flush-all
+	)
+	tpm2-abrmd "${tpm2_abrmd_args[@]}" &
+	export TCTI=tabrmd:bus_type=session
+
+	$1 $@ || die
+
+	# When swtpm dies, tmp2-abrmd will exit
+	kill $(< ${XDG_CONFIG_HOME}/mytpm1/swtpm.pid) || die
+}
+
+python_check_deps() {
+	if use introspection; then
+		has_version -b "dev-python/pygobject:3[${PYTHON_USEDEP}]" || return
+	fi
+	has_version -b "dev-python/mock[${PYTHON_USEDEP}]" &&
+	has_version -b "dev-python/dbus-python[${PYTHON_USEDEP}]"
+}
+
+pkg_setup() {
+	use test && python-any-r1_pkg_setup
+}
+
+src_prepare() {
+	use vala && vala_src_prepare
+	default
+}
+
+multilib_src_configure() {
+	local emesonargs=(
+		$(meson_native_true manpage)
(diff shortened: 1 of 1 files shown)


app-crypt/libsecret/libsecret-0.20.5.ebuild
```diff
@@ -47,10 +47,28 @@ BDEPEND="
...
+		tpm? (	app-crypt/swtpm
+			app-crypt/tpm2-abrmd
+			>=app-crypt/tpm2-tss-3.2.0 )
 	)
 	vala? ( $(vala_depend) )
 "
 
+tpm2_run_with_emulator() {
+	eval `dbus-launch --sh-syntax`
+	export XDG_CONFIG_HOME=${WORKDIR}/.config/swtpm
```
thesamesam, line 59: Use ${T}

app-crypt/libsecret/libsecret-0.20.5.ebuild
```diff
@@ -47,10 +47,28 @@ BDEPEND="
...
 		introspection? ( >=dev-libs/gjs-1.32 )
+		tpm? (	app-crypt/swtpm
+			app-crypt/tpm2-abrmd
+			>=app-crypt/tpm2-tss-3.2.0 )
 	)
 	vala? ( $(vala_depend) )
 "
 
+tpm2_run_with_emulator() {
+	eval `dbus-launch --sh-syntax`
```
thesamesam, line 58: $()

app-crypt/libsecret/libsecret-0.20.5.ebuild
```diff
@@ -47,10 +47,28 @@ BDEPEND="
...
+	export XDG_CONFIG_HOME=${WORKDIR}/.config/swtpm
+	/usr/share/swtpm/swtpm-create-user-config-files
```
thesamesam, line 60: BROOT or ESYSROOT, I guess the latter based on DEPEND, not that it really matters for tests anyway if needed in both

app-crypt/libsecret/libsecret-0.20.5.ebuild
```diff
@@ -47,10 +47,28 @@ BDEPEND="
...
+	mkdir -p ${XDG_CONFIG_HOME}/mytpm1
+	swtpm_setup --tpm2 --tpmstate ${XDG_CONFIG_HOME}/mytpm1 --createek --allow-signing --decryption --create-ek-cert --create-platform-cert --lock-nvram --overwrite --display
+	swtpm socket --tpm2 --tpmstate dir=${XDG_CONFIG_HOME}/mytpm1 --flags startup-clear --ctrl type=unixio,path=${XDG_CONFIG_HOME}/mytpm1/swtpm.socket.ctrl  --server type=unixio,path=${XDG_CONFIG_HOME}/mytpm1/swtpm.socket --pid file=${XDG_CONFIG_HOME}/mytpm1/swtpm.pid --daemon
+	tpm2-abrmd --logger=stdout --tcti=swtpm:path=${XDG_CONFIG_HOME}/mytpm1/swtpm.socket --session --flush-all &
```
thesamesam, line 64: Lots of missing || dies?

app-crypt/libsecret/libsecret-0.20.5.ebuild
```diff
@@ -47,10 +47,28 @@ BDEPEND="
...
+	export TCTI=tabrmd:bus_type=session
+
+	$1 $@
```
thesamesam, line 67: || die

thesamesam (commented):
Thanks! Please also put the long (good!) description from the PR in the commit message. Also, why --force?

mattst88:
Thank you. This looks really good. A few comments:

- I'd prefer to revbump the ebuild with this change since 0.20.5 is being stabilized  now
- I fixed bug 662376 as part of my stabilization testing, so some minor rebasing is needed
- I think we can make the long strings of arguments in `tpm2_run_with_emulator` more readable by putting them in bash arrays
- We're still running tests under `tpm2_run_with_emulator` if `USE=-tpm`

I've made these changes. Here's a diff. What do you think? If you're okay with them, I'll apply this under your authorship.

My only remaining question is about `tpm2-abrmd "${tpm2_abrmd_args[@]}" &`. What stops that process?

````diff
--- libsecret-0.20.5.ebuild	2022-03-15 13:01:15.872544242 -0700
+++ libsecret-0.20.5-r1.ebuild	2022-03-15 13:12:27.050707498 -0700
@@ -47,10 +47,59 @@
 			dev-python/dbus-python[${PYTHON_USEDEP}]
 			introspection? ( dev-python/pygobject:3[${PYTHON_USEDEP}] )')
 		introspection? ( >=dev-libs/gjs-1.32 )
+		tpm? (
+			app-crypt/swtpm
+			app-crypt/tpm2-abrmd
+			>=app-crypt/tpm2-tss-3.2.0
+		)
 	)
 	vala? ( $(vala_depend) )
 "
 
+tpm2_run_with_emulator() {
+	export XDG_CONFIG_HOME=${T}/.config/swtpm
+	${BROOT}/usr/share/swtpm/swtpm-create-user-config-files || die
+
+	mkdir -p ${XDG_CONFIG_HOME}/mytpm1 || die
+	swtpm_setup_args=(
+		--tpm2
+		--tpmstate ${XDG_CONFIG_HOME}/mytpm1
+		--createek
+		--allow-signing
+		--decryption
+		--create-ek-cert
+		--create-platform-cert
+		--lock-nvram
+		--overwrite
+		--display
+	)
+	swtpm_setup "${swtpm_setup_args[@]}" || die
+
+	swtpm_socket_args=(
+		--tpm2
+		--tpmstate dir=${XDG_CONFIG_HOME}/mytpm1
+		--flags startup-clear
+		--ctrl type=unixio,path=${XDG_CONFIG_HOME}/mytpm1/swtpm.socket.ctrl
+		--server type=unixio,path=${XDG_CONFIG_HOME}/mytpm1/swtpm.socket
+		--pid file=${XDG_CONFIG_HOME}/mytpm1/swtpm.pid
+		--daemon
+	)
+	swtpm socket "${swtpm_socket_args[@]}" || die
+
+	tpm2_abrmd_args=(
+		--logger=stdout
+		--tcti=swtpm:path=${XDG_CONFIG_HOME}/mytpm1/swtpm.socket
+		--session
+		--flush-all
+	)
+	tpm2-abrmd "${tpm2_abrmd_args[@]}" &
+	export TCTI=tabrmd:bus_type=session
+
+	$1 $@ || die
+
+	kill $(< ${XDG_CONFIG_HOME}/mytpm1/swtpm.pid) || die
+}
+
 python_check_deps() {
 	if use introspection; then
 		has_version -b "dev-python/pygobject:3[${PYTHON_USEDEP}]" || return
@@ -83,5 +132,9 @@
 }
 
 multilib_src_test() {
-	virtx dbus-run-session meson test -C "${BUILD_DIR}"
+	if use tpm; then
+		tpm2_run_with_emulator virtx dbus-run-session meson test -C "${BUILD_DIR}"
+	else
+		virtx dbus-run-session meson test -C "${BUILD_DIR}"
+	fi
 }
````

**Comment by mattst88** [comment]
Well, shit.

With my changes, I get this:

````
** (tpm2-abrmd:69): WARNING **: 13:38:07.678: Failed to get proxy for DBus daemon (org.freedesktop.DBus): Cannot autolaunch D-Bus without X11 $DISPLAY

** (tpm2-abrmd:69): CRITICAL **: 13:38:07.678: Failed to acquire DBus name com.intel.tss2.Tabrmd. UID 250 must be allowed to "own" this name. Check DBus config and check that this is running as user tss or root.
````

I guess something about the TPM setup needs `eval $(dbus-launch --sh-syntax) || die`?

**Comment by mattst88** [comment]
Something else strange. Without your patch, only one Xvfb is started:

````
 * Scanning for an open DISPLAY to start Xvfb ...
 * Starting Xvfb on $DISPLAY=5 ...
````

But with it, two are started:

````
 * Scanning for an open DISPLAY to start Xvfb ...
 * Starting Xvfb on $DISPLAY=5 ...
 * Scanning for an open DISPLAY to start Xvfb ...
 * Starting Xvfb on $DISPLAY=6 ...
````

This happens even without using `dbus-run-session`.

**Comment by mattst88** [comment]
> Well, shit.
> 
> With my changes, I get this:
> 
> ```
> ** (tpm2-abrmd:69): WARNING **: 13:38:07.678: Failed to get proxy for DBus daemon (org.freedesktop.DBus): Cannot autolaunch D-Bus without X11 $DISPLAY
> 
> ** (tpm2-abrmd:69): CRITICAL **: 13:38:07.678: Failed to acquire DBus name com.intel.tss2.Tabrmd. UID 250 must be allowed to "own" this name. Check DBus config and check that this is running as user tss or root.
> ```
> 
> I guess something about the TPM setup needs `eval $(dbus-launch --sh-syntax) || die`?

If so, I suspect we'll want something like how it's done here: https://gitweb.gentoo.org/repo/gentoo.git/plain/xfce-base/xfconf/xfconf-4.16.0.ebuild

**Comment by salahcoronya** [comment]
I'm OK with the cleanup. when swtpm is killed, tpm2-abrmd will exit on its own since the socket will be closed. Note that tpm2-abrmd is a dbus app, and it needs a bus too. The problem, however, is the both tpm2-abrmd and meson need to share the same session bus. Since I want tpm2_run_with_emulator() to be self-contained (like virtx), I kept the eval $(dbus-launch --sh-syntax) || die. Of course this pollutes the ebuild environment.  The then run-dbus-session can (and must) be skipped in the tpm path, since we'll already have a session bus.

**Comment by mattst88** [comment]
Thanks.

Any idea why (or how?) your patch causes multiple Xvfbs?

**Comment by salahcoronya** [comment]
Yeah its that $1 $@ thing. It just need to be $@ to get the desired result.

File: app-crypt/libsecret/libsecret-0.20.5-r1.ebuild [hunk 6]
```diff
@@ -0,0 +1,141 @@
...
+		$(meson_native_use_bool tpm tpm2)
+	)
+	meson_src_configure
+}
+
+multilib_src_test() {
+	if use tpm; then
+		dbus-run-session tpm2_run_with_emulator virtx meson test -C "${BUILD_DIR}"
+	else
+		dbus-run-session virtx meson test -C "${BUILD_DIR}"
```
mattst88, line 139: I've fixed this locally, but just for future reference: since `virtx` is a bash function, while `dbus-run-session` is a binary, you have to call them in the other order: `virtx dbus-run-session`.

mattst88: Makes me realize we should really have some kind of dbus eclass...

mattst88:
Really nice work. Thanks a bunch!

PR #24511, opened 2022-03-12
dev-java/snappy: Enable tests on arm, arm64 and ppc64

vaukai:
Closes: https://bugs.gentoo.org/834139

### DIFF
--- a/dev-java/snappy/snappy-1.1.7.8.ebuild
+++ b/dev-java/snappy/snappy-1.1.7.8.ebuild
@@ -23,27 +23,21 @@ KEYWORDS="amd64 ~arm arm64 ppc64 x86"
 # packaged.  Some extra steps are required before running the tests:
 # 1. Download hadoop-common 2.7.x from https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-common
 # 2. Set EANT_GENTOO_CLASSPATH_EXTRA to the path to hadoop-common-2.7.*.jar
-# 3. arm, arm64 and ppc64 only: Install test dependencies that are unkeyworded
-# 4. Set ALLOW_TEST="all"
+# 3. Set ALLOW_TEST="all"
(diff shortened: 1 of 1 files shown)


Leo3418 (commented):
Is a revision bump necessary? `RESTRICT="test"` is in effect (on every architecture) so I think it is safe to assume that nobody has enabled the `test` USE flag, thus these dependency changes in the `test? ( )` block effectively would not be propagated to anyone.

dev-java/snappy/snappy-1.1.7.8-r1.ebuild
```diff
@@ -0,0 +1,98 @@
...
+
+LICENSE="Apache-2.0"
+SLOT="1.1"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
+
+# Tests require org.apache.hadoop:hadoop-common:2.7.*, which is yet to be
+# packaged.  Some extra steps are required before running the tests:
+# 1. Download hadoop-common 2.7.x from https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-common
+# 2. Set EANT_GENTOO_CLASSPATH_EXTRA to the path to hadoop-common-2.7.*.jar
+# 3. arm, arm64 and ppc64 only: Install test dependencies that are unkeyworded
```
Leo3418, line 26: Please remove this step from the comment

Leo3418 (changes requested):

dev-java/snappy/snappy-1.1.7.8-r1.ebuild
vaukai: Done

vaukai:
Depends on https://bugs.gentoo.org/835237

fordfrog:
@vaukai merged, thanks!

PR #24512, opened 2022-03-12
dev-libs/mimalloc: set MI_BUILD_{STATIC,OBJECT}=OFF

MatthewGentoo:
Needed for solvespace (and any other future CMake-using consumers).

### DIFF
--- a/dev-libs/mimalloc/mimalloc-2.0.5-r2.ebuild
+++ b/dev-libs/mimalloc/mimalloc-2.0.5-r2.ebuild
@@ -1,0 +1,31 @@
+# Copyright 2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit cmake-multilib
+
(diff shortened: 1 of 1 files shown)


dev-libs/mimalloc/mimalloc-2.0.5-r2.ebuild
```diff
@@ -0,0 +1,31 @@
+# Copyright 2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit cmake-multilib
+
+DESCRIPTION="mimalloc is a compact general purpose allocator with excellent performance"
```
thesamesam, line 8: In a separate commit, let's do this just while I've noticed:
```suggestion
DESCRIPTION="A compact general purpose allocator with excellent performance"
```

Also, while not obligated, feel free to join me on mimalloc. It's only (initially) in tree for mold.

thesamesam (approved):

PR #22792, opened 2021-11-01
allow systemd-networkd to set hostname without polkit

gyakovlev:
@floppym as discussed on irc.

still WIP, not tested, just opening for visibility and initial feedback.
ofc it'll require a bit of keywording on dbus-broker, new GID assignement and flag masking and better comments, but later with that.
just drafting systemd change here.

not sure if it's ok to use `systemd_get_systemunitdir` here.

PR #24514, opened 2022-03-12
sys-apps/systemd: warn when dbus-broker is needed

Karlson2k:
Warn if hostnamed-fallback workaround is used, but dbus-broker.service is not enabled. This should give users a strong hint of  what needs to be done.

See https://github.com/gentoo/gentoo/pull/22792#issuecomment-1065918296

No revbumps.

### DIFF
--- a/sys-apps/systemd/systemd-250.2.ebuild
+++ b/sys-apps/systemd/systemd-250.2.ebuild
@@ -510,6 +510,14 @@ pkg_postinst() {
 		eerror "systemd again."
 		eerror
 	fi
+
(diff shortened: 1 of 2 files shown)


pinkflames:
Perhaps it would make more sense for dbus-broker.service to be vendor enabled? Then instead of checking for its presence, the systemd ebuild could just `systemctl preset dbus-broker.service` (or socket, if that's appropriate), since it's presumably only installed if the user intends to use it (and it's always possible to still opt-out by using `systemctl mask dbus-broker.service`).

PR #24517, opened 2022-03-12
media-libs/nv-codec-headers: drop versions

SoapGentoo:


ionenwks (approved):

SoapGentoo:
@thebitpit I'd like to unify how nvenc/nvdec are used, to you mind me renaming `nvdec` to `nvenc` in mythtv, seeing that they both pull in the same set of dependencies and make handling the flags on a global scale easier?

PR #24518, opened 2022-03-12
net-im/telegram-desktop: Bump to 3.6.0

mid-kid:
@gyakovlev


net-im/telegram-desktop/telegram-desktop-3.6.0.ebuild
```diff
@@ -3,20 +3,21 @@
 
 EAPI=8
 
-PYTHON_COMPAT=( python3_{7..10} )
+PYTHON_COMPAT=( python3_{8..10} )
 
-inherit xdg cmake python-any-r1 optfeature
+inherit xdg cmake python-any-r1 optfeature flag-o-matic
```
arthurzam, line 8: Are you sure `flag-o-matic` is needed?
CI and `pkgcheck scan --commit` show this as unused.

mid-kid: I needed it during development at some point, thanks for pointing it out.

PR #24519, opened 2022-03-12
sci-mathematics/z3: fix link atomic in riscv

Rabenda:
Bug: https://bugs.gentoo.org/835003
Signed-off-by: Han Gao <rabenda.cn@gmail.com>

### DIFF
--- a/sci-mathematics/z3/files/z3-4.8.14-libatomic.patch
+++ b/sci-mathematics/z3/files/z3-4.8.14-libatomic.patch
@@ -1,0 +1,60 @@
+https://github.com/Z3Prover/z3/commit/3d87d86c2825437780167ae6ff8f5b203f15ee7f.patch
+https://bugs.gentoo.org/835003
+
+From b2b407cdedc586f8d11f9dd029e1c3016c80bba7 Mon Sep 17 00:00:00 2001
+From: Revy <rabenda.cn@gmail.com>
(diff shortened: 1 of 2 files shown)


dlan17:
hi, thanks for the contribution and hunting this bug down, the patch LGTM

I don't kind it's a specific problem relate to riscv only which mean other arches may have the same issue, so drop "riscv" in commit message

what's the status of this patch in upstream? did someone report to z3 upstream? if not, please also keep them informed, basically we want to know if we (Gentoo side) have to maintain patch at downstream forever (or during next version bump) or not

Rabenda:
[cmake/modules/CheckAtomic.cmake](https://github.com/KDE/krita/blob/620fbcbdcb9b33a38de86641e6db333e98007bb3/cmake/modules/CheckAtomic.cmake)

the patch refers to this rewrite.

https://github.com/Z3Prover/z3/issues/5896 I report it.

thesamesam:
> what's the status of this patch in upstream? did someone report to z3 upstream? if not, please also keep them informed, basically we want to know if we (Gentoo side) have to maintain patch at downstream forever (or during next version bump) or not

Yeah, we want to avoid this as much as possible. Please also always include links to bug reports (both downstream and upstream) and any relevant patches/commit links upstream at the top of patches.

jsmolic:
I see that upstream PR got merged, which is great! @Rabenda could you just add a link to the PR in the patch header?

PR #24522, opened 2022-03-13
sci-physics/geant4_vmc: Fix build with `=sci-physics/geant-4.11.0.1`, adapt to dropped VMC for new `sci-physics/root`

olifre:
This backports a fix from upstream for compilation against `=sci-physics/geant-4.11.0.1`, and adapts to the dropped `vmc` USE flag in more recent `sci-physics/root` releases. 

@amadio That's probably a PR for you ;-).


olifre:
@amadio I've taken the chance to also add version `6.1` here, released shortly ago. 

I found during my tests that in case some parts of the VMC stack are not rebuilt after a Geant4 update, linker errors may be experienced (missing symbols), e.g. in caseGeant4 is updated and `vgm` is not rebuilt but `geant4_vmc` is rebuilt. 

I'm not sure about the best fix — rpobably, all VMC packages should use `sci-physics/geant:=` since the ABI of Geant4 may change in such ways. But this would also mean subslots may be needed for each Geant4 version to trigger these rebuilds via `:=`.

amadio:
Thanks! I will merge this tomorrow. Will probably try to get rid of ROOT versions with vmc USE flag ang rely only on the external package now.

olifre:
Sure, that sounds like a good idea :+1: . Thanks in advance!

amadio:
BTW, you don't need to tag me, the Gentoo bot already does it. Thanks!

olifre:
Thanks! 
Indeed, I realized I sent you extra-spam only after submitting the PRs, noted for the future ;-). 

Now only the `sci-physics/vmc` PR is left (I realized I had to adapt to the dropped `vmc` USE flag in https://github.com/gentoo/gentoo/pull/24523 , otherwise Portage would implicitly enforce `<sci-physics/root-6.26.00`).

PR #24524, opened 2022-03-13
backintime-1.3.2

Anard:
Now BackInTime < 1.3.2 didn't work if Python 3.10 was installed. v1.3.2 solved this issue.
Could BiT-1.3.2 be added to main Gentoo repositories ?


MatthewGentoo:
Thanks for your contribution :)

Before this can be merged, you need to add to agree to the [GCO](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin) by adding a `Signed-off-by` tag to your commit. Note that this needs to be done with your full name.

I think that the easiest way to do this is:

```bash
git config user.name 'Full Name'
git commit --amend --signoff
git push --force-with-lease
```

PR #24535, opened 2022-03-13
Add new version Graylog

serg-sg:


juippis (approved):
I'll squash the commits, thanks!

PR #24537, opened 2022-03-13
dev-util/Tensile: FHS and benchmark feature

littlewu2508:
The previous Tensile package installs non-python files
in python site directory. This change move Config files
to /usr/share. Various patches are applied for correcting
paths.

Also, enable gfx1031 target so people can run benchmark
on navi22 GPUs.


AndrewAmmerlaan:
Have you tested if rocBLAS can still compile correctly after these changes?

littlewu2508:
> 

I tested rocBLAS-4.3.0-r1. It seems that it does not build with Tensile (calling `TensileCreateLibrary` command) either using your original Tensile-4.3.0 nor this Tensile-4.3.0-r1, although Tensile package is found by cmake.

AndrewAmmerlaan:
> > 
> 
> I tested rocBLAS-4.3.0-r1. It seems that it does not build with Tensile (calling `TensileCreateLibrary` command) either using your original Tensile-4.3.0 nor this Tensile-4.3.0-r1, although Tensile package is found by cmake.

That is weird, it was working before, did https://github.com/gentoo/gentoo/pull/24679 fix this?

dev-util/Tensile/Tensile-4.3.0-r1.ebuild
```diff
@@ -0,0 +1,86 @@
...
+	popd || die
+
+	sed -e "/package_data/d" -e "/data_files/d" -i setup.py || die
+}
+
+python_install() {
+	distutils-r1_python_install
+
+	python_moduleinto Tensile
+	pushd Tensile
```
AndrewAmmerlaan, line 73: btw, there's a `|| die` missing here. No need to fix it now since the PR is already merged, we can fix it the next time we touch the ebuild.

dev-util/Tensile/Tensile-4.3.0-r1.ebuild
```diff
@@ -0,0 +1,86 @@
...
+KEYWORDS="~amd64"
+SLOT="0/$(ver_cut 1-2)"
+
+# Not compatible with recent versions of pytest
+RESTRICT="test"
+
+RDEPEND="${PYTHON_DEPS}
+	dev-python/pyyaml[${PYTHON_USEDEP}]
+	dev-python/msgpack[${PYTHON_USEDEP}]
+	>=dev-util/rocm-smi-4.3.0
```
AndrewAmmerlaan, line 25: Shouldn't this be `dev-util/rocm-smi:${SLOT}` instead?

littlewu2508:
> That is weird, it was working before, did #24679 fix this?

#24679 is the fix for rocBLAS.

AndrewAmmerlaan:
> > That is weird, it was working before, did #24679 fix this?
> 
> #24679 is the fix for rocBLAS.

Great, thanks :+1:

dev-util/Tensile/Tensile-4.3.0-r1.ebuild
littlewu2508: According to my observation there is no strict SLOT dependency with rocm-smi: it's just a library to monitor hardware, adjust fan and GPU clock which helps some GPU cool themselves when running benchmarks.

PR #24539, opened 2022-03-13
SuperTuxKart

Anard:
Added unstable (9999) ebuild based on git master version.
Corrected 1.3 ebuild which don't use glew anymore.


Anard:
I don't manage to signoff older commits... :(

PR #24541, opened 2022-03-13
dev-python/async_generator: fix installation for python3_9

telans:
Bug: https://bugs.gentoo.org/835090

Does not fix tests linked in bug, only installation.


zmedico:
Merged in cc862d165891d3699cb7dc2e056206dd69f8f240. Thanks!

PR #24542, opened 2022-03-13
dev-libs/liborcus: fix patch phase/naming

telans:


dev-libs/liborcus/liborcus-0.17.0.ebuild
```diff
@@ -44,7 +44,7 @@ pkg_setup() {
 
 src_prepare() {
 	# bug 713586
-	use test && eapply "${FILESDIR}/${P}-test-fix.patch"
+	use test && eapply "${FILESDIR}/${PN}-0.17.0-test-fix.patch"
```
a17r, line 47: What do we keep this ebuild for at all?

a17r (changes requested):

a17r:
Thanks, picked the fix to 0.17.2.

PR #24544, opened 2022-03-13
net-misc/seafile{,-client}: version bump to 8.0.6

CaseOf:


net-misc/seafile/seafile-8.0.6.ebuild
```diff
@@ -0,0 +1,58 @@
...
+
+PYTHON_COMPAT=( python3_{8..10} )
+
+RELEASE_COMMIT="7a61feaba63006d11c09d1020593ba3fcb3fc717"
+
+inherit autotools python-single-r1 vala
+
+DESCRIPTION="File syncing and sharing software with file encryption and group sharing"
+HOMEPAGE="https://www.seafile.com/ https://github.com/haiwen/seafile/"
+SRC_URI="https://github.com/haiwen/${PN}/archive/${RELEASE_COMMIT}.tar.gz -> ${P}.tar.gz"
```
MatthewGentoo, line 14: To save a bit of manual work each update, you can use `${PV}` here like this:

```suggestion
SRC_URI="https://github.com/haiwen/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
```

(`S="..."` will also need changing)

net-misc/seafile/seafile-8.0.6.ebuild
```diff
@@ -0,0 +1,58 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
MatthewGentoo, line 4: Maybe we could bump to EAPI-8 next time :)

```suggestion
EAPI=8
```

net-misc/seafile-client/seafile-client-8.0.6.ebuild
```diff
@@ -0,0 +1,51 @@
...
+
+EAPI=8
+
+RELEASE_COMMIT="1fb9ddd71fbf6f0252509aced527be459e240366"
+
+inherit xdg cmake
+
+DESCRIPTION="Seafile desktop client"
+HOMEPAGE="https://www.seafile.com/ https://github.com/haiwen/seafile-client/"
+SRC_URI="https://github.com/haiwen/${PN}/archive/${RELEASE_COMMIT}.tar.gz -> ${P}.tar.gz"
```
MatthewGentoo, line 12: Same as above, since the upstream are still creating tags for each release, you can use `${PV}` here.

MatthewGentoo (approved):
Thanks :) Some ideas for next time in the review.

I also filed a [stabilisation bug](https://bugs.gentoo.org/835154). Maybe once version 8.0.5 is marked as stable, you could go and delete all of the old versions.

net-misc/seafile/seafile-8.0.6.ebuild
CaseOf: Hi,
I was using commit hash since some versions ago because upstream was changing tags to other commit without warning.
Then I thought it was better to keep this behavior to avoid unnecessary issues.

ionenwks: Yes, seconding this. I've seen this upstream moving tags around before, and not only once.

fwiw maybe a note in the ebuild be nice so other developers don't wonder why every time?

MatthewGentoo:
Sorry for the delay, I saw https://bugs.gentoo.org/685920 and wanted to make sure that it was okay to keep in ::gentoo :)

PR #24545, opened 2022-03-14
net-libs/nodejs: keyword 16.13.1 for ~riscv

guyu2876:
Closes: https://bugs.gentoo.org/782913
Signed-off-by: Yu Gu <guyu2876@gmail.com>

### DIFF
--- a/profiles/package.mask
+++ b/profiles/package.mask
@@ -33,6 +33,13 @@
 
 #--- END OF EXAMPLES ---
 
+# John Helmert III <ajak@gentoo.org> (2022-03-13)
+# Arti is not a production-ready Tor client. According to upstream, "do
+# not use this program yet if you seriously need anonymity, privacy,
+# security, or stability." Feel free to unmask if this is not a
+# necessity.
+net-p2p/arti
+
 # Alfredo Tupone <tupone@gentoo.org> (2022-03-13)
 # No revdeps, last update on 2011
 # Removal on 2022-04-13.


net-libs/nodejs/files/nodejs-16.5.0-add-riscv.patch
```diff
@@ -0,0 +1,967 @@
...
+---
+ common.gypi                                   |   2 +-
+ .../riscv64/baseline-assembler-riscv64-inl.h  | 260 +++++-------------
+ .../riscv64/baseline-compiler-riscv64-inl.h   |  64 +----
+ .../src/builtins/riscv64/builtins-riscv64.cc  |  68 +++--
+ .../codegen/riscv64/macro-assembler-riscv64.h |   2 +-
+ 5 files changed, 120 insertions(+), 276 deletions(-)
+
+diff --git a/common.gypi b/common.gypi
+index b8d61f0d3270..89fac788d3d2 100644
```
jsmolic, line 27: Usually I like to drop these two lines as they don't really contain any useful data and make the patch larger overall (not needed to apply here if not convenient, just stating in general :) )

jsmolic (commented):
Few brief points (haven't looked at the patch closely):
- This is quite big patch, Gentoo has a policy regarding allowed [patch size](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html) in `./files` directory (20 kb max size) so if we add it we'll have to put it to one of developer's devspaces.
- Could you split the changes in 2 commits please? One for adding the patch, and another one for keywording the ebuild.

PR #24551, opened 2022-03-14
net-misc/ps3mediaserver: min java 1.8:*

vaukai:
@thesamesam

### DIFF
--- a/net-misc/ps3mediaserver/ps3mediaserver-1.90.1-r2.ebuild
+++ b/net-misc/ps3mediaserver/ps3mediaserver-1.90.1-r2.ebuild
@@ -1,0 +1,90 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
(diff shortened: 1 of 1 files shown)


net-misc/ps3mediaserver/ps3mediaserver-1.90.1-r2.ebuild
```diff
@@ -0,0 +1,90 @@
...
+RDEPEND=">=virtual/jre-1.8:*
+	media-libs/libmediainfo
+	media-libs/libzen
+	tsmuxer? ( media-video/tsmuxer )
+	transcode? ( media-video/mplayer[encode] )"
+
+S=${WORKDIR}/pms-${PV}
+PMS_HOME=/opt/${PN}
+
+src_prepare() {
```
Flowdalic, line 27: nit: all three "here documents" in this src_prepare are missing `die`. I know that this is already the case in the existing ebuild. But feel free to improve this ebuild while working on it. :)

PR #24558, opened 2022-03-14
[media-gfx/blender] Version bump and ffmpeg 5.0 fix

DarkDefender:
I bumped the blender version to the latest 3.1 release.
In addition to this I created two new ebuild revisions that patches in ffmpeg 5.0 support (for older blender versions).


DarkDefender:
@thesamesam I forgot to mention that the new ebuild hasn't had the test tarball packaged on the gentoo servers.
So that is and extra step that needs to be done.

Hopefully at Blender we will eventually manage to migrate from Phabricator to something that allows easy download of snapshots so you don't have to package and host the tests for the gentoo ebuilds.
Progress in this area has been slow though sadly :c

PR #24559, opened 2022-03-14
x11-wm/mutter: Fix typo: 'emsonargs' -> 'emesonargs'

Leo3418:
Currently, fixing this typo has no effect at all because the default values of `egl_device` and `wayland_eglstream` Meson options assigned by the upstream are `false` for every version we currently have. But just in case the upstream changes those options' default in a future release, or someone copies this block of script with the typo to use it elsewhere, fixing it right now could help avoid future gotchas.


ionenwks:
Oops, seems that was my typo from a while ago. I'll go ahead and merge it myself.

PR #24561, opened 2022-03-14
meson.eclass: print error logs for failing tests

djcampello:


mattst88:
cc: @floppym, @williamh

mattst88 (approved):

PR #24566, opened 2022-03-15
app-shells/starship: Drop 1.0.0, drop 1.2.1, and add 1.4.2

bowlofeggs:


ionenwks:
By the way, I just added commit f07cc77e6cb36ff934c2bc770356df60f0d1a289 because of this:
https://github.com/starship/starship/blob/52fa4bbab4393589aebbeded4351886e657a6275/Cargo.toml#L24

I didn't notice it at first given I used ~arch rust-1.59 but a user reported it failed to build with stable rust-1.58.1, app-misc/broot did something similar, guess it may be a new trend may have to watch out for :eyes:

bowlofeggs:
Thanks for catching that @ionenwks . I also have ~arch Rust, which is why I missed it too. I'm glad Rust now has the minimum version in `Cargo.toml` though. I wonder if it would be difficult to add a QA check for this.

PR #24573, opened 2022-03-15
www-misc/monitorix: Removed older versions

Dwosky:
Version 3.14.0 its stable, so I'm removing all other versions.

Signed-off-by: Pedro Arizmendi <dwosky@pm.me>


juippis (approved):
Thanks!

PR #24576, opened 2022-03-15
net-misc/gerbera: Bump to 1.10.0

tietzc:
I would also like to proxy maintain this, as it is currently marked maintainer-needed.

Closes: https://bugs.gentoo.org/834972


thesamesam:
Yay! Are we keeping the live ebuild? Please sync it with your changes or drop if not interested in maintaining it.

tietzc:
Upstream development is quite active, so let's keep it for now,

thesamesam:
Thank you!

PR #24579, opened 2022-03-15
net-libs/toxext: fix clang build && net-im/qtox: add 1.17.6 remove 1.16.3

Chiitoo:
I'm assuming here a little that this does indeed fix the clang build, since I've not been able to test it myself...

No ebuild changes for qTox (and not much changed for Linux overall, but I think there are at least some things there).


net-im/qtox/qtox-1.17.6.ebuild
```diff
@@ -0,0 +1,93 @@
...
+	dev-qt/qtsvg:5
+	dev-qt/qtwidgets:5
+	dev-qt/qtxml:5
+	media-gfx/qrencode:=
+	media-libs/libexif
+	media-libs/openal
+	media-video/ffmpeg:=[webp,v4l]
+	>=net-libs/tox-0.2.13:=[av]
+	>=net-libs/toxext-0.0.3
+	>=net-libs/tox_extension_messages-0.0.3
```
ionenwks, line 48: Seems this is still only used by 9999, I'll yank+amend the two lines on merge to be like 1.17.4.

.. merging which I guess may take a bit, need to wait for woodpecker to be in a good mood :eyes:

ionenwks (approved):
Tested toxext with clang+lld and does fix it indeed. This will probably be "fixed" by >=clang-14 too given it'll have default pie like gcc (which implies -fPIC, ends up "hiding" these lacking -fPIC issues while packages shouldn't actually assume default pie and so the fix will still matter).

net-im/qtox/qtox-1.17.6.ebuild
thesamesam: (Why need woodpecker for this?)

ionenwks: Oh right, for some reason I thought pushing went through it.

Chiitoo: Huh, yeah, not sure how I managed that.  Probably the sleepy.  :\

Thanks!

PR #24580, opened 2022-03-15
net-irc/weechat: Version updated to 3.4.1

jpds:
Tested on `amd64`.

Bug: https://bugs.gentoo.org/835133


thesamesam:
@mgorny pls test

net-irc/weechat/weechat-3.4.1.ebuild
```diff
@@ -84,6 +91,18 @@ pkg_setup() {
 	use python && python-single-r1_pkg_setup
 }
 
+src_unpack() {
+	if use verify-sig; then
+		cd "${DISTDIR}" || die
+		verify-sig_verify_detached ${P}.tar.xz ${P}.tar.xz.asc
+		verify-sig_verify_unsigned_checksums \
+			${P}-sha512.txt sha512 ${P}.tar.xz
```
mgorny, line 99: But why are you doing this? You've already verified the distfile against the signature, so checksums are redundant and since they're not even signed, they have no value.

mgorny (changes requested):

thesamesam:
Please sync the live ebuild then lgtm

mgorny:
The verify-sig part LGTM now.

PR #24581, opened 2022-03-15
verify-sig.eclass: Handle SHA512 checksums

jpds:
Needed as weechat.org only publish these.

### DIFF
--- a/eclass/verify-sig.eclass
+++ b/eclass/verify-sig.eclass
@@ -229,6 +229,10 @@ verify-sig_verify_unsigned_checksums() {
 			chksum_prog=sha256sum
 			chksum_len=64
 			;;
+		sha512)
+			chksum_prog=sha512sum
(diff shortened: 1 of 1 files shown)


mgorny:
Could you include the weechat update here as well?

jpds:
@mgorny I was going to stick at onto #24580 after a rebase (wasn't sure how to do this change there without it not being visible).

mgorny:
I don't really understand. Feel free to add the eclass change there and just ping me. I'd like to test it before approving (even though it looks obviously correct).

thesamesam:
Are we still doing this (see: https://github.com/gentoo/gentoo/pull/24580#discussion_r827397504). We've merged #24580.

mgorny:
This is trivial, so let's return to it when we have a real test case.

mgorny:
(no offense meant to the OP)

PR #24583, opened 2022-03-15
dev-python/cloudscraper: Bump to 1.2.60

vaartis:
No changes to the ebuild itself


juippis (approved):
In this case though we could've used **some** updates, like EAPI and copyright year! But I'll get those while merging, thanks!

PR #24585, opened 2022-03-15
dev-libs/mimetic: fix musl build

arkamar:
This applies a patch I have sent to upstream, which fixes musl build.


arkamar:
It is still not reviewed by upstream maintainer and I don't know if he will want more changes. Anyway, the presented PR to upstream contains two commits, the musl fix and then conversion of remaining `uints`. I believe that the standalone musl fix is better for package maintainers and I would stick with this patch here as well.

PR #24590, opened 2022-03-15
app-admin/syslog-ng: bump to 3.36.1

hydrapolic:
Also test deps boxfort/criterion.

Tested on ~amd64.


hydrapolic:
Thank you @thesamesam

PR #24595, opened 2022-03-15
Quote CC variables in various packages to allow spaces

fx-carton:
A common use case for this is things like "gcc -m32" or "gcc --sysroot=...".

This does not cover all packages, but rather (most of) the packages I've installed that fail to build in such situation.


SoapGentoo (approved):
this is definitely something we want, :+1:

PR #24596, opened 2022-03-16
sys-apps/man-db: replace su with su-exec

LeoQuote:
su is deprecated in shadow v4.10 https://github.com/shadow-maint/shadow/releases/tag/v4.10


floppym:
su is provided by util-linux by default. Most systems will not have su-exec installed at all.

LeoQuote:
Thanks for your review! I just added su keyword for linux-utils

PR #24603, opened 2022-03-16
www-servers/tomcat: Switch to ECJ_SLOT="4.15"

vaukai:


fordfrog:
i have to recheck the tomcat situation. there was something that prevented me from switching to the newer version of ecj but i need to verify whether the reason is still valid or not.

fordfrog:
my main concern is that tomcat must still work with java 1.8 as that is what the upstream supports. otherwise some apps might not work with tomcat because of the changes between java 1.8 and java 9+. when looking at it, `eclipse-ecj` needs java 11 for compilation, but i forced that to use source and target 1.8, so that might still work (unless some class would be missing in 1.8). `ant-eclipse-ecj` needs java 11 completely, but it doesn't seem to be part of the classpath. so it might work, but i need to do some tests to verify nothing is broken.

Alessandro-Barbieri:
What about dropping the json dependency?

dev-java/gwt/gwt-2.9.0-r1.ebuild
```diff
@@ -0,0 +1,96 @@
...
+# there are many tests that end with FAILED status. It is unclear if this is
+# expected from upstream or not, but in order to keep things honest, the test
+# suite is being restricted
+RESTRICT="mirror test"
+SLOT="2.8"
+LICENSE="Apache-2.0"
+KEYWORDS="~amd64 ~x86 ~amd64-linux ~x86-linux"
+
+CDEPEND="
+	dev-java/json:0
```
Alessandro-Barbieri, line 27: No more needed
see https://bugs.gentoo.org/818658

vaukai:
> so it might work, but i need to do some tests to verify nothing is broken.

@fordfrog Could you specify those test you have in mind?  Maybe I could assist testing.

fordfrog:
> > so it might work, but i need to do some tests to verify nothing is broken.
> 
> @fordfrog Could you specify those test you have in mind? Maybe I could assist testing.

i think it should be enough to compile the new versions of tomcat and run an instance of each with jdk 1.8. if there are no reports of unsupported classes during the startup, i would consider it safe to do the migration. it would be great if you could do the tests, i'm really busy and not sure when i would get to it. to speed the testing up, you can ping me on irc so that i could assist you.

vaukai:
@fordfrog 
Just added `8.5.77-r1` and `8.5.78-r1` and rebased.  Will start testing soon

fordfrog:
we still cannot drop `eclipse-ecj-4.5.1`, that we can do once the new versions of tomcat get stabilized and old are gone. and for the `8.5` slot we should add `8.5.78-r1` and keep the current versions. otherwise it looks good, thanks :-)

vaukai:
> we still cannot drop `eclipse-ecj-4.5.1`, that we can do once the new versions of tomcat get stabilized and old are gone. and for the `8.5` slot we should add `8.5.78-r1` and keep the current versions. otherwise it looks good, thanks :-)

All done.

fordfrog:
@vaukai thanks! can we at least drop ecj `4.10` and `4.13` now that we know we won't need them? or are they still used somewhere?

vaukai:
> @vaukai thanks! can we at least drop ecj `4.10` and `4.13` now that we know we won't need them? or are they still used somewhere?

https://github.com/gentoo/gentoo/pull/25277, but better please wait for green

PR #24606, opened 2022-03-16
riscv: keyword www-apps/nanoc, net-analyzer/{httping,2ping}

liangyongxiang:
knowns:
https://bugs.gentoo.org/738716  dev-ruby/parallel-1.19.2 fails tests
https://bugs.gentoo.org/801661 www-apps/nanoc-cli-4.12.2 fails tests

### DIFF
--- a/dev-ruby/colored/colored-1.2-r1.ebuild
+++ b/dev-ruby/colored/colored-1.2-r1.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=8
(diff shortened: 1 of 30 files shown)


dlan17:
there is a DroppedKeywords warning for dev-ruby/shoulda-matchers, since we only keyworded the slot 3 here.
but I think it's probably OK so far, since there is no other packages depend on slot4,5 as a quick grep in portage tree.

since the two tests failures are also exist in other Arch, so shouldn't be a block here.

jsmolic:
@liangyongxiang Can you also keyword the latest version of dev-ruby/shoulda-matchers?

liangyongxiang:
> @liangyongxiang Can you also keyword the latest version of dev-ruby/shoulda-matchers?

OK. I will keyword it together in the next PR.
I didn't notice that there is also a new version without the keyword.

PR #24611, opened 2022-03-16
dev-ruby/rubygems: Fix non-root bin dir

toshokan:
It seems the gem binary install path has changed with the new gentoo-defaults-5 configuration. 

Previously, gems were installed to `~/.gem/ruby/<version>/gem/bin`, now they are installed to `~/bin`. 
This seems like a bad idea, and I couldn't find the change documented anywhere, so I figured it might be a typo.

@graaff What do you think?

### DIFF
--- a/dev-ruby/rubygems/files/gentoo-defaults-5.rb
+++ b/dev-ruby/rubygems/files/gentoo-defaults-5.rb
@@ -13,7 +13,7 @@ module Gem
     end
 
     def gentoo_bindir
-      Process.euid.zero? ? '@GENTOO_PORTAGE_EPREFIX@/usr/local/bin' : File.join(user_home, 'bin')
+      Process.euid.zero? ? '@GENTOO_PORTAGE_EPREFIX@/usr/local/bin' : File.join(user_dir, 'bin')
     end
 
     def gentoo_local_dir


toshokan:
@graaff Would you be able to verify whether this is intended behaviour?

graaff:
I've added that default since it seemed to make sense to me as I was overhauling that code. So in that sense it's not a bug. But I don't have strong opinions on this and I don't use plain rubygems to install things, so I'm happy to discuss other defaults here.

@toshokan do you have additional arguments other than that your change reverts to the previous location? Just trying to get a better understanding of the best location, so that it can also be documented in the defaults file.

toshokan:
@graaff One thing that made me want to revert this is that if all the gems are installed to `~/bin`, I don't see a way to have different gem versions for different ruby versions (which is especially important when testing new ruby versions). With the old install location, we can just switch the bin directory in $PATH when using a different ruby version if we want to call binaries from the shell.

graaff:
That's also possible with `ruby -S` but that is perhaps less obvious and won't work for calls in scripts.

I'll have another look at what the ruby defaults actually are and I will probably stick with that.

konsolebox:
This behavior actually synchronizes with Gentoo dynamic selection of Ruby and correct gem executables but it requires `custom_shebang` to be set to `/usr/bin/env ruby`.
```
diff --git a/dev-ruby/rubygems/files/gentoo-defaults-5.rb b/dev-ruby/rubygems/files/gentoo-defaults-5.rb
index 817eb1eda4d5..4990872c9ceb 100644
--- a/dev-ruby/rubygems/files/gentoo-defaults-5.rb
+++ b/dev-ruby/rubygems/files/gentoo-defaults-5.rb
@@ -40,6 +40,7 @@ module Gem
       options = "--install-dir #{gentoo_install_dir} --bindir #{gentoo_bindir}"
 
       {
+        'custom_shebang' => '/usr/bin/env ruby',
         'install' => options,
         'uninstall' => options,
         'update' => options

```

konsolebox:
I currently have this as my modified defaults code.  It's more aggressive since it only allows gems to be installed in `/usr/local` when EUID is 0, but it's simpler and works transparently with bundler.  The current setting requires having `GEM_HOME` set to `user_dir` so bundler installs gems to it by default instead of `/usr/lib*`.

I also moved `vendor_dir` to the bottom (end) of the path so it's consistent with the original but I'm not sure about it.  I'm thinking it probably should have the highest priority instead.  I don't like the idea of placing it in between.  It's either package gems first, or current system gems first.  No user-gems-can-be-considered-a-temporary-patch-or-test-gems feature kind of thing.

I thought about making the install dir (which should be `/usr/local*`) different from the default dir (which could be `user_dir`) when EUID is 0 so bundler doesn't install to `/usr/local`, but it only makes things complicated.

`gem uninstall` for example will not uninstall gems in `user_dir`.
```
# frozen_string_literal: true

module Gem
  module OperatingSystemDefaults
    def user_dir
      gentoo_user_dir
    end

    def default_dir
      gentoo_user_dir
    end

    def default_bindir
      gentoo_user_bindir
    end

    def default_path
      path = []
      path << gentoo_user_dir unless gentoo_local_dir == gentoo_user_dir
      path << gentoo_local_dir
      path << gentoo_dir
      path << vendor_dir if vendor_dir && File.directory?(vendor_dir)
      path
    end

    def operating_system_defaults
      { 'custom_shebang' => '/usr/bin/env ruby' }
    end

  private
    def gentoo_dir
      @gentoo_dir ||= RbConfig::CONFIG['sitelibdir'].gsub('site_ruby', 'gems')
    end

    def gentoo_local_dir
      @gentoo_local_dir ||= gentoo_dir.gsub('/usr', '/usr/local')
    end

    def gentoo_user_dir
      @gentoo_user_dir ||= Process.euid.zero? ? gentoo_local_dir :
          method(:user_dir).super_method.call
    end

    def gentoo_user_bindir
      @gentoo_user_bindir ||= Process.euid.zero? ? '/usr/local/bin' : File.join(user_home, 'bin')
    end
  end

  self.singleton_class.prepend OperatingSystemDefaults
end
```

PR #24613, opened 2022-03-16
media-libs/libplacebo: Bump to version 4.192.1

orbea:
This update is needed for `media-video/mpv-9999` which fails with a libplacebo version check during configure otherwise.


thesamesam:
Thank you!

orbea:
Thanks!

PR #24617, opened 2022-03-17
dev-ada/libgpr: rename libgpr.so to libgnatprj.so

Alessandro-Barbieri:
Closes: https://bugs.gentoo.org/818505


atupone:
I only renamed the library in /usr/lib.
Thanks

PR #24621, opened 2022-03-17
Small fixes after VMAF enablement in FFmpeg

AdelKS:
Hello,

Small fixes after #23412 

Adel

### DIFF
--- a/media-libs/libvmaf/libvmaf-2.3.0.ebuild
+++ b/media-libs/libvmaf/libvmaf-2.3.0.ebuild
@@ -34,6 +34,11 @@ multilib_src_configure() {
 	meson_src_configure
 }
 
+multilib_src_install() {
+	meson_src_install
+	find "${D}" -name '*.la' -delete -o -name '*.a' -delete || die
+}
+
 multilib_src_install_all() {
 	einstalldocs
 
--- a/media-libs/libvmaf/libvmaf-2.3.0.ebuild
+++ b/media-libs/libvmaf/libvmaf-2.3.0.ebuild
@@ -21,6 +21,9 @@ fi
 LICENSE="BSD-2-with-patent"
 SLOT="0"
 
+# upstream issue, see bug #835471
+RESTRICT="test"
+
 BDEPEND="
 	dev-lang/nasm
 "
--- a/media-video/ffmpeg/ffmpeg-9999.ebuild
+++ b/media-video/ffmpeg/ffmpeg-9999.ebuild
@@ -339,8 +339,6 @@ src_prepare() {
 		export revision=git-N-${FFMPEG_REVISION}
 	fi
 
-	eapply "${FILESDIR}/vmaf-models-default-path.patch"
-
 	default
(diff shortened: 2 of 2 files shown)


media-libs/libvmaf/libvmaf-2.3.0.ebuild
```diff
@@ -21,6 +21,8 @@ fi
 LICENSE="BSD-2-with-patent"
 SLOT="0"
 
+RESTRICT="test"
+
```
juippis, line 25: We add a reason for disabling tests by adding a comment on top of it, preferably with an upstream issue so we can track when it gets fixed.

media-libs/libvmaf/libvmaf-2.3.0.ebuild
```diff
@@ -34,6 +36,11 @@ multilib_src_configure() {
 	meson_src_configure
 }
 
+multilib_src_install() {
+	meson_src_install
+	rm ${ED}/usr/$(get_libdir)/libvmaf.a || die "Failed to delete static library in /usr/$(get_libdir)"
```
juippis, line 41:
```suggestion
	find "${D}" -name '*.la' -delete -o -name '*.a' -delete || die
```

media-libs/libvmaf/libvmaf-2.3.0.ebuild
AdelKS: The related bug is this one https://bugs.gentoo.org/835471, some paths are hardcoded and don't work. @thesamesam prefers that I look into it first. Although I cannot seem to find where the test executables are produced to fix this.

AdelKS: Okay in the end it's an upstream issue, I opened a bug report https://github.com/Netflix/vmaf/issues/998

Should I disable tests for now or leave it be ? The library itself works for our use case anyway.

thesamesam: Please add a link to that issue above the restriction

juippis (approved):
Getting this, 
```
[29/150] /usr/lib/python-exec/python3.10/meson --internal vcstagger ../vmaf-2.3.0/libvmaf/include/  vcs_version.h.in include/vcs_version.h 2.3.0 /var/tmp/portage/media-libs/libvmaf-2.3.0/work/vmaf-2. 3.0/libvmaf/include @VCS_TAG@ '(.*)' /usr/bin/git --git-dir /var/tmp/portage/media-libs/libvmaf-2.3.0/work/vmaf-2.3.0/libvmaf/../.git describe --tags --long --match '?.*.*' --always
fatal: not a git repository: '/var/tmp/portage/media-libs/libvmaf-2.3.0/work/vmaf-2.3.0/libvmaf/../.git'
```

AdelKS:
I saw that message, it fails to determine its own version using `.git` folder, but it doesn't seem that fatal as the compilation continues and the testing too. As we can see in the [log](https://835471.bugs.gentoo.org/attachment.cgi?id=767205)

PR #24626, opened 2022-03-17
gkrellm-plugin.eclass: Support EAPI 8

laumann:
The most notable changes are:

 * BDEPEND on "virtual/pkgconfig" instead of DEPEND

 * ~~Remove the _GKRELLM_PLUGIN_1 variable as this eclass is not inherited by any
   other eclass~~ Was asked to keep it in.

 * Add @DEFAULT_UNSET to the variables PLUGIN_SERVER_SO and PLUGIN_DOCS

Looking over the eclass, it doesn't appear to use any functionality that was
modified or banned in EAPI 7 or 8.

### DIFF
--- a/eclass/gkrellm-plugin.eclass
+++ b/eclass/gkrellm-plugin.eclass
@@ -7,7 +7,8 @@
 # @AUTHOR:
 # Original author: Jim Ramsay <lack@gentoo.org>
 # EAPI 6 author: David Seifert <soap@gentoo.org>
-# @SUPPORTED_EAPIS: 6
+# EAPI 8 author: Thomas Bracht Laumann Jespersen <t@laumann.xyz>
+# @SUPPORTED_EAPIS: 6 8
 # @PROVIDES: multilib
 # @BLURB: Provides src_install used by (almost) all gkrellm plugins
 # @DESCRIPTION:
@@ -15,6 +16,8 @@
 # - Provides a common src_install method to avoid code duplication
 #
 # Changelog:
+#   17 March 2022: Thomas Bracht Laumann Jespersen <t@laumann.xyz>
+#     - Port to EAPI 8
 #   03 January 2018: David Seifert <soap@gentoo.org>
 #     - Port to EAPI 6, remove built_with_use, simplify a lot
 #   12 March 2007: Jim Ramsay <lack@gentoo.org>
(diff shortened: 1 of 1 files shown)


laumann:
Working with @thesamesam  on this one. The plan is have the eclass support EAPI 8, then do EAPI bumps (fixing bugs along the way) for the users of this eclass.

Given that the eclass only supports EAPI 6 at the moment, it follows that all the users are also EAPI 6. There are some 27 users, all in x11-plugins except for one.

eclass/gkrellm-plugin.eclass
```diff
@@ -96,6 +102,5 @@ gkrellm-plugin_src_install() {
 		fi
 	fi
 }
-
 _GKRELLM_PLUGIN_R1=1
```
SoapGentoo, line 105: move this directly after `if [[ ! ${_GKRELLM_PLUGIN_R1} ]]; then`

eclass/gkrellm-plugin.eclass
```diff
@@ -55,7 +62,7 @@ EXPORT_FUNCTIONS src_install
 
 if [[ ! ${_GKRELLM_PLUGIN_R1} ]]; then
 
-DEPEND="virtual/pkgconfig"
+BDEPEND="virtual/pkgconfig"
```
thesamesam, line 65: BDEPEND doesn't exist in EAPI 6

eclass/gkrellm-plugin.eclass
```diff
@@ -55,7 +62,7 @@ EXPORT_FUNCTIONS src_install
```
SoapGentoo, line 62: put the `EXPORT_FUNCTIONS` as the last line of this file

SoapGentoo (changes requested):

eclass/gkrellm-plugin.eclass
laumann: Done. Added just before the closing `fi`

eclass/gkrellm-plugin.eclass
laumann: Done.

eclass/gkrellm-plugin.eclass
laumann: Ouch! Thanks, fixed!

eclass/gkrellm-plugin.eclass
```diff
@@ -51,11 +58,14 @@ esac
...
 
-DEPEND="virtual/pkgconfig"
+if [[ ${EAPI} == 6 ]]; then
+	DEPEND="virtual/pkgconfig"
+else
+	BDEPEND="virtual/pkgconfig"
+fi
 
 # @FUNCTION: gkrellm-plugin_src_install
 # @DESCRIPTION:
```
laumann, line 71: According to the [eclass writing guide](https://devmanual.gentoo.org/eclass-writing/index.html#eclass-functions) the `@USAGE:` is required, but I've spotted a few other eclasses where `@USAGE:` is missing (I think it's the EXPORTED_FUNCTIONS).

Should we add it?

eclass/gkrellm-plugin.eclass
```diff
@@ -97,5 +107,6 @@ gkrellm-plugin_src_install() {
 	fi
 }
 
-_GKRELLM_PLUGIN_R1=1
+EXPORT_FUNCTIONS src_install
```
SoapGentoo, line 110: at the very end, that is, it needs to be **outside** the `fi` inherit guard

eclass/gkrellm-plugin.eclass
SoapGentoo: just add it, yes

SoapGentoo (changes requested):

eclass/gkrellm-plugin.eclass
laumann: done

eclass/gkrellm-plugin.eclass
laumann: alright, fixed.

SoapGentoo: (have an idea why we require this?)

SoapGentoo (approved):

eclass/gkrellm-plugin.eclass
laumann: I'm not entirely sure, but I would guess that we want to ensure that gkrellm-plugin_src_install() is defined only once, but EXPORT_FUNCTIONS should be executed for each `inherit`.

But I would like to know!

SoapGentoo: https://bugs.gentoo.org/399039

laumann: Thanks!

laumann: The devmanual doesn't mention this in any way (but also doesn't give examples of an eclass that inherits another eclass). Could be worthwhile addition maybe?

laumann: Was reading https://devmanual.gentoo.org/eclass-writing/#export-functions

eclass/gkrellm-plugin.eclass
```diff
@@ -96,6 +107,6 @@ gkrellm-plugin_src_install() {
 		fi
 	fi
 }
-
```
SoapGentoo, line 99: put the one empty vertical whitespace line back

laumann: Done.

laumann:
Updated with a simplification to the EAPI guard, as suggested on the [mailing list](https://archives.gentoo.org/gentoo-dev/message/767e32d254d4eeede4935dfd2c7c170e)

PR #24628, opened 2022-03-17
Qt 5.15.3 version bump

a17r:


thesamesam:
Some iwdevtools results, not verified them (based on VDB vs ELF metadata):
```
dev-qt:qtbluetooth-5.15.3:20220317-230020.log:35:VDB: detected possibly incorrect RDEPEND (dev-qt/qtbluetooth-5.15.3)
dev-qt:qtbluetooth-5.15.3:20220317-230020.log-36-dev-qt/qtconcurrent  <
dev-qt:qtbluetooth-5.15.3:20220317-230020.log-37-dev-qt/qtdbus        | dev-qt/qtdbus:5
dev-qt:qtbluetooth-5.15.3:20220317-230020.log-38-dev-qt/qtdeclarative | dev-qt/qtdeclarative:5
dev-qt:qtbluetooth-5.15.3:20220317-230020.log-39-                     > dev-qt/qtnetwork:5
```

```
dev-qt:qthelp-5.15.3:20220317-232601.log:36:VDB: detected possibly incorrect RDEPEND (dev-qt/qthelp-5.15.3)
dev-qt:qthelp-5.15.3:20220317-232601.log-37-dev-qt/qtgui     | dev-qt/qtgui:5
dev-qt:qthelp-5.15.3:20220317-232601.log-38-dev-qt/qtnetwork <
dev-qt:qthelp-5.15.3:20220317-232601.log-39-dev-qt/qtsql     | dev-qt/qtsql:5
dev-qt:qthelp-5.15.3:20220317-232601.log-40-dev-qt/qtwidgets | dev-qt/qtwidgets:5
```

```
dev-qt:qtwayland-5.15.3:20220317-230954.log:70:VDB: detected possibly incorrect RDEPEND (dev-qt/qtwayland-5.15.3)
dev-qt:qtwayland-5.15.3:20220317-230954.log-71- > dev-libs/glib:2
dev-qt:qtwayland-5.15.3:20220317-230954.log-72- > dev-qt/qtdbus:5
dev-qt:qtwayland-5.15.3:20220317-230954.log-73- > media-libs/fontconfig
dev-qt:qtwayland-5.15.3:20220317-230954.log-74- > media-libs/freetype
dev-qt:qtwayland-5.15.3:20220317-230954.log-75- > x11-libs/libxkbcommon
```

The rest were exclusively just missing :5.

ionenwks:
Yeah the missing :5 aren't really important if it's =QT5_PV.

I tend to be careful with qt and "possibly" unnecessary `<` deps, they're often checked anyway or used in some other way, but maybe really unnecessary. The missing `>` deps tend to be a no-brainer addition, may need to check if USE-specific or automagic at most.

Pesa:
> Some iwdevtools results, not verified them (based on VDB vs ELF metadata): [...]

* **qtbluetooth**
  * qtconcurrent is a weird one (in Qt land) because it's _almost_ header-only. I guess qtbluetooth is using just those parts, so it doesn't need `libQt5Concurrent.so` at runtime. That being said, I don't think Qt makes any promise regarding what's header-only and what is not, so things can change at any time. I would recommend keeping the RDEPEND here. Moreover qtconcurrent is ~300 KB on disk.
  * qtnetwork is indeed used at runtime since Qt 5.12, nice catch!
* **qthelp**
  * The build system does require qtnetwork, but I can't see where it's actually used... so it could be bogus. We can move it to DEPEND. Or we can patch the build system to stop looking for qtnetwork and drop the dependency altogether.
* **qtwayland**
  * I'm not sure what's going on here...
  * I believe libxkbcommon is inherited via `libQt5XkbCommonSupport.a` (qtgui), but do we need to add an explicit dep in qtwayland as well?
  * The rest I have no idea where they're coming from... they're not directly referenced by qtwayland AFAICS.

a17r:
> * **qthelp**
>   
>   * The build system does require qtnetwork, but I can't see where it's actually used... so it could be bogus. We can move it to DEPEND. Or we can patch the build system to stop looking for qtnetwork and drop the dependency altogether.

https://invent.kde.org/qt/qt/qttools/-/merge_requests/2

a17r:
> > Some iwdevtools results, not verified them (based on VDB vs ELF metadata): [...]

Fixed `dev-qt/qtbluetooth` and `dev-qt/qthelp`, thanks.

> * **qtwayland**
>   
>   * I'm not sure what's going on here...
>   * I believe libxkbcommon is inherited via `libQt5XkbCommonSupport.a` (qtgui), but do we need to add an explicit dep in qtwayland as well?
>   * The rest I have no idea where they're coming from... they're not directly referenced by qtwayland AFAICS.

I don't see that either.

AndrewAmmerlaan:
ACK for Scidavis :+1:

Pesa (approved):

thesamesam (approved):
I was hoping to get time to look at the SSE2 etc bugs but I doubt I'm going to have a chance to do that before we want to get this in (I wanted to coalesce the pushes).

I suppose in theory I have a bit of time before unmasking to get some changes ready for review, but no promises.

a17r:
Added package.mask to go together with mail https://gist.github.com/a17r/b622b0be3e1e613f1f45c0b35c5196ee

a17r:
Changes:
- Moved `dev-qt/qtxmlpatterns` back to unversioned symlink again, didn'T forget to adapt `dev-qt/qtchooser` accordingly
- Fixed `dev-qt/qtlocation` obsolete mapbox hack and added upstream 5.15.3 regression patch
- Added `dev-qt/qdoc` and `dev-qt/qtquicktimeline` to clean slot 5 list

Pesa (approved):

PR #24635, opened 2022-03-18
dev-java/rhino: Change slot of version 1.7.14 

vaukai:


dev-java/rhino/rhino-1.7.14-r1.ebuild
```diff
@@ -15,7 +15,7 @@ SRC_URI="https://github.com/mozilla/rhino/archive/Rhino${PV//./_}_Release.tar.gz
 
 LICENSE="MPL-1.1 GPL-2"
 KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
-SLOT="1.6"
+SLOT="1.7"
```
thesamesam, line 18: Maybe we should do $(ver_cut 1-2) instead to avoid future errors?

vaukai: @thesamesam I am not yet sure if I really want the slot change, causing more work on the other revdeps ...
Let's see what @fordfrog  and @Leo3418  think about

Leo3418: My two-cents on this is, if this package follows semantic versioning very poorly in the sense that they **constantly** introduce API-incompatible changes even between `MINOR` versions, then using `$(ver_cut 1-2)` is sensible.

fordfrog: so the only package that broke so far with 1.7 is libreoffice because it expects different jar file name?

vaukai: > so the only package that broke so far with 1.7 is libreoffice because it expects different jar file name?

Exactly this seems to be the case. 'plowshare' does fine and I am just retesting 'gwt' and 'bsf'. 'batik-1.14' is another PR and builds upon 'rhino-1.7.14'.

fordfrog: then i guess the fix should be done in libreoffice (according to your report at the bug stating that now we have it as the upstream has it)

vaukai: > then i guess the fix should be done in libreoffice (according to your report at the bug stating that now we have it as the upstream has it)

That's my idea too. I'll close this PR then.

PR #17049, opened 2020-08-08
net-nds/389-ds-base: reintroduce with version 1.4.4.4, new maintainers

Dessa:
cc @vashirov @thesamesam 

superceds #16627 

for those who have looked at the old PR and not at the changes i did, i will explain the changes i made, diff is against that very PRs last version

- ~~changed *ID to 389 to match Fedora~~ *ID needs to change to whatever is available as 389 is now taken
- add rust support, for now in a useflag but it will be mandatory at some point
- add accountpolicy useflag
- add tmpfiles.d for non-systemd
- add lib389, as a requirement for the installed python scripts
- changed to slotting. i opted for "1.4" since i was told by upstream that a replica between any 1.4 version would not cause issues. 1.5, or whatever the next version of that style would be will be in its own slot (but not co-installable). under that assumption i wanted to save myself a slotmove if that is ok.
- changed BDB dependency so 5.3 will come first as its the version that is preferred by upstream (as fedora uses it) but older versions will still build, comment applies even tho the linked bug is about openldap
- dependency changes
  - pin pam dependency to pam-passthru useflag
  - add some slot deps
  - raise icu dep matching configure
  - add missing selinux,libevent deps
  - move DEPEND to bdepend
  - perl module deps are gone, perl itself is still in BDEPEND as a perl script is called on build
  - moved blockers to RDEPEND
  - moved COMMON_DEPEND to DEPEND

```
--- 389-ds-base-1.4.4.4.ebuild.orig	2020-08-08 12:42:59.545795611 +0200
+++ 389-ds-base-1.4.4.4.ebuild	2020-08-07 19:41:10.000000000 +0200
@@ -3,83 +3,196 @@
 
 EAPI=7
 
-WANT_AUTOMAKE="1.16"
+CRATES="
+ansi_term-0.11.0
+atty-0.2.14
+autocfg-1.0.0
+base64-0.10.1
+bitflags-1.2.1
+byteorder-1.3.4
+cbindgen-0.9.1
+cc-1.0.54
+cfg-if-0.1.10
+clap-2.33.1
+fernet-0.1.3
+foreign-types-0.3.2
+foreign-types-shared-0.1.1
+getrandom-0.1.14
+hermit-abi-0.1.13
+itoa-0.4.5
+jobserver-0.1.21
+lazy_static-1.4.0
+libc-0.2.71
+log-0.4.8
+openssl-0.10.29
+openssl-sys-0.9.57
+paste-0.1.15
+paste-impl-0.1.15
[...]

PR #24637, opened 2022-03-18
net-nds/389-ds-base: version bump

expeditioneer:


Dessa:
diff against PR:
```
--- 389-ds-base-2.1.0.ebuild.1	2022-03-18 12:44:24.533743485 +0100
+++ 389-ds-base-2.1.0.ebuild	2022-03-18 12:34:18.431534009 +0100
@@ -1,7 +1,7 @@
 # Copyright 1999-2022 Gentoo Authors
 
-EAPI=7
+EAPI=8
 
 CRATES="
 	ahash-0.7.6
@@ -116,7 +116,7 @@
 DISTUTILS_SINGLE_IMPL=1
 
-inherit autotools distutils-r1 systemd tmpfiles db-use cargo
+inherit autotools distutils-r1 systemd tmpfiles cargo
 
 DESCRIPTION="389 Directory Server (core libraries and daemons)"
 HOMEPAGE="https://directory.fedoraproject.org/"
@@ -133,13 +133,12 @@
 # lib389 tests (which is most of the suite) can't find their own modules.
 RESTRICT="test"
 
-# always list newer first
 # Do not add any AGPL-3 BDB here!
 # See bug 525110, comment 15.
-BERKDB_SLOTS=( 5.3 )
 
 DEPEND="
 	>=app-crypt/mit-krb5-1.7-r100[openldap]
+	dev-db/lmdb:=
 	>=dev-libs/cyrus-sasl-2.1.19:2[kerberos]
 	>=dev-libs/icu-60.2:=
 	dev-libs/nspr
@@ -148,11 +147,9 @@
 	dev-libs/libpcre:3
 	dev-libs/openssl:0=
 	>=net-analyzer/net-snmp-5.1.2:=
-	>=net-nds/openldap[sasl]
-	|| (
-		$(for slot in ${BERKDB_SLOTS[@]} ; do printf '%s\n' "sys-libs/db:${slot}" ; done)
-	)
+	net-nds/openldap[sasl]
 	sys-libs/cracklib
+	sys-libs/db:5.3
 	|| ( sys-fs/e2fsprogs sys-libs/e2fsprogs-libs )
 	pam-passthru? ( sys-libs/pam )
 	selinux? (
@@ -176,7 +173,7 @@
 
 # perl dependencies are for logconv.pl
 RDEPEND="${DEPEND}
-	!net-nds/389-ds-base:0
+	!net-nds/389-ds-base:1.4
 	acct-user/dirsrv
 	acct-group/dirsrv
 	${PYTHON_DEPS}
@@ -243,7 +240,7 @@
 		--with-pythonexec="${PYTHON}"
 		--with-fhs
 		--with-openldap
-		--with-db-inc="$(db_includedir)"
+		--with-db-inc="${EPREFIX}"/usr/include/db5.3
 		--disable-cockpit
 	)
 
```
Please also in-place revbump 1.4 with a blocker against 2.1, just as i did here with 1.4

expeditioneer:
@Dessa if you are agree to the Code I would merge this PR.

Dessa:
please add the 2.1 blocker to the 1.4 ebuild as i said.
when that's done, it'll get my +1

Dessa (approved):

Dessa:
> Why is this slotted?
to avoid surprises with replication, i opted for mysql-like slotting initially #17049 comment 1

> also, revbump to add blockers.
i uh. said that but i didn't look close enough if it was done, my bad.

thesamesam:
Sorry, wrote previous comment on mobile, so bit terse.

I don't think the slots are really going to be useful for doing that -- they just make it look like they're parallel installable, even if they're not. I think it's confusing for UX. You can use subslost to pin to specific releases.

MySQL and MariaDB does this as does Firefox (I think it's less of an issue w/ Firefox because it's never even a library), but I think for MySQL and MariaDB it's particularly confusing given that PostgreSQL is _really_ slotted.

A news item should be given or something if we need people to exercise care. Slots don't really achieve anything special here -- the new version is still going to be installed if people have it in world w/o some more specific info, IMO.

PR #24638, opened 2022-03-18
profiles/base: mask USE=default-libcxx on sys-devel/clang due to ABI implications

Jannik2099:
Continuation of #24122 , which got closed by accident when I updated my fork...

Sorry for the delay, killed my IRC bouncer & file server for a few weeks.
Added a clang/libcxx subfeature (I hope correctly?) as suggested - I have no hard feelings whether to do a clang/libcxx or a clang/libstdc++ instead to preserve existing users, feedback welcome.


thesamesam:
I thought we agreed to ewarn instead to avoid breaking existing systems?

Jannik2099:
I don't remember seeing strong agreement for anything, hence asking.

What I'm picking up as most desirable so far is:
Mask on default + news item for affected users
Keep enabled on clang profile
Mask on new clang/libstdc++ suprofile

?

LinuxUserGD:
> Mask on default + news item for affected users Keep enabled on clang profile Mask on new clang/libstdc++ suprofile

If libc++ is removed from the default clang profile, would clang with libstdc++ depend on `sys-devel/gcc` instead?
(see [packages](https://github.com/gentoo/gentoo/pull/24638/files#diff-ff7d67d3704eb642189be293a030759faeb59ed47e8eb03bd23c881dcd6397acL5) and [packages.build](https://github.com/gentoo/gentoo/pull/24638/files#diff-29f452def752976443289d2d5c882be0e04864d6862c0147f21681ad9d9f6c8bL3))

PR #24642, opened 2022-03-18
dev-lang/python: save/restore PYTHONDONTWRITEBYTECODE for USE=pgo

thesamesam:
This avoids writing bytecode when we call has_version (or make
any other calls). But we do need it on for the 'emake' call
when building w/ USE=pgo.

Save & restore the value from the environment before/after
emake if building with PGO.

Closes: https://bugs.gentoo.org/831897

### DIFF
--- a/dev-lang/python/python-3.10.3.ebuild
+++ b/dev-lang/python/python-3.10.3.ebuild
@@ -236,6 +236,11 @@ src_compile() {
 	# https://bugs.gentoo.org/823728
 	export SETUPTOOLS_USE_DISTUTILS=stdlib
 
+	# Save PYTHONDONTWRITEBYTECODE so that 'has_version' doesn't
+	# end up writing bytecode & violating sandbox.
+	# bug #831897
+	local -x _PYTHONDONTWRITEBYTECODE=${PYTHONDONTWRITEBYTECODE}
+
 	if use pgo ; then
 		# bug 660358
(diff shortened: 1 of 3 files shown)


dev-lang/python/python-3.10.3.ebuild
```diff
@@ -236,6 +236,11 @@ src_compile() {
 	# https://bugs.gentoo.org/823728
 	export SETUPTOOLS_USE_DISTUTILS=stdlib
 
+	# Save PYTHONDONTWRITEBYTECODE so that 'has_version' doesn't
+	# end up writing bytecode & violating sandbox.
+	# bug #831897
+	local -x _PYTHONDONTWRITEBYTECODE=${PYTHONDONTWRITEBYTECODE}
```
mgorny, line 242: Is there any reason to actually export it and not just keep it as a short local var?

thesamesam: No, we can keep it as a short local. Any preference for name?

thesamesam:
I've merged as-is for now given somebody confirmed this resolved their issue & I don't really want to leave a sandbox issue left unfixed for much longer, but happy to change it if you want.

dev-lang/python/python-3.10.3.ebuild
mgorny: `orig_bytecode`?

PR #24644, opened 2022-03-18
*.eclass: @ECLASS-VARIABLE renamed to @ECLASS_VARIABLE

ulm:


mgorny:
Any reason for noci? I was curious to see whether it fails.

ulm:
> Any reason for noci? I was curious to see whether it fails.

I expected it to fail, but let's find this out experimentally.

SoapGentoo (approved):

thesamesam (approved):

PR #24647, opened 2022-03-18
python-r1.eclass: use python_has_version in examples of python_check_deps()

whitslack:
The `python_check_deps()` examples given in the documentation of several functions and variables in `python-r1.eclass` and `python-any-r1.eclass` are incorrect. The default behavior of `has_version()` is to check the *runtime* system (`RDEPEND`), but `python_check_deps()` is meant to check that the *build* system (`BDEPEND`) has a suitable Python environment. The examples fail to do this and thus likely are leading developers to write incorrect `python_check_deps()` functions in their ebuilds. Thankfully, `python-utils-r1.eclass` supplies a `python_has_version()` function that defaults to checking the build system and implements some other enhancements that make it nicer for use in `python_check_deps()`. Change the examples to use `python_has_version`.

This PR fixes [Bug 835462](https://bugs.gentoo.org/835462).

### DIFF
--- a/eclass/python-r1.eclass
+++ b/eclass/python-r1.eclass
@@ -163,7 +163,7 @@ fi
 # Example use:
 # @CODE
 # python_check_deps() {
-# 	has_version "dev-python/bar[${PYTHON_SINGLE_USEDEP}]"
+# 	python_has_version "dev-python/bar[${PYTHON_SINGLE_USEDEP}]"
 # }
 # @CODE
 #
@@ -483,9 +483,9 @@ python_gen_impl_dep() {
 #		dev-python/baz[${PYTHON_USEDEP}] )' -2)"
 #
 # python_check_deps() {
-#	has_version "dev-python/foo[${PYTHON_SINGLE_USEDEP}]" \
-#		&& { has_version "dev-python/bar[${PYTHON_USEDEP}]" \
-#			|| has_version "dev-python/baz[${PYTHON_USEDEP}]"; }
+# 	python_has_version "dev-python/foo[${PYTHON_SINGLE_USEDEP}]" &&
+# 		{ python_has_version "dev-python/bar[${PYTHON_USEDEP}]" ||
+# 			python_has_version "dev-python/baz[${PYTHON_USEDEP}]"; }
 # }
 #
 # src_compile() {
@@ -683,7 +683,7 @@ python_foreach_impl() {
 #	$(python_gen_any_dep 'dev-python/epydoc[${PYTHON_USEDEP}]' 'python2*') )"
 #
 # python_check_deps() {
-#	has_version "dev-python/epydoc[${PYTHON_USEDEP}]"
+# 	use !doc || python_has_version "dev-python/epydoc[${PYTHON_USEDEP}]"
 # }
 #
 # src_compile() {
--- a/eclass/python-any-r1.eclass
+++ b/eclass/python-any-r1.eclass
@@ -139,7 +139,7 @@ EXPORT_FUNCTIONS pkg_setup
 # Example use:
 # @CODE
 # python_check_deps() {
-# 	has_version "dev-python/foo[${PYTHON_USEDEP}]"
+# 	python_has_version "dev-python/foo[${PYTHON_USEDEP}]"
 # }
 # @CODE
 #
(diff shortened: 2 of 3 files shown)


eclass/python-r1.eclass
```diff
@@ -683,7 +683,7 @@ python_foreach_impl() {
 #	$(python_gen_any_dep 'dev-python/epydoc[${PYTHON_USEDEP}]' 'python2*') )"
 #
 # python_check_deps() {
-#	has_version "dev-python/epydoc[${PYTHON_USEDEP}]"
+# 	use !doc || python_has_version "dev-python/epydoc[${PYTHON_USEDEP}]"
```
mgorny, line 686: I think the newer examples prefer:

```
use doc || return 0
python_has_version ...
```

whitslack: Okay, I'll change it to that for consistency. It's functionally identical, though, right? And what if you had multiple different sets of use flags to check?
```
BDEPEND="$(python_gen_any_dep '
    foo? ( dev-python/foo[${PYTHON_USEDEP}] )
    bar? ( dev-python/bar[${PYTHON_USEDEP}] )
')"
```
Then what would you put in `python_check_deps()`? I like my way because it structurally mirrors the syntax of `BDEPEND`.
```
python_check_deps() {
    { use !foo || python_has_version "dev-python/foo[${PYTHON_USEDEP}]"; } &&
    { use !bar || python_has_version "dev-python/bar[${PYTHON_USEDEP}]"; }
}
```

mgorny: But that thing is horribly hard to read. Even the `foo && { bar || baz; }` thing is ugly. To be honest, I'm wondering if maybe we should just remove the examples from eclassdoc and expect people to read the Guide.

whitslack: > But that thing is horribly hard to read.

Is it? Maybe for a complete n00b of a programmer. But again, I really was asking what you would put in `python_check_deps()` in such a case. Would you write it all out like this?:
```
python_check_deps() {
    if use foo ; then
        python_has_version "dev-python/foo[${PYTHON_USEDEP}]" || return $?
    fi
    if use bar ; then
        python_has_version "dev-python/bar[${PYTHON_USEDEP}]" || return $?
    fi
}
```
Are you really arguing that that's less "horrible" than the compact form that I wrote using `&&`?

Anyway, I find the whole mechanism of `python_check_deps()` really awkward, as it requires duplicating the `python_gen_any_dep` dependencies of the package but with different syntax. It would be so much nicer if an ebuild author could write:
```
BDEPEND="$(python_gen_any_dep '
    dev-python/foo[${PYTHON_SINGLE_USEDEP}]
    || ( dev-python/bar[${PYTHON_USEDEP}]
         dev-python/baz[${PYTHON_USEDEP}] )
')"

python_get_any_deps() {
    echo "
        dev-python/foo[${PYTHON_SINGLE_USEDEP}]
        || ( dev-python/bar[${PYTHON_USEDEP}]
             dev-python/baz[${PYTHON_USEDEP}] )
    "
}
```
And then the eclass would handle parsing of the returned dependency spec, including the `|| ( … )` syntax. This would allow a straight copy+paste of the value of the `python_gen_any_dep` argument to the `python_get_any_deps()` output, with just a change from single-quoting to double-quoting to enable variable substitution.

This solution could co-exist with `python_check_deps()` simply by having the eclass define a default `python_check_deps()` that calls `python_get_any_deps` and checks the returned dependency spec if `python_get_any_deps` is defined or else does nothing if it's not defined. Existing ebuilds would be overriding the default `python_check_deps()` and thus would continue to work exactly as they do now.

whitslack: I have implemented the `python_get_any_deps()` alternative in #24677. Please have a look there.

mgorny:
I'm sorry, I have forgotten about this. I'll merge it now since the doc changes are good.

thesamesam:
commit f80b0acf7f78971b601aabb913e067a0dde70268
Author: Matt Whitlock <gentoo@mattwhitlock.name>
Date:   Fri Mar 18 17:44:05 2022 -0400

    python{,-any}-r1.eclass: use python_has_version in examples

    The python_check_deps() examples given in the documentation of several
    functions and variables in python-r1.eclass and python-any-r1.eclass are
    incorrect. The default behavior of has_version() is to check the
    *runtime* system (RDEPEND), but python_check_deps() is meant to check
    that the *build* system (BDEPEND) has a suitable Python environment. The
    examples fail to do this and thus likely are leading developers to write
    incorrect python_check_deps() functions in their ebuilds. Thankfully,
    python-utils-r1.eclass supplies a python_has_version() function that
    defaults to checking the build system and implements some other
    enhancements that make it nicer for use in python_check_deps(). Change
    the examples to use python_has_version.

    Closes: https://bugs.gentoo.org/835462
    Signed-off-by: Matt Whitlock <gentoo@mattwhitlock.name>
    Signed-off-by: Michał Górny <mgorny@gentoo.org>

PR #24650, opened 2022-03-19
dev-lang/erlang: Version updated to 24.3.2

jpds:


MatthewGentoo (approved):
Adding, but I've kept version 24.3 so that we can get it stable :)

PR #24659, opened 2022-03-19
sys-devel/dev86: support compiling sys-devel/dev86-0.16.21-r3 with clang

dpronin:
patched bootblocks/makeboot.c of sys-devel/dev86-0.16.21-r3

Signed-off-by: Denis Pronin <dannftk@yandex.ru>


juippis (approved):
Thanks!

PR #24660, opened 2022-03-19
AMDGPU AMF fixes

AdelKS:
I already had in mind to work on what has been reported on bug https://bugs.gentoo.org/834974, so here it is!

Something I need the reviewers to pay special attention to:
- I restricted to `video_cards_amdgpu` only at the level of `amdgpu-pro-vulkan` ebuild and I expect this to just propagated to the things that depend on it. I did not rev bump the packages as I think it's not needed.
- `amdgpu-pro-amf` only provides binaries for `amd64`, so ffmpeg `abi_x86_32` cannot use it ? But it can be built nonetheless since `amdgpu-pro-amf` is only an RDEPEND (at least on my machine it builds just fine) ?
  - to make `ffmpeg` be able to use AMF, it needs the env variable `VK_ICD_FILENAMES=/usr/share/vulkan/icd.d/amd_pro_icd64.json` so I made that with `doenvd`.

Something  I don't understand quite well is how `abi_x86_32` and `abi_x86_64` affects the fffmpeg binary, I initally thought two binaries would be produced, but only one seems to be produced. It uses either 32bit shared libs or 64bit shared libs, depending on what mode it's called on ? Any reading would be welcome!

### DIFF
--- a/media-libs/amdgpu-pro-vulkan/amdgpu-pro-vulkan-21.50.1.1376756.ebuild
+++ b/media-libs/amdgpu-pro-vulkan/amdgpu-pro-vulkan-21.50.1.1376756.ebuild
@@ -30,7 +30,9 @@ RESTRICT="bindist mirror"
 LICENSE="AMD-GPU-PRO-EULA"
 SLOT="0"
 KEYWORDS="-* ~amd64 ~x86"
-IUSE="abi_x86_32 abi_x86_64"
+IUSE="abi_x86_32 abi_x86_64 video_cards_amdgpu"
+
+REQUIRED_USE="video_cards_amdgpu"
 
 BDEPEND="dev-util/patchelf"
 
--- a/media-video/ffmpeg/ffmpeg-4.4.1-r4.ebuild
+++ b/media-video/ffmpeg/ffmpeg-4.4.1-r4.ebuild
@@ -172,6 +172,7 @@ IUSE="${IUSE} ${FFTOOLS[@]/#/+fftools_}"
 
 RDEPEND="
 	alsa? ( >=media-libs/alsa-lib-1.0.27.2[${MULTILIB_USEDEP}] )
+	amf? ( media-video/amdgpu-pro-amf )
 	amr? ( >=media-libs/opencore-amr-0.1.3-r1[${MULTILIB_USEDEP}] )
 	bluray? ( >=media-libs/libbluray-0.3.0-r1:=[${MULTILIB_USEDEP}] )
 	bs2b? ( >=media-libs/libbs2b-3.1.0-r1[${MULTILIB_USEDEP}] )
@@ -188,8 +189,8 @@ RDEPEND="
 		rav1e? ( >=media-video/rav1e-0.4:=[capi] )
 		snappy? ( >=app-arch/snappy-1.1.2-r1:=[${MULTILIB_USEDEP}] )
 		theora? (
-			>=media-libs/libtheora-1.1.1[encode,${MULTILIB_USEDEP}]
 			>=media-libs/libogg-1.3.0[${MULTILIB_USEDEP}]
+			>=media-libs/libtheora-1.1.1[encode,${MULTILIB_USEDEP}]
 		)
 		twolame? ( >=media-sound/twolame-0.3.13-r1[${MULTILIB_USEDEP}] )
 		webp? ( >=media-libs/libwebp-0.3.0:=[${MULTILIB_USEDEP}] )
@@ -249,10 +250,10 @@ RDEPEND="
 		gnome-base/librsvg:2=[${MULTILIB_USEDEP}]
 		x11-libs/cairo[${MULTILIB_USEDEP}]
 	)
+	nvenc? ( >=media-libs/nv-codec-headers-9.1.23.1[${MULTILIB_USEDEP}] )
 	svt-av1? ( >=media-libs/svt-av1-0.8.4[${MULTILIB_USEDEP}] )
 	truetype? ( >=media-libs/freetype-2.5.0.1:2[${MULTILIB_USEDEP}] )
 	vaapi? ( >=x11-libs/libva-1.2.1-r1:0=[${MULTILIB_USEDEP}] )
-	nvenc? ( >=media-libs/nv-codec-headers-9.1.23.1[${MULTILIB_USEDEP}] )
 	vdpau? ( >=x11-libs/libvdpau-0.7[${MULTILIB_USEDEP}] )
 	vidstab? ( >=media-libs/vidstab-1.1.0[${MULTILIB_USEDEP}] )
 	vmaf? ( media-libs/libvmaf[${MULTILIB_USEDEP}] )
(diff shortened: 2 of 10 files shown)


media-libs/amdgpu-pro-vulkan/amdgpu-pro-vulkan-21.50.1.1376756.ebuild
```diff
@@ -30,7 +30,9 @@ RESTRICT="bindist mirror"
 LICENSE="AMD-GPU-PRO-EULA"
 SLOT="0"
 KEYWORDS="-* ~amd64 ~x86"
-IUSE="abi_x86_32 abi_x86_64"
+IUSE="abi_x86_32 abi_x86_64 video_cards_amdgpu"
+
+REQUIRED_USE="video_cards_amdgpu"
```
juippis, line 35: While this isn't wrong, I wonder if this is right either. Maybe you just want to depend on x11-drivers/xf86-video-amdgpu?

media-video/ffmpeg/ffmpeg-4.4.1-r4.ebuild
```diff
@@ -273,6 +273,7 @@ RDEPEND="
 	zlib? ( >=sys-libs/zlib-1.2.8-r1[${MULTILIB_USEDEP}] )
 	zvbi? ( >=media-libs/zvbi-0.2.35[${MULTILIB_USEDEP}] )
 	postproc? ( !media-libs/libpostproc )
+	amf? ( media-video/amdgpu-pro-amf )
```
juippis, line 276: Could you put this into its correct place alphabetically?

media-video/ffmpeg/ffmpeg-4.4.1-r4.ebuild
```diff
@@ -563,4 +564,6 @@ multilib_src_install() {
 multilib_src_install_all() {
 	dodoc Changelog README.md CREDITS doc/*.txt doc/APIchanges
 	[ -f "RELEASE_NOTES" ] && dodoc "RELEASE_NOTES"
+
+	[ use amf ] && doenvd ${FILESDIR}/amf-env-vulkan-override
```
juippis, line 568:
```suggestion
	use amf && doenvd "${FILESDIR}"/amf-env-vulkan-override
```

media-video/ffmpeg/ffmpeg-4.4.1-r4.ebuild
```diff
@@ -563,4 +564,6 @@ multilib_src_install() {
 multilib_src_install_all() {
 	dodoc Changelog README.md CREDITS doc/*.txt doc/APIchanges
 	[ -f "RELEASE_NOTES" ] && dodoc "RELEASE_NOTES"
```
juippis, line 566: (I know this isn't your ebuild, but I want you to know we prefer double brackets, `[[ ]]`

media-video/ffmpeg/ffmpeg-5.0-r2.ebuild
```diff
@@ -98,7 +98,7 @@ FFMPEG_ENCODER_FLAG_MAP=(
 	amrenc:libvo-amrwbenc mp3:libmp3lame
 	kvazaar:libkvazaar libaom
 	openh264:libopenh264 rav1e:librav1e snappy:libsnappy svt-av1:libsvtav1 theora:libtheora twolame:libtwolame
-	webp:libwebp x264:libx264 x265:libx265 xvid:libxvid
+	webp:libwebp x264:libx264 x265:libx265 xvid:libxvid amf:amf
```
juippis, line 101: This also looks somewhat alphabetically sorted list.

media-video/ffmpeg/ffmpeg-5.0-r2.ebuild
```diff
@@ -291,6 +292,7 @@ BDEPEND="
 	cuda? ( >=sys-devel/clang-7[llvm_targets_NVPTX] )
 	doc? ( sys-apps/texinfo )
 	test? ( net-misc/wget sys-devel/bc )
+	amf? ( media-libs/amf-headers )
```
juippis, line 295: This too.

media-video/ffmpeg/ffmpeg-5.0-r2.ebuild
```diff
@@ -562,4 +564,6 @@ multilib_src_install() {
 multilib_src_install_all() {
 	dodoc Changelog README.md CREDITS doc/*.txt doc/APIchanges
 	[ -f "RELEASE_NOTES" ] && dodoc "RELEASE_NOTES"
+
+	[ use amf ] && doenvd ${FILESDIR}/amf-env-vulkan-override
```
juippis, line 568: Fix this too.

media-video/ffmpeg/ffmpeg-9999.ebuild
```diff
@@ -273,6 +273,7 @@ RDEPEND="
 	zlib? ( >=sys-libs/zlib-1.2.8-r1[${MULTILIB_USEDEP}] )
 	zvbi? ( >=media-libs/zvbi-0.2.35[${MULTILIB_USEDEP}] )
 	postproc? ( !media-libs/libpostproc )
+	amf? ( media-video/amdgpu-pro-amf )
```
juippis, line 276: Ditto.

media-video/ffmpeg/ffmpeg-9999.ebuild
```diff
@@ -562,4 +563,6 @@ multilib_src_install() {
 multilib_src_install_all() {
 	dodoc Changelog README.md CREDITS doc/*.txt doc/APIchanges
 	[ -f "RELEASE_NOTES" ] && dodoc "RELEASE_NOTES"
+
+	[ use amf ] && doenvd ${FILESDIR}/amf-env-vulkan-override
```
juippis, line 567: And here.

media-video/ffmpeg/files/amf-env-vulkan-override
```diff
@@ -0,0 +1 @@
+VK_ICD_FILENAMES=/usr/share/vulkan/icd.d/amd_pro_icd64.json
```
juippis, line 1: Will this work with other arches beside amd64?

media-libs/amdgpu-pro-vulkan/amdgpu-pro-vulkan-21.50.1.1376756.ebuild
AdelKS: I want to say that this package only works on systems running on an AMD GPU whose driver is `amdgpu`. I asked in the dev IRC channel and I got suggested this solution and I liked it.

> Maybe you just want to depend on x11-drivers/xf86-video-amdgpu?

I don't want to pull it in particular, although maybe it's needed ?

media-video/ffmpeg/ffmpeg-4.4.1-r4.ebuild
AdelKS: Thanks ! I would love to have a documentation that says which variables need to be quoted and which ones don't x)

media-video/ffmpeg/files/amf-env-vulkan-override
AdelKS: So basically, `ffmpeg` can be built with `amf` support in any arch a priori. But as of now it needs a library that is only offered by the `amd64` `amdgpu-pro-amf` package. So the `amf` useflag is masked in every arch except `amd64`. 

A little subtlety that I am not sure about is that `ffmpeg` can be built in multilib, abi 32 and 64, with `amf` support. But I don't know if the `amf` code path would work in the 32 bit abi. But I need to enable building ffmpeg with amf support in abi 32 otherwise, to have a multilib amf, one would need to disable the `amf` flag.

juippis (approved):

PR #24670, opened 2022-03-19
dev-util/desktop-file-utils: Backport support for spec 1.5

tastytea:
Add patch from upstream to enable support for version 1.5 of the Desktop
Entry Specification. Also bump EAPI and fix HOMEPAGE.

Upstream patch:
https://gitlab.freedesktop.org/xdg/desktop-file-utils/-/merge_requests/11 (copied
without modification).

Closes: https://bugs.gentoo.org/795570


dev-util/desktop-file-utils/files/desktop-file-utils-0.26-support-version-1.5.patch
```diff
@@ -0,0 +1,83 @@
+From 425177a28b6215e0745f95100160a08e810fd47c Mon Sep 17 00:00:00 2001
```
thesamesam, line 1: Nit: include a link to the commit at the top?

tastytea: Added.

thesamesam:
Great work!

PR #24671, opened 2022-03-19
sci-astronomy/wcslib: version bump 7.7

hamarituc:
* moved `flex` and `pkgconfig` to BDEPEND
* fixed race condition during test phase

Closes: https://bugs.gentoo.org/834504


sci-astronomy/wcslib/wcslib-7.7.ebuild
```diff
@@ -0,0 +1,75 @@
...
+		)
+	else
+		myconf+=( --without-pgplot )
+	fi
+	econf "${myconf[@]}"
+	sed -i -e 's/COPYING\*//' GNUmakefile || die
+}
+
+src_test() {
+	# -j1 to work around a race condition
```
thesamesam, line 63: Can this go now with your patch?

hamarituc: Unfortunately not. This workaround and the patch address different issues in the test suite.

PR #24677, opened 2022-03-20
python*-r1.eclass: introduce python_get_any_deps() function

whitslack:
This PR supersedes #24647.

While very powerful, the existing `python_check_deps()` mechanism is clunky to use, as it requires duplicating the dependency block passed to `python_gen_any_dep()` but using shell syntax that is ugly and may admit subtle differences in the logic. Indeed, the examples in the eclassdoc were wrong: all of them were calling `has_version` without `-b`, so they were checking the wrong sysroot, and one of them was inadventently missing a USE-conditional that was present in the corresponding `BDEPEND`.

As an optional alternative to `python_check_deps()`, introduce an ebuild-defined `python_get_any_deps()` function that prints to stdout the same dependency specification as the ebuild passes to `python_gen_any_dep()`. This is easier to use correctly and should cover practically all cases. Update the examples in the eclassdoc to use this new alternative instead of `python_check_deps()`.

### DIFF
--- a/eclass/python-any-r1.eclass
+++ b/eclass/python-any-r1.eclass
@@ -30,6 +30,13 @@
 # implementation fulfills the requirements, a false value (non-zero)
 # otherwise.
 #
+# As an alternative to (or in addition to) python_check_deps(), you can
+# define a python_get_any_deps() function, which should print to stdout
+# a dependency specification using the same syntax as in the *DEPEND
+# variables. The eclass will call python_get_any_deps() with EPYTHON set
+# to each matching Python implementation and with PYTHON_USEDEP and
+# PYTHON_SINGLE_USEDEP defined appropriately.
+#
 # Please note that python-any-r1 will always inherit python-utils-r1
 # as well. Thus, all the functions defined there can be used in the
 # packages using python-any-r1, and there is no need ever to inherit
@@ -130,7 +137,8 @@ EXPORT_FUNCTIONS pkg_setup
 # @OUTPUT_VARIABLE
 # @DESCRIPTION:
 # An eclass-generated USE-dependency string for the currently tested
-# implementation. It is set locally for python_check_deps() call.
+# implementation. It is set locally for python_check_deps() and
+# python_get_any_deps() calls.
 #
 # The generated USE-flag list is compatible with packages using
 # python-r1 eclass. For python-single-r1 dependencies,
@@ -138,8 +146,8 @@ EXPORT_FUNCTIONS pkg_setup
 #
 # Example use:
 # @CODE
-# python_check_deps() {
-# 	has_version "dev-python/foo[${PYTHON_USEDEP}]"
+# python_get_any_deps() {
+# 	echo "dev-python/foo[${PYTHON_USEDEP}]"
 # }
 # @CODE
 #
@@ -152,7 +160,8 @@ EXPORT_FUNCTIONS pkg_setup
 # @OUTPUT_VARIABLE
 # @DESCRIPTION:
 # An eclass-generated USE-dependency string for the currently tested
-# implementation. It is set locally for python_check_deps() call.
+# implementation. It is set locally for python_check_deps() and
+# python_get_any_deps() calls.
 #
 # The generated USE-flag list is compatible with packages using
 # python-single-r1 eclass. For python-r1 dependencies,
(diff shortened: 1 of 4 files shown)


whitslack:
An example of the kind of niceness this enables in an ebuild:
```
python_get_any_deps() {
	echo "
		doc? (
			>=dev-python/repoze-sphinx-autointerface-0.4[${PYTHON_USEDEP}]
			dev-python/sphinx[${PYTHON_USEDEP}]
		)
		|| (
			dev-python/pdm-pep517[${PYTHON_USEDEP}]
			dev-python/poetry-core[${PYTHON_USEDEP}]
			dev-python/setuptools[${PYTHON_USEDEP}]
		)
	"
}

BDEPEND="
	$(python_gen_any_dep "$(
		PYTHON_SINGLE_USEDEP='${PYTHON_SINGLE_USEDEP}' \
		PYTHON_USEDEP='${PYTHON_USEDEP}' \
		python_get_any_deps
	)")
"
```

And then at build-time, you might have:
```
>>> Compiling source in /var/tmp/portage/dev-python/proof-of-concept-0/work ...
 * Checking whether python3_10 is suitable ...
 *   >=dev-lang/python-3.10.0_p1-r1:3.10 ...                                          [ ok ]
 *   python_get_any_deps ...
 *     >=dev-python/repoze-sphinx-autointerface-0.4[python_targets_python3_10(-)] ... [ !! ]
 * Checking whether python3_9 is suitable ...
 *   >=dev-lang/python-3.9.9-r1:3.9 ...                                               [ ok ]
 *   python_get_any_deps ...
 *     >=dev-python/repoze-sphinx-autointerface-0.4[python_targets_python3_9(-)] ...  [ ok ]
 *     dev-python/sphinx[python_targets_python3_9(-)] ...                             [ ok ]
 *     dev-python/pdm-pep517[python_targets_python3_9(-)] ...                         [ !! ]
 *     dev-python/poetry-core[python_targets_python3_9(-)] ...                        [ ok ]
 * Using python3.9 in global scope
```

mgorny:
I'm sorry but no. The eclasses are complex enough as they are, and adding a dependency string parser in bash is not a viable solution.

whitslack:
> I'm sorry but no. The eclasses are complex enough as they are, and adding a dependency string parser in bash is not a viable solution.

Wow, that's incredibly closed-minded. The implementation I provided is compact and, as far as I'm aware, complete and correct. The whole point of eclasses is to move complexity out of ebuilds to a central place where it can be checked for correctness once, rather than having many ebuilds repeating the same mistakes. And I'm sure ebuilds are repeating the same mistakes. Hell, the `scons-utils.eclass` wasn't even correct.

If your objection is really just, "I don't understand that code, so I hate it," then that's very immature.

mgorny:
No, it's not and I find your comment offensive.

whitslack:
And I find your unceremonious, unsubstantiated, and unconditional dismissal of my novel contribution insulting, so it seems we are at an impasse. Good day, sir.

mattst88:
Okay, let's reset.

I agree that the `python_check_deps()` function is clunky, requiring duplicating package lists. Deduplicating that is good. I also agree that it's nice to print the dependencies when checking python version suitability. I think both of those are good goals.

At the same time, `_python_check_dep` is pretty scary.

Can we find a way to accomplish those two goals without parsing dependency specifications in bash?

thesamesam:
I admire the aim here (and I don't think the resultant interface is bad, but it looks a _bit_ clunky because of the Bash parsing needed).

I'm wondering  if we can compromise on a case which lets us win in the majority of cases -- defining a simple Bash array for BDEPEND-only stuff pre-inherit, including it in the eclass, and defining a simple python_check_deps default when it exists.

This avoids duplication between BDEPEND & python_check_deps for the easy cases.

whitslack:
> I also agree that it's nice to print the dependencies when checking python version suitability.

The existing `python_has_version` does that.

> This avoids duplication between BDEPEND & python_check_deps for the easy cases.

The easy cases are already easy. See [here](https://gitlab.com/bitcoin/gentoo/-/blob/76d26173b40257a496992721b8f33516bf052f58/dev-python/txtorcon/txtorcon-22.0.0.ebuild#L20-51) for an example with no duplication.

The tricky cases are when some of the build-time dependencies are USE-conditional or any-of-many.

PR #24684, opened 2022-03-20
net-im/telegram-desktop: Bump to 3.6.1

mid-kid:
@gyakovlev


arthurzam (approved):
We mostly prefer that the commits are separated (one doing bump, next doing drop) - makes it easier to revert the drop.
But will accept it as done currently.

juippis:
Also makes it possible to review the ebuild, since the old ebuilds can carry some old unwanted cruft with bumps. But this particular package is rather active so it's not EAPI-5 etc.

PR #24687, opened 2022-03-20
sys-libs/db: drop USE=java

thesamesam:
Flag was last-rited a while ago.

Closes: https://bugs.gentoo.org/713866
Closes: https://bugs.gentoo.org/615408


sys-libs/db/db-18.1.32-r1.ebuild
```diff
@@ -1,245 +0,0 @@
-# Copyright 1999-2022 Gentoo Authors
-# Distributed under the terms of the GNU General Public License v2
-
-EAPI=6
-inherit db flag-o-matic java-pkg-opt-2 autotools multilib multilib-minimal eapi7-ver toolchain-funcs
```
SoapGentoo, line 5: bring this ebuild back

SoapGentoo (changes requested):

SoapGentoo (approved):

PR #24692, opened 2022-03-21
app-containers/k3d: fix k3s version

meierfra:
my k3d tried to download a nonexistent k3s image. the reason for this seems to be, that the version in the ebuild has a suffix starting with '+' . After changing the ebuild locally my k3d worked flawlessly.


zmedico:
Thanks!

PR #24697, opened 2022-03-21
profiles/package.mask: Keep dev-java/jrobin

vaukai:
@jsmolic 
This package is EAPI 8

### DIFF
--- a/profiles/package.mask
+++ b/profiles/package.mask
@@ -44,7 +44,6 @@ dev-java/invokebinder
 dev-java/java-apicheck
 dev-java/jboss-marshalling-river
 dev-java/jboss-marshalling-serial
-dev-java/jrobin
 
 # David Seifert <soap@gentoo.org> (2022-03-21)
 # EAPI 5, last release 15 years ago, QA permission issues, no other


jsmolic:
Can you drop the EAPI 5 ebuild while at it?

ionenwks:
> Can you drop the EAPI 5 ebuild while at it?

It's still waiting for stabilization, but the EAPI=8 -r2 is from Feb 18 so it'd be about time to.

vaukai:
> > Can you drop the EAPI 5 ebuild while at it?
> 
> It's still waiting for stabilization, but the EAPI=8 -r2 is from Feb 18 so it'd be about time to.

author | Volkmar W. Pogatzki <gentoo@pogatzki.net> | 2022-02-18 16:35:25 +0100
committer | Sam James <sam@gentoo.org> | 2022-03-16 22:28:01 +0000

So we could stabilize on April 15, right?

ionenwks:
Yeah, see my edit. I noticed the right date after. Although note it's not always necessary to wait the full 30 days, I often do less when it's just ebuild changes / revbumps unless it's something big.

PR #24699, opened 2022-03-21
app-text/zathura-pdf-mupdf: fix sed call

rndxelement:
Fix sed call removing all lines mentioning mupdfthid in src_prepare()

Closes: https://bugs.gentoo.org/835592


thesamesam:
This feels like a good candidate for a patch given it can lead to silent runtime breakage if the context changes.

Also, link to the bug above the sed line to note to be careful (saying this is why we don't do mupdfthird).

rndxelement:
> This feels like a good candidate for a patch given it can lead to silent runtime breakage if the context changes.
> 
> Also, link to the bug above the sed line to note to be careful (saying this is why we don't do mupdfthird).

Thanks, that's a good idea. I also added a revision for the 0.3.8 ebuild, because it needs to rebuild for people that have it installed already

PR #24702, opened 2022-03-22
net-misc/rabbitmq-server: remove net dep from init

tgbugs:
rabbitmq does not have a hard requirement on the net service
therefore remove the dependency which causes the service to
fail to start if net has not been started or is running in an env
where there is not explicit net service (e.g. in a container).
Closes: https://bugs.gentoo.org/787815
Signed-off-by: Tom Gillespie <tgbugs@gmail.com>


ultrabug:
@dargor FYI

ultrabug:
Thank you

PR #24703, opened 2022-03-22
net-misc/tigervnc: Fix the init script

ceamac:
The ebuild is unchanged, only a revbump.  The changes are in `files/`

Remove $VNC_OPTS from conf.d and add a message instead
Since I had to change the init script, I also ran `checkbashisms` and `shellcheck` on it.

I only fixed `tigervnc-1.12`, 1.9 is untouched.


net-misc/tigervnc/files/tigervnc-1.12.0.initd
```diff
@@ -59,10 +55,10 @@ stop() {
...
 		# killing vncsession does not work, we have to kill Xvnc
 		# run pstree to see exactly what's going on
-		sessionpid=`cat /run/vncsession-\:${user##*:}.pid`
-		serverpid=`pstree -p $sessionpid | grep Xvnc | sed -e 's/^.*Xvnc(//
-			s/).*$//'`
-		kill $serverpid &>/dev/null
+		sessionpid=$(cat /run/vncsession-:"${user##*:}".pid)
+		serverpid=$(pstree -p "$sessionpid" | grep Xvnc | sed -e 's/^.*Xvnc(//
+			s/).*$//')
+		kill "$serverpid" >/dev/null 2>&1
```
tillschaefer, line 61: I guess sessionpid and serverpid should never contain whitespace or special characters. Thus, the quotations marks seem superfluous to me.

ceamac: Yes, I added them only to silence `shellcheck`.  They don't do any harm anyway.

thesamesam: Overquoting is generally better than underquoting, yep.

PR #24704, opened 2022-03-22
Fixed work with sddm

serg-sg:


a17r:
"Fixed work with sddm"?

serg-sg:
If the system uses sddm, then the service start fails. This fix in the init script solves the problem.

puleglot:
I was thinking about dropping this script entirely since it cannot be adapted to work with many modern login managers which launches separate user X sessions. Not sure why people still want to use it and avoid vnc tunneling via ssh. :)

x11-misc/x11vnc/files/x11vnc.init.d-r1
```diff
@@ -29,8 +29,13 @@ checkconfig() {
 	fi
 	checkpath -q -f -m 0600 -o root:root "${X11VNC_RFBAUTH}"
 
+	# Attempt to find SDDM
+	if [ -d /var/run/sddm/ ]; then
+		# If SDDM is found
+		X11VNC_AUTH_OPTS="-auth /var/run/sddm/*"
+		X11VNC_OPTS="-xkb -noxrecord"
```
puleglot, line 36: Overriding `X11VNC_OPTS` completely is a bad idea. Could you explain why these options are required here?

serg-sg: You're right. Now these parameters are applied by default and will be superfluous here. Thanks for the good note. Remove the bad line.

PR #24709, opened 2022-03-22
media-plugins/gkrellmpc: update EAPI 6 -> 8

laumann:
Convert src_prepare() to use the PATCHES array, and calling default
instead of eapply_user().

### DIFF
--- a/media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
+++ b/media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
@@ -1,0 +1,43 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit gkrellm-plugin toolchain-funcs
+
+DESCRIPTION="A gkrellm plugin to control the MPD (Music Player Daemon)"
+HOMEPAGE="http://mpd.wikia.com/wiki/Client:GKrellMPC"
+SRC_URI="http://mina.naguib.ca/dist/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE="threads"
+
+RDEPEND="
+	app-admin/gkrellm:2[X]
+	net-misc/curl"
+DEPEND="${RDEPEND}"
+
+# Will open gkrellm in X11 display
+RESTRICT="test"
+
+PATCHES=( "${FILESDIR}/${PF}"-make-test.patch )
+
+src_prepare() {
+	if use threads; then
+		PATCHES+=( "${FILESDIR}/${P}"-mt.patch )
+	fi
+	default
+}
+
+src_configure() {
+	tc-export CC
+}
+
+pkg_postinst() {
+	if use threads; then
+		elog "If you can't connect MPD, please unset USE threads."
+		elog "See https://bugs.gentoo.org/276970 for information."
+	fi
+}
--- a/media-plugins/gkrellmpc/files/gkrellmpc-0.1_beta10-r1-make-test.patch
+++ b/media-plugins/gkrellmpc/files/gkrellmpc-0.1_beta10-r1-make-test.patch
@@ -1,0 +1,10 @@
+Makefile: Run gkrellm, not gkrellm2 for test
+
+--- a/Makefile
++++ b/Makefile
+@@ -90,4 +90,4 @@ deinstall uninstall:
+ 	if [ -f ~/.gkrellm2/plugins/gkrellmpc.so ]; then rm -f ~/.gkrellm2/plugins/gkrellmpc.so; fi 
+ 
+ test: gkrellmpc.so
+-	gkrellm2 -p gkrellmpc.so
++	gkrellm -p gkrellmpc.so


media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
```diff
@@ -0,0 +1,43 @@
...
+
+RDEPEND="
+	app-admin/gkrellm:2[X]
+	net-misc/curl"
+DEPEND="${RDEPEND}"
+
+# Will open gkrellm in X11 display
+RESTRICT="test"
+
+PATCHES=( "${FILESDIR}/${PF}"-make-test.patch )
```
laumann, line 25: The patch only really has an effect for "make test", but then there's `RESTRICT="test"` right above. I converted the `sed` line from `-r0` to a patch here.

thesamesam: Can we unrestrict tests? How do they fail?

media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
```diff
@@ -0,0 +1,43 @@
...
+
+src_prepare() {
+	if use threads; then
+		PATCHES+=( "${FILESDIR}/${P}"-mt.patch )
```
thesamesam, line 29: It's not going to be a real issue for this, but conditional patching is to be avoided if we can help it because:
1. we don't notice if patches stop applying if not testing with that flag
2. it makes problems harder to debug (introduces more variance)

Patches should be written such that they can be applied unconditionally even if they only take effect conditionally based on build options.

laumann:
> +pkg_postinst() {
> +	if use threads; then
> +		elog "If you can't connect MPD, please unset USE threads."
> +		elog "See, https://bugs.gentoo.org/276970 for information."
> 
> Drop ,

Done.

media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
thesamesam: bonus: are these issues with conditional patching actually mentioned in devmanual? they should be. maybe you could add it ;)

laumann: Thanks for pointing this out. I don't think conditional patching is really mentioned in the devmanual - I'll take a look.

laumann: I found this in the devmanual: https://devmanual.gentoo.org/ebuild-writing/use-conditional-code/.

But I'm thinking [this page](https://devmanual.gentoo.org/ebuild-writing/misc-files/patches/index.html) could be expanded with a subsection called "conditional patching" that warns against the use of conditional patching for the reasons you gave.

media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
laumann: Just commenting out the RESTRICT line, produces:

```
$ FEATURES=test ebuild ./gkrellmpc-0.1_beta10-r1.ebuild clean install
[snip]
>>> Test phase: media-plugins/gkrellmpc-0.1_beta10-r1
make -j5   test
gkrellm -p gkrellmpc.so

(gkrellm:7901): Gtk-WARNING **: 09:18:39.522: cannot open display:
make: *** [Makefile:93: test] Error 1
 * ERROR: media-plugins/gkrellmpc-0.1_beta10-r1::gentoo failed (test phase):
 *   Make test failed. See above for details.
```

laumann: I suppose the ebuild process is not permitted access to interact with X11.

Running the test command manually reveals the following problem in the plugin Install Log:

```
*** Command line plugin:
./gkrellmpc.so
	Error: ./gkrellmpc.so: undefined symbol: curl_easy_getinfo
```

thesamesam: Can you try using the virtualx eclass?

laumann: Sure.

media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
thesamesam: How do you feel about drafting something up?

laumann: I'll do it!

media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
laumann: Sorry for being slow on this one. I changed the ebuild:

```diff
diff --git a/media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild b/media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
index 22adb571df5..37616f9222d 100644
--- a/media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
+++ b/media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=8
 
-inherit gkrellm-plugin toolchain-funcs
+inherit gkrellm-plugin toolchain-funcs virtualx
 
 DESCRIPTION="A gkrellm plugin to control the MPD (Music Player Daemon)"
 HOMEPAGE="http://mpd.wikia.com/wiki/Client:GKrellMPC"
@@ -20,7 +20,7 @@ RDEPEND="
 DEPEND="${RDEPEND}"
 
 # Will open gkrellm in X11 display
-RESTRICT="test"
+#RESTRICT="test"
 
 PATCHES=( "${FILESDIR}/${PF}"-make-test.patch )
 
@@ -35,6 +35,10 @@ src_configure() {
 	tc-export CC
 }
 
+src_test() {
+	virtx make test
+}
+
 pkg_postinst() {
 	if use threads; then
 		elog "If you can't connect MPD, please unset USE threads."

```
Result:
```
>>> Test phase: media-plugins/gkrellmpc-0.1_beta10-r1
 * ERROR: media-plugins/gkrellmpc-0.1_beta10-r1::gentoo failed (test phase):
 *   (no error message)
 *
 * Call stack:
 *                        ebuild.sh, line 127:  Called src_test
 *   gkrellmpc-0.1_beta10-r1.ebuild, line  39:  Called virtx 'make' 'test'
 *                      environment, line 1982:  Called die
 *       XVFB=$(type -p Xvfb) || die;
 *
```
(I guess I don't have USE=xvfb on x11-base/xorg-server)

media-plugins/gkrellmpc/gkrellmpc-0.1_beta10-r1.ebuild
laumann: https://github.com/gentoo/devmanual/pull/281

PR #24710, opened 2022-03-22
dev-libs/libpcre2: drop 'recursion-limit' USE flag

floppym:
Closes: https://bugs.gentoo.org/835796


floppym:
Pushed as 0f140c24292730f0fc225ff6683589fc84d63a45.

PR #24712, opened 2022-03-22
riscv: keyword x11-wm/fvwm and it's depends

liangyongxiang:


dlan17:
any reason why latest version (3.7.1) of x11-themes/fvwm-crystal is not keyworded?

PR #24713, opened 2022-03-22
app-admin/keepassxc: add missing dependencies

thesamesam:
- Botan exposes a subslot for its ABI so a subslot (:=) dep should be used
- Add missing argon2 dependency
- autotype: add missing libXi dependency

(Note that for autotype, it seems to always end up being built
(and finding dependencies?) even if the option is off?)


thesamesam:
@Polynomial-C

thesamesam:
USE=ccache should still go and := is still missing from Botan, so rebased.

PR #24714, opened 2022-03-22
games-engines/devilutionx: fix automagic ccache usage

thesamesam:
We let users enable this by themselves via e.g. FEATURES=ccache.

Closes: https://bugs.gentoo.org/813768


thesamesam:
@Polynomial-C

PR #24715, opened 2022-03-22
tpm2-pkcs11: Bump to 1.8.0 add new dependency

salahcoronya:
tpm2-pkcs11 bring a new dependency: dev-python/tpm2-pytss . So this PR includes the new package, as well. However, the tests all fail for dev-python/tpm2-pytss (ModuleNotFoundError: No module named 'tpm2_pytss._libtpm2_pytss', which is a CFFI module) and I can't figure out why. The tests work under pip. Either I'm missing a step. or this is a bug. (I've tried both PEP517 and traditional but both fail the tests in the same way).


salahcoronya:
OK, i finally got the test working Turned out 3 things here needed: To chdir out of ${S}. add ${S} and the virtualenv to PYTHONPATH, and change import mode of pytest. The test now run, They're may be a prettier way to to do, but it works. I'll leave ti draft for another day so. I'll let this stand for a while, then make the rest of my push request and mark it ready for pull.

salahcoronya:
Well, I had grander plans that included 2 more Python packages and more test coverage for tpm2-pkcs11, but there are failing tests on on of those 2, so I'm save that for another PR one day. Until then, will maintain the status quo (upstream did gain a new dependency, which is included in this PR)

salahcoronya:
dev-python/tpm2-pytss just pushed out a new version, so  I updated it. Unfortunately it also gained some new fragility:
```
Drop pkgconfig from runtime dependencies, thus no longer need dev packages of built bindings at runtime.
    NOTE: Version information is cached, a change in the TSS libraries requires a rebuild of the bindings.
```
I put a note in pkg_postinst about this fact. There's probably a better way though.

salahcoronya:
Well I ran a few "fragility tests" seeing what would happen if the version of tpm2_pytss didn't match tpm2-tss with 2 closely related versions (3.1.0 and 3.2.0, in both directions) (2.4->3.0 would trigger portage to rebuild the package anyway). Everything worked normally. So I removed the warning - portage should be able to check any breakage via the preserved-rebuild process normally, so no extra interaction should be required by the user other than the preserved-rebuild.

thesamesam:
> Well I ran a few "fragility tests" seeing what would happen if the version of tpm2_pytss didn't match tpm2-tss with 2 closely related versions (3.1.0 and 3.2.0, in both directions) (2.4->3.0 would trigger portage to rebuild the package anyway). Everything worked normally. So I removed the warning - portage should be able to check any breakage via the preserved-rebuild process normally, so no extra interaction should be required by the user other than the preserved-rebuild.

We can't rely on preserved-rebuild -- users aren't required to use it. If we need to rebuild based on diff. versions, we have to use subslots and subslot operators (:=).

juippis:
You do need to revbump them.

juippis (approved):
So can tpm2-tss-2.4.6 be removed? Let's monitor the future bumps closely if breaking ABI happens and add subslots then.

salahcoronya:
Yes, tpm2-tss-2.4.6 can be removed (I already had it removed in a different PR)

PR #24716, opened 2022-03-22
sys-process/cronie: fix cronie pidfile path

thesamesam:
cronie's build system is using runstatedir to determine
where to place the PID file. On one of my systems,
this path _happens_ to exist because of thermald,
but it won't usually exist.

We should set runstatedir to be "${EPREFIX}"/run
to store the PID file in the place the init script expects
(cronie hardcodes the value of runstatedir into the binary
to find where its pidfile is).

Closes: https://bugs.gentoo.org/835814
Bug: https://bugs.gentoo.org/685306

### DIFF
--- a/sys-process/cronie/cronie-1.6.0-r1.ebuild
+++ b/sys-process/cronie/cronie-1.6.0-r1.ebuild
@@ -42,6 +42,10 @@ src_prepare() {
 
 src_configure() {
 	local myeconfargs=(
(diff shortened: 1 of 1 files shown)


thesamesam:
@Polynomial-C

thesamesam:
Fixed in https://bugs.gentoo.org/685306#c1 and followup to add the ${EPREFIX}.

PR #24723, opened 2022-03-23
vala.eclass : fix EAPI 8 support (src_prepare to call vala_setup)

Miezhiko:
fixes building evolution-data-server-3.44.0

### DIFF
--- a/eclass/vala.eclass
+++ b/eclass/vala.eclass
@@ -180,7 +180,7 @@ vala_setup() {
 # @FUNCTION: vala_src_prepare
 # @DESCRIPTION:
 # For backwards compatibility in EAPIs 6 and 7.  Calls vala_setup.
-if [[ ${EAPI} == [67] ]]; then
+if [[ ${EAPI} == [678] ]]; then
 	vala_src_prepare() { vala_setup "$@"; }
 fi
 


Miezhiko:
@mattst88 @ulm

mattst88:
I'm unsure if we want to do this. I think it might be a good idea, since the reasoning behind renaming the function to "vala_setup" was to indicate that it could be called from `pkg_setup` -- but that was wrong.

I suspect it's probably just best to leave it as is.

FWIW, I fixed evolution-data-server.

eclass/vala.eclass
```diff
@@ -180,7 +180,7 @@ vala_setup() {
 # @FUNCTION: vala_src_prepare
 # @DESCRIPTION:
 # For backwards compatibility in EAPIs 6 and 7.  Calls vala_setup.
```
mattst88, line 182: *If* we want to do this, you should update the comment as well.

mattst88 (changes requested):

Miezhiko:
> 

thanks

ulm:
> I'm unsure if we want to do this. I think it might be a good idea, since the reasoning behind renaming the function to "vala_setup" was to indicate that it could be called from `pkg_setup` -- but that was wrong.
> 
> I suspect it's probably just best to leave it as is.

+1

If we reeable `src_prepare` then it should call `default` and the phase function should be exported. We discussed that before and decided against it.

About `vala_setup`, it's just a name. The commit message of commit bc13ab9233448a72eeb6551c948135a90fcf4502 introducing it says "can be called either from pkg_setup or from src_prepare" (and it _can_ be called from `pkg_setup` for source builds).

PR #24724, opened 2022-03-23
x11-plugins/gkwebmon: update EAPI 6 -> 8

laumann:
Changes:

 * { -> B}DEPEND="virtual/pkgconfig". The dependency is also provided by the eclass, but asking on #gentoo-dev it was recommended to keep it, because the ebuild explicitly does tc-export PKG_CONFIG.  

### DIFF
--- a/x11-plugins/gkwebmon/gkwebmon-0.2-r3.ebuild
+++ b/x11-plugins/gkwebmon/gkwebmon-0.2-r3.ebuild
@@ -1,0 +1,32 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit gkrellm-plugin toolchain-funcs
+
+DESCRIPTION="A web monitor plugin for GKrellM2"
+HOMEPAGE="http://gkwebmon.sourceforge.net/"
+SRC_URI="mirror://sourceforge/${PN}/${P}.tgz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~ppc ~sparc ~x86"
+
+# The Makefile links with -lssl.
+RDEPEND="
+	app-admin/gkrellm:2[X]
+	dev-libs/openssl"
+DEPEND="${RDEPEND}"
+BDEPEND="virtual/pkgconfig"
+
+PATCHES=(
+	"${FILESDIR}"/${P}-cc-cflags-ldflags.patch
+	"${FILESDIR}"/${P}-fno-common.patch
+)
+
+src_compile() {
+	tc-export PKG_CONFIG
+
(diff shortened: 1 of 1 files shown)


laumann:
@thesamesam Another one.

thesamesam:
Oh, while we're at it: you are welcome to (but it's not required) to do all of these in one PR to allow easier review and stuff.

x11-plugins/gkwebmon/gkwebmon-0.2-r3.ebuild
```diff
@@ -0,0 +1,32 @@
...
+SRC_URI="mirror://sourceforge/${PN}/${P}.tgz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~alpha ~amd64 ~ppc ~sparc ~x86"
+
+# The Makefile links with -lssl.
+RDEPEND="
+	app-admin/gkrellm:2[X]
+	dev-libs/openssl:0="
```
thesamesam, line 19: nit: just drop the 0, openssl hasn't been slotted for a long time (think there was a failed experiment for a bit)

laumann:
> Oh, while we're at it: you are welcome to (but it's not required) to do all of these in one PR to allow easier review and stuff.

I did consider it :-) I can one big PR with a commit per ebuild.

x11-plugins/gkwebmon/gkwebmon-0.2-r3.ebuild
laumann: Done.

juippis (approved):
Was this waiting for some eclass EAPI update? Seems to work fine at least already. 
```
 * VDB: detected possibly incorrect RDEPEND (x11-plugins/gkwebmon-0.2-r3)
 *                  > dev-libs/glib:2
 *                  > x11-libs/gtk+:2
 * dev-libs/openssl < 
```
Could you check these?

laumann:
> Was this waiting for some eclass EAPI update? 

No, the eclass was already updated in 5b1df6b067e0b1bfee6ba657c8df4d24a5b08d35

> Seems to work fine at least already.
> 
> ```
>  * VDB: detected possibly incorrect RDEPEND (x11-plugins/gkwebmon-0.2-r3)
>  *                  > dev-libs/glib:2
>  *                  > x11-libs/gtk+:2
>  * dev-libs/openssl < 
> ```
> 
> Could you check these?

Happily, but I'm not sure what I need to check here.

And now you made me curious :-) How did you get this report?

juippis:
I ended up merging this PR and doing a followup commit, c877f78c1fd4571694c1e80ff7b261d45755f952. It's from iwdevtools, https://packages.gentoo.org/packages/app-portage/iwdevtools

We don't depend on indirect deps. In theory, if gkgrellm dropped glib in a future release, it'd break gkwebmon. So always list exactly the needed deps.

PR #24729, opened 2022-03-23
x11-plugins/gkrellm-leds: update EAPI 6 -> 8

laumann:
Besides the EAPI bump, there's a bug for the referenced bug. The issue
seems to be simply a typo in src/Makefile.am resulting in a variable
@GK_LDFLAGS@ not being replaced appropriately.

Closes: https://bugs.gentoo.org/799176

---
TODO
 - [x] `ebuild install` -r1 and -r2 and compare the image folders

### DIFF
--- a/x11-plugins/gkrellm-leds/files/gkrellm-leds-0.8.2-r2-ldflags-typo.patch
+++ b/x11-plugins/gkrellm-leds/files/gkrellm-leds-0.8.2-r2-ldflags-typo.patch
@@ -1,0 +1,12 @@
+src/Makefile.am: fix type GK_FLAGS -> GKM_FLAGS
+--- a/src/Makefile.am
++++ b/src/Makefile.am
+@@ -10,7 +10,7 @@ gkleds_la_CFLAGS = -Wall -O2 -I../pixmaps @X_CFLAGS@ @GTK_CFLAGS@ @GKM_CFLAGS@
+ gkleds_la_LIBADD = @X_PRE_LIBS@ @X_LIBS@ @X_EXTRA_LIBS@ @GTK_LIBS@ @GKM_EXTRA_SRCS@
+ 
+ ## linker flags, we are building a plugin so skip lib prefix and lib versions
+-gkleds_la_LDFLAGS = -module -avoid-version @GK_LDFLAGS@
++gkleds_la_LDFLAGS = -module -avoid-version @GKM_LDFLAGS@
+ 
+ ## gkleds.la sources
+ gkleds_la_SOURCES = gkleds.c gkleds.h
--- a/x11-plugins/gkrellm-leds/gkrellm-leds-0.8.2-r2.ebuild
+++ b/x11-plugins/gkrellm-leds/gkrellm-leds-0.8.2-r2.ebuild
@@ -1,0 +1,36 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit gkrellm-plugin autotools
+
+MY_P="${P/rellm-/}"
+
+DESCRIPTION="GKrellM2 plugin for monitoring keyboard LEDs"
+HOMEPAGE="http://heim.ifi.uio.no/~oyvinha/gkleds/"
+SRC_URI="http://heim.ifi.uio.no/~oyvinha/e107_files/downloads/${MY_P}.tar.gz"
+S="${WORKDIR}/${MY_P}"
+
+LICENSE="GPL-2"
+SLOT="2"
+KEYWORDS="~alpha ~amd64 ~ppc ~sparc ~x86"
+
+RDEPEND="
+	app-admin/gkrellm:2[X]
+	x11-libs/libXtst"
+DEPEND="${RDEPEND}
+	x11-base/xorg-proto"
+BDEPEND="virtual/pkgconfig"
+
+PATCHES=( "${FILESDIR}/${P}-r2-ldflags-typo.patch" )
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	PLUGIN_SO=( src/.libs/gkleds$(get_modname) )
+	default
+}


x11-plugins/gkrellm-leds/gkrellm-leds-0.8.2-r2.ebuild
```diff
@@ -0,0 +1,37 @@
...
+
+RDEPEND="
+	app-admin/gkrellm:2[X]
+	x11-libs/libXtst"
+DEPEND="${RDEPEND}
+	x11-base/xorg-proto"
+
+S="${WORKDIR}/${MY_P}"
+
+PATCHES=( "${FILESDIR}/${P}-r2-ldflags-typo.patch" )
```
laumann, line 27: I'm not entirely sure which variables I should use in reference to the patch. My understanding so far is that one should refrain from `${PF}`, but the PMS lists several others (including `${P}` here) that aren't "consistent".

thesamesam: I _think_ it's "not consistent" in the sense that "it might be different in e.g. pkg_postrm". It's safe to use it in PATCHES (but don't use ${PF} for the reasons @ionenwks mentioned in #gentoo-dev earlier; we generally anticipate we can revbump packages en-masse without affecting their installability. Using ${PF} harms that).

ionenwks: Typically:
- Patch will likely keep across versions: ${PN}-0.0.0-name.patch (i.e. long standing issues, or Gentoo specific patches)
- Patch that should be either reviewed or dropped next version: ${P}-name.patch (i.e may be fixed / upstreamed, if not, can change to `${PN}` on bump)

And yeah, `${PF}` is just different given it includes revision making it too volatile over nothing. It shouldn't be thrown around casually and revbumps with no changes shouldn't be able to break things, *normally* only used for docs dir.

laumann: OK, seems like I was right to change `${PF}` to `${P}` here. I included `-r2` in the patch name, because I wanted to reflect that it's been created for that revision. But it should probably also apply to any potential future revisions of this version.

Thanks for the input!

laumann:
@thesamesam ping on this one. I pushed some minor edits to the ebuild.

x11-plugins/gkrellm-leds/files/gkrellm-leds-0.8.2-r2-ldflags-typo.patch
```diff
@@ -0,0 +1,12 @@
+src/Makefile.am: fix type GK_FLAGS -> GKM_FLAGS
+--- a/src/Makefile.am
++++ b/src/Makefile.am
+@@ -10,7 +10,7 @@ gkleds_la_CFLAGS = -Wall -O2 -I../pixmaps @X_CFLAGS@ @GTK_CFLAGS@ @GKM_CFLAGS@
+ gkleds_la_LIBADD = @X_PRE_LIBS@ @X_LIBS@ @X_EXTRA_LIBS@ @GTK_LIBS@ @GKM_EXTRA_SRCS@
+ 
+ ## linker flags, we are building a plugin so skip lib prefix and lib versions
+-gkleds_la_LDFLAGS = -module -avoid-version @GK_LDFLAGS@
++gkleds_la_LDFLAGS = -module -avoid-version @GKM_LDFLAGS@
```
SoapGentoo, line 9: in general, in variables you shouldn't reference `@VAR@` but `$(VAR)` because `AC_SUBST` already creates a separate `VAR = @VAR@` somewhere in `Makefile.in`.

x11-plugins/gkrellm-leds/gkrellm-leds-0.8.2-r2.ebuild
```diff
@@ -0,0 +1,35 @@
...
+
+src_prepare() {
+	default
+	eautoreconf
+}
+
+src_configure() {
+	PLUGIN_SO=( src/.libs/gkleds$(get_modname) )
```
SoapGentoo, line 33: is this really needed?

x11-plugins/gkrellm-leds/files/gkrellm-leds-0.8.2-r2-ldflags-typo.patch
laumann: OK, I wanted to go for the smallest change that would fix the problem. Do you want me make this change?

x11-plugins/gkrellm-leds/gkrellm-leds-0.8.2-r2.ebuild
laumann: This should be the [name of the plugin's .so file](https://devmanual.gentoo.org/eclass-reference/gkrellm-plugin.eclass/index.html#lbAG). The default is `${PN}$(get_modname)`, so I think it's needed, yes.

x11-plugins/gkrellm-leds/files/gkrellm-leds-0.8.2-r2-ldflags-typo.patch
SoapGentoo: no, it's just something to be aware of if you're fixing build systems

laumann: Alright, thanks! :smile: I appreciate it.

x11-plugins/gkrellm-leds/gkrellm-leds-0.8.2-r2.ebuild
thesamesam: FWIW, referencing .libs/ tends to break with slibtool

laumann: What's the solution to that? Move the file after it's been built?

thesamesam: Do we actually need to set it at all in configure? Can't we just set it to the right name on install!

PR #24734, opened 2022-03-24
dev-lang/zig: add memory requirement

BratishkaErik:
See https://github.com/ziglang/zig/issues/6485

Signed-off-by: BratishkaErik <bratishkaerik@getgoogleoff.me>


dev-lang/zig/zig-0.9.1.ebuild
```diff
@@ -13,7 +13,7 @@ if [[ ${PV} == 9999 ]]; then
 	inherit git-r3
 else
 	SRC_URI="https://github.com/ziglang/zig/archive/${PV}.tar.gz -> ${P}.tar.gz"
-	KEYWORDS="~amd64 ~arm ~arm64"
+	KEYWORDS="~amd64 ~arm64"
```
Flowdalic, line 16: Why drop `~arm`?

Flowdalic:
Could you split the commit into two commits: one "drop old" and one "add memory requirement via check-regs"?

dev-lang/zig/zig-0.9.1.ebuild
BratishkaErik: Oops, sorry...

BratishkaErik:
> Could you split the commit into two commits: one "drop old" and one "add memory requirement via check-reqs"?

Done.

PR #24735, opened 2022-03-24
app-arch/libarchive: Added verify-sig integration

jpds:


sec-keys/openpgp-keys-libarchive/openpgp-keys-libarchive-20220324.ebuild
```diff
@@ -0,0 +1,22 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+DESCRIPTION="OpenPGP keys used by the libarchive project"
+HOMEPAGE="https://www.libarchive.org/"
```
mgorny, line 7: Does the homepage list the key used anywhere?

jpds: Unfortunately not, I pulled it out of the `.asc`, and the release commits have the same key ID.

mgorny: Guessed so.

mgorny (approved):
Thanks. I'll do some reformatting and merge.

PR #24737, opened 2022-03-24
gkrellm-plugin.eclass users: Bump to EAPI 8

laumann:
@thesamesam Here's a big PR to track all the users of gkrellm-plugin.eclass, and bumping them to EAPI=8. I'll tick them off as I go, and maybe add comments as well.

 - [x] x11-plugins/gkrellm-cpupower-0.2-r1 - Add patch for Makefile, heavily inspired by the patch file for [x11-plugins/gkrellshoot](https://github.com/gentoo/gentoo/blob/master/x11-plugins/gkrellshoot/files/makefile-respect-flags.patch) by @Flowdalic 
 - [x] x11-plugins/gkrellshoot-0.4.4-r3 - plus PKG_CONFIG
 - [x] x11-plugins/gkrellm-countdown-0.1.2-r1 - plus PKG_CONFIG
 - [x] x11-plugins/gkrellmlaunch-0.5 - plus PKG_CONFIG
 - [x] x11-plugins/gkrellm-mailwatch-2.4.3-r2 -  plus PKG_CONFIG
 - [x] x11-plugins/gkrellm-trayicons-1.03-r1 - plus PKG_CONFIG
 - [x] x11-plugins/gkrellm-vaiobright-2.5-r2
 - [x] x11-plugins/gkrellkam-2.0.0 - plus PKG_CONFIG
 - [x] x11-plugins/gkrellm-xkb-1.05-r1 - plus PKG_CONFIG
 - [x] x11-plugins/gkrellmoon-0.6-r2 - converted `sed` line to patch, plus PKG_CONFIG
 - [ ] x11-plugins/gkrellfire-0.4.2-r1
 - [x] x11-plugins/gkrellsun-1.0.0-r4 - fixes: Don't call `gcc` directly, respect pkgconfig, ensure $(CC) and $(CFLAGS) are used
 - [x] x11-plugins/gkrellweather-2.0.8-r1 - fixes: Don't call `gcc` directly, respect pkgconfig. Suspect that flag-o-matic is not needed
 - [x] x11-plugins/gkrellm-bgchanger-0.1.11-r2
 - [x] x11-plugins/gkrellm-radio-2.0.4 - plus PKG_CONFIG
 - [x] x11-plugins/gkrelltop-2.2.13-r2 - plus PKG_CONFIG
 - [x] x11-plugins/gkrellm-imonc-0.2-r1 - plus PKG_CONFIG
 - [x] x11-plugins/gkrellstock-0.5.1-r1 - plus PKG_CONFIG, rm "-O2 -Wall" flags, don't call gcc directly
 - [x] x11-plugins/gkrellm-volume-2.1.13-r2 - plus PKG_CONFIG
 - [x] x11-plugins/gkrellmss-2.6-r4 - plus PKG_CONFIG, don't call cc directly, respect CFLAGS
 - [x] x11-plugins/i8krellm-2.5-r2 - plus PKG_CONFIG, don't call gcc directly, respect CFLAGS, remove makefile support for gkrellm1
 - [x] x11-plugins/gkrellmwireless-2.0.3-r2 - plus PKG_CONFIG
 - [ ] x11-plugins/gkrellm-bluez-0.2-r2
 - [ ] x11-plugins/gkrellaclock-0.3.4-r1 - bug [742836 (does not respect CFLAGS)](https://bugs.gentoo.org/742836). The compiled .so file fails to load with "gkrellm -p" - it expects a function: `Error: /var/tmp/portage/x11-plugins/gkrellaclock-0.3.4-r2/image/usr/lib64/gkrellm2/plugins/gkrellaclock.so: cannot open shared object file: No such file or directory`

--- 
Overall changes
 - Adjusting Makefiles where appropriate to use user-defined `PKG_CONFIG`

### DIFF
--- a/x11-plugins/gkrellm-cpupower/files/gkrellm-cpupower-0.2-makefile.patch
+++ b/x11-plugins/gkrellm-cpupower/files/gkrellm-cpupower-0.2-makefile.patch
@@ -1,0 +1,45 @@
+--- a/Makefile
++++ b/Makefile
+@@ -1,22 +1,29 @@
+ # Makefile for gkrellm cpupower plugin
+ 
+-GTK_INCLUDE = `pkg-config gtk+-2.0 --cflags`
+-GTK_LIB = `pkg-config gtk+-2.0 --libs`
++PKG_CONFIG ?= pkg-config
+ 
+-FLAGS = -O2 -Wall -fPIC $(GTK_INCLUDE)
++GTK_INCLUDE = $(shell ${PKG_CONFIG} gtk+-2.0 --cflags)
++GTK_LIB = $(shell ${PKG_CONFIG} gtk+-2.0 --libs)
++
++FLAGS = -fPIC $(GTK_INCLUDE)
+ LIBS = $(GTK_LIB)
+ 
+ LFLAGS = -shared -lcpupower
+ 
+-CC = gcc $(CFLAGS) $(FLAGS)
++CC = $(CC)
+ 
+ OBJS = cpupower.o
+ 
++all: cpupower.so
++
++%.o: %.c
++	$(CC) $(CFLAGS) $(FLAGS) -c -o $@ $<
++
+ cpupower.so: $(OBJS)
+-	$(CC) $(OBJS) -o cpupower.so $(LFLAGS) $(LIBS)
++	$(CC) $(CFLAGS) $(LDFLAGS) $(FLAGS) $(OBJS) -o cpupower.so $(LFLAGS) $(LIBS)
+ 
+ install: cpupower.so
+-	install -D -m 755 -s cpupower.so $(DESTDIR)/usr/lib/gkrellm2/plugins/cpupower.so
++	install -D -m 755 cpupower.so $(DESTDIR)/usr/lib/gkrellm2/plugins/cpupower.so
+ 
+ install-sudo:
+ 	mkdir -p $(DESTDIR)/etc/sudoers.d
+@@ -25,6 +32,3 @@ install-sudo:
+ 
+ clean:
+ 	rm -f *.o core *.so* *.bak *~
+-
+-cpupower.o: cpupower.c
+-
--- a/x11-plugins/gkrellm-cpupower/gkrellm-cpupower-0.2-r2.ebuild
+++ b/x11-plugins/gkrellm-cpupower/gkrellm-cpupower-0.2-r2.ebuild
@@ -1,0 +1,47 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit gkrellm-plugin toolchain-funcs
+
+MY_P="${P/gkrellm/gkrellm2}"
+
+DESCRIPTION="A Gkrellm2 plugin for displaying and manipulating CPU frequency"
+HOMEPAGE="https://github.com/sainsaar/gkrellm2-cpupower/"
+SRC_URI="https://github.com/sainsaar/gkrellm2-cpupower/archive/${PV}.tar.gz -> ${MY_P}.tar.gz"
+S="${WORKDIR}/${MY_P}"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+DEPEND="app-admin/gkrellm:2[X]"
+RDEPEND="
+	${DEPEND}
+	app-admin/sudo
+	sys-power/cpupower"
+BDEPEND="virtual/pkgconfig"
+
+PATCHES=(
+	"${FILESDIR}"/${P}-makefile.patch
+)
+
+src_compile() {
+	tc-export PKG_CONFIG
+	emake CC="$(tc-getCC)" LDFLAGS="${LDFLAGS}"
+}
+
+src_install() {
+	local PLUGIN_SO=( cpupower$(get_modname) )
+	gkrellm-plugin_src_install
+	emake DESTDIR="${D}" install-sudo
+}
+
+pkg_postinst() {
+	einfo
+	einfo "For changing the governor and CPU frequencies as a user, create the \"trusted\""
+	einfo "group, and add those users to that group who should be allowed to perform"
+	einfo "these changes."
+	einfo
+}
--- a/x11-plugins/gkrellshoot/gkrellshoot-0.4.4-r4.ebuild
+++ b/x11-plugins/gkrellshoot/gkrellshoot-0.4.4-r4.ebuild
@@ -1,0 +1,31 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit gkrellm-plugin toolchain-funcs
+
+DESCRIPTION="GKrellm2 plugin to take screen shots and lock screen"
+HOMEPAGE="http://gkrellshoot.sourceforge.net/"
+SRC_URI="mirror://sourceforge/gkrellshoot/${P}.tar.gz"
+S="${WORKDIR}/${P/s/S}"
+
+LICENSE="GPL-2"
+SLOT="2"
+KEYWORDS="~alpha ~amd64 ~ppc ~sparc ~x86"
+
+DEPEND="app-admin/gkrellm:2[X]"
+RDEPEND="
+	${DEPEND}
+	virtual/imagemagick-tools"
+BDEPEND="virtual/pkgconfig"
+
+PATCHES=(
+	"${FILESDIR}"/makefile-respect-flags.patch
+	"${FILESDIR}/${P}"-r4-pkgconfig.patch
+)
+
+src_compile() {
+	tc-export PKG_CONFIG
+	default
+}
--- a/x11-plugins/gkrellkam/files/gkrellkam-2.0.0-r1-pkgconfig.patch
+++ b/x11-plugins/gkrellkam/files/gkrellkam-2.0.0-r1-pkgconfig.patch
@@ -1,0 +1,13 @@
+--- a/Makefile
++++ b/Makefile
+@@ -12,8 +12,9 @@ DESTDIR =
+ # This should point to the GKrellM headers
+ GKRELLM_HDRS = /usr/include
+ 
++PKG_CONFIG ?= pkg-config
+ CC = $(CC)
+-GTKFLAGS := $(shell pkg-config gtk+-2.0 --cflags)
++GTKFLAGS := $(shell ${PKG_CONFIG} gtk+-2.0 --cflags)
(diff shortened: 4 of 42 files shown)


x11-plugins/gkrellm-cpupower/files/gkrellm-cpupower-0.2-makefile.patch
```diff
@@ -0,0 +1,35 @@
+--- a/Makefile
++++ b/Makefile
+@@ -3,17 +3,22 @@
+ GTK_INCLUDE = `pkg-config gtk+-2.0 --cflags`
```
thesamesam, line 4: While at it: replace `pkg-config` with `$(PKG_CONFIG)` or similar (defaulting to `pkg-config`) to respect user's choice of pkg-config.

While `pkg-config` should respect appropriate environment variables such that it's not needed to override it really, that doesn't mean much.

x11-plugins/gkrellm-cpupower/files/gkrellm-cpupower-0.2-makefile.patch
```diff
@@ -0,0 +1,35 @@
...
++
++%.o: %.c
++	$(CC) $(CFLAGS) $(FLAGS) -c -o $@ $<
++
+ cpupower.so: $(OBJS)
+-	$(CC) $(OBJS) -o cpupower.so $(LFLAGS) $(LIBS)
++	$(CC) $(CFLAGS) $(LDFLAGS) $(FLAGS) $(OBJS) -o cpupower.so $(LFLAGS) $(LIBS)
+ 
+ install: cpupower.so
+ 	install -D -m 755 -s cpupower.so $(DESTDIR)/usr/lib/gkrellm2/plugins/cpupower.so
```
thesamesam, line 28: It seems odd to me that it's always `lib` but I might be missing something. Also, doubt `-s` should be here (strip).

x11-plugins/gkrellm-cpupower/gkrellm-cpupower-0.2-r2.ebuild
```diff
@@ -0,0 +1,47 @@
...
+	app-admin/sudo
+	sys-power/cpupower"
+
+S="${WORKDIR}/${MY_P}"
+
+PATCHES=(
+	"${FILESDIR}"/${P}-makefile.patch
+)
+
+src_install() {
```
thesamesam, line 31: Change to match [phase execution order](https://devmanual.gentoo.org/ebuild-writing/functions/index.html)

x11-plugins/gkrellm-cpupower/gkrellm-cpupower-0.2-r2.ebuild
```diff
@@ -0,0 +1,47 @@
...
+MY_P="${P/gkrellm/gkrellm2}"
+
+DESCRIPTION="A Gkrellm2 plugin for displaying and manipulating CPU frequency"
+HOMEPAGE="https://github.com/sainsaar/gkrellm2-cpupower/"
+SRC_URI="https://github.com/sainsaar/gkrellm2-cpupower/archive/${PV}.tar.gz -> ${MY_P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
```
thesamesam, line 17: [Drop, as it's unnecessary?](Drop, as it's unnecessary?)

x11-plugins/gkrellshoot/gkrellshoot-0.4.4-r4.ebuild
```diff
@@ -0,0 +1,26 @@
...
+SLOT="2"
+KEYWORDS="~alpha ~amd64 ~ppc ~sparc ~x86"
+IUSE=""
+
+DEPEND="app-admin/gkrellm:2[X]"
+RDEPEND="
+	${DEPEND}
+	virtual/imagemagick-tools"
+
+S="${WORKDIR}/${P/s/S}"
```
thesamesam, line 22: I prefer S directly below SRC_URI b/c directly related

x11-plugins/gkrellshoot/gkrellshoot-0.4.4-r4.ebuild
```diff
@@ -0,0 +1,26 @@
...
+inherit gkrellm-plugin
+
+DESCRIPTION="GKrellm2 plugin to take screen shots and lock screen"
+HOMEPAGE="http://gkrellshoot.sourceforge.net/"
+SRC_URI="mirror://sourceforge/gkrellshoot/${P}.tar.gz"
+
+LICENSE="GPL-2"
+SLOT="2"
+KEYWORDS="~alpha ~amd64 ~ppc ~sparc ~x86"
+IUSE=""
```
thesamesam, line 15: Drop

x11-plugins/gkrellm-cpupower/files/gkrellm-cpupower-0.2-makefile.patch
```diff
@@ -0,0 +1,45 @@
...
+@@ -1,22 +1,29 @@
+ # Makefile for gkrellm cpupower plugin
+ 
+-GTK_INCLUDE = `pkg-config gtk+-2.0 --cflags`
+-GTK_LIB = `pkg-config gtk+-2.0 --libs`
++PKG_CONFIG = pkg-config
+ 
+-FLAGS = -O2 -Wall -fPIC $(GTK_INCLUDE)
++GTK_INCLUDE = $(shell ${PKG_CONFIG} gtk+-2.0 --cflags)
++GTK_LIB = $(shell ${PKG_CONFIG} gtk+-2.0 --libs)
```
laumann, line 12: @thesamesam This is what I figured was the "correct" way.

thesamesam: PKG_CONFIG ?= ..., no?

x11-plugins/gkrellm-cpupower/files/gkrellm-cpupower-0.2-makefile.patch
laumann: I replaced it with:

```makefile
PKG_CONFIG = pkg-config # is this fine for the default?

GTK_INCLUDE = $(shell ${PKG_CONFIG} gtk+-2.0 --cflags)
GTK_LIB = $(shell ${PKG_CONFIG} gtk+-2.0 --libs)
```

x11-plugins/gkrellm-cpupower/files/gkrellm-cpupower-0.2-makefile.patch
laumann: I'm not sure about the destination here, but I've gone ahead and removed `-s`.

laumann: I'll remember to check that for the other Makefiles.

x11-plugins/gkrellm-cpupower/gkrellm-cpupower-0.2-r2.ebuild
laumann: Done. Silly me.

x11-plugins/gkrellm-cpupower/gkrellm-cpupower-0.2-r2.ebuild
laumann: Done.

x11-plugins/gkrellshoot/gkrellshoot-0.4.4-r4.ebuild
laumann: OK, I was wondering where best to put - I had gathered between the *DEPEND variables and phase functions :-)

x11-plugins/gkrellshoot/gkrellshoot-0.4.4-r4.ebuild
laumann: Done.

x11-plugins/gkrellm-cpupower/files/gkrellm-cpupower-0.2-makefile.patch
laumann: Yes :-) Changed.

x11-plugins/gkrellm-cpupower/files/gkrellm-cpupower-0.2-makefile.patch
laumann: @thesamesam Should it be doing `tc-export PKG_CONFIG` as in the ebuild as well?

thesamesam: yes!

laumann: OK, then I will probably do:

```
BDEPEND="virtual/pkgconfig"
```
coupled with

```
tc-export PKG_CONFIG
```

in the ebuilds where this patch is added.

laumann: I've gone through the already bumped ebuilds and added patches + BDEPEND for pkg-config.

laumann:
@juippis made be aware of the fact that ebuilds shouldn't depend on transitive dependencies, and noticed in #24724 that two dependencies were missing from RDEPEND. I think this applies across most (if not all) of these ebuilds. I'll be sure to circle back and add these in where appropriate.

laumann:
> 1) I added gkrellfire.o to the end of line 50 of the fixed-fix-CC-CFLAGS-LDFLAGS-handling.patch. 2) I increased CHART_MAX_WIDTH from 120 to 150 in gkrellfire.c. 3) I changed the ebuild SRC_URI to "[http://distcache.FreeBSD.org/ports-distfiles/${P}.tgz](http://distcache.FreeBSD.org/ports-distfiles/$%7BP%7D.tgz)". The old upstream seems dead (as of 0.3.3?,) but I can't figure out where freebsd is getting their source.

1. This is a fix for bug [742839](https://bugs.gentoo.org/742839)? You can just open a separate PR for gkrellfire, then I'll either omit it from this one, or rebase to include your patch
2. The patch for CHART_MAX_WIDTH sounds like a [user patch](https://wiki.gentoo.org/wiki//etc/portage/patches)
3. Is it the same code as from the original SRC_URI?

genterminl:
1) not directly for bug 742839.  Not sure why/when I did this, but probably a local build failure.  I'll file separately if it's still needed after your changes.
2) agreed as user patch, but might be useful to avoid surprise if user widens the gkrellm display and grkrellfire doesn't expand to the full width.  Certainly not critical.
3) the updated SRC_URI is because freebsd moved the download source.  The only file at the old location is a README telling of the new location.

PR #24743, opened 2022-03-25
net-misc/tigervnc: check and clean up dependencies

ceamac:
I rearranged the dependencies.  I tested most USE flag combinations I could think of.  `qa-vdb` is happy.  I also added a few less obvious packages to RDEPEND that are needed if installing from binary package on a mostly bare system.

When I added [pkg_postinst()](https://github.com/gentoo/gentoo/pull/24252) I broke `xdg.eclass`, so now I call `xdg_pkg_postinst` to fix it.

I also added myself as a maintainer and I'll take care of the other bugs too at some point.


juippis (approved):
Seems to work, thanks!

PR #24744, opened 2022-03-25
www-client/seamonkey: Bump to version 2.53.11-r1

BioMike:
Requested in bug https://bugs.gentoo.org/835828#c6 : support for media-libs/dav1d-1.0.0, while retaining support for current stable.

No other changes were made to the ebuild, compared to 2.53.11.


BioMike:
Seems like that got released at about the same moment I did the PR. Checking the code, I see that this patch hasn't been included in that.

I'll bump the version and include a few other things in there that I was planning to put into the next version, if you don't mind.

thesamesam:
Sure!

juippis (approved):

PR #24751, opened 2022-03-25
x11-misc/xpad: add xdg_icon_cache_update

rndxelement:
The ebuild for xpad-5.7.0 was missing calls of xdg_icon_cache_update.

Closes: https://bugs.gentoo.org/835924

### DIFF
--- a/x11-misc/xpad/xpad-5.7.0.ebuild
+++ b/x11-misc/xpad/xpad-5.7.0.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=8
 
-inherit autotools
+inherit autotools xdg-utils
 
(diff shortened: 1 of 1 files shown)


x11-misc/xpad/xpad-5.7.0.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=8
 
-inherit autotools
+inherit autotools xdg-utils
```
juippis, line 6: You could've just inherited `xdg` here, then you wouldn't manually have to call the functions in pkg_post* phases. But this works too, thanks!

juippis (approved):

PR #24752, opened 2022-03-25
 app-misc/screen: Enable nls use flag for build dependency.

GrataHonoria:
Set the nls use flag for the  dependency sys-apps/texinfo.


swegener:
Could you provide some info about the issue this fixes? It builds fine with texinfo[-nls] for me.

PR #24755, opened 2022-03-26
dev-util/hip: add missing BDEPEND providing GL/glx.h

littlewu2508:
Previously this is the (build-time) dependency of rocclr, which has
DEPEND="virtual/opengl" to make sure GL/glx.h is present. Now hip-5
bundles rocclr and this BDEPEND shuold be added. Turned out headers
are provided by media-libs/libglvnd

Closes: https://bugs.gentoo.org/836136

### DIFF
--- a/dev-util/hip/hip-5.0.2.ebuild
+++ b/dev-util/hip/hip-5.0.2.ebuild
@@ -28,6 +28,7 @@ DEPEND="
 RDEPEND="${DEPEND}
 	dev-perl/URI-Encode
 	dev-libs/roct-thunk-interface:${SLOT}"
+BDEPEND="media-libs/libglvnd"
 
 PATCHES=(
 	"${FILESDIR}/${PN}-5.0.1-DisableTest.patch"


dev-util/hip/hip-5.0.2.ebuild
```diff
@@ -28,6 +28,7 @@ DEPEND="
 RDEPEND="${DEPEND}
 	dev-perl/URI-Encode
 	dev-libs/roct-thunk-interface:${SLOT}"
+BDEPEND="media-libs/libglvnd"
```
heroxbd, line 31: use `virtual/opengl` to align with rocclr.

heroxbd (changes requested):

heroxbd:
> ... this BDEPEND shuold be added. 

typo

thesamesam:
Headers should be in DEPEND, not BDEPEND.

heroxbd:
> 

Good point.  Will do a quick fix.

thesamesam:
Thank you! (For both that & all the merges of PRs)

littlewu2508:
> Headers should be in DEPEND, not BDEPEND, because they must be available:
> 
>     1. in the right location for cross (in (E)SYSROOT)
> 
>     2. may even vary per-architecture

Got it. Thanks for pointing out!

PR #24757, opened 2022-03-26
eclass: Use UID 0 instead of root and other fixes

ulm:


mgorny (approved):
LGTM. I find the prior mixing of `root` user with `0` gid somewhat curious (I guess one of the supported Gentoo variants used a different root group?).

thesamesam (approved):

PR #24760, opened 2022-03-26
www-client/seamonkey: bump version to 2.53.11.1

BioMike:
Additionally:

* Remove check for big endian and conditional --disable-skia configuration option.
* Include an upstream patch that fixes the skia build on big endian systems.
Bug: https://bugs.gentoo.org/723372

* Remove a sed that suppresses format errors on gcc-9.

Signed-off-by: Myckel Habets <gentoo-bugs@habets-dobben.nl>


thesamesam:
Excellent work. Thank you!

PR #24761, opened 2022-03-26
dev-libs/icinga-php-library: bump 0.8.0

antonfischl1980:
simple version bump

I also set EAPI=8 while at it

added upstream infos to metadata.xml


antonfischl1980:
no longer needed

PR #24765, opened 2022-03-26
media-gfx/gimp: 2.99.10 -> 2.99.10-r1, add USE="jpegxl"

band-a-prend:
Add new "jpegxl" USE-flag that previously was added to live-ebuild.


juippis (approved):
Thanks!

PR #24768, opened 2022-03-27
bump urbanterror

holgersson32644:
- games-fps/urbanterror: Bump to 4.3.4_p20211208
- games-fps/urbanterror: drop 4.3.4_p20210304


games-fps/urbanterror/urbanterror-4.3.4_p20211208.ebuild
```diff
@@ -1,34 +1,37 @@
...
 fi
 
+KEYWORDS="~amd64 ~x86"
 LICENSE="GPL-2"
 SLOT="0"
 IUSE="+altgamma +client +curl debug mumble openal +opus server +skeetshootmod voip vorbis"
-REQUIRED_USE="|| ( client server )
-		voip? ( opus )"
+REQUIRED_USE="
+	|| ( client server )
```
Flowdalic, line 27: Even though that is a non-semantic change, I like this style better. So that change is fine by me.

games-fps/urbanterror/urbanterror-4.3.4_p20211208.ebuild
```diff
@@ -1,34 +1,37 @@
...
+	voip? ( opus )
+"
 
 DOCS=( ChangeLog README.md README.ioq3.md md4-readme.txt )
 
 PATCHES=(
-	"${FILESDIR}"/${PN}-4.3.4_p20180708-fix-build_system.patch
+	"${FILESDIR}/${PN}-4.3.4_p20180708-fix-build_system.patch"
```
Flowdalic, line 34: You don't have to change the quote style here, as `PN` is guranteed to contain no space character. Hence, in the spirit of keeping the diff minimal, I suggest to revert that.

Flowdalic (changes requested):

games-fps/urbanterror/urbanterror-4.3.4_p20211208.ebuild
```diff
@@ -57,12 +60,16 @@ pkg_pretend() {
...
-
 	tc-export CC
+
+	default
 }
 
 src_compile() {
+	# -fstrict-aliasing breaks network connections to any server.
+	# Upstream bug is https://github.com/mickael9/ioq3/issues/40.
+	append-flags "-fno-strict-aliasing"
```
Flowdalic, line 71: It appears that upstream is already disabling strict aliasing: https://github.com/mickael9/ioq3/blob/c988f7669bd70b59beb384563d4c8304062ae011/Makefile#L321

Could it be that you ran into this on a system which had explicitly enabled this (which nobody should do)?

games-fps/urbanterror/urbanterror-4.3.4_p20211208.ebuild
```diff
@@ -1,34 +1,37 @@
...
 	EGIT_REPO_URI="https://github.com/mickael9/ioq3.git"
 	EGIT_BRANCH="urt"
 else
-	COMMIT_ID="0429c03056720523d27ca71d5a4aa3e8d00709e7"
 	SRC_URI="https://github.com/mickael9/ioq3/archive/${COMMIT_ID}.tar.gz -> ${P}.tar.gz"
 	S="${WORKDIR}/ioq3-${COMMIT_ID}"
-	KEYWORDS="~amd64 ~x86"
 fi
 
+KEYWORDS="~amd64 ~x86"
```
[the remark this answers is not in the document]
holgersson32644: I've seen at least one ebuild in the past with the "if PV == 9999, then" logic where someone added KEYWORDS and we ended up with at least two occurrences of them.
Therefor I'd like to align the ebuilds I maintain into a single KEYWORDS line outside of the conditional structure.

The COMMIT_ID needs to be edited manually each time, so a position as early in the file as reasonable possible looks beneficial to me (i.e. others don't have to search first). This aligns with the style I've seen in the go ebuilds inside the tree, aswell.

games-fps/urbanterror/urbanterror-4.3.4_p20211208.ebuild
holgersson32644: If you insist on that I'll revert it, however I find that quoting style ending in the middle of a path a bit irritating and prefer to see the quotation marks at first glance. (I know there's syntax highlighting etc, but sometimes you just see raw text.)

games-fps/urbanterror/urbanterror-4.3.4_p20211208.ebuild
holgersson32644: Sure, as far as I know we allow to set custom CFLAGS etc. in Gentoo for nearly every package and the old patch I carried from the previous maintainer hasufell strips the defaults off and replaces them with a variable set by portage (or the toolchain) of the build host. Actually I forgot to search for that in upstream's Makefile, so my bug at upstream is most probably invalid, thanks for pointing into the right direction there.
I'd stick with this approach and try to remember grepping the Makefiles for filtered flags next time.

games-fps/urbanterror/urbanterror-4.3.4_p20211208.ebuild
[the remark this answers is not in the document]
Flowdalic: Fine by me.

games-fps/urbanterror/urbanterror-4.3.4_p20211208.ebuild
Flowdalic: Yeah, I am also not a fan of that particular style, as I am not aware of any advantages. But I would prefer to keep the existing ebuild where possible. That said, it is not a blocker for me.

games-fps/urbanterror/urbanterror-4.3.4_p20211208.ebuild
Flowdalic: I wonder if we should use `filter-flag -fstrict-aliasing` instead.

Flowdalic: Uh, I just discovered that we remove `-fno-strict-aliasing` with a custom gentoo patch: https://github.com/gentoo/gentoo/blob/770a83e33e69c614e533028f9c85b7165ed1f077/games-fps/urbanterror/files/urbanterror-4.3.4_p20180708-fix-build_system.patch#L16-L18

I am pretty sure that we shouldn't do that.

ionenwks: Really... kept hearing about this package and strict-aliasing in #gentoo-games but didn't think it was because it was straight up removed :eyes: Yes, I'd change the patch to keep it personally. Warnings flags can be left alone as well, this isn't -Werror.

It's good to cleanup some overaggressive flags, but not flags that are necessary to function properly (even if it's not good practice to rely on -fno-strict-aliasing)

Flowdalic:
Thanks. I've bumped urbanterror based on this PR with some fixes.

thesamesam:
I would normally merge the PR then adapt it afterwards to preserve credit. Not that I really mind contributors making style changes to games stuff as long as they're likely to continue looking after it. But just my opinion.

Flowdalic:
Fair point. I am happy to give credit where credit is due. Will try to remember to add a Co-authored-by git trailer next time. @holgersson32644 please don't feel discouraged by me bumping the package, your contributions are always welcome. :)

holgersson32644:
@Flowdalic @thesamesam: Finally I've found some time and was going to finish this PR _now_ i.e. about 2 months later. So thanks a lot for just doing the necessary work and no worries about any credits here! ;)

PR #24778, opened 2022-03-27
media-tv/v4l-dvb-saa716x: fix pkgcheck VariableScope error

madmartin:
remove use of variable 'EROOT' in 'src_compile'
EAPI 8

Closes: https://bugs.gentoo.org/836080

### DIFF
--- a/media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
+++ b/media-tv/v4l-dvb-saa716x/v4l-dvb-saa716x-0.0.1_p20170225-r5.ebuild
@@ -1,7 +1,7 @@
-# Copyright 2020 Gentoo Authors
+# Copyright 2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
 
 inherit linux-info linux-mod
 
@@ -45,6 +45,6 @@ src_compile() {
 	kernel_is le 5 0 && BUILD_PARAMS="SUBDIRS" || BUILD_PARAMS="M"
 	BUILD_PARAMS+="=${S}/linux/drivers/media/common/saa716x CONFIG_SAA716X_CORE=m \
 		CONFIG_DVB_SAA716X_FF=m CONFIG_DVB_SAA716X_BUDGET=m CONFIG_DVB_SAA716X_HYBRID=m"
-	addpredict "${EROOT}"/usr/src/linux/
+	addpredict /usr/src/linux/
 	linux-mod_src_compile
 }


juippis (commented):
Can you show build.log for this package? I'm getting permission denied errors here and there, and I imagine I'm missing some kernel configure options too (even though on dist-kernel) and getting compile errors.
```
/var/tmp/portage/media-tv/v4l-dvb-saa716x-0.0.1_p20170225-r5/work/powARman-v4l-dvb-saa716x-83f3bfd93a95/linux/drivers/media/common/saa716x/saa716x_budget.c:36:10: fatal error: mb86a16.h: No such file or directory
   36 | #include "mb86a16.h"
      |          ^~~~~~~~~~~
compilation terminated.
make[1]: *** [scripts/Makefile.build:287: /var/tmp/portage/media-tv/v4l-dvb-saa716x-0.0.1_p20170225-r5/work/powARman-v4l-dvb-saa716x-83f3bfd93a95/linux/drivers/media/common/saa716x/saa716x_budget.o] Error 1
make[1]: *** Waiting for unfinished jobs....
/var/tmp/portage/media-tv/v4l-dvb-saa716x-0.0.1_p20170225-r5/work/powARman-v4l-dvb-saa716x-83f3bfd93a95/linux/drivers/media/common/saa716x/saa716x_hybrid.c:36:10: fatal error: zl10353.h: No such file or directory
   36 | #include "zl10353.h"
      |          ^~~~~~~~~~~
compilation terminated.
make[1]: *** [scripts/Makefile.build:287: /var/tmp/portage/media-tv/v4l-dvb-saa716x-0.0.1_p20170225-r5/work/powARman-v4l-dvb-saa716x-83f3bfd93a95/linux/drivers/media/common/saa716x/saa716x_hybrid.o] Error 1
make: *** [Makefile:1846: /var/tmp/portage/media-tv/v4l-dvb-saa716x-0.0.1_p20170225-r5/work/powARman-v4l-dvb-saa716x-83f3bfd93a95/linux/drivers/media/common/saa716x] Error 2
 * ERROR: media-tv/v4l-dvb-saa716x-0.0.1_p20170225-r5::gentoo failed (compile phase):
 *   emake failed
```

thesamesam:
Ping @madmartin.

madmartin:
> Ping @madmartin.

Hi, I'm investigating...

madmartin:
@juippis @thesamesam 


> I'm getting permission denied errors here and there, and I imagine I'm missing some kernel configure options too (even though on dist-kernel) and getting compile errors.

The "permission denied" errors are harmless, they have no negative impact on the build process. I'm not sure, but they are caused by the ebuild sandbox and non-root user during the build process. The `addpredict` statement is also related to this.

The assumption for missing kernel options is basically correct, but not the first problem to solve.

The fatal build error is caused by the missing `mb86a16.h` (and others). On my systems, I never used any of the three kernels in `virtual/dist-kernel`. I use only `sys-kernel/gentoo-sources`.

The headers needed to compile the modules are taken silently from `sys-kernel/gentoo-sources`. When I test with `sys-kernel/gentoo-kernel-bin` then I can also see this errors - I assume `sys-kernel/gentoo-kernel-bin` is the only package not providing module header files.

So I add a small patch to the `Makefile` that the header files inside the module source tree should be used. 
This works for one module (`saa716x_ff`) , but not for `saa716x_budget` and `saa716x_hybrid`. I see several errors like "unknown type name xxx" and "struct yyy has no member named zzz" - I assume the source tree is inconsistent in itself, most probably because the directory with the modules of my interest have been patched to compile against current kernels.

Now I have several options:
* only compile module `saa716x_ff`,  remove `saa716x_budget` and `saa716x_hybrid`. Fine for me, but other users?
* add a blocker against `sys-kernel/gentoo-kernel-bin`
* compile  `saa716x_budget` and `saa716x_hybrid` only when compiled not against `sys-kernel/gentoo-kernel-bin`
 
Whats your opinion?

juippis (approved):
I'm gonna merge this for the CI issue, but I think what you'll want to do is depend on kernel-sources exactly. E.g., 
```
DEPEND="|| (
		sys-kernel/gentoo-sources
		sys-kernel/vanilla-sources
		sys-kernel/git-sources
		sys-kernel/mips-sources
		sys-kernel/pf-sources
		sys-kernel/rt-sources
		sys-kernel/zen-sources
		sys-kernel/raspberrypi-sources
	)"
```
IF people compile their kernels manually, they've most likely package.provided a -sources package already, and this should work for them. 

Adding a DEPEND="" block doesn't need a revbump, so if you agree you can file a new PR anytime.

PR #24780, opened 2022-03-27
sys-block/partimage: update EAPI 6 -> 8

ConiKost:
Also migrating to GLEP-81,
fixing compilation with clang
and correcting license.

Closes: https://bugs.gentoo.org/732102
Closes: https://bugs.gentoo.org/781494

@gentoo/base-system Please review, If you are okay with merge.

### DIFF
--- a/sys-block/partimage/partimage-0.6.9-r3.ebuild
+++ b/sys-block/partimage/partimage-0.6.9-r3.ebuild
@@ -1,0 +1,141 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit autotools flag-o-matic pam
+
+DESCRIPTION="Console-based application to efficiently save raw partition data to image file"
+HOMEPAGE="https://www.partimage.org/"
+SRC_URI="mirror://sourceforge/${PN}/${P}.tar.bz2"
+
+LICENSE="GPL-2+"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~ppc ~sparc ~x86"
+IUSE="nls nologin pam ssl static"
+REQUIRED_USE="static? ( !pam )"
+
+COMMON_DEPEND="
+	acct-group/partimag
+	acct-user/partimag
+"
+LIBS_DEPEND="
+	app-arch/bzip2
+	>=dev-libs/newt-0.52
+	>=sys-libs/slang-2
+	sys-libs/zlib:=
+	!nologin? ( virtual/libcrypt:= )
+	ssl? ( dev-libs/openssl:0= )
+"
+PAM_DEPEND="pam? ( sys-libs/pam )"
+RDEPEND="
+	${COMMON_DEPEND}
+	${PAM_DEPEND}
+	!static? ( ${LIBS_DEPEND} )
+"
+DEPEND="
+	${PAM_DEPEND}
+	${LIBS_DEPEND}
+"
+BDEPEND="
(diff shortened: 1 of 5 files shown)


sys-block/partimage/partimage-0.6.9-r3.ebuild
```diff
@@ -0,0 +1,151 @@
...
+		partimagesslperms || die
+		einfo "Done"
+	else
+		einfo "SSL is disabled, not building certificates"
+	fi
+}
+
+partimagesslperms() {
+	local ret=0
+	fperms 600 ${privkey} 2>/dev/null
```
floppym, line 134: Why are we throwing away the error output here? Let's get rid all of the `>/dev/null` in this function.

sys-block/partimage/partimage-0.6.9-r3.ebuild
```diff
@@ -0,0 +1,151 @@
...
+	eautoreconf
+}
+
+src_configure() {
+	# XXX: Do we still need these?
+	filter-flags -fno-exceptions
+	use ppc && append-flags -fsigned-char
+
+	local myeconfargs=(
+		$(usex pam "$(usex !static '--enable-pam' '--disable-pam')" "--disable-pam")
```
floppym, line 63: Let's replace this crazy logic `REQUIRED_USE="static? ( !pam )"`. Also adjust PAM_DEPEND above.

sys-block/partimage/partimage-0.6.9-r3.ebuild
```diff
@@ -0,0 +1,151 @@
...
+		einfo "Done"
+	else
+		einfo "SSL is disabled, not building certificates"
+	fi
+}
+
+partimagesslperms() {
+	local ret=0
+	fperms 600 ${privkey} 2>/dev/null
+	ret=$((${ret}+$?))
```
floppym, line 135: Personally, I would replace this weird exit status arithmetic with `|| die` after each fperms call.

sys-block/partimage/partimage-0.6.9-r3.ebuild
ConiKost: Done.

sys-block/partimage/partimage-0.6.9-r3.ebuild
ConiKost: Sounds good. Changed. Since it was originally so definied, I haven't changed it :-) It yes, it looks that way better.

sys-block/partimage/partimage-0.6.9-r3.ebuild
ConiKost: Indeed. It does not work. Seems, it was even broken in current stable version :-) I changed it now, so it works. Tested on my local system.

sys-block/partimage/partimage-0.6.9-r3.ebuild
```diff
@@ -0,0 +1,142 @@
...
+		insinto /etc/partimaged
+		doins "${FILESDIR}"/servercert.cnf
+	fi
+
+	fowners partimag:root /etc/partimaged/partimagedusers
+}
+
+pkg_config() {
+	if use ssl; then
+		local confdir=${EROOT}/etc/partimaged
```
floppym, line 98: EROOT may contain whitespace, so any variables derived from it (confdir, privkey, cnf cert, csr) should be quoted when they are passed to commands below.

ConiKost: Ack, done.

floppym (approved):

PR #24781, opened 2022-03-27
sys-apps/utempter: update EAPI 6 -> 8

ConiKost:
This bump also migrates to GLEP-81.

Closes: https://bugs.gentoo.org/781584

@gentoo/base-system Please review, if you are fine with merging it.

### DIFF
--- a/sys-apps/utempter/utempter-0.5.5.6-r1.ebuild
+++ b/sys-apps/utempter/utempter-0.5.5.6-r1.ebuild
@@ -1,0 +1,49 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+MY_PN="${P%.*}"
+MY_P="${MY_PN}-${PV##*.}"
+
+inherit flag-o-matic rpm toolchain-funcs
+
+DESCRIPTION="App that allows non-privileged apps to write utmp (login) info"
+HOMEPAGE="https://www.redhat.com/"
+SRC_URI="mirror://gentoo/${MY_P}.src.rpm"
+S="${WORKDIR}/${MY_PN}"
+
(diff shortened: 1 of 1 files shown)


sys-apps/utempter/utempter-0.5.5.6-r1.ebuild
```diff
@@ -0,0 +1,60 @@
...
+	local myemakeargs=(
+		LIBDIR="/usr/$(get_libdir)"
+		RPM_BUILD_ROOT="${ED}"
+	)
+
+	emake "${myemakeargs[@]}" install
+
+	dobin utmp
+
+	fowners root:utmp /usr/sbin/utempter
```
floppym, line 46: Since we are utilizing the "utmp" group in src_install, we should add acct-group/utmp to BDEPEND.

sys-apps/utempter/utempter-0.5.5.6-r1.ebuild
```diff
@@ -0,0 +1,60 @@
...
+	fperms 2755 /usr/sbin/utempter
+}
+
+pkg_postinst() {
```
floppym, line 50: I think we should drop pkg_postinst. OpenRC and systemd already set the proper permissions on these files on boot.

floppym (changes requested):

sys-apps/utempter/utempter-0.5.5.6-r1.ebuild
ConiKost: Thanks, done.

sys-apps/utempter/utempter-0.5.5.6-r1.ebuild
ConiKost: Removed, done.

floppym (approved):

PR #24791, opened 2022-03-28
Dev python/pylatex

janko-j1:


Flowdalic:
Thanks. I believe nothing here requires a revision bump:
- certainly not the EAPI bump
- bug #798381 does not affect the installed image of the package
- for adding new python versions see: https://projects.gentoo.org/python/guide/package-maintenance.html?highlight=revision#adding-new-python-implementations-to-existing-packages
-  I believe this applies to python removals too. However, I am not sure what the motivation of removing python 3.7 is: It is still in the tree, and I suppose the package works with python 3.7? Are we supposed to remove 3.7 at this point?

Flowdalic:
> So, if no revision bump is needed : should I keep the stable keywords ?

Yes, that sounds sensible to me in this case.

janko-j1:
I found why I cleaned up py3.7: it was because of the following warning during emerge 
 * pylatex-1.4.1.tar.gz BLAKE2B SHA512 size ;-) ...                                                                              [ ok ]
 * Please clean PYTHON_COMPAT of obsolete implementations:
 *   python3_7

I have a test machine on which everything went fine but I cannot install the new version on the system I use daily, I'll investigate.

juippis:
> Are we supposed to remove 3.7 at this point?

Yes, with version/revision bumps whenever possible. It's been use.masked and unavailable for a long time already.

PR #24796, opened 2022-03-28
media-libs/libvmaf: add bdep to embed models within the shared libs

AdelKS:
Devs upstream [pointed me towards](https://github.com/Netflix/vmaf/issues/998) the correct fix for bug https://bugs.gentoo.org/835471

Signed-off-by: Adel KARA SLIMANE <adel.ks@zegrapher.com>

### DIFF
--- a/media-libs/libvmaf/libvmaf-2.3.0-r1.ebuild
+++ b/media-libs/libvmaf/libvmaf-2.3.0-r1.ebuild
@@ -21,12 +21,13 @@ fi
 LICENSE="BSD-2-with-patent"
 SLOT="0"
 
-# upstream issue, see bug #835471
-RESTRICT="test"
-
 BDEPEND="
 	dev-lang/nasm
+	app-editors/vim-core
 "
+# The app-editors/vim-core dep is needed to embed models within the library
+# could be made into a useflag if someones express the need for it
+# see https://github.com/Netflix/vmaf/blob/master/libvmaf/meson_options.txt#L21
 
 RDEPEND="${BDEPEND}"
 


media-libs/libvmaf/libvmaf-2.3.0-r1.ebuild
```diff
@@ -21,11 +21,9 @@ fi
 LICENSE="BSD-2-with-patent"
 SLOT="0"
 
-# upstream issue, see bug #835471
-RESTRICT="test"
-
 BDEPEND="
 	dev-lang/nasm
+	app-editors/vim-core
```
thesamesam, line 26: Is this always needed, even without tests?

If so, yeah, the revbump is correct, as AFAIK it didn't fail if vim-core was missing, and may have misinstalled files.

AdelKS: yes it's needed to have embedded libraries, if I understand things well. It is getting installed otherwise without the models being embedded. And starting ffmpeg 5 it becomes mandatory to give the path of models if they are not embedded, if I understand things well too.

I think it's best I add a use flag that controls this toggle and have it default enabled ?

AdelKS: It's for this options to properly work https://github.com/Netflix/vmaf/blob/master/libvmaf/meson_options.txt#L21

I will add the useflag and make it default enabled, otherwise it will get disabled and I believe the test won't be run.

AdelKS: Or we just keep it this way and deem it good, actually it hurts nobody to have the models embedded in the library actually.

PR #24801, opened 2022-03-29
net-misc/icaclient: Bump to version 22.3.0.24

henning-schild:


henning-schild:
This ebuild needs some version bumping, here i even jumped over 3 other upstream releases available since 2109

PR #24804, opened 2022-03-29
dev-util/gitlab-runner: drop dependency on acct-group/gitlab-runner

Flowdalic:
There is no need that dev-util/gitlab-runner depends on
acct-group/gitlab-runner since acct-user/gitlab-runner already depends on
acct-group/gitlab-runner.


Flowdalic:
The 'gitlab-runner' group is used by the ebuild in `src_install`. And since acct-user packages only declare a RDEPEND to their acct-group packages, the gitlab-runner ebuild needs to explicitly add an DEPEND on acct-group/gitlab-runner.

PR #24805, opened 2022-03-29
Bump games-strategy/freeciv to 3.0.0

Kangie:
Bump games-strategy/freeciv to 3.0.0; drop 2.6.6.

Changes:
- Source now compressed using xz, URI updated.
- Now generates html docs for all rulesets.


Kangie:
Looking into the ebuild some more, in the `src_install` section where `tools/freeciv-manual`  had to be updated for the new default ruleset - civ2civ3.

Noting that docs should generate a readme for each ruleset and that the `freeciv-manual`  wrapper accepts arguments we should probably into what options are used upstream in future and update the ebuild - current state is no worse than 2.6.6 though.

games-strategy/freeciv/freeciv-3.0.0.ebuild
```diff
@@ -169,10 +169,14 @@ src_install() {
 			# servers, because the 'freeciv-manual' tool is then not built. Also
 			# delete freeciv-manual from the GAMES_BINDIR, because it's useless.
 			# Note: to have it localized, it should be ran from _postinst, or
-			# something like that, but then it's a PITA to avoid orphan files...
-			./tools/freeciv-manual || die
-			docinto html
-			dodoc classic*.html
+		  # something like that, but then it's a PITA to avoid orphan files...
+			# freeciv-manual only supports one ruleset argument at a time.
```
laumann, line 173: Still some whitespace issue here, it looks like.

games-strategy/freeciv/freeciv-3.0.0.ebuild
```diff
@@ -170,9 +170,13 @@ src_install() {
...
 			# Note: to have it localized, it should be ran from _postinst, or
 			# something like that, but then it's a PITA to avoid orphan files...
-			./tools/freeciv-manual || die
-			docinto html
-			dodoc classic*.html
+			# freeciv-manual only supports one ruleset argument at a time.
+			for RULESET in alien civ1 civ2 civ2civ3 classic experimental multiplayer sandbox
+			do
+				./tools/freeciv-manual -r ${RULESET} || die
+				docinto rulesets/${RULESET}
```
chewi, line 177: The game works fine for me. My only concern is with the path for these rulesets. It is conventional to put HTML files in the `html` subdirectory, so either make this `docinto html/${RULESET}` or `docinto html/rulesets/${RULESET}`. I'm assuming the game itself doesn't use these files.

Kangie: Ack. I'll update the ebuild!

chewi:
Merged. Thanks!

PR #24807, opened 2022-03-29
media-libs/libjxl: version bump to 20220329 snapshot

novomesk:
@dlan17 Please test if it works for you.


media-libs/libjxl/libjxl-0.7.0_pre20220329.ebuild
```diff
@@ -0,0 +1,87 @@
+# Copyright 2021-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+CMAKE_ECLASS=cmake
```
mattst88, line 6: Can drop this line. `cmake` is the default.

media-libs/libjxl/libjxl-0.7.0_pre20220329.ebuild
```diff
@@ -0,0 +1,87 @@
...
+	>=dev-cpp/highway-0.16.0[${MULTILIB_USEDEP}]
+	media-libs/giflib:=[${MULTILIB_USEDEP}]
+	>=media-libs/lcms-2.13:=[${MULTILIB_USEDEP}]
+	media-libs/libpng:=[${MULTILIB_USEDEP}]
+	sys-libs/zlib[${MULTILIB_USEDEP}]
+	virtual/jpeg[${MULTILIB_USEDEP}]
+	x11-misc/shared-mime-info
+	openexr? ( media-libs/openexr:= )
+"
+
```
mattst88, line 30: Drop extra newline.

media-libs/libjxl/libjxl-0.7.0_pre20220329.ebuild
```diff
@@ -0,0 +1,87 @@
...
+	if multilib_is_native_abi; then
+		if use examples; then
+			dobin "${BUILD_DIR}/jxlinfo"
+		fi
+
+		insinto /usr/share/mime/packages
+		doins -r "${S}"/plugins/mime/image-jxl.xml
+	fi
+
+	find "${D}" -name '*.a' -delete || die
```
mattst88, line 86: I don't know cmake very well, but I wonder if there's a way to avoid building the static libraries.

media-libs/libjxl/libjxl-0.7.0_pre20220329.ebuild
```diff
@@ -0,0 +1,87 @@
...
+		-DJPEGXL_ENABLE_MANPAGES=OFF
+		-DJPEGXL_ENABLE_JNI=OFF
+		-DJPEGXL_ENABLE_TCMALLOC=OFF
+	)
+
+	if multilib_is_native_abi; then
+		mycmakeargs+=(
+			-DJPEGXL_ENABLE_TOOLS=ON
+			-DJPEGXL_ENABLE_EXAMPLES=$(usex examples)
+			-DJPEGXL_ENABLE_OPENEXR=$(usex openexr)
```
mattst88, line 61: Can use `multilib_native_usex`  here and drop the if statement.

media-libs/libjxl/libjxl-0.7.0_pre20220329.ebuild
```diff
@@ -0,0 +1,87 @@
...
+		-DJPEGXL_FORCE_SYSTEM_LCMS2=ON
+		-DJPEGXL_ENABLE_DOXYGEN=OFF
+		-DJPEGXL_ENABLE_MANPAGES=OFF
+		-DJPEGXL_ENABLE_JNI=OFF
+		-DJPEGXL_ENABLE_TCMALLOC=OFF
+	)
+
+	if multilib_is_native_abi; then
+		mycmakeargs+=(
+			-DJPEGXL_ENABLE_TOOLS=ON
```
mattst88, line 59: Can use `$(multilib_is_native_abi && echo ON || echo OFF)` here and pull it out of the if statement.

media-libs/libjxl/libjxl-0.7.0_pre20220329.ebuild
novomesk: There is an option, but it seems it is only a preference.

media-libs/libjxl/libjxl-0.7.0_pre20220329.ebuild
novomesk: I know about this possibility but it I'd like to keep it in this way. It's similar like in live ebuild. It is not possible to move all options outside there, because some options exist only when other are turned on.

media-libs/libjxl/libjxl-0.7.0_pre20220329.ebuild
novomesk: I will do it.

media-libs/libjxl/libjxl-0.7.0_pre20220329.ebuild
novomesk: You are referring to the place after openexr dependency?

novomesk:
`image/jxl` mime-type is now present in shared-mime-info-2.2 so we no longer need to install it via libjxl. I modified also live ebuild accordingly.

media-libs/libjxl/libjxl-0.7.0_pre20220329.ebuild
mattst88: Works for me. Thanks!

media-libs/libjxl/libjxl-0.7.0_pre20220329.ebuild
mattst88: There's a blank line between the `"` and the `RDEPEND` assignment. (I like the `"` on its own line)

PR #24812, opened 2022-03-30
www-apps/hugo: add 0.96.0, enable tests

tastytea:
- Enable tests
- Add test-full USE-flag to pull in (lots of) optional test dependencies
- Improve unbundling patch
- Don't set buildDate, it is expected to be the time of the git revision
  <https://github.com/gohugoio/hugo/commit/64afb7c>
- Add PVR to vendorInfo to match upstreams expectations
  <https://github.com/gohugoio/hugo/commit/e8df097>


juippis (commented):
The `test-full` USE flag seems weird, especially since it's not tied to `FEATURES="test"`. Could you just do integrate it into that? Do the tests work without additional deps, but run less tests? I think you should just depend on those with `IUSE="test"`. Don't overcomplicate.

tastytea:
I got it from other ebuilds. 😊 The issue is that the dependencies (mostly pandoc) pull in over 150 additional dependencies and I think that is a bit much for a few additional tests?

The tests work, the ones with the missing dependencies are skipped.

tastytea:
I don't know how I could integrate `test-full` with `FEATURES="test"` and the ebuilds I looked at didn't do it either. I admit that it isn't very elegant. So, I guess I just add the deps as normal test deps? Or is disabling the pandoc tests preferable?

juippis:
> I don't know how I could integrate `test-full` with `FEATURES="test"` and the ebuilds I looked at didn't do it either. I admit that it isn't very elegant. So, I guess I just add the deps as normal test deps? Or is disabling the pandoc tests preferable?

If they're just skipped when a dependency is missing, I'd add the test deps as `optfeature`. Or conditional disabling via `has_version`. But both ways really work for me, just some kind of advertisement about the massive deps.

Most of the time the tests are ran by different tinderboxes and they don't work very well with this custom USE flag.

tastytea:
I added pandoc as `optfeature`. The other deps are more reasonable so I directly depend on them.

tastytea:
> This way `test-full` USE flag would do nothing unless `test` is also enabled, which should only be enabled with `FEATURES="test"`. I guess this'd work. I can make the changes if you ack, let me know what you think about it.

Yeah, looks good to me.

juippis (approved):

PR #24813, opened 2022-03-30
sys-cluster/glusterfs: add USE=nfs

jkroonza:
FUSE is giving us major problems, looking to try NFS as alternative and
this is faster path than packages nfs-ganesha which will potentially be
the next step.


jkroonza:
If someone else has use for this, please let me know, the cache coherency issues and readlink performance is even worse than fuse, but at least it doesn't quite crash as frequently, so if you don't have fast, changing data and small folders it may work better for you.

PR #24815, opened 2022-03-30
games-util/lutris: add 0.5.10_beta2

AndrewAmmerlaan:
This fixes many bugs, both downstream (missing deps) and upstream (unable to launch steam)

Bug: https://github.com/lutris/lutris/issues/4175
Bug: https://bugs.gentoo.org/835897
Bug: https://bugs.gentoo.org/822969
Bug: https://bugs.gentoo.org/836332
Bug: https://bugs.gentoo.org/820860
Signed-off-by: Andrew Ammerlaan <andrewammerlaan@gentoo.org>


ionenwks:
Feel free to do as you please with lutris without asking games@ (or even add yourself as a maintainer if you're interested, I feel it could use closer attention with these accumulating bugs, but I've never used it personally).

ionenwks:
On a side-note, `Bug:` tags? Does it not `Closes:` any of these?

AndrewAmmerlaan:
> Feel free to do as you please with lutris without asking games@ (or even add yourself as a maintainer if you're interested, I feel it could use closer attention with these accumulating bugs, but I've never used it personally).

Thanks, will do :+1: 

> On a side-note, Bug: tags? Does it not Closes: any of these?

Those bugs are fixed in the new version 0.5.10_beta2, but the current in-tree 0.5.9.1 is still affected. After removing 0.5.9.1 we can close them, but considering that this is a beta release I think we should keep 0.5.9.1 around for a while.

PR #24821, opened 2022-03-30
app-i18n/transifex-client: add new version 1.1.0

robert7k:


juippis (commented):
Do we still need to define the `EGO_SUM` in the ebuild if those modules are vendored? I'd imagine you can clean the ebuild here.

robert7k:
> Do we still need to define the `EGO_SUM` in the ebuild if those modules are vendored? I'd imagine you can clean the ebuild here.

Thanks for the hint, removed EGO_SUM.

Also updated the PR to the most recent version 1.1.0.

app-i18n/transifex-client/transifex-client-1.1.0.ebuild
```diff
@@ -0,0 +1,34 @@
...
+
+inherit go-module
+
+DESCRIPTION="The Transifex command-line client"
+HOMEPAGE="https://github.com/transifex/cli"
+
+go-module_set_globals
+
+SRC_URI="https://github.com/transifex/cli/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz
+	https://gitlab.com/api/v4/projects/35204985/packages/generic/${PN}/${PV}/${PN}-1.1-deps.tar.xz"
```
juippis, line 14: Remember you can use bash or `ver_rs` for the `${PN}-1.1-deps.tar.xz` so you don't have to hardcode the version.

app-i18n/transifex-client/transifex-client-1.1.0.ebuild
```diff
@@ -0,0 +1,34 @@
...
+S="${WORKDIR}"/cli-${PV}
+
+LICENSE="Apache-2.0 BSD BSD-2 ISC LGPL-3 MIT MPL-2.0"
+SLOT="0"
+KEYWORDS="~amd64 ~ppc64 ~x86"
+
+src_compile() {
+	emake GOFLAGS="${GOFLAGS} -ldflags=-X="github.com/transifex/cli/internal/txlib.Version=${PV}
```
juippis, line 22: I haven't learned how to, but we have this function called `ego` to build go stuff. Take a look for the next bump!

juippis (approved):
Thanks!

PR #24822, opened 2022-03-30
media-libs/openjpeg: Fix segfault, security bug 832007

laumann:
See: https://github.com/uclouvain/openjpeg/commit/0afbdcf3e6d0d2bd2e16a0c4d513ee3cf86e460d
Bug: https://bugs.gentoo.org/832007

### DIFF
--- a/media-libs/openjpeg/files/openjpeg-2.4.0-r2-fix-segfault.patch
+++ b/media-libs/openjpeg/files/openjpeg-2.4.0-r2-fix-segfault.patch
@@ -1,0 +1,17 @@
+Upstream: https://github.com/uclouvain/openjpeg/commit/0afbdcf3e6d0d2bd2e16a0c4d513ee3cf86e460d
+From: xiaoxiaoafeifei <lliangliang2007@163.com>
+Date: Wed, 14 Jul 2021 09:35:13 +0800
+Subject: [PATCH] Fix segfault in src/bin/jp2/opj_decompress.c due to
+ uninitialized pointer (fixes #1368) (#1369)
+
+--- a/src/bin/jp2/opj_decompress.c
++++ b/src/bin/jp2/opj_decompress.c
+@@ -1356,7 +1356,7 @@ int main(int argc, char **argv)
+         int it_image;
+         num_images = get_num_images(img_fol.imgdirpath);
+ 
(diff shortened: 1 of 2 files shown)


media-libs/openjpeg/files/openjpeg-2.4.0-fix-segfault.patch
```diff
@@ -0,0 +1,23 @@
...
+Date: Wed, 14 Jul 2021 09:35:13 +0800
+Subject: [PATCH] Fix segfault in src/bin/jp2/opj_decompress.c due to
+ uninitialized pointer (fixes #1368) (#1369)
+
+---
+ src/bin/jp2/opj_decompress.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/src/bin/jp2/opj_decompress.c b/src/bin/jp2/opj_decompress.c
+index 0e0287350..18ead6721 100644
```
laumann, line 12: Normally, I would scrub a patch to remove some cruft, but as I am not the author of the patch, I figured it's maybe best to leave it as-is. Let me know if I should scrub it.

laumann:
Wasn't sure if it should be an `-r2`.

laumann:
> Wasn't sure if it should be an `-r2`.

From the [devmanual](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html):

> Examples of changes that warrant a new revision are:
>  - adding a patch to fix a runtime issue,

Guess that answers that :-)

media-libs/openjpeg/files/openjpeg-2.4.0-fix-segfault.patch
thesamesam: Scrubbing won't remove From or Subject, just git noise, so it should be fine to scrub.

laumann: Scrubbed!

Some output from scrub-patch. I'll add a link.
```
QA: openjpeg-2.4.0-r2-fix-segfault.patch has generated warnings:

001: From 0afbdcf3e6d0d2bd2e16a0c4d513ee3cf86e460d Mon Sep 17 00:00:00 2001
^^^ suggest replacing line with a http link pointing the upstream commit hash
```

PR #24829, opened 2022-03-31
net-mail/automx2: Bump to version 2022.0

rseichter:
Upstream feature release. This ebuild adds Python 3.10 to PYTHON_COMPAT.


net-mail/automx2/automx2-2022.0.ebuild
```diff
@@ -0,0 +1,46 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7,8,9,10} )
```
juippis, line 6: Can we clean the unsupported implementations out?

net-mail/automx2/automx2-2022.0.ebuild
```diff
@@ -0,0 +1,46 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
```
juippis, line 4: 8? :)

net-mail/automx2/automx2-2022.0.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+DESCRIPTION="Email client autoconfiguration service"
+HOMEPAGE="https://automx.org/"
+SRC_URI="https://github.com/rseichter/automx2/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64"
+
+BDEPEND="acct-user/automx2"
```
juippis, line 18: It doesn't look like `fowners` or `chown` is called in this ebuild, so this should be in RDEPEND I reckon.

juippis (commented):
I can make the suggested changes and merge if you wish, let me know if you agree to them.

net-mail/automx2/automx2-2022.0.ebuild
rseichter: Why not. The above list is just what I use for [CI testing](https://github.com/rseichter/automx2/actions/runs/2070824930).

net-mail/automx2/automx2-2022.0.ebuild
rseichter: I haven't yet written any ebuild with V8. However, if you think it works, feel free to bump the version.

net-mail/automx2/automx2-2022.0.ebuild
rseichter: I thought that acct-user/* was always used during build time, but if you wish to move it to RDEPEND, I have no objections.

juippis (approved):
Thanks!

PR #24831, opened 2022-03-31
sci-libs/vtk: find qt binaries with qt-5.15.3

waebbl:
Use the path for qmake as a reference for finding the path to Qt binaries
with Qt 5.15.3 and no qtchooser installed.

Closes: https://bugs.gentoo.org/836343

### DIFF
--- a/sci-libs/vtk/files/vtk-9.1.0-adjust-to-find-binaries.patch
+++ b/sci-libs/vtk/files/vtk-9.1.0-adjust-to-find-binaries.patch
@@ -1,0 +1,25 @@
+From e59b8ff7b83cd6a58c226cb4f5d9661bcb29002b Mon Sep 17 00:00:00 2001
+From: Bernd Waibel <waebbl-gentoo@posteo.net>
+Date: Thu, 31 Mar 2022 08:05:17 +0200
+Subject: [PATCH] adjust to find binaries
+
+Add compatibility for qt-5.15.3 with no qtchooser installed. QtCore exports
+Qt5::{qmake,moc,rcc} which we use to get the path for binaries.
+
+Signed-off-by: Bernd Waibel <waebbl-gentoo@posteo.net>
+--- a/GUISupport/QtQuick/qml/CMakeLists.txt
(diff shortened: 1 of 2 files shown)


a17r:
This is not Gentoo specific at all, and should be put to upstream.

waebbl:
> This is not Gentoo specific at all, and should be put to upstream.

You sure? I thought about not being Gentoo specific as well, but didn't found other distros having any issues or patches with building vtk-9.1 with qt-5.15.3

a17r:
If it relies on PATH and distro-specific guesses, it ain't right, we've fixed countless packages like that. Using the cmake provided target for qmake is the right thing to do.

a17r:
PS: This has nothing to do with Qt 5.15.3 itself - it is just the version we chose to change our packaging of it.

waebbl:
So other distro probably have all the qt5 binaries in their path. ARCH for example has the qmlplugindump in `/usr/bin`. Wouldn't having the binaries in `/usr/$(get_libdir)/qt5/bin` in PATH, eventually with an eselect module or a `5` suffix for compatibility with Qt6 be a better way? If other distros do have these binaries in PATH, wouldn't it qualify it for Gentoo specific again?

I'd guess, upstream suspects them to be available in PATH.

a17r:
No. We can't depend on PATH, packages must not depend on eselect for dependencies.

Upstreams guess a lot, and that's when distro people help them out.

waebbl:
Ok, I'm gonna change the PR and report upstream.

waebbl:
Upstream report at https://gitlab.kitware.com/vtk/vtk/-/issues/18502

a17r (approved):

a17r:
Thanks for filing upstream.

Again, a build fix only. Will drop the unnecessary revbump on merge.

waebbl:
> Thanks for filing upstream.
> 
> Again, a build fix only. Will drop the unnecessary revbump on merge.

Ah yes, didn't catched this. Thanks!

